PR #12283, opened 2020-09-16
Rework snapshot download logic to be more forgiving when ` --expected-shred-version` is not provided

mvines:
Follow-up to #12131

### DIFF
--- a/validator/src/main.rs
+++ b/validator/src/main.rs
@@ -81,34 +81,6 @@ fn hash_validator(hash: String) -> Result<(), String> {
         .map_err(|e| format!("{:?}", e))
 }
 
-fn get_shred_rpc_peers(
-    cluster_info: &ClusterInfo,
-    expected_shred_version: Option<u16>,
-) -> Vec<ContactInfo> {
-    let rpc_peers = cluster_info.all_rpc_peers();
-    match expected_shred_version {
-        Some(expected_shred_version) => {
-            // Filter out rpc peers that don't match the expected shred version
-            rpc_peers
-                .into_iter()
-                .filter(|contact_info| contact_info.shred_version == expected_shred_version)
-                .collect::<Vec<_>>()
-        }
-        None => {
-            if !rpc_peers
-                .iter()
-                .all(|contact_info| contact_info.shred_version == rpc_peers[0].shred_version)
-            {
-                eprintln!(
-                        "Multiple shred versions observed in gossip.  Restart with --expected-shred-version"
-                    );
-                exit(1);
-            }
-            rpc_peers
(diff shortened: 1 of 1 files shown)


validator/src/main.rs
```diff
@@ -168,21 +140,48 @@ fn start_gossip_node(
...
     loop {
-        info!(
-            "Searching for an RPC service, shred version={:?}...",
-            validator_config.expected_shred_version
-        );
         sleep(Duration::from_secs(1));
         info!("\n{}", cluster_info.contact_info_trace());
 
-        let rpc_peers = get_shred_rpc_peers(&cluster_info, validator_config.expected_shred_version);
+        let shred_version = validator_config
```
mvines, line 154: If `--expected-shred-version` is not provided, grab our current shred version from cluster info.  This will be set to non-zero shortly after receiving ContactInfo for our entrypoint.

ref: https://github.com/solana-labs/solana/blob/c8f03c7f6de9c334501f2e5541c1a2e501a8f59d/core/src/cluster_info.rs#L1492-L1493

validator/src/main.rs
```diff
@@ -168,21 +140,48 @@ fn start_gossip_node(
...
+            .expected_shred_version
+            .unwrap_or_else(|| cluster_info.my_shred_version());
+        if shred_version == 0 {
+            if let Some(entrypoint) =
+                cluster_info.lookup_contact_info_by_gossip_addr(entrypoint_gossip)
+            {
+                if entrypoint.shred_version == 0 {
```
mvines, line 161: If we're 0 and the entrypoint is also 0 (ie, a `solana-gossip spy` without a `--shred-version`), we need user input to proceed.  Abort the process

validator/src/main.rs
```diff
@@ -81,34 +81,6 @@ fn hash_validator(hash: String) -> Result<(), String> {
...
-    match expected_shred_version {
-        Some(expected_shred_version) => {
-            // Filter out rpc peers that don't match the expected shred version
-            rpc_peers
-                .into_iter()
-                .filter(|contact_info| contact_info.shred_version == expected_shred_version)
-                .collect::<Vec<_>>()
-        }
-        None => {
-            if !rpc_peers
```
mvines, line 98: This logic is too heavy handed now that the entrypoint advertises a proper shred version.

mvines:
(adding automerge so that hopefully this PR doesn't get blocked on me while I'm AFK in a bit.  But @t-nelson if you see anything that needs adjustment please remove automerge and I'll fix it up)

t-nelson (approved):
LGTM!

PR #12290, opened 2020-09-16
Remove client resends

### LINKED ISSUE #12280: Some cli transactions succeed even though they exit with simulation error
#### Problem
Twice recently transactions appeared to go through even though solana-cli exited with a simulate_transaction error:
```
$ solana stake-set-lockup ...
Error: rpc request error: RPC Error response: {"code":-32002,"message":"RPC node is unhealthy, unable to simulate transaction"}
```

```
$ solana pay 2x2nhDuwHptNE3j9Z2hhNdyeQffxk67XZRKE5aXXXXXX 0.001
Error: rpc request error: RPC Error response: {"code":-32002,"message":"Transaction simulation failed: TransactionError::DuplicateSignature"}
```

#### Proposed Solution
Identify bug and fix

CriesofCarrots:
#### Problem
RpcClient `send_transaction` methods contain a resend mechanism that triggers after approximately 8 secs of null signature statuses. Solana api nodes now run a SendTransactionService that handles retries of transactions received over RPC. This means that if a transaction is slow to land for some reason, the client may resend a transaction that is already in the SendTransactionService queue, and may receive a simulation error on the second send even though the original send actually succeeds.

#### Summary of Changes
- Remove send retries from rpc_client methods
- Retry status queries until blockhash expires

Fixes #12280

### DIFF
--- a/client/src/rpc_client.rs
+++ b/client/src/rpc_client.rs
@@ -412,12 +412,18 @@ impl RpcClient {
         transaction: &Transaction,
     ) -> ClientResult<Signature> {
         let signature = self.send_transaction(transaction)?;
-        let mut status_retries = 15;
+        let recent_blockhash = transaction.message.recent_blockhash;
         let status = loop {
             let status = self.get_signature_status(&signature)?;
             if status.is_none() {
-                status_retries -= 1;
-                if status_retries == 0 {
+                if self
+                    .get_fee_calculator_for_blockhash_with_commitment(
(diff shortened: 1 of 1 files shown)


CriesofCarrots:
@mvines , let me know what you think about the blockhash-expiration change. It could make the client sit for a lot longer, but it provides confidence that the transaction will not actually succeed after the client has given up.

mvines (approved):

Issue #11753, opened 2020-08-21
RPC: Cross program instruction details are not available


jstarry:
#### Problem
It's not possible to determine which CPI's were run during transaction execution from the RPC API. This is because we do not currently do not track this info outside of the message processor and do not store that information in RocksDB / BigTable.

#### Proposed Solution
Extract cross program instruction inputs, store them in transaction meta in rocks / bigtable, and then make this available through RPC `getConfirmedTransaction` / `getConfirmedBlock`

PR #12311, opened 2020-09-17
Record and store invoked instructions in transaction meta

jstarry:
#### Problem
It's not possible to determine which CPI's were run during transaction execution. This is because we do not currently do not track this info outside of the message processor and do not store that information in RocksDB / BigTable.

#### Summary of Changes
- Record each invoked instruction when `--enable-cpi-instruction-recording` flag is present
- Store compiled invoked instructions in transaction meta in rocksdb and bigtable
- Use `Option` to denote the difference between 0 invoked instructions and missing info

Fixes https://github.com/solana-labs/solana/issues/11753

### DIFF
--- a/docs/src/apps/jsonrpc-api.md
+++ b/docs/src/apps/jsonrpc-api.md
@@ -332,7 +332,7 @@ The result field will be an object with the following fields:
       - `fee: <u64>` - fee this transaction was charged, as u64 integer
       - `preBalances: <array>` - array of u64 account balances from before the transaction was processed
       - `postBalances: <array>` - array of u64 account balances after the transaction was processed
-      - `invoked: <array|undefined>` - List of [invoked instructions](#invoked-instructions-structure) or omitted if invoked instruction recording was not yet enabled during this transaction
+      - `innerInstructions: <array|undefined>` - List of [inner instructions](#inner-instructions-structure) or omitted if inner instruction recording was not yet enabled during this transaction
       - DEPRECATED: `status: <object>` - Transaction status
         - `"Ok": <null>` - Transaction was successful
         - `"Err": <ERR>` - Transaction failed with TransactionError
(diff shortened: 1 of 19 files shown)


jstarry:
Here's what the parsed json response looks like for a token swap (`8MGih...`) transaction

```json
{
  "jsonrpc": "2.0",
  "result": {
    "meta": {
      "err": null,
      "fee": 5000,
      "invoked": [
        {
          "index": 0,
          "instructions": [
            {
              "parsed": {
                "info": {
                  "amount": "100",
                  "authority": "CWrYspp3frz8NGduEwCiHDomBqPT866WzVFD8c7C8ZEX",
                  "destination": "BQnvHCVvKGEtdSdkB8QwEUKiFyQPnKFH3URbeQDCAxib",
                  "source": "4BAgTr94Kvr6eKjrKB1wuBogfgB6GiCsaVxJHhivcVoo"
                },
                "type": "transfer"
              },
              "program": "spl-token",
              "programId": "TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA"
            },
            {
              "parsed": {
                "info": {
                  "amount": "69",
                  "authority": "CWrYspp3frz8NGduEwCiHDomBqPT866WzVFD8c7C8ZEX",
                  "destination": "29YhBxg8HQVzqeFCv79N6mRM4dF6CJJaCthttHVutgcr",
                  "source": "3jaqTySyWs8h2fL3q2fEXRDWGRZAWSVXUSAG18zS2V4n"
                },
                "type": "transfer"
              },
              "program": "spl-token",
              "programId": "TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA"
            }
          ]
        }
      ],
      "postBalances": [...],
      "preBalances": [...],
      "status": {
        "Ok": null
      }
    },
    "slot": 2659,
    "transaction": {
      "message": {
        "accountKeys": [...],
        "instructions": [
          {
            "accounts": [...],
            "data": "22YeTxNWbUF9",
            "programId": "8MGih1VozWcQnkWoJJywviTonQBtUmMST3RiS19Ehiu7"
          }
        ],
        "recentBlockhash": "4S5GAfmoiJL8AEXT93xTvbeLDWd8yF45QquwApXVZJcP"
      },
      "signatures": [
        "5ue8BoHqGQg76pYukHA94LXK1GySZVpHwENmckQNLPv7stWokLnYHqeECs4DY29Es2THQqBhv5DMz4ZyKBgTBs5D"
      ]
    }
  },
  "id": "bfd582c7-9d87-4b88-a09f-b1d05c512c9c"
}
```

jstarry:
Note: this PR currently includes a commit from https://github.com/solana-labs/solana/pull/12404 in order to parallelize reviews 😬   feel free to skip that one

jstarry:
Just realized that `verify_instruction` doesn't restrict instruction data size so this could open an attack vector where instructions are invoked with large amounts of data and cause OOM's

jstarry:
I've disabled invoked instruction recording by default to address the OOM concern

jstarry:
Going to fix conflicts and rebase to remove the extra commit from #12404

mvines:
> Just realized that `verify_instruction` doesn't restrict instruction data size so this could open an attack vector where instructions are invoked with large amounts of data and cause OOM's

This seems like a more general problem.  For ease of deployment I'd prefer to not have a `--enable-cpi-instruction-recording` flag.   Maybe this issue can be addressed as a follow-up.

jstarry:
> This seems like a more general problem. For ease of deployment I'd prefer to not have a --enable-cpi-instruction-recording flag. Maybe this issue can be addressed as a follow-up.

Sounds good, I'll write up a follow up issue

mvines:
> Going to fix conflicts and rebase to remove the extra commit from #12404

I'll just wait for the rebase, that'll make reviewing way easier :)

jstarry:
@mvines rebased and checks are passing, should be safe to dive in!

docs/src/apps/jsonrpc-api.md
```diff
@@ -332,6 +332,7 @@ The result field will be an object with the following fields:
       - `fee: <u64>` - fee this transaction was charged, as u64 integer
       - `preBalances: <array>` - array of u64 account balances from before the transaction was processed
       - `postBalances: <array>` - array of u64 account balances after the transaction was processed
+      - `invoked: <array|undefined>` - List of [invoked instructions](#invoked-instructions-structure) or omitted if invoked instruction recording was not yet enabled during this transaction
```
mvines, line 335: To me `invoked` alone doesn't mean too much.  WDYT about `invokedInstructions`?   Although normal instructions are invoked too.  `innerInstructions`?

jstarry: Ah, I like inner!

runtime/src/message_processor.rs
```diff
@@ -714,6 +725,7 @@ impl MessageProcessor {
     /// Process a message.
     /// This method calls each instruction in the message over the set of loaded Accounts
     /// The accounts are committed back to the bank only if every instruction succeeds
+    #[allow(clippy::too_many_arguments)]
```
mvines, line 728: 🎰

jstarry: 🙈

mvines (approved):
Yeah, "inner instructions" looks great in code.  PR seems gtg to me

mvines:
(removed v1.2 label, I think we can avoid that extra backport)

oJshua:
@jstarry thanks for the reference on this; highly relevant to program logs. Looks good.

programs/bpf/c/src/invoke/invoke.c
```diff
@@ -251,9 +251,9 @@ extern uint64_t entrypoint(const uint8_t *input) {
     sol_assert(SUCCESS ==
                sol_invoke(&instruction, accounts, SOL_ARRAY_SIZE(accounts)));
 
+    // Signer privilege escalation will always fail the whole transaction
     instruction.accounts[0].is_signer = true;
-    sol_assert(SUCCESS !=
-               sol_invoke(&instruction, accounts, SOL_ARRAY_SIZE(accounts)));
+    sol_invoke(&instruction, accounts, SOL_ARRAY_SIZE(accounts));
```
jackcmay, line 256: Why was the assert removed?

jackcmay: Ah, moved out to `tests/program.rs`?

runtime/src/message_processor.rs
```diff
@@ -722,10 +734,16 @@ impl MessageProcessor {
         rent_collector: &RentCollector,
         log_collector: Option<Rc<LogCollector>>,
         executors: Rc<RefCell<Executors>>,
+        mut instruction_recorders: Option<&mut Vec<InstructionRecorder>>,
         cluster_type: ClusterType,
         epoch: Epoch,
     ) -> Result<(), TransactionError> {
         for (instruction_index, instruction) in message.instructions.iter().enumerate() {
+            let instruction_recorder = instruction_recorders.as_mut().map(|recorders| {
```
jackcmay, line 742: could these be constructed as a single shot earlier (in the bank) and then looked up here via `instruction_index`?

programs/bpf/c/src/invoke/invoke.c
jstarry: Not exactly. The assert wasn't doing anything because the invalid invoked instruction causes an error in message processing and control never returns to the caller, as I understand it.

runtime/src/message_processor.rs
jstarry: Ah yeah, that's nicer. PR here: https://github.com/solana-labs/solana/pull/12467

PR #11727, opened 2020-08-20
Squash supermajority root on blockstore replay at startup

carllin:
#### Problem
After long periods of no progress, a single node can have a long ledger it needs to replay, which is prohibitively taxing on memory and compute

#### Summary of Changes
Squash banks that supermajority has rooted on replay at startup

Fixes #

PR #12319, opened 2020-09-17
Ledger replay supermajority squash doesn't handle duplicate roots

carllin:
#### Problem
Detection of supermajority root on ledger replay does not sum the stake of vote accounts with the same root. Thanks @ryoqun for pointing this out: https://github.com/solana-labs/solana/pull/11727#discussion_r489975007

#### Summary of Changes
Fix and add test

Fixes #


ryoqun (approved):
LGTM! Thanks for fixing this quickly!

PR #12320, opened 2020-09-17
Give the duplicate send_transaction_service a different thread name

mvines:
There are two SendTransactionServices (🤕) and they both use the name thread name, so we can't tell which one (both?!) is misbehaving at https://github.com/solana-labs/solana/issues/12317


mvines:
Obsoleted by https://github.com/solana-labs/solana/pull/12333

PR #12340, opened 2020-09-18
validator/ cleanup

mvines:
`validator/src/main.rs` is messy, `core/src/validator.rs` is overexposed 

Rearrange some code to make things a little tidier.


mvines:
(adding v1.3 since there's a bunch of low-risk churn here and I don't want to deal merge conflicts for a couple months)

PR #12346, opened 2020-09-18
Add get_token_account methods

CriesofCarrots:
#### Problem
The spl-token-cli needs a quick way to get token account info beyond just the balance, cc https://github.com/solana-labs/solana-program-library/issues/467

#### Summary of Changes
Add RpcClient::get_token_account methods that return a parsed TokenAccountType. This allows for easy client handling of SPL token Accounts, Mints, or Multisigs

Usage for the issue above would look like:

```
if let TokenAccountType::Account(token_account) = config
  .rpc_client
  .get_token_account_with_commitment(&address, config.commitment_config)?
  .value {
    // print all fields, etc
  }
```

### DIFF
--- a/client/src/rpc_client.rs
+++ b/client/src/rpc_client.rs
@@ -15,7 +15,10 @@ use bincode::serialize;
 use indicatif::{ProgressBar, ProgressStyle};
 use log::*;
 use serde_json::{json, Value};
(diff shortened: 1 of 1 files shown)


t-nelson (approved):
`RpcError::ForUser` still makes me sad.  LGTM otherwise though!

PR #10718, opened 2020-06-19
Persistent tower

ryoqun:
#### Problem

tower isn't persisted.

#### Summary of Changes

Persist it

#### Todo

- [x] Address any remaining review comments from #7436 #9902 

Fixes #6936

PR #12350, opened 2020-09-19
Follow up to persistent tower with tests and API cleaning

ryoqun:
#### Problem

Tower::root() can get away of `Option<_>`.

#### Summary of Changes

Let's simplify code and reduce combinatorial complexes for humans.

And some minor follow up fixes.

Also, adds long-awaited interesting test: https://github.com/solana-labs/solana/pull/10718#discussion_r491253093

#### Context

follow up #10718

### DIFF
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -1174,15 +1174,12 @@ pub fn reconcile_blockstore_roots_with_tower(
                 ),
             })
             .collect();
-        // or check last_root or 0?
-        //assert!(
-        //    !new_roots.is_empty(),
-        //    "at least 1 parent slot must be found"
-        //);
+        assert!(
+            !new_roots.is_empty(),
+            "at least 1 parent slot must be found"
+        );
 
-        if !new_roots.is_empty() {
-            blockstore.set_roots(&new_roots)?
-        }
+        blockstore.set_roots(&new_roots)?
     }
     Ok(())
 }
@@ -2704,7 +2701,6 @@ pub mod test {
     }
 
     #[test]
-    #[ignore]
     #[should_panic(expected = "at least 1 parent slot must be found")]
     fn test_reconcile_blockstore_roots_with_tower_panic_no_parent() {
         solana_logger::setup();
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -387,17 +387,14 @@ impl Tower {
     pub fn record_bank_vote(&mut self, vote: Vote) -> Option<Slot> {
         let slot = vote.last_voted_slot().unwrap_or(0);
         trace!("{} record_vote for {}", self.node_pubkey, slot);
-        let root_slot = self.root();
+        let old_root = self.root();
         self.lockouts.process_vote_unchecked(&vote);
         self.last_vote = vote;
+        let new_root = self.root();
 
-        datapoint_info!(
-            "tower-vote",
-            ("latest", slot, i64),
-            ("root", self.root(), i64)
-        );
-        if root_slot != self.root() {
-            Some(self.root())
+        datapoint_info!("tower-vote", ("latest", slot, i64), ("root", new_root, i64));
+        if old_root != new_root {
+            Some(new_root)
         } else {
             None
         }
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -2704,6 +2704,7 @@ pub mod test {
     }
 
     #[test]
+    #[ignore]
     #[should_panic(expected = "at least 1 parent slot must be found")]
     fn test_reconcile_blockstore_roots_with_tower_panic_no_parent() {
         solana_logger::setup();
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -450,10 +450,6 @@ impl Tower {
         self.lockouts.root_slot.unwrap()
     }
 
-    pub fn node_pubkey(&self) -> Pubkey {
-        self.node_pubkey
-    }
-
     // a slot is recent if it's newer than the last vote we have
     pub fn is_recent(&self, slot: Slot) -> bool {
         if let Some(last_voted_slot) = self.lockouts.last_voted_slot() {
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -387,17 +387,17 @@ impl Tower {
     pub fn record_bank_vote(&mut self, vote: Vote) -> Option<Slot> {
         let slot = vote.last_voted_slot().unwrap_or(0);
         trace!("{} record_vote for {}", self.node_pubkey, slot);
-        let root_slot = self.lockouts.root_slot;
+        let root_slot = self.root();
         self.lockouts.process_vote_unchecked(&vote);
         self.last_vote = vote;
 
         datapoint_info!(
             "tower-vote",
             ("latest", slot, i64),
-            ("root", self.lockouts.root_slot.unwrap_or(0), i64)
+            ("root", self.root(), i64)
         );
-        if root_slot != self.lockouts.root_slot {
-            Some(self.lockouts.root_slot.unwrap())
+        if root_slot != self.root() {
+            Some(self.root())
         } else {
             None
         }
@@ -446,8 +446,8 @@ impl Tower {
     // which establishes the origin of trust (i.e. root) whether booting from genesis (slot 0) or
     // snapshot (slot N). In other words, there should be no possibility a Tower doesn't have
     // root, unlike young vote accounts.
-    pub fn root(&self) -> Option<Slot> {
-        self.lockouts.root_slot
+    pub fn root(&self) -> Slot {
+        self.lockouts.root_slot.unwrap()
     }
 
     // a slot is recent if it's newer than the last vote we have
@@ -514,7 +514,7 @@ impl Tower {
     ) -> SwitchForkDecision {
         self.last_voted_slot()
             .map(|last_voted_slot| {
-                let root = self.lockouts.root_slot.unwrap_or(0);
+                let root = self.root();
                 let empty_ancestors = HashSet::default();
 
                 let last_vote_ancestors =
@@ -837,8 +837,7 @@ impl Tower {
         slot_history: &SlotHistory,
     ) -> Result<Self> {
         // sanity assertions for roots
-        assert!(self.root().is_some());
-        let tower_root = self.root().unwrap();
+        let tower_root = self.root();
         info!(
             "adjusting lockouts (after replay up to {}): {:?} tower root: {}",
             replayed_root,
@@ -1160,26 +1159,28 @@ pub fn reconcile_blockstore_roots_with_tower(
     tower: &Tower,
     blockstore: &Blockstore,
 ) -> blockstore_db::Result<()> {
-    if let Some(tower_root) = tower.root() {
-        let last_blockstore_root = blockstore.last_root();
-        if last_blockstore_root < tower_root {
-            // Ensure tower_root itself to exist and be marked as rooted in the blockstore
-            // in addition to its ancestors.
-            let new_roots: Vec<_> = AncestorIterator::new_inclusive(tower_root, &blockstore)
-                .take_while(|current| match current.cmp(&last_blockstore_root) {
-                    Ordering::Greater => true,
-                    Ordering::Equal => false,
-                    Ordering::Less => panic!(
-                        "couldn't find a last_blockstore_root upwards from: {}!?",
-                        tower_root
-                    ),
-                })
-                .collect();
-            assert!(
-                !new_roots.is_empty(),
-                "at least 1 parent slot must be found"
-            );
+    let tower_root = tower.root();
+    let last_blockstore_root = blockstore.last_root();
+    if last_blockstore_root < tower_root {
+        // Ensure tower_root itself to exist and be marked as rooted in the blockstore
+        // in addition to its ancestors.
+        let new_roots: Vec<_> = AncestorIterator::new_inclusive(tower_root, &blockstore)
+            .take_while(|current| match current.cmp(&last_blockstore_root) {
+                Ordering::Greater => true,
+                Ordering::Equal => false,
+                Ordering::Less => panic!(
+                    "couldn't find a last_blockstore_root upwards from: {}!?",
+                    tower_root
+                ),
+            })
+            .collect();
+        // or check last_root or 0?
+        //assert!(
+        //    !new_roots.is_empty(),
+        //    "at least 1 parent slot must be found"
+        //);
 
+        if !new_roots.is_empty() {
             blockstore.set_roots(&new_roots)?
         }
     }
@@ -2744,13 +2745,13 @@ pub mod test {
             .unwrap();
 
         assert_eq!(tower.voted_slots(), vec![2, 3]);
-        assert_eq!(tower.root(), Some(replayed_root_slot));
+        assert_eq!(tower.root(), replayed_root_slot);
 
         tower = tower
             .adjust_lockouts_after_replay(replayed_root_slot, &slot_history)
             .unwrap();
         assert_eq!(tower.voted_slots(), vec![2, 3]);
-        assert_eq!(tower.root(), Some(replayed_root_slot));
+        assert_eq!(tower.root(), replayed_root_slot);
     }
 
     #[test]
@@ -2772,7 +2773,7 @@ pub mod test {
             .unwrap();
 
         assert_eq!(tower.voted_slots(), vec![2, 3]);
-        assert_eq!(tower.root(), Some(replayed_root_slot));
+        assert_eq!(tower.root(), replayed_root_slot);
     }
 
     #[test]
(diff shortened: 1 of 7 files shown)


ryoqun:
I already want #12739  ;)
```
error[E0277]: arrays only have std trait implementations for lengths 0..=32
   --> sdk/src/signature.rs:51:22
    |
51  |         bs58::encode(&self.0.to_bytes()).into_string()
    |                      ^^^^^^^^^^^^^^^^^^ the trait `std::array::LengthAtMost32` is not implemented for `[u8; 64]`
    |
   ::: /home/.cargo/registry/src/github.com-1ecc6299db9ec823/bs58-0.3.1/src/lib.rs:207:18
    |
207 | pub fn encode<I: AsRef<[u8]>>(input: I) -> encode::EncodeBuilder<'static, I> {
    |                  ----------- required by this bound in `bs58::encode`
    |
    = note: required because of the requirements on the impl of `std::convert::AsRef<[u8]>` for `[u8; 64]`
    = note: required because of the requirements on the impl of `std::convert::AsRef<[u8]>` for `&[u8; 64]`
```

core/src/consensus.rs
```diff
@@ -532,12 +546,22 @@ impl Tower {
...
                     // If the `switch_slot is a descendant of the last vote,
                     // no switching proof is necessary
-                    return SwitchForkDecision::NoSwitch;
+                    return SwitchForkDecision::SameFork;
                 }
 
-                // Should never consider switching to an ancestor
-                // of your last vote
-                assert!(!last_vote_ancestors.contains(&switch_slot));
+                // Generally, should never consider switching to an ancestor of your last vote
```
ryoqun, line 552: this should go to #12671

core/src/consensus.rs
```diff
@@ -520,9 +527,16 @@ impl Tower {
                             // all of them.
                             panic!("no ancestors found with slot: {}", last_voted_slot);
                         } else {
-                            // bank_forks doesn't have corresponding data for the stray restored last vote,
-                            // meaning some inconsistency between saved tower and ledger.
-                            // (newer snapshot, or only a saved tower is moved over to new setup?)
+                            // compare slots not to error! just because of newer snapshots
+                            if self.last_switch_threshold_check.is_none() && switch_slot < last_voted_slot {
```
ryoqun, line 531: define should_warn_on_first_switch_check()

core/src/consensus.rs
```diff
@@ -30,11 +30,11 @@ use std::{
...
 use thiserror::Error;
 
-#[derive(PartialEq, Clone, Debug)]
+#[derive(PartialEq, Clone, Debug, AbiExample)]
 pub enum SwitchForkDecision {
     SwitchProof(Hash),
-    NoSwitch,
-    FailedSwitchThreshold,
+    SameFork,
+    FailedSwitchThreshold(u64, u64),
```
ryoqun, line 37: extract this into separate pr.

core/src/replay_stage.rs
```diff
@@ -1552,7 +1552,19 @@ impl ReplayStage {
                     .epoch_vote_accounts(heaviest_bank.epoch())
                     .expect("Bank epoch vote accounts must contain entry for the bank's own epoch"),
             );
-            if switch_fork_decision == SwitchForkDecision::FailedSwitchThreshold {
+            let a = Some((heaviest_bank.slot(), switch_fork_decision.clone()));
+            if tower.last_switch_threshold_check != a {
```
ryoqun, line 1556: move to tower

core/src/replay_stage.rs
```diff
@@ -1598,11 +1610,13 @@ impl ReplayStage {
                 failure_reasons.push(HeaviestForkFailures::NoPropagatedConfirmation(bank.slot()));
             }
 
-            if !is_locked_out
-                && vote_threshold
-                && propagation_confirmed
-                && switch_fork_decision != SwitchForkDecision::FailedSwitchThreshold
-            {
+            if let SwitchForkDecision::FailedSwitchThreshold(_, _) = switch_fork_decision {
```
ryoqun, line 1613: deduplicate with the following `else` clause.

mvines:
@ryoqun - should we bring this into v1.4 as well?  I'm thinking why not, since v1.4 will live for a couple months

ryoqun:
> @ryoqun - should we bring this into v1.4 as well? I'm thinking why not, since v1.4 will live for a couple months

Yeah, definitely, I'll do. :)

ryoqun:
@carllin Could you review this?

local-cluster/tests/local_cluster.rs
```diff
@@ -1680,7 +1601,282 @@ fn test_validator_saves_tower() {
...
+//    generates a new fork starting at slot `S3` because `C` has more stake than `A`
+//    so `A` will eventually pick the fork `C` is on.
+//
+//    Furthermore `B`'s vote on `S3` is not observable because there are no
+//    descendants of slot `S3`, so that fork will not be chosen over `C`'s fork
+//
+// With the persisted tower:
+//    `A` should not be able to generate a switching proof.
+//
+fn do_test_optimistic_confirmation_violation_with_or_without_tower(with_tower: bool) {
```
carllin, line 1662: wow, this looked really involved/painful 🤕. Thanks for implementing that hacky suggestion for a test, hopefully this makes us feel better about the optimistic confirmation + saved tower interaction!

local-cluster/tests/local_cluster.rs
```diff
@@ -1431,12 +1438,6 @@ fn test_optimistic_confirmation_violation_with_no_tower() {
             prev_voted_slot
         );
         blockstore.set_dead_slot(prev_voted_slot).unwrap();
-
-        std::fs::remove_file(Tower::get_filename(
```
carllin, line 1435: Is this no longer necessary?

local-cluster/tests/local_cluster.rs
```diff
@@ -1680,7 +1601,282 @@ fn test_validator_saves_tower() {
...
+        std::fs::remove_dir_all(&validator_c_info.info.ledger_path).unwrap();
+        let mut opt = fs_extra::dir::CopyOptions::new();
+        opt.copy_inside = true;
+        fs_extra::dir::copy(
+            &validator_a_info.info.ledger_path,
+            &validator_c_info.info.ledger_path,
+            &opt,
+        )
+        .unwrap();
+        std::fs::remove_file(Tower::get_filename(
```
carllin, line 1746: Maybe a comment about what this is deleting? (the tower?)

local-cluster/tests/local_cluster.rs
```diff
@@ -1680,7 +1601,282 @@ fn test_validator_saves_tower() {
...
+            &validator_c_info.info.ledger_path,
+            &validator_a_pubkey,
+        ))
+        .unwrap();
+
+        let blockstore = Blockstore::open_with_access_type(
```
carllin, line 1752: Maybe we can extract this open -> purge logic into a separate function because the same logic is reused below?

local-cluster/tests/local_cluster.rs
```diff
@@ -1680,7 +1601,282 @@ fn test_validator_saves_tower() {
...
+        if highest_bank >= next_slot_on_a {
+            break;
+        }
+    }
+    let validator_a_info = cluster.exit_node(&validator_a_pubkey);
+    let client = cluster.get_validator_client(&validator_b_pubkey).unwrap();
+    loop {
+        sleep(Duration::from_millis(100));
+        let highest_bank = client
+            .get_slot_with_commitment(CommitmentConfig::recent())
```
carllin, line 1724: Is there a way to assert here that validator `b` actually voted on `next_slot_on_a` (maybe a variation of `last_vote_in_tower()` that checks `tower_contains_vote()`? 

I think CommitmentConfig::recent() implies the validator voted

local-cluster/tests/local_cluster.rs
```diff
@@ -1680,7 +1601,282 @@ fn test_validator_saves_tower() {
...
+        );
+        if !with_tower {
+            info!("Removing tower!");
+            std::fs::remove_file(Tower::get_filename(
+                &validator_a_info.info.ledger_path,
+                &validator_a_pubkey,
+            ))
+            .unwrap();
+
+            // For some reason, fork_choice always selects slot 27 over votes_on_c_fork.
```
carllin, line 1793: @ryoqun I think I see what's happening.

 If you don't delete slot 27 from the ledger, then the validator `A` will immediately vote for 27 on restart, because it hasn't gotten the heavier fork from validator `C` yet. Then it will be stuck on `27` unable to switch because `C` doesn't have enough stake to generate a switching proof

local-cluster/tests/local_cluster.rs
ryoqun: @carllin Yeah. As I figured out this in hard way, saved tower doesn't protect us from violating opt. conf., _if voted slot is marked dead, forcibly_. So, as you guessed, i'm effectively making this test harder to pass (=actually violate) by not removing the saved tower and indeed this test still passes. :)

https://github.com/solana-labs/solana/pull/10718#discussion_r490104064:
> Well, I found it's very hard to make a validator violate optimistic confirmation _if and only if tower is removed_.
> The hard part is that marking a slot as dead makes a validator to violate the optimistic conf. **even if tower is persited**. And eventually, the cluster makes new roots.
> ...

local-cluster/tests/local_cluster.rs
ryoqun: Yeah, to be frank, writing this was very hard and needed more logging improvement resulting in #12875. ;) Well, I'm actually surprised that I could write one. xD

local-cluster/tests/local_cluster.rs
ryoqun: cool. that makes sense. The hard part of writing this was that validator actually try very hard to move forward the chain and resolve any divergent. I guess I can't hold validator from voting after restart, right? I'll update this comment of `for some reason` with your comment. :)

local-cluster/tests/local_cluster.rs
```diff
@@ -1680,7 +1601,282 @@ fn test_validator_saves_tower() {
...
+//
+fn do_test_optimistic_confirmation_violation_with_or_without_tower(with_tower: bool) {
+    solana_logger::setup();
+
+    // First set up the cluster with 4 nodes
+    let slots_per_epoch = 2048;
+    let node_stakes = vec![31, 36, 33, 0];
+
+    // Each pubkeys are prefixed with A, B, C and D.
+    // D is needed to avoid NoPropagatedConfirmation erorrs
```
ryoqun, line 1670: @carllin oh, btw, NoPropagatedConfirmation can be worked around just by not staked votes by non-existing vote account. So, any validator can be tricked into the false illusion of propagation. And attacker can do this without fear of slashing, maybe? This might be actually be a bug? Should we exclude 0-lamport votes and introduce some threshold like 1 (or N) lamport(s) or 0.01% stake?

sdk/src/signature.rs
```diff
@@ -43,6 +43,15 @@ impl Keypair {
         self.0.to_bytes()
     }
 
+    pub fn from_base58_string(s: &str) -> Self {
+        Self::from_bytes(&bs58::decode(s).into_vec().unwrap()).unwrap()
+    }
+
+    pub fn to_base58_string(&self) -> String {
+        // Remove .iter() once we're rust 1.47+
```
ryoqun, line 51: FYI: @t-nelson Another `LengthAtMost32` artificial shackle. I'm looking forward to #12739. :)

```
error[E0277]: arrays only have std trait implementations for lengths 0..=32
   --> sdk/src/signature.rs:51:22
    |
51  |         bs58::encode(&self.0.to_bytes()).into_string()
    |                      ^^^^^^^^^^^^^^^^^^ the trait `std::array::LengthAtMost32` is not implemented for `[u8; 64]`
    |
   ::: /home/.cargo/registry/src/github.com-1ecc6299db9ec823/bs58-0.3.1/src/lib.rs:207:18
    |
207 | pub fn encode<I: AsRef<[u8]>>(input: I) -> encode::EncodeBuilder<'static, I> {
    |                  ----------- required by this bound in `bs58::encode`
    |
    = note: required because of the requirements on the impl of `std::convert::AsRef<[u8]>` for `[u8; 64]`
    = note: required because of the requirements on the impl of `std::convert::AsRef<[u8]>` for `&[u8; 64]`

```

local-cluster/tests/local_cluster.rs
ryoqun: done!

local-cluster/tests/local_cluster.rs
ryoqun: done

local-cluster/tests/local_cluster.rs
ryoqun: done

local-cluster/tests/local_cluster.rs
ryoqun: yeah, this suggestion makes sense. well, I was just lazy... ;)

local-cluster/tests/local_cluster.rs
```diff
@@ -1415,28 +1422,15 @@ fn test_optimistic_confirmation_violation_with_no_tower() {
...
-            );
-        });
+        let blockstore = open_blockstore(&exited_validator_info.info.ledger_path);
         info!(
             "Setting slot: {} on main fork as dead, should cause fork",
             prev_voted_slot
         );
+        // marking this voted slot as dead makes the saved tower garbage
+        // effectively. That's because its stray last vote becomes stale (= no
+        // ancestor in bank forks).
```
ryoqun, line 1432: here

local-cluster/tests/local_cluster.rs
ryoqun: I added explicit comment mentioning for the saved tower: https://github.com/solana-labs/solana/pull/12350/files#r506909261

ryoqun:
@carllin I think this pr is ready for another round of review. :)

local-cluster/tests/local_cluster.rs
carllin: @ryoqun yeah no way to prevent the validator from voting on restart, so I think this is the correct thing!

ryoqun:
https://github.com/solana-labs/solana/pull/12350#discussion_r506836502

> @carllin oh, btw, NoPropagatedConfirmation can be worked around just by not staked votes by non-existing vote account. So, any validator can be tricked into the false illusion of propagation. And attacker can do this without fear of slashing, maybe? This might be actually be a bug? Should we exclude 0-lamport votes and introduce some threshold like 1 (or N) lamport(s) or 0.01% stake?

we'll take a look at this later. I'll first merge this as this lgtm-ed by @carllin via discord.

carllin (approved):

local-cluster/tests/local_cluster.rs
```diff
@@ -1680,7 +1594,252 @@ fn test_validator_saves_tower() {
...
+fn open_blockstore(ledger_path: &Path) -> Blockstore {
+    Blockstore::open_with_access_type(ledger_path, AccessType::PrimaryOnly, None).unwrap_or_else(
+        |e| {
+            panic!("Failed to open ledger at {:?}, err: {}", ledger_path, e);
+        },
+    )
+}
+
+fn purge_slots(blockstore: &Blockstore, start_slot: Slot, slot_count: Slot) {
+    blockstore.purge_from_next_slots(start_slot, start_slot + slot_count);
```
ryoqun, line 1609: this was the correct order (cf: #13065 )

local-cluster/tests/local_cluster.rs
```diff
@@ -1680,7 +1594,252 @@ fn test_validator_saves_tower() {
...
+        } else {
+            panic!("tower restore failed...: {:?}", tower_err);
+        }
+    }
+    // actually saved tower must have at least one vote.
+    let last_vote = Tower::restore(&ledger_path, &node_pubkey)
+        .unwrap()
+        .last_voted_slot()
+        .unwrap();
+    Some(last_vote)
```
ryoqun, line 1627: As @CriesofCarrots noticed (https://discord.com/channels/428295358100013066/560503042458517505/858178548916027402):

> Also, why does this method [2] call Tower::restore twice?

this is oversight according to this local commit (fyi, I retain all edits since i cloned the repo locally for accountability...)
The root ancesty of the redundant Tower::restore calls originated from this commit and then evolved significantly, without the bug not being noticed to this day... ;)
```
$ git show 765ec9c462dbe4e2282426f9230b941dbf5cb8f6
commit 765ec9c462dbe4e2282426f9230b941dbf5cb8f6
Author: Ryo Onodera <ryoqun@gmail.com>
Date:   Thu Oct 8 04:13:46 2020 +0900

    save

diff --git a/local-cluster/tests/local_cluster.rs b/local-cluster/tests/local_cluster.rs
index f07d021614..128e5071c4 100644
--- a/local-cluster/tests/local_cluster.rs
+++ b/local-cluster/tests/local_cluster.rs
@@ -1717,6 +1717,8 @@ fn do_test_optimistic_confirmation_violation_with_or_without_tower(with_tower: b
     let mut bad_vote_detected = false;
     let mut retry = 100;
     loop {
+        let tower = Tower::restore(&val_a_ledger_path, &val_a_7B);
+        if tower.is_err() { continue }
         let s = Tower::restore(&val_a_ledger_path, &val_a_7B).unwrap().last_voted_slot().unwrap();
         if val_c_slots.contains(&s) {
             bad_vote_detected = true;
```

PR #12360, opened 2020-09-20
builds crds filters in parallel

behzadnouri:
#### Problem
Based on run-time profiles, the majority time of new_pull_requests is
spent building bloom filters, in hashing and bit-vec ops.

#### Summary of Changes
This commit builds crds filters in parallel using rayon constructs. The
added benchmark shows ~5x speedup (4-core machine, 8 threads).

### DIFF
--- a/core/src/crds_gossip_pull.rs
+++ b/core/src/crds_gossip_pull.rs
@@ -16,7 +16,8 @@ use crate::crds_gossip_error::CrdsGossipError;
 use crate::crds_value::{CrdsValue, CrdsValueLabel};
 use rand::distributions::{Distribution, WeightedIndex};
 use rand::Rng;
-use solana_runtime::bloom::Bloom;
+use rayon::{prelude::*, ThreadPool};
+use solana_runtime::bloom::{AtomicBloom, Bloom};
 use solana_sdk::hash::Hash;
 use solana_sdk::pubkey::Pubkey;
 use std::cmp;
@@ -116,35 +117,44 @@ impl CrdsFilter {
 }
 
 /// A vector of crds filters that together hold a complete set of Hashes.
-struct CrdsFilterSet(Vec<CrdsFilter>);
+struct CrdsFilterSet {
+    filters: Vec<AtomicBloom<Hash>>,
+    mask_bits: u32,
+}
 
 impl CrdsFilterSet {
     fn new(num_items: usize, max_bytes: usize) -> Self {
         let max_bits = (max_bytes * 8) as f64;
         let max_items = CrdsFilter::max_items(max_bits, FALSE_RATE, KEYS);
         let mask_bits = CrdsFilter::mask_bits(num_items as f64, max_items as f64);
-        // For each possible mask combination, generate a new filter.
-        let seeds = 0..2u64.pow(mask_bits);
-        let filters = seeds.map(|seed| {
-            let filter = Bloom::random(max_items as usize, FALSE_RATE, max_bits as usize);
-            let mask = CrdsFilter::compute_mask(seed, mask_bits);
-            CrdsFilter {
-                filter,
-                mask,
-                mask_bits,
-            }
-        });
-        Self(filters.collect())
+        let filters = std::iter::repeat_with(|| {
+            Bloom::random(max_items as usize, FALSE_RATE, max_bits as usize).into()
+        })
+        .take(1 << mask_bits)
+        .collect();
+        Self { filters, mask_bits }
     }
 
-    // Returns the filter within the vector of crds filters which corresponds
-    // to the given hash value.
-    fn get(&mut self, hash_value: &Hash) -> Option<&mut CrdsFilter> {
-        let shift = 64 - self.0.first()?.mask_bits.min(64);
-        let index = CrdsFilter::hash_as_u64(hash_value)
-            .checked_shr(shift)
-            .unwrap_or(0u64);
-        self.0.get_mut(index as usize)
+    fn add(&self, hash_value: Hash) {
+        let index = CrdsFilter::hash_as_u64(&hash_value)
+            .checked_shr(64 - self.mask_bits)
+            .unwrap_or(0);
+        self.filters[index as usize].add(&hash_value);
+    }
+}
+
+impl Into<Vec<CrdsFilter>> for CrdsFilterSet {
+    fn into(self) -> Vec<CrdsFilter> {
+        let mask_bits = self.mask_bits;
+        self.filters
+            .into_iter()
+            .enumerate()
+            .map(|(seed, filter)| CrdsFilter {
+                filter: filter.into(),
+                mask: CrdsFilter::compute_mask(seed as u64, mask_bits),
+                mask_bits,
+            })
+            .collect()
     }
 }
 
@@ -182,6 +192,7 @@ impl CrdsGossipPull {
     /// generate a random request
     pub fn new_pull_request(
         &self,
+        thread_pool: &ThreadPool,
         crds: &Crds,
         self_id: &Pubkey,
         self_shred_version: u16,
@@ -201,7 +212,7 @@ impl CrdsGossipPull {
         if options.is_empty() {
             return Err(CrdsGossipError::NoPeers);
         }
-        let filters = self.build_crds_filters(crds, bloom_size);
+        let filters = self.build_crds_filters(thread_pool, crds, bloom_size);
         let index = WeightedIndex::new(options.iter().map(|weighted| weighted.0)).unwrap();
         let random = index.sample(&mut rand::thread_rng());
         let self_info = crds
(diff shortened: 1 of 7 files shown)


core/src/crds_gossip_pull.rs
```diff
@@ -743,37 +760,41 @@ mod test {
...
-            let mut bytes = [0u8; HASH_BYTES];
-            rng.fill_bytes(&mut bytes);
-            let hash_value = Hash::new(&bytes);
-            let filter = crds_filter_set.get(&hash_value).unwrap().clone();
-            assert!(filter.test_mask(&hash_value));
-            // Validate that the returned filter is the *unique* filter which
-            // corresponds to the hash value (i.e. test_mask returns true).
+        let crds_filter_set =
+            CrdsFilterSet::new(/*num_items=*/ 9672788, /*max_bytes=*/ 8196);
+        let hashe_values: Vec<_> = std::iter::repeat_with(|| Hash::new_rand(&mut rng))
```
sakridge, line 767: `hash_values`

core/src/crds_gossip_pull.rs
```diff
@@ -391,25 +409,27 @@ impl CrdsGossipPull {
     // build a set of filters of the current crds table
     // num_filters - used to increase the likelyhood of a value in crds being added to some filter
     pub fn build_crds_filters(&self, crds: &Crds, bloom_size: usize) -> Vec<CrdsFilter> {
+        const PAR_MIN_LENGTH: usize = 1024;
```
sakridge, line 412: Maybe want to have some reasoning for this value.

sakridge (dismissed):
Looks pretty good to me, just couple of nits. I could think of some ideas to optimize around the lock critical section and maybe coalescing work going into the lock, but seems quite a bit better than what we have.

core/src/crds_gossip_pull.rs
behzadnouri: fixed, thanks

core/src/crds_gossip_pull.rs
behzadnouri: Just by experimenting using the added benchmark, I realized that there is no speed up for smaller arrays. It does not slow down the code, but I thought that probably better not to allocate threads for no gains.

core/src/crds_gossip_pull.rs
```diff
@@ -391,25 +409,27 @@ impl CrdsGossipPull {
...
-        }
-
-        filters.0
+        let filters = CrdsFilterSet::new(num, bloom_size);
+        rayon::join(
+            || {
+                crds.table
+                    .par_values()
+                    .with_min_len(PAR_MIN_LENGTH)
+                    .for_each(|v| filters.add(v.value_hash))
```
carllin, line 423: Does aggregating the values for each filter outside of the lock before calling `add` to bulk insert those values improve the bench at all?

core/src/crds_gossip_pull.rs
sakridge: Ok, yea I mean is there any experiments that show 1024 is better than 512 or 256 or another constant.

core/src/crds_gossip_pull.rs
```diff
@@ -391,25 +409,27 @@ impl CrdsGossipPull {
...
-            add_value_hash(&v.value_hash);
-        }
-        for (value_hash, _insert_timestamp) in &self.purged_values {
-            add_value_hash(&value_hash);
-        }
-
-        filters.0
+        let filters = CrdsFilterSet::new(num, bloom_size);
+        rayon::join(
+            || {
```
sakridge, line 419: I think using the gossip-specific threadpool here might be a good idea. The thread limit will keep it from using the whole machine's CPU and we've found that not sharing work between other instances of rayon generally performs better.

core/src/crds_gossip_pull.rs
behzadnouri: Initially I tried something similar to what I think you are suggesting, and it was slightly slower. I have some ideas about a lock-free implementation of the bloom filter construction, that I am going to test as well and compare.

core/src/crds_gossip_pull.rs
behzadnouri: Agreed, but I tried adding a ThreadPool to [`struct CrdsGossipPull`](https://github.com/solana-labs/solana/blob/02ba197a06184e0a6ae2254ad387ee9376fd18c4/core/src/crds_gossip_pull.rs#L160), and it was not compiling with the `#[derive(Clone)]` thing. [`struct CrdsGossip`](https://github.com/solana-labs/solana/blob/02ba197a06184e0a6ae2254ad387ee9376fd18c4/core/src/crds_gossip.rs#L20) also has `
#[derive(Clone)]`. Would it make sense to add the `ThreadPool` to [`struct ClusterInfo`](https://github.com/solana-labs/solana/blob/02ba197a06184e0a6ae2254ad387ee9376fd18c4/core/src/cluster_info.rs#L248) and pass it down the call-stack? or, do you have a different suggestion?

I am not sure those `#[derive(Clone)]` are used anywhere, so we may be able to drop them. Or, alternatively implement `Clone` manually.

sakridge: yea. I was thinking passing it down the call stack. removing `#[derive(Clone)]` or doing `Clone` manually doesn't seem like it would be too bad.

behzadnouri: added a ThreadPool to `struct ClusterInfo`. please take a look.

behzadnouri: After adding a dedicated ThreadPool to [`struct ClusterInfo`](https://github.com/solana-labs/solana/blob/afd9bfc45f7c467470d060dac892de878c9976c0/core/src/cluster_info.rs#L248), a number of tests started seg-faulting and crashing, apparently because of going out of memory. so I had to:
* add `#[serial]` to two of the tests in [core/tests/crds_gossip.rs](https://github.com/solana-labs/solana/blob/master/core/tests/crds_gossip.rs).
* limit number of threads in [`clone_with_id`](https://github.com/solana-labs/solana/blob/afd9bfc45f7c467470d060dac892de878c9976c0/core/src/cluster_info.rs#L434) when cloning cluster-info.

Should we also limit the number of threads for the threadpool in cluster-info as well?

sakridge: I see. I think we do have a limit there:
```
 let thread_pool = rayon::ThreadPoolBuilder::new()
     .num_threads(std::cmp::min(get_thread_count(), 8))
     .thread_name(|i| format!("sol-gossip-work-{}", i))
```

core/src/crds_gossip_pull.rs
sakridge: Yea, I was thinking of doing like a batch calculate a set of positions to set outside the lock.. then go into the lock and bulk set the positions. It might be more beneficial on a larger machine with more cores which would show more lock contention.

core/src/crds_gossip_pull.rs
behzadnouri: I was adding an entirely new ThreadPool here. But, from your comment, I realized that I am doing this wrong, because then the gossip service will have two separate thread pools.
At the same time I cannot use the [`"sol-gossip-work-{i}" threads`](https://github.com/solana-labs/solana/blob/afd9bfc45f7c467470d060dac892de878c9976c0/core/src/cluster_info.rs#L2330-L2334), because those are created in [`ClusterInfo::listen` thread](https://github.com/solana-labs/solana/blob/afd9bfc45f7c467470d060dac892de878c9976c0/core/src/gossip_service.rs#L54), whereas this new code is invoked in [`ClusterInfo::gossip` thread](https://github.com/solana-labs/solana/blob/afd9bfc45f7c467470d060dac892de878c9976c0/core/src/gossip_service.rs#L61).

We can reorganize the gossip service code to create the thread pool for all gossip service work in [`GossipService::new`](https://github.com/solana-labs/solana/blob/afd9bfc45f7c467470d060dac892de878c9976c0/core/src/gossip_service.rs#L31), and then share it between [listen and gossip threads](https://github.com/solana-labs/solana/blob/afd9bfc45f7c467470d060dac892de878c9976c0/core/src/gossip_service.rs#L54-L67). It will look like [this commit](https://github.com/behzadnouri/solana/commit/fdd668052dce64f16d442e052c5ee418ab8da9d0). If that looks right to you I will first send that as a separate pull request first.

behzadnouri: An alternative would be to actually have _two separate thread-pools_, one for [`ClusterInfo::listen` thread](https://github.com/solana-labs/solana/blob/afd9bfc45f7c467470d060dac892de878c9976c0/core/src/gossip_service.rs#L54), and the other for [`ClusterInfo::gossip` thread](https://github.com/solana-labs/solana/blob/afd9bfc45f7c467470d060dac892de878c9976c0/core/src/gossip_service.rs#L61).

It has the advantage that `listen` and `gossip` may not block each other, at the cost of higher memory requirements at max load.

sakridge: Yea. I hesitate to add another thread pool because we already have too many. The pool getting created in gossip service doesn't look too bad and shared between them. The work is not really blocking like an IO-heavy workload might be, so I think it might be fine to share the thread pool.

behzadnouri: sounds good. I sent out https://github.com/solana-labs/solana/pull/12402

sakridge: Actually, I think the method in this PR is fine where thread_pool is a member of `ClusterInfo` struct. Just move the thread pool creation of gossip-work-{i} to `ClusterInfo::new` and then use `self.thread_pool` in `ClusterInfo::listen`. For test configuration just create a 1 or 2 thread pool.

core/src/crds_gossip_pull.rs
behzadnouri: sent out https://github.com/solana-labs/solana/pull/12422 which adds an atomic variant of the bloom filter, and will use it here once that is merged. so wouldn't need locking and mutex here anymore.

core/src/crds_gossip_pull.rs
behzadnouri: update https://github.com/solana-labs/solana/pull/12402 accordingly

sakridge (approved):
lgtm

PR #12376, opened 2020-09-21
Runtime feature activation framework

mvines:
Runtime feature activation across all branches/clusters is too hard.   There can be a better way, and this PR moves towards it.

TODO:
* [x] Port the instruction sysvar and secp256k1 program activation logic over
* [x] Add cli side commands: `solana feature activate <keypair>`, and `solana feature status <keypair>`
* [x] Probably as a follow-up, expose `FeatureSet::default().id` in the node's gossip software version so it's easy to figure what nodes are running the same feature set (so we can get away from using the release version for this)
* [x] Tests


mvines:
This is starting to feel pretty nice.  Next up: cli support to request feature activation and report feature activation status

mvines:
Ok this is ready for review!  The PR is a little sprawling so here's where I think you can help:
* @sakridge - please check out the instruction sysvar and secp256k1 program activation logic changes, I basically touched everything that you touched in your recent activation PR
* @ryoqun - can you please review the new FeatureSet-related code in runtime/src/{feature.rs,feature_set.rs,bank.rs}
* @CriesofCarrots - `cli/src/feature.rs` please.  I tried something different here, going for an approach at minimizes touching `cli/src/cli.rs`.  I also plumbed the new feature set identifier through RPC, so that the cli is able to figure out if the cluster in which a feature is about to be activated on can actually handle it

cli/src/feature.rs
```diff
@@ -0,0 +1,269 @@
...
+    feature_subcommand: &FeatureCliCommand,
+) -> ProcessResult {
+    match feature_subcommand {
+        FeatureCliCommand::Status { features } => process_status(rpc_client, features),
+        FeatureCliCommand::Activate { feature } => process_activate(rpc_client, config, *feature),
+    }
+}
+
+// Feature activation is only allowed when 95% of the active stake is on the current feature set
+fn feature_activation_allowed(rpc_client: &RpcClient) -> Result<bool, Box<dyn std::error::Error>> {
```
CriesofCarrots, line 120: Return ClientError here?

cli/src/feature.rs
```diff
@@ -0,0 +1,269 @@
...
+                                .validator(is_valid_pubkey)
+                                .index(1)
+                                .help("Feature status to query [default: "),
+                        ),
+                )
+                .subcommand(
+                    SubCommand::with_name("activate")
+                        .about("Activate a runtime feature")
+                        .arg(
+                            Arg::with_name("feature")
```
CriesofCarrots, line 49: Make required?

version/src/lib.rs
```diff
@@ -3,19 +3,42 @@
 extern crate serde_derive;
 use serde_derive::{Deserialize, Serialize};
 use solana_sdk::sanitize::Sanitize;
-use std::fmt;
-
+use std::{convert::TryInto, fmt};
 #[macro_use]
 extern crate solana_sdk_macro_frozen_abi;
 
+// Older version structure used by 1.3.12 and earlier releases
```
CriesofCarrots, line 10:
```suggestion
// Older version structure used by 1.3.13 and earlier releases
```

cli/src/cli.rs
```diff
@@ -1954,6 +1964,7 @@ pub fn app<'ab, 'v>(name: &str, about: &'ab str, version: &'v str) -> App<'ab, '
         .cluster_query_subcommands()
         .nonce_subcommands()
         .stake_subcommands()
+        .feature_subcommands()
```
CriesofCarrots, line 1967:
```suggestion
        .cluster_query_subcommands()
        .feature_subcommands()
        .nonce_subcommands()
        .stake_subcommands()
```

runtime/src/message_processor.rs
```diff
@@ -678,12 +680,11 @@ impl MessageProcessor {
...
         instruction_recorder: Option<InstructionRecorder>,
         instruction_index: usize,
-        cluster_type: ClusterType,
-        epoch: Epoch,
+        feature_set: &FeatureSet,
     ) -> Result<(), InstructionError> {
         // Fixup the special instructions key if present
         // before the account pre-values are taken care of
-        if solana_sdk::sysvar::instructions::is_enabled(epoch, cluster_type) {
+        if feature_set.active(&feature_set::instructions_sysvar_enabled::id()) {
```
jackcmay, line 687: nit, convention is usually `is_active`

cli/src/feature.rs
```diff
@@ -0,0 +1,269 @@
...
+            }
+        })
+        .sum();
+
+    Ok(total_compatible_stake * 100 / total_active_stake >= 95)
+}
+
+fn process_status(rpc_client: &RpcClient, feature_ids: &[Pubkey]) -> ProcessResult {
+    if feature_ids.len() > 1 {
+        println!(
```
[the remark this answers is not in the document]
mvines: Sure, follow-up sounds great here.

cli/src/feature.rs
```diff
@@ -0,0 +1,269 @@
...
+    Ok("".to_string())
+}
+
+fn process_activate(
+    rpc_client: &RpcClient,
+    config: &CliConfig,
+    feature_id: Pubkey,
+) -> ProcessResult {
+    let account = rpc_client
+        .get_multiple_accounts(&[feature_id])?
```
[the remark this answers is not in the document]
mvines: `get_account` returns `Err(...)` when the account doesn't exist and didn't want to have to distinguish between error.  Looks like maybe `get_account_with_commitment()` doesn't have this problem but then I need to `use CommitmentConfig` so not a fan of that path either

cli/src/feature.rs
```diff
@@ -0,0 +1,269 @@
...
+                .setting(AppSettings::SubcommandRequiredElseHelp)
+                .subcommand(
+                    SubCommand::with_name("status")
+                        .about("Query runtime feature status")
+                        .arg(
+                            Arg::with_name("feature")
+                                .value_name("ADDRESS")
+                                .validator(is_valid_pubkey)
+                                .index(1)
+                                .help("Feature status to query [default: "),
```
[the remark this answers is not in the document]
mvines: Sure, done.

runtime/src/message_processor.rs
mvines: sure

mvines:
All review feedback applied

CriesofCarrots (dismissed):
Cli, client, and rpc changes lgtm!

jackcmay:
@mvines Phase two is passing the feature set into `InvokeContext`?

version/src/lib.rs
```diff
@@ -3,19 +3,42 @@
...
 
+impl Sanitize for LegacyVersion {}
+
+#[derive(Serialize, Deserialize, Clone, PartialEq, AbiExample)]
+pub struct Version {
+    pub major: u16,
+    pub minor: u16,
+    pub patch: u16,
+    pub commit: Option<u32>, // first 4 bytes of the sha1 commit hash
+    pub feature_set: u32,    // first 4 bytes of the FeatureSet identifier
```
jackcmay, line 27: Nice!

jackcmay (dismissed):

mvines:
> @mvines Phase two is passing the feature set into `InvokeContext`?

Oh yeah that could be useful and allow us to even cleanly sneak features into builtin programs.

ryoqun:
> @ryoqun - can you please review the new FeatureSet-related code in runtime/src/{feature.rs,feature_set.rs,bank.rs}

Sorry for being late to review this... I just started to review this. I'm going to review this weekend. If that's too late, could you ask someone else to look at?

mvines:
> I just started to review this. I'm going to review this weekend.

That sounds perfects :)

mvines:
@ryoqun / @sakridge - I'm going to land this PR on master since it's causing a backup.  But I'll hold off on landing the v1.3 backport until Monday in case there's post-merge review feedback

runtime/src/bank.rs
```diff
@@ -3455,25 +3463,80 @@ impl Bank {
...
+                            if allow_new_activations {
+                                // Feature has been requested, activate it now
+                                feature.activated_at = Some(slot);
+                                if feature.to_account(&mut account).is_some() {
+                                    self.store_account(feature_id, &account);
+                                }
+                                newly_activated.insert(*feature_id);
+                                active.insert(*feature_id);
+                                info!("Feature {} activated at slot {}", feature_id, slot);
+                                continue;
```
ryoqun, line 3504: nits: I think these two `continue`s' control flow is a bit hard to follow. At the same time, I feel the urge to skip the general catch-all `inactive.insert(...)` for these cases in some way. So, I'd like this to be something like this:

```rust
diff --git a/runtime/src/bank.rs b/runtime/src/bank.rs
index 078d2402e4..daab6dd347 100644
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -3484,6 +3484,7 @@ impl Bank {
         let slot = self.slot();
 
         for feature_id in &self.feature_set.inactive {
+            let mut activated = false;
             if let Some(mut account) = self.get_account(feature_id) {
                 if let Some(mut feature) = Feature::from_account(&account) {
                     match feature.activated_at {
@@ -3497,20 +3498,22 @@ impl Bank {
                                 newly_activated.insert(*feature_id);
                                 active.insert(*feature_id);
                                 info!("Feature {} activated at slot {}", feature_id, slot);
-                                continue;
+                                activated = true;
                             }
                         }
                         Some(activation_slot) => {
                             if slot >= activation_slot {
                                 // Feature is already active
                                 active.insert(*feature_id);
-                                continue;
+                                activated = true;
                             }
                         }
                     }
                 }
             }
-            inactive.insert(*feature_id);
+            if !activated {
+                inactive.insert(*feature_id);
+            }
         }
 
         self.feature_set = Arc::new(FeatureSet { active, inactive });
```

ryoqun: I further edited my diff; specifically moved the `active` handling as well. This way, the dichotomy of `active` and `inactive` is clear in my opinion. :)

mvines: Thanks, I applied your suggestion at https://github.com/solana-labs/solana/pull/12510

runtime/src/feature.rs
```diff
@@ -0,0 +1,42 @@
...
+    }
+    pub fn to_account(&self, account: &mut Account) -> Option<()> {
+        bincode::serialize_into(&mut account.data[..], self).ok()
+    }
+    pub fn create_account(&self, lamports: u64) -> Account {
+        let data_len = Self::size_of().max(bincode::serialized_size(self).unwrap() as usize);
+        let mut account = Account::new(lamports, data_len, &id());
+        self.to_account(&mut account).unwrap();
+        account
+    }
```
ryoqun, line 41: (Sorry for further post-merge review comments; I'm finishing this off)

nits: According to this way of [creating new `Feature` accounts](https://github.com/solana-labs/solana/pull/12376/files#diff-2b291ef836c6fc9eeaafd31e6a38d639R256) (skipping special system instruction for `Feature`), this is assuming `Feature::size_of()`-sized zero byte arrays should be always to be deserialized as `Feature { activated_at: None }` successfully? This is sensible bet. However, this could be change `bincode` version up, or switch to completely new other deserialization implementation (protobuf, borsh).

How about writing a tiny unit-test to assume that? Or am I worrying too much?

```
  #[test]
  fn test_deserialize_blank_new_account() {
      let just_initialized = Account::new(42, Feature::size_of(), &id());
      assert_eq!(Feature::from_account(&just_initialized), Some(Feature{activated_at: None}));
  }
```

runtime/src/feature.rs
```diff
@@ -0,0 +1,42 @@
...
+            None
+        } else {
+            bincode::deserialize(&account.data).ok()
+        }
+    }
+    pub fn to_account(&self, account: &mut Account) -> Option<()> {
+        bincode::serialize_into(&mut account.data[..], self).ok()
+    }
+    pub fn create_account(&self, lamports: u64) -> Account {
+        let data_len = Self::size_of().max(bincode::serialized_size(self).unwrap() as usize);
```
ryoqun, line 37: I couldn't decipher the subtlety of needing `{integer}.max()` here. Commenting might be desired. I think `data_len` just straightly can be set to `Self::size_of()`. Is there any possibility that actually `bincode::serialized_size(self)` could be larger than `bincode::serialized_size(activated_at: Some(Slot::MAX))`?

Or is this considering future fields changes like that?

cli/src/feature.rs
```diff
@@ -0,0 +1,273 @@
...
+        .map(|vote_account| {
+            if Some(&Some(my_feature_set)) == feature_set_map.get(&vote_account.node_pubkey) {
+                vote_account.activated_stake
+            } else {
+                0
+            }
+        })
+        .sum();
+
+    Ok(total_compatible_stake * 100 / total_active_stake >= 95)
```
ryoqun, line 154: I think this is very cool thanks to this runtime feature activation framework.

Also, I don't think this check needs to be on-chain. So, this guard is good to reside on cli here. :)

ryoqun (commented):
LGTM with nits! I'm really looking forward to this for the ease of feature gate ops. :)

runtime/src/feature.rs
mvines: Done in https://github.com/solana-labs/solana/pull/12524

version/src/lib.rs
```diff
@@ -27,27 +50,40 @@ fn compute_commit(sha1: Option<&'static str>) -> Option<u32> {
...
     }
 }
 
 impl fmt::Display for Version {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}.{}.{}", self.major, self.minor, self.patch,)
+    }
+}
+
+impl fmt::Debug for Version {
```
ryoqun, line 74: @mvines btw, I'm going to tweak this a abit. Well, I miss `solana-XXX --version`'s commit hash output:

before:
```
$ solana-ledger-tool --version
solana-ledger-tool 1.4.0 775340c6

```

after:
```
$ solana-ledger-tool --version
solana-ledger-tool 1.4.0
```

Well, it's very useful for development purpose at least for me to know which build commit from bunch of laying binaries with `--version` and log output. Both has gone now.

How about this change?

- Add new macro `crate::full_version!` which `fmt("{:?}", Version)` and replace all over place in various `main.rs`.

or
 
- Add new macro `crate::simple_version!` and restore `crate::version!` and change  what you really wanted the `fmt("{}.{}.{}")` behavior.

mvines: I've been lamenting it's removal as well. 

I think changing
https://github.com/solana-labs/solana/blob/b2fc7f7caa2061bd7eb3485175f2ebdcb299c053/version/src/lib.rs#L96
to `{:?}` will do the trick actually.

CriesofCarrots: +1 for commit hash 😁

ryoqun: @mvines @CriesofCarrots just created #12636 :)

PR #12380, opened 2020-09-22
Fix solana-tokens check_payer_balances for distribute-stake

CriesofCarrots:
#### Problem
The `check_payer_balances` method was making assumptions about distribution and fees amounts and payers that were true for the `distribute-tokens` subcommand, but not for `distribute-stakes`. As a result, it was returning erroneous errors for `distribute-stakes` on dry-run.

#### Summary of Changes
- Add proper handling for distribute-stakes when `args.stake_args.is_some()`

### DIFF
--- a/tokens/src/commands.rs
+++ b/tokens/src/commands.rs
@@ -43,10 +43,36 @@ struct Allocation {
 }
 
 #[derive(Debug, PartialEq)]
-pub enum FundType {
-    Distribution,
-    Fees,
-    SolForFees,
+pub enum FundingSource {
+    FeePayer,
+    StakeAccount,
+    SystemAccount,
+}
+
+pub struct FundingSources(Vec<FundingSource>);
+
+impl std::fmt::Debug for FundingSources {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        for (i, source) in self.0.iter().enumerate() {
+            if i > 0 {
+                write!(f, "/")?;
+            }
+            write!(f, "{:?}", source)?;
+        }
+        Ok(())
+    }
+}
+
+impl PartialEq for FundingSources {
+    fn eq(&self, other: &Self) -> bool {
+        self.0 == other.0
+    }
+}
+
+impl From<Vec<FundingSource>> for FundingSources {
+    fn from(sources_vec: Vec<FundingSource>) -> Self {
+        Self(sources_vec)
+    }
 }
 
 #[derive(thiserror::Error, Debug)]
@@ -61,10 +87,8 @@ pub enum Error {
     TransportError(#[from] TransportError),
     #[error("Missing lockup authority")]
     MissingLockupAuthority,
-    #[error("insufficient funds for {0:?} ({1} SOL)")]
-    InsufficientFunds(FundType, f64),
-    #[error("insufficient funds for {0:?} ({1} SOL) and {2:?} ({3} SOL)")]
-    InsufficientFundsForMultiple(FundType, f64, FundType, f64),
+    #[error("insufficient funds in {0:?}, requires {1} SOL")]
+    InsufficientFunds(FundingSources, f64),
 }
 
 fn merge_allocations(allocations: &[Allocation]) -> Vec<Allocation> {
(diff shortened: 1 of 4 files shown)


tokens/src/commands.rs
```diff
@@ -42,6 +42,13 @@ struct Allocation {
     lockup_date: String,
 }
 
+#[derive(Debug, PartialEq)]
+pub enum FundType {
```
garious, line 46: Strange name.  Implies a "type of fund".  How about `FundingSource`?

tokens/src/commands.rs
```diff
@@ -42,6 +42,13 @@ struct Allocation {
...
+    Distribution,
+    Fees,
+    SolForFees,
```
garious, line 49: These names confuse me, could these be "StakeAccount", "FeePayer", and "SystemAccount"?

tokens/src/commands.rs
```diff
@@ -990,4 +1048,309 @@ mod tests {
...
+            let recipient = Pubkey::new_rand();
+            let allocations = vec![Allocation {
+                recipient: recipient.to_string(),
+                amount: 1000.0,
+                lockup_date: "".to_string(),
+            }];
+
+            // Test distribute tokens with fully funded single payer
+            let mut args = DistributeTokensArgs {
+                sender_keypair: read_keypair_file(&sender_keypair_file).unwrap().into(),
```
garious, line 1081: What's with all the files?

tokens/src/commands.rs
```diff
@@ -990,4 +1048,309 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_check_payer_balances() {
+        let (mut genesis_config, sender_keypair) =
+            create_genesis_config(sol_to_lamports(9_000_000.0));
+        genesis_config.fee_rate_governor = FeeRateGovernor::new(10_000, 0);
+        let bank_forks = Arc::new(RwLock::new(BankForks::new(Bank::new(&genesis_config))));
+        Runtime::new().unwrap().block_on(async {
+            let transport = start_local_server(&bank_forks).await;
```
garious, line 1066: Theses are parallel-friendly tests. No need to stuff these independent cases in one enormous test.

tokens/src/commands.rs
CriesofCarrots: Easiest way I know of to be able to use the same Keypair in more than one signing role, since Keypair doesn't support clone. Open to another pattern, if you have a suggestion.

tokens/src/commands.rs
CriesofCarrots: Currently, these names describe what the amounts are for, not which account is paying them. I'll take a look and see if I can make helpful error messages using the FundingSource instead.

Honestly, I find `sol_for_fees` generally obscure, since it's not paying for any fees reflected in solana-tokens. I infer it is bonus SOL that you are giving recipients to pay for fees for some sort of transactions after the distribution, but the critical bit is that it's going directly to the `recipient` in the csv file as opposed to the new stake account that will controlled by `recipient`, right? Perhaps it should be named accordingly.

garious: Would it help to use the term "fee pool" instead?

CriesofCarrots: I think the word "fee" is the problematic part for a user who doesn't have history with this tool, or may not be using it as part of the same external process you are.

garious: Ah, how about something like "unlocked sol"?

CriesofCarrots: Definitely better!

tokens/src/commands.rs
```diff
@@ -261,16 +290,7 @@ async fn distribute_allocations(
...
-        let undistributed_tokens: f64 = allocations.iter().map(|x| x.amount).sum();
-        check_payer_balances(
-            num_signatures,
-            sol_to_lamports(undistributed_tokens),
-            client,
-            args,
-        )
-        .await?;
-    }
+    check_payer_balances(num_signatures, allocations, client, args).await?;
```
CriesofCarrots, line 293: @garious there's an outstanding issue here, and I'd appreciate your thoughts. Now that this check happens on real-deal runs, it needs to move, otherwise the first allocation gets attempted and logged in the db. It probably makes more sense to for this check to be outside `distribute_allocations` anyway.

Can you recommend a neat way to determine the number of signatures for the distribution? (ideally without duplicating all of this https://github.com/solana-labs/solana/pull/12380/files#diff-f5393ff0891e002c6d8f29fe8b90fb07R235-L223)

garious: I don't think I understand the problem. As I recall, everything logged to the db during a dry run is discarded.

CriesofCarrots: Yep, but this PR promotes `check_payer_balances` to run on non-dry runs as well.

garious: I'm still a little lost, but if you refactor such that you collect all the Messages upfront, then you can calculate `num_signatures` using pub field `message.header.num_required_signatures`.  The only important aspect of the ordering here is that the Transaction is sent immediately after getting the blockhash and signing so that the blockhash won't expire.

garious (approved):
Looks good, thanks!

Issue #12305, opened 2020-09-17
explorer: Missing support for new token instruction types


jstarry:
#### Problem
Lacking support for instructions:
```
  "freezeAccount",
  "thawAccount",
  "transfer2",
  "approve2",
  "mintTo2",
  "burn2",
```

Example: https://explorer.solana.com/address/2z29e854PYavXz6Pnm2FMjg1bXckcXoo77hiz6FzjKyA/tokens

#### Proposed Solution
Add support and decode instructions for display in transaction details view

PR #12381, opened 2020-09-22
introduce additional token instructions

oJshua:
#### Problem
Lacking support for instructions:
```
  "freezeAccount",
  "thawAccount",
  "transfer2",
  "approve2",
  "mintTo2",
  "burn2",
```

Example: https://explorer.solana.com/address/2z29e854PYavXz6Pnm2FMjg1bXckcXoo77hiz6FzjKyA/tokens

#### Proposed Solution
Add support and decode instructions for display in transaction details view

Fixes https://github.com/solana-labs/solana/issues/12305


mvines:
@jackcmay is thinking about changing the `2` instructions to something else, maybe `Checked` prefix.  It'd be worth syncing with him quickly on this

jackcmay:
I don't really care either way, security audit complained about the undescriptive `2` naming and they have a minor point.  We could change them by replacing `2` with `Checked`.  I'm fine doing this but could also easily drop it if no one else cares.  (Hows that for noncommittal 😁)

mvines:
Let's do it.  The audit had a point, the `2` suffix means something to me but probably to nobody else on this earth

jackcmay:
@oJshua Instruction rename here: https://github.com/solana-labs/solana-program-library/pull/512

CriesofCarrots:
Can go ahead with the Checked names on the display side, but rpc will not be returning them until https://github.com/solana-labs/solana/issues/12391 is done and rolled out on api nodes.

oJshua:
Updated display name.

explorer/src/components/instruction/token/types.ts
```diff
@@ -140,4 +212,10 @@ export const IX_TITLES = {
   mintTo: "Mint To",
   burn: "Burn",
   closeAccount: "Close Account",
+  freezeAccount: "Freeze Account",
+  thawAccount: "Thaw Account",
+  transfer2: "Transfer Checked",
+  approve2: "Approve Checked",
+  mintTo2: "Mint To Checked",
+  burn2: "Burn Checked",
```
jstarry, line 220: nit: maybe add parens around `Checked`?

explorer/src/components/instruction/token/TokenDetailsCard.tsx
```diff
@@ -96,15 +96,38 @@ function TokenInstruction(props: InfoProps) {
...
+  for (let key in info) {
+    const value = info[key];
     if (value === undefined) continue;
 
     let tag;
     let labelSuffix = "";
     if (value instanceof PublicKey) {
       tag = <Address pubkey={value} alignRight link />;
+    } else if (key === "tokenAmount") {
+      buildAttributes(attributes, info.tokenAmount, decimals);
```
jstarry, line 129: I don't think it's necessary to flatten the full token amount object like this:

Screen Shot 2020-09-23 at 11 36 09 AM

How about we stick to displaying the ui amount nicely and skip the other fields?

explorer/src/components/instruction/token/types.ts
```diff
@@ -102,6 +108,60 @@ const CloseAccount = object({
   signers: optional(array(Pubkey)),
 });
 
+const FreezeAccount = object({
+  account: Pubkey,
+  mint: Pubkey,
+  freezeAuthority: optional(Pubkey),
+  multisigOwner: optional(Pubkey),
```
jstarry, line 115:
```suggestion
  multisigFreezeAuthority: optional(Pubkey),
```

explorer/src/components/instruction/token/types.ts
```diff
@@ -102,6 +108,60 @@ const CloseAccount = object({
...
+  signers: optional(array(Pubkey)),
+});
+
+const ThawAccount = object({
+  account: Pubkey,
+  mint: Pubkey,
+  freezeAuthority: optional(Pubkey),
+  multisigOwner: optional(Pubkey),
```
jstarry, line 123:
```suggestion
  multisigFreezeAuthority: optional(Pubkey),
```

oJshua: Thanks for this!

jstarry (dismissed):
Looks good!

Issue #5418, opened 2019-08-05
Add support for creating libra transactions in a Wasm SDK


jstarry:
#### Problem
Use the smallest amount of JS code to create and send librapay transactions. 

#### Proposed Solution
Use `wasm-bindgen` to create a JS interface to the `solana-sdk` Rust SDK. Specifically, we want to expose Libra types in a JS API.

The `solana_libra_types` crate currently has a lot of wasm incompatible deps so there would be quite a bit of work to make it wasm friendly. We could also create a `solana_libra_types_lite` crate to avoid some of these issues.

---
Using the `solana_libra_types` crate as is presents a lot of problems
* Libra uses [`proptest`](https://crates.io/crates/proptest) to do property based testing but it uses OS calls to create forks for parallelized testing which is not wasm compatible. This should be a dev-dependency but many libra crates have it as a regular dependency. There is a temporary workaround [here](https://altsysrq.github.io/proptest-book/proptest/wasm.html)
* The [`threshold_crypto`](https://crates.io/crates/threshold_crypto) crate has wasm incompatible dependencies: `errno` and `memsec`. `errno` is not directly used but may be present for side-effects. I temporarily [forked](https://github.com/jstarry/threshold_crypto) the repo to get around these issues
* The [`memsec`](https://crates.io/crates/memsec) crate is a Rust port of `libsodium` and the only method used in `threshold_crypto` is the `memzero` function.
  * This function is used to zero memory such that the compiler does not elide those calls. The crate can be used if its `"use_os"` feature is disabled.
  * It can also just be replaced with [`zeroize`](https://crates.io/crates/zeroize) which is wasm compatible and likely to be used throughout the dalek crates: [PR Discussion](https://github.com/dalek-cryptography/curve25519-dalek/pull/258)
* The `logger` dependency needs to be optional or restructured to be able to disable IO/threading functionality.

##### Nightly is required for at least one crate:
* `ed25519-dalek` 

##### Blocker:
* I'm getting really opaque Rust compiler errors even after resolving all of the above issues. I have to use a nightly from April to even get a stacktrace. We'll have to submit a bug report and get help from the compilers team because I'm at a loss as to how to resolve this:

```
$ RUST_BACKTRACE=full cargo +nightly-2019-04-15 build --target wasm32-unknown-unknown
```

```
thread 'rustc' panicked at 'src/librustc_resolve/resolve_imports.rs:916: inconsistent resolution for an import', src/librustc/util/bug.rs:37:26
[...]

Issue #3693, opened 2019-04-09
Research Rust <-> WASM for solana-web3.js


mvines:
The current [solana-web3.js](https://github.com/solana-labs/solana-web3.js) implementation is completely disjoint from [sdk/](https://github.com/solana-labs/solana/tree/master/sdk), which means we need to continually play catchup to transaction format changes, and will need to reimplement new features such as the FeeCalculator in Javascript.

Let's experiment with goodies like https://rustwasm.github.io/docs/wasm-bindgen/ to avoid this duplication, and ideally automatically reflect the same sdk/ API surface into the Javascript world for web and node.js users.

CriesofCarrots:
Yes, please. I'm moving in this direction already, so I will assign myself for the time being.

jstarry:
#### Proposal
Use `wasm_bindgen` to generate JS bindings for loading programs and serialization (tx, instruction, validator info, etc) and use the `wasm-pack` tool to produce a new NPM package `@solana/wasm`. Refactor `@solana/web3.js` to become a hybrid of js and wasm by asynchronously delegating to the `@solana/wasm` package under the hood for selected operations.

#### Details
- NPM packages with `.wasm` files need to be asynchronously imported so all methods which leverage `@solana/wasm`  will have to be `async`.
- `wasm_bindgen` attributes can be added directly to the `solana-sdk` crate but we need a new wrapper crate `solana-sdk-wasm` of lib-type `cdylib` to play nicely with `wasm-pack`
- `wasm_bindgen` generates jsdocs automatically as well as typescript definition files. We could use a tool to generate a flow def from TS?
- `wasm_bindgen` cannot be added to traits so some trait impls will have to be re-exported for binding
```rust
#[cfg(target_arch = "wasm32")]
#[wasm_bindgen(js_name = toString)]
pub fn wasm_to_string(&self) -> String {
    self.to_string()
}
```

#### Releases
- Release `@solana/wasm` along with `solana-sdk`  crate releases
- Potentially rely on Dependabot to automatically bump `@solana/wasm` 

#### References
- [`wasm-bindgen` book](https://rustwasm.github.io/docs/wasm-bindgen/)
- [`rustwasm` book](https://rustwasm.github.io/docs/book/)
- [`wasm-pack` book](https://rustwasm.github.io/docs/wasm-pack/)
- [`rustwasm` project templates](https://rustwasm.github.io/docs/book/reference/project-templates.html)

#### Bundling Research
- Assumes that
[...]

PR #12383, opened 2020-09-22
WIP: Web3.js/signing performance

bartosz-lipinski:
#### Problem

Right now, within web3.js, we are using a javascript port of [tweetnacl](https://github.com/dchest/tweetnacl-js), which result in performance degradation. 

As proposed in the https://github.com/solana-labs/solana-web3.js/issues/974, I have created a new package @solana/wasm using rust and compile it using [wasm-pack](https://github.com/rustwasm/wasm-pack) into WebAssembly. 

This PR is also an initial prototype that could address:
* "Research Rust <-> WASM for solana-web3.js" https://github.com/solana-labs/solana/issues/3693
* "Add support for creating libra transactions in a Wasm SDK"  https://github.com/solana-labs/solana/issues/5418

#### Summary of Changes

This is still a work in progress but wanted to start a discussion around interface changes. 

Main points of concern:
* `@solana/web3.js` will now requite clients to await the initialization function `waitReady` - I decided to create a single initialization function instead of converting every available method that we have into Promise.
* Existing client would also need to change their webpack configuration to ensure that the wasm file is properly loaded (I created Babel transformation and Webpack plugin to make that transition easier) - please take a look at the example: `wasm/examples/webpack`
* I needed to create a custom rollup plugin to build an isomorphic library that could be used in both Web (webpack) and Node.JS environment. This is primarily related to how the wasm-pack generates code used as a wasm npm package.  It is possible to simplify our build process at the expense of maintaining two separate packages, one for Node.js and one for the Web. 

Smaller changes:
* Use `customize-cra` to change CRA
* Update unit test with `beforeAll` method and move some initialization code from `describe` to `before*` method per [jest doc](https://jestjs.io/docs/en/setup-teardown#order-of-execution-of-describe-and-test-blocks)
* 

TODO:
- [x] Remove any references to `tweetnacl`
    - [x] Implement missing methods in Rust
- [x] Add benchmark unit-tests to verify performance improvements - based on the initial benchmarks I am expecting a 3-4x improvement in singing performance 
- [ ] Finish IIFE package setup - right now ESM and CJS works
- [ ] Add plain old web example with just scripts (no-bundlers)
- [x] Add readme to @solana/wasm and setup linter and unit-tests
- [x] Investigating linking with solana-core to share code
- [ ] Add build optimization to reduce the size of the wasm file

Fixes #
https://github.com/solana-labs/solana-web3.js/issues/974
https://github.com/solana-labs/solana/issues/5418


bartosz-lipinski:
Looks like I cannot reference `solana-sdk` directly because it is dependent on `net2` which doesn't support WASM as a compilation target. If we want to increase code sharing between wasm and code/sdk we would need to somehow decouple those references. 

Two targets that we would need to support: `wasm32-unknown-unknown` and `wasm32-wasi`

bartosz-lipinski:
@jstarry, @sakridge after running a basic benchmark on sign method it looks like we will be able to achieve on avg. 20-25x performance improvement.

jstarry:
Hey @bartosz-lipinski this is really exciting and awesome perf improvements! It's hard to find the time to power through an 8k+ line code review so that's the reason for the review delay.

> This is still a work in progress but wanted to start a discussion around interface changes.

Sounds good, I agree that we should have more discussion about this!

> @solana/web3.js will now requite clients to await the initialization function waitReady - I decided to create a single initialization function instead of converting every available method that we have into Promise.

Yeah, this is the trickiest part of integrating wasm right now. This initialization function is going to be a big migration pain, I think we could think more about the best dev integration experience. No matter what we do, I think that wasm integration will probably result in a major version bump for the web3 sdk. It's hard to imagine this being an incremental change.

bartosz-lipinski:
@jstarry fair enough I will split this PR into two separate ones. The new one will only include @solana/wasm and this one will contain all the changes required to start using it which could be gradual migration.

bartosz-lipinski:
I've created PR with only @solana/wasm (https://github.com/solana-labs/solana/pull/12581) once integrated I will update this one.

PR #9212, opened 2020-04-01
Allow to run testnet-automation.sh directly from PRs

ryoqun:
#### Problem

While doing #9161, I noticed that I can't test testnet automation scripts with my changes in that PR.

#### Summary of Changes

I'm surprised testnet-automation.sh is really solid to this extent; just a small more wiring was needed to fix the problem.

_As a bonus, this paves the way to run preset of testnet runs from PRs directly and automatically._

Well, adequately streamlined, yet invoking bunch of `colo.sh`/`gcp.sh` and `net.sh` is a bit tedious for me just to run normal perf. tests. You know, engineers are lazy.. :p

PR #12390, opened 2020-09-22
Moar rolling upgrade

ryoqun:
#### Problem

testing upgrade version test is hard and required local patches to be merged into stable and beta release channels
and supporting the combination of newer `net/*.sh` and older validator seems tedious and basically unmaintained.

#### Summary of Changes

basic idea is like this

1. git reset --hard `$HEAD_OF_STABLE_OR_WHATEVER_PR_ON_STABLE_BRANCH`
1. build locally
1. scp binaries
1. ./net/start.sh ...
1. git reset --hard `$HEAD_OF_BETA_OR_WHATEVER_PR_ON_STABLE_BRANCH`
1. build locally
1. scp binaries
1. ./net/start.sh ...

This is reboot of #9212 op top of #12337 . #9212 originally [closed without merging because building local version is supported without that pr](https://github.com/solana-labs/solana/pull/9212#issuecomment-607015733). But now we want to support to building multiple versions locally. So I think this justifies the need of this pr.

Fixes #

### DIFF
--- a/multinode-demo/bootstrap-validator.sh
+++ b/multinode-demo/bootstrap-validator.sh
@@ -93,7 +93,6 @@ ledger_dir="$SOLANA_CONFIG_DIR"/bootstrap-validator
 args+=(
   --enable-rpc-exit
   --enable-rpc-set-log-filter
-  --require-tower
   --ledger "$ledger_dir"
   --rpc-port 8899
   --snapshot-interval-slots 200
--- a/multinode-demo/setup.sh
+++ b/multinode-demo/setup.sh
@@ -4,7 +4,7 @@ here=$(dirname "$0")
 # shellcheck source=multinode-demo/common.sh
 source "$here"/common.sh
 
-set -e
+set -ex
 
 rm -rf "$SOLANA_CONFIG_DIR"/bootstrap-validator
 mkdir -p "$SOLANA_CONFIG_DIR"/bootstrap-validator
@@ -48,4 +48,5 @@ default_arg --faucet-lamports 500000000000000000
 default_arg --hashes-per-tick auto
 default_arg --cluster-type development
 
+echo "genesis args: ${args[@]}"
 $solana_genesis "${args[@]}"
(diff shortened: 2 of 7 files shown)


ryoqun:
@t-nelson Yay, this worked nicely: https://buildkite.com/solana-labs/rolling-upgrade-tests/builds/113

with

```
CHANNEL=local
LOCAL_BUILD_BRANCH=pull/12387/head
UPGRADE_CHANNEL=local
UPGRADE_LOCAL_BUILD_BRANCH=v1.3
```

So, the rolling upgrade test passed only when #12387 applied.

ryoqun:
> supporting the combination of newer `net/*.sh` and older validator seems tedious and basically unmaintained.

Well, this pr currently doesn't solve this part

t-nelson:
Hey thanks for taking a stab at this.  I have a somewhat embarrassing hack to get similar.  We can actually avoid the local builds entirely, passing them off as optional buildkite jobs that store the release as artifacts.  Then testnet-automation can resolve and download the right builds at runtime

Gross script: https://github.com/t-nelson/solana/commit/b17c75dadcf8946c71e238284462d69ce7008b09

Buildkite:
Screen Shot 2020-09-22 at 10 39 03

t-nelson:
> > supporting the combination of newer `net/*.sh` and older validator seems tedious and basically unmaintained.
> 
> Well, this pr currently doesn't solve this part

Hehe, true.  I believe this is one of our, perhaps unstated, devops goals, though.  We no intention of actually supporting features back to stable, just that the master/HEAD scripts executing on the stable/HEAD bins should work.  This mostly just means backporting hidden stubs so clap doesn't blow up when parsing the command line.

ryoqun:
>  We can actually avoid the local builds entirely, passing them off as optional buildkite jobs that store the release as artifacts. ...

@t-nelson I see. Glad to see that you had similar one. Yeah, these're very similar. So, it seems that we can polish your separate buildkite job approach. Are you going to polish it at #12337? Alternatively, I can work on polishing it if you're busy with other things. :)

t-nelson:
> @t-nelson I see. Glad to see that you had similar one. Yeah, these're very similar. So, it seems that we can polish your separate buildkite job approach. Are you going to polish it at #12337? Alternatively, I can work on polishing it if you're busy with other things. :)

I was hoping to get back to it after cleaning up the SPL Token CLI (probably EOW), but if you've got room on your plate, fee free to take over!  I'll consolidate my changes into the branch pointed at #12337 (and rip out the junk debugging commits).

PR #12396, opened 2020-09-22
Add RPC notify and banking keys debug

sakridge:
#### Problem

No way to enable debug for notifications or transaction processing.

#### Summary of Changes

Add a `--debug-key` feature to allow the operator to print all transactions which touch certain keys and also to when RPC subscriptions are notified. This can help debug latency issues by passing a flag and enabling a debug mask.

Fixes #

### DIFF
--- a/core/src/rpc_subscriptions.rs
+++ b/core/src/rpc_subscriptions.rs
@@ -17,6 +17,7 @@ use solana_client::{
         RpcResponseContext, RpcSignatureResult, SlotInfo,
     },
 };
+use solana_measure::measure::Measure;
 use solana_runtime::{
     bank::Bank,
     bank_forks::BankForks,
@@ -445,7 +446,7 @@ impl RpcSubscriptions {
         account_subscriptions: Arc<RpcAccountSubscriptions>,
         notifier: &RpcNotifier,
         commitment_slots: &CommitmentSlots,
-    ) {
+    ) -> HashSet<SubscriptionId> {
         let subscriptions = account_subscriptions.read().unwrap();
         check_commitment_and_notify(
             &subscriptions,
@@ -455,7 +456,7 @@ impl RpcSubscriptions {
             Bank::get_account_modified_slot,
             filter_account_result,
             notifier,
-        );
+        )
     }
 
     fn check_program(
@@ -464,7 +465,7 @@ impl RpcSubscriptions {
         program_subscriptions: Arc<RpcProgramSubscriptions>,
         notifier: &RpcNotifier,
         commitment_slots: &CommitmentSlots,
-    ) {
+    ) -> HashSet<SubscriptionId> {
         let subscriptions = program_subscriptions.read().unwrap();
         check_commitment_and_notify(
             &subscriptions,
@@ -474,7 +475,7 @@ impl RpcSubscriptions {
             Bank::get_program_accounts_modified_since_parent,
             filter_program_results,
             notifier,
-        );
+        )
     }
 
     fn check_signature(
@@ -483,7 +484,7 @@ impl RpcSubscriptions {
         signature_subscriptions: Arc<RpcSignatureSubscriptions>,
         notifier: &RpcNotifier,
         commitment_slots: &CommitmentSlots,
-    ) {
+    ) -> HashSet<SubscriptionId> {
         let mut subscriptions = signature_subscriptions.write().unwrap();
         let notified_ids = check_commitment_and_notify(
             &subscriptions,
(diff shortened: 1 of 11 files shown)


runtime/src/bank.rs
```diff
@@ -2015,6 +2022,14 @@ impl Bank {
         let mut tx_count: u64 = 0;
         let err_count = &mut error_counters.total;
         for ((r, _hash_age_kind), tx) in executed.iter().zip(txs.iter()) {
+            if let Some(debug_keys) = &self.transaction_debug_keys {
+                for key in &tx.message.account_keys {
+                    if debug_keys.contains(key) {
+                        info!("result: {:?} tx: {:?}", r, tx);
```
carllin, line 2028: should we also log which slot, this is, i.e. `self.slot`

carllin:
so this will tell us when a bank processed a transaction (which will tell us the signature), should we also add a debug log here: https://github.com/carllin/solana/commit/f92c5bc3288fbbff14975500f66fb2dff2412c9b#diff-a7549f152920d85fb44e6a784b8e5e1fR2498-R2504, to let us know when the signature notification occurs (in case there's some weird latency to notification)

runtime/src/bank.rs
sakridge: I thought we had prints which tell us which slot is processing.

carllin: yeah, but it can be hard sometimes to figure out which slot is being replayed when you have parts of multiple slots being processed at the same time (don't get the "frozen" separator between them)

sakridge: ok, added

sakridge:
> 
> 
> so this: https://github.com/solana-labs/solana/pull/12396/files#diff-a7549f152920d85fb44e6a784b8e5e1fR2026-R2029 will tell us when a bank processed a transaction (which will tell us the signature), should we also add a debug log here: [carllin@f92c5bc#diff-a7549f152920d85fb44e6a784b8e5e1fR2498-R2504](https://github.com/carllin/solana/commit/f92c5bc3288fbbff14975500f66fb2dff2412c9b#diff-a7549f152920d85fb44e6a784b8e5e1fR2498-R2504), to let us know when the signature notification occurs (in case there's some weird latency to notification)

I think this place will give us any visibility into weird latencies and be generic for the various notifications:
https://github.com/solana-labs/solana/pull/12396/files#diff-153526d7b6a2099938f98c6cfd2e98a9R980-R992

core/src/rpc_subscriptions.rs
```diff
@@ -913,46 +922,74 @@ impl RpcSubscriptions {
...
                 signature,
                 bank_forks,
                 signature_subscriptions.clone(),
                 &notifier,
                 &commitment_slots,
+            )
+            .len();
+        }
+        signatures_time.stop();
+        if (num_pubkeys_notified + num_programs_notified + num_signatures_notified) > 0 {
```
CriesofCarrots, line 980: It might be nice to know how long it's taking to iterate through the various current subscriptions, even when no notifications are sent, no? Every slot is arguably too often, but might be nice to sample sometimes.

carllin:
> > so this: https://github.com/solana-labs/solana/pull/12396/files#diff-a7549f152920d85fb44e6a784b8e5e1fR2026-R2029 will tell us when a bank processed a transaction (which will tell us the signature), should we also add a debug log here: [carllin@f92c5bc#diff-a7549f152920d85fb44e6a784b8e5e1fR2498-R2504](https://github.com/carllin/solana/commit/f92c5bc3288fbbff14975500f66fb2dff2412c9b#diff-a7549f152920d85fb44e6a784b8e5e1fR2498-R2504), to let us know when the signature notification occurs (in case there's some weird latency to notification)
> 
> I think this place will give us any visibility into weird latencies and be generic for the various notifications:
> https://github.com/solana-labs/solana/pull/12396/files#diff-153526d7b6a2099938f98c6cfd2e98a9R980-R992

Ah ok, I think I see. Walking through a common scenario. If we wanted to see why a "Single" confirmation for a signature is taking a long time to notify:

We have logging to see:
1. When the subscription came into RPC
2. When the signature was processed in a specific slot.
3. When the confirmation notification for `Single` was processed for that slot in `notify_accounts_programs_signatures`.

Which should give us a good end-to-end indication of what's happening

core/src/rpc_subscriptions.rs
sakridge: how about a time cutoff.. it can log if it takes over 10ms

CriesofCarrots: Sounds good

CriesofCarrots (approved):
lgtm

carllin (approved):

Issue #12273, opened 2020-09-16
explorer: Token transfer instructions don't display which token is transfered


jstarry:
#### Problem
When viewing a transaction involving a token transfer, it's difficult to know which token is being transferred because SPL Token "Transfer" instructions don't include the mint address ("Transfer2" does). 

Example:

The following transaction displays an instruction which transfers wrapped ETH:
https://explorer.solana.com/tx/afPaFtbXfK8iE5D6P3F2KiPC917EfCq8yA6awT9q73iCo3MU8kn6UNd8sy3K77wZda7S9twvbMrPL5BUL2ARNNV
We know that because we can inspect the source account:
https://explorer.solana.com/address/EgrroRsUDmwDLeWFd1zZ4Yej7Jhu6yYqUEwT2RLgTdQW

#### Proposed Solution
Display which token is being transferred and include the token unit.

jstarry:
This shouldn't be too bad because we already fetch the mint account to determine the decimals to use in the transfer amount. @oJshua are you interested in taking this on?

oJshua:
@jstarry yep, no problem.

PR #12398, opened 2020-09-22
add token type on token instructions + symbol on amount

oJshua:
#### Problem
When viewing a transaction involving a token transfer, it's difficult to know which token is being transferred because SPL Token "Transfer" instructions don't include the mint address ("Transfer2" does). 

Example:

The following transaction displays an instruction which transfers wrapped ETH:
https://explorer.solana.com/tx/afPaFtbXfK8iE5D6P3F2KiPC917EfCq8yA6awT9q73iCo3MU8kn6UNd8sy3K77wZda7S9twvbMrPL5BUL2ARNNV
We know that because we can inspect the source account:
https://explorer.solana.com/address/EgrroRsUDmwDLeWFd1zZ4Yej7Jhu6yYqUEwT2RLgTdQW

#### Proposed Solution
Display which token is being transferred and include the token unit.

Fixes https://github.com/solana-labs/solana/issues/12273

### DIFF
--- a/explorer/src/components/instruction/token/TokenDetailsCard.tsx
+++ b/explorer/src/components/instruction/token/TokenDetailsCard.tsx
@@ -19,6 +19,8 @@ import {
 } from "providers/accounts";
 import { normalizeTokenAmount } from "utils";
 import { reportError } from "utils/sentry";
+import { useCluster } from "providers/cluster";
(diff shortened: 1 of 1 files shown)


jstarry (approved):
So much nicer!

Screen Shot 2020-09-23 at 11 54 29 AM

PR #12402, opened 2020-09-23
shares gossip thread-pool across gossip service threads

behzadnouri:
#### Problem
We want to parallelize the work in build_crds_filters using rayon thread
pools. But the gossip-work threads:
https://github.com/solana-labs/solana/blob/afd9bfc45/core/src/cluster_info.rs#L2330-L2334
are not usable for this purpose, because they are created in the
ClusterInfo::listen thread:
https://github.com/solana-labs/solana/blob/afd9bfc45/core/src/gossip_service.rs#L54
whereas build_crds_filter is invoked in ClusterInfo::gossip thread:
https://github.com/solana-labs/solana/blob/afd9bfc45/core/src/gossip_service.rs#L61

#### Summary of Changes
~~This commit creates the thread-pool in GossipService::new :~~
~~https://github.com/solana-labs/solana/blob/afd9bfc45/core/src/gossip_service.rs#L31~~
This commit moves the gossip-work ThreadPool from ClusterInfo::listen:
https://github.com/solana-labs/solana/blob/afd9bfc45/core/src/cluster_info.rs#L2330-L2334
to ClusterInfo::new as a new field in the ClusterInfo struct:
https://github.com/solana-labs/solana/blob/afd9bfc45/core/src/cluster_info.rs#L248-L260
so that it can be shared between listen and gossip threads:
https://github.com/solana-labs/solana/blob/afd9bfc45/core/src/gossip_service.rs#L54-L67

### DIFF
--- a/core/src/cluster_info.rs
+++ b/core/src/cluster_info.rs
@@ -36,7 +36,7 @@ use core::cmp;
 use itertools::Itertools;
 use rayon::iter::IntoParallelIterator;
 use rayon::iter::ParallelIterator;
-use rayon::ThreadPool;
+use rayon::{ThreadPool, ThreadPoolBuilder};
 use solana_ledger::staking_utils;
 use solana_measure::measure::Measure;
 use solana_measure::thread_mem_usage;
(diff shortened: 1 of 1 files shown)


sakridge (dismissed):
lgtm

PR #12404, opened 2020-09-23
Remove transaction encoding from storage layer

jstarry:
#### Problem
Transaction encoding is done in the storage layer (bigtable / blockstore) when it really only needs to be done at the api layer (RPC). The current implementation couples the storage types with "UI"/"Encoded" types which adds extra data conversion overhead that isn't necessary. It also does unnecessary transaction serialization

Context: This makes it easier to separate API invoked instruction types from storage invoked instruction types here: https://github.com/solana-labs/solana/pull/12311

#### Summary of Changes
- Introduce new `Encoded..` flavors of `ConfirmedBlock`, `ConfirmedTransaction`, etc.
- Remove all encoding logic from blockstore and bigtable libraries

Fixes #

### DIFF
--- a/client/src/rpc_client.rs
+++ b/client/src/rpc_client.rs
@@ -35,7 +35,7 @@ use solana_sdk::{
     transaction::{self, Transaction},
 };
 use solana_transaction_status::{
-    ConfirmedBlock, ConfirmedTransaction, TransactionStatus, UiTransactionEncoding,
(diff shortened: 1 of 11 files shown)


jstarry:
Adding backports since this is a prereq for surfacing CPI's

mvines (approved):
Looks much cleaner, thanks!

jstarry:
skipped the v1.2 backport

PR #12194, opened 2020-09-11
Fix rooted accounts cleanup, simplify locking

carllin:
#### Problem
1) There were some race conditions between accounts cleanup and accounts store
2) Cleanup logic was divided into too many areas getting very complicated/hard to understand

#### Summary of Changes
1) No more cleanup of old rooted storage entries in store(), store() now only purges storage entries from the same slot 
2) Cleaning up rooted storage entries only occurs in clean() and do_shrink_slots(), via a single `handle_reclaims()` function
3) `handle_reclaims` no longer blocks the store() pipe;ine, so it will always force clean, removing the need for the dead slots counter

Fixes #

PR #12408, opened 2020-09-23
Move dropping AppendVecs outside lock

carllin:
#### Problem
Dropping AppendVecs incurs a flush, which can take a long time on GCE hardware, from several hundred ms to a few seconds for large batches of cleans. This blocks any locks that are held during these drops

#### Summary of Changes
Move drop outside of locks, add some timing metrics

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -1327,14 +1327,14 @@ impl AccountsDB {
 
         let mut remove_storages_elapsed = Measure::start("remove_storages_elapsed");
         for slot in non_roots {
-            let slot_removed_storages = storage.0.remove(&slot);
-            total_removed_storage_entries +=
-                slot_removed_storages.as_ref().map(|x| x.len()).unwrap_or(0);
-            total_removed_bytes += slot_removed_storages
-                .as_ref()
-                .map(|x| x.values().map(|i| i.accounts.capacity()).sum())
-                .unwrap_or(0);
-            all_removed_slot_storages.push(slot_removed_storages);
+            if let Some(slot_removed_storages) = storage.0.remove(&slot) {
+                total_removed_storage_entries += slot_removed_storages.len();
+                total_removed_bytes += slot_removed_storages
+                    .values()
+                    .map(|i| i.accounts.capacity())
+                    .sum::<u64>();
+                all_removed_slot_storages.push(slot_removed_storages);
+            }
         }
         remove_storages_elapsed.stop();
         drop(storage);
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -856,9 +856,10 @@ impl AccountsDB {
         trace!("shrink_stale_slot: slot: {}", slot);
 
         let mut stored_accounts = vec![];
+        let mut storage_read_elapsed = Measure::start("storage_read_elapsed");
         {
-            let storage = self.storage.read().unwrap();
-            if let Some(stores) = storage.0.get(&slot) {
+            let slot_storages = self.storage.read().unwrap().0.get(&slot).cloned();
+            if let Some(stores) = slot_storages {
                 let mut alive_count = 0;
                 let mut stored_count = 0;
                 for store in stores.values() {
@@ -897,7 +898,9 @@ impl AccountsDB {
                 }
             }
         }
+        storage_read_elapsed.stop();
 
+        let mut index_read_elapsed = Measure::start("index_read_elapsed");
         let alive_accounts: Vec<_> = {
             let accounts_index = self.accounts_index.read().unwrap();
             stored_accounts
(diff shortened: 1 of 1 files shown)


runtime/src/accounts_db.rs
```diff
@@ -952,29 +965,78 @@ impl AccountsDB {
...
+                    store.count() > 0
+                });
             }
+            start.stop();
+            write_storage_elapsed = start.as_ms();
         }
+        rewrite_elapsed.stop();
 
+        let mut drop_storage_entries_elapsed = Measure::start("drop_storage_entries_elapsed");
+        drop(dead_storages);
```
carllin, line 1016: moved drop here

runtime/src/accounts_db.rs
```diff
@@ -1255,10 +1317,57 @@ impl AccountsDB {
...
         }
+        remove_storages_elapsed.stop();
+        drop(storage);
+
+        let num_slots_removed = all_removed_slot_storages.len();
+
+        let mut drop_storage_entries_elapsed = Measure::start("drop_storage_entries_elapsed");
+        // Backing mmaps for removed storages entries explicitly dropped here outside
+        // of any locks
+        drop(all_removed_slot_storages);
```
carllin, line 1347: moved drop here

runtime/src/accounts_db.rs
```diff
@@ -1255,10 +1317,57 @@ impl AccountsDB {
...
+
+        let mut all_removed_slot_storages = vec![];
+        let mut total_removed_storage_entries = 0;
+        let mut total_removed_bytes = 0;
+
+        let mut remove_storages_elapsed = Measure::start("remove_storages_elapsed");
         for slot in non_roots {
-            storage.0.remove(&slot);
+            let slot_removed_storages = storage.0.remove(&slot);
+            total_removed_storage_entries +=
```
ryoqun, line 1331: nits: `if let Some(slot_removed_storages) = ...` would make the code a bit cleaner, perhaps?

runtime/src/accounts_db.rs
```diff
@@ -1255,10 +1317,57 @@ impl AccountsDB {
...
+                slot_removed_storages.as_ref().map(|x| x.len()).unwrap_or(0);
+            total_removed_bytes += slot_removed_storages
+                .as_ref()
+                .map(|x| x.values().map(|i| i.accounts.capacity()).sum())
```
ryoqun, line 1335: nits: I think we could write this as `flat_map(...).sum()`.

runtime/src/accounts_db.rs
```diff
@@ -1255,10 +1317,57 @@ impl AccountsDB {
...
+
+        let num_slots_removed = all_removed_slot_storages.len();
+
+        let mut drop_storage_entries_elapsed = Measure::start("drop_storage_entries_elapsed");
+        // Backing mmaps for removed storages entries explicitly dropped here outside
+        // of any locks
+        drop(all_removed_slot_storages);
+        drop_storage_entries_elapsed.stop();
+
+        datapoint_info!(
```
ryoqun, line 1350: (just a note): I think the frequency of submissions of this new metrics should be safe because the freq. of `::purge_slots()` is bound to the slot rate (i.e. 400ms).

runtime/src/accounts_db.rs
ryoqun: if this is taking considerable time, how about handing these over AccountsBackgroundService? I think `drop()`-ing them can really be delayed as much as possible. :)

ryoqun (approved):
lgtm with nits! Thanks for polishing the shrinking mechanism.

runtime/src/accounts_db.rs
carllin: @ryoqun, with this PR: https://github.com/solana-labs/solana/pull/12194, all dropping will be done in AccountsBackgroundService, b/c clean + shrink are handled there.

runtime/src/accounts_db.rs
```diff
@@ -952,29 +965,78 @@ impl AccountsDB {
...
+            write_storage_elapsed = start.as_ms();
         }
+        rewrite_elapsed.stop();
 
+        let mut drop_storage_entries_elapsed = Measure::start("drop_storage_entries_elapsed");
+        drop(dead_storages);
+        drop_storage_entries_elapsed.stop();
+
+        datapoint_info!(
+            "do_shrink_slot_time",
```
sakridge, line 1020: I would do `us` for the timings

sakridge (approved):

runtime/src/accounts_db.rs
carllin: yup definitely, fixed!

runtime/src/accounts_db.rs
carllin: switched over!

PR #12414, opened 2020-09-23
Explorer: The length of history doesn't change after Refresh.

### LINKED ISSUE #11934: explorer: Account history is cut short when refreshing
#### Problem
After loading more account history, clicking the "refresh" button will reset the already loaded history

Steps to reproduce
1. Open https://explorer.solana.com/address/11111111111111111111111111111111?cluster=devnet
2. Click "Load more"
3. Click "Refresh"
4. Observe that the extended history is no longer visible

#### Proposed Solution
Refresh account history while retaining previously fetched history

diman007:
It's simple solution to fix #11934
If your click "Load more" 3 times, then after refreshing your will see 100 detail rows of history (25 - from page load + 25 * 3 from 'load more')

### DIFF
--- a/explorer/src/providers/accounts/history.tsx
+++ b/explorer/src/providers/accounts/history.tsx
@@ -166,7 +166,9 @@ export function useFetchAccountHistory() {
           limit: 25,
         });
       } else {
-        fetchAccountHistory(dispatch, pubkey, cluster, url, { limit: 25 });
+        fetchAccountHistory(dispatch, pubkey, cluster, url, { 
+          limit: before?.data?.fetched.length || 25 
+        });
       }
     },
     [state, dispatch, cluster, url]


explorer/src/providers/accounts/history.tsx
```diff
@@ -166,7 +166,9 @@ export function useFetchAccountHistory() {
           limit: 25,
         });
       } else {
-        fetchAccountHistory(dispatch, pubkey, cluster, url, { limit: 25 });
+        fetchAccountHistory(dispatch, pubkey, cluster, url, { 
+          limit: before?.data?.fetched.length || 25
```
jstarry, line 170: This isn't the fix I was looking for. We don't need to refetch the transactions that were already fetched

jstarry:
@diman007 thanks for looking into this issue! When reviewing your PR I looked a little closer at the code and fixed it myself. Let me know if you would like to take on a different task, I'd be happy to find one for you 👍

explorer/src/providers/accounts/history.tsx
diman007: Ok. Anyway, I was happy to help)

PR #12415, opened 2020-09-23
chore: make `lint` for checking, `lint:fix` for fixing

joncinque:
#### Problem

`npm run lint` actually does fixing through prettier, and `npm run lint:fix` does too, which is not very clear.

#### Summary of Changes

`npm run lint` now only does checks.

Fixes #


jstarry:
look good but the commit message needs to be fixed.. should prefix with "chore: <lower case letter>..." (chore will forgo the patch release)

joncinque:
Noted, updated the commit to reflect that

jstarry (approved):

PR #12422, opened 2020-09-23
adds an atomic variant of the bloom filter

behzadnouri:
#### Problem
For crds_gossip_pull, we want to parallelize build_crds_filters, which
requires concurrent writes to bloom filters.

#### Summary of Changes
This commit implements a variant of the bloom filter which uses atomics
for its bits vector, and so is thread-safe.

### DIFF
--- a/runtime/benches/bloom.rs
+++ b/runtime/benches/bloom.rs
@@ -3,7 +3,8 @@
 extern crate test;
 use bv::BitVec;
 use fnv::FnvHasher;
-use solana_runtime::bloom::{Bloom, BloomHashIndex};
+use rand::Rng;
+use solana_runtime::bloom::{AtomicBloom, Bloom, BloomHashIndex};
 use solana_sdk::{
     hash::{hash, Hash},
     signature::Signature,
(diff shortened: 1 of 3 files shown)


runtime/src/bloom.rs
```diff
@@ -121,9 +122,59 @@ impl<T: AsRef<[u8]>> BloomHashIndex for T {
...
+            _phantom: PhantomData::default(),
+        }
+    }
+}
+
+impl<T: BloomHashIndex> AtomicBloom<T> {
+    pub fn add(&self, key: &T) {
+        for k in &self.keys {
+            let pos = key.hash_at_index(*k) % self.num_bits;
+            let index = pos >> 6;
```
carllin, line 152: Just for future clarity, could we add some comments to these two lines, something like:

pos >> 6 divides by 64 to figure out which of the Atomic64 bit chunks we need to modify

(pos & 63) is equivalent to mod 63 so that we can find the index of the bit within the Atomic64 to modify

runtime/benches/bloom.rs
```diff
@@ -97,3 +98,48 @@ fn bench_sigs_hashmap(bencher: &mut Bencher) {
...
+    let mut rng = rand::thread_rng();
+    let hash_values: Vec<_> = std::iter::repeat_with(|| Hash::new_rand(&mut rng))
+        .take(1200)
+        .collect();
+    let mut fail = 0;
+    bencher.iter(|| {
+        let bloom: AtomicBloom<_> = Bloom::random(1287, 0.1, 7424).into();
+        // Intentionally not using parallelism here, so that this and above
+        // benchmark only compare the bit-vector ops.
+        // For benchmarking the parallel code, change bellow for loop to:
```
carllin, line 133: curious, how did this compare to the above `bench_add_hash`?

carllin:
cool stuff! Interested in knowing how it performs when integrated with: https://github.com/solana-labs/solana/pull/12360/files#diff-8b2c295066af153b1108d9db32997509R426-R429 if you have the numbers

carllin (approved):

runtime/src/bloom.rs
behzadnouri: done! `pos & 63` is equivalent to mod 64 though.

runtime/benches/bloom.rs
behzadnouri: Even without parallism, the atomic one is faster:
```
test bench_add_hash        ... bench:     153,569 ns/iter (+/- 6,109)
test bench_add_hash_atomic ... bench:     135,103 ns/iter (+/- 6,918)
```

With `par_iter`:
```
test bench_add_hash        ... bench:     151,852 ns/iter (+/- 9,219)
test bench_add_hash_atomic ... bench:      65,663 ns/iter (+/- 4,359)
```

behzadnouri:
> cool stuff! Interested in knowing how it performs when integrated with: https://github.com/solana-labs/solana/pull/12360/files#diff-8b2c295066af153b1108d9db32997509R426-R429 if you have the numbers

For the benchmark added there, atomics will be ~40% faster than the mutex/lock version. So the total speed-up compared to current master is like 5x. (obviously on my machine)

runtime/src/bloom.rs
carllin: ah yeah :)

mvines:
Do we strictly need this on v1.2?  Looking to wind down v1.2 pretty hard now, blockers only at this point

behzadnouri:
> Do we strictly need this on v1.2? Looking to wind down v1.2 pretty hard now, blockers only at this point

it will be needed when we want to backport https://github.com/solana-labs/solana/pull/12360

mvines:
> > Do we strictly need this on v1.2? Looking to wind down v1.2 pretty hard now, blockers only at this point
> 
> it will be needed when we want to backport #12360

Let's not backport #12360 either then.   We'll start moving mainnet-beta up to v1.3 next week

behzadnouri:
> > > Do we strictly need this on v1.2? Looking to wind down v1.2 pretty hard now, blockers only at this point
> > 
> > 
> > it will be needed when we want to backport #12360
> 
> Let's not backport #12360 either then. We'll start moving mainnet-beta up to v1.3 next week

sure, removed v1.2 tag

PR #12434, opened 2020-09-24
introduce RpcPerfSample and modify getPerformanceSamples output

oJshua:
#### Problem
`getRecentPerformanceSamples` currently outputs a vector of ordered pairs `(slot, PerfSample)`. This isn't a common pattern in the rpc api, and it would be better if the results were flattened (i.e. a list of objects).

#### Summary of Changes
Move slot number into RpcPerfSample.

### DIFF
--- a/client/src/rpc_response.rs
+++ b/client/src/rpc_response.rs
@@ -275,6 +275,15 @@ pub struct RpcConfirmedTransactionStatusWithSignature {
     pub memo: Option<String>,
 }
 
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct RpcPerfSample {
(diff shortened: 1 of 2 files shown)


client/src/rpc_response.rs
```diff
@@ -275,6 +275,15 @@ pub struct RpcConfirmedTransactionStatusWithSignature {
     pub memo: Option<String>,
 }
 
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
```
CriesofCarrots, line 279: Another good reason for the Rpc struct!

CriesofCarrots (approved):
Adding v1.3 so this rolls out to mainnet-beta with the rest of your work

Issue #12419, opened 2020-09-23
web3.js: support getRecentPerformanceSamples rpc


oJshua:
#### Problem
The `getRecentPerformanceSamples` rpc was introduced in 1.3 and needs an associated call in solana-web3.js. 

#### Proposed Solution
Implement `getRecentPerformanceSamples` in solana-web3.js.

PR #12442, opened 2020-09-24
feat: introduce getRecentPerformanceSamples rpc

oJshua:
#### Problem
The `getRecentPerformanceSamples` rpc was introduced in 1.3 and needs an associated call in solana-web3.js. 

#### Proposed Solution
Implement `getRecentPerformanceSamples` in solana-web3.js.

Fixes https://github.com/solana-labs/solana/issues/12419

### DIFF
--- a/web3.js/src/connection.js
+++ b/web3.js/src/connection.js
@@ -506,7 +506,7 @@ type ConfirmedBlock = {
 };
 
 /**
- * A PerfSample (performance sample)
+ * A performance sample
  *
  * @typedef {Object} PerfSample
  * @property {number} slot Slot number of sample
@@ -2361,9 +2361,9 @@ export class Connection {
    * @return {Promise<Array<PerfSample>>}
    */
   async getRecentPerformanceSamples(
-    limit: ?Number,
+    limit: ?number,
   ): Promise<Array<PerfSample>> {
-    const args = this._buildArgs([limit]);
+    const args = this._buildArgs(limit ? [limit] : []);
     const unsafeRes = await this._rpcRequest(
       'getRecentPerformanceSamples',
       args,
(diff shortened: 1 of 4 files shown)


oJshua:
Just needs a test.

jstarry:
Looks like the CI failure is from some upstream RPC changes to the `getVersion` method

web3.js/src/connection.js
```diff
@@ -1178,6 +1194,20 @@ const GetRecentBlockhashAndContextRpcResult = jsonRpcResultAndContext(
   }),
 );
 
+/*
+ * Expected JSON RPC response for "getRecentPerformanceSamples" message
+ */
+const GetRecentPerformanceSamplesRpcResult = jsonRpcResult(
+  struct.array([
+    struct.object({
```
jstarry, line 1202: nit: let's use `struct.pick({` in case we add more perf stats in the future

web3.js/test/connection.test.js
```diff
@@ -1279,6 +1279,69 @@ test('get supply', async () => {
   expect(supply.nonCirculatingAccounts.length).toBeGreaterThan(0);
 });
 
+test('get performance samples', async () => {
+  if (!mockRpcEnabled) {
```
jstarry, line 1283: I personally prefer live tests over the mock because they serve as nice integration tests for the rpc api as well as smoke tests for breaking changes that require a web3 fix. Is there something about perf sampling that makes it hard to add a live test?

jstarry (commented):
I couldn't find docs for the `getPerformanceSamples` RPC method on edge.docs.solana.com, do you mind adding some? It helps with code reviews like this

web3.js/test/connection.test.js
oJshua: I'll check into this again. At the moment the perf samples API wasn't present during the live test.

jstarry (approved):

PR #12447, opened 2020-09-24
Use publish=false for benches

CriesofCarrots:
#### Problem
Benches don't need to be published to crates.io. In some cases, this happens indirectly because the Cargo.toml does not include a description. But `publish = false` is more clear. Also, it sames 1-2min in solana-secondary CI per crate.

#### Summary of Changes
Use `publish = false` for all benches


CriesofCarrots:
Looks like CI-gate isn’t getting updates, but CI is running here: https://buildkite.com/solana-labs/solana/builds/31425

PR #12450, opened 2020-09-24
Fix clippy too many args warning for process_message

jstarry:
#### Problem
Clippy was ignored

#### Summary of Changes
- Add `ProcessMessageArgs` struct to reduce function argument count for `MessageProcessor::process_message`

Fixes #


mvines:
I think I'd prefer the clippy suppression in this case actually.    It emits a bad code smell signal to everybody who scrolls past, and is thus an opportunity to reflect on what could be done to reduce the args overall.

jstarry:
True, it's more of a sign that code complexity has gone too far than anything else

PR #12451, opened 2020-09-24
feat: add preflightCommitment support

jstarry:
#### Problem
Lack of preflightCommitment support. This means that skipPreflight usage is probably much higher than it needs to be.

#### Summary of Changes
- Add `preflightCommitment` to send options and confirm options

Fixes #


jstarry:
Didn't get around to switching over the loader and tests to use this option over skipPreflight yet. Hope to find time soon though

CriesofCarrots (approved):

PR #12455, opened 2020-09-24
Track inserted repair shreds

sakridge:
#### Problem

Not much visibility into the number of shreds inserted by repair vs. non-repair.

#### Summary of Changes

Add a counter to track which inserted shreds came from the repair port.

Fixes #

### DIFF
--- a/core/src/window_service.rs
+++ b/core/src/window_service.rs
@@ -111,7 +111,7 @@ fn run_check_duplicate(
     Ok(())
 }
 
-fn verify_repair(_shred: &Shred, repair_info: &Option<RepairMeta>) -> bool {
+fn verify_repair(repair_info: &Option<RepairMeta>) -> bool {
     repair_info
         .as_ref()
         .map(|repair_info| repair_info.nonce == DEFAULT_NONCE)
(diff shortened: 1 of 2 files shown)


ledger/src/blockstore.rs
```diff
@@ -789,6 +792,7 @@ impl Blockstore {
                             end_index,
                         },
                     ));
+                    inserted_indicies.push(i);
```
carllin, line 795: typo: `indices`

core/src/window_service.rs
```diff
@@ -140,13 +140,20 @@ where
     let mut i = 0;
     shreds.retain(|shred| (verify_repair(&shred, &repair_infos[i]), i += 1).0);
```
carllin, line 141: I think we should also probably run the `retain` on the indexes in `repair_infos` here as well so that the shreds and `repair_infos` still line up

ledger/src/blockstore.rs
sakridge: fixed

carllin (approved):

PR #12462, opened 2020-09-24
Fix ref-count for multiple stores to the same pubkey in a slot, fixes zero lamport purge detection

carllin:
#### Problem
Storing to the same key multiple times in a slot only keeps one entry alive for that (pubkey, slot)  pair in the AccountsIndex, but increments the ref count multiple times. This means the zero-lamport purge logic will not accurately detect if an account can be purged

#### Summary of Changes
Only increment ref count once per store per pubkey/slot.

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -2064,32 +2064,26 @@ impl AccountsDB {
                 }
                 drop(storage);
                 datapoint_debug!("clean_dead_slots", ("stores", stores.len(), i64));
-                let slot_pubkeys: Vec<(Slot, Vec<Pubkey>)> = {
+                let slot_pubkeys: HashSet<(Slot, Pubkey)> = {
                     self.thread_pool_clean.install(|| {
                         stores
                             .into_par_iter()
                             .map(|store| {
                                 let accounts = store.accounts.accounts(0);
-                                (
-                                    store.slot,
-                                    accounts
-                                        .into_iter()
-                                        .map(|account| account.meta.pubkey)
-                                        .collect::<Vec<Pubkey>>(),
-                                )
+                                accounts
+                                    .into_iter()
+                                    .map(|account| (store.slot, account.meta.pubkey))
+                                    .collect::<HashSet<(Slot, Pubkey)>>()
                             })
-                            .collect()
+                            .reduce(HashSet::new, |mut reduced, store_pubkeys| {
+                                reduced.extend(store_pubkeys);
+                                reduced
+                            })
                     })
                 };
                 let index = self.accounts_index.read().unwrap();
-                let mut cleaned_slot_keys: HashSet<(Slot, Pubkey)> = HashSet::new();
-                for (slot, pubkey_v) in slot_pubkeys {
-                    for pubkey in pubkey_v {
-                        if !cleaned_slot_keys.contains(&(slot, pubkey)) {
-                            index.unref_from_storage(&pubkey);
-                            cleaned_slot_keys.insert((slot, pubkey));
-                        }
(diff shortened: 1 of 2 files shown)


runtime/src/accounts_index.rs
```diff
@@ -159,15 +159,31 @@ impl<'a, T: 'a + Clone> AccountsIndex<T> {
...
-
-            lock.0.fetch_add(1, Ordering::Relaxed);
+            let list = &mut lock.1.write().unwrap();
+            // filter out other dirty entries from the same slot
+            let mut same_slot_previous_updates: Vec<(usize, (Slot, T))> = list
+                .iter()
+                .enumerate()
+                .filter_map(|(i, (f, value))| {
+                    if *f == slot {
+                        Some((i, (*f, value.clone())))
```
ryoqun, line 169: nits: maybe this `.clone()` can be moved inside this `if` clause?: https://github.com/solana-labs/solana/pull/12462/files#diff-ef68c28d9d63e66355c44904d9877825R178

i.e. we can `.clone()` only if there is non-empty `same_slot_previous_updates` as a micro-optimization?

runtime/src/accounts_index.rs
```diff
@@ -159,15 +159,31 @@ impl<'a, T: 'a + Clone> AccountsIndex<T> {
...
-            reclaims.extend(list.iter().filter(|(f, _)| *f == slot).cloned());
-            list.retain(|(f, _)| *f != slot);
-
-            lock.0.fetch_add(1, Ordering::Relaxed);
+            let list = &mut lock.1.write().unwrap();
+            // filter out other dirty entries from the same slot
+            let mut same_slot_previous_updates: Vec<(usize, (Slot, T))> = list
+                .iter()
+                .enumerate()
+                .filter_map(|(i, (f, value))| {
```
ryoqun, line 167: nits: rename `f` => `s`? (That `f` stood for the `fork`).

runtime/src/accounts_index.rs
carllin: agreed, updated!

runtime/src/accounts_index.rs
carllin: sure thing, updated!

runtime/src/accounts_db.rs
```diff
@@ -2064,24 +2064,31 @@ impl AccountsDB {
...
-                                    .collect::<Vec<Pubkey>>()
+                                (
+                                    store.slot,
+                                    accounts
+                                        .into_iter()
+                                        .map(|account| account.meta.pubkey)
+                                        .collect::<Vec<Pubkey>>(),
+                                )
                             })
                             .collect()
```
ryoqun, line 2081: how about `reduce()`-ing here like this, which will be done in parallel by rayon?:

```rust
let pubkeys: HashSet<(Slot, Pubkey)> = 
   ...
   .reduce(|| HashSet<_>::new(), |mut reduced, pubkeys| reduced.extend(pubkeys)

```

so that we can remove the need to construct `cleaned_slot_keys` later?

ryoqun (dismissed):
LGTM with nits, awesome bug fixing. I guess you found this by carefully reading the code to fight with the snapshot bloat issue at #12194 .

I guess this bug doesn't manifest as clear errors. Only this causes is dangling appendvecs, which doesn't reduce the snapshot size as expected.

runtime/src/accounts_db.rs
carllin: updated!

carllin:
@ryoqun I actually found it while writing those snapshot hash mismatch tests :)

Issue #11771, opened 2020-08-21
explorer: Remove dependency on solana beach API


jstarry:
#### Problem
We currently use the solana beach api for:
- real time transaction count
- average tps and blocktime
- tps history

But it's not ideal to have an external dependency that we can't fix quickly if something gets broken

#### Proposed Solution
Build Solana equivalents of these APIs and migrate explorer to use those instead

oJshua:
@jstarry This might be a good one for me to tackle next. Did you envision using the same socket.io-style api (drop in replacement for solana beach) or is there an existing service we might attach this to?

jstarry:
@oJshua we have an [existing jsonrpc pubsub service](https://github.com/solana-labs/solana/blob/master/core/src/rpc_pubsub.rs) which we could add to. Some of these APIs may need some sort of backing time series db for historical calculations, which we don't have a precedent for

oJshua:
Just spoke with @mvines about writing tps averages per minute to RocksDB. Should do the trick.

aditya-manit:
Hey, is there any doc on all the apis provided by solana beach?

mvines:
👋🏼  Solana Beach is provided by Staking Facilities, so separate organization.  As far as I know they haven't published a public API

PR #12463, opened 2020-09-24
Remove Solanabeach dependency from Explorer

oJshua:
#### Problem
We currently use the solana beach api for:
- real time transaction count
- average tps and blocktime
- tps history

But it's not ideal to have an external dependency that we can't fix quickly if something gets broken

#### Summary of Changes
- Poll various RPC endpoints to build up cluster stats
- Render performance samples in 3 timeframes
- Convert state management into using reducers
- Remove socket.io dependency

Fixes https://github.com/solana-labs/solana/issues/11771

**Merge will need to wait until 1.3 is deployed across clusters and new web3 version is pushed**


oJshua:
Adding a timeout / retry button if node doesn't respond.

explorer/src/providers/stats/solanaDashboardInfo.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const status =
+    state.avgBlockTime_1h !== 0 && state.epochInfo.absoluteSlot !== 0
+      ? ClusterStatsStatus.Ready
+      : ClusterStatsStatus.Loading;
+
+  switch (action.type) {
+    case DashboardInfoActionType.SetPerfSamples:
+      const samples = action.data.map((sample) => {
+        return sample.samplePeriodSecs / sample.numSlots;
+      });
```
jstarry, line 63: The average block time stat used to filter out skipped slots. This looks like it's measuring the average time elapsed per slot rather than average block time. Slot time is useful for measuring the progress of an epoch but I think it needs to be coupled with some indication of the skipped slot percentage to give the full picture of how long it takes for a transaction to land.

explorer/src/providers/stats/solanaDashboardInfo.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+      const avgBlockTime_1h =
+        samples.reduce((sum: number, cur: number) => {
+          return sum + cur;
+        }, 0) / 60;
```
jstarry, line 68: What happens if there are fewer (or more) than 60 samples?

explorer/src/providers/stats/solanaDashboardInfo.tsx
```diff
@@ -0,0 +1,96 @@
...
+        status: ClusterStatsStatus.Error,
+      };
+    case DashboardInfoActionType.Reset:
+      return {
+        ...action.data,
+      };
+    default:
+      return {
+        ...state,
+      };
```
jstarry, line 94: If nothing changed, you should return `state` directly in order to avoid causing unnecessary renders

explorer/src/providers/stats/solanaDashboardInfo.tsx
```diff
@@ -0,0 +1,96 @@
...
+}
+
+export type DashboardInfoActionSetPerfSamples = {
+  type: DashboardInfoActionType.SetPerfSamples;
+  data: PerfSample[];
+};
+
+export type DashboardInfoActionSetEpochInfo = {
+  type: DashboardInfoActionType.SetEpochInfo;
+  data: any;
```
jstarry, line 31: Any reason why this isn't typed?

explorer/src/providers/stats/solanaPerformanceInfo.tsx
```diff
@@ -0,0 +1,120 @@
...
+      };
+    default:
+      return {
+        ...state,
+      };
+  }
+}
+
+function downsampleByFactor(series: number[], factor: number) {
+  return series.reduce((result: number[], num: number, i: number) => {
```
jstarry, line 106: This implementation looks fine but I think that an approach that chunks the array would be easier to read

jstarry (commented):
Looks good to me, I'd like to try it out with the web3 changes when those are released!

Main concern is over the change from average block time to average slot time. Maybe we can get by with a copy change in the explorer to address that. But I'm curious if this was intentional

explorer/src/providers/stats/solanaDashboardInfo.tsx
oJshua: This is interesting. I guess I didn't know how average block time was computed. Since we are using the performance samples in this manner, would it make sense for us to filter out the skipped slots on the backend when we write PerfSamples to the blockstore?

explorer/src/providers/stats/solanaDashboardInfo.tsx
oJshua: No. I think it may have been untyped when I moved it over. Maybe not; I'll fix!

explorer/src/providers/stats/solanaDashboardInfo.tsx
oJshua: Just noticing "Epoch time remaining" is off from the Solana Beach version.

explorer/src/providers/stats/solanaPerformanceInfo.tsx
oJshua: I agree! I must have been in a reducing mood that day. I might rewrite in the morning.

explorer/src/providers/stats/solanaDashboardInfo.tsx
jstarry: Oh interesting, Solana Beach now shows "Slot Time" instead of "Block Time". Seems fine for now to do the same in this PR.

> would it make sense for us to filter out the skipped slots on the backend when we write PerfSamples to the blockstore?

Maybe we can add that later. Slot time seems fine for now.

> Just noticing "Epoch time remaining" is off from the Solana Beach version.

On this branch?

jstarry:
I just tried it out and it looks awesome!

explorer/src/providers/stats/solanaPerformanceInfo.tsx
```diff
@@ -0,0 +1,122 @@
...
+    case PerformanceInfoActionType.SetPerfSamples:
+      if (action.data.length < 1) {
+        return state;
+      }
+
+      let short = action.data.map((sample) => {
+        return sample.numTransactions / sample.samplePeriodSecs;
+      });
+
+      const historyMaxTps = Math.max(...short);
```
jstarry, line 68: I think this should be calculated from the final `short`, `medium`, and `long`. Otherwise if the max value happened over 30min ago, it will get downsampled away and the chart max will be exaggerated

oJshua: This actually happens before the short array is resized (so we're dealing with I believe 360 samples). Should be good.

explorer/src/providers/stats/solanaDashboardInfo.tsx
oJshua: Yeah, it's happening in this branch. I will investigate.

explorer/src/providers/stats/solanaPerformanceInfo.tsx
jstarry: Right, I'm saying you should check for the max after the short array is resized. The max is used to determine the chart y axis and your code can pick up a max TPS that won't be charted because it's in the past / will not be visible after downsampling

oJshua: Oh, I misunderstood! Sure thing.

oJshua: Wait, I'm not sure we're on the same page here. The max TPS is being derived from the whole 6 hour history (before all the downsampling, it's 360 samples 6 * 60). Since I'm grabbing the max TPS from largest possible window (6h), this shouldn't get downsampled away. It should represent the highest value over the final short, medium, and long arrays.

oJshua: Okay, I do see one problem. The average of a downsampled max could be lower than the original max TPS. That might be your point lol. I will pull this from the short, medium, long arrays

explorer/src/providers/stats/solanaDashboardInfo.tsx
jstarry: @oJshua are you still looking into this?

oJshua: Yes, so the hourly average block time being provided by Solana Beach is around 10% off from our computed value. I seem to trust our data more. I can monitor when the epoch transition happens and see who is more accurate.

jstarry: Sounds good. I think this PR just needs a change from "Block Time" to "Slot Time" then

jstarry (dismissed):

explorer/src/providers/stats/solanaDashboardInfo.tsx
oJshua: Thanks! Pushed.

PR #5291, opened 2019-07-26
Fix metrics when leader does not report metrics

jstarry:
#### Summary of Changes

*Metrics Reporting*:
* Added `host_id` influx tag to enable aggregating metrics across nodes
* All nodes will now announce how many nodes are in the cluster
* In addition to announcing when they become leader, nodes will also announce when another node becomes leader
* Transaction counts are reported different ways to have both fine-grained and aggregate metrics:
  * From leader as they execute transactions
  * From validators as they replay transactions
  * From the banking stage as it processes transactions (all nodes report this)

*Metrics Dashboard*:
* Both the node count stat and graph now show up to date values as long as one validator has metrics enabled. The median of all reported node counts will be displayed
* Transaction metrics use the reported tx count from the banking stage. The median of counts from all nodes is used
* Vote counts were being double counted before, now we take the median of the processed vote counts from all nodes
* Current leader is calculated from the mode of all reported leaders
* Leader change graph now determines the current slot by taking the median reported slot from all nodes

Fixes #5170

PR #12465, opened 2020-09-25
Update Current Leader query

CriesofCarrots:
#### Problem
The current-leader datapoint submitted by all nodes (added in https://github.com/solana-labs/solana/pull/5291) was removed more than a year ago. The dashboard Current Leader query has been broken since then, and always returns the lowest host_id in `replay_stage-new_leader`

#### Summary of Changes
Fix query

### DIFF
--- a/metrics/scripts/grafana-provisioning/dashboards/cluster-monitor.json
+++ b/metrics/scripts/grafana-provisioning/dashboards/cluster-monitor.json
@@ -136,7 +136,7 @@
           "hide": false,
           "orderByTime": "ASC",
           "policy": "default",
-          "query": "SELECT MODE(last) FROM ( SELECT last(\"leader\") FROM \"$testnet\".\"autogen\".\"replay_stage-new_leader\" WHERE $timeFilter GROUP BY host_id )\n",
+          "query": "SELECT last(\"leader\") FROM \"$testnet\".\"autogen\".\"replay_stage-new_leader\" WHERE $timeFilter ORDER BY \"time\" desc \n",
           "rawQuery": true,
           "refId": "A",
           "resultFormat": "table",


metrics/scripts/grafana-provisioning/dashboards/cluster-monitor.json
```diff
@@ -136,7 +136,7 @@
           "hide": false,
           "orderByTime": "ASC",
           "policy": "default",
-          "query": "SELECT MODE(last) FROM ( SELECT last(\"leader\") FROM \"$testnet\".\"autogen\".\"replay_stage-new_leader\" WHERE $timeFilter GROUP BY host_id )\n",
+          "query": "SELECT last(\"leader\") FROM \"$testnet\".\"autogen\".\"replay_stage-new_leader\" WHERE $timeFilter ORDER BY \"time\" desc \n",
```
jstarry, line 139: Ah, thanks for fixing this! Is there a chance we could keep the "MODE" to ensure one lagging node doesn't influence the metrics?

CriesofCarrots: Sorry, I don't follow how `MODE` would work in this case. If a node isn't reporting metrics or hasn't sent their metrics yet even though it is currently the leader, this query will just return the previous leader's id. I don't see how to get more-current data without re-adding the current-leader datapoint from all nodes. What am I missing?

CriesofCarrots: Obviously it would be better to order by slot desc, but influxdb only permits order by time.

jstarry: From the linked PR:
> In addition to announcing when they become leader, nodes will also announce when another node becomes leader

So since every node announces the leader, we need to filter out any datapoints that were really recent but came from a delinquent node. `MODE` acts as that filter by using the most common reported leader

jstarry: > Obviously it would be better to order by slot desc, but influxdb only permits order by time.

Exactly, since we can't do that. We can't know if the latest datapoint was for an old slot

CriesofCarrots: > So since every node announces the leader, we need to filter out any datapoints that were really recent but came from a delinquent node. `MODE` acts as that filter by using the most common reported leader

The whole point of this PR is that nodes do not announce when another node becomes leader. They only announce when they themselves become leader. So there is only at most one datapoint per slot. If datapoints do share the same time, mode won't show us which is correct.

jstarry: Oh, sorry totally missed this "was removed more than a year ago" part. That sounds good then

jstarry (approved):

metrics/scripts/grafana-provisioning/dashboards/cluster-monitor.json
CriesofCarrots: Awesome, thank you!

PR #12490, opened 2020-09-25
Port BPFLoader2 activation to FeatureSet

mvines:
Along the way delete genesis-programs/ and the Bank's EnteredEpochCallback, significantly cleaning up how built-in programs are added and activated.


cli/Cargo.toml
```diff
@@ -37,6 +37,7 @@ solana-faucet = { path = "../faucet", version = "1.4.0" }
 solana-logger = { path = "../logger", version = "1.4.0" }
 solana-net-utils = { path = "../net-utils", version = "1.4.0" }
 solana-remote-wallet = { path = "../remote-wallet", version = "1.4.0" }
+solana-runtime = { path = "../runtime", version = "1.4.0" }
```
jackcmay, line 40: Would the feature set make more sense in the sdk?

mvines: I did consider that, not totally opposed to it.  I didn't have a strong reason for why FeatureSet should generally be exposed in an SDK at this point, so I went for keeping it more narrowly scoped (at the expense of adding this new cli dependency!).  

Only the two too commits are relevant for this PR.  The rest of this PR is just a clone of #12376 (including what you're commenting on here)

jackcmay: Yup, I know :-)

mvines: There, fixed.  This PR now only contains the BPFLoader activation rework

jackcmay: In order for us to pass the `FeatureSet` to InvokeContext we could either move `FeatureSet` to the sdk, or `InvokeContext` to runtime.  The latter is looking pretty good: https://github.com/solana-labs/solana/pull/12507

mvines:
Ugh, getting stuck on `Bank` AbiExample stuff.    Not sure this is really even needed anymore `Bank` @ryoqun

runtime/src/bank.rs
```diff
@@ -186,6 +188,26 @@ impl CowCachedExecutors {
     }
 }
 
+#[cfg(RUSTC_WITH_SPECIALIZATION)]
+impl AbiExample for Builtin {
```
ryoqun, line 192: :eyes:

core/src/builtins.rs
```diff
@@ -0,0 +1,44 @@
...
+/// Builtin programs activated dynamically by feature
+fn feature_builtins() -> Vec<(Builtin, Pubkey)> {
+    let builtins = vec![(
+        solana_bpf_loader_program!(),
+        feature_set::bpf_loader2_program::id(),
+    )];
+
+    builtins
+        .into_iter()
+        .map(|(b, p)| (Builtin::new(&b.0, b.1, Entrypoint::Loader(b.2)), p))
```
jackcmay, line 35: `declare_builtin` should probably be a struct so we don't have reference it via tuple indexes.  If agreed we can do it a different PR

jackcmay: Or just expand out the map to:
`.map(|((name, id, entrypoint), p)| (Builtin::new(&b.0, b.1, Entrypoint::Loader(b.2)), p))`

runtime/src/bank.rs
```diff
@@ -186,6 +188,26 @@ impl CowCachedExecutors {
...
+        }
+    }
+}
+
+#[derive(Clone)]
+pub struct Builtins {
+    /// Builtin programs that are always available
+    pub genesis_builtins: Vec<Builtin>,
+
+    /// A builtin program activated dynamically by feature
```
jackcmay, line 207: nit: `/// Builtin programs activated dynamically by feature`

core/src/builtins.rs
mvines: > declare_builtin should probably be a struct so we don't have reference it via tuple indexes. If agreed we can do it a different PR

Yeah I agree that'll be much nicer.  I didn't want to take on that churn in this PR

runtime/src/bank.rs
mvines: fixed

mvines:
Looks good otherwise @jackcmay?

jackcmay (approved):
Much cleaner

PR #12456, opened 2020-09-24
Add feature to resolve the spl-token v2 multisig bug

mvines:
Add mechanism that'll allow us to fix the spl-token v2 multisig issue across the existing clusters

PR #12502, opened 2020-09-25
Add copy-on-write executor cache

jackcmay:
#### Problem

Different forks using the same shared executor cache may add/remove entries that are fork specific.

- Load the BPF program but don’t finalize it yet.
- On a minority fork, modify the BPF program with another transaction then finalize it and execute it.  - - CachedExecutors  loads the BPF program from the minority fork.
- Now finalize the BPF program (without the minority fork modification) on the rooted path.  
CachedExecutors will not update the program and continue using the version from the minority fork.

Depending on the ordering in which (2) happens, and it may be different between validators, the BPF program that is executed will differ and could produce different results for the same instructions.

#### Summary of Changes

Wrap the executor cache with a copy-on-write object

Fixes #

### DIFF
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -251,6 +251,9 @@ impl CachedExecutors {
                 .insert(*pubkey, (AtomicU64::new(0), executor));
         }
     }
+    fn remove(&mut self, pubkey: &Pubkey) {
+        let _ = self.executors.remove(pubkey);
+    }
 }
 
 #[derive(Default)]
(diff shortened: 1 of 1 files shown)


mvines:
lgtm, would you mind adding a `CachedExecutors` method that'll allow me to easily delete an entry for me to use in #12456 while you're around these parts?

jackcmay:
> lgtm, would you mind adding a `CachedExecutors` method that'll allow me to easily delete an entry for me to use in #12456 while you're around these parts?

Done

mvines (approved):

PR #12507, opened 2020-09-25
Move process_instruction defs to runtime

jackcmay:
#### Problem

- Bunch of definitions that don't need to be in the sdk
- We are going to need to pipe `FeatureSet` into `InvokeContext` but `FeatureSet` is local to runtime

#### Summary of Changes

These definitions don't need to be in the SDK and moving them into runtime brings them in with their users and allows us to leave `FeatureSet` where it is.

Fixes #

### DIFF
--- a/runtime/src/process_instruction.rs
+++ b/runtime/src/process_instruction.rs
@@ -1,5 +1,3 @@
-#[cfg(RUSTC_WITH_SPECIALIZATION)]
-use solana_sdk::abi_example::AbiExample;
 use solana_sdk::{
     account::{Account, KeyedAccount},
     instruction::{CompiledInstruction, Instruction, InstructionError},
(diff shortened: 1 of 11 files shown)


genesis-programs/src/lib.rs
```diff
@@ -7,10 +7,12 @@ extern crate solana_exchange_program;
 #[macro_use]
 extern crate solana_vest_program;
```
mvines, line 9: Can you rebase on https://github.com/solana-labs/solana/pull/12490 (review for that PR would be great too!).  `genesis-programs/...` is no more!

jackcmay: Done

jackcmay:
@ryoqun Any ideas on how to fix this CI? 

Background, I moved types from the sdk into runtime.  The bank depends on `Executor` which was moved. Rust was happy before because `AbiExample` was. in the same crate as `impl AbiExample for ArcExecutor {`.  How to define an AbiExample for a dynamic trait in this case?

runtime/src/process_instruction.rs
```diff
@@ -0,0 +1,150 @@
...
+    fn execute(
+        &self,
+        program_id: &Pubkey,
+        keyed_accounts: &[KeyedAccount],
+        instruction_data: &[u8],
+        invoke_context: &mut dyn InvokeContext,
+    ) -> Result<(), InstructionError>;
+}
+#[cfg(RUSTC_WITH_SPECIALIZATION)]
+impl AbiExample for Arc<dyn Executor> {
```
jackcmay, line 134: @ryoqun Here is the code that breaks:

```
error[E0117]: only traits defined in the current crate can be implemented for arbitrary types
   --> runtime/src/process_instruction.rs:134:1
    |
134 | impl AbiExample for Arc<dyn Executor> {
    | ^^^^^^^^^^^^^^^^^^^^-----------------
    | |                   |
    | |                   `Arc` is not defined in the current crate
    | impl doesn't use only types from inside the current crate
    |
    = note: define and implement a trait or new type instead

error: aborting due to previous error; 1 warning emitted
```

jackcmay:
@ryoqun Thanks!

PR #12510, opened 2020-09-26
bank: Remove continues from compute_active_feature_set()

mvines:
Post-merge feedback from #12376


ryoqun:
thanks for taking my suggestion!

PR #12513, opened 2020-09-27
retains hash value of outdated responses received from pull requests

behzadnouri:
#### Problem
pull_response_fail_inserts has been increasing:
https://cdn.discordapp.com/attachments/478692221441409024/759096187587657778/pull_response_fail_insert.png
but for outdated values which fail to insert:
https://github.com/solana-labs/solana/blob/a5c3fc14b3/core/src/crds_gossip_pull.rs#L332-L344
https://github.com/solana-labs/solana/blob/a5c3fc14b3/core/src/crds.rs#L104-L108
are not recorded anywhere, and so the next pull request may obtain the
same redundant payload again, unnecessary taking bandwidth.

#### Summary of Changes
This commit holds on to the hashes of failed-inserts for a while, similar
to purged_values:
https://github.com/solana-labs/solana/blob/a5c3fc14b3/core/src/crds_gossip_pull.rs#L380
and filter them out for the next pull request:
https://github.com/solana-labs/solana/blob/a5c3fc14b3/core/src/crds_gossip_pull.rs#L204

### DIFF
--- a/core/src/crds.rs
+++ b/core/src/crds.rs
@@ -105,19 +105,13 @@ impl Crds {
         &self,
         value: CrdsValue,
         local_timestamp: u64,
-    ) -> Option<VersionedCrdsValue> {
+    ) -> (bool, VersionedCrdsValue) {
         let new_value = self.new_versioned(local_timestamp, value);
         let label = new_value.value.label();
-        let would_insert = self
-            .table
-            .get(&label)
-            .map(|current| new_value > *current)
-            .unwrap_or(true);
-        if would_insert {
-            Some(new_value)
-        } else {
-            None
-        }
+        // New value is outdated and fails to insert, if it already exists in
+        // the table with a more recent wallclock.
+        let outdated = matches!(self.table.get(&label), Some(current) if new_value <= *current);
+        (!outdated, new_value)
     }
     /// insert the new value, returns the old value if insert succeeds
     pub fn insert_versioned(
--- a/core/src/crds_gossip_pull.rs
+++ b/core/src/crds_gossip_pull.rs
@@ -29,6 +29,8 @@ use std::ops::Index;
 pub const CRDS_GOSSIP_PULL_CRDS_TIMEOUT_MS: u64 = 15000;
 // The maximum age of a value received over pull responses
 pub const CRDS_GOSSIP_PULL_MSG_TIMEOUT_MS: u64 = 60000;
+// Retention period of hashes of received outdated values.
+const FAILED_INSERTS_RETENTION_MS: u64 = 20_000;
 pub const FALSE_RATE: f64 = 0.1f64;
 pub const KEYS: f64 = 8f64;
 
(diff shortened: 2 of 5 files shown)


core/src/crds.rs
```diff
@@ -105,19 +105,11 @@ impl Crds {
...
-            .table
-            .get(&label)
-            .map(|current| new_value > *current)
-            .unwrap_or(true);
-        if would_insert {
-            Some(new_value)
-        } else {
-            None
-        }
+        let outdated = matches!(self.table.get(&label), Some(current) if new_value <= *current);
```
sakridge, line 111: Kind of find the `matches` way harder to read.

core/src/crds_gossip_pull.rs
```diff
@@ -353,47 +364,65 @@ impl CrdsGossipPull {
...
-                })
-            });
         }
         owners.insert(*from);
         for owner in owners {
             crds.update_record_timestamp(&owner, now);
         }
+        stats.failed_insert += failed_inserts.len();
+        self.purge_failed_inserts(now);
+        self.failed_inserts
```
sakridge, line 403: better to just create it with the timestamp originally, no?

core/src/crds_gossip_pull.rs
```diff
@@ -408,7 +437,9 @@ impl CrdsGossipPull {
         for (value_hash, _insert_timestamp) in &self.purged_values {
             add_value_hash(&value_hash);
         }
-
+        for (value_hash, _) in &self.failed_inserts {
```
sakridge, line 440: Hmm, since the filter is changed, does it need any kind of gating for mixed-network nodes?

sakridge:
Cool, what kind of numbers are you seeing for failed inserts before/after this?

behzadnouri (commented):
> Cool, what kind of numbers are you seeing for failed inserts before/after this?

With a small cluster that I can start on gce, `failed_inserts` are already mostly zero. So I do not see any improvements there. But it might help on mainnet, which sees a lot of failed inserts. It can be because some nodes have fallen behind and they send old values.

behzadnouri:
> > Cool, what kind of numbers are you seeing for failed inserts before/after this?
> 
> With a small cluster that I can start on gce, `failed_inserts` are already mostly zero. So I do not see any improvements there. But it might help on mainnet, which sees a lot of failed inserts. It can be because some nodes have fallen behind and they send old values.

cluster_info_stats2 pull_response_fail_insert_gce40_zoom

With a 40 node gce cluster I see a slight improvement. Though the difference is certainly not big enough to say that this is not just noise. The green line is the average `pull_response_fail_insert`, and the red line is 90% percentile. The left chunk is master branch, and the right chunk is this patch.

sakridge (approved):
I'm a bit curious what the overlap with purged values might be.

carllin (approved):

behzadnouri:
> I'm a bit curious what the overlap with purged values might be.

There can be some overlap, but only on the values which were purged since the pull-request was sent out. For the nodes which see 10k+ failed-inserts, _probably_ that overlap is a small portion.

PR #12514, opened 2020-09-27
Rpc -> proper optimistic confirmation

CriesofCarrots:
#### Problem
Rpc uses the same bank for `Single` and `SingleGossip` commitment levels.

#### Summary of Changes
- Add service to track optimistically confirmed banks (currently being done inside RpcSubscriptions)
- Remove tracking from RpcSubscriptions
- Use service to update OptimisticallyConfirmedBank, which caches a reference to the Bank
- Use OptimisticallyConfirmedBank in Rpc

### DIFF
--- a/core/src/optimistically_confirmed_bank_tracker.rs
+++ b/core/src/optimistically_confirmed_bank_tracker.rs
@@ -1,0 +1,275 @@
+//! The `optimistically_confirmed_bank_tracker` module implements a threaded service to track the
+//! most recent optimistically confirmed bank for use in rpc services, and triggers gossip
+//! subscription notifications
+
+use crate::rpc_subscriptions::RpcSubscriptions;
+use solana_runtime::{bank::Bank, bank_forks::BankForks};
+use solana_sdk::clock::Slot;
+use std::{
+    collections::HashSet,
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        mpsc::{Receiver, RecvTimeoutError},
+        Arc, RwLock,
+    },
+    thread::{self, Builder, JoinHandle},
+    time::Duration,
+};
+
+pub struct OptimisticallyConfirmedBank {
+    pub bank: Arc<Bank>,
+}
+
+impl OptimisticallyConfirmedBank {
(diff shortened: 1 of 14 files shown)


core/src/optimistically_confirmed_bank_tracker.rs
```diff
@@ -0,0 +1,278 @@
...
+                {
+                    let mut w_optimistically_confirmed_bank =
+                        optimistically_confirmed_bank.write().unwrap();
+                    if bank.slot() > w_optimistically_confirmed_bank.bank.slot() {
+                        w_optimistically_confirmed_bank.bank = bank.clone();
+                    }
+                    drop(w_optimistically_confirmed_bank);
+                    subscriptions.notify_gossip_subscribers(slot);
+                } else {
+                    pending_optimistically_confirmed_banks.insert(slot);
```
carllin, line 129: should we only add this to the pending queue if it's greater than the current optimistic bank and the current root?

Edit:  I see the `BankNotification::Root(bank)` clears outdated items from the `pending_optimistically_confirmed_banks`, but may still be good to check in case a root doesn't get set for a while

core/src/optimistically_confirmed_bank_tracker.rs
```diff
@@ -0,0 +1,278 @@
...
+            }
+            BankNotification::Root(bank) => {
+                let root_slot = bank.slot();
+                let mut w_optimistically_confirmed_bank =
+                    optimistically_confirmed_bank.write().unwrap();
+                if root_slot > w_optimistically_confirmed_bank.bank.slot() {
+                    w_optimistically_confirmed_bank.bank = bank;
+                }
+                drop(w_optimistically_confirmed_bank);
+                *pending_optimistically_confirmed_banks = pending_optimistically_confirmed_banks
```
carllin, line 152: I think a `retain` over a `filter` would be more efficient here

core/src/optimistically_confirmed_bank_tracker.rs
```diff
@@ -0,0 +1,278 @@
...
+                    .iter()
+                    .cloned()
+                    .filter(|&s| s > root_slot)
```
carllin, line 155: I think a retain over a filter would be more efficient here

carllin (approved):

PR #12528, opened 2020-09-28
patches bug in Crds::find_old_labels with pubkey specific timeout

behzadnouri:
#### Problem
Current code only returns values which are expired based on the default
timeout. Example from the added unit test:
  - value inserted at time 0
  - pubkey specific timeout = 1
  - default timeout = 3

Then at now = 2, the value is expired, but the function fails to return
the value because it compares with the default timeout.

#### Summary of Changes
Compares local-timestamp against pubkey specific timeout when available.

### DIFF
--- a/core/src/crds.rs
+++ b/core/src/crds.rs
@@ -185,18 +185,14 @@ impl Crds {
         now: u64,
         timeouts: &HashMap<Pubkey, u64>,
     ) -> Vec<CrdsValueLabel> {
-        let min_ts = *timeouts
+        let default_timeout = *timeouts
             .get(&Pubkey::default())
             .expect("must have default timeout");
         self.table
             .iter()
             .filter_map(|(k, v)| {
-                if now < v.local_timestamp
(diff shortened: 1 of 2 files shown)


carllin (dismissed):

carllin:
good catch. Luckily, this shouldn't be affecting the clusters right now because the `pubkey specific timeout == Epoch` is always greater than `default timeout == CRDS_GOSSIP_PULL_CRDS_TIMEOUT_MS` right?

behzadnouri:
> good catch! Luckily, this shouldn't be affecting the clusters right now because the `pubkey specific timeout == Epoch` is always greater than `default timeout == CRDS_GOSSIP_PULL_CRDS_TIMEOUT_MS` right?

seems like the only pubkey with override has timeout `u64::MAX`.
so I think you are right. this shouldn't impact current clusters.
https://github.com/solana-labs/solana/blob/06f84c65f/core/src/crds_gossip_pull.rs#L471

PR #12536, opened 2020-09-28
Pipe FeatureSet though InvokeContext

jackcmay:
#### Problem

Runtime and the loader have no access to bank's feature set

#### Summary of Changes

- Pipe the bank's feature set through InvokeContext
- Remove the cross-program invocation feature, already enabled in all networks
- Add a feature to update the compute budget and move the compute budget into MessageProcessor
- Remove cap on maximum program size, not hittable now but could be when the new compute budget configuration feature is enabled.

Fixes #

### DIFF
--- a/programs/bpf_loader/src/bpf_verifier.rs
+++ b/programs/bpf_loader/src/bpf_verifier.rs
@@ -54,12 +54,14 @@ pub enum VerifierError {
     InvalidRegister(usize),
 }
 
-fn check_prog_len(prog: &[u8]) -> Result<(), BPFError> {
+fn check_prog_len(prog: &[u8], is_program_size_cap: bool) -> Result<(), BPFError> {
     if prog.len() % ebpf::INSN_SIZE != 0 {
         return Err(VerifierError::ProgramLengthNotMultiple.into());
     }
-    if prog.len() > ebpf::PROG_MAX_SIZE {
-        return Err(VerifierError::ProgramTooLarge(prog.len() / ebpf::INSN_SIZE).into());
+    if is_program_size_cap {
+        if prog.len() > ebpf::PROG_MAX_SIZE {
+            return Err(VerifierError::ProgramTooLarge(prog.len() / ebpf::INSN_SIZE).into());
+        }
     }
 
     if prog.is_empty() {
@@ -139,8 +141,8 @@ fn check_imm_register(insn: &ebpf::Insn, insn_ptr: usize) -> Result<(), Verifier
 }
 
 #[rustfmt::skip]
-pub fn check(prog: &[u8]) -> Result<(), BPFError> {
-    check_prog_len(prog)?;
+pub fn check(prog: &[u8], is_program_size_cap: bool) -> Result<(), BPFError> {
+    check_prog_len(prog, is_program_size_cap)?;
 
     let mut insn_ptr: usize = 0;
     while insn_ptr * ebpf::INSN_SIZE < prog.len() {
--- a/programs/bpf_loader/src/bpf_verifier.rs
+++ b/programs/bpf_loader/src/bpf_verifier.rs
@@ -58,10 +58,8 @@ fn check_prog_len(prog: &[u8], is_program_size_cap: bool) -> Result<(), BPFError
     if prog.len() % ebpf::INSN_SIZE != 0 {
         return Err(VerifierError::ProgramLengthNotMultiple.into());
     }
-    if is_program_size_cap {
-        if prog.len() > ebpf::PROG_MAX_SIZE {
-            return Err(VerifierError::ProgramTooLarge(prog.len() / ebpf::INSN_SIZE).into());
-        }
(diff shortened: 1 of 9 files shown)


runtime/src/bank.rs
```diff
@@ -3627,6 +3621,12 @@ impl Bank {
                         }
                     }
                 }
+            } else {
```
jackcmay, line 3624: What do you think about this as a way to consolidate existing epoch dependent features into FeatureSet?  Doing it this way means the feature would turn on at some epoch or we could enable it at anytime before.

runtime/src/bank.rs
```diff
@@ -3648,7 +3635,12 @@ impl Bank {
                         }
                     }
                 }
+            } else if *feature_id == feature_set::cross_program_invocation_enabled::id()
```
jackcmay, line 3638: @mvines What do you think about this as a way to consolidate existing epoch dependent features into FeatureSet? Doing it this way means the feature would turn on at some epoch or we could enable it at anytime before.

mvines: Is this just a transition problem because CPI not fully rolled out on all clusters?   Once we're fully migrated over I feel like these special cases go away. I ran into a similar case at 5d6410c1cb662a31b4c5bb04efced666d7eff29a and solved it a little differently.

The problem I see with adding that info into `FeatureSet` is that then the `Feature` account won't be setup correctly (without further Bank integration), so `solana feature status` won't reflect the actual status of that feature.

In the case of `consistent_recent_blockhashes_sysvar`, I still need to `solana feature activate` this feature on devnet/testnet even though it's already actually active.  This is just for accounting purposes to ensure the `Feature` account is configured correctly at the next epoch on devnet/testnet.

jackcmay: @mvines CPI (at least this feature) is rolled out on all the clusters already, on test/devnet its on by default and on mainnet is after epoch 63.  Maybe we don't even need this feature anymore, just always on?

I want this feature in FeatureSet so it's available to the runtime without threading a special case bool.  Looks like the way you solved `consistent_recent_blockhashes_sysvar` is very similar to this given the feature may be activated via account but also activated without an account existing.  So both solutions an account can be created to satisfy the `solana feature status`.

mvines: > Maybe we don't even need this feature anymore, just always on?

Oh yeah, if it's already enabled on all three clusters then I think the feature can be removed from the code entirely.

runtime/src/feature_set.rs
```diff
@@ -38,6 +42,7 @@ lazy_static! {
         (pico_inflation::id(), "pico-inflation"),
         (spl_token_v2_multisig_fix::id(), "spl-token multisig fix"),
         (bpf_loader2_program::id(), "bpf_loader2 program"),
+        (compute_budget_config2::id(), "compute budget configuration 2"),
```
mvines, line 45: Why "2"?

jackcmay: there was an original compute budget config and there will be more as time goes on

programs/bpf_loader/src/bpf_verifier.rs
```diff
@@ -58,10 +58,6 @@ fn check_prog_len(prog: &[u8]) -> Result<(), BPFError> {
     if prog.len() % ebpf::INSN_SIZE != 0 {
         return Err(VerifierError::ProgramLengthNotMultiple.into());
     }
-    if prog.len() > ebpf::PROG_MAX_SIZE {
```
mvines, line 61: This has a risk of diverging behaviour right?   A new BPF program could be accepted by less than 66.6% of the nodes during an upgrade.

runtime/src/feature_set.rs
mvines: How about "1ms compute budget configuration", that's what `ComputeBudget::default()` now is I think

programs/bpf_loader/src/bpf_verifier.rs
mvines: Yeah I think this needs to be feature gated :-/

jackcmay: Hmm, yeah, I was thinking about it as there would be no programs on the chain larger then this that would also be able to execute successfully with the current budget, but yeah, I guess finalize could diverge.  Divergence avoidance is hard

runtime/src/feature_set.rs
jackcmay: There might be tweaks etc.. that might also align with 1ms, not sure we want to try and name each tweak we make, numeric versioning seems to be more straight forward

mvines: I feel like we could continue to come up with more meaningful names for future budget tweaks. 

I'm thinking that the output from `solana feature status` is also for general 3rd party developer/validator consumption.
Somebody asking about the status of "1ms compute budget configuration" is going to be much more useful than "compute budget configuration 2" (especially if there's a 2, 3, 4, 5 as well).

PR #12555, opened 2020-09-29
removes invalid/outdated pending push messages early

behzadnouri:
#### Problem
In CrdsGossipPush::new_push_messages:
https://github.com/solana-labs/solana/blob/972619edb/core/src/crds_gossip_push.rs#L211-L228
we already have paid the cost of looking-up the label in crds table and
checking the hash value and wallclock only to find out that in some
cases the value is invalid or is outdated. So might as well remove the
value here rather than wait for the next call to
purge_old_pending_push_messages:
https://github.com/solana-labs/solana/blob/972619edb/core/src/crds_gossip_push.rs#L372

#### Summary of Changes
Removes invalid/outdated pending push messages early in CrdsGossipPush::new_push_messages.

### DIFF
--- a/core/src/crds_gossip_push.rs
+++ b/core/src/crds_gossip_push.rs
@@ -216,52 +216,59 @@ impl CrdsGossipPush {
     /// The list of push messages is created such that all the randomly selected peers have not
     /// pruned the source addresses.
     pub fn new_push_messages(&mut self, crds: &Crds, now: u64) -> HashMap<Pubkey, Vec<CrdsValue>> {
-        let mut total_bytes: usize = 0;
-        let mut values = vec![];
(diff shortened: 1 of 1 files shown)


sakridge (approved):
looks good, did you check the timings?

behzadnouri:
> looks good, did you check the timings?

on a small network of 10 nodes I do not see any impact.
it might be different for larger networks.

behzadnouri:
> > looks good, did you check the timings?
> 
> on a small network of 10 nodes I do not see any impact.
> it might be different for larger networks.

`new_push_requests` looks slightly better.

cluster_info_stats2 new_push_requests_12555

PR #12561, opened 2020-09-29
Include post balance information for rewards

mvines:
The rewards information surfaced in RPC `getConfirmedBlock` currently doesn't include the account balance before or after the reward was credited.   This makes it impossible to easily answer the simple question, "What's the % staking rewards my account received in previous epochs".

This draft PR adds a `post_balance` field to the `Rewards` struct that comes out of the Bank.  The plumbing into Blockstore/Bigtable/RPC is TBD.

### DIFF
--- a/core/src/rewards_recorder_service.rs
+++ b/core/src/rewards_recorder_service.rs
@@ -1,5 +1,6 @@
 use crossbeam_channel::{Receiver, RecvTimeoutError, Sender};
 use solana_ledger::blockstore::Blockstore;
+use solana_runtime::bank::RewardInfo;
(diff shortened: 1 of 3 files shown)


core/src/rewards_recorder_service.rs
```diff
@@ -49,9 +50,9 @@ impl RewardsRecorderService {
         let (slot, rewards) = rewards_receiver.recv_timeout(Duration::from_secs(1))?;
         let rpc_rewards = rewards
             .into_iter()
-            .map(|(pubkey, lamports)| Reward {
+            .map(|(pubkey, reward_info)| Reward {
```
mvines, line 53: `reward_info.post_balance` gets dropped here...

PR #12569, opened 2020-09-29
add sha256 syscall

jackcmay:
#### Problem

Expensive to hash in BPF

#### Summary of Changes

Provide a syscall for hashing

Fixes #


jackcmay:
@garious No SHA512 support, but on the SHA256 side of things, this what you were expecting?

PR #12572, opened 2020-09-29
Tighten docs publishing flow

### LINKED ISSUE #11387: set-solana-release-tag.sh does not stick to beta branch
#### Problem
For non-release docs publishing to [beta|edge].docs.solana.com, we are substituting in the latest-in-time release version.  If we make a release on the stable branch last, this value gets subbed in, when it should only be the latest release on beta branch, that is, the release with the greatest numerical version number.

https://github.com/solana-labs/solana/blob/master/docs/set-solana-release-tag.sh#L10

#### Proposed Solution
 - Set release tag based on numerical precedence, rather than newest
 - Make sure we don't accidentally set a tag for a pre-release build.

### LINKED ISSUE #11505: Docs build process publishes on any branch tag
#### Problem
We are maintaining and publishing releases to multiple non-master branches.  Travis CI publishes to docs.solana.com on ANY release tag, but it should happen only on a release tag on the beta branch.  This causes the published docs to bounce between two different branches of source code.

#### Proposed Solution
Tighten up branch checking logic in the travis publishing scripts/configs.

danpaul000:
#### Problem
 - We are substituting in a release tag into the docs content based on the latest-in-time release on any branch.  We should only sub in tags from the latest tag on the beta branch.
- Docs publishing to docs.solana.com occurs when a tag is made on any branch, but this should only update docs.solana.com when a tag is made on beta branch.

#### Summary of Changes
 - Use `ci/channel-info.sh` to get latest beta tag, rather than querying github API.
 - Add a check to `.travis.yml` not to proceed with docs publishing if we are not on `beta` or `edge` channels.

Fixes #11505
Fixes #11387

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -124,6 +124,7 @@ jobs:
           - ~/.npm
 
       before_install:
+        - .travis/channel_restriction.sh edge beta || travis_terminate 0
         - .travis/affects.sh docs/ .travis || travis_terminate 0
         - cd docs/
         - source .travis/before_install.sh
--- a/.travis/channel_restriction.sh
+++ b/.travis/channel_restriction.sh
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+#
+# Only proceed if we are on one of the channels passed in when calling this file
+#
+
+set -ex
+
+eval "$(ci/channel-info.sh)"
+
+for acceptable_channel in "$@"; do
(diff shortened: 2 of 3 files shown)


danpaul000:
Commented out the "skip if PR" logic in `.travis.yml` to test logic.  Will revert before merge.

mvines (approved):
lgtm

t-nelson (approved):
:tada:

PR #12581, opened 2020-09-30
@solana/wasm

### LINKED ISSUE #3693: Research Rust <-> WASM for solana-web3.js

bartosz-lipinski:
#### Problem

Splitting PR: https://github.com/solana-labs/solana/pull/12383. This includes only @solana/wasm library that is addressing: "Research Rust <-> WASM for solana-web3.js" #3693

Once this PR is approved we can consider using @solana/wasm in @solana/web3.js. 

#### Summary of Changes

@solana/wasm exposes: 
* KeyGeneration: random, fromSeed, fromSecretKey
* Verify, Sign and isOnCurve for ED25519
* Sha256

Additionally, the library provides webpack plugins that simplify integration with wasm for React projects. 

Fixes #3693


jstarry:
I'm thinking ahead about how this all gets integrated and I'm starting to imagine wasm simply fitting in to the perf critical parts of the sdk as well as areas that we don't want to write twice (serialization, transaction message encoding, etc). I think most of these areas would be used via async methods.

With that being said, I think we can get by without a `waitForReady` method and hide all wasm loading behind async methods. I think this would be a lot more better for dev experience. What do you think?

jstarry:
> Additionally, the library provides webpack plugins that simplify integration with wasm for React projects.

I would prefer if we didn't provide extra plugins if we don't need to. Can you provide some more details on the benefits of providing these rather than instructing devs how to use the community (webpack, rollup, babel) wasm plugins directly?

wasm/src/lib.rs
```diff
@@ -0,0 +1,148 @@
...
+	let secret = SecretKey::from_bytes(&seed).unwrap();
+	let public: PublicKey = (&secret).into();
+	let keypair = Keypair{secret, public};
+
+	// use all 64bit for secret key to maintain compatibility with tweetnacl
+	let secret_key = keypair.to_bytes().to_vec();
+	let public_key = keypair.public.as_bytes().to_vec();
+
+	let result = Pair{ public_key, secret_key};
+	Ok(JsValue::from_serde(&result).unwrap())
```
jstarry, line 56: You can return `Pair` directly, the serialization will be done automatically and then you will get nice typescript typedefs

wasm/src/lib.rs
```diff
@@ -0,0 +1,148 @@
...
+	return Ok(JsValue::from_serde(&result).unwrap());
+}
+
+/// Sign a message using ED25519 from ed25519_dalek
+/// * pubkey: UIntArray with 32 elements
+/// * private: UIntArray with 32 or 64 elements
+/// * message: Arbitrary length UIntArray
+///
+/// * returned vector is the signature consisting of 64 bytes.
+#[wasm_bindgen(js_name = signED25519)]
```
jstarry, line 97: nit: Ed25519 is not usually written in all caps. Also, since we only support `ed25519` signatures, we don't need to specify the scheme in the method name, just `sign` is sufficient

wasm/src/lib.rs
```diff
@@ -0,0 +1,148 @@
...
+/// * returned vector is the signature consisting of 64 bytes.
+#[wasm_bindgen(js_name = isOnCurveED25519)]
+pub fn ed25519_is_on_curve(pubkey: &[u8]) -> bool {
+	return curve25519_dalek::edwards::CompressedEdwardsY::from_slice(&pubkey)
+	.decompress()
+	.is_some();
+}
+
+#[wasm_bindgen(js_name = sha256)]
+pub fn hash_sha256(data: &[u8]) -> String {
```
jstarry, line 144: Is this intended for derived addresses or did you have something else in mind? I think we would be better off not exporting this since it will likely only be used internally for derived addresses

wasm/src/ts/crypto/ed25519.ts
```diff
@@ -0,0 +1,25 @@
+import {getWASM, ensureReady} from './../wrapper';
```
jstarry, line 1: I think we can get by without writing TS wrapper code by hand. `wasm-pack` generates typedefs for interacting with the wasm module directly

wasm/tests/crypto.test.ts
```diff
@@ -0,0 +1,138 @@
...
+  return finish - start;
+};
+
+describe('Benchmark in Node.js WASM implementation of ED25519 vs tweetnacl', () => {
+  beforeAll(async () => {
+    await waitReady();
+  });
+
+  // on avg. WASM is 20-25x faster than tweetnacl but I am testing conservative assumption
+  test('WASM sign method should be faster than nacl by at least 15x', () => {
```
jstarry, line 28: This is great for proving out why we should add wasm support, but I don't think we need to include it in this PR

wasm/tests/crypto.test.ts
```diff
@@ -0,0 +1,138 @@
...
+    }, ITERATIONS);
+    const mod = benchmark(() => {
+      nacl.sign.detached(data, keypair.secretKey);
+    }, ITERATIONS);
+
+    expect(mod / wasm).toBeGreaterThanOrEqual(15);
+  });
+});
+
+describe('Key generation', () => {
```
jstarry, line 46: If we drop the TS wrapper code, we can test the Rust code directly and setup browser integration tests with wasm-bindgen rather than testing here

wasm/scripts/rollup-wasm-plugin.js
```diff
@@ -0,0 +1,434 @@
+const $fs = require("fs");
```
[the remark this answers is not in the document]
bartosz-lipinski: Sure. The main purpose here is to customize how wasm is loaded in different environments. wasm-pack generated code will work in either node.js or browser and by default will require creating multiple packages. This plugin takes a wasm-pack generated code and creates binding that can be shared between node and browser see line 225 loadWasm method. Additionally, it creates a base64 version of the file that can be imported in plain HTML.

bartosz-lipinski (commented):
> Can you provide some more details on the benefits of providing these rather than instructing devs how to use the community (webpack, rollup, babel) wasm plugins directly?

Currently, all the wasm plugins focus on loading wasm in the user space (not provided by npm package if they load wasm from the package usually it happens via inlining which is not ideal from bundle size. I found that only through combination of babel plugin to transform fetch and webpack plugin to copy required files I was able to loaded properly within client project. 

If we don't provide the plugins directly it will require users to create those plugins manually which is not ideal in my opinion.

jstarry:
I'm starting to think we could make web3.wasm be totally separate from web3.js but still complimentary. That way, performance critical apps could leverage the wasm sdk for things like signing, sigverify, and key generation if needed but those apps could then create web3.js `Account`'s, `Transaction`'s, or `PublicKey`'s from the `Buffer`'s created with web3.wasm. Developers for these apps would also be willing to do more work to get wasm setup properly, whereas the average developer might get turned off by the extra setup and wouldn't really benefit from it anyway.

For example:

Apps continue creating transactions with web3.js but when it comes to signing they have two choices:
1. Sign directly with web3.js using the current `sign(..)` method (slow but convenient)
2. Serialize the transaction and then sign with web3.wasm and then add the signature via the web3.js `addSignature` method (fast but requires more setup to get wasm working)

---

This way, web3.wasm can continue growing organically without worrying about making breaking changes to the web3.js SDK or causing headaches for existing devs. Eventually, when wasm is more widely supported, we can direct people over the web3.wasm SDK and deprecate the web3.js repo

wasm/src/lib.rs
bartosz-lipinski: I was pllaning to use this instead of crypto-hash to reduce external dependecies.

wasm/src/lib.rs
```diff
@@ -0,0 +1,148 @@
...
+
+/// Generate key pair for use with ED25519 from ed25519_dalek
+///
+/// * returned struct has two fields public and secret.
+#[wasm_bindgen(js_name = generateKeyPair)]
+pub fn generate_keypair() -> Result<JsValue, JsValue> {
+	// using get random directly instead of OsRng since it's easier to integrate with WASM
+	let mut seed = [0u8; 32];
+	getrandom::getrandom(&mut seed).unwrap();
+	let secret = SecretKey::from_bytes(&seed).unwrap();
```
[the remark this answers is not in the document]
bartosz-lipinski: unwrap will just bubble as exception to javascript with callstack

wasm/src/lib.rs
```diff
@@ -0,0 +1,148 @@
...
+	let public: PublicKey = (&secret).into();
```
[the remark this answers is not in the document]
bartosz-lipinski: done

wasm/src/lib.rs
```diff
@@ -0,0 +1,148 @@
+extern crate ed25519_dalek;
+extern crate hex;
```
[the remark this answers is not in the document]
bartosz-lipinski: done

wasm/src/lib.rs
```diff
@@ -0,0 +1,148 @@
...
+
+mod utils;
+
+use wasm_bindgen::prelude::*;
+
+use std::convert::TryFrom;
+use ed25519_dalek::{Keypair, PublicKey, SecretKey, Signature, Signer as _, Verifier as _, SECRET_KEY_LENGTH, PUBLIC_KEY_LENGTH};
```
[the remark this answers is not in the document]
bartosz-lipinski: done

wasm/tests/crypto.test.ts
bartosz-lipinski: We can add browser integration tests but this test should stay in my opinion as they verify node.js side of the backage.

wasm/tests/crypto.test.ts
bartosz-lipinski: done

wasm/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+name = "solana-wasm"
+version = "0.1.0"
+description = "Solana WASM bindings"
+authors = ["Solana Maintainers <maintainers@solana.com>"]
+repository = "https://github.com/solana-labs/solana"
+homepage = "https://solana.com/"
+license = "Apache-2.0"
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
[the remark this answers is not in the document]
bartosz-lipinski: done

wasm/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+
+[lib]
+name = "solana_web3_wasm"
```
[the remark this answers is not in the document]
bartosz-lipinski: dpne

wasm/src/lib.rs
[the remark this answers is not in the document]
jstarry: Ah yeah, you're totally right. A panic in wasm will bubble up through the wasm-bindgen bindings. The panic does look quite scary though from the wasm-bindgen error logging so I think it would be great if you returned errors properly. `wasm-bindgen` methods return `Result` so that Result::Err variants can be turned into thrown exceptions.

```
  console.error
    panicked at 'called `Result::unwrap()` on an `Err` value: ()', src/lib.rs:49:25
    
    Stack:
    
    Error: 
        at /Users/jstarry/Workspace/solana/solana/wasm/dist/solana-wasm.cjs.js:515:15
        at /Users/jstarry/Workspace/solana/solana/wasm/dist/solana-wasm.cjs.js:370:22
        at console_error_panic_hook::Error::new::h8677d9788b6f7488 (wasm-function[1080]:0x768bd)
        at console_error_panic_hook::hook_impl::hd059b7ee5c601848 (wasm-function[202]:0x5847e)
        at console_error_panic_hook::hook::h2376c682cbcd7b97 (wasm-function[1210]:0x78614)
        at core::ops::function::Fn::call::hfc902b7dbd494ca7 (wasm-function[981]:0x75058)
        at std::panicking::rust_panic_with_hook::h3e6619a8809a443a (wasm-function[346]:0x629cb)
        at rust_begin_unwind (wasm-function[1099]:0x76d6e)
        at core::panicking::panic_fmt::ha8209066b105fcdd (wasm-function[1311]:0x79a37)
        at core::option::expect_none_failed::ha72d60de35a76889 (wasm-function[569]:0x6ba30)

      506 | const __wbg_error_4bb6c2a97407129a = logError(function(arg0, arg1) {
      507 |     try {
    > 508 |         console.error(getStringFromWasm0(arg0, arg1));
          |                 ^
      509 |     } finally {
      510 |         wasm.__wbindgen_free(arg0, arg1);
      511 |     }

      at dist/solana-wasm.cjs.js:508:17
      at dist/solana-wasm.cjs.js:370:22
      at console_error_panic_hook::error::hcd67c963013e0f67 (wasm-function[412]:0x65e91)
      at console_error_panic_hook::hook_impl::hd059b7ee5c601848 (wasm-function[202]:0x58567)
      at console_error_panic_hook::hook::h2376c682cbcd7b97 (wasm-function[1210]:0x78614)
      at core::ops::function::Fn::call::hfc902b7dbd494ca7 (wasm-function[981]:0x75058)
      at std::panicking::rust_panic_with_hook::h3e6619a8809a443a (wasm-function[346]:0x629cb)
```

Issue #12494, opened 2020-09-25
v1.3: Bigtable inner transaction plumbing is incompatible with v1.2


mvines:
I'm pretty sure that the new field added to `StoredConfirmedBlockTransactionStatusMeta` here:
https://github.com/solana-labs/solana/blob/179dd6ee5993e27844df750f8d9e9836f1f00fdd/storage-bigtable/src/lib.rs#L165
will break the deserialization of existing `StoredConfirmedBlock` entries in BigTable:
https://github.com/solana-labs/solana/blob/179dd6ee5993e27844df750f8d9e9836f1f00fdd/storage-bigtable/src/lib.rs#L76-L89

It's impractical to reprocess the existing BigTable transaction contents, so we'll need two versions of `StoredConfirmedBlockTransactionStatusMeta`/`StoredConfirmedBlock` in the v1.3 branch.     A legacy version for existing data, and the current version.

mvines:
To prevent this from accidentally occurring again, the `#[frozen_abi]` attribute should be added to all the structs in `storage-bigtable/` that are written to BigTable as well

jstarry:
@mvines the whole point of option wrapping `inner_instructions` was to prevent this issue. The only issue I see is if the warehouse nodes are updated before the api nodes which would cause v1.2 api nodes to try to serialize v1.3 serialized transaction metas which would break.

CriesofCarrots:
I think `inner_instructions` might need a `#[serde(default)]` tag for v1.3 code to be able to deserialize existing `StoredConfirmedBlock` entries in BigTable.

jstarry:
Got it, fiddling with this now

mvines:
Here's some easy STR:
```
$ solana-ledger-tool -l . bigtable block 37147442
```

Works on v1.2 but master dies with:
```
$ solana-ledger-tool -l . bigtable block 37147442
[2020-09-26T01:44:29.402773000Z INFO  solana_storage_bigtable::access_token] Requesting token for BigTableData scope
[2020-09-26T01:44:29.594210000Z INFO  solana_storage_bigtable::access_token] Token expires in 3599 seconds
[2020-09-26T01:44:30.652954000Z WARN  solana_storage_bigtable::bigtable] Failed to deserialize blocks/000000000236d332: invalid length 3, expected a multi-byte length
BigTableError(ObjectCorrupt("blocks/000000000236d332"))
```

jstarry:
[...]

PR #12587, opened 2020-09-30
Fix TransactionStatusMeta breakage in blockstore

CriesofCarrots:
#### Problem
As of https://github.com/solana-labs/solana/pull/12311 (and the v1.3 backport), nodes running on current code cannot deserialize legacy TransactionStatusMeta data.

#### Summary of Changes
- Add a deserialization helper function that sets remaining fields to default on end of file error. This enables extending structs by adding new optional fields to the end.

cc #12494

### DIFF
--- a/sdk/src/deserialize_utils.rs
+++ b/sdk/src/deserialize_utils.rs
@@ -1,0 +1,91 @@
+use serde::{Deserialize, Deserializer};
+
+pub fn default_on_eof<'de, T, D>(d: D) -> Result<T, D::Error>
+where
+    D: Deserializer<'de>,
+    T: Deserialize<'de> + Default,
+{
+    let result = T::deserialize(d);
+    match result {
+        Err(err) if err.to_string() == "io error: unexpected end of file" => Ok(T::default()),
+        result => result,
+    }
+}
+
+#[cfg(test)]
+pub mod tests {
+    use super::*;
+    use bincode::deserialize;
+
+    #[test]
+    fn test_default_on_eof() {
+        #[derive(Serialize, Deserialize, Debug, PartialEq)]
+        struct Foo {
+            bar: u16,
+            #[serde(deserialize_with = "default_on_eof")]
+            baz: Option<u16>,
+            #[serde(deserialize_with = "default_on_eof")]
+            quz: String,
+        }
+
+        let data = vec![1, 0];
+        assert_eq!(
+            Foo {
+                bar: 1,
+                baz: None,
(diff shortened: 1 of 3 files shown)


sdk/src/deserialize_utils.rs
```diff
@@ -0,0 +1,91 @@
+use serde::{Deserialize, Deserializer};
+
+pub fn default_on_eof<'de, T, D>(d: D) -> Result<T, D::Error>
+where
+    D: Deserializer<'de>,
+    T: Deserialize<'de> + Default,
+{
+    let result = T::deserialize(d);
+    match result {
+        Err(err) if err.to_string() == "io error: unexpected end of file" => Ok(T::default()),
```
CriesofCarrots, line 10: @mvines I spent a bit of time trying to check the error type here, but didn't make a lot of headway, due to the boxed generic error that serde deserialize returns (to accommodate different formats, etc). Got any ideas?

jstarry: I also spent time there and didn't make any progress, I think this is the best way

mvines: ok, wfm

sdk/src/deserialize_utils.rs
```diff
@@ -0,0 +1,91 @@
...
+    use bincode::deserialize;
+
+    #[test]
+    fn test_default_on_eof() {
+        #[derive(Serialize, Deserialize, Debug, PartialEq)]
+        struct Foo {
+            bar: u16,
+            #[serde(deserialize_with = "default_on_eof")]
+            baz: Option<u16>,
+            #[serde(deserialize_with = "default_on_eof")]
```
mvines, line 27: I'm curious what would happen if this second `default_on_eof` was omitted?   Does `deserialize` blow up?

CriesofCarrots: If the second `default_on_eof` is omitted: if `bar` and `baz` are both present, `deserialize` works fine, but if `deserialize` on a partial struct blows up.

CriesofCarrots: I can add a should-panic unit test to emphasize usage, if that would be helpful...?

mvines: Yeah a test demoing the blow-up that we can point people at when their code blows up seems helpful.  Sucks that we can't have serde attributes depend on each other

mvines (dismissed):

PR #12598, opened 2020-09-30
Include post balance information for rewards

CriesofCarrots:
The rewards information surfaced in RPC getConfirmedBlock currently doesn't include the account balance before or after the reward was credited. This makes it impossible to easily answer the simple question, "What's the % staking rewards my account received in previous epochs".

This PR adds a post_balance field to the Rewards struct that comes out of the Bank, and adds handling for Blockstore and Bigtable such that legacy Rewards without that field will still be deserialized properly. Post-balance is set to 0 if not available.

Closes #12561


mvines (approved):
protobuf and `default_on_eof` made this super easy!

CriesofCarrots:
Yep! Thanks @jstarry for laying the groundwork!

PR #12599, opened 2020-09-30
Add GetConfirmedBlocksWithLimit RPC method

mvines:
`GetConfirmedBlocks` requires that I provide an `end_slot` or else I get fed up to 500,000 blocks.   This is wasteful when I just want to poke around the ledger and return a block or two after a given slot (eg, the first block in an epoch).  

To get around this add `GetConfirmedBlocksWithLimit`, which requires a starting slot and a limit.   

Adding a new method seemed conceptually much simpler than trying to shoehorn an optional limit argument at the end of `GetConfirmedBlocks`.    Also `GetConfirmedBlocks` still has it's uses (ie, when I want to fetch all confirmed blocks in a given epoch) so it deserves to continue to exist as a non-deprecated method alongside `GetConfirmedBlocksWithLimit`


mvines:
@CriesofCarrots - wdyt?

CriesofCarrots (approved):
Wfm, I will definitely make use of it! The request_processor method internals could be deduped, but they are simple enough that it feels like more trouble than it's worth.

mvines:
> The request_processor method internals could be deduped, but they are simple enough that it feels like more trouble than it's worth.

My thinking as well, not worth the trouble.

PR #12600, opened 2020-09-30
Add nonced-tx check to RpcClient

CriesofCarrots:
#### Problem
If a transaction status isn't found on the first query, RpcClient send-and-confirm methods use the recent blockhash and continue polling until the transaction has expired. This doesn't work for nonced transactions, since the recent blockhash has deliberately already expired. The result is that the RpcClient method may return early that the transaction failed when in fact it has yet to land.

#### Summary of Changes
If the transaction is a nonced transaction, grab a recent blockhash from the chain and use that for initial timeout.

### DIFF
--- a/client/src/rpc_client.rs
+++ b/client/src/rpc_client.rs
@@ -32,7 +32,7 @@ use solana_sdk::{
     hash::Hash,
     pubkey::Pubkey,
     signature::Signature,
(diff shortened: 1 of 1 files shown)


t-nelson (approved):
Much cleaner!

Could de-replicode via a helper, but up to you

PR #12605, opened 2020-10-01
implements drain_filter for crds table

behzadnouri:
#### Problem
Crds::find_old_labels clones and re-allocates all the outdated crds
labels on the heap:
https://github.com/solana-labs/solana/blob/0d5258b6d/core/src/crds.rs#L181-L203
CrdsGossipPull::purge_active then looks up the values again from the
crds table and removes them:
https://github.com/solana-labs/solana/blob/0d5258b6d/core/src/crds_gossip_pull.rs#L485-L507
The heap allocation and the 2nd lookup are redundant and unnecessary.

Similarly CrdsValue::record_labels allocates ~300 CrdsValueLabel on the
heap by buffering into a vector, whereas an iterator will suffice at the
call-site.
https://github.com/solana-labs/solana/blob/0d5258b6d/core/src/crds_value.rs#L426-L439

#### Summary of Changes
The commit implements a drain_filter method, similar to
Vec::drain_filter:
https://doc.rust-lang.org/std/vec/struct.Vec.html#method.drain_filter
to remove values more efficiently without buffering into a vector and
redundant second lookup.

Also, updated Crds::find_old_labels to return an iterator.

### DIFF
--- a/core/src/crds_gossip_pull.rs
+++ b/core/src/crds_gossip_pull.rs
@@ -541,20 +541,12 @@ impl CrdsGossipPull {
         now: u64,
         timeouts: &HashMap<Pubkey, u64>,
     ) -> usize {
-        let old = crds.find_old_labels(now, timeouts);
-        let mut purged: VecDeque<_> = old
-            .iter()
-            .filter_map(|label| {
-                let rv = crds
-                    .lookup_versioned(label)
-                    .map(|val| (val.value_hash, val.local_timestamp));
-                crds.remove(label);
-                rv
-            })
-            .collect();
-        let ret = purged.len();
-        self.purged_values.append(&mut purged);
-        ret
+        let num_purged_values = self.purged_values.len();
+        self.purged_values.extend(
+            crds.drain_old_values(now, timeouts)
+                .map(|(_, v)| (v.value_hash, v.local_timestamp)),
+        );
+        self.purged_values.len() - num_purged_values
     }
     /// Purge values from the `self.purged_values` queue that are older then purge_timeout
     pub fn purge_purged(&mut self, min_ts: u64) {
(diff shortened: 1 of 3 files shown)


core/src/crds_gossip_pull.rs
```diff
@@ -542,20 +542,12 @@ impl CrdsGossipPull {
         now: u64,
         timeouts: &HashMap<Pubkey, u64>,
     ) -> usize {
-        let old = crds.find_old_labels(now, timeouts);
```
sakridge, line 545: I wonder what the time separation is here, because we could more easily split the work into a read-lock/write-lock split here. `purged_values` could be something like `Vec<Vec<_>>` and just appending a `Vec` onto the end would be super cheap within the write lock part. You also have `crds.remove` which doesn't seem easy to refactor outside the write lock, so if that is the main time then maybe it doesn't make sense. But you don't have to do a full table scan within the write lock.

behzadnouri: yeah this was done given that we already have a mutable reference here, so might implement the same logic without the extra heap allocation, cloning and look-ups.
From profiling run-time of a small gce network, it seems like vast majority of the time is spent in scanning the table. if that is also the case on mainnet, then yes you are right, better not to do a full table scan holding the write-lock.
From the metrics seems like the purge-count tends to be very small in general. So again the table scan part should be dominant. Though perhaps maybe one node is having large `purge-count` compared to rest of the nodes (which seems odd), and that would involve different trade-offs.

Alternatives are implement a finer grained locking in gossip (and perhaps one dedicated read/write lock for crds table), or utilize this thread_pool:
https://github.com/solana-labs/solana/blob/87de82ac9/core/src/cluster_info.rs#L1562
in the scan part of handle_purge further down:
https://github.com/solana-labs/solana/blob/87de82ac9/core/src/cluster_info.rs#L1606
cluster_info-purge-count-tail

behzadnouri:
closing this in favor of https://github.com/solana-labs/solana/pull/13073 which scans the crds table in parallel.

behzadnouri:
closing this in favor of https://github.com/solana-labs/solana/pull/13073 which scans the crds table in parallel.

PR #8436, opened 2020-02-25
Do periodic inbound cleaning for rooted slots

ryoqun:
#### Problem

After much investigation, I've realized just removing storage entries with count being 0 nor tweaking the [purge logic](https://github.com/solana-labs/solana/blob/73a278dc647d12cf43a6c14dccfd1edb42cc0ee5/runtime/src/accounts_db.rs#L899) isn't enough.

The root cause of snapshot bloat is that we leak StorageEntry/AppendVec too easily, first of all. Even just running `solana ping` causes that situation.

In this regard, much like the zero-lamport accounts RAM exhaustion DOS attack, this behavior can be easily used to mount a DOS with rather small amount of SOL.

#### Summary of Changes

We were too lazy to GC older storage entries. Especially, we do literally nothing for any account states while not touched.

So, introduce a bit eager compaction for non-zero lamports accounts at the snapshot generation intervals.

#### Background

This is one of two PRs towards #8168. This is an inbound compaction. The other (#8337) is an outbound (when loading snapshot) compaction. To really fix enlarged snapshot on TdS, we need to land both.

Part of #8168

PR #7157, opened 2019-11-27
More conservative purge_zero_lamport_accounts purge logic

sakridge:
#### Problem

Purging zero lamport accounts where the parent account's appendvec is not removed is not compatible with snapshots. Since the index is not serialized in the snapshot, the old view of the account could be revived.

#### Summary of Changes

Don't purge zero lamport accounts unless all AppendVecs containing parent accounts can be also removed.

Fixes #7117

PR #12606, opened 2020-10-01
Fix zero-lamport accounts preventing slot cleanup

carllin:
#### Problem
Some slot `S`  containing an update for an account with key `K` may not be cleaned up if at the time of `clean_slots`, the latest update in `K` is in some rooted slot `S' > S` and `K` has zero lamports. This is because in `clean_slots()`

`if account_info.lamports == 0 {`

and 

`else if accounts_index.uncleaned_roots.contains(slot)`

are mutually exclusive, and `uncleaned_roots` is purged at the end of the clean_slots() loop, so those slots will not be re-examined if the pubkey is not updated again in some future root that is added to `uncleaned root`

#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -138,6 +138,13 @@ impl AccountStorage {
             })
             .map(|account| (account, slot))
     }
+
+    fn slot_store_count(&self, slot: Slot, store_id: AppendVecId) -> Option<usize> {
+        self.0
+            .get(&slot)
+            .and_then(|slot_storages| slot_storages.get(&store_id))
+            .map(|store| store.count_and_status.read().unwrap().0)
+    }
 }
 
 #[derive(Debug, Eq, PartialEq, Copy, Clone, Deserialize, Serialize, AbiExample, AbiEnumVisitor)]
@@ -541,13 +548,15 @@ impl AccountsDB {
         )
     }
 
-    // Reclaim older states of rooted non-zero lamport accounts as a general
-    // AccountsDB bloat mitigation and preprocess for better zero-lamport purging.
+    // Reclaim older states of rooted accounts for AccountsDB bloat mitigation
     fn clean_old_rooted_accounts(
         &self,
         purges_in_root: Vec<Pubkey>,
         max_clean_root: Option<Slot>,
     ) -> (PurgedAccountSlots, ReclaimedAccounts) {
+        if purges_in_root.is_empty() {
+            return (HashMap::new(), HashMap::new());
+        }
         // This number isn't carefully chosen; just guessed randomly such that
         // the hot loop will be the order of ~Xms.
         const INDEX_CLEAN_BULK_COUNT: usize = 4096;
@@ -736,13 +745,8 @@ impl AccountsDB {
         accounts_scan.stop();
 
         let mut clean_old_rooted = Measure::start("clean_old_roots");
-        let (purged_account_slots, removed_accounts) = {
-            if !purges_in_root.is_empty() {
-                self.clean_old_rooted_accounts(purges_in_root, max_clean_root)
-            } else {
-                (HashMap::new(), HashMap::new())
-            }
-        };
+        let (purged_account_slots, removed_accounts) =
+            self.clean_old_rooted_accounts(purges_in_root, max_clean_root);
         self.do_reset_uncleaned_roots(&mut candidates, max_clean_root);
         clean_old_rooted.stop();
 
@@ -784,13 +788,13 @@ impl AccountsDB {
                 } else {
                     let mut key_set = HashSet::new();
                     key_set.insert(*key);
-                    let count: usize = {
-                        let storage = self.storage.read().unwrap();
-                        let slot_storage = storage.0.get(&slot).unwrap();
-                        let store = slot_storage.get(&account_info.store_id).unwrap();
-                        let count = store.count_and_status.read().unwrap().0;
-                        count - 1
-                    };
+                    let count = self
+                        .storage
+                        .read()
+                        .unwrap()
+                        .slot_store_count(*slot, account_info.store_id)
+                        .unwrap()
+                        - 1;
                     debug!(
                         "store_counts, inserting slot: {}, store id: {}, count: {}",
                         slot, account_info.store_id, count
@@ -899,17 +903,16 @@ impl AccountsDB {
                 assert!(dead_slots.contains(&expected_single_dead_slot));
             }
         }
-        let purged_account_slots = if !dead_slots.is_empty() {
-            self.process_dead_slots(&dead_slots)
-        } else {
-            HashMap::new()
-        };
+        let purged_account_slots = self.process_dead_slots(&dead_slots);
         (purged_account_slots, removed_accounts)
     }
 
     // Must be kept private!, does sensitive cleanup that should only be called from
     // supported pipelines in AccountsDb
     fn process_dead_slots(&self, dead_slots: &HashSet<Slot>) -> PurgedAccountSlots {
+        if dead_slots.is_empty() {
+            return HashMap::new();
+        }
         let mut clean_dead_slots = Measure::start("reclaims::purge_slots");
         let purged_account_slots = self.clean_dead_slots(&dead_slots);
         clean_dead_slots.stop();
@@ -3444,7 +3447,7 @@ pub mod tests {
         let account = Account::new(0, 0, &Account::default().owner);
         accounts.store(latest_slot, &[(&pubkeys[30], &account)]);
 
-        // Create 10 new accounts in slot 1, should now have 11 = 10 = 21
+        // Create 10 new accounts in slot 1, should now have 11 + 10 = 21
         // accounts
         create_account(&accounts, &mut pubkeys1, latest_slot, 10, 0, 0);
 
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -100,8 +100,9 @@ pub type SnapshotStorages = Vec<SnapshotStorage>;
 // Each slot has a set of storage entries.
 pub(crate) type SlotStores = HashMap<usize, Arc<AccountStorageEntry>>;
 
-type PurgedAccountSlots = HashMap<Pubkey, HashSet<Slot>>;
-type ReclaimedAccounts = HashMap<AppendVecId, HashSet<usize>>;
+type AccountSlots = HashMap<Pubkey, HashSet<Slot>>;
+type AppendVecOffsets = HashMap<AppendVecId, HashSet<usize>>;
+type ReclaimResult = (AccountSlots, AppendVecOffsets);
 
 trait Versioned {
     fn version(&self) -> u64;
@@ -553,7 +554,7 @@ impl AccountsDB {
         &self,
         purges_in_root: Vec<Pubkey>,
         max_clean_root: Option<Slot>,
-    ) -> (PurgedAccountSlots, ReclaimedAccounts) {
+    ) -> ReclaimResult {
         if purges_in_root.is_empty() {
             return (HashMap::new(), HashMap::new());
         }
@@ -578,11 +579,12 @@ impl AccountsDB {
         inc_new_counter_info!("clean-old-root-par-clean-ms", clean_rooted.as_ms() as usize);
 
         let mut measure = Measure::start("clean_old_root_reclaims");
-        let (purged_account_slots, removed_accounts) = self.handle_reclaims(&reclaims, None, false);
+        let mut reclaim_result = (HashMap::new(), HashMap::new());
+        self.handle_reclaims(&reclaims, None, false, Some(&mut reclaim_result));
         measure.stop();
         debug!("{} {}", clean_rooted, measure);
         inc_new_counter_info!("clean-old-root-reclaim-ms", measure.as_ms() as usize);
-        (purged_account_slots, removed_accounts)
+        reclaim_result
     }
 
     fn do_reset_uncleaned_roots(
@@ -839,7 +841,7 @@ impl AccountsDB {
 
         self.handle_dead_keys(dead_keys);
 
-        self.handle_reclaims(&reclaims, None, false);
+        self.handle_reclaims(&reclaims, None, false, None);
 
         reclaims_time.stop();
         datapoint_info!(
@@ -888,13 +890,19 @@ impl AccountsDB {
         reclaims: SlotSlice<AccountInfo>,
         expected_single_dead_slot: Option<Slot>,
         no_dead_slot: bool,
-    ) -> (PurgedAccountSlots, ReclaimedAccounts) {
+        reclaim_result: Option<&mut ReclaimResult>,
+    ) {
         if reclaims.is_empty() {
-            return (HashMap::new(), HashMap::new());
+            return;
         }
-
-        let (dead_slots, removed_accounts) =
-            self.remove_dead_accounts(reclaims, expected_single_dead_slot);
+        let (purged_account_slots, reclaimed_offsets) =
+            if let Some((ref mut x, ref mut y)) = reclaim_result {
+                (Some(x), Some(y))
+            } else {
+                (None, None)
+            };
+        let dead_slots =
+            self.remove_dead_accounts(reclaims, expected_single_dead_slot, reclaimed_offsets);
         if no_dead_slot {
             assert!(dead_slots.is_empty());
         } else if let Some(expected_single_dead_slot) = expected_single_dead_slot {
@@ -903,18 +911,21 @@ impl AccountsDB {
                 assert!(dead_slots.contains(&expected_single_dead_slot));
             }
         }
-        let purged_account_slots = self.process_dead_slots(&dead_slots);
-        (purged_account_slots, removed_accounts)
+        self.process_dead_slots(&dead_slots, purged_account_slots);
     }
 
     // Must be kept private!, does sensitive cleanup that should only be called from
     // supported pipelines in AccountsDb
-    fn process_dead_slots(&self, dead_slots: &HashSet<Slot>) -> PurgedAccountSlots {
+    fn process_dead_slots(
+        &self,
+        dead_slots: &HashSet<Slot>,
+        purged_account_slots: Option<&mut AccountSlots>,
+    ) {
         if dead_slots.is_empty() {
-            return HashMap::new();
+            return;
         }
         let mut clean_dead_slots = Measure::start("reclaims::purge_slots");
-        let purged_account_slots = self.clean_dead_slots(&dead_slots);
+        self.clean_dead_slots(&dead_slots, purged_account_slots);
         clean_dead_slots.stop();
 
         let mut purge_slots = Measure::start("reclaims::purge_slots");
@@ -928,7 +939,6 @@ impl AccountsDB {
             purge_slots,
             dead_slots,
         );
-        purged_account_slots
     }
 
     fn do_shrink_stale_slot(&self, slot: Slot) -> usize {
@@ -1090,7 +1100,7 @@ impl AccountsDB {
             update_index_elapsed = start.as_us();
 
             let mut start = Measure::start("update_index_elapsed");
-            self.handle_reclaims(&reclaims, Some(slot), true);
+            self.handle_reclaims(&reclaims, Some(slot), true, None);
             start.stop();
             handle_reclaims_elapsed = start.as_us();
 
@@ -1492,7 +1502,7 @@ impl AccountsDB {
 
         // 1) Remove old bank hash from self.bank_hashes
         // 2) Purge this slot's storage entries from self.storage
-        self.handle_reclaims(&reclaims, Some(remove_slot), false);
+        self.handle_reclaims(&reclaims, Some(remove_slot), false, None);
         assert!(self.storage.read().unwrap().0.get(&remove_slot).is_none());
     }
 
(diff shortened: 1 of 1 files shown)


runtime/src/accounts_db.rs
```diff
@@ -681,7 +686,8 @@ impl AccountsDB {
                         let (slot, account_info) = &list[index];
                         if account_info.lamports == 0 {
                             purges.insert(*pubkey, accounts_index.would_purge(pubkey));
-                        } else if accounts_index.uncleaned_roots.contains(slot) {
+                        }
+                        if accounts_index.uncleaned_roots.contains(slot) {
```
carllin, line 690: changed from `else if` to `if` here, deal with conflicting cases below

runtime/src/accounts_db.rs
```diff
@@ -713,26 +723,35 @@ impl AccountsDB {
...
-        let storage = self.storage.read().unwrap();
-        for (key, (account_infos, _ref_count)) in &purges {
-            for (slot, account_info) in account_infos {
-                let slot_storage = storage.0.get(&slot).unwrap();
-                let store = slot_storage.get(&account_info.store_id).unwrap();
+        for (key, (account_infos, ref_count)) in purges.iter_mut() {
+            account_infos.retain(|(slot, account_info)| {
+                if old_rooted_dead_slots.contains(slot) {
+                    *ref_count = self
+                        .accounts_index
```
carllin, line 730: removed holding storage lock above so that wouldn't dead lock with other logic that grabs `accounts_index()` lock and then the `storage lock`

runtime/src/accounts_db.rs
```diff
@@ -681,7 +686,8 @@ impl AccountsDB {
                         let (slot, account_info) = &list[index];
                         if account_info.lamports == 0 {
                             purges.insert(*pubkey, accounts_index.would_purge(pubkey));
-                        } else if accounts_index.uncleaned_roots.contains(slot) {
```
ryoqun, line 684: Thanks for fixing another AccountsDB bug.. ;)

iirc, there was some reason that I intentionally chained with `else if`. But I can't remember the exact reason. And, sorry for the lack of comment. I'll try to remember and report back here. Anyway, a lot happened since then. So, this is just ok as long as the ci is green and running validator causes no bad snapshot.

carllin: @ryoqun, no worries! Was it to avoid deleting the slot such that an AppendVec added earlier to `purges` wouldn't be deleted, i.e. so this wouldn't panic? https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_db.rs#L720. I suspect this might be the reason because if you change the `else if` to an `if` on master, and try running this test: https://github.com/solana-labs/solana/pull/12606/files#diff-2099c5256db4eb5975c8834af38f6456R3118, that unwrap() will panic

ryoqun: (for the record) this line is added #8436.

runtime/src/accounts_db.rs
```diff
@@ -713,26 +723,35 @@ impl AccountsDB {
...
+                if old_rooted_dead_slots.contains(slot) {
+                    *ref_count = self
+                        .accounts_index
+                        .read()
+                        .unwrap()
+                        .ref_count_from_storage(&key);
+                    return false;
+                }
                 if let Some(store_count) = store_counts.get_mut(&account_info.store_id) {
                     store_count.0 -= 1;
```
carllin, line 737: This and the matching `count - 1` below should only be decremented if the `clean_old_rooted_accounts` above didn't clean those storage entries already

runtime/src/accounts_db.rs
ryoqun: @carllin Hmm, it seems some of other tests are failing. Could you look at it? Those failing tests will be very cryptic to you. So don't hesitate to ask questions. At first glance, I can't tell these are actually failing by being broken or these test assertions need to just be updated.

I'm still trying to remember something. Still, the memory is blur but it was like this:

- `clean_old_rooted_accounts` should conceptually be complement to the cleaning of zero lamport accounts. So, for every given live account, we do either `clean_old_rooted_accounts` or cleaning of zero lamport accounts. Even if we don't clean old_rooted accounts for zero lamport accounts, the cleaning of zero lamport logic can purge them eventually. So there is no need to do both; it's kind of waste doing both for each account every time we run `clean_accounts()`.
- There was a problem / bug when we're too eager to clean old index entries of zero lamport accounts. So combined the above bullet point concern, I've avoided doing both and chose `if else` approach. (That's why tests failing I guess).
  - ... And I failed to comment as such because that assumption was so prevalent at the time. It was obvious thing to me. (my bad).

ryoqun: > @ryoqun, no worries! Was it to avoid deleting the slot such that an AppendVec added earlier to `purges` wouldn't be deleted, i.e. so this wouldn't panic? https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_db.rs#L720. I suspect this might be the reason because if you change the `else if` to an `if` on master, and try running this test: https://github.com/solana-labs/solana/pull/12606/files#diff-2099c5256db4eb5975c8834af38f6456R3118, that unwrap() will panic

As said in the my above comment, I don't think my thought for the original implementation decision started from these concerns. At the time, I didn't even tried to do `if` approach. Anyway, I think your observation is correct for the interrelation of two cleaning and panic of wrap with the `if` approach..

carllin: > @carllin Hmm, it seems some of other tests are failing. Could you look at it? Those failing tests will be very cryptic to you. So don't hesitate to ask questions. At first glance, I can't tell these are actually failing by being broken or these test assertions need to just be updated.
> 
> I'm still trying to remember something. Still, the memory is blur but it was like this:
> 
> * `clean_old_rooted_accounts` should conceptually be complement to the cleaning of zero lamport accounts. So, for every given live account, we do either `clean_old_rooted_accounts` or cleaning of zero lamport accounts. Even if we don't clean old_rooted accounts for zero lamport accounts, the cleaning of zero lamport logic can purge them eventually. So there is no need to do both; it's kind of waste doing both for each account every time we run `clean_accounts()`.
> * There was a problem / bug when we're too eager to clean old index entries of zero lamport accounts. So combined the above bullet point concern, I've avoided doing both and chose `if else` approach. (That's why tests failing I guess).
>   
>   * ... And I failed to comment as such because that assumption was so prevalent at the time. It was obvious thing to me. (my bad).

Yeah, working on those tests! I think the ones that are failing are a mix of assertions failing and me not handling: https://github.com/solana-labs/solana/pull/12606#discussion_r498093305, which is breaking the issue described here: https://github.com/solana-labs/solana/pull/7157

runtime/src/accounts_db.rs
carllin: Fixed and updated in latest commits.

runtime/src/accounts_db.rs
```diff
@@ -100,6 +100,9 @@ pub type SnapshotStorages = Vec<SnapshotStorage>;
 // Each slot has a set of storage entries.
 pub(crate) type SlotStores = HashMap<usize, Arc<AccountStorageEntry>>;
 
+type PurgedAccountSlots = HashMap<Pubkey, HashSet<Slot>>;
+type ReclaimedAccounts = HashMap<AppendVecId, HashSet<usize>>;
```
ryoqun, line 104: how about `AppendVecOffsets`/`reclaimed_offsets` over the status-quo?

- this type itself doesn't necessarily imply anything about being `reclaimed`. It's just some shaped container type alias.
- imply the `reclaimed` implication by variable names.
- avoid inconsistency between `reclaim`/`removed`. `reclaim` is preferred because this is related to `handle_reclaim`. `remove` is a bit too generic term; also actually we don't `remove` anything as appendvecs are always append only. So `reclaim` sounds good for me for this conceptual/logical delete here.

runtime/src/accounts_db.rs
```diff
@@ -100,6 +100,9 @@ pub type SnapshotStorages = Vec<SnapshotStorage>;
 // Each slot has a set of storage entries.
 pub(crate) type SlotStores = HashMap<usize, Arc<AccountStorageEntry>>;
 
+type PurgedAccountSlots = HashMap<Pubkey, HashSet<Slot>>;
```
ryoqun, line 103: similar to https://github.com/solana-labs/solana/pull/12606/files#r498668833, how about `AccountSlots`/`purged_account_slots`

runtime/src/accounts_db.rs
```diff
@@ -702,9 +736,13 @@ impl AccountsDB {
...
         let mut clean_old_rooted = Measure::start("clean_old_roots");
-        if !purges_in_root.is_empty() {
-            self.clean_old_rooted_accounts(purges_in_root, max_clean_root);
-        }
+        let (purged_account_slots, removed_accounts) = {
+            if !purges_in_root.is_empty() {
+                self.clean_old_rooted_accounts(purges_in_root, max_clean_root)
+            } else {
+                (HashMap::new(), HashMap::new())
+            }
```
ryoqun, line 744: nits: how about moving this `is_empty` check inside `clean_old_rooted_accounts` like https://github.com/solana-labs/solana/pull/12606/files#diff-2099c5256db4eb5975c8834af38f6456R888?

runtime/src/accounts_db.rs
```diff
@@ -540,7 +543,11 @@ impl AccountsDB {
 
     // Reclaim older states of rooted non-zero lamport accounts as a general
```
ryoqun, line 544: now, `non-zero lamports` isn't correct, thanks to your work. :)

runtime/src/accounts_db.rs
```diff
@@ -3194,41 +3429,44 @@ pub mod tests {
...
-        // Modify the first 10 of the slot 0 accounts as updates in slot 1
+        // Modify the first 10 of the accounts from slot 0 in slot 1
         modify_accounts(&accounts, &pubkeys, latest_slot, 10, 3);
+        // Overwrite account 30 from slot 0 with lamports=0 into slot 1.
+        // Slot 1 should now have 10 + 1 = 11 accounts
+        let account = Account::new(0, 0, &Account::default().owner);
+        accounts.store(latest_slot, &[(&pubkeys[30], &account)]);
 
-        // Create 10 new accounts in slot 1
+        // Create 10 new accounts in slot 1, should now have 11 = 10 = 21
```
ryoqun, line 3439: nits: `11 + 10 = 21`?

runtime/src/accounts_db.rs
```diff
@@ -816,40 +884,48 @@ impl AccountsDB {
...
+            assert!(dead_slots.len() <= 1);
+            if dead_slots.len() == 1 {
+                assert!(dead_slots.contains(&expected_single_dead_slot));
             }
         }
+        let purged_account_slots = if !dead_slots.is_empty() {
+            self.process_dead_slots(&dead_slots)
+        } else {
+            HashMap::new()
+        };
```
ryoqun, line 906: how about moving this code block into the callee like this?: https://github.com/solana-labs/solana/pull/12606/files#r498703041

runtime/src/accounts_db.rs
```diff
@@ -713,26 +751,56 @@ impl AccountsDB {
...
                     let mut key_set = HashSet::new();
                     key_set.insert(*key);
-                    store_counts.insert(
-                        account_info.store_id,
-                        (store.count_and_status.read().unwrap().0 - 1, key_set),
+                    let count: usize = {
+                        let storage = self.storage.read().unwrap();
+                        let slot_storage = storage.0.get(&slot).unwrap();
+                        let store = slot_storage.get(&account_info.store_id).unwrap();
+                        let count = store.count_and_status.read().unwrap().0;
```
ryoqun, line 791: how about defining a helper function at `AccountStorage`? I remember we have close one in tests.

ryoqun:
@carllin not strictly, related to this pr, but I found rather worried flaky test failure: https://buildkite.com/solana-labs/solana/builds/32038#b65ccd3b-3b4d-40c6-82bc-51b685ed5948/247-1671

Could you take a look at that before merging this?

```
---- bank::tests::test_clean_zero_lamport_account_different_hash stdout ----
thread 'bank::tests::test_clean_zero_lamport_account_different_hash' panicked at 'No bank hash was found for this bank, that should not be possible', runtime/src/accounts.rs:671:14
stack backtrace:
   0: rust_begin_unwind
             at /rustc/7e6d6e5f535321c2223f044caba16f97b825009c/library/std/src/panicking.rs:475
   1: core::panicking::panic_fmt
             at /rustc/7e6d6e5f535321c2223f044caba16f97b825009c/library/core/src/panicking.rs:85
   2: core::option::expect_failed
             at /rustc/7e6d6e5f535321c2223f044caba16f97b825009c/library/core/src/option.rs:1211
   3: core::option::Option<T>::expect
             at /rustc/7e6d6e5f535321c2223f044caba16f97b825009c/library/core/src/option.rs:331
   4: solana_runtime::accounts::Accounts::bank_hash_info_at
             at ./src/accounts.rs:669
   5: solana_runtime::bank::Bank::hash_internal_state
             at ./src/bank.rs:3236
   6: solana_runtime::bank::Bank::rehash
             at ./src/bank.rs:1345
   7: solana_runtime::bank::tests::test_clean_zero_lamport_account_different_hash
             at ./src/bank.rs:9365
   8: solana_runtime::bank::tests::test_clean_zero_lamport_account_different_hash::{{closure}}
             at ./src/bank.rs:9359
   9: core::ops::function::FnOnce::call_once
             at /rustc/7e6d6e5f535321c2223f044caba16f97b825009c/library/core/src/ops/function.rs:233
  10: core::ops::function::FnOnce::call_once
             at /rustc/7e6d6e5f535321c2223f044caba16f97b825009c/library/core/src/ops/function.rs:233
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
 
failures:
    bank::tests::test_clean_zero_lamport_account_different_hash
```

runtime/src/accounts_db.rs
```diff
@@ -816,40 +884,48 @@ impl AccountsDB {
...
+            if dead_slots.len() == 1 {
+                assert!(dead_slots.contains(&expected_single_dead_slot));
             }
         }
+        let purged_account_slots = if !dead_slots.is_empty() {
+            self.process_dead_slots(&dead_slots)
+        } else {
+            HashMap::new()
+        };
+        (purged_account_slots, removed_accounts)
```
ryoqun, line 907: I wonder rustc is clever enough to optimize away these collection construction spanning some call-graph altogether only when this `fn` (= `handle_reclaims`) called from `store_with_hashes`.

I'm a bit worried about the wasted computation if the optimization isn't possible because `store_with_hashes` are the one of code path, which should be the fastest.

I think it's decent guess for lively blockchain with very high tps to overwrite existing accounts over and over (otherwise, some program is leaking accounts very badly...). So, we'll end up creating somewhat large these unused bookkeeping collections.

How about passing around `Option<&mut ReclaimResult>` as an argument for the alternative approach?

ryoqun: oh, note that `type ReclaimResult = (AccountSlots, AppendVecOffsets);`.

ryoqun:
@carllin Sorry for late review as always... ;) Also, this time I'm afraid but I have to assign some homework to you.

runtime/src/accounts_db.rs
ryoqun: I see! It looks that you passed the CI fully. So happy. :)

So, the once independent cleaning mechanisms finally start to interact each other in one pass after battle-tested for half a year. That sounds cool.

ryoqun:
Anyway, I see this pr is great progress.

I'm really looking forward to the less bloated snapshots. And slightly increased write thanks to more granular `storage` lock? I think zero lamport tombstones will be reclaimed much faster with this pr.

Also, please test this for rather long duration, pointing to mainnet-beta and testnet.

carllin:
@ryoqun for sure, thanks for the review! Definitely letting this run against mainnet for a while. 

> @carllin not strictly, related to this pr, but I found rather worried flaky test failure: https://buildkite.com/solana-labs/solana/builds/32038#b65ccd3b-3b4d-40c6-82bc-51b685ed5948/247-1671
> 
> Could you take a look at that before merging this?
> 
> ```
> ---- bank::tests::test_clean_zero_lamport_account_different_hash stdout ----
> thread 'bank::tests::test_clean_zero_lamport_account_different_hash' panicked at 'No bank hash was found for this bank, that should not be possible', runtime/src/accounts.rs:671:14
> stack backtrace:
>    0: rust_begin_unwind
>              at /rustc/7e6d6e5f535321c2223f044caba16f97b825009c/library/std/src/panicking.rs:475
>    1: core::panicking::panic_fmt
>              at /rustc/7e6d6e5f535321c2223f044caba16f97b825009c/library/core/src/panicking.rs:85
>    2: core::option::expect_failed
>              at /rustc/7e6d6e5f535321c2223f044caba16f97b825009c/library/core/src/option.rs:1211
>    3: core::option::Option<T>::expect
>              at /rustc/7e6d6e5f535321c2223f044caba16f97b825009c/library/core/src/option.rs:331
>    4: solana_runtime::accounts::Accounts::bank_hash_info_at
>              at ./src/accounts.rs:669
>    5: solana_runtime::bank::Bank::hash_internal_state
>              at ./src/bank.rs:3236
>    6: solana_runtime::bank::Bank::rehash
>              at ./src/bank.rs:1345
>    7: solana_runtime::bank::tests::test_clean_zero_lamport_account_different_hash
>              at ./src/bank.rs:9365
>    8: solana_runtime::bank::tests::test_clean_zero_lamport_account_different_hash::{{closure}}
>              at ./src/bank.rs:9359
>    9: core::ops::function::FnOnce::call_once
>              at /rustc/7e6d6e5f535321c2223f044caba16f97b825009c/library/core/src/ops/function.rs:233
>   10: core::ops::function::FnOnce::call_once
>              at /rustc/7e6d6e5f535321c2223f044caba16f97b825009c/library/core/src/ops/function.rs:233
> note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
>  
> failures:
>     bank::tests::test_clean_zero_lamport_account_different_hash
> ```

Addressed this here: https://github.com/solana-labs/solana/pull/12658

runtime/src/accounts_db.rs
carllin: Fixed in latest commit! https://github.com/solana-labs/solana/pull/12606/commits/69092648864a17e4a21d5699a51b5c2b6b83c16e

runtime/src/accounts_db.rs
carllin: Fixed in latest commit! https://github.com/solana-labs/solana/pull/12606/commits/69092648864a17e4a21d5699a51b5c2b6b83c16e

runtime/src/accounts_db.rs
carllin: Fixed in latest commit! https://github.com/solana-labs/solana/pull/12606/commits/69092648864a17e4a21d5699a51b5c2b6b83c16e

runtime/src/accounts_db.rs
```diff
@@ -713,26 +757,56 @@ impl AccountsDB {
...
                     store_count.1.insert(*key);
                 } else {
                     let mut key_set = HashSet::new();
                     key_set.insert(*key);
-                    store_counts.insert(
-                        account_info.store_id,
-                        (store.count_and_status.read().unwrap().0 - 1, key_set),
+                    let count = self
+                        .storage
+                        .read()
```
ryoqun, line 795: @carllin Opps, Sorry to spot this at later stage of this pr. But, I think there is `panic!` possibility here, because of the shrinker thread, which is running in parallel.

By moving this `self.storage.read()` lock inside here from the outer loop prologue, now shrinker thread are free to update account storages while this is running. `unwrap()` at L798 here bite us.

This is because shrinker  effectively replaces old AccountStorageEntry with new one while removing the old AccountStorageEntry (we might be accessing this from here). `remove_dead_accounts()` never cleans up any empty AccountStorageEntries unless all of them for given slot is empty. But the `do_shrink_slot` do its work diligently by removing any empty StorageEntry manually after `handle_reclaims`, knowing they can almost certainly can remove empty one of them at the very least.

As a matter of fact, we're not affected from other store codepath, however. Because we're mangling rooted slots. other store codepath only touches non-rooted new slots. The only two which touches the rooted slots are this cleaner and shrinker.

carllin: @ryoqun, shrink and clean run serially in AccountsBackgroundService, no?

ryoqun: Yeah, you're right; Even more, the above comment is wrong: we're actually serializing the executions: https://github.com/solana-labs/solana/pull/12606/files?diff=split&w=1#diff-2099c5256db4eb5975c8834af38f6456R703

nis: So, I think this lock can go away now.

ryoqun (approved):
LGTM with nits

Great work! Also thanks for addressing various review comments! This got a lot clearer. :+1:
And, sorry for the false alarm about the shrinker<=>cleaner data race bug... ;)

ryoqun:
@carllin btw, I'd rather like to wait backporting this to v1.4 for the duration of 1 or 2 point releases. I want your previous grand work (#12194) to rip a bit on v1.4. What do you think?

carllin:
@ryoqun I think that makes sense! I'll hold off on the backport and let the other PR bake a bit, thanks for the review!

ryoqun:
@carllin Time to backport this? it looks like testnet is doing well with #12194.

Also, we're queueing #12126 and #12787 (preferably as the next patch release after this pr). Also, I rather want to spend most of time with all these prs backported on v1.4 so that we could be confident when the mainnet finally moves to the v1.4.

carllin:
@ryoqun definitely, tagged!

ryoqun:
image

it looks like snapshot got bigger a bit.

ryoqun:
https://metrics.solana.com:3000/d/V5LPmn_Zk/testnet-monitor-edge-ryoqun?panelId=67&fullscreen&orgId=2&from=now-30d&to=now&var-datasource=Solana%20Metrics%20(read-only)&var-testnet=tds&var-hostid=All

ryoqun:
image

As expected?,  clean-old-root is reduced alot: https://metrics.solana.com:3000/d/V5LPmn_Zk/testnet-monitor-edge-ryoqun?panelId=45&fullscreen&orgId=2&from=now-30d&to=now&var-datasource=Solana%20Metrics%20(read-only)&var-testnet=tds&var-hostid=All

PR #12615, opened 2020-10-01
limits number of threads in core/tests/crds_gossip.rs

behzadnouri:
#### Problem
crds_gossip tests start large networks, which with large thread-pools
will exhaust system resources, causing failures in ci tests:
https://buildkite.com/solana-labs/solana/builds/31953

#### Summary of Changes
The commit limits size of thread-pools in the test.


mvines (approved):
Thanks, I added the v1.3 label since I've seen this error occur in CI on the v1.3 branch as well

PR #12620, opened 2020-10-01
Weight push peers by how long we haven't pushed to them

### LINKED ISSUE #12409: Large amounts of traffic are being sent to offline nodes
#### Problem

After shutting down a node, the remainder of the cluster keeps sending traffic. This has been the source of frequent complaints by users who ran Solana on a home connection, shut it down, and then continued to get DDoSed by the cluster.

#### Proposed Solution

Stop sending traffic to nodes that aren't up-to-date in gossip.

This will only help with nodes that are fully offline. Congestion control is needed as well: #12410

A malicious attacker could still create fake entries since the gossip IP isn't authenticated via a three-way handshake that proves that there's a valid return path, see https://github.com/solana-labs/solana/issues/9491.

sakridge:
#### Problem

If a node goes offline, everyone starts to weight them in push because their contact info local_timestamp is not updated. This causes a DOS on the node.

#### Summary of Changes

Weight the nodes according to when we actually last sent them a push.

Fixes #12409

### DIFF
--- a/core/src/crds_gossip_push.rs
+++ b/core/src/crds_gossip_push.rs
@@ -37,6 +37,9 @@ pub const CRDS_GOSSIP_PRUNE_MSG_TIMEOUT_MS: u64 = 500;
 pub const CRDS_GOSSIP_PRUNE_STAKE_THRESHOLD_PCT: f64 = 0.15;
 pub const CRDS_GOSSIP_PRUNE_MIN_INGRESS_NODES: usize = 2;
 
+// 10 minutes
+const MAX_PUSHED_TO_TIMEOUT_MS: u64 = 10 * 60 * 1000;
+
 #[derive(Clone)]
 pub struct CrdsGossipPush {
     /// max bytes per message
@@ -50,6 +53,8 @@ pub struct CrdsGossipPush {
     /// This cache represents a lagging view of which validators
     /// currently have this node in their `active_set`
     received_cache: HashMap<Pubkey, HashMap<Pubkey, (bool, u64)>>,
+    last_pushed_to: HashMap<Pubkey, u64>,
+    last_pushed_to_cleanup_ts: u64,
     pub num_active: usize,
     pub push_fanout: usize,
     pub msg_timeout: u64,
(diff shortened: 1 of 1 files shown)


core/src/crds_gossip_push.rs
```diff
@@ -249,11 +254,14 @@ impl CrdsGossipPush {
                         trace!("new_push_messages insert {} {:?}", *p, v);
                         push_messages.entry(*p).or_default().push(v.clone());
                         self.num_pushes += 1;
+                        *self.last_pushed_to.entry(*p).or_insert(0) = now;
```
behzadnouri, line 257: this is invoked for every value which is pushed to this peer. probably would be better just to do it once outside the for loops once `push_messages` is done (line 263). something like:
```
for (pubkey, _) in &push_messages {
   self.last_pushed_to.insert(*pubkey, now);
}
```

core/src/crds_gossip_push.rs
```diff
@@ -249,11 +254,14 @@ impl CrdsGossipPush {
...
                     }
                 }
                 self.push_messages.remove(&v.label());
             }
         }
+        self.last_pushed_to
+            .retain(|_id, timestamp| now - *timestamp > MAX_PUSHED_TO_TIMEOUT_MS);
```
behzadnouri, line 264: maybe we can just use an lru cache here

behzadnouri (approved):

core/src/crds_gossip_push.rs
sakridge: Yea, I changed to that.

core/src/crds_gossip_push.rs
sakridge: yea.. is it worth the extra work? I guess it's harder to DOS and has a fixed memory footprint.

behzadnouri: i think it would also save us from iterating over the entire hashmap here

sakridge: Wouldn't you need to search the map for the highest item to reject when you hit capacity?

behzadnouri: There are O(1) implementations: https://docs.rs/lru/0.6.0/lru/
We can also switch to lru later. Probably will not make much difference for now.

carllin (approved):

behzadnouri:
lgtm

behzadnouri (dismissed):

PR #12633, opened 2020-10-02
Check CPI program is executable

jackcmay:
#### Problem

cross-program invocation's don't enforce that the program be executable

#### Summary of Changes

- Enforce that programs are executable (mirrors this check: https://github.com/solana-labs/solana/blob/42aeead6b4e97663c8941ea5761b03b56eaa4be0/runtime/src/accounts.rs#L250)
- Remove long-standing `assert`

Fixes #


jackcmay:
Note: `cargo audit` has been disabled in CI

mvines (approved):

t-nelson (approved):

PR #12634, opened 2020-10-02
Fix zero-len slice translations

jackcmay:
#### Problem

Zero length vectors have not been allocated yet so calling `as_ptr()` returns address `0x1` which is not in the translation map and thus fails with an access violation

#### Summary of Changes

If translating a zero-length vector then don't attempt to translate

Fixes #


jackcmay:
Note: `cargo audit` has been disabled in CI

mvines (approved):

t-nelson (approved):

mvines:
Manually landed 8c989da68342918f1717c60aa60fdfab7d1e676e

PR #12650, opened 2020-10-02
`solana stakes` now employs server-side filtering if only one vote account is provided

mvines:
The time to run `solana stakes <vote_pubkey>` on testnet drops from over 30 seconds to 1 second with this PR.

### DIFF
--- a/cli/src/cluster_query.rs
+++ b/cli/src/cluster_query.rs
@@ -18,7 +18,11 @@ use solana_client::{
     client_error::ClientErrorKind,
     pubsub_client::PubsubClient,
     rpc_client::{GetConfirmedSignaturesForAddress2Config, RpcClient},
-    rpc_config::{RpcLargestAccountsConfig, RpcLargestAccountsFilter},
+    rpc_config::{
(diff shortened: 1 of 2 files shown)


cli/src/cluster_query.rs
```diff
@@ -1276,17 +1280,53 @@ pub fn process_show_stakes(
...
+                                encoding: Some(rpc_filter::MemcmpEncoding::Binary),
+                            }),
+                            // Filter by `Delegation::voter_pubkey`, which begins at byte offset 124
+                            rpc_filter::RpcFilterType::Memcmp(rpc_filter::Memcmp {
+                                offset: 124,
+                                bytes: rpc_filter::MemcmpEncodedBytes::Binary(
+                                    vote_account_pubkeys[0].to_string(),
+                                ),
+                                encoding: Some(rpc_filter::MemcmpEncoding::Binary),
+                            }),
```
mvines, line 1307: It was a little clunky to build this filter by hand...

CriesofCarrots: Could definitely add some tooling here, now that I'm not the only one using this!

CriesofCarrots (approved):

PR #12653, opened 2020-10-02
Program consumable stake program

jackcmay:
#### Problem

The stake program is not structured in a way that allows other programs pull in its instruction/state definitions

#### Summary of Changes

Hackfest to show what changes might need to be made, mostly just commenting out stuff not needed or compatible with the program.

With these changes, you can pull in the stake program via:

```
[features]
no-entrypoint = []
skip-no-mangle = ["solana-sdk/skip-no-mangle"]
program = ["solana-sdk/program", "solana-stake-program/program", "solana-stake-program/no-entrypoint"]
default = ["program", "solana-sdk/default", "solana-stake-program/default"]

[dependencies]
solana-sdk = { path = "../../../../sdk/", version = "1.4.0", default-features = false, optional = true }
solana-stake-program = { path = "../../../../programs/stake/", version = "1.4.0", default-features = false, optional = true }=
```

Fixes #

### DIFF
--- a/programs/bpf/Cargo.lock
+++ b/programs/bpf/Cargo.lock
@@ -1910,6 +1910,7 @@ name = "solana-bpf-rust-sanity"
 version = "1.4.0"
 dependencies = [
  "solana-sdk",
+ "solana-stake-program",
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


aeyakovenko:
@jackcmay does this change the staking ABI?  we already have wallets using the serde based interface

jackcmay:
This is just a hackfest to give a sense of what will probably have to be reorganized in order for bpf programs to pull in stake instructions and state.  These changes don't include any abi changes that I recall.

PR #12661, opened 2020-10-03
Fix error in max root calculation

carllin:
#### Problem
My recent change to `get_max_root ()` in accounts cleanup makes the following error possible:

Given roots 1,2,5, and a slot list:
`vec![(1, true), (2, true), (5, true), (9, true)];`

Calling `purge_older_root_entries(&mut slot_list, &mut reclaims, Some(7))` will purge/reclaim `1, 2, 5` which are needed for a snapshot of slot `7`.

#### Summary of Changes
Properly find max root < given maximum allowed root

Fixes #

### DIFF
--- a/runtime/src/accounts_index.rs
+++ b/runtime/src/accounts_index.rs
@@ -140,9 +140,19 @@ impl<'a, T: 'a + Clone> AccountsIndex<T> {
         })
     }
 
-    pub fn get_max_root(roots: &HashSet<Slot>, slice: SlotSlice<T>) -> Slot {
+    // Get the maximum root <= `max_allowed_root` from the given `slice`
+    pub fn get_max_root(
+        roots: &HashSet<Slot>,
+        slice: SlotSlice<T>,
+        max_allowed_root: Option<Slot>,
+    ) -> Slot {
         let mut max_root = 0;
         for (f, _) in slice.iter() {
+            if let Some(max_allowed_root) = max_allowed_root {
+                if *f > max_allowed_root {
+                    continue;
+                }
+            }
             if *f > max_root && roots.contains(f) {
                 max_root = *f;
             }
@@ -229,10 +239,7 @@ impl<'a, T: 'a + Clone> AccountsIndex<T> {
     ) {
         let roots = &self.roots;
 
-        let mut max_root = Self::get_max_root(roots, &list);
-        if let Some(max_clean_root) = max_clean_root {
-            max_root = std::cmp::min(max_root, max_clean_root);
-        }
+        let max_root = Self::get_max_root(roots, &list, max_clean_root);
 
         reclaims.extend(
             list.iter()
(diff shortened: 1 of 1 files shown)


runtime/src/accounts_index.rs
```diff
@@ -604,4 +611,79 @@ mod tests {
...
+        assert_eq!(slot_list, vec![(1, true), (2, true), (5, true), (9, true)]);
+
+        // Add a later root, earlier slots should be reclaimed
+        slot_list = vec![(1, true), (2, true), (5, true), (9, true)];
+        index.add_root(1);
+        // Note 2 is not a root
+        index.add_root(5);
+        reclaims = vec![];
+        index.purge_older_root_entries(&mut slot_list, &mut reclaims, None);
+        let expected_reclaims = vec![(1, true), (2, true)];
```
ryoqun, line 632: nits: no need to assign to `expected_reclaims`?

ryoqun: And there is some other occurrences.

runtime/src/accounts_index.rs
```diff
@@ -229,10 +239,7 @@ impl<'a, T: 'a + Clone> AccountsIndex<T> {
     ) {
         let roots = &self.roots;
 
-        let mut max_root = Self::get_max_root(roots, &list);
-        if let Some(max_clean_root) = max_clean_root {
-            max_root = std::cmp::min(max_root, max_clean_root);
```
ryoqun, line 234: innocent looking code does hurt us... Sorry for missing this while reviewing #12194 .

runtime/src/accounts_index.rs
```diff
@@ -140,9 +140,19 @@ impl<'a, T: 'a + Clone> AccountsIndex<T> {
         })
     }
 
-    pub fn get_max_root(roots: &HashSet<Slot>, slice: SlotSlice<T>) -> Slot {
+    // Get the maximum root <= `max_allowed_root` from the given `slice`
+    pub fn get_max_root(
```
ryoqun, line 144: nits no `pub`?

runtime/src/accounts_index.rs
```diff
@@ -604,4 +611,79 @@ mod tests {
             1
         );
     }
+
+    #[test]
+    fn test_purge_older_root_entries() {
```
ryoqun, line 616: :+1: This test has good test coverage. :)

ryoqun (dismissed):
lgtm with nits.

runtime/src/accounts_index.rs
carllin: No worries, I was an idiot because I did the correct thing here (which is almost exactly the same): https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_index.rs#L107-L110 and then forgot to change this 🤦

PR #12663, opened 2020-10-05
Only fetch snapshot if it's newer than local

### LINKED ISSUE #12591: Crash from tower root assertion on validator startup
#### Problem

```
[2020-09-30T15:59:58.256677117Z INFO  solana_ledger::blockstore_processor] ledger processed in 50ms. 94 MB allocated. root=38185448, 1 fork at 38185448,
 with 1 frozen bank
[2020-09-30T15:59:58.498848424Z INFO  solana_core::consensus] adjusting lockouts (after replay up to 38185448): [38186913, 38186914, 38186915, 38186916,
 38186917, 38186918, 38186919, 38186920, 38186921, 38186922, 38186923, 38186924, 38186925, 38186926, 38186927, 38186928, 38186929, 38186930, 38186931, 3
8186932, 38186933, 38186934, 38186935, 38186936, 38186937, 38186938, 38186939, 38186940, 38186941, 38186942, 38186943]
thread 'main' panicked at 'tower root: 38186912 >= replayed root slot: 38185448', core/src/consensus.rs:778:9
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.46/src/backtrace/libunwind.rs:86
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.46/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src/libstd/sys_common/backtrace.rs:78
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
             at src/libstd/sys_common/backtrace.rs:59
   4: core::fmt::write
             at src/libcore/fmt/mod.rs:1076
   5: std::io::Write::write_fmt
             at src/libstd/io/mod.rs:1537
   6: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:62
   7: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:49
   8: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:198
  12: std::panicking::begin_panic
  13: solana_core::consensus::Tower::adjust_lockouts_after_replay
  14: solana_core::validator::Validator::new
  15: solana_validator::main
  16: std::rt::lang_start::{{closure}}
  17: std::rt::lang_start_internal::{{closure}}
             at src/libstd/rt.rs:52
  18: std::panicking::try::do_call
             at src/libstd/panicking.rs:348
  19: std::panicking::try
             at src/libstd/panicking.rs:325
  20: std::panic::catch_unwind
             at src/libstd/panic.rs:394
  21: std::rt::lang_start_internal
             at src/libstd/rt.rs:51
  22: main
  23: __libc_start_main
  24: _start
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
[2020-09-30T15:59:58.553233401Z ERROR solana_metrics::metrics] datapoint: panic program="validator" thread="main" one=1i message="panicked at core/src/consensus.rs:778:9" location="core/src/consensus.rs:778:9"
```

#### Proposed Solution

Debug and fix

ryoqun:
#### Problem

When a booting validator is fetching snapshot from the cluster, it's possible to fetch older snapshot than local...
Depending on gossip status, it's possible...

If that's the case, the persisted tower ultimately vomits a panic because it now can detect root got warped to the past.

#### Summary of Changes

Don't ever download older snapshot.

Fixes #12591

### DIFF
--- a/validator/src/main.rs
+++ b/validator/src/main.rs
@@ -29,6 +29,7 @@ use solana_perf::recycler::enable_recycler_warming;
 use solana_runtime::{
     bank_forks::{CompressionType, SnapshotConfig, SnapshotVersion},
     hardened_unpack::{unpack_genesis_archive, MAX_GENESIS_ARCHIVE_UNPACKED_SIZE},
+    snapshot_utils::get_highest_snapshot_archive_path,
 };
 use solana_sdk::{
     clock::Slot,
@@ -146,8 +147,10 @@ fn get_rpc_node(
     blacklisted_rpc_nodes: &mut HashSet<Pubkey>,
     snapshot_not_required: bool,
     no_untrusted_rpc: bool,
+    ledger_path: &std::path::Path,
 ) -> (ContactInfo, Option<(Slot, Hash)>) {
     let mut blacklist_timeout = Instant::now();
+    let mut retry_reason = None;
     loop {
         sleep(Duration::from_secs(1));
         info!("\n{}", cluster_info.contact_info_trace());
@@ -174,8 +177,12 @@ fn get_rpc_node(
         }
 
         info!(
-            "Searching for an RPC service with shred version {}...",
-            shred_version
+            "Searching for an RPC service with shred version {}{}...",
+            shred_version,
+            retry_reason
+                .clone()
+                .map(|s| format!(" (Retrying: {})", s))
+                .unwrap_or_default()
         );
 
         let rpc_peers = cluster_info
@@ -204,17 +211,21 @@ fn get_rpc_node(
         );
 
         if rpc_peers_blacklisted == rpc_peers_total {
-            // If all nodes are blacklisted and no additional nodes are discovered after 60 seconds,
-            // remove the blacklist and try them all again
             if blacklist_timeout.elapsed().as_secs() > 60 {
-                info!("Blacklist timeout expired");
+                // If all nodes are blacklisted and no additional nodes are discovered after 60 seconds,
+                // remove the blacklist and try them all again
+                retry_reason = Some("Blacklist timeout expired".to_owned());
                 blacklisted_rpc_nodes.clear();
+            } else {
+                retry_reason = Some("Wait for trusted rpc peers".to_owned());
             }
             continue;
         }
         blacklist_timeout = Instant::now();
 
-        let mut highest_snapshot_hash: Option<(Slot, Hash)> = None;
+        let mut highest_snapshot_hash: Option<(Slot, Hash)> =
+            get_highest_snapshot_archive_path(ledger_path)
+                .map(|(_path, (slot, hash, _compression))| (slot, hash));
         let eligible_rpc_peers = if snapshot_not_required {
             rpc_peers
         } else {
@@ -256,9 +267,16 @@ fn get_rpc_node(
             match highest_snapshot_hash {
                 None => {
                     assert!(eligible_rpc_peers.is_empty());
-                    info!("No snapshots available");
                 }
                 Some(highest_snapshot_hash) => {
+                    if eligible_rpc_peers.is_empty() {
+                        retry_reason = Some(format!(
+                            "Wait for newer snapshot than local: {:?}",
+                            highest_snapshot_hash
+                        ));
+                        continue;
+                    }
+
                     info!(
                         "Highest available snapshot slot is {}, available from {} node{}: {:?}",
                         highest_snapshot_hash.0,
@@ -282,6 +300,8 @@ fn get_rpc_node(
             let contact_info =
                 &eligible_rpc_peers[thread_rng().gen_range(0, eligible_rpc_peers.len())];
             return (contact_info.clone(), highest_snapshot_hash);
+        } else {
+            retry_reason = Some("No snapshots available".to_owned());
         }
     }
 }
@@ -636,6 +656,7 @@ fn rpc_bootstrap(
             &mut blacklisted_rpc_nodes,
             bootstrap_config.no_snapshot_fetch,
             bootstrap_config.no_untrusted_rpc,
+            ledger_path,
         );
         info!(
             "Using RPC service from node {}: {:?}",
--- a/validator/src/main.rs
+++ b/validator/src/main.rs
@@ -180,7 +180,7 @@ fn get_rpc_node(
             "Searching for an RPC service with shred version {}{}...",
             shred_version,
             retry_reason
-                .clone()
+                .as_ref()
                 .map(|s| format!(" (Retrying: {})", s))
                 .unwrap_or_default()
         );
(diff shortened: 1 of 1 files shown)


validator/src/main.rs
```diff
@@ -282,6 +300,8 @@ fn get_rpc_node(
             let contact_info =
                 &eligible_rpc_peers[thread_rng().gen_range(0, eligible_rpc_peers.len())];
             return (contact_info.clone(), highest_snapshot_hash);
+        } else {
+            retry_reason = Some("No snapshots available".to_owned());
```
ryoqun, line 304: I think this is more clearer place to put this message in terms of control flow..

validator/src/main.rs
```diff
@@ -256,9 +267,16 @@ fn get_rpc_node(
...
-                    info!("No snapshots available");
                 }
                 Some(highest_snapshot_hash) => {
+                    if eligible_rpc_peers.is_empty() {
+                        retry_reason = Some(format!(
+                            "Wait for newer snapshot than local: {:?}",
+                            highest_snapshot_hash
+                        ));
+                        continue;
+                    }
```
ryoqun, line 278: :meat_on_bone:

validator/src/main.rs
```diff
@@ -204,17 +211,21 @@ fn get_rpc_node(
...
+            } else {
+                retry_reason = Some("Wait for trusted rpc peers".to_owned());
             }
             continue;
         }
         blacklist_timeout = Instant::now();
 
-        let mut highest_snapshot_hash: Option<(Slot, Hash)> = None;
+        let mut highest_snapshot_hash: Option<(Slot, Hash)> =
+            get_highest_snapshot_archive_path(ledger_path)
```
ryoqun, line 227: :triangular_flag_on_post:

validator/src/main.rs
```diff
@@ -174,8 +177,12 @@ fn get_rpc_node(
...
 
         info!(
-            "Searching for an RPC service with shred version {}...",
-            shred_version
+            "Searching for an RPC service with shred version {}{}...",
+            shred_version,
+            retry_reason
+                .as_ref()
+                .map(|s| format!(" (Retrying: {})", s))
+                .unwrap_or_default()
```
ryoqun, line 185: I slightly improved logging to indicate retries more clearly.

ryoqun:
And sadly, there is no tests...

ryoqun:
Although, v1.3 doesn't cause persistent-tower-triggered panics because of this bug, I'm going to backport this there anyway because this is still a correctness error in the v1.3 as well even if v1.3 doesn't implement persistent tower.

validator/src/main.rs
carllin: Nit: I this log message doesn't account for
1) `trusted_snapshot_hashes` is empty: https://github.com/solana-labs/solana/blob/master/validator/src/main.rs#L221-L222  or  
2)`rpc_peers` is empty: https://github.com/solana-labs/solana/blob/master/validator/src/main.rs#L181-L185

carllin (dismissed):
Awesome, thanks!

validator/src/main.rs
mvines: If local has a newer snapshot why don't we just use it.

We can do this generically when trusted validators exist:  if all available snapshots from trusted validators are older than what local, use local.   There could even be a new cli argument, `--use-local-snapshot-if-newer-than <NUMBER_OF_SLOTS>` that would use a local snapshot if it's at most `NUMBER_OF_SLOTS` slots behind the newest trusted validator.  Default value of `NUMBER_OF_SLOTS` is maybe 100-500.

ryoqun: @carllin 

> Nit: I this log message doesn't account for:

Well, I misread the code too but these `continue`s are for the inner loop: https://github.com/solana-labs/solana/blob/master/validator/src/main.rs#L226

ryoqun: > If local has a newer snapshot why don't we just use it.

The main reason is that **original snapshot RPC selection logic doesn't wait for the full set of specified rpc trusted nodes to be available**. So, we can't determine that predicate to begin with.

So, if a booting validator sees a subset of trusted rpc nodes with older snapshot, it considers the snapshot to be the newest one the cluster has to offer and fetches it.


Definitely, we can wait to determine ` if all available snapshots from trusted validators are older than what local, ...`, but this requires 100% liveness of trusted validators. Given two or three trusted nodes for mainnet-beta/testnet, using 90% or 80% liveness threshold doesn't quite work well.

Also, it's operationally rare to have newer local snapshot assuming solid setup uses `--no-snapshot-fetch` anyway and/or the connected cluster aren't making newer snapshots.

So this new change is rather for corner-casing. Also it's for development purpose, where frequent and forcible restarts are rampant. ;)

So, I opted for simplicity.

ryoqun: Also, I'll consider `--use-local-snapshot-if-newer-than` for a bit...

ryoqun: > `--use-local-snapshot-if-newer-than <NUMBER_OF_SLOTS>` that would use a local snapshot if it's at most `NUMBER_OF_SLOTS` slots behind the newest trusted validator. Default value of `NUMBER_OF_SLOTS` is maybe 100-500.

Well, sadly supporting this takes a bit of effort than just plumbling (reasons below). How about merging this pr first? So that v1.4 (=~ testnet) doesn't suffer this known persistent-tower bug? (Heh, I was completely unaware of this snapshot-fetch-related failure scenario...).

The detailed reason of more efforts:

First, persistent tower never supports older root bank than voted banks to prevent double vote and other general slashing events. So, we can't use local snapshots even if it's just slightly older (practically, like 100-500) while voting is enabled. 

Persistent tower always assumes local availability of voted slots which aren't isolated like snapshot's root banks. The rationale is that you should have the slots and have replayed them if you voted on them (duh).

The only exception is to start over newer snapshots. This is for some disk failure or corrupted rocksdb.

ryoqun:
> ryoqun added the v1.3 label 2 days ago

I'll wait this for a while. I'll backport this after the mainnet-beta validators have fully transitioned to the v1.3.

validator/src/main.rs
ryoqun: @mvines Could you check my previous comment?

Especially,

> Well, sadly supporting this takes a bit of effort than just plumbling (reasons below). How about merging this pr first? So that v1.4 (=~ testnet) doesn't suffer this known persistent-tower bug? (Heh, I was completely unaware of this snapshot-fetch-related failure scenario...).

ryoqun:
> > ryoqun added the v1.3 label 2 days ago
> 
> I'll wait this for a while. I'll backport this after the mainnet-beta validators have fully transitioned to the v1.3.

Condition cleared! :)

validator/src/main.rs
mvines: Works for me.  This gets us is more in the right direction

mvines (approved):

Issue #8232, opened 2020-02-12
Leaders continue building on their own fork after partitioning from the network.


aeyakovenko:
#### Problem

Under normal operation a validator receives the block through turbine.  Because the packets are retransmitted by the entire network a validator sees that a vast majority of the network has received at least some portion of the block and will try to repair it.  Because of how Turbine works, there is a high probability that if one validator receives the block, then so did at least 50% of the network.

This property doesn’t hold for leaders own blocks.  When a leader decides to vote on its own block it does so without any signal that the rest of the network has received the block and will vote as well.

Thus a leader may continue building on its own fork while completely partitioned from the rest of the network.  If the leader has a high stake, this could lead to a lockout at a very large height away from the network.

#### Proposed Solution

Update consensus design to handle block availability, and how it relates to tendermints propose stage.

1. Banking stage need its own PoH
2. Banking stage produces a block on a fork of PoH
3. Replay stage doesn’t vote on its own block or reset PoH until it sees 1/3+ validators vote on it. 
4. Replay stage doesn’t vote or reset PoH on foreign blocks unless it receives the block with 1/3+ validators retransmitting it.
5. First retransmit needs to be signed.

Tag: @carllin @sakridge @sagar-solana 

This can be broken up into a couple PRs.

Alternative.  Bank weight doesn’t do a good job accounting for a fork that has the highest change of being confirmed based on the current set of lockouts and leader schedules.  A 1/3- minority bank should never weigh more then 2/3+ bank no matter the lockout.

aeyakovenko:
Seems like we have consensus around alt 2

ryoqun:
Because now we have a persisted tower, we can also prevent leaders starting older snapshot to create duplicate blocks at the banking stage.

ref: https://github.com/solana-labs/solana/issues/9369#issuecomment-611178234

PR #12671, opened 2020-10-05
Fix tower/blockstore unsync due to external causes

### LINKED ISSUE #13128: Solana validator v1.4.2 is not starting: thread 'main' panicked at 'slot_in_tower(43526861) < checked_slot(43526861)'
#### Problem

When I'm trying to start the validator, this error can be observed in logs:

```
Oct 24 00:26:28 ubuntu-s-8vcpu-16gb-lon1-01 solana-validator[8787]: [2020-10-24T07:26:28.053215234Z INFO  solana_core::consensus] adjusting lockouts (after replay up to 43615565): [43526861] tower root: 43526861 replayed root: 43615565
Oct 24 00:26:28 ubuntu-s-8vcpu-16gb-lon1-01 solana-validator[8787]: thread 'main' panicked at 'slot_in_tower(43526861) < checked_slot(43526861)', core/src/consensus.rs:931:21
Oct 24 00:26:28 ubuntu-s-8vcpu-16gb-lon1-01 solana-validator[8787]: stack backtrace:
Oct 24 00:26:28 ubuntu-s-8vcpu-16gb-lon1-01 solana-validator[8787]:    0: backtrace::backtrace::libunwind::trace
Oct 24 00:26:28 ubuntu-s-8vcpu-16gb-lon1-01 solana-validator[8787]:              at ./cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.46/src/backtrace/libunwind.rs
Oct 24 00:26:28 ubuntu-s-8vcpu-16gb-lon1-01 solana-validator[8787]:    1: backtrace::backtrace::trace_unsynchronized
Oct 24 00:26:28 ubuntu-s-8vcpu-16gb-lon1-01 solana-validator[8787]:              at ./cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.46/src/backtrace/mod.rs:66
Oct 24 00:26:28 ubuntu-s-8vcpu-16gb-lon1-01 solana-validator[8787]:    2: std::sys_common::backtrace::_print_fmt
Oct 24 00:26:28 ubuntu-s-8vcpu-16gb-lon1-01 solana-validator[8787]:              at src/libstd/sys_common/backtrace.rs:78
Oct 24 00:26:28 ubuntu-s-8vcpu-16gb-lon1-01 solana-validator[8787]:    3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
Oct 24 00:26:28 ubuntu-s-8vcpu-16gb-lon1-01 solana-validator[8787]:              at src/libstd/sys_common/backtrace.rs:59
Oct 24 00:26:28 ubuntu-s-8vcpu-16gb-lon1-01 solana-validator[8787]:    4: core::fmt::write
Oct 24 00:26:28 ubuntu-s-8vcpu-16gb-lon1-01 solana-validator[8787]:              at src/libcore/fmt/mod.rs:1076
Oct 24 00:26:28 ubuntu-s-8vcpu-16gb-lon1-01 solana-validator[8787]:    5: std::io::Write::write_fmt
Oct 24 00:26:28 ubuntu-s-8vcpu-16gb-lon1-01 solana-validator[8787]:              at src/libstd/io/mod.rs:1537
Oct 24 00:26:28 ubuntu-s-8vcpu-16gb-lon1-01 solana-validator[8787]:    6: std::sys_common::backtrace::_print
Oct 24 00:26:28 ubuntu-s-8vcpu-16gb-lon1-01 solana-validator[8787]:              at src/libstd/sys_common/backtrace.rs:62
Oct 24 00:26:28 ubuntu-s-8vcpu-16gb-lon1-01 solana-validator[8787]:    7: std::sys_common::backtrace::print
Oct 24 00:26:28 ubuntu-s-8vcpu-16gb-lon1-01 solana-validator[8787]:              at src/libstd/sys_common/backtrace.rs:49
Oct 24 00:26:28 ubuntu-s-8vcpu-16gb-lon1-01 solana-validator[8787]:    8: std::panicking::default_hook::{{closure}}
Oct 24 00:26:28 ubuntu-s-8vcpu-16gb-lon1-01 solana-validator[8787]:              at src/libstd/panicking.rs:198
Oct 24 00:26:28 ubuntu-s-8vcpu-16gb-lon1-01 solana-validator[8787]:    9: std::panicking::default_hook
Oct 24 00:26:28 ubuntu-s-8vcpu-16gb-lon1-01 solana-validator[8787]:              at src/libstd/panicking.rs:217
Oct 24 00:26:28 ubuntu-s-8vcpu-16gb-lon1-01 solana-validator[8787]:   10: solana_metrics::metrics::set_panic_hook::{{closure}}::{{closure}}
Oct 24 00:26:28 ubuntu-s-8vcpu-16gb-lon1-01 solana-validator[8787]:   11: std::panicking::rust_panic_with_hook
Oct 24 00:26:28 ubuntu-s-8vcpu-16gb-lon1-01 solana-validator[8787]:              at src/libstd/panicking.rs:530
Oct 24 00:26:28 ubuntu-s-8vcpu-16gb-lon1-01 solana-validator[8787]:   12: rust_begin_unwind
Oct 24 00:26:28 ubuntu-s-8vcpu-16gb-lon1-01 solana-validator[8787]:              at src/libstd/panicking.rs:437
Oct 24 00:26:28 ubuntu-s-8vcpu-16gb-lon1-01 solana-validator[8787]:   13: std::panicking::begin_panic_fmt
Oct 24 00:26:28 ubuntu-s-8vcpu-16gb-lon1-01 solana-validator[8787]:              at src/libstd/panicking.rs:391
Oct 24 00:26:28 ubuntu-s-8vcpu-16gb-lon1-01 solana-validator[8787]:   14: solana_core::consensus::Tower::adjust_lockouts_after_replay
Oct 24 00:26:28 ubuntu-s-8vcpu-16gb-lon1-01 solana-validator[8787]:   15: solana_core::validator::Validator::new
Oct 24 00:26:28 ubuntu-s-8vcpu-16gb-lon1-01 solana-validator[8787]:   16: solana_validator::main
Oct 24 00:26:28 ubuntu-s-8vcpu-16gb-lon1-01 solana-validator[8787]:   17: std::rt::lang_start::{{closure}}
Oct 24 00:26:28 ubuntu-s-8vcpu-16gb-lon1-01 solana-validator[8787]:   18: std::rt::lang_start_internal::{{closure}}
Oct 24 00:26:28 ubuntu-s-8vcpu-16gb-lon1-01 solana-validator[8787]:              at src/libstd/rt.rs:52
```

v1.4.2

ryoqun:
#### Problem

there are various persisted tower assertions which shouldn't be triggered (thanks for reporting, @sakridge )

All of them is related to flushing timing issue or unintended ledger manual manipulation:

|FKFE8cSNZ5AzTNq4VsC7y6wEeoYWbzVRfxuzc36Sqr5y | validator | solana-replay-stage | panicked at 'assertion failed: !last_vote_ancestors.contains(&switch_slot)', core/src/consensus.rs:540:17
-- | -- | -- | --

|twP716sMjTFynrZRSCKkHhoeHaQEXnR2bzqDEzuxP81 |	validator	|main	|panicked at 'at least 1 parent slot must be found', core/src/consensus.rs:1094:13
-- | -- | -- | --

#### Summary of Changes

Adjust those assertions to accomodate any kind of incomplete unsync between tower <=> blockstore. Also, support a future tower which is descendant of ledger root.

Fixes #13128

context: #10718

### DIFF
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -554,8 +554,8 @@ impl Tower {
                     // validator SEGV, OS/HW crash, or plain No Free Space FS error.
 
                     // When we're in this clause, it basically means validator is badly running
-                    // with a future tower and RE-CREATING a block for one of past slots as the leader
-                    // of the slot, especially problematic is last_voted_slot.
+                    // with a future tower while replaying past slots, especially problematic is
+                    // last_voted_slot.
                     // So, don't re-vote on it by returning pseudo FailedSwitchThreshold, otherwise
                     // there would be slashing because of double vote on one of last_vote_ancestors.
                     // (Well, needless to say, re-creating the duplicate block must be handled properly
@@ -563,10 +563,9 @@ impl Tower {
                     //
                     // To be specific, the replay stage is tricked into a false perception where
                     // last_vote_ancestors is AVAILABLE for descendant-of-`switch_slot`,  stale, and
-                    // stray slots (which should always be empty_ancestors). This is caused by the banking
-                    // stage's incorrectly creating new duplicate block.
+                    // stray slots (which should always be empty_ancestors).
                     //
-                    // This is covered by test_future_tower in local_cluster
+                    // This is covered by test_future_tower_* in local_cluster
                     SwitchForkDecision::FailedSwitchThreshold(0, total_stake)
                 };
 
@@ -903,10 +902,11 @@ impl Tower {
                 self.initialize_root(replayed_root);
             } else {
                 // This should never occur under normal operation.
-                // If this validator has leader slots while voting suspended this way,
+                // While this validator's voting is suspended this way,
                 // suspended_decision_due_to_major_unsynced_ledger() will be also touched.
                 let message = format!(
-                    "For some reason, we're REPROCESSING already voted and ROOTED slots; \
+                    "For some reason, we're REPROCESSING slots which has already been \
+                     voted and ROOTED by us; \
                      VOTING will be SUSPENDED UNTIL {}!",
                     last_voted_slot,
                 );
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -547,7 +547,7 @@ impl Tower {
                     &empty_ancestors
                 };
 
-                let suspended_decision_due_to_major_unsynced_ledger = |total_stake| {
+                let suspended_decision_due_to_major_unsynced_ledger = || {
                     // This peculiar corner handling is needed mainly for a tower which is newer than
                     // blockstore. (Yeah, we tolerate it for ease of maintaining validator by operators)
                     // This condition could be introduced by manual ledger mishandling,
@@ -604,7 +604,7 @@ impl Tower {
                             last_voted_slot
                         );
                     } else {
-                        return suspended_decision_due_to_major_unsynced_ledger(total_stake);
+                        return suspended_decision_due_to_major_unsynced_ledger();
                     }
                 }
 
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -515,6 +515,60 @@ impl Tower {
             .map(|last_voted_slot| {
                 let root = self.root();
                 let empty_ancestors = HashSet::default();
+                let empty_ancestors_due_to_minor_unsynced_ledger = || {
+                    // This condition (stale stray last vote) shouldn't occur under normal validator
+                    // operation, indicating something unusual happened.
+                    // This condition could be introduced by manual ledger mishandling,
+                    // validator SEGV, OS/HW crash, or plain No Free Space FS error.
+
+                    // However, returning empty ancestors as a fallback here shouldn't result in
+                    // slashing by itself (Note that we couldn't fully preclude any kind of slashing if
+                    // the failure was OS or HW level).
+
+                    // Firstly, lockout is ensured elsewhere.
+
+                    // Also, there is no risk of optimistic conf. violation. Although empty ancestors
+                    // could result in incorrect (= more than actual) locked_out_stake and
+                    // false-positive SwitchProof later in this function, there should be no such a
+                    // heavier fork candidate, first of all, if the last vote (or any of its
+                    // unavailable ancestors) were already optimistically confirmed.
+                    // The only exception is that other validator is already violating it...
+                    if self.is_first_switch_check() && switch_slot < last_voted_slot {
+                        // `switch < last` is needed not to warn! this message just because of using
+                        // newer snapshots on validator restart
+                        let message = format!(
+                          "bank_forks doesn't have corresponding data for the stray restored \
+                           last vote({}), meaning some inconsistency between saved tower and ledger.",
+                          last_voted_slot
+                        );
+                        warn!("{}", message);
+                        datapoint_warn!("tower_warn", ("warn", message, String));
+                    }
+                    &empty_ancestors
+                };
+
+                let suspended_decision_due_to_major_unsynced_ledger = |total_stake: u64| -> SwitchForkDecision {
+                    // This peculiar corner handling is needed mainly for a tower which is newer than
+                    // blockstore. (Yeah, we tolerate it for ease of maintaining validator by operators)
+                    // This condition could be introduced by manual ledger mishandling,
+                    // validator SEGV, OS/HW crash, or plain No Free Space FS error.
+
+                    // When we're in this clause, it basically means validator is badly running
+                    // with a future tower and RE-CREATING a block for one of past slots as the leader
+                    // of the slot, especially problematic is last_voted_slot.
+                    // So, don't re-vote on it by returning pseudo FailedSwitchThreshold, otherwise
+                    // there would be slashing because of double vote on one of last_vote_ancestors.
+                    // (Well, needless to say, re-creating the duplicate block must be handled properly
+                    // at the banking stage: https://github.com/solana-labs/solana/issues/8232)
+                    //
+                    // To be specific, the replay stage is tricked into a false perception where
+                    // last_vote_ancestors is AVAILABLE for descendant-of-`switch_slot`,  stale, and
+                    // stray slots (which should always be empty_ancestors). This is caused by the banking
+                    // stage's incorrectly creating new duplicate block.
+                    //
+                    // This is covered by test_future_tower in local_cluster
+                    SwitchForkDecision::FailedSwitchThreshold(0, total_stake)
+                };
 
                 let last_vote_ancestors =
                     ancestors.get(&last_voted_slot).unwrap_or_else(|| {
@@ -529,35 +583,7 @@ impl Tower {
                             // all of them.
                             panic!("no ancestors found with slot: {}", last_voted_slot);
                         } else {
-                            // This condition (stale stray last vote) shouldn't occur under normal validator
-                            // operation, indicating something unusual happened.
-                            // Possible causes include: OS/HW crash, validator process crash, only saved tower
-                            // is moved over to a new setup, etc...
-
-                            // However, returning empty ancestors as a fallback here shouldn't result in
-                            // slashing by itself (Note that we couldn't fully preclude any kind of slashing if
-                            // the failure was OS or HW level).
-
-                            // Firstly, lockout is ensured elsewhere.
-
-                            // Also, there is no risk of optimistic conf. violation. Although empty ancestors
-                            // could result in incorrect (= more than actual) locked_out_stake and
-                            // false-positive SwitchProof later in this function, there should be no such a
-                            // heavier fork candidate, first of all, if the last vote (or any of its
-                            // unavailable ancestors) were already optimistically confirmed.
-                            // The only exception is that other validator is already violating it...
-                            if self.is_first_switch_check() && switch_slot < last_voted_slot {
-                                // `switch < last` is needed not to warn! this message just because of using
-                                // newer snapshots on validator restart
-                                let message = format!(
-                                  "bank_forks doesn't have corresponding data for the stray restored \
-                                   last vote({}), meaning some inconsistency between saved tower and ledger.",
-                                  last_voted_slot
-                                );
-                                warn!("{}", message);
-                                datapoint_warn!("tower_warn", ("warn", message, String));
-                            }
-                            &empty_ancestors
+                            empty_ancestors_due_to_minor_unsynced_ledger()
                         }
                     });
 
@@ -578,25 +604,7 @@ impl Tower {
                             last_voted_slot
                         );
                     } else {
-                        // This peculiar corner handling is needed mainly for a tower which is newer than
-                        // blockstore. (Yeah, we tolerate it for ease of maintaining validator by operators)
-                        // This condition could be introduced by manual ledger mishandling,
-                        // validator SEGV, OS/HW crash, or plain No Free Space FS error.
-                        // When we're in this clause, it basically means validator is badly running
-                        // with a future tower and RE-CREATING a block for one of past slots as the leader
-                        // of the slot, especially problematic is last_voted_slot.
-                        // So, don't re-vote on it by returning pseudo FailedSwitchThreshold, otherwise
-                        // there would be slashing because of double vote on one of last_vote_ancestors.
-                        // (Well, needless to say, re-creating the duplicate block must be handled properly
-                        // at the banking stage: https://github.com/solana-labs/solana/issues/8232)
-                        //
-                        // To be specific, the replay stage is tricked into a false perception where
-                        // last_vote_ancestors is AVAILABLE for descendant-of-`switch_slot`,  stale, and
-                        // stray slots (which should always be empty_ancestors). This is caused by the banking
-                        // stage's incorrectly creating new duplicate block.
-                        //
-                        // This is covered by test_future_tower in local_cluster
-                        return SwitchForkDecision::FailedSwitchThreshold(0, total_stake);
+                        return suspended_decision_due_to_major_unsynced_ledger(total_stake);
                     }
                 }
 
@@ -880,7 +888,8 @@ impl Tower {
 
         if let Some(last_voted_slot) = self.last_voted_slot() {
             if tower_root <= replayed_root {
-                // Normally, we goes into this clause with possible help of reconcile_blockstore_roots_with_tower()
+                // Normally, we goes into this clause with possible help of
+                // reconcile_blockstore_roots_with_tower()
                 if slot_history.check(last_voted_slot) == Check::TooOld {
                     // We could try hard to anchor with other older votes, but opt to simplify the
                     // following logic
@@ -892,6 +901,9 @@ impl Tower {
                 self.adjust_lockouts_with_slot_history(slot_history)?;
                 self.initialize_root(replayed_root);
             } else {
+                // This should never occur under normal operation.
+                // If this validator has leader slots while voting suspended this way,
+                // suspended_decision_due_to_major_unsynced_ledger() will be also touched.
                 let message = format!(
                     "For some reason, we're REPROCESSING already voted and ROOTED slots; \
                      VOTING will be SUSPENDED UNTIL {}!",
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -569,13 +569,36 @@ impl Tower {
                     return SwitchForkDecision::SameFork;
                 }
 
-                assert!(
-                    !last_vote_ancestors.contains(&switch_slot),
-                    "Should never consider switching to slot ({}), which is ancestors({:?}) of last vote: {}",
-                    switch_slot,
-                    last_vote_ancestors,
-                    last_voted_slot
-                );
+                if last_vote_ancestors.contains(&switch_slot) {
+                    if !self.is_stray_last_vote() {
+                        panic!(
+                            "Should never consider switching to slot ({}), which is ancestors({:?}) of last vote: {}",
+                            switch_slot,
+                            last_vote_ancestors,
+                            last_voted_slot
+                        );
+                    } else {
+                        // This peculiar corner handling is needed mainly for a tower which is newer than
+                        // blockstore. (Yeah, we tolerate it for ease of maintaining validator by operators)
+                        // This condition could be introduced by manual ledger mishandling,
+                        // validator SEGV, OS/HW crash, or plain No Free Space FS error.
+                        // When we're in this clause, it basically means validator is badly running
+                        // with a future tower and RE-CREATING a block for one of past slots as the leader
+                        // of the slot, especially problematic is last_voted_slot.
+                        // So, don't re-vote on it by returning pseudo FailedSwitchThreshold, otherwise
+                        // there would be slashing because of double vote on one of last_vote_ancestors.
+                        // (Well, needless to say, re-creating the duplicate block must be handled properly
+                        // at the banking stage: https://github.com/solana-labs/solana/issues/8232)
+                        //
+                        // To be specific, the replay stage is tricked into a false perception where
+                        // last_vote_ancestors is AVAILABLE for descendant-of-`switch_slot`,  stale, and
+                        // stray slots (which should always be empty_ancestors). This is caused by the banking
+                        // stage's incorrectly creating new duplicate block.
+                        //
+                        // This is covered by test_future_tower in local_cluster
+                        return SwitchForkDecision::FailedSwitchThreshold(0, total_stake);
+                    }
+                }
 
                 // By this point, we know the `switch_slot` is on a different fork
                 // (is neither an ancestor nor descendant of `last_vote`), so a
@@ -846,14 +869,6 @@ impl Tower {
         );
         assert_eq!(slot_history.check(replayed_root), Check::Found);
 
-        // reconcile_blockstore_roots_with_tower() should already have aligned these.
-        assert!(
-            tower_root <= replayed_root,
-            format!(
-                "tower root: {:?} >= replayed root slot: {}",
-                tower_root, replayed_root
-            )
-        );
         assert!(
             self.last_vote == Vote::default() && self.lockouts.votes.is_empty()
                 || self.last_vote != Vote::default() && !self.lockouts.votes.is_empty(),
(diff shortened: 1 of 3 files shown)


core/src/consensus.rs
```diff
@@ -855,11 +857,11 @@ impl Tower {
             }
 
             if let Some(checked_slot) = checked_slot {
-                // This is really special, only if tower is initialized (root = slot 0) for genesis and contains
-                // a vote (= slot 0) for the genesis, the slot 0 can repeat only once
-                let voting_from_genesis = *slot_in_tower == checked_slot && *slot_in_tower == 0;
+                // This is really special, only if tower is initialized and contains
+                // a vote for the root, the root slot can repeat only once
+                let voting_for_root = *slot_in_tower == checked_slot && *slot_in_tower == tower_root;
```
ryoqun, line 862: I've found this also:

```

BarcaR5KvTdESod5UZTbPVdcjZD2ZSLFMQws3N5Gsqdp | validator | main | panicked at 'assertion failed: *slot_in_tower < checked_slot', core/src/consensus.rs:865:21
-- | -- | -- | --



```

ryoqun:
FYI @carllin

core/src/consensus.rs
```diff
@@ -537,7 +537,7 @@ impl Tower {
 
                 // Should never consider switching to an ancestor
                 // of your last vote
-                assert!(!last_vote_ancestors.contains(&switch_slot));
+                assert!(!last_vote_ancestors.contains(&switch_slot), "{} in {:?}", switch_slot, last_vote_ancestors);
```
ryoqun, line 540: more proper message please:
```
[2020-10-06T10:59:16.409837934Z WARN  solana_sys_tuner] Failed to send tuning request, is `solana-sys-tuner` running? Os { code: 2, kind: NotFound, message: "No such file or directory" }
[2020-10-06T10:59:16.413971684Z WARN  solana_core::replay_stage] PARTITION DETECTED waiting to join heaviest fork: 50 last vote: 51, reset slot: 50
thread 'solana-replay-stage' panicked at '50 in {33, 40, 37, 49, 50, 41, 38, 39, 25, 20, 43, 42, 28, 29, 36, 31, 32, 46, 22, 45, 21, 34, 26, 23, 48, 30, 24, 35, 27, 44, 47}', core/src/consensus.rs:540:17
stack backtrace:
   0:     0x55d4a4656310 - std::backtrace_rs::backtrace::libunwind::trace::ha1eabb142084ea37
                               at /rustc/f44c6e4e288e4f49c5c257b2e2cb2ad5918f26a3/library/std/src/../../backtrace/src/backtrace/libunwind.rs:96
   1:     0x55d4a4656310 - std::backtrace_rs::backtrace::trace_unsynchronized::hbb8825661d997281

```

ryoqun:
> FKFE8cSNZ5AzTNq4VsC7y6wEeoYWbzVRfxuzc36Sqr5y | validator | solana-replay-stage | panicked at 'assertion failed: !last_vote_ancestors.contains(&switch_slot)', core/src/consensus.rs:540:17
> -- | -- | -- | --

After some extensitve testing, I've failed to reproduce with #12663. So, I think the pr manages to fix this assertion as well.

ryoqun:
> > FKFE8cSNZ5AzTNq4VsC7y6wEeoYWbzVRfxuzc36Sqr5y
> > validator
> > solana-replay-stage
> > panicked at 'assertion failed: !last_vote_ancestors.contains(&switch_slot)', core/src/consensus.rs:540:17
> 
> After some extensitve testing, I've failed to reproduce with #12663. So, I think the pr manages to fix this assertion as well.

I've finally managed to produce this: https://github.com/solana-labs/solana/pull/12350/files#r502317201

core/src/consensus.rs
ryoqun: this is also found on the wild: #13128

core/src/consensus.rs
```diff
@@ -864,16 +879,39 @@ impl Tower {
...
+            } else {
+                let message = format!(
+                    "For some reason, we're REPROCESSING already voted and ROOTED slots; \
+                     VOTING will be SUSPENDED UNTIL {}!",
                     last_voted_slot,
-                    slot_history.oldest(),
-                ));
+                );
+                error!("{}", message);
+                datapoint_error!("tower_error", ("error", message, String));
```
ryoqun, line 901: Originally, I was planning to abort the `solana-validator`. But I think we should prefer operation easiness over perfection of uptime. I'll write about this a bit later.

Anyway, might be good adding `solana-validator --force-reboot`?

ryoqun: this is original implementation: https://github.com/ryoqun/solana/commit/d8c838686fb0493320977a0889cafcacb0e973a3#r43557231

core/src/consensus.rs
```diff
@@ -569,13 +569,36 @@ impl Tower {
...
+                if last_vote_ancestors.contains(&switch_slot) {
+                    if !self.is_stray_last_vote() {
+                        panic!(
+                            "Should never consider switching to slot ({}), which is ancestors({:?}) of last vote: {}",
+                            switch_slot,
+                            last_vote_ancestors,
+                            last_voted_slot
+                        );
+                    } else {
+                        // This peculiar corner handling is needed mainly for a tower which is newer than
```
ryoqun, line 581: @carllin hehe, this was quite hard... ;)

local-cluster/tests/local_cluster.rs
```diff
@@ -1865,6 +1869,88 @@ fn do_test_optimistic_confirmation_violation_with_or_without_tower(with_tower: b
     }
 }
 
+#[test]
+#[serial]
+fn test_future_tower() {
```
ryoqun, line 1874: this test covers over all functioning of all of affected assertions.

core/src/consensus.rs
```diff
@@ -569,13 +569,36 @@ impl Tower {
...
+                        // there would be slashing because of double vote on one of last_vote_ancestors.
+                        // (Well, needless to say, re-creating the duplicate block must be handled properly
+                        // at the banking stage: https://github.com/solana-labs/solana/issues/8232)
+                        //
+                        // To be specific, the replay stage is tricked into a false perception where
+                        // last_vote_ancestors is AVAILABLE for descendant-of-`switch_slot`,  stale, and
+                        // stray slots (which should always be empty_ancestors). This is caused by the banking
+                        // stage's incorrectly creating new duplicate block.
+                        //
+                        // This is covered by test_future_tower in local_cluster
```
ryoqun, line 598: FYI, SameFork causes test failure as intended.

core/src/consensus.rs
```diff
@@ -569,13 +569,36 @@ impl Tower {
...
 
-                assert!(
-                    !last_vote_ancestors.contains(&switch_slot),
-                    "Should never consider switching to slot ({}), which is ancestors({:?}) of last vote: {}",
-                    switch_slot,
-                    last_vote_ancestors,
-                    last_voted_slot
-                );
+                if last_vote_ancestors.contains(&switch_slot) {
+                    if !self.is_stray_last_vote() {
```
ryoqun, line 573: Ideally, I wanted to move this into heaviest_subtree_fork_choice, but failed. no investigation done. I'm now out of stamina.

core/src/consensus.rs
```diff
@@ -1172,12 +1213,23 @@ pub fn reconcile_blockstore_roots_with_tower(
                 ),
             })
             .collect();
-        assert!(
```
ryoqun, line 1175: Removed entirely this assertion

core/src/consensus.rs
```diff
@@ -846,14 +869,6 @@ impl Tower {
         );
         assert_eq!(slot_history.check(replayed_root), Check::Found);
 
-        // reconcile_blockstore_roots_with_tower() should already have aligned these.
-        assert!(
```
ryoqun, line 850: entirely removed this assertion.

core/src/consensus.rs
```diff
@@ -569,13 +569,36 @@ impl Tower {
                     return SwitchForkDecision::SameFork;
                 }
 
-                assert!(
```
ryoqun, line 572: this assertion was relaxed to accommodate new observed legitimate behavior.

ryoqun:
Pheeeew, I've finally make this pr out of draft status. :) Could you review this when you're available. Thanks! :)

core/src/consensus.rs
```diff
@@ -569,13 +569,36 @@ impl Tower {
...
+                        // blockstore. (Yeah, we tolerate it for ease of maintaining validator by operators)
+                        // This condition could be introduced by manual ledger mishandling,
+                        // validator SEGV, OS/HW crash, or plain No Free Space FS error.
+                        // When we're in this clause, it basically means validator is badly running
+                        // with a future tower and RE-CREATING a block for one of past slots as the leader
+                        // of the slot, especially problematic is last_voted_slot.
+                        // So, don't re-vote on it by returning pseudo FailedSwitchThreshold, otherwise
+                        // there would be slashing because of double vote on one of last_vote_ancestors.
+                        // (Well, needless to say, re-creating the duplicate block must be handled properly
+                        // at the banking stage: https://github.com/solana-labs/solana/issues/8232)
```
ryoqun, line 591: surface #8232

core/src/consensus.rs
ryoqun: this is already fixed and merged by #13135

core/src/consensus.rs
ryoqun: > Originally, I was planning to abort the `solana-validator`. But I think we should prefer operation easiness over perfection of uptime. I'll write about this a bit later.

@carllin Generally speaking, I'm at the rigorous school of fail-first-loudly. So, It's usual for me to abort here because future persistent tower is really odd thing at the very least. 

However, I bent my usual philosophy this time. Because correcting this manually is rather hard (for rough idea of recommended recovery steps, see: https://github.com/ryoqun/solana/commit/d8c838686fb0493320977a0889cafcacb0e973a3#r43557231). And I think we shouldn't require/expect validator operators to have that level of skill and acquittance of our validator workings. Instead, I think they should care about their external environment like DC, security, operation assertiveness, etc.

Based on this assumption, I thought it's worthwhile to make validator robust and self-healing as much as possible with being low-maintainance as long as it doesn't increase the likelihood of slashing. Fortunately, we can make this case so, just by allowing future tower. Thus, here is the reason of current form of this pr.


I think usual occurrences of future tower will be caused by unclean machine shutdown or kill -9 or no free space. So, I guess the gap of slots between blockstore and tower will relatively be  small. On the other hand, I think mixed tower <=> blockstore due to manual mis operation should be very rare, compared to the crashes.

So, I think the opportunity cost (due to required human introspection) would outweigh the right and good practice of correcting and validating the aborted validator. Doing the right and correct here doesn't make sense economically if they can just recover just by waiting a min or several.

Then, the only remaining concern of not aborting is there are false negatives for the detection of erroneous conditions.

But, I think this is Ok. Even if validator don't immediately abort on future tower, I think we can expect operators can detect significant downtime of their validator with `solana-watchtower` and other their methods. Because largely-gaped future tower disables voting for significant duration.

And here is the unavoidable false negative: validator could indefinitely be stalled if an operator mixes up saved tower file and the saved tower is from future and they use save identity key across different clusters and they don't use `solana-watchtower`. I think this is so rare to occur in the wild. And we could safely blame operators if that should occur. ;)

ryoqun:
@carllin Could you review this in your free time? Thanks!

local-cluster/tests/local_cluster.rs
```diff
@@ -1865,6 +1869,88 @@ fn do_test_optimistic_confirmation_violation_with_or_without_tower(with_tower: b
...
+    }
+
+    cluster.restart_node(&validator_a_pubkey, validator_a_info);
+
+    let mut newly_rooted = false;
+    for _ in 0..300 {
+        sleep(Duration::from_millis(100));
+
+        if let Some(root) = root_in_tower(&val_a_ledger_path, &validator_a_pubkey) {
+            if root >= 25 {
```
carllin, line 1931: maybe a comment on why `25` is special, it seems like anything `>= 10 will work right`?

local-cluster/tests/local_cluster.rs
```diff
@@ -1865,6 +1869,88 @@ fn do_test_optimistic_confirmation_violation_with_or_without_tower(with_tower: b
...
+    let _validator_a_info = cluster.exit_node(&validator_a_pubkey);
+    if newly_rooted {
+        // there should be no forks; i.e. monotonically increasing voted slots
+        let last_vote = last_vote_in_tower(&val_a_ledger_path, &validator_a_pubkey).unwrap();
+        let blockstore = open_blockstore(&val_a_ledger_path);
+        assert_eq!(
+            AncestorIterator::new(last_vote, &blockstore)
+                .take_while(|a| *a >= 25)
+                .collect::<Vec<_>>(),
+            (25..=58).rev().collect::<Vec<_>>()
```
carllin, line 1946: maybe `25..=25 + MAX_LOCKOUT_HISTORY + 1` to be more explicit?

core/src/consensus.rs
```diff
@@ -569,13 +569,36 @@ impl Tower {
...
+                            last_vote_ancestors,
+                            last_voted_slot
+                        );
+                    } else {
+                        // This peculiar corner handling is needed mainly for a tower which is newer than
+                        // blockstore. (Yeah, we tolerate it for ease of maintaining validator by operators)
+                        // This condition could be introduced by manual ledger mishandling,
+                        // validator SEGV, OS/HW crash, or plain No Free Space FS error.
+                        // When we're in this clause, it basically means validator is badly running
+                        // with a future tower and RE-CREATING a block for one of past slots as the leader
```
carllin, line 586: For my clarification, does the case described here look something like:

Validator `A` is generated blocks for slots `1` through `5`, but only voted on `1` through `4` before it crashed.

`Tower: 1 -> 2 -> 3 -> 4`

Blockstore is corrupted deleted, validator starts up again from snapshot of slot `1`.

After receiving the blocks from the network, the validator freezes bank `3`, it cannot vote on it due to lockouts.
When the validator creates bank `4`, now bank forks has the ancestors of bank 4, so `ancestors(4).contains(3)` will return true.

Now before bank `4` can be frozen, heaviest for choice rule will try to pick `3` again, this time `ancestors(4).contains(3)` returns true so the panic occurs.

If this is the case, luckily the current `is_locked_out` check: https://github.com/solana-labs/solana/blob/master/core/src/replay_stage.rs#L1578,  specifically this check:
https://github.com/solana-labs/solana/blob/master/core/src/consensus.rs#L474-L476, should at least make sure the vote doesn't happen.

Unfortunately, I don't think this prevents creating a duplicate block. For instance in the example above, the validator previously created a block `5`, but has no idea that it's done so because the ledger is gone. Then even without voting for block `4`, PoH will reset to `4`:  https://github.com/ryoqun/solana/blob/master/core/src/replay_stage.rs#L529 because it's the heaviest bank, and the leader may generate another version of block `5`.

core/src/consensus.rs
carllin: thanks for the detective work! Was this caused after a deleted/corrupted ledger?

core/src/consensus.rs
```diff
@@ -864,16 +879,39 @@ impl Tower {
...
+                    // following logic
+                    return Err(TowerError::TooOldTower(
+                        last_voted_slot,
+                        slot_history.oldest(),
+                    ));
+                }
+                self.adjust_lockouts_with_slot_history(slot_history)?;
+                self.initialize_root(replayed_root);
+            } else {
+                let message = format!(
```
carllin, line 895: Maybe we should add a comment that this should only happen on ledger deletion

core/src/consensus.rs
carllin: I think another alternative to this is if we signal the operator that they must go fetch a snapshot > their tower root, it might be easier to reason about in the future as this grows in complexity 👻

core/src/consensus.rs
ryoqun: basically yes. :)

core/src/consensus.rs
ryoqun: @carllin yeah. all are true! So glad you understand correctly. And thanks for writing explation back to me for me, taking my work. xD

core/src/consensus.rs
ryoqun: > I think another alternative to this is if we signal the operator that they must go fetch a snapshot >= their tower root, it might be easier to reason about in the future as this grows in complexity

You're right. I also considered it.

https://github.com/ryoqun/solana/commit/d8c838686fb0493320977a0889cafcacb0e973a3#diff-8aa35f64ce162f53fcdda9823deb9a87ac8a0c35d0738533e35c2241c726e769R755-R756:

>                          use snapshot (newer than {}) fetched from the cluster by running without \
>                         --no-snapshot-fertch or populate missing data in blockstore by running \


But, recommending newer snapshot might not be an option for seriously operating _mainnet_ validators. That's because they have to trust the newer snapshot once more after their initial TOFU. (that's because secured snapshot isn't still a thing) And I hope they insist to keep their ledger integrity so even after the catastrophic failure like this, as the guardian of quite large amount of assets..

So, I took the red pill (complexity at our side). ;) I think these still are manageable.

local-cluster/tests/local_cluster.rs
ryoqun: yeah, that's right. I clarified: 24c66239e01f7be32b36524fafb9ca68af06dcab

local-cluster/tests/local_cluster.rs
ryoqun: done: 24c66239e01f7be32b36524fafb9ca68af06dcab

core/src/consensus.rs
ryoqun: yeah, I re-organized the comments a bit; it got a longer 75e76fd519021b72fd5ef48c0cb5d1f023cb3535

ryoqun:
@carllin Thanks for the review! I think I've addressed all your comments.

core/src/consensus.rs
```diff
@@ -515,6 +515,60 @@ impl Tower {
...
+                };
+
+                let suspended_decision_due_to_major_unsynced_ledger = |total_stake: u64| -> SwitchForkDecision {
+                    // This peculiar corner handling is needed mainly for a tower which is newer than
+                    // blockstore. (Yeah, we tolerate it for ease of maintaining validator by operators)
+                    // This condition could be introduced by manual ledger mishandling,
+                    // validator SEGV, OS/HW crash, or plain No Free Space FS error.
+
+                    // When we're in this clause, it basically means validator is badly running
+                    // with a future tower and RE-CREATING a block for one of past slots as the leader
```
carllin, line 557: hmm, the validator doesn't even have to be leader/recreating a block for this to happen right? For example in this example here: https://github.com/solana-labs/solana/pull/12671#discussion_r513830724, validator `A` doesn't necessarily have to have generated blocks 1-4, those could be any validator's blocks, the key is it will still try to call the switching logic on a previously seen block.

Side note, could we include a simple diagram like the one in https://github.com/solana-labs/solana/pull/12671#discussion_r513830724 in the comment? Helps to clarify these points 😃

core/src/consensus.rs
```diff
@@ -864,16 +887,43 @@ impl Tower {
...
                     last_voted_slot,
-                    slot_history.oldest(),
-                ));
+                );
+                error!("{}", message);
+                datapoint_error!("tower_error", ("error", message, String));
+                // pass-through adjust_lockouts_with_slot_history just to sanitize without
+                // any adjustment using a synthesized SlotHistory.
+                let mut warped_slot_history = (*slot_history).clone();
+                warped_slot_history.add(tower_root);
```
carllin, line 917: Follow-up to this: https://github.com/solana-labs/solana/pull/12671/files#r513942038, for maintainability, could we add a 
comment on why this `warped_slot_history.add(tower_root); ` is necessary and sufficient for correctness (I think to find an anchored slot)?

core/src/consensus.rs
```diff
@@ -515,6 +515,60 @@ impl Tower {
             .map(|last_voted_slot| {
                 let root = self.root();
                 let empty_ancestors = HashSet::default();
+                let empty_ancestors_due_to_minor_unsynced_ledger = || {
```
carllin, line 518: nice, I like this new structure! I think if possible we should add a simple diagram/explanation to the comment to help clarify these points/add proof to the correctness. 

Maybe we can import from here
1) Safety: https://github.com/solana-labs/solana/pull/10718#discussion_r474468319

2) Corner case consideration: https://github.com/solana-labs/solana/pull/10718/files#r478740859

carllin (dismissed):
Approved with some nits!

core/src/consensus.rs
```diff
@@ -864,16 +887,43 @@ impl Tower {
...
+                    return Err(TowerError::TooOldTower(
+                        last_voted_slot,
+                        slot_history.oldest(),
+                    ));
+                }
+                self.adjust_lockouts_with_slot_history(slot_history)?;
+                self.initialize_root(replayed_root);
+            } else {
+                // This should never occur under normal operation.
+                // If this validator has leader slots while voting suspended this way,
```
carllin, line 905: Same as comment: https://github.com/solana-labs/solana/pull/12671/files#r514554545, validator doesn't necessarily have to be the leader

core/src/consensus.rs
ryoqun: added a comment: 2aba6a13eefab135acf6337d6d030b6918cf7293

core/src/consensus.rs
ryoqun: > the validator doesn't even have to be leader/recreating a block for this to happen right?

You're correct. it seems validator doesn't have to be leader. I got false impression of leader requirement. Because, unstaked validators seldom reached to this closure while testing locally and manually.

Comment updated: 677887aafabc819ebbebd7959239648af8ce4e78

Now, I improved test_future_tower to test for leader and non-leader to validate this: 7d60a0b89b13c8e8af0e4f9f9baf22efd6bc0c10

core/src/consensus.rs
ryoqun: yeah, https://github.com/solana-labs/solana/pull/12671/files#r514958798

Comment updated: 677887a

ryoqun:
@carllin FYI, I'm merging this albeit there is still desirable comment additions and adjustments because I believe there should be no remaining outstanding implementation change and I want to deliver this rather now. Currently, I'm rather bugy with inflation inspection/tweak, vote program change, and `solana-genesis` debugging...

I'll create another further follow-up comment-only pr later after my task queue got shallow. :) Also note that I have already 1 nano bug fix pr and logging cleaning pr relating to tower. So, there are 3 remaining PRs I'd like to create.

core/src/consensus.rs
ryoqun: Yeah, I'll do this later: https://github.com/solana-labs/solana/pull/12671#issuecomment-719473558

core/src/consensus.rs
ryoqun: > Side note, could we include a simple diagram like the one in [#12671 (comment)](https://github.com/solana-labs/solana/pull/12671#discussion_r513830724) in the comment? Helps to clarify these points smiley

I'll do this later: https://github.com/solana-labs/solana/pull/12671#issuecomment-719473558

ryoqun:
@carllin Anyway, thanks so much for careful review!

core/src/consensus.rs
```diff
@@ -515,6 +515,59 @@ impl Tower {
...
+                    // When we're in this clause, it basically means validator is badly running
+                    // with a future tower while replaying past slots, especially problematic is
+                    // last_voted_slot.
+                    // So, don't re-vote on it by returning pseudo FailedSwitchThreshold, otherwise
+                    // there would be slashing because of double vote on one of last_vote_ancestors.
+                    // (Well, needless to say, re-creating the duplicate block must be handled properly
+                    // at the banking stage: https://github.com/solana-labs/solana/issues/8232)
+                    //
+                    // To be specific, the replay stage is tricked into a false perception where
+                    // last_vote_ancestors is AVAILABLE for descendant-of-`switch_slot`,  stale, and
```
ryoqun, line 565: improve wording here

core/src/consensus.rs
```diff
@@ -864,16 +886,59 @@ impl Tower {
         );
 
         if let Some(last_voted_slot) = self.last_voted_slot() {
-            if slot_history.check(last_voted_slot) == Check::TooOld {
-                // We could try hard to anchor with other older votes, but opt to simplify the
-                // following logic
-                return Err(TowerError::TooOldTower(
+            if tower_root <= replayed_root {
+                // Normally, we goes into this clause with possible help of
```
ryoqun, line 890: `possibly with`.

PR #12674, opened 2020-10-05
filters out inactive nodes from push options

behzadnouri:
#### Problem
https://github.com/solana-labs/solana/pull/12620
patched the DDOS issue with nodes which go offline:
https://github.com/solana-labs/solana/issues/12409

However, offline nodes still see (much lesser) traffic spike, likely
because no origins are pruned from their bloom filter in active set:
https://github.com/solana-labs/solana/blob/aaf3790d8/core/src/crds_gossip_push.rs#L276-L286
and so multiple nodes push redundant duplicate messages to them
simultaneously:
https://github.com/solana-labs/solana/blob/aaf3790d8/core/src/crds_gossip_push.rs#L254-L255

#### Summary of Changes
This commit will filter out inactive peers from potential push targets
entirely. To mitigate eclipse attacks, staked nodes are retried
periodically.

### DIFF
--- a/core/src/crds_gossip_pull.rs
+++ b/core/src/crds_gossip_pull.rs
@@ -281,8 +281,7 @@ impl CrdsGossipPull {
     ) {
         requests.into_iter().for_each(|(caller, _)| {
             let key = caller.label().pubkey();
-            let old = crds.insert(caller, now);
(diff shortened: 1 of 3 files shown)


sakridge (approved):
lgtm

carllin (approved):

Issue #12209, opened 2020-09-14
Include program logs in the explorer


mvines:
Transaction instruction logs are not easy to view, and not at all from the explorer.  This would be a great debug tool for program developers.

PR #12678, opened 2020-10-05
Store program logs in blockstore / bigtable (TransactionWithStatusMeta)

oJshua:
#### Problem
Transaction instruction logs are not easy to view, and not at all from the explorer. This would be a great debug tool for program developers. 

#### Summary of Changes
- Collects logging from transaction batches
- Stores log messages in TransactionWithStatusMeta
- Stores as Option for backwards deserialization compatibility 

Fixes https://github.com/solana-labs/solana/issues/12209

### DIFF
--- a/ledger/src/blockstore_processor.rs
+++ b/ledger/src/blockstore_processor.rs
@@ -109,6 +109,7 @@ fn execute_batch(
             MAX_PROCESSING_AGE,
             transaction_status_sender.is_some(),
             transaction_status_sender.is_some(),
+            transaction_status_sender.is_some(),
         );
 
     bank_utils::find_and_send_votes(batch.transactions(), &tx_results, replay_vote_sender);
@@ -2899,6 +2900,7 @@ pub mod tests {
         ) = batch.bank().load_execute_and_commit_transactions(
             &batch,
             MAX_PROCESSING_AGE,
+            false,
             false,
             false,
         );
--- a/ledger/src/blockstore_processor.rs
+++ b/ledger/src/blockstore_processor.rs
@@ -16,8 +16,8 @@ use solana_metrics::{datapoint_error, inc_new_counter_debug};
 use solana_rayon_threadlimit::get_thread_count;
 use solana_runtime::{
     bank::{
-        Bank, Builtins, InnerInstructionsList, TransactionBalancesSet, TransactionProcessResult,
-        TransactionResults,
+        Bank, Builtins, InnerInstructionsList, TransactionBalancesSet, TransactionLogMessages,
+        TransactionProcessResult, TransactionResults,
     },
     bank_forks::BankForks,
     bank_utils,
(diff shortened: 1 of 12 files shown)


oJshua:
Fixing that one test, stand by.

jstarry:
no worries, looks like there is still one more test failure

storage-bigtable/src/bigtable.rs
```diff
@@ -654,6 +654,7 @@ mod tests {
                 pre_balances: vec![43, 0, 1],
                 post_balances: vec![0, 42, 1],
                 inner_instructions: Some(vec![]),
+                log_messages: Some(vec![]),
             }),
         };
         let block = ConfirmedBlock {
```
CriesofCarrots, line 660: Looks like you need to set `log_messages: None` for the expected bincode block ~line 705 below

oJshua: Ah, thank you!

ledger/src/blockstore_processor.rs
```diff
@@ -103,7 +103,7 @@ fn execute_batch(
     transaction_status_sender: Option<TransactionStatusSender>,
     replay_vote_sender: Option<&ReplayVoteSender>,
 ) -> Result<()> {
-    let (tx_results, balances, inner_instructions) =
+    let (tx_results, balances, inner_instructions, transaction_logs) =
```
jstarry, line 106: Only API nodes really need to track transaction logs, validators don't need to. How about disabling log collection when `transaction_status_sender.is_none()`?

runtime/src/bank.rs
```diff
@@ -2198,6 +2204,12 @@ impl Bank {
                         self.feature_set.clone(),
                     );
 
+                    transaction_logs.push(
+                        Rc::try_unwrap(log_collector.unwrap())
```
jstarry, line 2208: This `unwrap()` is safe right now but it's generally a good idea to avoid it when possible

runtime/src/bank.rs
```diff
@@ -2198,6 +2204,12 @@ impl Bank {
                         self.feature_set.clone(),
                     );
 
+                    transaction_logs.push(
```
jstarry, line 2207: I think now is a good time to truncate long log messages now that we will be storing them in rocksdb and bigtable. Maybe 100KB? (it's 1.3KB to log a single public key right now)

runtime/src/bank.rs
oJshua: Yeah, I'm aware of the dangers and it felt sketchy using it here. Do you have an alternative for this case?

oJshua: Maybe unwrap_or_default. I'll test this out.

runtime/src/bank.rs
```diff
@@ -2211,11 +2224,24 @@ impl Bank {
...
+                                .into();
+
+                        let line_count = log_messages.len();
+
+                        Self::truncate_log_messages(
+                            &mut log_messages,
+                            TRANSACTION_LOG_MESSAGES_BYTES_LIMIT,
                         );
+
+                        if log_messages.len() != line_count {
```
CriesofCarrots, line 2239: It would be nice if this were inside `truncate_log_messages` or similar helper so that you could unit-test this branch.

CriesofCarrots (approved):
Looks good! One suggestion to consider.

runtime/src/bank.rs
oJshua: I agree. I'll refactor that bit.

CriesofCarrots (approved):

PR #12679, opened 2020-10-05
Local program allocator

jackcmay:
#### Problem

Programs use a syscall based rust heap.  Doing so is less efficient than implementing a local heap over the heap memory region.  Doing it local also give developers a good reference for how they can implement the heap to their own needs

#### Summary of Changes

- Provide a program local heap by default (don't use the syscall one)
- Provide a feature that allows developers to turn off the default heap
- Provide an example that shows a program implementing their own heap

Fixes #


jackcmay:
@mvines I don't think this requires any gating, agreed?

sdk/src/entrypoint.rs
```diff
@@ -17,24 +19,39 @@ pub type ProgramResult = ResultGeneric<(), ProgramError>;
...
 pub type ProcessInstruction =
     fn(program_id: &Pubkey, accounts: &[AccountInfo], instruction_data: &[u8]) -> ProgramResult;
 
 /// Programs indicate success with a return value of 0
 pub const SUCCESS: u64 = 0;
 
-/// Declare the entry point of the program.
+/// Start address of the memory region used for program heap.
+pub const HEAP_START_ADDRESS: usize = 0x300000000;
+/// Length of the heap memory regionused for program heap.
```
mvines, line 32:
```suggestion
/// Length of the heap memory region used for program heap.
```

mvines: Why 32k?

jackcmay:
@mvines, I'm going to backport this change to v.13 so that folks using the v1.3 sdk can take advantage of the local heap.  Any objections?

mvines:
sgtm

PR #12684, opened 2020-10-06
Remove unnecessary skip-no-mangle feature from sdk

jackcmay:
#### Problem

The sdk feature `skip-no-mangle` isn't actually used in the sdk, just by the callers of the entrypoint macro.

#### Summary of Changes

Remove the feature from the sdk

Fixes #

### DIFF
--- a/sdk/Cargo.toml
+++ b/sdk/Cargo.toml
@@ -11,8 +11,6 @@ edition = "2018"
 [features]
 # On-chain program specific dependencies
 program = []
-# Program elements to optionally skip
-skip-no-mangle = []
 # Dependencies that are not compatible or needed for on-chain programs
 default = [
     "assert_matches",


jackcmay:
@CriesofCarrots I don't think we need to publish any SPL crates and we keep this change to solana v1.4 and newer.  Any programs that move to v1.4 will have to stop depending on `solana-sdk/skip-no-mangle`.  The SPL programs remove this dependency here: https://github.com/solana-labs/solana-program-library/pull/567

Am I missing something?

CriesofCarrots:
The thing that skip-no-mangle was accomplishing is now being done by the `program` feature in the spl programs: https://github.com/solana-labs/solana-program-library/blob/8131bec332741862644dff6c98394bfe240d01e8/token/program/src/entrypoint.rs#L3

So I think there's no reason to keep these tags:
https://github.com/solana-labs/solana/blob/ae0fd3043a7a4aef23adf52e63caa9866a351680/sdk/src/entrypoint.rs#L39
https://github.com/solana-labs/solana/blob/ae0fd3043a7a4aef23adf52e63caa9866a351680/sdk/src/entrypoint_deprecated.rs#L40

You're right, no reason to publish new spl crates if we aren't going to backport this to v1.3

CriesofCarrots:
Why not remove those tags?

jackcmay:
@CriesofCarrots Oh, I missed that point, so you think we don't need them at all anymore.  Did you try removing then entirely in your double sdk test?

CriesofCarrots:
@jackcmay , yep, I removed them entirely.

jackcmay:
@CriesofCarrots https://github.com/solana-labs/solana/pull/12696

jackcmay:
@CriesofCarrots and https://github.com/solana-labs/solana-program-library/pull/570

Issue #12670, opened 2020-10-05
vote hash mismatch in testnet with high tps


ryoqun:
#### Problem

STR:

- `./net/gce.sh create -n 4 -c 1 -p <prefix>`
- `./net/net.sh start`
- `gcloud beta compute ssh --zone "us-west1-b" <validator node> -- sudo pkill -f solana-validator`

I haven't investigated well. but restoring snapshotted root bank succeeds. But immediately with the next of the bank, the bank hash starts to diverge like this:

```
[2020-10-05T09:31:31.679508257Z INFO  solana_runtime::bank] bank frozen: 801 hash: DgZKMR56GB2TyxoB2BVzHujMfWgShGhswAqpDXTa442 accounts_delta: 2GxJuagoNVHfrcbGQ1Qo9cTxAwcDfQasvM5x4QbsAPTi signature_count: 16119 last_blockhash: 34afVht5TSaJbZNM7ax1QhBeCL4KdY6Jh8CQ1p7q7wjT capitalization: 1000000808342242059
[2020-10-05T09:31:31.679546395Z INFO  solana_runtime::bank] accounts hash slot: 801 stats: BankHashStats { num_updated_accounts: 32248, num_removed_accounts: 0, num_lamports_stored: 36106811887435, total_data_len: 176296, num_executable_accounts: 0 }
```

```
[2020-10-05T09:33:37.253939880Z INFO  solana_runtime::bank] bank frozen: 801 hash: EQNias25stAcHYjEPcQW9dQPPJmLW8t9yvfgGo8kwnvN accounts_delta: 7p6svSrwTjgvPitEF5FS2aSFGeHatnyn7vyMhMf6jkbW signature_count: 16119 last_blockhash: 34afVht5TSaJbZNM7ax1QhBeCL4KdY6Jh8CQ1p7q7wjT capitalization: 1000000808342242059
[2020-10-05T09:33:37.253982797Z INFO  solana_runtime::bank] accounts hash slot: 801 stats: BankHashStats { num_updated_accounts: 32248, num_removed_accounts: 0, num_lamports_stored: 36106811887435, total_data_len: 176296, num_executable_accounts: 0 }
```

#### Proposed Solution

debug and fix.

also we should stop the validator if our computed bank hash is different than the rooted supermajority of the cluster.

ryoqun:
[...]

PR #11927, opened 2020-08-31
Check bank capitalization

ryoqun:
#### Problem

- No check for bank capitalization.

- Also, the current offline bank capitalization check doesn't catch overflows.

#### Summary of Changes

On accounts hashing, load the total accounts capitalization and compare with bank.capitalization.

Fixes #

#### context

reboot of https://github.com/solana-labs/solana/pull/10937

PR #12690, opened 2020-10-06
Detect vote hash mismatch even at run-sanity.sh

ryoqun:
#### Problem

https://github.com/solana-labs/solana/issues/12670#issuecomment-704134218

#### Summary of Changes

note that this test expected to fail.

### DIFF
--- a/ci/run-sanity.sh
+++ b/ci/run-sanity.sh
@@ -29,7 +29,8 @@ curl -X POST -H 'Content-Type: application/json' -d '{"jsonrpc":"2.0","id":1, "m
 
 wait $pid
 
-$solana_ledger_tool create-snapshot --ledger config/ledger "$snapshot_slot" config/snapshot-ledger
+$solana_ledger_tool create-snapshot --ledger config/ledger "$snapshot_slot" config/snapshot-ledger 2>&1 \
+  | tee -a "ledger-tool.log"
 cp config/ledger/genesis.tar.bz2 config/snapshot-ledger
-$solana_ledger_tool verify --ledger config/snapshot-ledger 2>&1 | tee "ledger-verify.log"
-! grep "failed to match hash" "ledger-verify.log"
+$solana_ledger_tool verify --ledger config/snapshot-ledger 2>&1 | tee -a "ledger-tool.log"
+! grep "failed to match hash" "ledger-tool.log"
--- a/ci/run-sanity.sh
+++ b/ci/run-sanity.sh
@@ -31,5 +31,5 @@ wait $pid
 
 $solana_ledger_tool create-snapshot --ledger config/ledger "$snapshot_slot" config/snapshot-ledger
 cp config/ledger/genesis.tar.bz2 config/snapshot-ledger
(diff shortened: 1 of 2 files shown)


sakridge:
Would it better to have a flag to panic on this case?

mvines:
> Would it better to have a flag to panic on this case?

Flag would be cool.   Plumbing that flag from `solana-validator` into the vote state seems really gnarly though, I don't know how we'd do that nicely

sakridge:
> 
> 
> > Would it better to have a flag to panic on this case?
> 
> Flag would be cool. Plumbing that flag from `solana-validator` into the vote state seems really gnarly though, I don't know how we'd do that nicely

That's true. Plumbing would be a pain. Also security reviews don't like these kinds of flags to turn off checks.

On the other hand, these kind of log checks are kind of brittle and have a habit of being disabled without anyone noticing. Also seems like more of a pain to enable in more test situations. Generally we would always want to panic in this situation unless it's a specifically malicious test case to insert duplicate conflicting blocks into the system.

ryoqun:
> > > Would it better to have a flag to panic on this case?
> > 
> > 
> > Flag would be cool. Plumbing that flag from `solana-validator` into the vote state seems really gnarly though, I don't know how we'd do that nicely
> 
> That's true. Plumbing would be a pain. Also security reviews don't like these kinds of flags to turn off checks.
> 
> On the other hand, these kind of log checks are kind of brittle and have a habit of being disabled without anyone noticing. Also seems like more of a pain to enable in more test situations. Generally we would always want to panic in this situation unless it's a specifically malicious test case to insert duplicate conflicting blocks into the system.

Yeah, I don't like the current state of this. After some thought, I think I'll just add a validator kill-switch if the validator's voted slot hash is mismatched to the supermajority rest of cluster.

This will be like combination of #11927 and #11727, which is done at the snapshot interval.

ryoqun:
@mvines @sakridge (no rush) please take a look at the more proper solution for this in my previous comment.

sakridge:
@ryoqun is there something wrong with this? Can we prioritize enabling this checking?

ryoqun:
@sakridge There is nothing wrong with this. it's just that I couldn't get my hands on it. Currently, I'm focusing on various inflation and staking bugs. I talked about this a bit with @carllin, we can implement this with VoteTracker.

sakridge:
> 
> 
> @sakridge There is nothing wrong with this. it's just that I couldn't get my hands on it. Currently, I'm focusing on various inflation and staking bugs. I talked about this a bit with @carllin, we can implement this with VoteTracker.

Is the builtins part of this change needed for the vote hash mismatch?

ryoqun:
> > @sakridge There is nothing wrong with this. it's just that I couldn't get my hands on it. Currently, I'm focusing on various inflation and staking bugs. I talked about this a bit with @carllin, we can implement this with VoteTracker.
> 
> Is the builtins part of this change needed for the vote hash mismatch?

@sakridge oops, sorry. these was rebase leftover; fixed.

PR #12692, opened 2020-10-06
fix: use Node.js https agent when endpoint uses https

jstarry:
#### Problem
Cannot use `http.Agent` for https fetch requests in Node.js

```
    TypeError: Protocol "https:" not supported. Expected "http:"TypeError [ERR_INVALID_PROTOCOL]: Protocol "https:" not supported. Expected "http:"
```

#### Summary of Changes
- Check the endpoint protocol to determine which agent to use
- Add https test against devnet api to prevent this in the future :/
- Silence rollup dependency warnings

Fixes #

### DIFF
--- a/web3.js/rollup.config.js
+++ b/web3.js/rollup.config.js
@@ -98,6 +98,8 @@ function generateConfig(configType) {
         'bs58',
         'buffer-layout',
         'crypto-hash',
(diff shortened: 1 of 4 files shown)


mvines (approved):
lgtm

PR #12695, opened 2020-10-06
Web3.js/validation performance

### LINKED ISSUE #12664: web3: JSON validation is very slow for large responses
#### Problem
JSON validation in the web3 sdk is very slow for large responses

For example:
- `getProgramAccounts` is slow when many accounts are returned
- Had to disable full validation for `getLeaderSchedule` due to slowness: https://github.com/solana-labs/solana/pull/12049

#### Proposed Solution
Seems like `superstruct.array(..)` is the culprit so I think it's worth investigating its implementation. However, this could just be an overall issue with `superstruct` and so it might be worth investigating different libraries since superstruct development hasn't been very active

bartosz-lipinski:
#### Problem

JSON validation in the web3 sdk is very slow for large responses

For example:

getProgramAccounts is slow when many accounts are returned
Had to disable full validation for getLeaderSchedule due to slowness: #12049

#### Summary of Changes

- [x] Move schema definition from javascript code to schema format in a js file 
- [x] Create a code generator that reads schema and generates validators using `ajv` and `ajv-pack`
- [x] Create typescript definition for JSON schema
- [x] Add build script that compiles schema into validation functions that can be used at runtime
- [x] Migrate all references of superstruct to JSON schema

All these changes will lead to Improved performance of validation code (between 4x-20x improvement depending on the payload).

Future improvements (Outside of the scope of this PR):
* Remove JsonRPC types as a part of web3.js migration to typescript
* Use extracted schema to deploy Swagger API page:
    * Enables client generator
    * Enables online API tester
* migrate schema to https://api.mainnet-beta.solana.com/schema.json
* generate json schema directly from rust code

Fixes #12664


bartosz-lipinski:
I verified performance using `getProgramAccounts` function which returns 9.5MB for the stake program. It was validated in the browser under 0.1ms.

jstarry:
@bartosz-lipinski can you please fix the CI failure and conflicts?

PR #12696, opened 2020-10-06
Remove skip-no-mangle entirely

jackcmay:
#### Problem

Programs now use other features to decide if the program's entrypoint should be pulled in or not.

#### Summary of Changes

Remove this feature usage from the `entrypoint` macro

Fixes #


jackcmay:
@CriesofCarrots Peg'd the token and memo versions.  Next time we upgrade either the feature `skip-no-mangle` will need to be removed

CriesofCarrots (approved):

PR #10466, opened 2020-06-09
runtime: fix possible deadlock in bank

BurtonQin:
#### Problem
There are two kinds of possible deadlock bugs:

1. locks in conflicting order:
e.g.

|compare_bank|freeze|
|---|---|
|stakes.read()||
||hash.write()|
||[collect_return_eagerly](https://github.com/solana-labs/solana/blob/9d07a23c46f30894d5a0b5e55d3b266664a7c84f/runtime/src/bank.rs#L794)|
||[collect_rent_in_partition](https://github.com/solana-labs/solana/blob/9d07a23c46f30894d5a0b5e55d3b266664a7c84f/runtime/src/bank.rs#L1671)|
||[store_account](https://github.com/solana-labs/solana/blob/9d07a23c46f30894d5a0b5e55d3b266664a7c84f/runtime/src/bank.rs#L1720)|
||[stakes.write()](https://github.com/solana-labs/solana/blob/9d07a23c46f30894d5a0b5e55d3b266664a7c84f/runtime/src/bank.rs#L2026)//deadlock!|
|hash.read()//deadlock!||

The fix is to swap the order of `stakes` and `hash` in `compare_bank()`.

 2. double read lock if comparing with itself
When compare_bank() compares itself, there are four cases where the first read lock is not released before the second lock. A deadlock may happen when the two read locks are interleaved by a write lock from another thread. The reason is that the priority policy of `std::sync::RwLock` is [dependent on the underlying operating system's implementation](https://doc.rust-lang.org/std/sync/struct.RwLock.html). AFAIK, [Windows and macOS use fair policy](https://www.reddit.com/r/rust/comments/f4zldz/i_audited_3_different_implementation_of_async/) which leads to this kind of deadlock.

The fix is to add a shortcut by checking the ptr equality of `self` and `dbank` first.

#### Summary of Changes
1. Swap the order of `stakes` and `hash` in `compare_bank()`.
2. Add a shortcut by checking the ptr equality of `self` and `dbank` in `compare_bank()`.

PR #12697, opened 2020-10-06
Fix fee mismatch on snapshot deserialize

carllin:
#### Problem
https://github.com/solana-labs/solana/issues/12670#issuecomment-704064936

#### Summary of Changes
As I see it, two options:

1) Proper fix: Serialize `lamports_per_signature` field and add gating to roll this out
2) Just take the `lamports_per_signature` value from the fee calculator on deserialize

Fixes #

### DIFF
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -878,7 +878,11 @@ impl Bank {
         );
         assert_eq!(bank.epoch_schedule, genesis_config.epoch_schedule);
         assert_eq!(bank.epoch, bank.epoch_schedule.get_epoch(bank.slot));
-
+        bank.fee_rate_governor.lamports_per_signature = bank.fee_calculator.lamports_per_signature;
+        assert_eq!(
+            bank.fee_rate_governor.create_fee_calculator(),
+            bank.fee_calculator
+        );
         bank
     }
 
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -71,6 +71,7 @@ use std::{
     mem,
     ops::RangeInclusive,
     path::PathBuf,
+    ptr,
     rc::Rc,
     sync::{
         atomic::{AtomicBool, AtomicU64, Ordering::Relaxed},
@@ -477,8 +478,12 @@ pub(crate) struct BankFieldsToSerialize<'a> {
     pub(crate) is_delta: bool,
 }
 
+// Can't derive PartialEq because RwLock doesn't implement PartialEq
 impl PartialEq for Bank {
     fn eq(&self, other: &Self) -> bool {
+        if ptr::eq(self, other) {
+            return true;
+        }
         *self.blockhash_queue.read().unwrap() == *other.blockhash_queue.read().unwrap()
             && self.ancestors == other.ancestors
             && *self.hash.read().unwrap() == *other.hash.read().unwrap()
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -71,7 +71,6 @@ use std::{
     mem,
     ops::RangeInclusive,
     path::PathBuf,
-    ptr,
     rc::Rc,
     sync::{
         atomic::{AtomicBool, AtomicU64, Ordering::Relaxed},
@@ -478,18 +477,18 @@ pub(crate) struct BankFieldsToSerialize<'a> {
     pub(crate) is_delta: bool,
 }
 
-impl<'a> PartialEq for BankFieldsToSerialize<'a> {
+impl PartialEq for Bank {
     fn eq(&self, other: &Self) -> bool {
         *self.blockhash_queue.read().unwrap() == *other.blockhash_queue.read().unwrap()
             && self.ancestors == other.ancestors
-            && self.hash == other.hash
+            && *self.hash.read().unwrap() == *other.hash.read().unwrap()
             && self.parent_hash == other.parent_hash
             && self.parent_slot == other.parent_slot
             && *self.hard_forks.read().unwrap() == *other.hard_forks.read().unwrap()
-            && self.transaction_count == other.transaction_count
-            && self.tick_height == other.tick_height
-            && self.signature_count == other.signature_count
-            && self.capitalization == other.capitalization
+            && self.transaction_count.load(Relaxed) == other.transaction_count.load(Relaxed)
+            && self.tick_height.load(Relaxed) == other.tick_height.load(Relaxed)
+            && self.signature_count.load(Relaxed) == other.signature_count.load(Relaxed)
+            && self.capitalization.load(Relaxed) == other.capitalization.load(Relaxed)
             && self.max_tick_height == other.max_tick_height
             && self.hashes_per_tick == other.hashes_per_tick
             && self.ticks_per_slot == other.ticks_per_slot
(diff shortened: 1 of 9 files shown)


sdk/src/fee_calculator.rs
```diff
@@ -67,7 +67,6 @@ impl FeeCalculator {
 pub struct FeeRateGovernor {
     // The current cost of a signature  This amount may increase/decrease over time based on
     // cluster processing load.
-    #[serde(skip)]
```
carllin, line 70: this is where gating would be necessary. Alternatively we just set this value on deserialize to `bank.fee_calculator.lamports_per_signature`

mvines:
(+v1.3 label since this could also affect mainnet-beta once the pico-inflation feature is enabled)

sdk/src/fee_calculator.rs
carllin: @ryoqun after talking with @t-nelson, it seems removing the skip here directly also breaks RPC, seems like the path of least resistance is to just manually assign this field for now. Thoughts?

ryoqun: > the path of least resistance is to just manually assign this field for now. Thoughts?

@carllin yeah, I think this solution is good.

ryoqun:
> Proper fix: Serialize `lamports_per_signature` field and add gating to roll this out

@carllin Also, note that _gating_ here would not be the pretty `solana features`, rather would be a snapshot version bump.

ryoqun:
@carllin First of all, thanks for debugging and actually fixing this inconsistency bug. :)

However, what I'm a bit wondering is that I felt that the vote mismatches were very likely to occur _only when the `solana-tps-bench` is running (i.e. under high load)._

For me, this looks like a deterministic inconsystency issue. Could this bug explain the indermisticic behavior I've observed? For one thing, my observation isn't so well formal. So I'm not conclusive. (I just sensed this way when tackling another issue (tower assertion fixes #12671)).

sdk/src/fee_calculator.rs
ryoqun: Also, when did this bug started to occur? Since beginning? ;)

carllin: hmmm yeah I would guess since it was first checked in xD

carllin:
@ryoqun, you're right in that this is a deterministic bug.

For me when following your steps to repro yesterday, I was seeing this bug repro on *restart* under high tps, I wonder if it has something to do with how the fee is calculated here: https://github.com/solana-labs/solana/blob/master/sdk/src/fee_calculator.rs#L138, which is derived from `latest_signatures_per_slot`, which needs to hit some threshold before it changes.

ryoqun:
@carllin this is LGTM for me even if this is still a draft. Is there any remaining work like write a unit-test? I think this is a bit hard.
Or you might be able to write one by hooking up a new path-through call flow with `new_from_fields` & `get_fields_to_serialize` and `impl From<BankFieldsToDeserialize> for BankFieldsToSerialize`

carllin:
@ryoqun I modified the existing `test_bank_serialize_newer` to cover this case, what do you think?

runtime/src/bank.rs
```diff
@@ -3590,51 +3630,6 @@ impl Bank {
         }
     }
 
-    pub fn compare_bank(&self, dbank: &Bank) {
-        if ptr::eq(self, dbank) {
-            return;
-        }
```
ryoqun, line 3596: maybe this should be kept to deadlock free? #10466 

Yeah, we should have added an comment...

runtime/src/bank.rs
```diff
@@ -477,6 +477,42 @@ pub(crate) struct BankFieldsToSerialize<'a> {
...
+            && self.collector_fees.load(Relaxed) == other.collector_fees.load(Relaxed)
+            && self.fee_calculator == other.fee_calculator
+            && self.fee_rate_governor == other.fee_rate_governor
+            && self.collected_rent.load(Relaxed) == other.collected_rent.load(Relaxed)
+            && self.rent_collector == other.rent_collector
+            && self.epoch_schedule == other.epoch_schedule
+            && *self.inflation.read().unwrap() == *other.inflation.read().unwrap()
+            && *self.stakes.read().unwrap() == *other.stakes.read().unwrap()
+            && self.epoch_stakes == other.epoch_stakes
+            && self.is_delta.load(Relaxed) == other.is_delta.load(Relaxed)
```
ryoqun, line 512: I wonder how we can maintain this for the addition of newer fields...

runtime/src/bank.rs
```diff
@@ -477,6 +477,42 @@ pub(crate) struct BankFieldsToSerialize<'a> {
     pub(crate) is_delta: bool,
 }
 
+impl PartialEq for Bank {
```
ryoqun, line 480: By the way, why can't we just `derive` `PartialEq`? A comment might be desirable to explain the limitation.

runtime/src/serde_snapshot/tests.rs
```diff
@@ -219,7 +219,7 @@ fn test_bank_serialize_style(serde_style: SerdeStyle) {
     assert_eq!(dbank.get_balance(&key1.pubkey()), 0);
     assert_eq!(dbank.get_balance(&key2.pubkey()), 10);
     assert_eq!(dbank.get_balance(&key3.pubkey()), 0);
-    bank2.compare_bank(&dbank);
+    assert!(bank2 == dbank);
```
ryoqun, line 222: nits: how about `assert_eq!`?

runtime/src/serde_snapshot/tests.rs
```diff
@@ -219,7 +219,7 @@ fn test_bank_serialize_style(serde_style: SerdeStyle) {
     assert_eq!(dbank.get_balance(&key1.pubkey()), 0);
     assert_eq!(dbank.get_balance(&key2.pubkey()), 10);
     assert_eq!(dbank.get_balance(&key3.pubkey()), 0);
-    bank2.compare_bank(&dbank);
```
ryoqun, line 222: :tada: 

(I wonder why this was like this not defining `==`.)..

ryoqun:
@carllin reviewed. :)

ryoqun:
> @ryoqun, looks like the `compare_bank()` function should have covered this case but was missing some cases. I modified the existing `test_bank_serialize_newer` to cover this case, what do you think?

Really thanks for adding this. I'm very happy knowing that we're now covered under tests. :)

runtime/src/bank.rs
carllin: ah yeah good catch, thanks!

runtime/src/bank.rs
carllin: hehe so the issue was RwLock doesn't implement PartialEq, and we can't implement it for RwLock because it doesn't belong to this crate.

runtime/src/serde_snapshot/tests.rs
carllin: The issue I ran into with `assert_eq!` was that it requires `Bank` implement `Debug`, which means all the program stack `CowCachedExecutors -> Executor -> InvokeContext` have to all implement Debug , and I didn't want to plumb all of that through 😩

ryoqun (approved):
LGTM!

runtime/src/serde_snapshot/tests.rs
ryoqun: @carllin FYI, I've took the initiative here after stressed with not being able to freely `dbg!()` bunch of bank fields: https://github.com/solana-labs/solana/pull/13017/files#diff-b5dead0a38e5f63447595bbeda1428d3aa851d33954343bb1bc38b4b00add5fdR159

PR #12709, opened 2020-10-07
Inflation cli fix and parameter update

ericlwilliams:
#### Problem
`solana inflation` printed fractional Foundation allocation rather than %.
Initial inflation rate governor default didn't match current design (15% rather than 8%)

#### Summary of Changes
Fixed & updated

Fixes #

### DIFF
--- a/cli/src/inflation.rs
+++ b/cli/src/inflation.rs
@@ -57,7 +57,7 @@ pub fn process_inflation_subcommand(
         println!("Rate reduction per year: {:>5.2}%", governor.taper * 100.);
     }
     if governor.foundation_term > 0. {
-        println!("Foundation percentage:   {:>5.2}%", governor.foundation);
+        println!("Foundation percentage:   {:>5.2}%", governor.foundation * 100.);
         println!(
             "Foundation term:         {:.1} years",
             governor.foundation_term
--- a/cli/src/inflation.rs
+++ b/cli/src/inflation.rs
@@ -57,7 +57,7 @@ pub fn process_inflation_subcommand(
         println!("Rate reduction per year: {:>5.2}%", governor.taper * 100.);
     }
     if governor.foundation_term > 0. {
-        println!("Foundation percentage:   {:>5.2}%", governor.foundation * 100.);
+        println!("Foundation percentage:   {:>5.2}%", governor.foundation);
         println!(
             "Foundation term:         {:.1} years",
             governor.foundation_term
--- a/sdk/src/inflation.rs
+++ b/sdk/src/inflation.rs
@@ -22,7 +22,7 @@ pub struct Inflation {
     __unused: f64,
 }
 
-const DEFAULT_INITIAL: f64 = 0.15;
+const DEFAULT_INITIAL: f64 = 0.08;
 const DEFAULT_TERMINAL: f64 = 0.015;
 const DEFAULT_TAPER: f64 = 0.15;
 const DEFAULT_FOUNDATION: f64 = 0.05;


cli/src/inflation.rs
```diff
@@ -57,7 +57,7 @@ pub fn process_inflation_subcommand(
         println!("Rate reduction per year: {:>5.2}%", governor.taper * 100.);
     }
     if governor.foundation_term > 0. {
-        println!("Foundation percentage:   {:>5.2}%", governor.foundation);
+        println!("Foundation percentage:   {:>5.2}%", governor.foundation * 100.);
```
mvines, line 60: This would produce a value that's not accurate, as implemented.   ` governor.foundation` is currently the percentage of the total inflation amount that goes to the foundation.

sdk/src/inflation.rs
```diff
@@ -22,7 +22,7 @@ pub struct Inflation {
     __unused: f64,
 }
 
-const DEFAULT_INITIAL: f64 = 0.15;
+const DEFAULT_INITIAL: f64 = 0.08;
```
mvines, line 25: Any point in actually changing this?   It would only affect new development clusters, and have no affect for devnet/testnet/mainnet-beta.   That being said, no objections to the change itself.

ericlwilliams: Just to keep alignment with design but I'm fine with waiting if it would lead to confusion now. I defer to you

mvines: sg, thanks.  Changing it wfm then

cli/src/inflation.rs
mvines: I suspect that might not have been the intention though :)

ericlwilliams: From what I can tell, currently governor.foundation is stored as the _fraction_ of total inflation that goes to the Foundation
https://github.com/solana-labs/solana/blob/b5faa11f736fd836d4afdee67d827e86344b06e8/sdk/src/inflation.rs#L28

And, while it is hard to tell with pico-inflation turned on (not enough sig figs in `solana inflation`), the `testnet` example here: https://github.com/solana-labs/solana/pull/12623 suggests that `solana inflation` is printing out the _fraction_ as a percentage.  This fix was just to convert the fraction to percentage.

ericlwilliams: > I suspect that might not have been the intention though :)

It is. 5% (as currently designed) of the inflation (e.g. 8%) should be allocated to Foundation, not 5% of total supply.  
Whether this distinction is clear enough in the cli is a different question :)

mvines: Right now `governor.foundation`% of inflation is allocated to Foundation.  At year 0 with 15%, that means 0.0075% goes to the Foundation.   Basically the code in sdk/src/inflation.rs is not calculating the Foundation % correctly at all but what the cli currently outputs is the correct amount.  100x what the cli outputs doesn't fix the real bug.

Practically foundation is at 0% for now, and there's no plumbing to direct a foundation percentage over 0% anywhere.  So if/when we want the foundation to receive more than 0 there's a couple code changes that need to happen, including fixing the bug here

ericlwilliams: I see, so there is a factor of 100 in there somewhere.  In that case, I'll just close this PR and we can revisit all of this later.

mvines: Landing your change to DEFAULT_INITIAL sgtm still though, if you want to get that in

ericlwilliams: okay, I removed this factor of 100, kept the update to DEFAULT_INITIAL with https://github.com/solana-labs/solana/pull/12709/commits/c3c15252bd0010b93f3debb069797d1e31e2e187

PR #12712, opened 2020-10-07
Add bank-timestamp design proposal

CriesofCarrots:
#### Problem
`Bank::unix_timestamps()` are drifting further and further from real-world time.

#### Summary of Changes
Propose a method for correcting this drift using the validator-timestamp oracle

### DIFF
--- a/docs/src/proposals/bank-timestamp-correction.md
+++ b/docs/src/proposals/bank-timestamp-correction.md
@@ -1,0 +1,64 @@
+---
+title: Bank Timestamp Drift Correction
+---
+
+## Problem
+
+Currently, each Bank estimates its timestamp as an offset from genesis using a
+theoretical slots-per-second constant. This timestamp gets inputted into the
+Clock sysvar and is used to assess time-based stake account lockups. However the
+actual slots processed per second on the public clusters has been much slower
+than the goal, meaning lockups will not actually expire on (or anytime near) the
+date the lockup is set to expire.
+
+## Proposed Solution
+
+The [validator timestamp oracle](../implemented-proposals/validator-timestamp-oracle.md)
+provides real-world time information, and validators now timestamp every vote.
+As a result vote-account state can be used to get the stake-weighted timestamp
+of any slot and correct the Bank/clock sysvar drift.
+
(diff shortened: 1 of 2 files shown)


docs/src/proposals/bank-timestamp-correction.md
```diff
@@ -0,0 +1,64 @@
...
+The map of slots to sample needs to be generated deterministically to maintain
+consensus, so it does still present an opportunity for exploit. One option would
+be to use only some of the populated samples to determine the corrected
+timestamp; choosing which ones at the end of the epoch based on the last
+blockhash. But a malicious actor could still affect the corrected timestamp by
+manipulating timestamps for all of the sample slots, or indeed all of the slots
+in the epoch. So there doesn't appear to be much benefit to the additional
+complexity.
+
+### Correction Math
```
mvines, line 61: I think the major part that we're missing here is how the drift is actually corrected:
1. If the bank clock is ahead of wallclock, then at the end of the epoch the bank clock is perhaps incremented by `1` (the smallest amount larger than 0)
2. If the bank clock is behind of wallclock (like it is currently), how quickly to we warp ahead?  I feel like we probably shouldn't warp ahead too much.  In the case of a cabal of validators attempting influencing the clock, the only way to reduce their impact will be to deactivate stake from them.  This will take days/weeks.   So perhaps we only warp head by at most 1 epoch of wallclock?   But if we limit to this, will we reach close to realtime by the end of the year?

CriesofCarrots: I'll add a section on this.

1. I was thinking we wouldn't ever correct the bank clock backwards, so if the bank clock is ahead of wallclock, it isn't adjusted at all.
2. Assuming that the real average slot duration remains similar to that over the past 40M blocks, one additional epoch's-worth of correction will not be enough to get us synced by the end of the year. According to my math, we'll need to correct approximately 2.5 additional epochs of wallclock each epoch to reach real time by the end of the year.

CriesofCarrots:
Closing in lieu of a simpler approach: https://github.com/solana-labs/solana/pull/12737
Will re-open if needed

PR #12713, opened 2020-10-07
Add separate push queue to reduce push lock contention

sakridge:
#### Problem

Local push messages into gossip can have lock contention stalling replay stage for many `ms` while getting a vote into the push queue.

#### Summary of Changes

Add a separate vec for local push messages. Drain it when creating new push messages.

Fixes #

### DIFF
--- a/core/src/cluster_info.rs
+++ b/core/src/cluster_info.rs
@@ -257,6 +257,7 @@ pub struct ClusterInfo {
     id: Pubkey,
     stats: GossipStats,
     socket: UdpSocket,
+    local_message_pending_push_queue: RwLock<Vec<(CrdsValue, u64)>>,
 }
 
 impl Default for ClusterInfo {
@@ -414,6 +415,7 @@ impl ClusterInfo {
             id,
             stats: GossipStats::default(),
             socket: UdpSocket::bind("0.0.0.0:0").unwrap(),
+            local_message_pending_push_queue: RwLock::new(vec![]),
         };
         {
             let mut gossip = me.gossip.write().unwrap();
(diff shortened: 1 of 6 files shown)


sakridge:
voting

Hopefully can get rid of these voting spikes.

behzadnouri:
I think these lines have to be reversed in order to ensure pull requests are generated off an updated crds table:
https://github.com/solana-labs/solana/blob/aea979c32/core/src/cluster_info.rs#L1463-L1468

Which makes me wonder what other places are assuming crds table is up to date.

sakridge:
> I think these lines have to be reversed in order to ensure pull requests are generated off an updated crds table:
> https://github.com/solana-labs/solana/blob/aea979c32/core/src/cluster_info.rs#L1463-L1468
> 
> Which makes me wonder what other places are assuming crds table is up to date.

Well, these pending push requests are local ones that we haven't sent to anyone, so no other node would have them yet to be able to respond to our pull request with.

In general, it seems like a slightly stale table should not be a problem, it would just be updated on the next go-around of the gossip generation loop.

core/src/cluster_info.rs
```diff
@@ -1425,11 +1426,21 @@ impl ClusterInfo {
             })
             .collect()
     }
+    fn drain_push_queue(&self) -> Vec<(CrdsValue, u64)> {
+        let mut push_queue = self.pending_push_queue.write().unwrap();
+        std::mem::replace(&mut *push_queue, vec![])
```
behzadnouri, line 1431: maybe `std::mem::take` here instead

behzadnouri (dismissed):
lgtm if you have not observed any issues on a test network.

> these pending push requests are local ones that we haven't sent to anyone

maybe add this as a comment to `ClusterInfo.pending_push_queue` field definition.

> In general, it seems like a slightly stale table should not be a problem, it would just be updated on the next go-around of the gossip generation loop.

if the staleness lag is added up on each hop, then could become a significant delay when the message has to go through multiple links. (obviously not applicable here if these are only local push messages).

core/src/cluster_info.rs
sakridge: yea, take is a bit cleaner, changed

carllin (approved):

Issue #10093, opened 2020-05-17
Consider correcting Bank::slots_per_year drift


mvines:
Similarly to https://github.com/solana-labs/solana/issues/9874, inflation and rent calculations are based on a theoretical number of slots per year:
https://github.com/solana-labs/solana/blob/5e89bd88681da2f331e752ccbf670df77b600789/runtime/src/bank.rs#L939-L943

At the time of this issue, the actual slots_per_year rate appears to be about 1 week ahead of the theoretical, meaning that inflation will taper faster than expected and rent will become due sooner:
```
$ solana block-time
Block: 11525626
Date: 2020-05-17T17:07:27Z (UnixTimestamp: 1589735247)

$ solana cluster-date 
Block: 11525640
Date: 2020-05-08T23:06:38Z (UnixTimestamp: 1588979198)
```

Let's decide if this even is a problem first.

mvines:
> Let's decide if this even is a problem first.

Yep it's a problem.    At every Epoch we probably should adjust `solana cluster-date` towards `solana block-time`, but it must not go backwards.

garious:
Per conversation with @mvines, we can change `Clock::unix_timestamp` from PoH time to block time, but only update when block time increases.  We can make the transition on Mainnet Beta now because PoH time is currently behind block time.

CriesofCarrots:
Update on this: `solana cluster-date` is now corrected toward `solana block-time`, with various caveats (time cannot got backward, changes are bounded to 25% of the poh estimate since the start of the slot).

However, it is not certain how the slots-per-year calculation ought to be corrected. Presumably we want the annualized interest rate to be correct, but we need to consider how much and how often we want the rate to drift (correct). Since inflation is only just about to begin on mainnet-beta, it seems like we will want to observe and collect data for a while to make these determinations.

PR #12737, opened 2020-10-08
Correct Bank timestamp drift every slot

### LINKED ISSUE #9874: Correct bank::unix_timestamp() drift
The timestamp returned by `bank::unix_timestamp()` ends up in the clock sysvar and is used for time-based stake account lockouts.   However it's based on a theoretical slots-per-second instead of reality, so it's very likely quite inaccurate already.   This could post a problem for the month 9 stake unlocks scheduled for Jan 7, 2021.   

Steps:
1. Implement https://github.com/solana-labs/solana/issues/9847 to gain visibility into how bad the drift currently is
2. Consider recalibrating the clock every Epoch based on the vote timestamps?

Also note this comment that perhaps needs to be adjusted along the way:
https://github.com/solana-labs/solana/blob/6deaf649eff9363b506e8f62ede4e5914b436571/core/src/rpc.rs#L416-L420

CriesofCarrots:
#### Problem
The timestamp returned by `Bank::unix_timestamp()` ends up in the clock sysvar and is used to assess time-based stake account lockouts. However it's based on a theoretical slots-per-second instead of reality, so it's quite inaccurate. This will pose a problem for lockups, since the accounts will not register as lockup-free on (or anytime near) the date the lockup is set to expire.

Block times are estimated using a validator timestamp oracle; this data provides an opportunity to align the bank timestamp more closely with real-world time.

#### Summary of Changes
- Adjust `Bank::update_clock` to set the Clock sysvar timestamp to the stake-weighted timestamp estimate if it is greater than the unix_timestamp from genesis; feature-gated

This adds 50us to each `Bank::new_from_parent()`, which was not distinguishable in confirmation-time or other metrics on a testnet under moderate TPS load.

Fixes #9874 
Toward #10093 (but requires some additional handling to update Bank::slots_per_year)

### DIFF
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -1044,12 +1044,23 @@ impl Bank {
     }
 
     fn update_clock(&self) {
+        let mut unix_timestamp = self.unix_timestamp();
+        if self
+            .feature_set
+            .is_active(&feature_set::timestamp_correction::id())
+        {
+            if let Some(timestamp_estimate) = self.get_timestamp_estimate() {
+                if timestamp_estimate > unix_timestamp {
+                    unix_timestamp = timestamp_estimate
+                }
+            }
+        }
         let clock = sysvar::clock::Clock {
             slot: self.slot,
             unused: Self::get_unused_from_slot(self.slot, self.unused),
             epoch: self.epoch_schedule.get_epoch(self.slot),
             leader_schedule_epoch: self.epoch_schedule.get_leader_schedule_epoch(self.slot),
-            unix_timestamp: self.unix_timestamp(),
+            unix_timestamp,
         };
         self.update_sysvar_account(&sysvar::clock::id(), |account| {
             clock.create_account(self.inherit_sysvar_account_balance(account))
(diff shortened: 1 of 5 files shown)


sdk/src/stake_weighted_timestamp.rs
```diff
@@ -0,0 +1,150 @@
...
+    account::Account,
+    clock::{Slot, UnixTimestamp},
+    pubkey::Pubkey,
+};
+use std::{collections::HashMap, time::Duration};
+
+pub const TIMESTAMP_SLOT_RANGE: usize = 16;
+
+pub fn calculate_stake_weighted_timestamp(
+    unique_timestamps: HashMap<Pubkey, (Slot, UnixTimestamp)>,
```
garious, line 13: Can this take a reference?

runtime/src/bank.rs
```diff
@@ -1311,6 +1327,47 @@ impl Bank {
...
+                        ))
+                    } else {
+                        None
+                    }
+                })
+            })
+            .collect();
+        let slot_duration = Duration::from_nanos(self.ns_per_slot as u64);
+        let epoch = self.epoch_schedule().get_epoch(self.slot());
+        let stakes = HashMap::new();
```
garious, line 1351: I wonder if an empty HashMap implies a heap allocation.  Can we avoid answering that by returning None when epoch_vote_accounts returns None?

runtime/src/bank.rs
```diff
@@ -1311,6 +1327,47 @@ impl Bank {
         self.update_recent_blockhashes_locked(&blockhash_queue);
     }
 
+    fn get_timestamp_estimate(&self) -> Option<UnixTimestamp> {
+        let mut get_timestamp_estimate_time = Measure::start("get_timestamp_estimate");
+        let recent_timestamps: HashMap<Pubkey, (Slot, UnixTimestamp)> = self
+            .vote_accounts()
```
garious, line 1333: Is this the rooted vote accounts? That is, are we sure all validators will calculate the same timestamp?

garious (commented):
Looks about right

runtime/src/bank.rs
CriesofCarrots: Ah, yep definitely

runtime/src/bank.rs
CriesofCarrots: These are the vote accounts in the new (from parent) Bank. Same info that is used for rewards, epoch-stakes, etc. I think that if we have nodes coming up with different Bank/accounts state for the same slot, we have a bigger problem?

runtime/src/bank.rs
```diff
@@ -985,19 +987,33 @@ impl Bank {
...
+
+    fn update_clock(&self) {
+        let mut unix_timestamp = self.unix_timestamp_from_genesis();
+        if self
+            .feature_set
+            .is_active(&feature_set::timestamp_correction::id())
+        {
+            if let Some(timestamp_estimate) = self.get_timestamp_estimate() {
+                if timestamp_estimate > unix_timestamp {
+                    unix_timestamp = timestamp_estimate
```
mvines, line 1004: Add a metric to monitor the corrections easily?

mvines (approved):
Pretty simple, I like it

garious (dismissed):

CriesofCarrots:
Oh https://github.com/solana-labs/solana/pull/12742 💔

PR #12750, opened 2020-10-09
Solarweave Implementation

TheLoneRonin:
*Moved from https://github.com/solana-labs/explorer/pull/168*

## Key Features

This pull request implements Solarweave in a completely unobtrusive way. It has automatic fallback compatibility to the original implementation. **Therefore, no actual cache from Solarweave needs to be present for the Explorer to work.** However, it will try to query Solarweave first.

*Please note that the block page will not work unless an existing Solarweave Cache is implemented.*

### Search by blockhash and slot

You can now search for blocks using slots:

image

You can also search for blocks by using blockhashes:

image

### The new block page

The new block page showcases an overview of the block as well as all the transactions in the block.

image

If an invalid block is inputted. It will simply just display this page.

### Custom Solarweave Cluster Info

If you wanted to retrieve data from a custom Solarweave Database you can use the updated Cluster Info page as seen here.

### Full fallback compatibility

There is automatic fallback compatibility for the original block explorer. This means that even if there is no Solarweave cache. The Explorer will still work as intended. Just remember that block pages will not work if there is no Solarweave cache.

## Other notes

- Block History does not use Solarweave. Please see updates on: https://github.com/ArweaveTeam/arweave/issues/268 if considering using Arweave for Block History in the future.


explorer/src/components/ClusterModal.tsx
```diff
@@ -66,36 +66,66 @@ function CustomClusterInput({ activeSuffix, active }: InputProps) {
...
-    </Link>
+      </Link>
+
+      <Link
+        to={(location) => clusterLocation(location)}
+        className="btn input-group input-group-merge p-0"
+      >
+        <input
+          type="text"
+          placeholder="Solarweave Database"
```
jstarry, line 112: Do you expect it to be common to use other databases? I was expecting there to be a single arweave database for each cluster.

explorer/src/components/SearchBar.tsx
```diff
@@ -189,6 +189,32 @@ function buildOptions(search: string, cluster: Cluster) {
...
+      label: "Blockhash",
+      options: [
+        {
+          label: `Blockhash ${search}`,
+          value: search,
+          pathname: `/blockhash/${search}`,
+        },
+      ],
+    });
+  }
```
jstarry, line 216: Blockhash's look just like accounts actually. They are both base58 encoded 32 byte values. So this option should be added whenever the account option is added below

explorer/src/components/account/BlockHistoryCard.tsx
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  function invalidTransactions() {
+    return (
+      <tr>
+        <td>ERROR</td>
+        <td>Could not retrieve transactions</td>
+      </tr>
+    );
+  }
```
jstarry, line 38: Please match the error state for accounts / transactions. Looks like this for invalid value

Screen Shot 2020-10-09 at 2 16 03 PM

Looks like this when not found
Screen Shot 2020-10-09 at 2 18 32 PM

explorer/src/App.tsx
```diff
@@ -43,6 +44,20 @@ function App() {
               <TransactionDetailsPage signature={match.params.signature} />
             )}
           />
+          <Route
+            exact
+            path={"/slot/:slot"}
+            render={({ match }) => (
+              <BlockDetailsPage slot={match.params.slot} blockhash={""} />
+            )}
```
jstarry, line 52: Sorry, I changed my mind from before. I don't think we need the `/slot/:slot` path because blocks can be uniquely identified by both slot number and hash. So I think it's best to combine these under a single route path `/block/:id` where id can be either a number slot or string hash.

explorer/src/components/SearchBar.tsx
```diff
@@ -40,7 +40,7 @@ export function SearchBar() {
             ref={(ref) => (selectRef.current = ref)}
             options={buildOptions(search, cluster)}
             noOptionsMessage={() => "No Results"}
-            placeholder="Search for accounts, transactions, programs, and tokens"
+            placeholder="Search for slots, blocks, accounts, transactions, programs, and tokens"
```
jstarry, line 43:
```suggestion
            placeholder="Search for blocks, accounts, transactions, programs, and tokens"
```

explorer/src/components/SearchBar.tsx
```diff
@@ -189,6 +189,32 @@ function buildOptions(search: string, cluster: Cluster) {
...
+      label: "Slot",
+      options: [
+        {
+          label: `Slot #${search}`,
+          value: search,
+          pathname: `/slot/${search}`,
+        },
+      ],
+    });
+  }
```
jstarry, line 203:
```suggestion
  if (!isNaN(Number(search))) {
    options.push({
      label: "Block",
      options: [
        {
          label: `Block #${search}`,
          value: search,
          pathname: `/block/${search}`,
        },
      ],
    });
  }
```

explorer/src/components/account/BlockHistoryCard.tsx
```diff
@@ -0,0 +1,112 @@
...
+            ? block.data?.Tags
+              ? block.data?.Tags.filter(
+                  (tag: any) =>
+                    tag.name === "slot" ||
+                    tag.name === "parentSlot" ||
+                    tag.name === "blockhash"
+                ).map((tag: any) => {
+                  return (
+                    <tr>
+                      <td>{tag.name}</td>
```
jstarry, line 60: Please uppercase the first character of the tag

explorer/src/components/account/BlockHistoryCard.tsx
```diff
@@ -0,0 +1,112 @@
...
+                      <td className="text-lg-right">{tag.value}</td>
```
jstarry, line 61: Parent slot value should be "linkified" for easy navigation. In fact, it would be great if we create a `BlockSlot` component (similar to `Address`) which has some options for turning slot numbers into links.

explorer/src/providers/accounts/block.tsx
```diff
@@ -0,0 +1,116 @@
+import React from "react";
```
jstarry, line 1: Can you please move this file from `./providers/accounts/block.tsx` to `./providers/blocks.tsx`?

explorer/src/providers/accounts/block.tsx
```diff
@@ -0,0 +1,116 @@
...
+  React.useEffect(() => {
+    dispatch({ type: ActionType.Clear, url });
+  }, [dispatch, url]);
+
+  return (
+    <StateContext.Provider value={state}>
+      <DispatchContext.Provider value={dispatch}>
+        <TokensProvider>
+          <HistoryProvider>{children}</HistoryProvider>
+        </TokensProvider>
```
jstarry, line 36: please remove the token and history providers, I think they were copy pasted 😉

explorer/src/providers/cluster.tsx
```diff
@@ -67,11 +67,25 @@ export function clusterUrl(cluster: Cluster, customUrl: string): string {
...
 }
 
+export function solarweaveUrl(cluster: Cluster, solarweaveUrl: string): string {
+  switch (cluster) {
+    case Cluster.Devnet:
+      return "solana-devnet-solarweave";
+    case Cluster.MainnetBeta:
+      return "solana-mainnet-solarweave";
+    case Cluster.Testnet:
+      return "solana-testnet-solarweave";
```
jstarry, line 77: Do these already have data or are they placeholders?

explorer/package.json
```diff
@@ -9,6 +9,7 @@
     "@testing-library/jest-dom": "^5.11.4",
     "@testing-library/react": "^11.0.4",
     "@testing-library/user-event": "^12.1.7",
+    "@theronin/solarweave": "^1.4.0",
```
jstarry, line 12: Maybe we publish this under `@solana`? I'm not sure what the handoff / maintenance going forward looks like :)

explorer/src/providers/accounts/block.tsx
```diff
@@ -0,0 +1,116 @@
...
+type Dispatch = Cache.Dispatch<Block>;
+
+const StateContext = React.createContext<State | undefined>(undefined);
+const DispatchContext = React.createContext<Dispatch | undefined>(undefined);
+
+type BlockProviderProps = { children: React.ReactNode };
+
+export interface Block {
+  BlockData: any;
+  Tags: any;
```
jstarry, line 20:
```suggestion
  blockData: any;
  tags: any;
```

explorer/src/providers/accounts/block.tsx
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  dispatch({
+    type: ActionType.Update,
+    url,
+    key,
+    status,
+    data: {
+      BlockData: result?.BlockData,
+      Tags: result?.Tags,
```
jstarry, line 93: We should use superstruct to validate the response here. There are a few examples around the codebase of how to set this up

explorer/src/providers/accounts/block.tsx
```diff
@@ -0,0 +1,116 @@
...
+  });
+}
+
+export function useFetchBlock() {
+  const { cluster, url, solarweave } = useCluster();
+  const state = React.useContext(StateContext);
+  const dispatch = React.useContext(DispatchContext);
+
+  if (!state || !dispatch) {
+    throw new Error(`useBlock must be used within a BlockProvider`);
```
jstarry, line 104:
```suggestion
    throw new Error(`useFetchBlock must be used within a BlockProvider`);
```

explorer/src/providers/transactions/details.tsx
```diff
@@ -55,9 +116,16 @@ async function fetchDetails(
   let fetchStatus;
   let transaction;
   try {
-    transaction = await new Connection(url).getParsedConfirmedTransaction(
-      signature
-    );
+    const tx = await RetrieveSignature(signature, `${solarweave}-index`);
```
jstarry, line 119: Is this database specific to signatures? I would expect the name to be something like `${solarweave}-signatures`

jstarry: I'd like this to be in a nested try catch so that if for some reason the arweave api request fails, we still fall back to the RPC api

jstarry (commented):
Looks awesome! Very excited to finally have a block details page soon!

explorer/src/components/ClusterModal.tsx
TheLoneRonin: If someone is querying a custom RPC node.
It might also be a forked chain.

Thus, I think a custom Solarweave Database would make sense!

explorer/src/components/SearchBar.tsx
TheLoneRonin: Okay perfect, thank you!
Just updated that.

explorer/src/App.tsx
TheLoneRonin: Oh no worries! It's actually a very easy check. Just gotta run an `isNaN` check!
Will refactor!

TheLoneRonin: Also, agree it's better semantics wise.

explorer/src/providers/transactions/details.tsx
TheLoneRonin: The reason it's `index` is because the `index` database is able to retrieve blocks based on account keys and signatures. So it encapsulates multiple tags.

The `try` `catch` statement is essential though!

explorer/src/components/account/BlockHistoryCard.tsx
TheLoneRonin: Ohhhh I like this idea a lot!
Thanks, will add that in.

explorer/src/components/account/BlockHistoryCard.tsx
TheLoneRonin: I'm going to change it `Slot` and `Parent Slot` and `Blockhash`.
So will have to enumerate it accordingly.

explorer/src/providers/accounts/block.tsx
TheLoneRonin: Indeed they were hahaha.

explorer/src/providers/cluster.tsx
TheLoneRonin: Only devnet is partially populated.
These are currently placeholders.

I still need archive nodes that have block data from block 0 for each network.

explorer/package.json
TheLoneRonin: Yea would be great! Feel free to make a new repo and then add me as a Collaborator and I will move the codebase over there.

jstarry:
> Only devnet is partially populated.
These are currently placeholders.
I still need archive nodes that have block data from block 0 for each network.




@TheLoneRonin since it will be awhile before block data is stored on arweave, I think it's best to keep arweave calls out of the production explorer. I think you've proven how easy it is to drop in the arweave api which is great and sufficient for the requirements of this project.

You've already laid the groundwork for adding a block details page so I think we should continue forward with that but have it call the RPC API `getConfirmedBlock` instead of arweave since we know that won't have data for awhile. Thoughts? If you enjoy the task, feel free to make those changes! Otherwise I think we could take over your PR changes. Up to you 😄

TheLoneRonin:
@jstarry,

Agree with this!
I think the only difference in changes is that you would only be able to retrieve by slot number.

Furthermore, if the RPC node does not have the full chain stored.
There are some problems that may arise from getting *older* blocks.

In this case, what I will do is keep this pull request open.
Open a new pull request that is a fork of this branch.

And have it query the RPC command instead.

---

Expanding on this, in the future. We do plan on having events stored in Solarweave.
This is a critical feature for reading smart contracts.

Maybe when this is successfully implemented. We can then look into getting Solarweave fully integrated into the Solana Explorer. Since when that is implemented, Solarweave will have a lot more utility.

PR #12768, opened 2020-10-09
Expose all rewards (fees, rent, voting and staking) in RPC getConfirmedBlock and the cli

mvines:
The rewards array in `getConfirmedBlock` has a couple limitations that this PR addresses:
1. No indication of the type of reward -- was it for voting and staking?  This can be inferred by looking at the type of account of course but that's clumsy
2. Transaction fees and rent distributions are not surfaced

With this PR `getConfirmedBlock` will now report at minimum one reward per block, the transaction fees paid to the leader.

### DIFF
--- a/cli/src/cluster_query.rs
+++ b/cli/src/cluster_query.rs
@@ -73,8 +73,7 @@ impl ClusterQuerySubCommands for App<'_, '_> {
                         .validator(is_slot)
                         .value_name("SLOT")
                         .takes_value(true)
-                        .index(1)
-                        .required(true),
+                        .index(1),
                 ),
         )
         .subcommand(
@@ -363,7 +362,7 @@ pub fn parse_cluster_ping(
 }
 
 pub fn parse_get_block(matches: &ArgMatches<'_>) -> Result<CliCommandInfo, CliError> {
-    let slot = value_t_or_exit!(matches, "slot", Slot);
+    let slot = value_of(matches, "slot");
     Ok(CliCommandInfo {
         command: CliCommand::GetBlock { slot },
         signers: vec![],
@@ -700,7 +699,17 @@ pub fn process_leader_schedule(rpc_client: &RpcClient) -> ProcessResult {
     Ok("".to_string())
 }
 
-pub fn process_get_block(rpc_client: &RpcClient, _config: &CliConfig, slot: Slot) -> ProcessResult {
+pub fn process_get_block(
+    rpc_client: &RpcClient,
+    _config: &CliConfig,
+    slot: Option<Slot>,
+) -> ProcessResult {
+    let slot = if let Some(slot) = slot {
+        slot
+    } else {
+        rpc_client.get_slot()?
+    };
+
     let mut block =
         rpc_client.get_confirmed_block_with_encoding(slot, UiTransactionEncoding::Base64)?;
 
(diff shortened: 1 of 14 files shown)


cli/src/cluster_query.rs
```diff
@@ -73,8 +73,7 @@ impl ClusterQuerySubCommands for App<'_, '_> {
                         .validator(is_slot)
```
mvines, line 73: `solana block` now dumps the latest confirmed block instead an error, which is a nice feature when watching for transaction fees.

core/src/replay_stage.rs
```diff
@@ -291,7 +291,6 @@ impl ReplayStage {
                         &bank_forks,
```
mvines, line 291: ReplayStage needed some rejigging.  `record_rewards` needs to happen after the bank is frozen instead of immediately upon Bank creation

storage-bigtable/src/convert.rs
```diff
@@ -24,6 +25,13 @@ impl From<Reward> for generated::Reward {
             pubkey: reward.pubkey,
```
mvines, line 25: Getting enums in/out of protobuf is a little clunky

transaction-status/src/lib.rs
```diff
@@ -241,6 +242,8 @@ pub struct Reward {
     pub lamports: i64,
     #[serde(deserialize_with = "default_on_eof")]
     pub post_balance: u64, // Account balance in lamports after `lamports` was applied
+    #[serde(default, deserialize_with = "default_on_eof")]
```
mvines, line 245: `#[serde(deserialize_with = "default_on_eof")]` was insufficient to satisfy `serde_json` when trying to fetch an existing confirmed block from testnet without the `reward_type` field.  Adding `#[serde(default)]` as well made it happy

storage-bigtable/src/convert.rs
mvines: That doesn't seem much better, I'll keep the existing code I think

transaction-status/src/lib.rs
mvines: With this PR, remove the `default` and then:
```
$ cargo run -- --verbose block 40700256
    Finished dev [unoptimized + debuginfo] target(s) in 0.34s
     Running `/Users/mvines/ws/s2/target/debug/solana --verbose block 40700256`
RPC URL: http://testnet.solana.com
Default Signer Path: /Users/mvines/.config/solana/id.json
Error: missing field `rewardType`
```

serde_json doesn't even hit the `default_on_eof` deserializer, it already decided that there's a missing field in the response and it can't do anything about it.  Deriving `Default` for the entire `Reward` struct produces the same error

runtime/src/bank.rs
```diff
@@ -523,8 +523,32 @@ impl PartialEq for Bank {
     }
 }
 
-#[derive(Debug, PartialEq, Serialize, Deserialize, AbiExample, Default, Clone, Copy)]
+#[derive(Debug, PartialEq, Serialize, Deserialize, AbiExample, AbiEnumVisitor, Clone, Copy)]
+pub enum RewardType {
+    Fee,
+    Rent,
+    Voting,
+    Staking,
```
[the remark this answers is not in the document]
mvines: Sure, done.  Of course the next time we add a reward, it needs to start W, X, Y or Z in order to keep things alphabetical

runtime/src/bank.rs
```diff
@@ -5717,14 +5791,15 @@ mod tests {
 
         // verify validator rewards show up in bank1.rewards vector
         assert_eq!(
-            bank1.rewards,
-            Some(vec![(
+            *bank1.rewards.read().unwrap(),
+            vec![(
                 stake_id,
                 RewardInfo {
+                    reward_type: RewardType::Staking,
```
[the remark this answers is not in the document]
mvines: I snuck a little something into `test_bank_update_rewards_determinism()`

transaction-status/src/lib.rs
CriesofCarrots: Yeah, `serde_json` doesn't need the `default_on_eof`. That's for bincode going into and out of blockstore.

...Ah, okay, figured it out. I was curious why this isn't affection `log_messages`, which also aren't on testnet right now. It's because the meta are being parsed by serde_json into `UiTransactionStatusMeta`, which don't include the `default_on_eof` tag.
Similarly, the serde_json parsing of `Reward` from rpc works fine with the `default_on_eof` tag removed. 

tl;dr, we'll need to remember `#[serde(default)]` for new `default_on_eof` fields on structs that are used directly in rpc.

runtime/src/bank.rs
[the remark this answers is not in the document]
CriesofCarrots: Challenge accepted! 😉

CriesofCarrots (approved):

PR #12778, opened 2020-10-09
Retry inflation parameter update along with documentation updates

ericlwilliams:
#### Problem
Apparently inflation parameter PR didn't go through so those commits still on this fork. Primarily wanted to update some old inflation documentation in docs.


mvines:
Hey LMK if you need help getting past the CI gauntlet here @ericlwilliams!

mvines:
@ericlwilliams - it's ok to just update the bank hash in `test_bank_hash_consistency`.  It's just different due to the change to DEFAULT_INITIAL.   There's four hashes that need to be updated though.  Best way to figure them is just to change the first one, run the test again, and let the second hash fail.  Repeat

sdk/src/inflation.rs
```diff
@@ -22,7 +22,7 @@ pub struct Inflation {
     __unused: f64,
 }
 
-const DEFAULT_INITIAL: f64 = 0.15;
+const DEFAULT_INITIAL: f64 = 0.08;
```
ryoqun, line 25: well, this needed backporting..

PR #9527, opened 2020-04-16
Introduce eager rent collection

ryoqun:
#### Problem

rent-due (!= rent-exempt) accounts are never garbage-collected unless updated (i.e. lazy collection), which accumulates over time and can be a DoS attack vector.

Also, the current rent collection mechanism has some gaming factor due to being lazy. (noted below)

Even more, there is [some UX issue for the lazy rent collection](https://github.com/solana-labs/solana/issues/7413#issuecomment-615136356).

#### Summary of Changes

- scan the whole pubkey domain space over an epoch (or 2 day (if epoch is less than 2 day and not warmed-up); ~not implemented yet~ _implemented_) for each slot progress, while updating all accounts.
  - (cons) **`AccountIndex` was switched from `HashMap` to `BTreeMap` for range queries to realize that**
- (pro) thus, maximum number of `AppendVec` is now bound to the number of slots of given epoch, combined with the recently-introduced per-slot shrinking #9219.
  - Also, this fixes unbound increase of `AccountsIndex.roots`.
- as a side effect, the rent collection schedule is now universally consistent: all existing accounts and newly-created accounts both alike are collected for the next epoch for each epoch exactly once.
  - (pro) technically, we can remove `account.rent_epoch` as a bonus. (`ACCOUNT_STORAGE_OVERHEAD`: 128 -> 120, -6.25% for IO bandwidth)
- the iteration algorithm intentionally designed to be deterministic as a requirement because rent collection generally affects account delta bank hash, distribution, and capitalization. Also it's stete-less as desired to avoid ABI breakage.
- (pro) the system load is well spread across an epoch because we can assume vast of pubkey space is uniformly distributed.
- (pro) this doesn't introduce ABI break; only an epoch-triggered enactment `if` guard is needed.

##### minor problem of current rent colletion

_Also I found that the lazy rent collection is problematic because delayed rent collection doesn't account for past rent fees were different or not._
That means the current calculation is based on the latest rent fee multiple of epochs. it's not the integral of past actual (dynamic) rent fees over epochs.
This means it introduces some gaming much like recently removed `redeem-vote-credits` (#7916):
A validator or an account owner might send 0 lamports to rent-due accounts to collect/save rent when the at-the-time dynamic rent is more profitable.
Although, the dynamic rent isn't still implemented...

##### (cons) new dos vector

[...]

PR #12787, opened 2020-10-10
Break up AccountsIndex lock

carllin:
#### Problem
Holding AccountsIndex lock during scans blocks account stores

#### Summary of Changes
Builds on top of https://github.com/solana-labs/solana/pull/12126

Benchmark `bench_concurrent_scan_write` here shows 1000 inserts of new pubkeys into the storage taking about 9-10ms (80% still waiting on the AccountsIndex write lock), even as the index/accountsdb storage size grows. Before this change, the inserts tarting taking hundreds of ms.

Maybe we would use something like a https://github.com/crossbeam-rs/crossbeam/tree/master/crossbeam-skiplist to optimize contention the core lock farther, but there does not seem to be any stable alternatives yet.
Fixes #


runtime/Cargo.toml
```diff
@@ -31,6 +32,7 @@ num_cpus = "1.13.0"
 rand = "0.7.0"
 rayon = "1.4.1"
 regex = "1.3.9"
+rental = "0.5.4"
```
ryoqun, line 35: :eyes: https://crates.io/crates/rental

runtime/benches/accounts.rs
```diff
@@ -141,3 +142,103 @@ fn bench_delete_dependencies(bencher: &mut Bencher) {
         accounts.accounts_db.clean_accounts(None);
     });
 }
+
+#[bench]
+#[ignore]
+fn bench_concurrent_read_write(bencher: &mut Bencher) {
+    let num_readers = 5;
+    let accounts = Arc::new(Accounts::new(
+        vec![PathBuf::from("concurrent_read_write")],
```
ryoqun, line 151: nits recognize FARF_DIR?

runtime/benches/accounts.rs
```diff
@@ -141,3 +142,103 @@ fn bench_delete_dependencies(bencher: &mut Bencher) {
...
+            accounts.store_slow(slot + 1, &Pubkey::new_rand(), &account);
+        }
+    })
+}
+
+#[bench]
+fn bench_concurrent_scan_write(bencher: &mut Bencher) {
+    let num_readers = 5;
+    let accounts = Arc::new(Accounts::new(
+        vec![PathBuf::from("concurrent_read_write")],
```
ryoqun, line 201: nits recognize FARF_DIR?

runtime/benches/accounts.rs
```diff
@@ -141,3 +142,103 @@ fn bench_delete_dependencies(bencher: &mut Bencher) {
...
+    for _ in 0..num_readers {
+        let accounts = accounts.clone();
+        let pubkeys = pubkeys.clone();
+        Builder::new()
+            .name("readers".to_string())
+            .spawn(move || {
+                let mut rng = rand::thread_rng();
+                loop {
+                    let i = rng.gen_range(0, num_keys);
+                    accounts.load_slow(&HashMap::new(), &pubkeys[i]);
```
ryoqun, line 177: nits: `unwrap` should be safe, right?

nits: also how about using `test::black_box` in the fear of human-level clever `rustc` in the distant future? ;)

runtime/benches/accounts.rs
```diff
@@ -141,3 +142,103 @@ fn bench_delete_dependencies(bencher: &mut Bencher) {
...
+    ));
+    let num_keys = 1000;
+    let slot = 0;
+    accounts.add_root(slot);
+    let default_owner = Account::default().owner;
+    let pubkeys: Arc<Vec<_>> = Arc::new(
+        (0..num_keys)
+            .map(|_| {
+                let pubkey = Pubkey::new_rand();
+                let account = Account::new(1, 0, &Account::default().owner);
```
ryoqun, line 212: nits: can use `default_owner`?

runtime/benches/accounts.rs
```diff
@@ -141,3 +142,103 @@ fn bench_delete_dependencies(bencher: &mut Bencher) {
...
+    );
+
+    for _ in 0..num_readers {
+        let accounts = accounts.clone();
+        let pubkeys = pubkeys.clone();
+        Builder::new()
+            .name("readers".to_string())
+            .spawn(move || {
+                loop {
+                    accounts.load_by_program(&HashMap::new(), &default_owner);
```
ryoqun, line 226: nits: like https://github.com/solana-labs/solana/pull/12787/files#r503695336, how about `assert!(!loaded_accounts.is_empty())` and `test::black_box()`?

runtime/benches/accounts.rs
```diff
@@ -141,3 +142,103 @@ fn bench_delete_dependencies(bencher: &mut Bencher) {
...
+
+    let num_new_keys = 1000;
+    let new_accounts: Vec<_> = (0..num_new_keys)
+        .map(|_| Account::new(1, 0, &Account::default().owner))
+        .collect();
+    bencher.iter(|| {
+        for account in &new_accounts {
+            // Write to a different slot than the one being read from. Because
+            // there's a new account pubkey being written to every time, will
+            // compete for the accounts index lock on every store
```
ryoqun, line 191: good comment!

runtime/benches/accounts.rs
```diff
@@ -141,3 +142,103 @@ fn bench_delete_dependencies(bencher: &mut Bencher) {
...
+            accounts.store_slow(slot + 1, &Pubkey::new_rand(), &account);
+        }
```
ryoqun, line 193: nits: how about extracting to something like this and share with `bench_concurrent_scan_write`, too?: `store_accounts_with_possible_contention`. (`possible` is explicitly picked in the consideration of the improvement in this pr itself). :)

runtime/benches/accounts.rs
carllin: good suggestion, updated on other PR as well!

runtime/src/accounts.rs
```diff
@@ -302,8 +302,7 @@ impl Accounts {
     ) -> Vec<(Result<TransactionLoadResult>, Option<HashAgeKind>)> {
         //PERF: hold the lock to scan for the references, but not to clone the accounts
         //TODO: two locks usually leads to deadlocks, should this be one structure?
```
ryoqun, line 304: Grab your fruit: now prime time to remove this outdated comment :)

runtime/src/accounts_index.rs
```diff
@@ -1,38 +1,276 @@
...
 use std::{
     collections::{BTreeMap, HashMap, HashSet},
     ops::RangeBounds,
-    sync::{RwLock, RwLockReadGuard},
+    sync::{Arc, RwLock},
 };
 
+const ITER_BATCH_SIZE: usize = 1000;
+
+rental! {
```
ryoqun, line 15: interesting. :)

runtime/src/accounts_index.rs
```diff
@@ -1,38 +1,276 @@
...
+            .read()
+            .unwrap()
+            .range((self.start_bound, self.end_bound))
+            .map(|(pubkey, account_map_entry)| {
+                (*pubkey, account_map_entry.clone())
+            })
+            .take(ITER_BATCH_SIZE)
+            .collect();
+
+        if chunk.is_empty() {
```
ryoqun, line 185: nits: there is very and very small corner case of violating `iter` invariant: we could return `Some` even after we returned `None` here if we got newer Pubkey inserted, which would be compared after current `self.start_bound`. how about closing this hole now? Consider data parallelism from `rayon` and imagine how hard to spot this bug in the future from insane behavior. :)

Sadly, (start, end) = (unbound, unbound) or anything like that wouldn't work to mark the state. So, we must add another flag `finished` or like that and set it to `true` here.

runtime/src/accounts_index.rs
```diff
@@ -1,38 +1,276 @@
...
 
 pub type RefCount = u64;
-type AccountMapEntry<T> = (AtomicU64, RwLock<SlotList<T>>);
 pub type AccountMap<K, V> = BTreeMap<K, V>;
 
+#[derive(Clone, Debug)]
+pub struct AccountMapEntry<T> {
+    ref_count: Arc<AtomicU64>,
+    pub slot_list: Arc<RwLock<SlotList<T>>>,
+}
```
ryoqun, line 52: I wonder we could reduce Arc to 1 from 2 by line this, it's a bit hideous but maybe worth effort? (we should be keen the size of AcccontIndex memory footprint (Arc costs 2-word memory for each, so we're increasing entry size by `4 * usize` here per `Account`.) and reduced general update cost from updating Arc twice everytime accessed).

```rust
pub struct AccountMapEntryInner<T> {
    ref_count: AtomicU64,
    pub slot_list: RwLock<SlotList<T>>,
}
type AccountMapEntry<T> = Arc<AccountMapEntryInner>;
```

ryoqun:
> Benchmark `bench_concurrent_scan_write` here shows 1000 inserts of new pubkeys into the storage taking about 9-10ms (80% still waiting on the AccountsIndex write lock), even as the index/accountsdb storage size grows. Before this change, the inserts tarting taking hundreds of ms.

I'm really happy to see this report. :)


How about any perf changes of `colo-gpu-perf-4-val-1-client.yml` with this(#12787) & previous dashmap (#12126),  compared to the base commit?

runtime/benches/accounts.rs
carllin: Updated!

runtime/src/accounts_index.rs
carllin: Fixed and added test!

runtime/benches/accounts.rs
```diff
@@ -196,3 +193,31 @@ fn bench_concurrent_read_write(bencher: &mut Bencher) {
...
+}
+
+#[bench]
+#[ignore]
+fn bench_concurrent_scan_write(bencher: &mut Bencher) {
+    store_accounts_with_possible_contention(
+        "concurrent_scan_write",
+        bencher,
+        |accounts, pubkeys| loop {
+            accounts.load_by_program(&HashMap::new(), &Account::default().owner);
```
ryoqun, line 220: nits: test::black_box here as well?

runtime/src/accounts_index.rs
```diff
@@ -160,72 +405,31 @@ impl<'a, T: 'a + Clone> AccountsIndex<T> {
...
-
-    // Try to update an item in account_maps. If the account is not
-    // already present, then the function will return back Some(account_info) which
-    // the caller can then take the write lock and do an 'insert' with the item.
-    // It returns None if the item is already present and thus successfully updated.
-    pub fn update(
+    // Updates the given pubkey at the given slot with the new account information.
+    // Returns true if the pubkey was newly inserted into the index, otherwise, if the
+    // pubkey updates an existing entry in the index, returns false.
+    pub fn update_or_create_if_missing(
```
ryoqun, line 411: nits: I welcome less confusing name so thanks for taking a bite on this. but this is a bit long relatively for frequently called method. how about just `upsert`?

runtime/src/accounts_index.rs
ryoqun: thinking on this.... hmm.

carllin: Updated with the ouroboros suggestion, thanks!

runtime/Cargo.toml
carllin: gone!

runtime/src/accounts_index.rs
carllin: that works for me!

runtime/benches/accounts.rs
carllin: ah forgot, thanks!

core/src/validator.rs
```diff
@@ -1056,7 +1056,6 @@ fn report_target_features() {
         // Validator binaries built on a machine with AVX support will generate invalid opcodes
         // when run on machines without AVX causing a non-obvious process abort.  Instead detect
         // the mismatch and error cleanly.
-        #[target_feature(enable = "avx")]
```
ryoqun, line 1059: nits: unrelated change?

carllin: uh super annoying, my `cargo fmt` keeps deleting this line... thanks!

runtime/src/accounts_db.rs
```diff
@@ -1311,37 +1303,52 @@ impl AccountsDB {
         accounts_index: &AccountsIndex<AccountInfo>,
         pubkey: &Pubkey,
     ) -> Option<(Account, Slot)> {
-        let (lock, index) = accounts_index.get(pubkey, Some(ancestors), None)?;
-        let slot = lock[index].0;
+        let (slot, store_id, offset) = {
+            let (lock, index) = accounts_index.get(pubkey, Some(ancestors), None)?;
+            let slot_list = lock.slot_list();
+            let slot = slot_list[index].0;
```
ryoqun, line 1309: nits: can't we do like this destructuring?

```rust
let (slot, {store_id, offset, ...}) = slot_list[index];
```

runtime/src/accounts_db.rs
```diff
@@ -1311,37 +1303,52 @@ impl AccountsDB {
...
-        let info = &lock[index].1;
+        let (slot, store_id, offset) = {
+            let (lock, index) = self
+                .accounts_index
+                .get(pubkey, Some(ancestors), None)
+                .unwrap();
+            let slot_list = lock.slot_list();
+            let slot = slot_list[index].0;
+            let store_id = slot_list[index].1.store_id;
+            let offset = slot_list[index].1.offset;
```
ryoqun, line 1337: same as this: https://github.com/solana-labs/solana/pull/12787/files?diff=split&w=1#r505171426

core/src/validator.rs
ryoqun: protip: `(unset RUSTFLAGS && cargo +1.45.0 fmt --all)` This way I can live with ever-evolving nightly. :)

runtime/src/accounts_index.rs
```diff
@@ -1,38 +1,305 @@
...
-    sync::{RwLock, RwLockReadGuard},
+    ops::{Deref, DerefMut, Range, RangeBounds},
+    sync::{Arc, RwLock, RwLockReadGuard, RwLockWriteGuard},
 };
+const ITER_BATCH_SIZE: usize = 1000;
+
+#[self_referencing]
+pub struct OwningReadGuard<T>
+where
+    T: 'static,
```
ryoqun, line 18: nits: no need `where`?:

```rust
pub struct OwningReadGuard<T: 'static>
{
   ...
```

runtime/src/accounts_index.rs
```diff
@@ -1,38 +1,305 @@
...
+    type Target = T;
+    fn deref(&self) -> &Self::Target {
+        &*self.guard
+    }
+}
+
+#[self_referencing]
+pub struct OwningWriteGuard<T>
+where
+    T: 'static,
```
ryoqun, line 35: same as https://github.com/solana-labs/solana/pull/12787/files?diff=split&w=1#r505174649

runtime/src/accounts_db.rs
carllin: yup totally!

carllin: yup totally, updated!

carllin:
@ryoqun been running for past 20 hours or so with snapshot verification: https://metrics.solana.com:3000/d/monitor-edge/cluster-telemetry-edge?orgId=2&from=1602723051575&to=1602796347486&panelId=35&fullscreen&var-datasource=Solana%20Metrics%20(read-only)&var-testnet=mainnet-beta&var-hostid=aeypRbHck9k5zGuftrdSaUMfjJyTrjtaGJEtALt835y, no issues yet!

runtime/src/accounts_index.rs
carllin: @ryoqun  this is doable, but I think it makes the self-referential struct a little uglier because now you have to wrap the entire `AccountsMapEntryInner` instead of just the lock inside the `self-reference` tag. Here's an example:  https://github.com/solana-labs/solana/pull/12787/commits/50a9b2647d6a47c0758b78e277731e5731accd5c

What do you think?

carllin: @ryoqun this is much cleaner with the ouroboros crate here: https://github.com/solana-labs/solana/pull/12787/commits/970b2c78aa1a04a980500591fd2cc5cbc6af6be5

runtime/src/accounts_index.rs
```diff
@@ -1,38 +1,297 @@
...
+    }
+
+    // Try to update an item in account_maps. If the account is not
+    // already present, then the function will return back Some(account_info) which
+    // the caller can then take the write lock and do an 'insert' with the item.
+    // It returns None if the item is already present and thus successfully updated.
+    pub fn update(&mut self, slot: Slot, account_info: T, reclaims: &mut SlotList<T>) {
+        // filter out other dirty entries from the same slot
+        let mut same_slot_previous_updates: Vec<(usize, (Slot, &T))> = self
+            .slot_list_mut()
```
ryoqun, line 120: nits: doesn't need to be `_mut`? So define `.slot_list()` and use it?

runtime/src/accounts_index.rs
```diff
@@ -1,38 +1,297 @@
...
+            .iter()
+            .enumerate()
+            .filter_map(|(i, (s, value))| {
+                if *s == slot {
+                    Some((i, (*s, value)))
```
ryoqun, line 125: nits: this looks so similar to the enclosing closure signature. Can we use `filter` instead of `filter_map` somehow? Because this is fairly in the tight loop. I want to squeeze out anything unneeded as much as possible. :)

runtime/src/accounts_index.rs
```diff
@@ -1,38 +1,297 @@
...
+    }
+
+    pub fn ref_count(&self) -> &AtomicU64 {
+        &self.owned_entry.ref_count
+    }
+
+    // Try to update an item in account_maps. If the account is not
+    // already present, then the function will return back Some(account_info) which
+    // the caller can then take the write lock and do an 'insert' with the item.
+    // It returns None if the item is already present and thus successfully updated.
```
ryoqun, line 116: outdated comment?

ryoqun:
> @ryoqun been running for past 20 hours or so with snapshot verification: https://metrics.solana.com:3000/d/monitor-edge/cluster-telemetry-edge?orgId=2&from=1602723051575&to=1602796347486&panelId=35&fullscreen&var-datasource=Solana%20Metrics%20(read-only)&var-testnet=mainnet-beta&var-hostid=aeypRbHck9k5zGuftrdSaUMfjJyTrjtaGJEtALt835y, no issues yet!

cool. :)

also, had the validator been slummed by `getProgramAccounts` RPC reqs? and no slowdowns of replay and no odd `panic!`s?

> How about any perf changes of `colo-gpu-perf-4-val-1-client.yml` with this(#12787) & previous dashmap (#12126), compared to the base commit?

As a ground wrapping-up of past various AccountsDB related improvements by you, how about writing like this? Too much hustle?  This is done for AccountIndex's `HashMap` => `BTreeMap` change.:
>  https://github.com/solana-labs/solana/pull/9527#issuecomment-627210347

you can peek around this build jobs to know how to create these pdfs or ask me :) : https://buildkite.com/solana-labs/system-performance-tests/builds?branch=pull%2F9527%2Fhead

also, please take a look at my previous similar reports for inspiration:

> https://github.com/solana-labs/solana/pull/8436#issuecomment-591865918
> https://github.com/solana-labs/solana/pull/8436#issuecomment-592390251

runtime/src/accounts_index.rs
carllin: yeah for sure, removed!

runtime/src/accounts_index.rs
```diff
@@ -1,38 +1,297 @@
...
+            ));
+        }
+
+        (w_account_entry.unwrap(), is_newly_inserted)
+    }
+
+    pub fn handle_dead_keys(&self, dead_keys: Vec<Pubkey>) {
+        if !dead_keys.is_empty() {
+            for key in &dead_keys {
+                let mut w_index = self.account_maps.write().unwrap();
```
ryoqun, line 289: just making sure: `w_index` is moved inside loop body to mitigate longer write lock?

runtime/src/accounts_index.rs
```diff
@@ -1,38 +1,297 @@
...
+            });
+            w_account_entry = Some(WriteAccountMapEntry::from_account_map_entry(
+                account_entry.clone(),
+            ));
+        }
+
+        (w_account_entry.unwrap(), is_newly_inserted)
+    }
+
+    pub fn handle_dead_keys(&self, dead_keys: Vec<Pubkey>) {
```
ryoqun, line 286: nits: how about taking a `&[Pubkey]`? I donno why clippy didn't warn for this. ;)

runtime/src/accounts_index.rs
```diff
@@ -1,38 +1,297 @@
...
+        if !dead_keys.is_empty() {
+            for key in &dead_keys {
+                let mut w_index = self.account_maps.write().unwrap();
+                if let Some(account_entry) = w_index.get(key) {
+                    if account_entry.slot_list.read().unwrap().is_empty() {
+                        w_index.remove(key);
+                    }
```
ryoqun, line 293: nits: I think we could reduce the lookup from 2 times to once like this:

```rust
if let Some(index_entry) = w_index.entry(key) {
    if (...) {
        index_entry.remove();
    }
}
```

ryoqun:
2nd review pass is done. I think this is really reaching to the lgtm. Thanks for defending my wave of nits attacks patiently. Well, I'm very tend to be interrupted by other concurrent tasks..

runtime/src/accounts_index.rs
carllin: hmm yeah, it should be a pretty fast loop though, so maybe it's not necessary?

runtime/src/accounts_index.rs
carllin: slight variation on this using `Occupied` instead of `Some`, but updated!

runtime/src/accounts_index.rs
```diff
@@ -1,38 +1,297 @@
...
+    fn deref(&self) -> &Self::Target {
+        &*self.guard
+    }
+}
+
+impl<T> DerefMut for OwningWriteGuard<T> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut *self.guard
+    }
+}
```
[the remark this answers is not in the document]
ryoqun: maybe, my joke was so subtle... Well, you can remove this 30+ lines because these aren't used anymore at all. :)

runtime/src/accounts_index.rs
ryoqun: hmm, I'm not sure too for the perf. difference.. Anyway, I just wanted to confirm this was somewhat intentional or just typo.

carllin:
@ryoqun, I couldn't get the PDF's to scale properly, but I was able to collect some metrics:

With the DashMap + Index change: https://metrics.solana.com:3000/d/monitor-edge/cluster-telemetry-edge?orgId=2&from=1602901242387&to=1602902152277&panelId=35&var-datasource=Solana%20Metrics%20(read-only)&var-testnet=testnet-dev-carl&var-hostid=All

Without DashMap + Index change: https://metrics.solana.com:3000/d/monitor-edge/cluster-telemetry-edge?orgId=2&from=1602902319061&to=1602903305200&panelId=35&var-datasource=Solana%20Metrics%20(read-only)&var-testnet=testnet-dev-carl&var-hostid=All

The confirmation + TPS are about the same, the biggest difference seems to be replay has fewer spikes with the changes: https://metrics.solana.com:3000/d/monitor-edge/cluster-telemetry-edge?orgId=2&from=1602901207646&to=1602903332757&panelId=35&fullscreen&var-datasource=Solana%20Metrics%20(read-only)&var-testnet=testnet-dev-carl&var-hostid=All&refresh=1m

ryoqun:
> @ryoqun, I couldn't get the PDF's to scale properly

Thanks for running these perf results, anyway! I created them for future reference.

[break-up-accounts-index-before.pdf](https://github.com/solana-labs/solana/files/5397459/break-up-accounts-index-before.pdf)
[break-up-accounts-index-after.pdf](https://github.com/solana-labs/solana/files/5397458/break-up-accounts-index-after.pdf)

I'm still looking the result. But, in general, there should be no bad perf. degradation and spike reduced a lot for some reason. Maybe the root contributor to it is the reduction of squashing, which I didn't expected but a good outcome. :)

runtime/src/accounts_db.rs
```diff
@@ -725,28 +723,37 @@ impl AccountsDB {
...
+                                *pubkey,
+                                self.accounts_index.roots_and_ref_count(&locked_entry),
+                            );
                         }
-                        if accounts_index.uncleaned_roots.contains(slot) {
-                            // Assertion enforced by `accounts_index.get()`, the latest slot
-                            // will not be greater than the given `max_clean_root`
-                            if let Some(max_clean_root) = max_clean_root {
-                                assert!(*slot <= max_clean_root);
-                            }
```
ryoqun, line 748: nits: this assert! is no longer needed?

ryoqun (approved):
LGTM with nits!

Thanks for working so hard on AccountsDB.

runtime/src/accounts_db.rs
carllin: oh good catch, readded!

PR #12791, opened 2020-10-10
bank: Don't bother paying 0 rent

mvines:


runtime/src/bank.rs
```diff
@@ -2676,17 +2676,19 @@ impl Bank {
...
-                self.store_account(&pubkey, &account);
-                rewards.push((
-                    pubkey,
-                    RewardInfo {
-                        reward_type: RewardType::Rent,
-                        lamports: rent_to_be_paid as i64,
-                        post_balance: account.lamports,
-                    },
-                ));
+                if rent_to_be_paid > 0 {
```
ryoqun, line 2679: @mvines I'm certain this will create testnet cluster crash because this changes some accounts are written or not, depending on validator updated or not.. testing on #12175. Also, I'll just make `enforce_fix` cover this new behavior.

ryoqun: I mean, this ultimately produces different bank hashes.

mvines: Oh really?  Wow.  Because `self.store_account(&pubkey, &account);` was skipped here?  My main goal was to skip `rewards.push(...)`, but then didn't see a reason to not skip the entire block and no test told me otherwise!

Mainnet-beta should be fine since rent is still 100% burned.   There's been rent payments on testnet though since this PR is in 1.3.16, testnet hasn't blown apart yet.  Have we just been lucky?

mvines: Oh my bad.  This PR missed v1.3.16 so it's not live on testnet/mainnet-beta yet 😅

ryoqun: > Because `self.store_account(&pubkey, &account);` was skipped here?

Yeah, I think so. Just preparing to test this locally as well.

> Oh my bad. This PR missed v1.3.16 so it's not live on testnet/mainnet-beta yet sweat_smile

:sweat_smile:

ryoqun: @mvines I've confirmed this locally and on [CI](https://buildkite.com/solana-labs/solana/builds/32596) and created a fix for this: https://github.com/solana-labs/solana/pull/12804.

Issue #9491, opened 2020-04-14
Gossip is vulnerable to UDP reflection


leoluk:
#### Problem

Gossip responses are sent to the source IP, rather than the IP address in gossip. This means that the mechanism is vulnerable to source IP spoofing, with responses sent to the spoofed IP.

The real-world impact of this ranges from "nuisance" to "very bad" depending on the amplification factor.

Spoofing is still trivial in today's internet - many providers still don't do proper filtering.

#### Proposed Solution

In a connectionless protocol, the source IP can never be trusted. The canonical solution to this is a three-way handshake that authenticates the source IP.

Issue #9492, opened 2020-04-14
Gossip pull request amplification


leoluk:
#### Problem

Some gossip responses are much larger than the request. Sending a single 1175 byte pull request packet results in a 907100 byte response in ~1k individual packets, which represents an impressive amplification factor of 772x.

Example payload: [reflect.bin.txt](https://github.com/solana-labs/solana/files/4476464/reflect.bin.txt)

This allows an attacker to cause excessive egress bandwidth costs for a validator at little cost to herself.

Combined with #9491, it can also be used to repurpose a cluster into a very effective DDoS cannon.

The large UDP send/receive buffers set by sys-tuner likely exacerbate this. After sending a burst of packets to our tds validator, it got stuck sending gossip responses at 1 Gbps until I restarted it:

![image](https://user-images.githubusercontent.com/859697/79246297-78242a00-7e79-11ea-83e6-7fd5cc28f8fc.png)

#### Proposed Solution

Zero-pad request packets to have the same size/be larger than the responses? Not sure.

behzadnouri:
For reference, we do we have an [outbound data budget](https://github.com/solana-labs/solana/blob/87311cce7/core/src/cluster_info.rs#L251) and also some ongoing work https://github.com/solana-labs/solana/pull/11807 to better allocate that budget to pull requests. but, yeah, this definitely needs more work and tuning.

PR #12794, opened 2020-10-10
implements ping-pong packets between nodes

behzadnouri:
#### Problem
https://hackerone.com/reports/991106
> It’s possible to use UDP gossip protocol to amplify DDoS attacks. An attacker can spoof IP address in UDP packet when sending PullRequest to the node. There's no any validation if provided source IP address is not spoofed and the node can send much larger PullResponse to victim's IP. As I checked, PullRequest is about 290 bytes, while PullResponse is about 10 kB. It means that amplification is about 34x. This way an attacker can easily perform DDoS attack both on Solana node and third-party server.
>
> To prevent it, need for example to implement ping-pong mechanism similar as in Ethereum: Before accepting requests from remote client needs to validate his IP. Local node sends Ping packet to the remote node and it needs to reply with Pong packet that contains hash of matching Ping packet. Content of Ping packet is unpredictable. If hash from Pong packet matches, local node can remember IP where Ping packet was sent as correct and allow further communication.
>
> More info:
> https://github.com/ethereum/devp2p/blob/master/discv4.md#endpoint-proof
> https://github.com/ethereum/devp2p/blob/master/discv4.md#wire-protocol

#### Summary of Changes
The commit adds a PingCache, which maintains records of remote nodes
which have returned a valid response to a ping message, and on-the-fly
ping messages pending a pong response from the remote node.

When handling pull-requests, those from addresses which have not passed
the ping-pong check are filtered out, and additionally ping packets are
added for addresses which need to be (re)verified.


behzadnouri:
@sakridge @carllin The code is complete except for tests. I wanted to get the initial review before adding more tests.

also @mvines since this is an abi change.

aeyakovenko:
Can you expand the problem?  the hacker one links are behind a login.  I'd like to have the details in the GitHub PR

behzadnouri:
> Can you expand the problem? the hacker one links are behind a login. I'd like to have the details in the GitHub PR

sure, done

aeyakovenko:
@behzadnouri doesn't the PullRequest/PullResponse have the same effect?  Do we need another set of messages, or can we infer the same information form the pulls?

behzadnouri:
> @behzadnouri doesn't the PullRequest/PullResponse have the same effect? Do we need another set of messages, or can we infer the same information form the pulls?

I am not sure we can. An attacker can keep sending fake pull/push traffic, without an indication that it is spoofing an ip address. With ping/pong packets, the pong package can be verified as soon as it is received.

leoluk:
That Hackerone report is a duplicate of https://github.com/solana-labs/solana/issues/9492.

Glad to see it's getting fixed!

aeyakovenko:
Yea. You age right.  We need a signed cookie to be sent back to authenticate the address.

sakridge:
Looks like a good direction to me, just need to figure out the backwards-compatibility part.

core/src/ping_pong.rs
```diff
@@ -0,0 +1,263 @@
...
+    // Time-to-live of received pong messages.
+    ttl: Duration,
+    // Timestamp of last ping message sent to a remote node.
+    // Used to rate limit pings to remote nodes.
+    pings: LruCache<(Pubkey, SocketAddr), Instant>,
+    // Verified pong responses from remote nodes.
+    pongs: LruCache<(Pubkey, SocketAddr), Instant>,
+    // Hash of ping tokens sent out to remote nodes,
+    // pending a pong response back.
+    cache: LruCache<Hash, (Pubkey, SocketAddr)>,
```
carllin, line 40: how about calling this the `pending_cache` to be more explicit?

core/src/ping_pong.rs
```diff
@@ -0,0 +1,263 @@
...
+        &mut self,
+        now: Instant,
+        node: (Pubkey, SocketAddr),
+        pingf: F,
+    ) -> (bool, Option<Ping<T>>)
+    where
+        T: Serialize,
+        F: FnMut() -> Option<Ping<T>>,
+    {
+        let (check, ping) = match self.pongs.get(&node) {
```
carllin, line 206: nit: rename ping-> should_ping to communicate it's a boolean?

core/src/ping_pong.rs
```diff
@@ -0,0 +1,263 @@
...
+                // Pop if the pong message has expired.
+                if age > self.ttl {
+                    self.pongs.pop(&node);
+                }
+                // If the pong message is not too recent, generate a new ping
+                // message to extend remote node verification.
+                (true, age > self.ttl / 8)
+            }
+        };
+        let ping = if ping {
```
carllin, line 219: nit: rename ping-> did_ping to communicate it's a boolean

core/src/cluster_info.rs
```diff
@@ -1796,15 +1858,21 @@ impl ClusterInfo {
...
-            addrs.push(pull_data.from_addr);
-        }
+        let mut packets = Packets::new_with_recycler(recycler.clone(), 64, "handle_pull_requests");
+        let (caller_and_filters, addrs): (Vec<_>, Vec<_>) = {
+            let mut rng = rand::thread_rng();
+            let check_pull_request =
+                self.check_pull_request(Instant::now(), &mut rng, &mut packets);
+            requests
+                .into_iter()
+                .filter(check_pull_request)
```
carllin, line 1872: It might be helpful for debugging to have datapoint to track the number of requests dropped here. We currently have https://github.com/solana-labs/solana/blob/master/core/src/cluster_info.rs#L1898-L1900 in this function, but this new filter isn't accounted for inside of that

core/src/cluster_info.rs
```diff
@@ -1788,6 +1812,44 @@ impl ClusterInfo {
...
+            }
+            // TODO: For backward compatibility, this unconditionally returns
+            // true for now. It has to return _check, once nodes start
+            // responding to ping messages.
+            true
+        };
+        // Because pull-responses are sent back to packet.meta.addr() of
+        // incoming pull-requests, pings are also sent to request.from_addr (as
+        // opposed to caller.gossip address).
+        move |request| {
```
carllin, line 1845: cool,  function currying!

core/src/cluster_info.rs
```diff
@@ -95,6 +96,9 @@ const MAX_GOSSIP_TRAFFIC: usize = 128_000_000 / PACKET_DATA_SIZE;
 
 /// Keep the number of snapshot hashes a node publishes under MAX_PROTOCOL_PAYLOAD_SIZE
 pub const MAX_SNAPSHOT_HASHES: usize = 16;
+const GOSSIP_PING_TOKEN_SIZE: usize = 32;
+const GOSSIP_PING_CACHE_CAPACITY: usize = 16384;
```
carllin, line 100: Should this be based on the number of nodes in gossip instead of a constant?

core/src/ping_pong.rs
carllin, line 206: nit: rename ping-> should_ping to communicate it's a boolean

carllin:
@behzadnouri looks good! Maybe we can use the pongs to do the upgrade? I.e. once a peer responds to a pong message, then we can mark that peer as upgraded and use the hard verify path in `check_pull_request` for that peer

core/src/cluster_info.rs
```diff
@@ -1788,6 +1812,44 @@ impl ClusterInfo {
...
+            let (_check, ping) = ping_cache.check(now, node, &mut pingf);
+            if let Some(ping) = ping {
+                let ping = Protocol::PingMessage(ping);
+                let ping = Packet::from_data(&node.1, ping);
+                packets.packets.push(ping);
+            }
+            // TODO: For backward compatibility, this unconditionally returns
+            // true for now. It has to return _check, once nodes start
+            // responding to ping messages.
+            true
```
mvines, line 1840: We can plumb in the final code here as well, making it conditional on a new feature added to `runtime/src/feature_set.rs`.   This way we can dynamically activate this feature on clusters once 95% of the validator fleet has upgraded.   Right now `FeatureSet` is not plumbed though cluster_info though, so I suggest adding this plumbing first as a forerunner PR to this one.  `core/` should probably use the `FeatureSet` from the highest rooted bank.

metrics/src/counter.rs
```diff
@@ -76,7 +76,7 @@ macro_rules! inc_counter_info {
 #[macro_export]
 macro_rules! inc_new_counter {
     ($name:expr, $count:expr, $level:expr, $lograte:expr, $metricsrate:expr) => {{
-        if log_enabled!($level) {
+        if $count != 0 && log_enabled!($level) {
```
mvines, line 79: nit: separate out little changes like this into their own PRs.  They can land independently, and sooner

core/src/cluster_info.rs
```diff
@@ -95,6 +96,9 @@ const MAX_GOSSIP_TRAFFIC: usize = 128_000_000 / PACKET_DATA_SIZE;
 
 /// Keep the number of snapshot hashes a node publishes under MAX_PROTOCOL_PAYLOAD_SIZE
 pub const MAX_SNAPSHOT_HASHES: usize = 16;
+const GOSSIP_PING_TOKEN_SIZE: usize = 32;
```
mvines, line 99: A comment would be nice explaining why 32 hre

core/src/cluster_info.rs
behzadnouri: yes, that is a good idea. added a new stats counter here.

core/src/ping_pong.rs
behzadnouri: done

core/src/ping_pong.rs
behzadnouri: done

metrics/src/counter.rs
behzadnouri: I actually reverted this change and instead put the `if ... != 0` at the call-site.
Though the counter value does not change when event count is zero, some of the logging logic in the implementation relies on how often that function is called regardless of events counts being zero or not:
https://github.com/solana-labs/solana/blob/f5ed017d6/metrics/src/counter.rs#L172-L203
It might be somewhere in the code relies on that behavior.

core/src/cluster_info.rs
behzadnouri: sure, working on it.

brianlong:
This looks like a great feature. Are you also retaining the round-trip ping-pong times? That data can help operators to trouble-shoot networking problems.

core/src/cluster_info.rs
carllin: @behzadnouri, got it, thanks! I think the most convenient available way would be to use `cluster_info.gossip_peers()`. It's not the most efficient because it scans the table (broadcast does something similar with `cluster_info.tvu_peers()`) but may not be too bad if done periodically.

VnceB:
wouldn't that also fix #9491 ?

leoluk:
Yes, it solves #9491, but not #9492.

core/src/cluster_info.rs
behzadnouri: thanks, since the alternatives would require table scan and locking, i lean toward keeping this constant for now, and later test alternatives.

core/src/cluster_info.rs
behzadnouri: updated the diff to use the feature-set added in https://github.com/solana-labs/solana/pull/12878

behzadnouri:
> This looks like a great feature. Are you also retaining the round-trip ping-pong times? That data can help operators to trouble-shoot networking problems.

@brianlong The code is retaining the time the pings were sent out but for the purpose of rate-limiting. but, agreed, we can also easily get the round-trip time to measure latencies. Once this is rolled out and if we do not hit any backward compatibility issues I can experiment with that.

behzadnouri:
> @behzadnouri looks good! Maybe we can use the pongs to do the upgrade? I.e. once a peer responds to a pong message, then we can mark that peer as upgraded and use the hard verify path in `check_pull_request` for that peer

Per [Michael's comment](https://github.com/solana-labs/solana/pull/12794#discussion_r503683449), I have added a new feature-id https://github.com/solana-labs/solana/pull/12878 and will use that to dynamically turn on the feature.  We can wait for both 95% of fleet to upgrade, and also once the newly added metric `pull_request_ping_pong_check_failed_count` looks fine.

behzadnouri:
> @sakridge @carllin The code is complete except for tests. I wanted to get the initial review before adding more tests.
> 
> also @mvines since this is an abi change.

I have added more test coverage. Should be ready for full review.

core/src/ping_pong.rs
```diff
@@ -0,0 +1,401 @@
...
+            self.maybe_ping(now, node, pingf)
+        } else {
+            None
+        };
+        (check, ping)
+    }
+}
+
+impl Clone for PingCache {
+    fn clone(&self) -> Self {
```
sakridge, line 229: Is it just for tests?

core/src/cluster_info.rs
```diff
@@ -97,6 +98,10 @@ const MAX_GOSSIP_TRAFFIC: usize = 128_000_000 / PACKET_DATA_SIZE;
 
 /// Keep the number of snapshot hashes a node publishes under MAX_PROTOCOL_PAYLOAD_SIZE
 pub const MAX_SNAPSHOT_HASHES: usize = 16;
+/// Number of bytes in the randomly generated token sent with ping messages.
+const GOSSIP_PING_TOKEN_SIZE: usize = 32;
```
sakridge, line 102: Just 32-bytes seems like overkill. What about just u64 which seems like enough and allows for better memory efficiency and easy atomic use if needed, less cycles for compare/randomization generation etc.

core/src/ping_pong.rs
behzadnouri: This is needed for `ClusterInfo::clone_with_id` function:
https://github.com/solana-labs/solana/blob/e666272ab/core/src/cluster_info.rs#L477-L499
which is only for tests and simulations.

I have to implement this manually because `LruCache` does not implement `Clone`.

core/src/cluster_info.rs
behzadnouri: I am worried that will compromise security in some way, and limit usability of these tokens in the future for other security related applications beyond the current one (i.e. ip-spoofing prevention). For example responding with sha256 hash of ping token which itself was only 64 bits is kind of degenerate.

These ping/pong packets will be super low traffic, since once received they will be cached for a while. So I doubt that this will cause any noticeable performance difference.

sakridge (dismissed):
lgtm

core/src/ping_pong.rs
behzadnouri: I removed the `Clone` implementation in favor of a separate `mock_clone` function which is more explicit and emphasizes that it is only for tests and simulations. Also, updated the clone code because it has to traverse the iterators in reverse order to maintain the original order of insertions.

PR #12796, opened 2020-10-11
client: simulate_transaction_with_config() now passes full config to server

### LINKED ISSUE #12795: RpcClient should expose the `commitment` param for `simulate_transaction`
#### Problem

Currently the `simulate_transaction` method on the `RpcClient` doesn't allow for passing in the `commitment` parameter. This leads to simulated transactions failing when they rely on previous transactions that use `singleGossip` or `single`.

https://github.com/solana-labs/solana/blob/master/client/src/rpc_client.rs#L183

#### Proposed Solution

Add the ability to pass in the commitment field.

mvines:
Fixes #12795

@armaniferrante - would you mind confirming this PR works for you instead of https://github.com/project-serum/serum-dex/blob/89ac8ea77b2f0310156578e1ec52add3a7e03e79/common/src/client/rpc.rs#L201?

### DIFF
--- a/client/src/rpc_client.rs
+++ b/client/src/rpc_client.rs
@@ -161,26 +161,20 @@ impl RpcClient {
     pub fn simulate_transaction(
         &self,
         transaction: &Transaction,
-        sig_verify: bool,
     ) -> RpcResult<RpcSimulateTransactionResult> {
-        self.simulate_transaction_with_config(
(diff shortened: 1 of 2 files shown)


mvines:
Blocked by #12797

armaniferrante:
Works on my end.

PR #12804, opened 2020-10-12
Move no-0-rent rent dist. behavior under feature

ryoqun:
#### Problem

#12791 introduces differing behavior.

#### Summary of Changes

Properly gate it. Fortunately, we have still not enabled runtime feature (`no_overflow_rent_distribution`); let's free-ride on it

Fixes #12791


mvines:
Thank you @ryoqun, this was a close one!

PR #12826, opened 2020-10-13
`solana vote-account`/`solana stake-account` now works with RPC servers without `--enable-rpc-transaction-history`

mvines:
Vote/stake epoch rewards are only fetchable from RPC servers with `--enable-rpc-transaction-history` enabled.  Cleanly handle when this is not the case

Partial fix for #12843 (the cli failure)

### DIFF
--- a/cli/src/stake.rs
+++ b/cli/src/stake.rs
@@ -23,8 +23,12 @@ use solana_cli_output::{
     CliStakeType,
 };
 use solana_client::{
-    blockhash_query::BlockhashQuery, nonce_utils, rpc_client::RpcClient,
-    rpc_request::DELINQUENT_VALIDATOR_SLOT_DISTANCE,
+    blockhash_query::BlockhashQuery,
+    client_error::{ClientError, ClientErrorKind},
+    nonce_utils,
+    rpc_client::RpcClient,
+    rpc_custom_error,
+    rpc_request::{self, DELINQUENT_VALIDATOR_SLOT_DISTANCE},
 };
 use solana_remote_wallet::remote_wallet::RemoteWalletManager;
 use solana_sdk::{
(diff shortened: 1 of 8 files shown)


cli/src/stake.rs
```diff
@@ -1605,10 +1609,26 @@ pub(crate) fn fetch_epoch_rewards(
...
+            Err(ClientError {
+                kind:
+                    ClientErrorKind::RpcError(rpc_request::RpcError::RpcResponseError {
+                        code: rpc_custom_error::JSON_RPC_SERVER_ERROR_BLOCK_NOT_AVAILABLE,
+                        message: _,
+                    }),
+                request: _,
+            }) => {
+                // RPC node doesn't have this block
+                break;
```
mvines, line 1626: All the work in this PR is just to enable a `break` here!

client/src/http_sender.rs
```diff
@@ -27,6 +27,13 @@ impl HttpSender {
     }
 }
 
+#[derive(Deserialize, Debug)]
+struct RpcErrorObject {
+    code: i64,
+    message: String,
+    /*data field omitted*/
```
mvines, line 34: One day it'd be cool to return a full `RpcCustomError` object here.  But for now I'm content with just having access to the error `code`

client/src/rpc_custom_error.rs
```diff
@@ -1,13 +1,15 @@
...
-const JSON_RPC_SERVER_ERROR_3: i64 = -32003;
-const JSON_RPC_SERVER_ERROR_4: i64 = -32004;
-const JSON_RPC_SERVER_ERROR_5: i64 = -32005;
-const JSON_RPC_SERVER_ERROR_6: i64 = -32006;
+pub const JSON_RPC_SERVER_ERROR_BLOCK_CLEANED_UP: i64 = -32001;
+pub const JSON_RPC_SERVER_ERROR_SEND_TRANSACTION_PREFLIGHT_FAILURE: i64 = -32002;
+pub const JSON_RPC_SERVER_ERROR_TRANSACTION_SIGNATURE_VERIFICATION_FAILURE: i64 = -32003;
+pub const JSON_RPC_SERVER_ERROR_BLOCK_NOT_AVAILABLE: i64 = -32004;
+pub const JSON_RPC_SERVER_ERROR_NODE_UNHEALTHLY: i64 = -32005;
+pub const JSON_RPC_SERVER_ERROR_TRANSACTION_PRECOMPILE_VERIFICATION_FAILURE: i64 = -32006;
```
CriesofCarrots, line 12: 😍

CriesofCarrots (approved):

PR #12840, opened 2020-10-13
terminology update, nonce to bump seed

jackcmay:
#### Problem

When deriving program addresses an extra seed may be required to push the address off the curve.  The extra seed is called a "nonce" but it's used more than once.

#### Summary of Changes

Call it a "bump seed" instead

Fixes #


jackcmay:
@mvines fyi

mvines (approved):

PR #12852, opened 2020-10-13
Calculate accounts hash async in accounts background service

sakridge:
#### Problem

Full accounts hash is computed synchronously in bank_forks::set_root which blocks other replay processing and voting.

#### Summary of Changes

Calculate hash async in accounts background service. Use the max_root parameter such that a potential future root doesn't affect the result of the calculation.

Fixes #

### DIFF
--- a/runtime/src/accounts_background_service.rs
+++ b/runtime/src/accounts_background_service.rs
@@ -51,6 +51,10 @@ impl SnapshotRequestHandler {
                     status_cache_slot_deltas,
                 } = snapshot_request;
 
+                let mut hash_time = Measure::start("hash_time");
+                snapshot_root_bank.update_accounts_hash();
+                hash_time.stop();
+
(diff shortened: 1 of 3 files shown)


ryoqun:
> Use the max_root parameter such that a potential future root doesn't affect the result of the calculation.

Cool, a nice side effect of @carllin's #12194

carllin (approved):

carllin:
Would be nice to run this on a validator connected to mainnet for a while to make sure nothing unexpected happens.

sakridge:
> Would be nice to run this on a validator connected to mainnet for a while to make sure nothing unexpected happens.

I ran it for about 5 hours yesterday with the `--trusted-validator` flags on and `--halt-on-trusted-validators-accounts-hash-mismatch` with mainnet and I was seeing that all the hashes are matching. I will run it for a couple days to be sure.

PR #12855, opened 2020-10-13
Don't report RewardType::Fee when none was awarded

mvines:
`solana block` on mainnet-beta is reporting unhelpful rewards such as:
```
Rewards:
  Address                                            Type        Amount           New Balance           Percent Change
  Awes4Tr6TX8JDzEhCZY2QVNimT6iD1zWHzf1vNyGvpLM        fee        ◎0.000000000     ◎314.357516520          0.000000000%
```
If there's no unburned fee, skip posting the fee reward

### DIFF
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -1484,14 +1484,16 @@ impl Bank {
             );
 
             let post_balance = self.deposit(&self.collector_id, unburned);
-            self.rewards.write().unwrap().push((
-                self.collector_id,
-                RewardInfo {
-                    reward_type: RewardType::Fee,
-                    lamports: unburned as i64,
-                    post_balance,
-                },
-            ));
+            if unburned != 0 {
+                self.rewards.write().unwrap().push((
+                    self.collector_id,
+                    RewardInfo {
+                        reward_type: RewardType::Fee,
+                        lamports: unburned as i64,
+                        post_balance,
+                    },
+                ));
+            }
             self.capitalization.fetch_sub(burned, Relaxed);
         }
     }


mvines:
Special review request from @ryoqun to ensure I don't accidentally almost break consensus a second time in under a week!

runtime/src/bank.rs
```diff
@@ -1484,14 +1484,16 @@ impl Bank {
...
-            self.rewards.write().unwrap().push((
-                self.collector_id,
-                RewardInfo {
-                    reward_type: RewardType::Fee,
-                    lamports: unburned as i64,
-                    post_balance,
-                },
-            ));
+            if unburned != 0 {
+                self.rewards.write().unwrap().push((
```
ryoqun, line 1488: The change of behavior shouldn't cause any problem for cluster consensus.

But how about defining a method like `fn {record_positive_reward|record_reward_if_any}(&mut rewards, reward_type, pubkey, lamports, post_balance)`? The behavior of skipping to push for 0-lamport rewards seems quite generic and solid invariant tied to `Bank::rewards`. The same code is repeated thrice in `collect_fees`, `pay_validator_rewards` and  `distribute_rent_to_validators`.

ryoqun:
(Also, I just remembered all of fee rewards are currently 100% burned on testnet and mainnet-beta.)

runtime/src/bank.rs
mvines: Nice idea. I think that trying to create that generic function will not work out well, each is slightly different so I think I'd rather just leave it as is.

ryoqun (approved):
LGTM!

PR #12859, opened 2020-10-13
Add log_messages to proto file

CriesofCarrots:
#### Problem
Rebuilding `storage-bigtable/src/confirmed_block.proto` breaks `generated::TransactionStatusMeta` because it is regenerated without the `log_messages` field. `log_messages` was added into proto/solana.bigtable.confirmed_block.rs by hand. 

#### Summary of Changes
Add `log_messages` to confirmed_block.proto


oJshua (approved):
Thanks for looping me in on this @CriesofCarrots. So that's a generated file? edit: just read about it, makes sense.

Issue #12843, opened 2020-10-13
solana [stake-account|vote-account] returns an error: Error: invalid type: null, expected struct EncodedConfirmedBlock


max-block:
```
$ solana vote-account /path/to/vote.json
Error: invalid type: null, expected struct EncodedConfirmedBlock
```

solana stake-account returns the same error

solana --version
solana-cli 1.3.16 (src:b31ec057; feat:2415322184)

CriesofCarrots:
Hi @max-block ; what cluster are you targeting? Are you willing to share the pubkey of `/path/to/vote.json`?

max-block:
@CriesofCarrots testnet TDS, 95xoBAM5kHDANac3rim6ka1pUwrtVKFLdwzVLVTf8hZK

max-block:
I've upgraded to v1.3.17, but the problem is still present.

CriesofCarrots:
@max-block , you should find that `solana vote-account` works for you now.
@ericlwilliams , `block 40268260` is currently in limbo... it's been pruned from blockstore on the api nodes, but hasn't made it into bigtable from the warehouse node yet. However, the other block you were looking at, `block 40700256`, now works again.

ericlwilliams:
Thanks @CriesofCarrots !

CriesofCarrots:
Issue was caused by a bug in deserialization of legacy Reward data from rocksdb.
Consider this resolved: Long-term solution is at #12860 . Any remaining data gaps (like the block mentioned above) should be resolved as the testnet warehouse node writes these blocks/rewards into bigtable.

PR #12860, opened 2020-10-13
Convert Blockstore Rewards cf to protobuf

CriesofCarrots:
#### Problem
Extending the Reward struct breaks deserialization from Blockstore, even with the `default_on_eof` deserializer. This is because the data is stored as a Vec, and bincode doesn't distinguish between the extended field and the start of the next Vec item.

#### Summary of Changes
- Add a ProtobufColumn trait to blockstore; get_protobuf_or_bincode method supports falling back to bincode deserialization for historical data
- impl Rewards as a ProtobufColumn

Depends on #12859 
Prevents #12843 on future Reward fields


CriesofCarrots:
@mvines preview of how this is shaping up. I'm pleased with how light the changes need to be.

mvines:
I like the direction.  The prospect of re-using parts of `confirmed_block.proto` for both rocksdb and bigtable is nice too

CriesofCarrots:
The protobuf-encoded Reward struct is smaller than bincode: ~60bytes vs ~70bytes
Decoding protobuf is slower than deserializing bincode, but encoding protobuf is slightly faster than serializing bincode:

For a Rewards of len = 100
```
test bench_read_bincode             ... bench:       9,546 ns/iter (+/- 2,562)
test bench_read_protobuf            ... bench:      16,518 ns/iter (+/- 1,586)
test bench_serialize_write_bincode  ... bench:      31,716 ns/iter (+/- 65,080)
test bench_serialize_write_protobuf ... bench:      28,598 ns/iter (+/- 53,521)
```

CriesofCarrots:
@mvines this is feeling pretty solid. Can you please take a look?
Do you have any concerns about the slower decoding?

storage-bigtable/src/confirmed_block.proto
```diff
@@ -73,6 +73,10 @@ message Reward {
     RewardType reward_type = 4;
 }
 
+message Rewards {
```
CriesofCarrots, line 76: Thoughts on creating a `solana-proto` crate that both `solana-ledger` and `solana-storage-bigtable` would depend on?
It was easy not to, since `solana-ledger` already depends on `solana-storage-bigtable`

sakridge:
make it faster: :-) https://capnproto.org

storage-bigtable/src/confirmed_block.proto
mvines: `solana-ledger` depending on `solana-storage-bigtable` for rocksdb storage doesn't feel great.

`solana-ledger-proto` or `solana-storage-proto` maybe?  `solana-proto` feels too generic.   Then `solana-*-capnproto` slides in nicely later

CriesofCarrots: Incidentally, I'm happy to give capnproto a go for this, but as you pointed out, the rocksdb and bigtable reuse seems nice...

mvines: IMO let's go for reuse here.  If capnproto works out for nicely for program instruction/account serde then we can potentially use it here as well if we're really concerned about the additional cost that protobuf brings

CriesofCarrots: @mvines how does this look?

mvines (approved):

PR #12875, opened 2020-10-14
Better tower logs for SwitchForkDecision and etc

ryoqun:
#### Problem

When writing test for #12350, debugging was so painful that I've added some logs here and there.

#### Summary of Changes

All of this is just some logging improvements. I separated this form the original pr to make logically separated prs independent.

There is no functional change.

Fixes #

### DIFF
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -530,15 +530,34 @@ impl Tower {
                             // all of them.
                             panic!("no ancestors found with slot: {}", last_voted_slot);
                         } else {
-                            // compare slots not to error! just because of newer snapshots
+                            // This condition shouldn't occur under normal validator operation, indicating
+                            // something unusual happened.
+                            // Possible causes include: OS/HW crash, validator process crash, only saved tower
+                            // is moved over to a new setup, etc...
+
+                            // However, returning empty ancestors as a fallback here shouldn't result in
+                            // slashing by itself (Note that we couldn't fully preclude any kind of slashing if
+                            // the failure was OS or HW level).
+
+                            // Firstly, lockout is ensured elsewhere.
+
+                            // Also, there is no risk of optimistic conf. violation. Although empty ancestors
+                            // could result in incorrect (= more than actual) locked_out_stake and
+                            // false-positive SwitchProof later in this function, there should be no such a
+                            // heavier fork candidate, first of all, if the last vote (or any of its
+                            // unavailable ancestors) were already optimistically confirmed.
+                            // The only exception is that other validator is already violating it...
                             if self.is_first_switch_check() && switch_slot < last_voted_slot {
-                                error!(
+                                // `switch < last` is needed not to warn! this message just because of using
+                                // newer snapshots on validator restart
+                                let message = format!(
                                   "bank_forks doesn't have corresponding data for the stray restored \
                                    last vote({}), meaning some inconsistency between saved tower and ledger.",
                                   last_voted_slot
                                 );
+                                warn!("{}", message);
+                                datapoint_warn!("tower_warn", ("warn", message, String));
                             }
-                            // (system crash? process crash? newer snapshot, or only a saved tower is moved over to new setup?)
                             &empty_ancestors
                         }
                     });
@@ -614,7 +633,7 @@ impl Tower {
                             }
 
                             // Only count lockouts on slots that are:
-                            // 1) Not ancestors of `last_vote`
+                            // 1) Not ancestors of `last_vote`, meaning being on different fork
                             // 2) Not from before the current root as we can't determine if
                             // anything before the root was an ancestor of `last_vote` or not
                             if !last_vote_ancestors.contains(lockout_interval_start)
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -30,11 +30,11 @@ use std::{
 };
 use thiserror::Error;
 
-#[derive(PartialEq, Clone, Debug)]
+#[derive(PartialEq, Clone, Debug, AbiExample)]
 pub enum SwitchForkDecision {
     SwitchProof(Hash),
-    NoSwitch,
-    FailedSwitchThreshold,
+    SameFork,
+    FailedSwitchThreshold(u64, u64),
 }
 
 impl SwitchForkDecision {
@@ -45,8 +45,11 @@ impl SwitchForkDecision {
         authorized_voter_pubkey: &Pubkey,
     ) -> Option<Instruction> {
         match self {
-            SwitchForkDecision::FailedSwitchThreshold => None,
-            SwitchForkDecision::NoSwitch => Some(vote_instruction::vote(
+            SwitchForkDecision::FailedSwitchThreshold(_, total_stake) => {
+                assert_ne!(*total_stake, 0);
+                None
+            }
+            SwitchForkDecision::SameFork => Some(vote_instruction::vote(
                 vote_account_pubkey,
                 authorized_voter_pubkey,
                 vote,
@@ -60,6 +63,10 @@ impl SwitchForkDecision {
                 ))
             }
         }
+    }
+
+    pub fn can_vote(&self) -> bool {
+        !matches!(self, SwitchForkDecision::FailedSwitchThreshold(_, _))
     }
 }
 
@@ -101,6 +108,8 @@ pub struct Tower {
     // This could be emptied after some time; but left intact indefinitely for easier
     // implementation
     stray_restored_slot: Option<Slot>,
+    #[serde(skip)]
+    pub last_switch_threshold_check: Option<(Slot, SwitchForkDecision)>,
 }
 
 impl Default for Tower {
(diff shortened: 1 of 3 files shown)


core/src/consensus.rs
```diff
@@ -30,11 +30,11 @@ use std::{
 };
 use thiserror::Error;
 
-#[derive(PartialEq, Clone, Debug)]
+#[derive(PartialEq, Clone, Debug, AbiExample)]
 pub enum SwitchForkDecision {
     SwitchProof(Hash),
-    NoSwitch,
```
ryoqun, line 36: I've always been bothered with this naming `NoSwitch` everytime I see this on logs like this:

- At the first glance, it sounds something bad happened due to the connotation of `No`. (it's not negating errorous thing like `Error` as in `NoError`) But this is in fact pretty normal enum variant.
- Then, my next wondering is: If there isn't something called `Switch`, what is this exactly?

So, I'm proposing here to rename to `SameFork`:

- `Same` sounds good thing or boring/usual thing at least.
- It doesn't introduce new term; it borrows the `Fork` term in the enum name. So this doesn't introduce any additional cognitive load.
 - `SameFork` adequately implies we're voting on same fork without `Switch` if the reader has basic understaing of validator's voting behavior. 
- From my understanding, I think `SameFork` correctly represents what this variant is meant for.


@carllin Time for bike shedding? ;)

core/src/consensus.rs
```diff
@@ -30,11 +30,11 @@ use std::{
...
-    FailedSwitchThreshold,
+    SameFork,
+    FailedSwitchThreshold(u64, u64),
```
ryoqun, line 37: let's start passing around locked_out_stake and total_stake for logging. I found this very useful to debug.

core/src/consensus.rs
```diff
@@ -520,9 +530,16 @@ impl Tower {
                             // all of them.
                             panic!("no ancestors found with slot: {}", last_voted_slot);
                         } else {
-                            // bank_forks doesn't have corresponding data for the stray restored last vote,
-                            // meaning some inconsistency between saved tower and ledger.
-                            // (newer snapshot, or only a saved tower is moved over to new setup?)
+                            // compare slots not to error! just because of newer snapshots
+                            if self.is_first_switch_check() && switch_slot < last_voted_slot {
+                                error!(
```
ryoqun, line 535: The reason I didn't `error!` here at #10718 is that doing so printed so many duplicate errors because `check_switch_threshold` is called frequently. Thanks to `is_first_switch_check()` (and `last_switch_threshold_check` behind it) this isn't true anymore.

core/src/consensus.rs
```diff
@@ -45,8 +45,11 @@ impl SwitchForkDecision {
         authorized_voter_pubkey: &Pubkey,
     ) -> Option<Instruction> {
         match self {
-            SwitchForkDecision::FailedSwitchThreshold => None,
-            SwitchForkDecision::NoSwitch => Some(vote_instruction::vote(
+            SwitchForkDecision::FailedSwitchThreshold(_, total_stake) => {
+                assert_ne!(*total_stake, 0);
```
ryoqun, line 49: place some sanity check as a bonus.

core/src/consensus.rs
```diff
@@ -622,10 +639,44 @@ impl Tower {
...
+            switch_slot,
+            ancestors,
+            descendants,
+            progress,
+            total_stake,
+            epoch_vote_accounts,
+        );
+        let new_check = Some((switch_slot, decision.clone()));
+        if new_check != self.last_switch_threshold_check {
+            trace!(
```
ryoqun, line 667: I just wanted to `trace!` this. But because of very frequent call of `check_switch_threshold` with same arguments with same check result, I have to implement this very simple cache for old `new_check`.

core/src/consensus.rs
```diff
@@ -932,9 +983,9 @@ impl Tower {
             self.lockouts = vote_state;
             self.do_initialize_lockouts(root, |v| v.slot > root);
             trace!(
-                "{} lockouts initialized to {:?}",
+                "Lockouts in tower for {} is initialized using bank {}",
                 self.node_pubkey,
-                self.lockouts
```
ryoqun, line 937: Currently, we are `{:?}`-ing a whole Tower once at validator startup later with `info!` to ease any persistent tower related debuggng.

core/src/consensus.rs
carllin: Sounds fine by me!

core/src/consensus.rs
ryoqun: Also note that, this assumes there is no way to control `total_stake` remotely. So there should be no risk of dos.

ryoqun:
@carllin Could you review this? (Cool, it seems that you've always reading this pr..) This blocks #12671 and #12350. So I rather want to merge this quickly. :)

core/src/consensus.rs
```diff
@@ -520,9 +530,16 @@ impl Tower {
                             // all of them.
                             panic!("no ancestors found with slot: {}", last_voted_slot);
                         } else {
-                            // bank_forks doesn't have corresponding data for the stray restored last vote,
-                            // meaning some inconsistency between saved tower and ledger.
-                            // (newer snapshot, or only a saved tower is moved over to new setup?)
+                            // compare slots not to error! just because of newer snapshots
+                            if self.is_first_switch_check() && switch_slot < last_voted_slot {
```
carllin, line 534: hmmm will `switch_slot < last_voted_slot` trigger if there's a heavier fork at a lesser slot? I.e. something that looks like:

```
                                       5 (last vote)
                                    /
0 - 1 (snapshot root)
                                  \
                                      2 (new heaviest fork)
```


This might happen because `check_switch_threshold` is called here: https://github.com/solana-labs/solana/blob/master/core/src/replay_stage.rs#L1545-L1553 even if we are locked out detected here: https://github.com/solana-labs/solana/blob/master/core/src/replay_stage.rs#L1578

core/src/consensus.rs
```diff
@@ -493,7 +503,7 @@ impl Tower {
         false
     }
 
-    pub(crate) fn check_switch_threshold(
+    fn calculate_check_switch_threshold(
```
carllin, line 506: maybe rename to `make_check_switch_threshold_decision`?

core/src/consensus.rs
ryoqun: @carllin yeah that condition will trigger. Even under that condition, I think we should do `error!` a log (note: not `panic!`) because it's still a bad condition where there is no data in bank_forks for switch_slot (this is in the fallback closure for unwrap_or_else). So, I'm saying that this case isn't a false positive. What I wanted to avoid is a different false positive where we're resuming from snapshot (=~ `switch_slot`) which is newer than  or equal to `last_vote`.

Also, fyi, this `else` clause is entered only if we haven't any single vote since restart and make_check_switch_threshold_decision is called for the first time.

ryoqun: oh, you're saying that we should avoid `error!` here if we're locked out under that situation because there is no possibility of bad voting? So, are you thinking this is kind of a false positive in fact?

Well, regarding for that thinking, that's the reason I made this `error!` not `panic!`.

carllin: @ryoqun yeah, was mostly just concerned about the false positive.

How about lowering to a `warn` and adding a comment?

ryoqun: @carllin done accordingly: b51b9daf67. How does that look? I also added `datapoint_warn` to monitor for fun. :)

carllin (approved):

core/src/consensus.rs
```diff
@@ -520,9 +530,34 @@ impl Tower {
                             // all of them.
                             panic!("no ancestors found with slot: {}", last_voted_slot);
                         } else {
-                            // bank_forks doesn't have corresponding data for the stray restored last vote,
-                            // meaning some inconsistency between saved tower and ledger.
-                            // (newer snapshot, or only a saved tower is moved over to new setup?)
+                            // This condition shouldn't occur under normal validator operation, indicating
+                            // something unusual happened.
```
ryoqun, line 534: For the record, this is the well-written proof of this:


> pleb:
> This is my line of reasoning/understanding of why that log shouldn't ever trigger under normal operation.
> 
>  https://github.com/solana-labs/solana/pull/12875/files#diff-0b654209e209924bfc70e8dd4c84d0979e36b6ca76e85a5899ba4572521a8170R550, so here: last_voted_slot is a stray with no ancestors.
> 
> So assume last_voted_slot is a stray with no ancestors. Assume that when the vote for last_voted_slot was made, the snapshot was S_old. So when we restart from a snapshot S:
> 
> 1) S we start from is > last_voted_slot. But then the warning added in that PR will not trigger because switch_slot > S > last_voted_slot
> 2) If the snapshot S <= last_voted_slot then 
>          a) if S >= S_old is the latest snapshot across all restarts, then you're guaranteed to have ancestors (unless something went wrong) because last_voted_slot is descended from S and the ledger should contain all the descendants of S. 
>          b) If S < S_old, the do_adjust_lockouts_after_replay() should panic because it can't find any anchored slots
> 
> So in all cases under normal operation, that log shouldn't tigger, is that right?


And this proof is correct as far as I peer-reviewed this.

PR #12878, opened 2020-10-14
passes through feature-set to gossip requests handling

behzadnouri:
#### Problem
For https://github.com/solana-labs/solana/pull/12794 we want to activate ping/pong check of addresses dynamically once super majority of nodes have updated to the new code. see: https://github.com/solana-labs/solana/pull/12794#discussion_r503683449

#### Summary of Changes
The commit adds a new feature id for ping/pong check and passes through the feature-set construct down to request handler.

### DIFF
--- a/core/src/cluster_info.rs
+++ b/core/src/cluster_info.rs
@@ -52,6 +52,7 @@ use solana_perf::packet::{
 };
 use solana_rayon_threadlimit::get_thread_count;
 use solana_runtime::bank_forks::BankForks;
+use solana_runtime::feature_set::{self, FeatureSet};
 use solana_sdk::hash::Hash;
 use solana_sdk::{
     clock::{Slot, DEFAULT_MS_PER_SLOT, DEFAULT_SLOTS_PER_EPOCH},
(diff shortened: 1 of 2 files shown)


core/src/cluster_info.rs
```diff
@@ -2171,13 +2180,22 @@ impl ClusterInfo {
         }
 
         let (stakes, epoch_time_ms) = Self::get_stakes_and_epoch_time(bank_forks);
-
+        let feature_set = bank_forks.map(|bank_forks| {
+            bank_forks
+                .read()
+                .unwrap()
+                .working_bank()
```
mvines, line 2187: It'll be better to use `.root_bank()`.  With the `.working_bank()`, we could see a feature get enabled and then disabled again (if the feature happens to get enabled in a minority fork).    Otherwise lgtm!

mvines (dismissed):

core/src/cluster_info.rs
behzadnouri: done.

There are other places in this file which are using `.working_bank` (e.g. for stakes). Do you think they should also use `.root_bank` instead?

mvines: I'd say so, but only for consistency.  It looks like it shouldn't matter in practice either way though

PR #12885, opened 2020-10-14
Re-use accounts stores

sakridge:
#### Problem

File/mmap creation can be expensive (100s of ms) and mmap deletion causes a flush to disk which increases unnecessary writes to disk when that data is not actually needed.

#### Summary of Changes

Save stores for re-use later.

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -4535,6 +4535,19 @@ pub mod tests {
     }
 
     #[test]
+    fn test_storage_finder() {
+        solana_logger::setup();
+        let db = AccountsDB::new_sized(Vec::new(), 16 * 1024);
+        let key = Pubkey::new_rand();
+        let lamports = 100;
+        let data_len = 8190;
+        let account = Account::new(lamports, data_len, &Pubkey::new_rand());
+        // pre-populate with a smaller empty store
+        db.create_and_insert_store(1, 8192);
+        db.store(1, &[(&key, &account)]);
+    }
+
+    #[test]
     fn test_get_snapshot_storages_empty() {
         let db = AccountsDB::new(Vec::new(), &ClusterType::Development);
         assert!(db.get_snapshot_storages(0).is_empty());
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -56,6 +56,7 @@ const PAGE_SIZE: u64 = 4 * 1024;
 pub const DEFAULT_FILE_SIZE: u64 = PAGE_SIZE * 1024;
 pub const DEFAULT_NUM_THREADS: u32 = 8;
 pub const DEFAULT_NUM_DIRS: u32 = 4;
+const MAX_RECYCLE_STORES: usize = 5000;
 
 lazy_static! {
     // FROZEN_ACCOUNT_PANIC is used to signal local_cluster that an AccountsDB panic has occurred,
@@ -172,7 +173,7 @@ pub enum BankHashVerificationError {
 /// Persistent storage structure holding the accounts
 #[derive(Debug)]
 pub struct AccountStorageEntry {
-    pub(crate) id: AppendVecId,
+    pub(crate) id: AtomicUsize,
 
     pub(crate) slot: AtomicU64,
 
@@ -202,7 +203,7 @@ impl AccountStorageEntry {
         let accounts = AppendVec::new(&path, true, file_size as usize);
 
         Self {
-            id,
+            id: AtomicUsize::new(id),
             slot: AtomicU64::new(slot),
             accounts,
             count_and_status: RwLock::new((0, AccountStorageStatus::Available)),
@@ -212,7 +213,7 @@ impl AccountStorageEntry {
 
     pub(crate) fn new_empty_map(id: AppendVecId, accounts_current_len: usize) -> Self {
         Self {
-            id,
+            id: AtomicUsize::new(id),
             slot: AtomicU64::new(0),
             accounts: AppendVec::new_empty_map(accounts_current_len),
             count_and_status: RwLock::new((0, AccountStorageStatus::Available)),
@@ -241,11 +242,12 @@ impl AccountStorageEntry {
         *count_and_status = (count, status);
     }
 
-    pub fn recycle(&self, slot: Slot) {
+    pub fn recycle(&self, slot: Slot, id: usize) {
         let mut count_and_status = self.count_and_status.write().unwrap();
         self.accounts.reset();
         *count_and_status = (0, AccountStorageStatus::Available);
         self.slot.store(slot, Ordering::Release);
+        self.id.store(id, Ordering::Relaxed);
         self.approx_store_count.store(0, Ordering::Relaxed);
     }
 
@@ -270,7 +272,7 @@ impl AccountStorageEntry {
     }
 
     pub fn append_vec_id(&self) -> AppendVecId {
-        self.id
+        self.id.load(Ordering::Relaxed)
     }
 
     pub fn flush(&self) -> Result<(), IOError> {
@@ -330,7 +332,7 @@ impl AccountStorageEntry {
             count > 0,
             "double remove of account in slot: {}/store: {}!!",
             self.slot(),
-            self.id
+            self.append_vec_id(),
         );
 
         count -= 1;
@@ -1020,7 +1022,7 @@ impl AccountsDB {
                             account.clone_account(),
                             *account.hash,
                             next - start,
-                            (store.id, account.offset),
+                            (store.append_vec_id(), account.offset),
                             account.meta.write_version,
                         ));
                         start = next;
@@ -1066,7 +1068,7 @@ impl AccountsDB {
                 },
             )
             .sum();
-        let aligned_total: u64 = (alive_total + (PAGE_SIZE - 1)) & !(PAGE_SIZE - 1);
+        let aligned_total: u64 = self.page_align(alive_total);
 
         debug!(
             "shrinking: slot: {}, stored_accounts: {} => alive_accounts: {} ({} bytes; aligned to: {})",
@@ -1113,7 +1115,7 @@ impl AccountsDB {
                 slot,
                 &accounts,
                 &hashes,
-                |_| shrunken_store.clone(),
+                |_, _| shrunken_store.clone(),
                 write_versions.into_iter(),
             );
             start.stop();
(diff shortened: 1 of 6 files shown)


runtime/src/accounts_db.rs
```diff
@@ -259,7 +267,7 @@ impl AccountStorageEntry {
     }
 
     pub fn slot(&self) -> Slot {
-        self.slot
+        self.slot.load(Ordering::Acquire)
```
carllin, line 270: 👀

sakridge:
Comparison of find storage time, new is on left. After warmup, it's improved from 10-15ms to 2-4ms.

find-storage

runtime/src/accounts_db.rs
```diff
@@ -1475,13 +1650,34 @@ impl AccountsDB {
...
         }
         remove_storages_elapsed.stop();
 
         let num_slots_removed = all_removed_slot_storages.len();
 
+        let mut recycled_count = 0;
+        let mut recycle_stores = self.recycle_stores.write().unwrap();
+        for slot_entries in &all_removed_slot_storages {
+            let entry = slot_entries.read().unwrap();
+            for (_slot, stores) in entry.iter() {
```
carllin, line 1664: nit: the key here is the `store_id`, not the `_slot`

runtime/src/accounts_db.rs
```diff
@@ -1381,9 +1416,68 @@ impl AccountsDB {
...
+        let store = self.try_recycle_store(slot, min_size, max_size)?;
+        self.insert_store(slot, store.clone());
+        Some(store)
+    }
+
+    fn try_recycle_store(
+        &self,
+        slot: Slot,
+        min_size: u64,
+        max_size: u64,
```
carllin, line 1434: is there any reason to have a `max_size` if we're always passing `std::u64::MAX`?

runtime/src/accounts_db.rs
```diff
@@ -1475,13 +1650,34 @@ impl AccountsDB {
...
+            }
+            if recycle_stores.len() > MAX_RECYCLE_STORES {
+                let dropped_count = total_removed_storage_entries - recycled_count;
+                self.stats
+                    .dropped_stores
+                    .fetch_add(dropped_count as u64, Ordering::Relaxed);
+                break;
+            }
+        }
+        drop(recycle_stores);
```
carllin, line 1679: I think it would be cleaner to move this block into a `recycle_slot_stores(&self, removed_slot_storages: &[SlotStores])` function. 

That way we can move the outer break into the inner `break` and `return` instead, and we don't need to explicitly drop the `recycle_stores` lock

runtime/src/accounts_db.rs
```diff
@@ -1381,9 +1416,68 @@ impl AccountsDB {
...
+        for (i, store) in recycle_stores.iter().enumerate() {
+            if Arc::strong_count(store) == 1 {
+                max = std::cmp::max(store.accounts.capacity(), max);
+                min = std::cmp::min(store.accounts.capacity(), min);
+                avail += 1;
+
+                if store.accounts.capacity() >= min_size && store.accounts.capacity() < max_size {
+                    let old_id = store.append_vec_id();
+                    store.recycle(slot, self.next_id.fetch_add(1, Ordering::Relaxed));
+                    let ret = recycle_stores.swap_remove(i);
```
carllin, line 1449: nit: I think we could `swap_remove()`, drop the lock, then `store.recycle()`

runtime/src/accounts_db.rs
```diff
@@ -1791,17 +1993,30 @@ impl AccountsDB {
             if rvs.is_empty() {
                 storage.set_status(AccountStorageStatus::Full);
 
-                // See if an account overflows the default append vec size.
-                let data_len = (with_meta[infos.len()].1.data.len() + 4096) as u64;
-                if data_len > self.file_size {
-                    self.create_and_insert_store(slot, data_len * 2);
+                // See if an account overflows the append vecs in the slot.
+                let data_len = (with_meta[infos.len()].1.data.len() + STORE_META_OVERHEAD) as u64;
```
carllin, line 1997: In the old implementation, this was `(with_meta[infos.len()].1.data.len() + 4096` (I think to account for all the metas stored here: https://github.com/solana-labs/solana/blob/master/runtime/src/append_vec.rs#L420-L425), but `STORE_META_OVERHEAD = 256 < 4096`

runtime/src/accounts_db.rs
```diff
@@ -1381,9 +1416,68 @@ impl AccountsDB {
...
+        &self,
+        slot: Slot,
+        min_size: u64,
+        max_size: u64,
+    ) -> Option<Arc<AccountStorageEntry>> {
+        let mut max = 0;
+        let mut min = std::u64::MAX;
+        let mut avail = 0;
+        let mut recycle_stores = self.recycle_stores.write().unwrap();
+        for (i, store) in recycle_stores.iter().enumerate() {
```
carllin, line 1440: If `recycle_stores` was a `BTreeMap<usize, Vec<Arc<AccountStorageEntry>>` keyed by the store size, then we would only need a lookup and a `pop()` and can avoid this scan under the lock.

runtime/src/accounts_db.rs
```diff
@@ -1133,7 +1158,17 @@ impl AccountsDB {
         rewrite_elapsed.stop();
 
         let mut drop_storage_entries_elapsed = Measure::start("drop_storage_entries_elapsed");
-        drop(dead_storages);
+        let mut recycle_stores = self.recycle_stores.write().unwrap();
```
carllin, line 1161: I'm curious about the contention on this lock, so would be nice to factor out the `self.recycle_stores.write().unwrap();` into a separate timing field outside of `Measure::start("drop_storage_entries_elapsed");`

runtime/src/accounts_db.rs
```diff
@@ -1475,13 +1650,34 @@ impl AccountsDB {
...
-                all_removed_slot_storages.push(slot_removed_storages);
+                all_removed_slot_storages.push(slot_removed_storages.clone());
             }
         }
         remove_storages_elapsed.stop();
 
         let num_slots_removed = all_removed_slot_storages.len();
 
+        let mut recycled_count = 0;
+        let mut recycle_stores = self.recycle_stores.write().unwrap();
```
carllin, line 1661: same as https://github.com/solana-labs/solana/pull/12885/files#r514094338, would be nice to add a timing field here as part of the `purge_slots_time` metric

runtime/src/accounts_db.rs
```diff
@@ -1791,17 +1993,30 @@ impl AccountsDB {
...
-                // See if an account overflows the default append vec size.
-                let data_len = (with_meta[infos.len()].1.data.len() + 4096) as u64;
-                if data_len > self.file_size {
-                    self.create_and_insert_store(slot, data_len * 2);
+                // See if an account overflows the append vecs in the slot.
+                let data_len = (with_meta[infos.len()].1.data.len() + STORE_META_OVERHEAD) as u64;
+                if !self.has_space_available(slot, data_len) {
+                    let special_store_size = std::cmp::max(data_len * 2, self.file_size);
+                    if self
+                        .try_recycle_and_insert_store(slot, special_store_size, std::u64::MAX)
```
carllin, line 2001: Should this call to  `try_recycle_and_insert_store ` also be part of the `self.stats.store_find_store` metric, or would it be better off in a separate metric (just accounting for all the recycle locks here)

runtime/src/accounts_db.rs
carllin: Our first acquire-release semantics! To celebrate maybe we should include a comment about why it's necessary here?

runtime/src/accounts_db.rs
sakridge: When shrinking, we don't pass in u64::MAX.

runtime/src/accounts_db.rs
sakridge: I added a test to verify the constant value.

runtime/src/accounts_db.rs
sakridge: I think we use it in gossip as well. Most of our atomics are not critical for functionality like perf counters which are fine if they are incremented or loaded out of order. I thought it would be bad if the load/store is re-ordered for the slot value, a thread could see a stale slot value.

runtime/src/accounts_db.rs
sakridge: Added the helper.

runtime/src/accounts_db.rs
sakridge: re-ordered

runtime/src/accounts_db.rs
sakridge: When I look at the perf, the current implementation is quite fast. I will experiment with this in a follow-up.

runtime/src/accounts_db.rs
sakridge: added

runtime/src/accounts_db.rs
sakridge: added

runtime/src/accounts_db.rs
carllin: Just to clarify my understanding and future reference about the interaction between the atomics `slot` and `id` fields in `AccountStorageEntry`:

A recycle scenario can be summarized with something like the following:

Initial state: 
```
self.slot = 0
self.id = 0
```

Thread A grabs and holds a copy of the storage entry `Arc` and reads
`Thread A`
```
let ref_count = storage.lookup(store_id_x);  // lookup in the storage for some store_id_x, may be the old store id, or the new recycled id
ref_count.store(2, Ordering::SeqCst);      // storage clone bumps the Arc refcount, Arc uses `SeqCst`   
slot.load(Ordering::Acquire);                             
id.load(Ordering::Relaxed);
ref_count.store(1, Ordering::SeqCst);
```

Thread `B` performing the `recycle`:
`Thread B`
```
// do the recycle
if refcount.load(Ordering::SeqCst) == 1 {    // Arc `strong_count()` uses `SeqCst`   
    slot.store(5, Ordering::Release);                             
    id.store(5, Ordering::Relaxed);
}
// Insert the store
storage.lock().insert(store_id_y);   // insert recycled store into storage with new store id
```

So now looking at possible combinations of `(slot, id) ` that thread `A` can see: 

1) If Thread `A` is reading directly grabbing and holding copy of the `Arc` for the storage entry from the `recycled_stores`, then it can see  any combination of (0, 0), (0, 5), (5, 5) (5, 0), because the loads can interleave with the recycle logic in `B` that starts at `if Arc::strong_count(store) == 1 {`

2) If Thread `A` is finding the storage entry through a read on `self.storages`:

Can thread `A` still see a mixed combination like (0, 5), or (5, 0)? I think the storage lock and the `strong_count() == 1` check prevent this from happening?

For example for `(5, 0)`, this means `A` must have seen `B`'s store to `slot`, which means store_id_x == store_id_y,
so the loads in thread `A` must have occurred *after* the new storage with `store_id_y` was inserted
into the store at `storage.lock().insert(store_id_y);`. And because locks have full fences, both loads in 
thread `A` cannot be ordered before the `storage.lock().insert(store_id_y);`, and must see both stores from thread `B`.

`(0, 5)` follows a similar reasoning.

runtime/src/accounts_db.rs
```diff
@@ -1791,17 +2017,30 @@ impl AccountsDB {
...
```
carllin, line 2025: nit: Should we wrap this call to `try_recycle_and_insert_store` in a `Measure` and be add it to the self.stats.store_find_store timing metric?

carllin (approved):
Great stuff!

PR #12931, opened 2020-10-15
Report compute budget usage

jackcmay:
#### Problem

Developers don't have a nice way to determine how much of their compute budget a program used

#### Summary of Changes

On program completion the compute budget usage will be reported.  When making cross-program invocations each invocation's usage will also be reported.

Fixes #


jstarry:
@jackcmay now that we have a syscall for reporting compute units, I think we should remove this log. It's not really necessary and it's a bit annoying. I would like to add this to the tx metadata instead

jstarry:
I created an issue for this here: https://github.com/solana-labs/solana/issues/21489

PR #12952, opened 2020-10-16
Check payer balance for program account rent as needed

### LINKED ISSUE #12937: Cli: solana-deploy balance check doesn't include rent
#### Problem
`solana deploy` checks to see if the funding account can cover all the write transactions of a program deployment, but doesn't include the rent-exempt balance that the program account must be initialized with: https://github.com/solana-labs/solana/blob/359707c85eaf9fb3dd793adb4038160e036aba0a/cli/src/cli.rs#L1247

As a result, users get this monstrosity of an error:
```
Error: dynamic program error: Program account allocation failed: RPC response error -32002: Transaction simulation failed: Error processing Instruction 0: custom program error: 0x1
```

#### Proposed Solution
Fix balance check to include rent-exempt amount
Also see about fixing up the error here so that it doesn't stomp on the SystemError parsing: https://github.com/solana-labs/solana/blob/359707c85eaf9fb3dd793adb4038160e036aba0a/cli/src/cli.rs#L1275

CriesofCarrots:
#### Problem
`solana deploy` checks to see if the funding account can cover all the write transactions of a program deployment, but doesn't include the rent-exempt balance that the program account must be initialized with.

#### Summary of Changes
- Extend function to sum balance with multiple fees, when provided. Use in process_deploy.

Fixes #12937

### DIFF
--- a/cli/src/checks.rs
+++ b/cli/src/checks.rs
@@ -55,11 +55,41 @@ pub fn check_account_for_multiple_fees_with_commitment(
     messages: &[&Message],
     commitment: CommitmentConfig,
 ) -> Result<(), CliError> {
+    check_account_for_spend_multiple_fees_with_commitment(
+        rpc_client,
+        account_pubkey,
+        0,
+        fee_calculator,
+        messages,
+        commitment,
+    )
+}
+
+pub fn check_account_for_spend_multiple_fees_with_commitment(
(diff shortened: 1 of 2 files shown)


t-nelson (approved):
Nice!  I'm surprised we didn't run into this earlier

PR #12958, opened 2020-10-17
feat: add Secp256k1 support to solana-web3.js

### LINKED ISSUE #12672: Add Secp256k1 support to solana-web3.js
The newly added Secp256k1 program/instruction is not yet exposed in solana-web3.js.  It should be easy to create this `Instruction` in web3.sj: https://github.com/solana-labs/solana/blob/3930cb865a866d162c7023129f0bd4318bcd1ed8/programs/secp256k1/src/lib.rs#L22

Other context:
* https://github.com/solana-labs/solana/blob/master/sdk/src/secp256k1.rs
* https://github.com/solana-labs/solana/blob/master/sdk/src/secp256k1_program.rs

The `Sysvar1nstructions1111111111111111111111111` sysvar constant should be exposed as well: https://github.com/solana-labs/solana/blob/master/sdk/src/sysvar/instructions.rs

oJshua:
#### Problem
The newly added Secp256k1 program/instruction is not yet exposed in solana-web3.js. It should be easy to create this Instruction in web3.js.

#### Summary of Changes
- Create Secp256k1 program and instruction classes
- Allow creating instruction with private key + message or public key + signature + message (support Ren use case)

Fixes #12672

### DIFF
--- a/web3.js/module.d.ts
+++ b/web3.js/module.d.ts
@@ -531,6 +531,7 @@ declare module '@solana/web3.js' {
   export const SYSVAR_RENT_PUBKEY: PublicKey;
   export const SYSVAR_REWARDS_PUBKEY: PublicKey;
   export const SYSVAR_STAKE_HISTORY_PUBKEY: PublicKey;
+  export const SYSVAR_INSTRUCTIONS_PUBKEY: PublicKey;
 
   // === src/vote-account.js ===
   export const VOTE_PROGRAM_ID: PublicKey;
--- a/web3.js/module.d.ts
+++ b/web3.js/module.d.ts
@@ -966,6 +966,54 @@ declare module '@solana/web3.js' {
     ): AuthorizeNonceParams;
   }
 
+  // === src/secp256k1-program.js ===
+  export type CreateSecp256k1InstructionWithPublicKeyParams = {
+    publicKey: Buffer | Uint8Array | Array<number>;
+    message: Buffer | Uint8Array | Array<number>;
+    signature: Buffer | Uint8Array | Array<number>;
+    recoveryId: number;
+  };
+
+  export type CreateSecp256k1InstructionWithPrivateKeyParams = {
+    privateKey: Buffer | Uint8Array | Array<number>;
+    message: Buffer | Uint8Array | Array<number>;
+  };
+
+  export type DecodedSecp256k1Instruction = {
+    numSignatures: number;
+    signatureOffset: number;
+    signatureInstructionOffset: number;
+    ethAddressOffset: number;
+    ethAddressInstructionIndex: number;
+    messageDataOffset: number;
+    messageDataSize: number;
+    messageInstructionIndex: number;
+    signature: Buffer;
+    ethPublicKey: Buffer;
+    recoveryId: number;
+    message: Buffer;
+  };
+
+  export class Secp256k1Instruction {
+    static decodeInstruction(
+      instruction: TransactionInstruction,
+    ): DecodedSecp256k1Instruction;
+
+    static checkProgramId(programId: PublicKey): void;
+  }
+
+  export class Secp256k1Program {
+    static get programId(): PublicKey;
+
+    static createInstructionWithPublicKey(
+      params: CreateSecp256k1InstructionWithPublicKeyParams,
+    ): TransactionInstruction;
+
+    static createInstructionWithPrivateKey(
+      params: CreateSecp256k1InstructionWithPrivateKeyParams,
+    ): TransactionInstruction;
+  }
+
   // === src/loader.js ===
   export class Loader {
     static getMinNumSignatures(dataLength: number): number;
--- a/web3.js/module.d.ts
+++ b/web3.js/module.d.ts
@@ -980,29 +980,6 @@ declare module '@solana/web3.js' {
     message: Buffer | Uint8Array | Array<number>;
   };
 
-  export type DecodedSecp256k1Instruction = {
-    numSignatures: number;
-    signatureOffset: number;
-    signatureInstructionOffset: number;
-    ethAddressOffset: number;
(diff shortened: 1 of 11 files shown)


jstarry:
> should be easy

534 lines changed 😆

mvines:
@oJshua - as a separate PR can you please also fix that simulateTransaction limitation you discovered along the way

web3.js/module.d.ts
```diff
@@ -966,6 +966,54 @@ declare module '@solana/web3.js' {
     ): AuthorizeNonceParams;
```
jstarry, line 966: Please export `SYSVAR_INSTRUCTIONS_PUBKEY` in this file and in the flow file

web3.js/src/secp256k1-program.js
```diff
@@ -0,0 +1,210 @@
...
+ * @property {Buffer | Uint8Array | Array<number>} privateKey
+ * @property {Buffer | Uint8Array | Array<number>} message
+ */
+export type CreateSecp256k1InstructionWithPrivateKeyParams = {|
+  privateKey: Buffer | Uint8Array | Array<number>,
+  message: Buffer | Uint8Array | Array<number>,
+|};
+
+/**
+ * A decoded Secp256k instruction
```
jstarry, line 46:
```suggestion
 * A decoded Secp256k1 instruction
```

web3.js/src/secp256k1-program.js
```diff
@@ -0,0 +1,210 @@
...
+      message,
+    };
+  }
+
+  /**
+   * @private
+   */
+  static checkProgramId(programId: PublicKey) {
+    if (!programId.equals(Secp256k1Program.programId)) {
+      throw new Error('invalid instruction; programId is not Secp256kProgram');
```
jstarry, line 102:
```suggestion
      throw new Error('invalid instruction; programId is not Secp256k1Program');
```

web3.js/test/secp256k1-program.test.js
```diff
@@ -0,0 +1,142 @@
...
+    message,
+  } = Secp256k1Instruction.decodeInstruction(instruction);
+
+  expect(numSignatures).toEqual(1);
+  expect(signatureOffset).toEqual(32);
+  expect(signatureInstructionOffset).toEqual(0);
+  expect(ethAddressOffset).toEqual(12);
+  expect(ethAddressInstructionIndex).toEqual(0);
+  expect(messageDataOffset).toEqual(97);
+  expect(messageDataSize).toEqual('Test message'.length);
```
jstarry, line 56: this should be testing the buffer length right?

web3.js/test/secp256k1-program.test.js
```diff
@@ -0,0 +1,142 @@
...
+    connection,
+    transaction,
+    [from],
+    {
+      commitment: 'single',
+      skipPreflight: true,
+    },
+  );
+
+  expect(transactionSignature.length > 0).toEqual(true);
```
jstarry, line 105: nit: don't think this is needed, awaiting on the sendAndConfirm is sufficient because it will throw if the transaction fails

web3.js/src/secp256k1-program.js
```diff
@@ -0,0 +1,210 @@
...
+  BufferLayout.u16('messageDataSize'),
+  BufferLayout.u8('messageInstructionIndex'),
+  BufferLayout.blob(20, 'ethPublicKey'),
+  BufferLayout.blob(64, 'signature'),
+  BufferLayout.u8('recoveryId'),
+]);
+
+export class Secp256k1Instruction {
+  /**
+   * Decode a secp256k1 instruction
```
jstarry, line 84: Is the intention to only decode a subset of valid secp256k1 instructions?

web3.js/module.d.ts
oJshua: Thanks! Late addition

web3.js/src/secp256k1-program.js
oJshua: That is the intention. I should get some handling for invalid instructions.

oJshua: I noticed the other programs had decoders, so I was following suit with this. Maybe it's not necessary until we need it?

jstarry: If it's not needed right now, feel free to skip. Since you already made some progress, could be nice to wrap it up. Totally up to you!

> That is the intention. Do you think I need some handling for invalid cases?

Yes, I think the decoding should be comprehensive. But the thing I was getting at is that you make a lot of assumptions about structure of the instruction which would fail to properly decode valid instructions outside of those assumptions. For example, the instruction could include multiple signatures or the message and signature order could be flipped (should be using offsets to decode these).

oJshua: Very good point about multiple signatures! The implementation I ported over to web3 of course only deals with a single signature, but it's true we need to support >1 signatures as well. I will shelve this for now.

web3.js/src/secp256k1-program.js
```diff
@@ -0,0 +1,162 @@
...
+ */
+export type CreateSecp256k1InstructionWithPrivateKeyParams = {|
+  privateKey: Buffer | Uint8Array | Array<number>,
+  message: Buffer | Uint8Array | Array<number>,
+|};
+
+const SECP256K1_INSTRUCTION_LAYOUT = BufferLayout.struct([
+  BufferLayout.u8('numSignatures'),
+  BufferLayout.u16('signatureOffset'),
+  BufferLayout.u8('signatureInstructionOffset'),
```
jstarry, line 48:
```suggestion
  BufferLayout.u8('signatureInstructionIndex'),
```

jstarry (dismissed):
I missed one small thing in the first pass, after fixing feel free to merge!

PR #12961, opened 2020-10-17
Solenoid compiler

0b01:
Solenoid is a compiler for translating EVM to BPF. [Currently it can compile a simple hello world contract into bpf.]( https://gist.github.com/0b01/7495ffc3c9c97b09f666d629e21fecf5)


solenoid/src/compiler.rs
```diff
@@ -0,0 +1,1163 @@
...
+                self.push_label(name, builder);
+                builder.build_unconditional_branch(self.errbb.unwrap());
+                self.pop_label();
+
+                if is_runtime {
+                    warn!("Invalid instruction encountered. Continuing compilation!");
+                    return Some(())
+                } else {
+                    warn!("Invalid instruction encountered. Halting compilation!");
+                    return None;
```
0b01, line 680: This may be problematic if the constructor has complex branching logic. Should compile all the instructions.

garious:
@0b01, no need to put this in the `solana` repo. That'll only slow you down.  How about a standalone repo of your own for now, and once it's more full-featured, we either transfer it to the solana-labs organization or integrate it into the `solana-program-library` repo?  Also, take a look at the `token` program and its `build.rs` file for an example of compiling a Rust crate into BPF and running it.

PR #12962, opened 2020-10-19
improves threads' utilization in processing gossip packets

behzadnouri:
#### Problem
ClusterInfo::process_packets handles incoming packets in a thread_pool:
https://github.com/solana-labs/solana/blob/87311cce7/core/src/cluster_info.rs#L2118-L2134

However, profiling runtime shows that threads are not well utilized and
a lot of the processing is done sequentially.

#### Summary of Changes
This commit redistributes the work done in parallel. Testing on a gce
cluster shows 20%+ improvement in processing gossip packets with much
smaller variations.

### DIFF
--- a/core/src/cluster_info.rs
+++ b/core/src/cluster_info.rs
@@ -211,6 +211,7 @@ struct GossipStats {
     new_push_requests2: Counter,
     new_push_requests_num: Counter,
     filter_pull_response: Counter,
+    process_gossip_packets_time: Counter,
     process_pull_response: Counter,
     process_pull_response_count: Counter,
     process_pull_response_len: Counter,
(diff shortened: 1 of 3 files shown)


behzadnouri:
cluster_info_stats2 process_gossip_packets_time_2
Left chunk is this commit.

core/src/cluster_info.rs
```diff
@@ -2125,12 +2146,31 @@ impl ClusterInfo {
...
+                .filter_map(|packet| {
+                    let protocol: Protocol =
+                        limited_deserialize(&packet.data[..packet.meta.size]).ok()?;
+                    protocol.sanitize().ok()?;
+                    let protocol = protocol.par_verify()?;
+                    Some((packet.meta.addr(), protocol))
+                })
+                .collect()
         });
+        self.handle_packets(
```
sakridge, line 2163: Hmmm, so doing the handle_packets in parallel doesn't give benefit?

sakridge (approved):
lgtm, I like how verify is separated out better. Can hook up a GPU path easier :)

core/src/cluster_info.rs
behzadnouri: probably it does, but not as much. i will update with more of the packet handling work done on thread pool.

behzadnouri:
> lgtm, I like how verify is separated out better. Can hook up a GPU path easier :)

Actually from runtime profiles, those signature verifications takes significant time. So using GPU might be a good idea.

PR #12984, opened 2020-10-19
sdk: Add SyscallStubs to enable syscall interception when building programs for non-BPF

mvines:
There's no way to override `sol_invoke_signed()` from a program unit-test environment, resulting in unsightly program-specific stubs.

With this PR, program unit tests can use `set_syscall_stubs()` to swap in their own `sol_invoke_signed()`, which would allow for code like [this](https://github.com/solana-labs/solana-program-library/blob/0e8c4fb5bf13229c3d23a5b5d0b09d4d24bfc306/token-swap/program/src/processor.rs#L534-L568) to live  under `#[cfg(test)]`.

Note there's still a ton of boilerplate that's required to supply an implementation of `SyscallStubs::sol_invoke_signed()`.  This'll be worked on as (multiple) follow-up PRs.   Generate thought is that `runtime/` supplies a `SyscallStubs` implementation with a full-blown `InvokeContext`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4440,6 +4440,7 @@ dependencies = [
  "hex",
  "hmac",
  "itertools 0.9.0",
+ "lazy_static",
  "libsecp256k1",
  "log 0.4.8",
  "memmap",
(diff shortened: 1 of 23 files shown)


jackcmay:
This looks good with the side effect of removing the unmangled stub names for non-bpf programs, woot!

PR #12985, opened 2020-10-19
sdk: add "everything" feature

mvines:
Cargo features are intended to be additive.  Once the "program" feature is added there's no way to get back to the feature set of solana-sdk that includes everything.   To avoid this add an explicit "everything" feature (enabled by default).

### DIFF
--- a/sdk/Cargo.toml
+++ b/sdk/Cargo.toml
@@ -11,12 +11,14 @@ edition = "2018"
 [features]
 # On-chain program specific dependencies
 program = []
-# Dependencies that are not compatible or needed for on-chain programs
+# Everything includes functionality that is not compatible or needed for on-chain programs
 default = [
+  "everything"
+]
+everything = [
     "assert_matches",
     "byteorder",
     "chrono",
-    "curve25519-dalek",
     "generic-array",
     "memmap",
     "rand",
@@ -65,6 +67,9 @@ rustversion = "1.0.3"
 libsecp256k1 = { version = "0.3.5", optional = true }
 sha3 = { version = "0.9.1", optional = true }
 digest = { version = "0.9.0", optional = true }
+
+[target.'cfg(not(target_arch = "bpf"))'.dependencies]
+curve25519-dalek = { version = "2.1.0" }
 
 [dev-dependencies]
 curve25519-dalek = "2.1.0"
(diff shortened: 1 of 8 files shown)


sdk/Cargo.toml
```diff
@@ -66,6 +68,9 @@ libsecp256k1 = { version = "0.3.5", optional = true }
 sha3 = { version = "0.9.1", optional = true }
 digest = { version = "0.9.0", optional = true }
 
+[target.'cfg(not(target_arch = "bpf"))'.dependencies]
+curve25519-dalek = { version = "2.1.0" }
```
mvines, line 72: freebie: this fixes a compile error when using `Pubkey::create_program_address()` from a program when not targeting BPF

mvines:
@jackcmay - care to give this a quick 👀?

sdk/src/program_error.rs
```diff
@@ -5,7 +5,7 @@ use thiserror::Error;
 
 #[cfg(feature = "program")]
 use crate::info;
-#[cfg(not(feature = "program"))]
+#[cfg(all(feature = "everything", not(feature = "program")))]
```
jackcmay, line 8: Was this the underlying motivation for this change?

sdk/src/pubkey.rs
```diff
@@ -130,7 +124,7 @@ impl Pubkey {
         // not supported
         #[cfg(not(all(feature = "program", target_arch = "bpf")))]
         {
-            let mut hasher = Hasher::default();
+            let mut hasher = crate::hash::Hasher::default();
```
jackcmay, line 127: Were you seeing issues with this?

sdk/src/program_error.rs
mvines: Not directly.  When trying to test stuff over in SPL land, the semantics of enabling "program" causing code to be removed was generating confusing errors.

sdk/src/pubkey.rs
mvines, line 127: Na, my goal here was to remove all those feature-specific `use` statements at the top of pubkey.rs, to clear the way for easier future refactoring at #12987 and #12989

jackcmay (approved):

PR #12987, opened 2020-10-19
Add replacements for Pubkey::new_rand()/Hash::new_rand()

mvines:
`Pubkey::new_rand()` and `Hash::new_rand()` are problematic.  They are unavailable when compiling `solana-sdk` with the "program" feature, but will come back when the "everything" (https://github.com/solana-labs/solana/pull/12985) feature.   

Due to how Cargo feature addition works, a program developer could inadvertently add a dependency on `Pubkey::new_rand()` due to the "everything" feature getting activated in a `[dev-dependency]` when building native only to later have their build fail when targetting BPF.

When `Pubkey` and `Hash` move into a new `solana-program-sdk` create (that `solana-sdk` depends on),  `Pubkey::new_rand()` and `Hash::new_rand()` cannot come along.   This will make it impossible to support the `new_rand()` functions where they currently live.

The migration plan:
1. Land this PR in master, v1.4 and v1.3
2. In v1.4, as `Pubkey` and `Hash` move into a new `solana-program-sdk` supply a stub for `Pubkey::new_rand()` when !BPF for now. `Hash::new_rand()` is believed to only be in use internally so remove it entirely in v1.4
3. v1.5/v1.6 timeframe, remove `Pubkey::new_rand()`

### DIFF
--- a/account-decoder/src/parse_account_data.rs
+++ b/account-decoder/src/parse_account_data.rs
@@ -111,8 +111,8 @@ mod test {
 
     #[test]
     fn test_parse_account_data() {
-        let account_pubkey = Pubkey::new_rand();
-        let other_program = Pubkey::new_rand();
+        let account_pubkey = solana_sdk::pubkey::new_rand();
+        let other_program = solana_sdk::pubkey::new_rand();
         let data = vec![0; 4];
         assert!(parse_account_data(&account_pubkey, &other_program, &data, None).is_err());
 
--- a/account-decoder/src/parse_config.rs
+++ b/account-decoder/src/parse_config.rs
@@ -117,7 +117,7 @@ mod test {
             }))
             .unwrap(),
         };
-        let info_pubkey = Pubkey::new_rand();
+        let info_pubkey = solana_sdk::pubkey::new_rand();
         let validator_info_config_account = create_config_account(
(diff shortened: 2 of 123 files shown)


account-decoder/src/parse_account_data.rs
```diff
@@ -111,8 +111,8 @@ mod test {
 
     #[test]
     fn test_parse_account_data() {
-        let account_pubkey = Pubkey::new_rand();
-        let other_program = Pubkey::new_rand();
+        let account_pubkey = solana_sdk::pubkey::new_rand();
```
jackcmay, line 114: Stating `solana_sdk` everytime is pretty verbose, we would probably want:

```rust
use solana_sdk::pubkey;

...

let account_pubkey = pubkey::new_rand();
```

jackcmay:
Looks like except one nit

account-decoder/src/parse_account_data.rs
mvines: that was just an artifact of mass search/replace

mvines: I'm going to leave this as is, it can be cleaned up later.   It'll be a painful manual edit to over a hundred files

PR #12988, opened 2020-10-19
adds more parallel processing to gossip packets handling

behzadnouri:
#### Problem
`ClusterInfo::handle_packets` can do some parallel processing:
https://github.com/solana-labs/solana/blob/c0675968b/core/src/cluster_info.rs#L1703-L1812

#### Summary of Changes
Incoming pull-requests and pull-responses are now processed using parallel iterators.

### DIFF
--- a/core/src/cluster_info.rs
+++ b/core/src/cluster_info.rs
@@ -36,13 +36,12 @@ use solana_sdk::sanitize::{Sanitize, SanitizeError};
 use bincode::{serialize, serialized_size};
 use core::cmp;
 use itertools::Itertools;
-use rayon::iter::IntoParallelIterator;
-use rayon::iter::ParallelIterator;
+use rayon::prelude::*;
 use rayon::{ThreadPool, ThreadPoolBuilder};
 use solana_ledger::staking_utils;
 use solana_measure::measure::Measure;
 use solana_measure::thread_mem_usage;
-use solana_metrics::{datapoint_debug, inc_new_counter_debug, inc_new_counter_error};
+use solana_metrics::{inc_new_counter_debug, inc_new_counter_error};
 use solana_net_utils::{
     bind_common, bind_common_in_range, bind_in_range, find_available_port_in_range,
     multi_bind_in_range, PortRange,
@@ -67,7 +66,7 @@ use solana_streamer::streamer::{PacketReceiver, PacketSender};
 use std::{
     borrow::Cow,
     cmp::min,
-    collections::{HashMap, HashSet},
+    collections::{hash_map::Entry, HashMap, HashSet},
     fmt,
     net::{IpAddr, Ipv4Addr, SocketAddr, TcpListener, UdpSocket},
     ops::{Deref, DerefMut},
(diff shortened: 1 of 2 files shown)


core/src/cluster_info.rs
```diff
@@ -1951,6 +1939,62 @@ impl ClusterInfo {
...
+        thread_pool: &ThreadPool,
+        stakes: &HashMap<Pubkey, u64>,
+        epoch_time_ms: u64,
+    ) {
+        type HashMap = std::collections::HashMap<Pubkey, Vec<CrdsValue>>;
+        if responses.is_empty() {
+            return;
+        }
+        fn extend(hash_map: &mut HashMap, (from, mut data): (Pubkey, Vec<CrdsValue>)) {
+            match hash_map.entry(from) {
```
carllin, line 1954: I think in these cases it's more idiomatic to use `if let Some(value) = hash_map.get(&from) {..} else {..}`

carllin, line 1954: I think it's more idiomatic to write:
```
hash_map.entry(from)
    .and_modify(|entry| {..})
    .or_insert(data)
```

core/src/cluster_info.rs
```diff
@@ -1951,6 +1939,62 @@ impl ClusterInfo {
...
+            }
+            for kv in other {
+                extend(&mut hash_map, kv);
+            }
+            hash_map
+        }
+        let responses = thread_pool.install(|| {
+            responses
+                .into_par_iter()
+                .with_min_len(1024)
```
carllin, line 1979: This is an interesting way to join the responses together faster. Do we know how large the "responses" vector typically is? Wondering if it's enough to take advantage of these 1024-len chunks

carllin:
@behzadnouri looks correct! Just interested in knowing if we're using the appropriate batching size `1024` in `handle_batch_pull_requests` and `handle_batch_pull_responses`

core/src/cluster_info.rs
behzadnouri: There are these two metrics:
https://github.com/solana-labs/solana/blob/e6b821c39/core/src/cluster_info.rs#L1997-L2000
in the following function, which may give an approximate _lower-bound_ for the length of responses here. Based on that approximate, on "average", this particular instance probably will not do much parallel processing today on mainnet.

But, nevertheless, I think it is good to do this using parallel iterators here, because it will help this scale should the network grow much bigger. It will also keep things more robust should some nodes fall behind and receive abnormal traffic. Even today on mainnet,  99% percentile is more than twice the average (and "max" is so far out).

cluster_info_stats3 process_pull_response_len_tail

core/src/cluster_info.rs
```diff
@@ -2016,6 +2004,62 @@ impl ClusterInfo {
...
     }
 
+    fn handle_batch_pull_responses(
+        &self,
+        responses: Vec<(Pubkey, Vec<CrdsValue>)>,
+        thread_pool: &ThreadPool,
+        stakes: &HashMap<Pubkey, u64>,
+        epoch_time_ms: u64,
+    ) {
+        type HashMap = std::collections::HashMap<Pubkey, Vec<CrdsValue>>;
```
sakridge, line 2014: How about something like `ResponseMap`, bit confusing to conflict with the generic `HashMap` name.

behzadnouri: removed it entirely. it was only a type alias to make things shorter, but i think writing the actual type would be more explicit.

carllin (approved):

sakridge:
How does the performance compare?

Only worry of mine is the heavy re-ordering of requests may put older requests (or different types of requests) at the back of the queue vs. newer ones. If everything is fast enough then I suppose it doesn't matter. Also the extra copy of requests into these separate vecs, although we are already doing that in certain cases anyway.

behzadnouri:
> How does the performance compare?
> 
> Only worry of mine is the heavy re-ordering of requests may put older requests (or different types of requests) at the back of the queue vs. newer ones. If everything is fast enough then I suppose it doesn't matter. Also the extra copy of requests into these separate vecs, although we are already doing that in certain cases anyway.

When I tested this on gce it looked fine. The size of the network definitely has an impact, so we need to further monitor on testnet. But I think overall trade-offs are positive. The re-ordering is not ideal, but it also allows more efficient processing. The copying I expect should be fine, because the data are moved.

sakridge (approved):
lgtm

Issue #12812, opened 2020-10-12
explorer: Decode BPF Loader 2 instructions


mvines:
The BPF Loader instructions show up as Raw in the explorer:
* Example [Write](https://github.com/solana-labs/solana/blob/2f5bb7e507279d7eff94e8a25bc931c3d2711c39/sdk/src/loader_instruction.rs#L13) instruction: https://explorer.solana.com/tx/5fFkQ71GhhrGnfXXoJ9H4E7bybN6mgM5397CT83Ukc3WnD23i9KbnhkL8ARVrr6K3WbjCXssVUe5Td4JM9jUp3Lq
* Example [Finalize](https://github.com/solana-labs/solana/blob/2f5bb7e507279d7eff94e8a25bc931c3d2711c39/sdk/src/loader_instruction.rs#L30) instruction: https://explorer.solana.com/tx/3ET4ogNzoBEKYubBM1vo3T5PeGBKyXKmDRVUpoTBMzpQxbWKYQqcCDGmUSxPRUedTVqXnNLaojMWsiVj7r8oKUqD

oJshua:
@mvines I don't seem to be getting any raw instruction data back from the API with those transactions you referenced. Do you have other examples?

mvines:
Strange, they worked before.  @CriesofCarrots - can you please check the explorer links above?  I think this is just due to the protobuf/rewards-related bugs we've been working through right? 

@oJshua - just select testnet, then search for "BPF Loader 2".  First transaction for the loader, at the moment, is a Finalize: https://explorer.solana.com/tx/5CV4tRvdrhoiUQKgD6stjW9VGQCszEZFBQUDf4gefZUWQA3XgFgs9hEduJQwdxZFtMPBqQKYZdpMX8r2Rj4yodNb?cluster=testnet

CriesofCarrots:
@mvines it's the issue fixed here: https://github.com/solana-labs/solana/pull/12923
It'd be nice to get that deployed onto our api nodes. v1.3.18?

@oJshua , I'm going to start working on rpc support for more parsed instructions this week, which should help: https://github.com/solana-labs/solana/issues/11316

mvines:
> @mvines it's the issue fixed here: #12923
> It'd be nice to get that deployed onto our api nodes. v1.3.18?

Thanks, yeah a v1.3.18 only for API node update sounds great.

CriesofCarrots:
[...]

PR #12998, opened 2020-10-19
Parse bpf loader instructions

CriesofCarrots:
#### Problem
The BPF Loader instructions show up as Raw in the explorer.

#### Summary of Changes
- Add json parsing for LoaderInstructions

Toward #12812

```json
// Write example ix

"instructions": [
    {
      "parsed": {
        "info": {
          "bytes": "AAAAAAAAAAABAAAAAAAAAAAAAABgrAEAAAAAAOAnAgAgAAAAAAAAAAAAAABbQwIAEgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABAAAAEAAAAAAAAAAAAAAAAAAAAAAAAAAHAAAAEAAAAAAAAAAAAAAAAAAAAAAAAAAbAAAAEAAAAAAAAAAAAAAAAAAAAAAAAAAmAAAAEAAAAAAAAAAAAAAAAAAAAAAAAAAQAAAAEgABAMj2AAAAAAAAgAEAAAAAAAAAYWJvcnQAc29sX2xvZ18AZW50cnlwb2ludABzbw==",
          "offset": 152056
        },
        "type": "write"
      },
      "program": "bpf-loader",
      "programId": "BPFLoader2111111111111111111111111111111111"
    }
  ]

// Finalize example ix

"instructions": [
    {
      "parsed": {
        "type": "finalize"
      },
      "program": "bpf-loader",
      "programId": "BPFLoader2111111111111111111111111111111111"
    }
  ]
```

### DIFF
--- a/transaction-status/src/parse_bpf_loader.rs
+++ b/transaction-status/src/parse_bpf_loader.rs
@@ -1,27 +1,38 @@
 use crate::parse_instruction::{ParsableProgram, ParseInstructionError, ParsedInstructionEnum};
 use bincode::deserialize;
-use serde_json::{json, Value};
-use solana_sdk::{instruction::CompiledInstruction, loader_instruction::LoaderInstruction};
+use serde_json::json;
+use solana_sdk::{
+    instruction::CompiledInstruction, loader_instruction::LoaderInstruction, pubkey::Pubkey,
+};
 
 pub fn parse_bpf_loader(
     instruction: &CompiledInstruction,
+    account_keys: &[Pubkey],
 ) -> Result<ParsedInstructionEnum, ParseInstructionError> {
     let bpf_loader_instruction: LoaderInstruction =
         deserialize(&instruction.data).map_err(|err| {
             println!("{:?}", err);
             ParseInstructionError::InstructionNotParsable(ParsableProgram::BpfLoader)
         })?;
+    if instruction.accounts.is_empty() || instruction.accounts[0] as usize >= account_keys.len() {
+        return Err(ParseInstructionError::InstructionKeyMismatch(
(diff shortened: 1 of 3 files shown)


CriesofCarrots:
@oJshua can you please take a look and see if this will suit you for #12812 ? In particular:
- do you need the destination account address in the `info` map?
- any reason not to return `bytes` as base64-encoded string?

oJshua:
> * do you need the destination account address in the `info` map? ... thinking on this more, I think it will be more consistent with parse_token if we do include this, so I will make that change, unless you object

No objection here!
> * any reason not to return `bytes` as base64-encoded string?

Works for me.
> Also, let me know if we need to return `info: null` instead of skipping that field when null

Skipping will be just fine!

This all works perfectly for #12812 . Thanks!

oJshua (approved):

transaction-status/src/parse_bpf_loader.rs
```diff
@@ -0,0 +1,102 @@
...
+    instruction::CompiledInstruction, loader_instruction::LoaderInstruction, pubkey::Pubkey,
+};
+
+pub fn parse_bpf_loader(
+    instruction: &CompiledInstruction,
+    account_keys: &[Pubkey],
+) -> Result<ParsedInstructionEnum, ParseInstructionError> {
+    let bpf_loader_instruction: LoaderInstruction =
+        deserialize(&instruction.data).map_err(|err| {
+            println!("{:?}", err);
```
mvines, line 14: 🖨️

CriesofCarrots: 🤦‍♀️

PR #13007, opened 2020-10-20
Don't hold dashmap write lock in store create

carllin:
#### Problem
@sakridge  noticed that `get_slot_storages()` in AccountsDb was occasionally very slow: 
Screen Shot 2020-10-19 at 11 22 09 PM

#### Summary of Changes
At first I thought it might be some bug in `DashMap`, but I ran

1) The benchmarks as included in this PR here
2) Variations of this benchmark with multiple readers/no writers
3) Varying numbers of keys in the maps
4) Reading randomly chosen keys in the map instead of the same key 

And throughout DashMap seemed to outperform/perform as well as the `Arc<RwLock>`

Turns out the issue is holding the write lock on the dashmap while waiting on a separate write lock here: https://github.com/solana-labs/solana/blob/942e4273ba52cade3d0a3e93dd7661230c98466f/runtime/src/accounts_db.rs#L1418-L1421, where the second write lock is taking a while (probably contention with reads?). This makes me think it might be worthwhile to turn the inner SlotStore into a `DashMap` as well 😄 

The simplest fix is just add a clone here: https://github.com/solana-labs/solana/blob/942e4273ba52cade3d0a3e93dd7661230c98466f/runtime/src/accounts_db.rs#L1413-L1417 (tested and the results look good), but I think you can also avoid taking the write lock every time as this PR implements.

After results:
Screen Shot 2020-10-19 at 11 34 44 PM

Fixes #

### DIFF
--- a/runtime/benches/accounts.rs
+++ b/runtime/benches/accounts.rs
@@ -3,6 +3,7 @@
 extern crate test;
 
 use dashmap::DashMap;
+use rand::Rng;
 use solana_runtime::{
     accounts::{create_test_accounts, Accounts},
     bank::*,
@@ -146,6 +147,60 @@ fn bench_delete_dependencies(bencher: &mut Bencher) {
     bencher.iter(|| {
         accounts.accounts_db.clean_accounts(None);
     });
+}
+
+#[bench]
+#[ignore]
+fn bench_concurrent_read_write(bencher: &mut Bencher) {
+    let num_readers = 5;
+    let accounts = Arc::new(Accounts::new(
+        vec![
+            PathBuf::from(std::env::var("FARF_DIR").unwrap_or_else(|_| "farf".to_string()))
+                .join("concurrent_read_write"),
+        ],
+        &ClusterType::Development,
+    ));
+    let num_keys = 1000;
+    let slot = 0;
+    accounts.add_root(slot);
+    let pubkeys: Arc<Vec<_>> = Arc::new(
+        (0..num_keys)
+            .map(|_| {
+                let pubkey = Pubkey::new_rand();
+                let account = Account::new(1, 0, &Account::default().owner);
+                accounts.store_slow(slot, &pubkey, &account);
+                pubkey
+            })
+            .collect(),
+    );
+
+    for _ in 0..num_readers {
+        let accounts = accounts.clone();
+        let pubkeys = pubkeys.clone();
+        Builder::new()
+            .name("readers".to_string())
+            .spawn(move || {
+                let mut rng = rand::thread_rng();
+                loop {
+                    let i = rng.gen_range(0, num_keys);
+                    test::black_box(accounts.load_slow(&HashMap::new(), &pubkeys[i]).unwrap());
+                }
+            })
+            .unwrap();
+    }
+
+    let num_new_keys = 1000;
+    let new_accounts: Vec<_> = (0..num_new_keys)
+        .map(|_| Account::new(1, 0, &Account::default().owner))
+        .collect();
+    bencher.iter(|| {
+        for account in &new_accounts {
+            // Write to a different slot than the one being read from. Because
+            // there's a new account pubkey being written to every time, will
+            // compete for the accounts index lock on every store
+            accounts.store_slow(slot + 1, &Pubkey::new_rand(), &account);
+        }
+    })
 }
 
 #[bench]
--- a/runtime/benches/accounts.rs
+++ b/runtime/benches/accounts.rs
@@ -2,7 +2,7 @@
 
 extern crate test;
 
-use rand::Rng;
+use dashmap::DashMap;
 use solana_runtime::{
     accounts::{create_test_accounts, Accounts},
     bank::*,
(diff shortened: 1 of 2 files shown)


runtime/src/accounts_db.rs
```diff
@@ -130,9 +130,8 @@ impl AccountStorage {
         slot: Slot,
         store_id: AppendVecId,
     ) -> Option<Arc<AccountStorageEntry>> {
-        self.0
-            .get(&slot)
-            .and_then(|storage_map| storage_map.value().read().unwrap().get(&store_id).cloned())
+        self.get_slot_stores(slot)
```
ryoqun, line 133: maybe is this proactive measure for the unrelated possible slow down?

runtime/src/accounts_db.rs
```diff
@@ -1410,12 +1409,18 @@ impl AccountsDB {
         let store =
             Arc::new(self.new_storage_entry(slot, &Path::new(&self.paths[path_index]), size));
         let store_for_index = store.clone();
-        let slot_storage = self
```
ryoqun, line 1413: DashMap returns RefMut here and the `.clone()` in new code makes it (and the write lock on DashMap backing it) drop as soon as it goes out of scope (well temporary expression in this case), right?

Could you comment about that?

Well, so hard finding and we need some commenting I guess.

ryoqun: Anyway, thanks for finding these very naughty bugs!

sakridge:
A bit unfortunate that `DashMap` has these performance pitfalls, but good find.

runtime/src/accounts_db.rs
carllin: Added a comment and a type for clarity!

runtime/src/accounts_db.rs
carllin: yeah this was just something else I suspected while investigating the root cause

runtime/src/accounts_db.rs
```diff
@@ -1410,12 +1409,18 @@ impl AccountsDB {
...
-            .storage
-            .0
-            .entry(slot)
-            .or_insert(Arc::new(RwLock::new(HashMap::new())));
-        slot_storage
+        let mut slot_storages = self.storage.get_slot_stores(slot);
```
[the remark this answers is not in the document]
carllin: fixed, thanks!

runtime/src/accounts_db.rs
```diff
@@ -1410,12 +1409,23 @@ impl AccountsDB {
...
-            .0
-            .entry(slot)
-            .or_insert(Arc::new(RwLock::new(HashMap::new())));
-        slot_storage
+            .get_slot_stores(slot)
+            .or_else(||
+            // DashMap entry.or_insert() returns a RefMut, essentially a write lock,
+            // which is dropped after this block ends, minimizing time held by the lock.
+            // However, we still want to persist the reference to the `SlotStores` behind
+            // the lock, hence we clone it out, (`SlotStores` is an Arc so is cheap to clone).
```
ryoqun, line 1420: thanks for explaining this. :)

runtime/src/accounts_db.rs
```diff
@@ -1410,12 +1409,23 @@ impl AccountsDB {
...
+            Some(self.storage
+                .0
+                .entry(slot)
+                .or_insert(Arc::new(RwLock::new(HashMap::new())))
+                .clone()))
+            .unwrap();
```
ryoqun, line 1426: well, I have a tendency of being wary of plain `unwrap()`. How about using `unwrap_or_else`? Well, I should have say so with my original comment...

runtime/benches/accounts.rs
```diff
@@ -145,54 +150,50 @@ fn bench_delete_dependencies(bencher: &mut Bencher) {
 
 #[bench]
 #[ignore]
-fn bench_concurrent_read_write(bencher: &mut Bencher) {
```
ryoqun, line 148: how about reverting back these nice AccountsDb-level bench tests? I think keeping and maintain these doesn't cost us much and might be handy to lay around the source tree in addition to the shinny direct DashMap/RwLock<HashMap> showdown. 

I guess, you just rewrote the body of these bench body code in a haste? ;)

runtime/src/accounts_db.rs
```diff
@@ -1410,12 +1409,23 @@ impl AccountsDB {
         let store =
             Arc::new(self.new_storage_entry(slot, &Path::new(&self.paths[path_index]), size));
         let store_for_index = store.clone();
-        let slot_storage = self
+
+        let slot_storages: SlotStores = self
```
ryoqun, line 1413: `: SlotStore  = `: i like the taste of well nuanced explicit type annotation at the age of rampant type inference xD

runtime/src/accounts_db.rs
carllin: thanks, good catch!

runtime/benches/accounts.rs
carllin: thanks, good catch!

ryoqun (approved):
lgtm! super good find!

sakridge (approved):

PR #13014, opened 2020-10-20
Fix signature index check in secp256k1 verify

jstarry:
#### Problem
Index check doesn't prevent invalid indices properly. If `signature_index` is 1 and there is only one instruction data, the check doesn't fail.

#### Summary of Changes
- Fix off-by-one

Fixes #

### DIFF
--- a/sdk/src/secp256k1.rs
+++ b/sdk/src/secp256k1.rs
@@ -72,7 +72,7 @@ pub fn verify_eth_addresses(
 
         // Parse out signature
         let signature_index = offsets.signature_instruction_index as usize;
-        if signature_index > instruction_datas.len() {
+        if signature_index >= instruction_datas.len() {
             return Err(Secp256k1Error::InvalidInstructionDataSize);
         }
         let signature_instruction = instruction_datas[signature_index];


sakridge (approved):
lgtm

ryoqun:
@sakridge @jstarry Uuum. is this need to be runtime-featured?

```
$ ./target/release/solana --url http://api.mainnet-beta.solana.com feature status

Stake By Feature Set:
2415322184 - 90% 
2674527397 - 9% 
Feature                                      Description                              Status
MoqiU1vryuCGQSxFKA1SZ316JdLEFFhoAu6cKUNk7dN  pubkey log syscall                       inactive
SECCKV5UVUsr8sTVSVAzULjdm87r7mLPaqH2FGZjevR  inflation kill switch                    inactive
YCKSgA6XmjtkQrHBQjpyNrX6EMhJPcYcLWMVgWn36iv  max program call depth 64                active since slot 41904000
3h1BQWPDS5veRsq6mDBWruEpgPxRJkfwGexg5iiQ9mYg consistent recentblockhashes sysvar      active since slot 41040000
3zydSLUwuqqsV3wL5wBsaVgyvMox3XTHx7zLEuQf1U2Z correct bank timestamps                  inactive
4kpdyrcj5jS47CZb2oJGfVxjYbsMm2Kx97gFyZrxxwXz no overflow rent distribution            inactive
5RzEHTnf6D7JPZCvwEzjM19kzBsyjSU3HoMfXaQmVgnZ ping-pong packet check #12794            inactive
D7KfP7bZxpkYtD4Pc38t9htgs1k5k47Yhxe4rp6WDVi8 sha256 syscall                           inactive
DFBnrgThdzH4W6wZ12uGPoWcMnvfZj11EHnxHcVxLPhD bpf_loader2 program                      active since slot 41040000
E3PHP7w8kB7np3CTQ1qQ2tW3KCtjRSXBQgW9vM2mWv2Y secp256k1 program                        active since slot 41040000
E5JiFDQCwyC6QfT9REFyMpfK2mHcmv1GUDySU1Ue7TYv spl-token multisig fix                   active since slot 41040000
EdM9xggY5y7AhNMskRG8NgGMnaP4JFNsWi8ZZtyT1af5 max invoke call depth 4                  active since slot 41904000
EnvhHCLvg55P7PDtbvR1NwuTuAeodqpusV3MR5QEK8gs instructions sysvar                      active since slot 41040000
FtjnuAtJTWwX3Kx9m24LduNEhzaGuuPfDW6e14SX2Fy5 rent fixes (#10206, #10468, #11342)      inactive
GaBtBJvmS4Arjj5W1NmFcyvPjsHN38UGYDq2MDwbs9Qu pico-inflation                           inactive
HRe7A6aoxgjKzdjbBv6HTy7tJ4YWqE6tVmYCGho6S9Aq ristretto multiply syscall               inactive
HxvjqDSiF5sYdSYuCXsUnS8UeAoWsMT9iGoFP8pgV1mB compute budget balancing                 active since slot 41904000

Feature activation is not allowed at this time
```

> E3PHP7w8kB7np3CTQ1qQ2tW3KCtjRSXBQgW9vM2mWv2Y secp256k1 program                        active since slot 41040000

because this is already enabled on mainnet-beta and these handling differencies before and after this pr create different transaction results?

mvines:
Yep this affects banking_stage.  Needs a feature gate :(

sdk/src/secp256k1.rs
```diff
@@ -72,7 +72,7 @@ pub fn verify_eth_addresses(
 
         // Parse out signature
         let signature_index = offsets.signature_instruction_index as usize;
-        if signature_index > instruction_datas.len() {
+        if signature_index >= instruction_datas.len() {
```
ryoqun, line 75: I also think adding a test for this won't hurt us, considering this is **the signature verification code**. Maybe this should be easy?

ryoqun:
Also, considering we found this on the wild, maybe should we feed more bunch of eth sigs picked from the real eth network into this code to expose more bugs?

ryoqun:
> Also, considering we found this on the wild, maybe should we feed more bunch of eth sigs picked from the real eth network into this code to expose more bugs?

So that we can pack bunch of any bugs under a single feature gate. :)

jstarry:
Wouldn't this cause a panic? I don't think it needs a feature gate. I caught it while reading code, don't think it's happened in the wild

sdk/src/secp256k1.rs
sakridge: This actually a more complete fix and adds tests:
https://github.com/solana-labs/solana/pull/13026

sakridge (changes requested):
needs another fix as in: https://github.com/solana-labs/solana/pull/13026

PR #13016, opened 2020-10-20
Explorer: speed up cluster stat loading

jstarry:
#### Problem
Cluster stats are not shown until data has been fetched twice

#### Summary of Changes
Adjust reducer status calculation so that data is shown right away

Fixes #


oJshua (approved):

oJshua:
@jstarry thanks, I had noticed it was kind of slow to start.

PR #13018, opened 2020-10-20
feat: disable loader rate limiter for non solana endpoints

jstarry:
#### Problem
The loader tries to avoid rate limiting by sleeping after each transaction send but rate limiting is only enabled on solana.com endpoints.

#### Summary of Changes
- Use `singleGossip` commitment instead of `single` for load instructions
- Disabled transaction batch awaits since we don't have to worry about `AccountInUse` errors anymore
- Added a simple check to disable rate limiting when the endpoint is not hosted on solana.com

The ideal change is to react to 429 with a connection scoped rate limiter, leaving that for the future.

Fixes #

### DIFF
--- a/web3.js/src/loader.js
+++ b/web3.js/src/loader.js
@@ -67,7 +67,7 @@ export class Loader {
       // Fetch program account info to check if it has already been created
       const programInfo = await connection.getAccountInfo(
         program.publicKey,
-        'single',
+        'singleGossip',
       );
 
       let transaction: Transaction | null = null;
@@ -127,7 +127,7 @@ export class Loader {
           transaction,
           [payer, program],
           {
-            commitment: 'single',
+            commitment: 'singleGossip',
             skipPreflight: true,
           },
         );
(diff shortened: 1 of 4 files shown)


jstarry:
Did a quick test and here's a rough breakdown of the improvements:

19.2s - current web3
17.7s - switch from single to singleGossip
15.9s - disable batch waiting
3.1s - disable rate limiting for non solana.com endpoints

web3.js/src/loader.js
```diff
@@ -169,22 +169,15 @@ export class Loader {
...
-      await sleep(1000 / REQUESTS_PER_SECOND);
-
-      // Run up to 8 Loads in parallel to prevent too many parallel transactions from
-      // getting retried due to AccountInUse errors.
-      //
-      // TODO: 8 was selected empirically and should probably be revisited
-      if (transactions.length === 8) {
-        await Promise.all(transactions);
-        transactions = [];
+      if (connection._rpcEndpoint.includes('solana.com')) {
```
mvines, line 178: 🙈 - looking forward to a more dynamic 429-response based approach :)

mvines (approved):

web3.js/src/loader.js
jstarry: Will get to it soon, here's an issue: https://github.com/solana-labs/solana/issues/13061

PR #13027, opened 2020-10-20
Force unset CARGO to use correct version of cargo

joncinque:
#### Problem

If `build.sh` is run from within `cargo` on Linux environments, it can fail because CARGO is set to the wrong executable.

#### Summary of Changes

Force CARGO to be unset

Fixes https://github.com/solana-labs/solana-program-library/issues/646

### DIFF
--- a/sdk/bpf/rust/build.sh
+++ b/sdk/bpf/rust/build.sh
@@ -34,6 +34,10 @@ export RUSTFLAGS="
     -C link-arg=-no-threads \
     -C linker=$bpf_sdk/dependencies/llvm-native/bin/ld.lld"
 
+# CARGO may be set if build.sh is run from within cargo, causing
+# incompatibilities between cargo and xargo versions
+unset CARGO
+
 # Setup xargo
 export XARGO_HOME="$bpf_sdk/dependencies/xargo"
 export XARGO_RUST_SRC="$bpf_sdk/dependencies/rust-bpf-sysroot/src"


sdk/bpf/rust/build.sh
```diff
@@ -34,6 +34,10 @@ export RUSTFLAGS="
     -C link-arg=-no-threads \
     -C linker=$bpf_sdk/dependencies/llvm-native/bin/ld.lld"
 
+# CARGO may be set if build.sh is run from within cargo, causing
+# incompatibilities between cargo and xargo versions
+unset CARGO
```
mvines, line 39: How does this manifest?   I don't see where in https://github.com/japaric/xargo/ it looks for the CARGO env var

jackcmay: It's probably set by cargo itself

joncinque: https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates this is done from cargo first

mvines: and then xargo uses cargo internally?  What cargo is xargo wanting to use?   Can I please get a link to the actual error that occurs?

joncinque: I don't know the internals of xargo fully, but it seems that it's calling cargo.  When CARGO is set, it uses that one, causing some potential incompatibilities.  From Jack's PR to test this out in CI, you can see the error at the end: https://travis-ci.com/github/solana-labs/solana-program-library/jobs/402691047

The key is `error: unknown codegen option: embed-bitcode`, which is a codegen flag added in rust 1.45.  Since a wrong / newer version of cargo is being used to build dependencies, it's passing the default for that flag to the older rustc from bpf-sdk, which is at 1.39, causing the build failure.

jackcmay: @joncinque unsetting `CARGO` in `do.sh` fixed the issue in the SPL CI.  I'm going to close that PR and we can open a new once these changes make it into a release to pull in a newer bpf-sdk in spl

joncinque: Sounds good, I'll merge this once CI passes

PR #13073, opened 2020-10-21
scans crds table in parallel for finding old labels

behzadnouri:
#### Problem
From runtime profiles, the majority time of ClusterInfo::handle_purge
https://github.com/solana-labs/solana/blob/0776fa05c/core/src/cluster_info.rs#L1605-L1626
is spent scanning crds table finding old labels:
https://github.com/solana-labs/solana/blob/0776fa05c/core/src/crds.rs#L175-L197

This can be done in parallel given that gossip thread-pool:
https://github.com/solana-labs/solana/blob/0776fa05c/core/src/cluster_info.rs#L1637-L1641
is idle when handle_purge is invoked:
https://github.com/solana-labs/solana/blob/0776fa05c/core/src/cluster_info.rs#L1681

#### Summary of Changes
Updated the code to use rayon parallel iterators.
For a crds table roughly the size of current mainnet, I see ~4x speed up on my machine with 4 cores.

### DIFF
--- a/core/benches/crds.rs
+++ b/core/benches/crds.rs
@@ -16,26 +16,16 @@ fn bench_find_old_labels(bencher: &mut Bencher) {
     let thread_pool = ThreadPoolBuilder::new().build().unwrap();
     let mut rng = thread_rng();
     let mut crds = Crds::default();
-    let mut num_inserts = 0;
     let now = CRDS_GOSSIP_PULL_CRDS_TIMEOUT_MS + CRDS_GOSSIP_PULL_CRDS_TIMEOUT_MS / 1000;
-    for _ in 0..50_000 {
-        if crds
-            .insert(CrdsValue::new_rand(&mut rng), rng.gen_range(0, now))
-            .is_ok()
-        {
-            num_inserts += 1;
-        }
-    }
-    assert_eq!(num_inserts, 50_000);
+    std::iter::repeat_with(|| (CrdsValue::new_rand(&mut rng), rng.gen_range(0, now)))
+        .take(50_000)
+        .for_each(|(v, ts)| assert!(crds.insert(v, ts).is_ok()));
     let mut timeouts = HashMap::new();
     timeouts.insert(Pubkey::default(), CRDS_GOSSIP_PULL_CRDS_TIMEOUT_MS);
-    let mut fail = 0;
     bencher.iter(|| {
-        let out = crds.find_old_labels(&thread_pool, now, &timeouts).len();
-        if out < 10 || out > 250 {
-            fail += 1;
-        }
+        let out = crds.find_old_labels(&thread_pool, now, &timeouts);
+        assert!(out.len() > 10);
+        assert!(out.len() < 250);
         out
     });
-    assert_eq!(fail, 0);
 }
--- a/core/benches/crds.rs
+++ b/core/benches/crds.rs
@@ -1,0 +1,41 @@
+#![feature(test)]
+
+extern crate test;
+
+use rand::{thread_rng, Rng};
+use rayon::ThreadPoolBuilder;
+use solana_core::crds::Crds;
+use solana_core::crds_gossip_pull::CRDS_GOSSIP_PULL_CRDS_TIMEOUT_MS;
+use solana_core::crds_value::CrdsValue;
+use solana_sdk::pubkey::Pubkey;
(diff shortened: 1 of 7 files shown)


core/benches/crds.rs
```diff
@@ -0,0 +1,41 @@
...
+    let mut rng = thread_rng();
+    let mut crds = Crds::default();
+    let mut num_inserts = 0;
+    let now = CRDS_GOSSIP_PULL_CRDS_TIMEOUT_MS + CRDS_GOSSIP_PULL_CRDS_TIMEOUT_MS / 1000;
+    for _ in 0..50_000 {
+        if crds
+            .insert(CrdsValue::new_rand(&mut rng), rng.gen_range(0, now))
+            .is_ok()
+        {
+            num_inserts += 1;
```
carllin, line 26: Can we remove the `num_inserts` and just panic if 
`crds.insert(CrdsValue::new_rand(&mut rng), rng.gen_range(0, now)).is_err()`

core/src/crds_gossip_pull.rs
```diff
@@ -537,24 +537,21 @@ impl CrdsGossipPull {
...
-            .collect();
-        let ret = purged.len();
-        self.purged_values.append(&mut purged);
-        ret
+        let num_purged_values = self.purged_values.len();
+        self.purged_values.extend(
+            crds.find_old_labels(thread_pool, now, timeouts)
+                .into_iter()
+                .filter_map(|label| {
+                    let val = crds.remove(&label)?;
```
carllin, line 550: is this `remove() `guaranteed to return `Some` because no other thread is touching the table (I think `purge` is happening under the gossip write lock). If so can we unwrap here()?

core/benches/crds.rs
```diff
@@ -0,0 +1,41 @@
...
+        }
+    }
+    assert_eq!(num_inserts, 50_000);
+    let mut timeouts = HashMap::new();
+    timeouts.insert(Pubkey::default(), CRDS_GOSSIP_PULL_CRDS_TIMEOUT_MS);
+    let mut fail = 0;
+    bencher.iter(|| {
+        let out = crds.find_old_labels(&thread_pool, now, &timeouts).len();
+        if out < 10 || out > 250 {
+            fail += 1;
```
carllin, line 36: same as above, I think we can panic here if fail occurs.

core/benches/crds.rs
behzadnouri: done

core/benches/crds.rs
behzadnouri: done

core/src/crds_gossip_pull.rs
behzadnouri: You are right, but there is nothing locally here that guarantees this is the case and will stay so in the future. For example, we may later in the future want to switch to a concurrent hash-map which does not require explicit locking. But then someone should remember that the code here was relying on single threaded access to the crds table. So I think it would be safer and more forward compatible if we do not rely on that assumption here.

carllin: @behzadnouri hehe I was actually thinking along the same scenario, i.e. where we switch to a concurrent hash-map. I wondering if it would be better to add an assertion here in that case as well, so that we remember to enforce the guarantee  that with respect to a particular label,`check timestamp->remove` is an atomic operation even in a concurrent hashmap implementation. Otherwise, for instance, an insertion might add a newer version of that label and then we remove the newer label by accident.

behzadnouri: That is a fair point, and might be the right direction to go. But my suggestion is that, if we want to make that change, lets do it separately in its own commit so that this commit implements [the same logic as the existing code](https://github.com/solana-labs/solana/blob/368aeb2ce/core/src/crds_gossip_pull.rs#L547-L553), but only in parallel.

behzadnouri: Another thing is that `crds.remove(&label) == None` is not necessarily an error. For example hypothetically if crds table was thread-safe, then 2 threads could be simultaneously doing the same thing; then either one which removes the old value from the table and store it in `purged_values` has already done the job, and the other one can skip.

carllin: makes sense to me!

carllin (approved):

Issue #13079, opened 2020-10-22
RPC: simulate_transaction needs to perform precompile verification


oJshua:
#### Problem
While attempting to simulate a secp256k1 instruction transaction, it was discovered that an invalid secp256k1 instruction won't be handled due to simulate_transaction's lack of precompile verification.

#### Proposed Solution
Include the lines: https://github.com/solana-labs/solana/blob/b6bfed64cb159ee67bb6bdbaefc7f833bbed3563/core/src/rpc.rs#L2281-L2283 in simulate_transaction.

PR #13080, opened 2020-10-22
add precompile verification to simulate_transaction

oJshua:
#### Problem
While attempting to simulate a secp256k1 instruction transaction, it was discovered that an invalid secp256k1 instruction won't be handled due to simulate_transaction's lack of precompile verification.

#### Summary of Changes
Introduces precompile verification to simulate_transaction call.

Fixes https://github.com/solana-labs/solana/issues/13079

### DIFF
--- a/core/src/rpc.rs
+++ b/core/src/rpc.rs
@@ -1267,6 +1267,18 @@ impl JsonRpcRequestProcessor {
     }
 }
 
+fn verify_transaction(transaction: &Transaction) -> Result<()> {
+    if transaction.verify().is_err() {
+        return Err(RpcCustomError::TransactionSignatureVerificationFailure.into());
+    }
+
+    if let Err(e) = transaction.verify_precompiles() {
+        return Err(RpcCustomError::TransactionPrecompileVerificationFailure(e).into());
+    }
+
+    Ok(())
+}
+
 fn verify_filter(input: &RpcFilterType) -> Result<()> {
     input
         .verify()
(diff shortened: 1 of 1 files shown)


core/src/rpc.rs
```diff
@@ -2322,6 +2322,10 @@ impl RpcSol for RpcSolImpl {
             Ok(())
         };
 
+        if let Err(e) = transaction.verify_precompiles() {
```
mvines, line 2325: I think this should be under the `if config.sig_verify` condition above

oJshua: Roger!

core/src/rpc.rs
```diff
@@ -2317,7 +2317,11 @@ impl RpcSol for RpcSolImpl {
         let (_, transaction) = deserialize_transaction(data, encoding)?;
 
         let mut result = if config.sig_verify {
-            transaction.verify()
+            if let Err(e) = transaction.verify() {
```
mvines, line 2320: Let's use the exact form as `send_transaction`:
```
            if transaction.verify().is_err() {
                return Err(RpcCustomError::TransactionSignatureVerificationFailure.into());
            }
            if let Err(e) = transaction.verify_precompiles() {
            }
```
No need for possible return value differences between these two APIs.  Consider moving the verification into a shared function even

oJshua: Okay, good plan

mvines (approved):

PR #12110, opened 2020-09-08
Prevent unbound memory growth by blockstore_processor

ryoqun:
#### Problem

#11727 wasn't enough. When replaying large number of slots, the memory growth is unbound at relatively quick rate.

#### Summary of Changes

Free any garbage resource laying around AccountsDB and really `Drop::drop()` banks by cleaning them from `all_banks` too.

#### Context

found via capitalization work.

PR #13089, opened 2020-10-22
cleanup accounts shrink debug message

sakridge:
#### Problem

Shrink message could be more informative and better formatted.

#### Summary of Changes

Add slot and fix `<alive> / <stored` to make sense with `(forced)` optional at the end.

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -985,7 +985,8 @@ impl AccountsDB {
                 }
                 if alive_count == stored_count && stores.values().len() == 1 {
                     trace!(
-                        "shrink_stale_slot: not able to shrink at all{}: {} / {}",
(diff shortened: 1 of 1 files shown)


ryoqun (approved):
lgtm!

runtime/src/accounts_db.rs
```diff
@@ -985,15 +985,17 @@ impl AccountsDB {
                 }
                 if alive_count == stored_count && stores.values().len() == 1 {
                     trace!(
-                        "shrink_stale_slot: not able to shrink at all{}: {} / {}",
```
ryoqun, line 988: err, last minute change at my #12110 broke this log... ;)

ryoqun:
>  @ryoqun ryoqun added v1.3 v1.4 labels now 

I quickly checked `git log` and it seems that this pr should be backported to v1.3 and v1.4 as well.

PR #12989, opened 2020-10-19
Separate the "program" feature of `solana-sdk` into a new crate called `solana-program`

mvines:
The "program" feature of `solana-sdk` needs to go away.  

Source backwards compatibility is retained by making `solana-sdk` re-export `solana-program` definitions, and the `solana-sdk` "program" feature becomes vestigial. 

This work will allow a program to declare `solana-program` as a `[dependency]` and then the full-blown `solana-sdk` as a `[dev-dependency]`, enabling access to useful things like `BanksClient` directly from program unit tests.

PR #13094, opened 2020-10-23
cargo-build-bpf: Remove legacy program feature support

mvines:
Once https://github.com/solana-labs/solana/pull/12989 lands, `cargo-build-bpf` can drop support for the detection of non-ported program crates that require `--no-default-features --features program`.   This was added just as temporary convenience and it's reasonable to require that `cargo-build-bpf` users use the new style of depending on `solana-program-sdk` instead of `solana-sdk`


mvines:
Expected to fail until #12989 lands

PR #13112, opened 2020-10-23
ci: Add downstream project build testing

mvines:
It's not nice to break downstream projects.  Add a new job in the CI build pipeline that fetches and builds example-helloworld, spl-token, and serum-dex against the PR.


mvines:
Serum DEX build is expected to fail until this downstream PR merged: https://github.com/project-serum/serum-dex/pull/39

PR #13113, opened 2020-10-23
marks pull request creation time only once per peer

behzadnouri:
#### Problem
mark_pull_request_creation time requires an exclusive lock on gossip:
https://github.com/solana-labs/solana/blob/16944e218/core/src/cluster_info.rs#L1547-L1548
Current code is redundantly marking each peer once for each request.
There are at most only 2 unique peers, whereas there are hundreds of
requests per each. So the lock is acquired hundreds of time longer than
necessary.

#### Summary of Changes
The commit de-duplicates peers and acquires the lock on gossip only once.

### DIFF
--- a/core/src/cluster_info.rs
+++ b/core/src/cluster_info.rs
@@ -1541,6 +1541,8 @@ impl ClusterInfo {
         self.stats
             .new_pull_requests_count
             .add_relaxed(pulls.len() as u64);
+        // There are at most 2 unique peers here: The randomly
+        // selected pull peer, and possibly also the entrypoint.
         let peers: Vec<Pubkey> = pulls.iter().map(|(peer, _, _, _)| *peer).dedup().collect();
         {
             let mut gossip =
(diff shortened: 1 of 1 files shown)


core/src/cluster_info.rs
```diff
@@ -1541,11 +1541,17 @@ impl ClusterInfo {
         self.stats
             .new_pull_requests_count
             .add_relaxed(pulls.len() as u64);
+        let peers: Vec<Pubkey> = pulls.iter().map(|(peer, _, _, _)| *peer).dedup().collect();
```
carllin, line 1544: good point! How about a comment/assert that there should only be <= 2 peers (entrypoint + randomly selected pull peer), for future reference?

carllin (dismissed):

core/src/cluster_info.rs
behzadnouri: added comment

PR #13120, opened 2020-10-23
Add bank timestamp bounding

CriesofCarrots:
#### Problem
The multiplication in the `calculate_stake_weighted_timestamp` method allows a node with very small stake to have an outsize effect on the resulting timestamp.

#### Summary of Changes
- Add a bounded calculation method that selects the stake-weighted media timestamp estimate, limited to 25% difference from the expected elapsed time since the start of the epoch.
- Also prevent block times from ever going backward relative to their ancestor blocks.
- Feature-gate these changes.

### DIFF
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -1114,7 +1114,16 @@ impl Bank {
                         ("from_genesis", unix_timestamp, i64),
                         ("corrected", timestamp_estimate, i64),
                     );
-                    unix_timestamp = timestamp_estimate
+                    unix_timestamp = timestamp_estimate;
+
+                    let ancestor_timestamp = self.clock().unix_timestamp;
+                    if self
+                        .feature_set
+                        .is_active(&feature_set::timestamp_bounding::id())
+                        && timestamp_estimate < ancestor_timestamp
+                    {
+                        unix_timestamp = ancestor_timestamp;
+                    }
                 }
             }
         }
@@ -9921,7 +9930,7 @@ mod tests {
             voting_keypair,
             ..
         } = create_genesis_config_with_leader(5, &leader_pubkey, 3);
-        let bank = Bank::new(&genesis_config);
+        let mut bank = Bank::new(&genesis_config);
         assert_eq!(
             bank.clock().unix_timestamp,
             bank.unix_timestamp_from_genesis()
@@ -9973,6 +9982,22 @@ mod tests {
         assert_eq!(
             bank.clock().unix_timestamp,
             bank.unix_timestamp_from_genesis() + 1
+        );
+
+        // Timestamp cannot go backward from ancestor Bank to child
+        bank = new_from_parent(&Arc::new(bank));
+        update_vote_account_timestamp(
+            BlockTimestamp {
+                slot: bank.slot(),
+                timestamp: bank.unix_timestamp_from_genesis() - 1,
+            },
+            &bank,
+            &voting_keypair.pubkey(),
+        );
+        bank.update_clock(None);
+        assert_eq!(
+            bank.clock().unix_timestamp,
+            bank.unix_timestamp_from_genesis()
         );
     }
 
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -60,7 +60,7 @@ use solana_sdk::{
     slot_hashes::SlotHashes,
     slot_history::SlotHistory,
     stake_weighted_timestamp::{
-        calculate_stake_weighted_timestamp, EstimateType, TIMESTAMP_SLOT_RANGE,
+        calculate_stake_weighted_timestamp, EstimateType, DEPRECATED_TIMESTAMP_SLOT_RANGE,
     },
     system_transaction,
     sysvar::{self},
@@ -1489,7 +1489,13 @@ impl Bank {
             .filter_map(|(pubkey, (_, account))| {
                 VoteState::from(&account).and_then(|state| {
                     let timestamp_slot = state.last_timestamp.slot;
-                    if self.slot().checked_sub(timestamp_slot)? <= TIMESTAMP_SLOT_RANGE as u64 {
+                    if (self
+                        .feature_set
+                        .is_active(&feature_set::timestamp_bounding::id())
+                        && self.ancestors.contains_key(&timestamp_slot))
+                        || self.slot().checked_sub(timestamp_slot)?
+                            <= DEPRECATED_TIMESTAMP_SLOT_RANGE as u64
+                    {
                         Some((
                             pubkey,
                             (state.last_timestamp.slot, state.last_timestamp.timestamp),
@@ -9694,7 +9700,7 @@ mod tests {
         let validator_vote_keypairs1 = ValidatorVoteKeypairs::new_rand();
         let validator_keypairs = vec![&validator_vote_keypairs0, &validator_vote_keypairs1];
         let GenesisConfigInfo {
-            genesis_config,
+            mut genesis_config,
             mint_keypair: _,
             voting_keypair: _,
         } = create_genesis_config_with_vote_accounts(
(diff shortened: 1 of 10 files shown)


runtime/src/feature_set.rs
```diff
@@ -93,6 +97,7 @@ lazy_static! {
         (cumulative_rent_related_fixes::id(), "rent fixes (#10206, #10468, #11342)"),
         (pubkey_log_syscall_enabled::id(), "pubkey log syscall"),
         (pull_request_ping_pong_check::id(), "ping-pong packet check #12794"),
+        (timestamp_bounding::id(), "add timestamp-correction bounding"),
```
mvines, line 100:
```suggestion
        (timestamp_bounding::id(), "add timestamp-correction bounding #13120"),
```

sdk/src/stake_weighted_timestamp.rs
```diff
@@ -5,15 +5,46 @@ use solana_sdk::{
...
+use std::{
+    collections::{BTreeMap, HashMap},
+    time::Duration,
+};
 
 pub const TIMESTAMP_SLOT_RANGE: usize = 16;
 
+pub enum EstimateType {
+    Bounded,
+    Unbounded,
```
mvines, line 17:
```suggestion
    Unbounded, // Deprecated.  Remove in the Solana v1.6.0 timeframe
```

sdk/src/stake_weighted_timestamp.rs
```diff
@@ -36,6 +67,48 @@ pub fn calculate_stake_weighted_timestamp(
     }
 }
 
+fn calculate_bounded_stake_weighted_timestamp(
```
mvines, line 70: @sconybeare - care to review this function?

sdk/src/stake_weighted_timestamp.rs
```diff
@@ -36,6 +67,48 @@ pub fn calculate_stake_weighted_timestamp(
...
+    let mut stake_accumulator = 0;
+    let mut stake_weighted_timestamps_sum = 0;
+    let mut included_stake = 0;
+    // Only include timestamp estimates from the middle 1/3 stake
+    for (timestamp, stake) in stake_per_timestamp.iter() {
+        stake_accumulator += *stake;
+        if stake_accumulator <= (total_stake / 3) as u128 {
+            continue;
+        }
+        stake_weighted_timestamps_sum += *timestamp as u128 * *stake as u128;
```
mvines, line 103: Even in the middle 1/3 of the stake, I think we should throw away outliers.  I can still sneak in a i64::MAX, perhaps by accident, and cause havoc.  Require timestamps be within an hour of the median?  (even an hour is absurdly generious given all nodes should be synced with NTP time)

sdk/src/stake_weighted_timestamp.rs
CriesofCarrots: Actually, hold off a bit longer. I'm going to rework this tomorrow based on our subsequent discussion

sdk/src/stake_weighted_timestamp.rs
CriesofCarrots: Agreed. @aeyakovenko proposed tracking the timestamp at the beginning of the epoch and bounding based on the poh estimate from the beginning of the epoch. This seems a little safer to me than a set time-delta from median.

CriesofCarrots:
This is ready for a proper look, if you will oblige @mvines

runtime/src/bank.rs
```diff
@@ -1091,26 +1093,78 @@ impl Bank {
...
+            let (estimate_type, epoch_start_timestamp) = if self
+                .feature_set
+                .is_active(&feature_set::timestamp_bounding::id())
+            {
+                let timestamp_bounding_activation_slot = self
+                    .get_account(&feature_set::timestamp_bounding::id())
+                    .as_ref()
+                    .and_then(|account| Feature::from_account(account))
+                    .and_then(|feature| feature.activated_at)
+                    .expect("feature has activated");
```
CriesofCarrots, line 1111: Is there an api that does this more elegantly?

sdk/src/stake_weighted_timestamp.rs
CriesofCarrots: @sconybeare, here is the updated bounding function

runtime/src/bank.rs
aeyakovenko: tendermint + cosmos just use a plain median.  how are you ensuring we have 2/3+ samples for any slot?  there is no guarantee that we have that many votes

CriesofCarrots: @aeyakovenko no, we aren't ensuring we have 2/3+ samples for each slot. Do we need to? I think that could be done by extending the `TIMESTAMP_SLOT_RANGE` to include slots back to the root, since we should always have votes from 2/3+ on that slot.

Either way, it still seems like a stake-weighted median is better than a flat median to minimize the possible effect of low-staked validators.

garious:
This looks good to me, but can you add something to docs.solana.com that explains the fields in the clock sysvar in excruciating detail?

runtime/src/bank.rs
aeyakovenko: could we use all the votes in Tower?  that would ensure that the sample is based on rooted data

aeyakovenko: and to summarize, do we use a weighted 1/3 median?  so drop 1/3 highs and 1/3 lows?

CriesofCarrots: > could we use all the votes in Tower? that would ensure that the sample is based on rooted data

We currently don't have access to Tower from the Bank, so that'd require a bit of re-architecting. Not sure I follow why it needs to be rooted data, since the votes will all be in ancestors Banks. Can you tell me more?

As I mentioned above, we can use Bank::ancestors to include all votes/timestamps since the current root, which I think should guarantee 2/3+ stake.

> and to summarize, do we use a weighted 1/3 median? so drop 1/3 highs and 1/3 lows?

I had that written originally, but I updated to just use the stake-weighted median value, so:  the timestamp at which 50% of the stake estimates a greater-or-equal timestamp and 50% of the stake estimates a lesser-or-equal timestamp.

CriesofCarrots: @mvines , while this thread has gone off-trail, I did actually want to know if there is a simpler/better way to get an activated feature's activation slot...

mvines: No better way right now, this is the first case where the activation slot matters.   Can we make it not matter? :)

If no, then I think FeatureSet::active should change from a `HashSet<Pubkey>` to `HashMap<Pubkey,Slot>`, avoid ing all this work on every single new Bank and setting things up for the next time the activation slot matters

CriesofCarrots: The ways we could make it not matter involve making assumptions about the realistic range of `epoch_start_timestamp` vs `Bank::get_unused_from_slot`, which I'm not super excited about.

I'll start on that FeatureSet::active edit, if okay with you.

mvines: sgtm!

CriesofCarrots:
@garious added: https://github.com/solana-labs/solana/pull/13120/commits/ad423ad5eb7678dea40ee3081022179978bcd88e

CriesofCarrots:
@garious @mvines anything else you'd like to see here?
Pushes today were just to rebase on master bank.rs changes

aeyakovenko:
@CriesofCarrots can you add a design doc, it think this is important enough to try to keep documented.

CriesofCarrots:
> @CriesofCarrots can you add a design doc, it think this is important enough to try to keep documented.

Certainly, added. There's also documentation in the Clock sysvar section, as per @garious 's request.

aeyakovenko (approved):

mvines (approved):
great docs!

mvines:
@CriesofCarrots - I'll merge this to get the v1.4 backport going

CriesofCarrots:
> @CriesofCarrots - I'll merge this to get the v1.4 backport going

@mvines , thank you! And thanks for handling that last conflict/rebase!

PR #13129, opened 2020-10-24
feat: allow setting explicit fee payer for transaction

jstarry:
#### Problem
When transactions are serialized and a fee payer cannot be determined, an error is thrown that says: `"Transaction feePayer required"`. The transaction API only has a getter for `feePayer` and so it's unclear to devs on how to specify one.

#### Summary of Changes
- Allow setting `transaction.feePayer`
- Deprecate `setSigners`
- Deprecate and add warning for when devs sign a transaction with an account that isn't required to sign
- Fix some type definitions

Fixes #


mvines (approved):

t-nelson (approved):
Looks great!  One step closer to parity with the Rust interface :tada:

PR #13175, opened 2020-10-26
check sysvar id for AccountInfo

jackcmay:
#### Problem

When deserializing a Sysvar from AccountInfo the Sysvar's id is not validated to match

#### Summary of Changes

Check the AccountInfo's id to ensure it matches the expected sysvar.

Fixes #


ryoqun:
@jackcmay how about backporting this to v1.3 as well? I think every third party programs are affected by this too. CC: @mvines

mvines:
v1.4 backport sgtm

PR #13199, opened 2020-10-27
Account for forward delay in transaction simulation

jstarry:
#### Problem
- Bank transaction simulation doesn't account for forwarding delays which means that a transaction could pass simulation but fail processing after forwarding

#### Summary of Changes
- Account for forwarding delay in simulation

Fixes #

### DIFF
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -39,7 +39,8 @@ use solana_sdk::{
     account::Account,
     clock::{
         Epoch, Slot, SlotCount, SlotIndex, UnixTimestamp, DEFAULT_TICKS_PER_SECOND,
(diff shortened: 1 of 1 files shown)


jstarry:
I just removed the "fix" for last valid slot and this PR is just about transaction simulation now

CriesofCarrots (dismissed):
Looks good. Could you add a comment about that adjustment? Maybe something about simulating worst-possible forwarding scenario

jstarry:
Added a comment

ryoqun:
(it seems this pr reduced its scope after I saw this for the first time; anyway meddling with `MAX_PROCESSING_AGE` and the like reminds me of #9423, well writing tests for it has been on my todo list for toooo long time. I haven't still get there...)

PR #13206, opened 2020-10-27
Make bad_bank_hash test faster

sakridge:
#### Problem

bad_bank_hash test is slow

#### Summary of Changes

Optimize, re-use account set and randomly sample from it for the store loop and reduce iterations. Reduces from about 2:30 into 30 seconds.

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -4441,35 +4441,59 @@ pub mod tests {
 
     #[test]
     fn test_bad_bank_hash() {
+        solana_logger::setup();
         use solana_sdk::signature::{Keypair, Signer};
         let db = AccountsDB::new(Vec::new(), &ClusterType::Development);
 
         let some_slot: Slot = 0;
         let ancestors: Ancestors = [(some_slot, 0)].iter().copied().collect();
 
-        for _ in 0..10_000 {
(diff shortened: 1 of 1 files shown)


runtime/src/accounts_db.rs
```diff
@@ -4441,35 +4441,59 @@ pub mod tests {
...
-                .map(|_| {
-                    let key = Keypair::new().pubkey();
-                    let lamports = thread_rng().gen_range(0, 100);
-                    let some_data_len = thread_rng().gen_range(0, 1000);
-                    let account = Account::new(lamports, some_data_len, &key);
-                    (key, account)
-                })
-                .collect();
-            let account_refs: Vec<_> = accounts_keys
+            (0..num_accounts).into_iter().for_each(|_| {
```
carllin, line 4471: Ran some quick maths, and I think we're expected to take about 138 samples to get 100 items, which isn't too bad.

Alternatively we could shuffle the `account_keys` vector on every iteration via 

```
use rand::seq::SliceRandom;
accounts_keys.shuffle(&mut thread_rng());
```

and take the first `num_accounts`. Curious, but might not be any better for small vectors like this though 😛

carllin (approved):

Issue #13212, opened 2020-10-28
Web3: SimulateTransaction does not consider CPI errors


michaelhly:
#### Problem
When executing token-swap's withdrawal method, I would occasionally hit this error:
```
    failed to send transaction: Transaction simulation failed: Error processing Instruction 0: custom program error: 0x1

      at Connection._callee53$ (node_modules/@solana/web3.js/src/connection.js:2860:13)
      at tryCatch (node_modules/regenerator-runtime/runtime.js:63:40)
      at Generator.invoke [as _invoke] (node_modules/regenerator-runtime/runtime.js:293:22)
      at Generator.next (node_modules/regenerator-runtime/runtime.js:118:21)
      at asyncGeneratorStep (node_modules/@babel/runtime/helpers/asyncToGenerator.js:3:24)
      at _next (node_modules/@babel/runtime/helpers/asyncToGenerator.js:25:9)
```

After checking node logs, it turns out this error was occurring inside a CPI:
```
ocalnet_1  | [2020-10-28T01:40:31.086748900Z INFO  solana_runtime::message_processor] BPF program consumed 4959 of 200000 units
localnet_1  | [2020-10-28T01:40:31.086864400Z INFO  solana_runtime::message_processor] BPF program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA success
localnet_1  | [2020-10-28T01:40:31.925527100Z INFO  solana_runtime::message_processor] Call BPF program 6aYDwejProFEqV12fQAedRL2Gv5P1EZhA8Y9kXnJeKo1
localnet_1  | [2020-10-28T01:40:31.926232900Z INFO  solana_runtime::message_processor] Program log: Instruction: Withdraw
localnet_1  | [2020-10-28T01:40:31.926649500Z INFO  solana_runtime::message_processor] Program log: Nonce: 255
localnet_1  | [2020-10-28T01:40:31.927363600Z INFO  solana_runtime::message_processor] Call BPF program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA
localnet_1  | [2020-10-28T01:40:31.927464800Z INFO  solana_runtime::message_processor] Program log: Instruction: Transfer
localnet_1  | [2020-10-28T01:40:31.927646100Z DEBUG solana_rbpf::vm] BPF instructions executed: 7498
[...]

PR #13213, opened 2020-10-28
feat: print program logs when transaction send fails

jstarry:
#### Problem
It's not clear when an instruction fails from an inner instruction because there is no log trace returned from `sendTransaction` in the `Connection` api

#### Summary of Changes
- Error log the program logs if the transaction fails and logs are present

```
Transaction simulation failed: Error processing Instruction 1: custom program error: 0x2 
    Call BPF program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA
    Program log: Instruction: InitializeAccount
    Program log: Error: Invalid Mint
    BPF program consumed 3579 of 200000 units
    BPF program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA failed: custom program error: 0x2
```

Fixes https://github.com/solana-labs/solana/issues/13212

### DIFF
--- a/web3.js/src/connection.js
+++ b/web3.js/src/connection.js
@@ -2857,6 +2857,14 @@ export class Connection {
     const unsafeRes = await this._rpcRequest('sendTransaction', args);
     const res = SendTransactionRpcResult(unsafeRes);
     if (res.error) {
+      if (res.error.data) {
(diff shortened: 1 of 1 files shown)


mvines (approved):
nice

PR #13224, opened 2020-10-28
feat: add support for inner instructions metadata

vtsurka:
#### Problem
Web3 does not support inner transactions meta
https://docs.solana.com/apps/jsonrpc-api#getconfirmedtransaction

#### Summary of Changes
Add information about inner transactions in rpc responses

### DIFF
--- a/web3.js/module.d.ts
+++ b/web3.js/module.d.ts
@@ -151,14 +151,9 @@ declare module '@solana/web3.js' {
     err: TransactionError | null;
   };
 
-  export type PartiallyDecodedInnerInstruction = {
-    index: number;
-    instructions: PartiallyDecodedInstruction[];
-  };
-
   export type ParsedInnerInstruction = {
     index: number;
-    instructions: (ParsedInstruction | PartiallyDecodedInnerInstruction)[];
+    instructions: (ParsedInstruction | PartiallyDecodedInstruction)[];
   };
 
   export type ParsedConfirmedTransactionMeta = {
--- a/web3.js/module.d.ts
+++ b/web3.js/module.d.ts
@@ -1,5 +1,4 @@
 declare module '@solana/web3.js' {
-  import {Buffer} from 'buffer';
   import * as BufferLayout from 'buffer-layout';
 
   // === src/publickey.js ===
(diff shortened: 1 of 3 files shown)


vtsurka:
`  export type ParsedTransaction = {
    signatures: Array<string>;
    message: {
      accountKeys: ParsedMessageAccount[];
      instructions: (ParsedInstruction | PartiallyDecodedInstruction)[];
      recentBlockhash: string;
    };
  };

  export type ParsedConfirmedTransaction = {
    slot: number;
    transaction: ParsedTransaction;
    meta: ParsedConfirmedTransactionMeta | null;
  };`

Is it correct to put meta strict to ParsedConfirmedTransactionMeta? when transaction.message.instructions could be both of type?

web3.js/module.d.ts
```diff
@@ -138,8 +137,28 @@ declare module '@solana/web3.js' {
     logs: Array<string> | null;
   };
 
+  export type PartiallyDecodedInnerInstruction = {
+    index: number;
+    instructions: PartiallyDecodedInstruction[];
+  };
```
jstarry, line 143: Oops, just noticed that this isn't quite right
```suggestion
  export type CompiledInnerInstruction = {
    index: number;
    instructions: CompiledInstruction[];
  };
```

web3.js/module.d.ts
```diff
@@ -138,8 +137,28 @@ declare module '@solana/web3.js' {
...
+    innerInstructions?: PartiallyDecodedInnerInstruction[];
+    preBalances: Array<number>;
+    postBalances: Array<number>;
+    logMessages?: Array<string>;
+    err: TransactionError | null;
+  };
+
+  export type ParsedInnerInstruction = {
+    index: number;
+    instructions: ParsedInstruction[];
```
jstarry, line 156:
```suggestion
    instructions: (ParsedInstruction | PartiallyDecodedInnerInstruction)[];
```

web3.js/src/connection.js
```diff
@@ -1126,6 +1158,60 @@ const ConfirmedTransactionMetaResult = struct.union([
   struct.pick({
     err: TransactionErrorResult,
     fee: 'number',
+    innerInstructions: struct.union(
+      struct.array([
+        struct({
+          index: 'number',
+          instructions: struct.array([
+            struct.union([
```
jstarry, line 1166: nit: unnecessary union

jstarry (commented):
> Is it correct to put meta strict to ParsedConfirmedTransactionMeta? when transaction.message.instructions could be both of types?

It is correct, but you just helped me realize that there are a few things that should be fixed. Hopefully it all makes sense with the ~above~ below comments

web3.js/module.d.ts
```diff
@@ -137,23 +137,28 @@ declare module '@solana/web3.js' {
...
 
+  export type PartiallyDecodedInnerInstruction = {
+    index: number;
+    instructions: PartiallyDecodedInstruction[];
+  };
+
   export type ParsedInnerInstruction = {
     index: number;
-    instructions: ParsedInstruction[];
+    instructions: (ParsedInstruction | PartiallyDecodedInnerInstruction)[];
```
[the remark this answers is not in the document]
jstarry: Apologies, I typo'd the previous suggestion. This is what it should be ^ `PartiallyDecodedInnerInstruction` is not needed

PR #13231, opened 2020-10-28
fix: broken inner instruction struct validator

vtsurka:
#### Problem
Failed build of web3 because connection missed & struct.union has wrong args
#### Summary of Changes
Import CompiledInstruction in connection.js & fix struct.union for inner instruction
Fixes #

### DIFF
--- a/web3.js/src/connection.js
+++ b/web3.js/src/connection.js
@@ -12,7 +12,7 @@ import {NonceAccount} from './nonce-account';
 import {PublicKey} from './publickey';
 import {MS_PER_SLOT} from './timing';
 import {Transaction} from './transaction';
-import {Message, CompiledInstruction} from './message';
+import {Message} from './message';
 import {sleep} from './util/sleep';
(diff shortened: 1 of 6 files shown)


web3.js/src/connection.js
```diff
@@ -12,7 +12,7 @@ import {NonceAccount} from './nonce-account';
 import {PublicKey} from './publickey';
 import {MS_PER_SLOT} from './timing';
 import {Transaction} from './transaction';
-import {Message} from './message';
+import {Message, CompiledInstruction} from './message';
```
jstarry, line 15: When running `npm run flow:watch` I get this error, can you fix?
```
Cannot import the type CompiledInstruction as a value. Use import type instead. [import-type-as-value]

     12│ import {PublicKey} from './publickey';
     13│ import {MS_PER_SLOT} from './timing';
     14│ import {Transaction} from './transaction';
     15│ import {Message, CompiledInstruction} from './message';
```

PR #13243, opened 2020-10-28
Add sol_log_bpf_compute_units syscall

mvines:
The runtime outputs a nice log message at the end of executing my program stating the amount of compute units consumed.   But now I want to reduce the usage and have no idea where to look.   My only option is to start carving out parts of my program and seeing what that does to the final compute units consumed.   This is painful.

But now I can pepper my program with `sol_log_consumption()`s tracing and easily see the compute units get consumed.

### DIFF
--- a/programs/bpf/c/src/sanity/sanity.c
+++ b/programs/bpf/c/src/sanity/sanity.c
@@ -18,5 +18,7 @@ extern uint64_t entrypoint(const uint8_t *input) {
   // program, no account keys or input data are expected but real
   // programs will have specific requirements so they can do their work.
   sol_log_params(&params);
+
+  sol_log_bpf_compute_units();
   return SUCCESS;
 }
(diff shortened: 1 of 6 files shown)


mvines:
@jackcmay - wdyt about also adding a `find_program_address()` syscall?  It will be common for programs to want to find an address instead of making the caller provide the bump seed.    Currently we run out of compute budget after ~84 iterations within the BPF `find_program_address()`

jackcmay:
> @jackcmay - wdyt about also adding a `find_program_address()` syscall? It will be common for programs to want to find an address instead of making the caller provide the bump seed. Currently we run out of compute budget after ~84 iterations within the BPF `find_program_address()`

Sure, no objection to this

jackcmay:
this lgtm

mvines:
@jackcmay - I renamed this to `sol_log_bpf_compute_units()` based on our discussion at #13245, to make it more clear that this log is for BPF-programs only

mvines:
Rolled this into #13286

PR #13245, opened 2020-10-28
Add Bank::set_bpf_compute_budget_override()

mvines:
From a program unit test environment, I want to be able to ratchet down the compute budget to avoid future regressions to my optimized program code.

### DIFF
--- a/programs/bpf/benches/bpf_loader.rs
+++ b/programs/bpf/benches/bpf_loader.rs
@@ -15,7 +15,7 @@ use solana_runtime::{
     genesis_utils::{create_genesis_config, GenesisConfigInfo},
     loader_utils::load_program,
     process_instruction::{
-        ComputeBudget, ComputeMeter, Executor, InvokeContext, Logger, ProcessInstruction,
+        BpfComputeBudget, ComputeMeter, Executor, InvokeContext, Logger, ProcessInstruction,
     },
 };
 use solana_sdk::{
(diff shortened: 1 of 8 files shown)


jackcmay:
Not too happy about moving the bpf_loader specific compute budget back into the bank.  Maybe instead we write a utility that digests the logs and reports the consumption which the test could then validate against?

mvines:
> Not too happy about moving the bpf_loader specific compute budget back into the bank.

Seems like we need to rename ComputeBudget/compute_budget to BpfComputeBudget/bpf_compute_budget.

ComputeBudget as it's currently named does not sound BPF-specific at all.  Ideally we could configure the BpfLoader *before* it's given to the bank/message_processor

I'm pretty opposed to log parsing to fail unit tests.  That's a pretty unergonomic solution vs. simply adjusting the compute limit upfront.

jackcmay:
Agreed, message parsing is fragile.  I like the idea of configuring the bpf loader loader and handing it off to the message processor, maybe as part of a "loaders" list

mvines:
Can I tempt you with the top commit?   It renames ComputeBudget to BpfComputeBudget, which I think is a major improvement to readability.  

I'm still adding `Bank::set_bpf_compute_budget_override()` in this PR, but at least it's now more clearly scoped.

jackcmay:
One of the ideas around ComputeBudget was that it was a concept that applied to whatever loader was running, so a generic syscall namespace would apply and not need to be bpf specific.  Underneath renaming the struct sounds fine since the current implementation is specific, but mainly because that's the only loader we have.  That's one of the reasons I wanted to move it into the loader, it then naturally becomes bpf loader specific and its scope is clear.

mvines:
Folding this into https://github.com/solana-labs/solana/pull/13286

PR #9490, opened 2020-04-14
Add append-vec subcommand & another env knob

ryoqun:
#### Problem

There is no handy way to display the content of `AppendVec` and there is no easy way to view historical old state of written accounts by `solana-validator`. Both are handy when debugging.

#### Summary of Changes

- Add such a subcommand to `ledger-tool` and such a environment variable, respectively.

Needed this while investigation toward to #9423 / #9357

PR #13254, opened 2020-10-29
Add store-tool

sakridge:
#### Problem

No standalone way to inspect account stores offline.

#### Summary of Changes

Add a tool to open account stores and print their contents.

Fixes #

### DIFF
--- a/runtime/store-tool/Cargo.toml
+++ b/runtime/store-tool/Cargo.toml
@@ -1,0 +1,22 @@
+[package]
+authors = ["Solana Maintainers <maintainers@solana.foundation>"]
+edition = "2018"
+name = "solana-store-tool"
+description = "Tool to inspect append vecs"
+version = "1.5.0"
+repository = "https://github.com/solana-labs/solana"
+license = "Apache-2.0"
+homepage = "https://solana.com/"
+publish = false
+
+[dependencies]
+log = { version = "0.4.8" }
+solana-logger = { path = "../../logger", version = "1.5.0" }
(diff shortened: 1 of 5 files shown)


runtime/store-tool/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+authors = ["Solana Maintainers <maintainers@solana.foundation>"]
+edition = "2018"
+name = "solana-store-tool"
+description = "Tool to inspect append vecs"
+version = "1.5.0"
+repository = "https://github.com/solana-labs/solana"
+license = "Apache-2.0"
+homepage = "https://solana.com/"
+
```
mvines, line 10: `publish = false`?

sakridge: added, thanks

runtime/store-tool/src/main.rs
```diff
@@ -0,0 +1,47 @@
+use clap::{crate_description, crate_name, value_t_or_exit, App, Arg};
+use log::*;
+use solana_runtime::append_vec::AppendVec;
+
+fn main() {
```
ryoqun, line 5: well, there is never-merged similar pr: https://github.com/solana-labs/solana/pull/9490/files#diff-fe20bb9062954767913766205f72e8eb5ae88909c0d2529ec706e27a3e0eb800R589

sakridge: and?

PR #13256, opened 2020-10-29
Update FeatureSet::active to include slot-activated

CriesofCarrots:
#### Problem
In some cases (at least [this one case](https://github.com/solana-labs/solana/pull/13120)), a Bank method needs to know the slot a feature activation occurred. Currently, this involves deserializing the Feature account, which is expensive if it needs to be done every slot.

#### Summary of Changes
Extend FeatureSet::active to include the slot an active feature was activated

I do not think this needs to be gated, as it doesn't change the behavior of current code. But would appreciate the sanity-check.

### DIFF
--- a/runtime/src/feature_set.rs
+++ b/runtime/src/feature_set.rs
@@ -133,7 +133,7 @@ impl FeatureSet {
     }
 
     pub fn activated_slot(&self, feature_id: &Pubkey) -> Option<Slot> {
-        self.active.get(feature_id).map(|slot| *slot)
+        self.active.get(feature_id).copied()
     }
 
     pub fn cumulative_rent_related_fixes_enabled(&self) -> bool {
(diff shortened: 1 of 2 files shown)


runtime/src/feature_set.rs
```diff
@@ -138,7 +143,7 @@ impl FeatureSet {
     /// All features enabled, useful for testing
     pub fn all_enabled() -> Self {
         Self {
-            active: FEATURE_NAMES.keys().cloned().collect(),
+            active: FEATURE_NAMES.keys().cloned().map(|key| (key, 0)).collect(),
```
CriesofCarrots, line 146: Any concerns about defaulting to zero here?

mvines (approved):
nice, this slide in pretty smoothly!

runtime/src/feature_set.rs
mvines: Nope, I think that's correct.  Enabled since genesis

PR #13264, opened 2020-10-29
fix: fix TypeError when confirmed block is not found

jstarry:
#### Problem
When a confirmed block is not found, the web3 sdk throws a TypeError error is not defined due to referencing the wrong "result"

#### Summary of Changes
- Fix bug

Fixes #


CriesofCarrots (approved):
Nice catch

PR #13265, opened 2020-10-29
Remove solana_sdk::builtins module

mvines:
Dead code. 💀


mvines:
Oops, not dead code!  But only used by bpf_loader so it can be inlined

PR #13285, opened 2020-10-29
fix: fallback to connection commitment when confirming transactions

jstarry:
#### Problem
When confirming transactions without an explicit commitment level, the web3 sdk falls back to `max` commitment. This behavior is not consistent to how any of the other `Connection` methods work. When an app chooses a particular commitment level for their `Connection`, the expectation should be that the commitment is applied unless explicitly overridden.

#### Summary of Changes
- Fallback to `Connection.commitment` rather than `max`

Fixes #

### DIFF
--- a/web3.js/src/connection.js
+++ b/web3.js/src/connection.js
@@ -2178,7 +2178,7 @@ export class Connection {
     assert(decodedSignature.length === 64, 'signature has invalid length');
 
     const start = Date.now();
-    const subscriptionCommitment: Commitment = commitment || this.commitment;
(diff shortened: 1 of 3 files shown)


jstarry:
This is technically a breaking change, but I would argue the previous behavior was unexpected. Would appreciate a second opinion on whether merging this is a good idea. It came up once from a dev in Discord and I just bumped into when updating the hello world example

mvines (dismissed):
lgtm, this isn't a breaking change, it's a bug fix. :)

PR #13286, opened 2020-10-29
Native/builtin programs now receive an InvokeContext

mvines:
By plumbing InvokeContext into native and builtin programs, we open the door for:
1) Feature gating *within* the program.  🎉 
2) Enabling CPI, which is needed for unit testing

Downsides:
* FeatureSet moves into sdk/  - ok not really a downside
* `runtime/src/process_instruction.rs` → `sdk/src/process_instruction.rs` -- not great but now that `solana-program` exists, this seems fine
* `runtime/src/bpf_test_utils.rs` gets inlined into `sdk/src/process_instruction.rs`.  These aren't BPF-specific test utils at all right now


sdk/src/process_instruction.rs
```diff
@@ -24,7 +24,6 @@ pub type LoaderEntrypoint = unsafe extern "C" fn(
     invoke_context: &dyn InvokeContext,
 ) -> Result<(), InstructionError>;
 
-pub type ProcessInstruction = fn(&Pubkey, &[KeyedAccount], &[u8]) -> Result<(), InstructionError>;
 pub type ProcessInstructionWithContext =
```
jackcmay, line 27: Why not just call this `ProcessInstruction` now since there only needs to be one?

mvines: will do!

runtime/src/bank.rs
```diff
@@ -10065,22 +10077,15 @@ mod tests {
...
             _data: &[u8],
-            _context: &mut dyn InvokeContext,
+            _invoke_context: &mut dyn InvokeContext,
         ) -> std::result::Result<(), InstructionError> {
             Ok(())
         }
-        assert!(!format!("{:?}", Entrypoint::Program(mock_process_instruction)).is_empty());
-        assert!(!format!("{:?}", Entrypoint::Loader(mock_ix_processor)).is_empty());
+        assert!(!format!("{:?}", Entrypoint::Program(mock_processor)).is_empty());
+        assert!(!format!("{:?}", Entrypoint::Loader(mock_processor)).is_empty());
```
jackcmay, line 10089: Do we need `Entrypoint::Loader` any longer?

mvines: I guess only to add typing to distinguish a loader from a program.   I can try pulling it out and see what that does to the code though?

jackcmay: Its originally there to handle the two entrypoints, now that there is one lets clean that up too

sdk/src/process_instruction.rs
mvines: I think there's value in keeping ProcessInstructionWithContext.
There's `solana_program::entrypoint::ProcessInstruction`, which has no context.   So keeping this as 
`ProcessInstructionWithContext` improves readability especially in files that use both types

runtime/src/bank.rs
mvines: @jackcmay - do you see any issue with deleting the `MessageProcessor::loaders` field entirely?   The only impact seems to be in `MessageProcessor::process_instruction`

jackcmay: If loaders and programs have the same entrypoint signature now then we don't need to make the distinction between the two anymore.  So yeah, we can probably roll `loaders` into `programs`

mvines: I just added b7d184c759754595aa560af76d5e5585c525f659 to merge them in MessageProcessor.  Once that passes CI, I'll make the larger changes need across the code base needed to de-Entrypoint

mvines: Looks good, Entrypoint is now gone

PR #13140, opened 2020-10-26
Cache account stores, flush from AccountsBackgroundService

carllin:
#### Problem
Storing same account multiple times in a slot, especially large accounts, is inefficient for storage

#### Summary of Changes
1) Introduce `AccountsCache` for accounts
2) Flush rooted slots from `AccountsBackgroundService` periodically
3) Flush slots on startup from `BlockstoreProcessor`

Based on initial validator test runs, so haven't seen much improvement in replay times and snapshot sizes, but may be a bigger issue if the same accounts are touched more often in the same slot.

Fixes #

PR #13294, opened 2020-10-29
Refactor store/update index function

carllin:
#### Problem
Store -> update index -> handle reclaims is a common pattern used by both regular account store and shrink.

#### Summary of Changes
Refactor this functionality into common functions shared with shrink and regular account store.

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -40,6 +40,7 @@ use solana_sdk::{
 };
 use std::convert::TryFrom;
 use std::{
+    boxed::Box,
     collections::{HashMap, HashSet},
     convert::TryInto,
     io::{Error as IOError, Result as IOResult},
@@ -106,6 +107,7 @@ pub(crate) type SlotStores = Arc<RwLock<HashMap<usize, Arc<AccountStorageEntry>>
 type AccountSlots = HashMap<Pubkey, HashSet<Slot>>;
 type AppendVecOffsets = HashMap<AppendVecId, HashSet<usize>>;
 type ReclaimResult = (AccountSlots, AppendVecOffsets);
+type StorageFinder<'a> = Box<dyn Fn(Slot, usize) -> Arc<AccountStorageEntry> + 'a>;
 
 trait Versioned {
     fn version(&self) -> u64;
@@ -408,6 +410,12 @@ struct FrozenAccountInfo {
     pub lamports: u64, // Account balance cannot be lower than this amount
 }
 
+#[derive(Default)]
+pub struct StoreAccountsTiming {
+    store_accounts_elapsed: u64,
+    update_index_elapsed: u64,
+    handle_reclaims_elapsed: u64,
+}
 // This structure handles the load/store of the accounts
 #[derive(Debug)]
 pub struct AccountsDB {
@@ -1089,10 +1097,8 @@ impl AccountsDB {
         let mut dead_storages = vec![];
         let mut find_alive_elapsed = 0;
         let mut create_and_insert_store_elapsed = 0;
-        let mut store_accounts_elapsed = 0;
-        let mut update_index_elapsed = 0;
-        let mut handle_reclaims_elapsed = 0;
         let mut write_storage_elapsed = 0;
+        let mut store_accounts_timing = StoreAccountsTiming::default();
         if aligned_total > 0 {
             let mut start = Measure::start("find_alive_elapsed");
             let mut accounts = Vec::with_capacity(alive_accounts.len());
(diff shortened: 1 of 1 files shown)


runtime/src/accounts_db.rs
```diff
@@ -2715,12 +2697,53 @@ impl AccountsDB {
...
+        write_version_producer: Option<Box<dyn Iterator<Item = u64>>>,
+    ) -> StoreAccountsTiming {
+        let storage_finder: StorageFinder<'a> = storage_finder
+            .unwrap_or_else(|| Box::new(move |slot, size| self.find_storage_candidate(slot, size)));
+
+        let write_version_producer: Box<dyn Iterator<Item = u64>> = write_version_producer
+            .unwrap_or_else(|| {
+                let mut current_version = self.bulk_assign_write_version(accounts.len());
+                Box::new(std::iter::from_fn(move || {
+                    let ret = current_version;
```
ryoqun, line 2730: @carllin Do you like fun rust ride?  :) Could you try `impl Fn(...)` instead of `Box<dyn Fn(...)>` like I played at #13345. What I'm worrying is that this closure will be called inside a tight loop. With `Box`, I think this incurrs dynamic dispatch for each invocation. I dunno how that affects the performance. Or rustc is smart enough to optimize away it. But I'd like to avoid that altogether if could by using `impl Fn` (which does static diapatch; also no heap allocation by not using `Box`). Abstraction with zero runtime cost. Beware, compiler- lang- support was rather edgy for `impl Fn`...

runtime/src/accounts_db.rs
```diff
@@ -2715,12 +2697,53 @@ impl AccountsDB {
...
+    fn store_accounts_custom<'a>(
+        &'a self,
+        slot: Slot,
+        accounts: &[(&Pubkey, &Account)],
+        hashes: &[Hash],
+        storage_finder: Option<StorageFinder<'a>>,
+        write_version_producer: Option<Box<dyn Iterator<Item = u64>>>,
+    ) -> StoreAccountsTiming {
+        let storage_finder: StorageFinder<'a> = storage_finder
+            .unwrap_or_else(|| Box::new(move |slot, size| self.find_storage_candidate(slot, size)));
```
ryoqun, line 2724: I think we can avoid `unwrap` here by moving this into its sole call-site: `store_accounts_default`. Also, this makes `store_accounts_default` hold the definition of default behavior, which is easier to read.

And, `store_account_custom` now can take `StorageFinder<'a>` not wrapped in `Option<_>`. :)

runtime/src/accounts_db.rs
```diff
@@ -2715,12 +2697,53 @@ impl AccountsDB {
...
+
+        let write_version_producer: Box<dyn Iterator<Item = u64>> = write_version_producer
+            .unwrap_or_else(|| {
+                let mut current_version = self.bulk_assign_write_version(accounts.len());
```
ryoqun, line 2728: we could similar for this `unwrap_or_else`: https://github.com/solana-labs/solana/pull/13294/files#r518311759

ryoqun:
@carllin yay for less duplicated code. Also, sorry for long delay. I think I crated enough homework for you until I get another free time to look. ;)

runtime/src/accounts_db.rs
carllin: yeah this would be true based only on this PR 😄 , but I'm actually using these options in a few places in my other work here:

https://github.com/solana-labs/solana/pull/13140/files#diff-1090394420d51617f3233275c2b65ed706b35b53b115fe65f82c682af8134a6fR1178-R1179

https://github.com/solana-labs/solana/pull/13140/files#diff-1090394420d51617f3233275c2b65ed706b35b53b115fe65f82c682af8134a6fR2184-R2185

https://github.com/solana-labs/solana/pull/13140/files#diff-1090394420d51617f3233275c2b65ed706b35b53b115fe65f82c682af8134a6fR2994-R2995

runtime/src/accounts_db.rs
carllin: hehe @ryoqun I actually spent quite some time tying to avoid this `Box` as well, and I think it's tradeoff for code size (alternative is creating a case for every one of the combinations). 

The problem with using `impl Fn(...)` is that the `write_version_producer` needs to be able to infer a type at compile time, so something like `write_version_producer` cannot hold two different types here (different in the `if` and `else` branch).

ryoqun: I see. Thanks for trying so hard and sharing the result. :)

ryoqun (approved):
lgtm; sorry for being late to review!

PR #13315, opened 2020-10-30
Add EpochVoteAccounts sysvar

mvines:
The EpochVoteAccounts sysvar contains the address and total stake for all vote accounts with active stake in the current epoch.

TODO:
* [ ] Add a test with hundreds of vote accounts


mvines:
Might not need this...

mvines:
Probably not going to carry this forward, I can do what I need off-chain instead

PR #13318, opened 2020-10-30
Flesh out development docs

jackcmay:
#### Problem

The development documentation is sorely lacking...

#### Summary of Changes

Re-organize and start fleshing out the docs

The general framework is in place but more content is wip

Old layout:

Screen Shot 2020-10-30 at 2 34 22 PM

New layout:

Screen Shot 2020-10-30 at 2 33 33 PM

Fixes #


jackcmay:
@mvines @garious 

Probably a bit hard to review and a rough draft but would love some initial feedback on the overall structure and the scope of topics covered (and anything else for that matter)

garious:
This is amazing! I'm okay with doing whatever you think is best on the `master` branch. Post-merge, I can do an end-to-end pass and look to clean up an inconsistencies.

garious:
Also, can you check in with @rmshea about how to preserve backward-compatibility with existing URLs?  Some of those URLs are more important than others, like say `transaction.md`, which has probably been shared with integration teams.

danpaul000:
Yeah, this is going to break every URL currently under docs.solana.com/apps . I think this should be easy to set the needed redirects when we deploy with Vercel (https://vercel.com/docs/configuration#project/redirects) 

I think the redirect array should go into the vercel config, which is pretty minimal right now, and auto-generated at build time in https://github.com/solana-labs/solana/blob/master/docs/publish-docs.sh

jackcmay:
Thanks @danpaul000 I'll do that as part of this PR

jackcmay:
Barring the nits, any objection to committing this once it passes CI and then iterating on it?

docs/publish-docs.sh
```diff
@@ -30,6 +30,25 @@ cat > "$CONFIG_FILE" <<EOF
   "name": "$PROJECT_NAME",
   "scope": "solana-labs"
 }
+{
+  "redirects": [
```
jackcmay, line 34: @danpaul000 Can you look over these, is there any way to test this before they are released?

docs/src/developing/deployed-programs/debugging.md
```diff
@@ -0,0 +1,118 @@
+---
+title: "Debugging"
+---
+
+Solana programs run on-chain, so debugging them in the wild can be challenging.
+To make debugging programs easier, developers can write unit tests that directly
+test their program's execution via the Solana runtime, or run a local cluster
+that will allow RPC clients to interact with their program.
+
+## Running unit tests
```
jackcmay, line 10: @mvines Leaving this as a TODO since the test stuff is influx

docs/src/developing/deployed-programs/debugging.md
```diff
@@ -0,0 +1,118 @@
...
+
+TODO
+
+## Running on a Local Cluster
```
jackcmay, line 14: Need help about how we should communicate the best way for folks to run on a local cluster vs unit tests and the next section about how to best do transaction simulation.

mvines:
@jackcmay - I feel like this should go into v1.4 as well, so it hits docs.solana.com this month

jackcmay:
@mvines Agreed, I want this to go into docs.solana.com asap, nobody looks at edge/beta.  Thinking I get this into master then we add the test updates and iterate on feedback and other additions

jackcmay:
@mvines @garious 1st swipe merged, looking for feedback :-)

@jstarry, Your changes are included

PR #13319, opened 2020-10-30
Reduce retransmit debug

sakridge:
#### Problem

Retransmit messages print count for all peers which can be a lot.

#### Summary of Changes

Just print top 5 peers.

Fixes #


behzadnouri (approved):
thank you

carllin (approved):

Issue #13171, opened 2020-10-26
explorer: Parse Serum DEX and swap instructions


oJshua:
#### Problem
Currently, Serum DEX instructions appear as Unknown (see: https://explorer.solana.com/address/ErF6ptYb3tKzSZjLKBoTqYJPUkMQtPUeAGo16kGLdGVt/tokens). It would be useful for Explorer to decode the instructions to provide more detail.

#### Proposed Solution
Decode the DEX instructions to provide insight into the instructions used in Serum transactions. 

spl-token-wallet implements transaction parsing for DEX instructions and others the UI commonly throws in (creating new accounts, initializing token accounts, etc.). https://github.com/project-serum/spl-token-wallet/blob/master/src/utils/transactions.js#L20

oJshua:
Related: https://github.com/solana-labs/solana/issues/11935

oJshua:
Swap instructions: https://github.com/solana-labs/oyster-swap/blob/master/src/models/tokenSwap.ts

PR #13320, opened 2020-10-30
explorer: Parse Serum DEX and swap instructions for TokenHistory

oJshua:
#### Problem
Currently, Serum DEX instructions appear as Unknown (see: https://explorer.solana.com/address/ErF6ptYb3tKzSZjLKBoTqYJPUkMQtPUeAGo16kGLdGVt/tokens). It would be useful for Explorer to decode the instructions to provide more detail.

#### Summary of Changes
- maps serum instructions on TokenHistory and updates SerumDetails card
- maps token swap instructions on TokenHistory and adds TokenSwapDetails card

Fixes https://github.com/solana-labs/solana/issues/13171

### DIFF
--- a/explorer/src/components/account/TokenHistoryCard.tsx
+++ b/explorer/src/components/account/TokenHistoryCard.tsx
@@ -3,7 +3,6 @@ import {
   PublicKey,
   ConfirmedSignatureInfo,
   ParsedInstruction,
-  PartiallyDecodedInstruction,
 } from "@solana/web3.js";
 import { CacheEntry, FetchStatus } from "providers/cache";
 import {
@@ -37,6 +36,10 @@ import {
   isSerumInstruction,
   parseSerumInstructionTitle,
 } from "utils/tx";
+import {
+  isTokenSwapInstruction,
+  parseTokenSwapInstructionTitle,
+} from "components/instruction/token-swap/types";
 
 export function TokenHistoryCard({ pubkey }: { pubkey: PublicKey }) {
   const address = pubkey.toBase58();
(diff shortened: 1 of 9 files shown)


explorer/src/components/instruction/token-swap/types.ts
```diff
@@ -0,0 +1,42 @@
...
+export function isTokenSwapInstruction(
+  instruction: TransactionInstruction
+): boolean {
+  return PROGRAM_IDS.includes(instruction.programId.toBase58());
+}
+
+export function parseTokenSwapInstructionTitle(
+  instruction: TransactionInstruction
+): string {
+  try {
```
jstarry, line 31: nit: unneeded outer try catch

explorer/src/components/instruction/serum/types.ts
```diff
@@ -0,0 +1,42 @@
...
+  5: "Settle Funds",
+  6: "Cancel Order By Client Id",
+  7: "Disable Market",
+  8: "Sweep Fees",
+};
+
+export function parseSerumInstructionTitle(
+  instruction: TransactionInstruction
+): string {
+  try {
```
jstarry, line 31: nit: same here, unnecessary try catch

explorer/src/components/account/TokenHistoryCard.tsx
```diff
@@ -301,13 +310,41 @@ const TokenTransactionRow = React.memo(
...
+          );
+        }
+
         if ("parsed" in ix) {
           if (ix.program === "spl-token") {
             return instructionTypeName(ix, tx);
           } else {
             return undefined;
           }
+        } else if (
```
jstarry, line 328: I was was originally thinking that we would flatten inner instructions here rather than show higher level instructions like swap / serum. I think this is pretty useful though.

jstarry (approved):
Looks good, it's a step forward to a vision I have for a much nicer token history table. I don't think this closes https://github.com/solana-labs/solana/issues/13171 though. Having the titles for the instructions is nice, but we need more context from the instruction data for this to be really useful

jstarry:
> it's a step forward to a vision I have for a much nicer token history table

Here's an issue for problems with the current token history table: it's a step forward to a vision I have for a much nicer token history table

PR #13322, opened 2020-10-31
Add validators to SoM list

danpaul000:
New validators on testnet should be added to stake-o-matic list.


mvines:
If you rebase this PR, it should get past `cargo audit`

PR #13324, opened 2020-10-31
Add solana-program-test crate

mvines:
Program test development is too hard.   This crate starts to improve the situation.  

This code has been living in https://github.com/solana-labs/solana-program-library/pull/626 for a while now but it really belongs in the monorepo.  The CPI support out of a BPF program compiled natively is not complete, lots of clean up needed before this can be fully supported (later). 

Once this ships in a release, it'll be imported into SPL and will unblock a number of SPL PRs.


mvines:
Hello world test case that uses `solana-program-test`: https://github.com/mvines/bpf-program-template/blob/5b98c98ab8e36852168eab177a69752983f3b37c/src/lib.rs#L20-L52

PR #13325, opened 2020-10-31
Small code cleanup and typo fixes

ryoqun:
#### Problem

There is some outdated code, typo, confusing variable names, spotted by me or audit report information suggestions.

#### Summary of Changes

Clarify them all. There should be absolutely no functional changes.

These code are extremely important. Let's squeeze out every part of noise and make it beautifully-coded.

Context #13233

### DIFF
--- a/programs/stake/src/stake_state.rs
+++ b/programs/stake/src/stake_state.rs
@@ -407,37 +407,45 @@ impl Stake {
     ///   for credits_observed were the points paid
     pub fn calculate_points_and_credits(
         &self,
-        vote_state: &VoteState,
+        new_vote_state: &VoteState,
         stake_history: Option<&StakeHistory>,
     ) -> (u128, u64) {
-        if self.credits_observed >= vote_state.credits() {
+        // if there is no newer credits since observed, return no point
+        if new_vote_state.credits() <= self.credits_observed {
             return (0, 0);
         }
 
-        let mut credits_observed = self.credits_observed;
-        let mut points = 0u128;
-        for (epoch, credits, prev_credits) in vote_state.epoch_credits() {
+        let mut points = 0;
+        let mut new_credits_observed = self.credits_observed;
+
+        for (epoch, final_epoch_credits, initial_epoch_credits) in
+            new_vote_state.epoch_credits().iter().copied()
+        {
+            let stake = u128::from(self.delegation.stake(epoch, stake_history));
+
             // figure out how much this stake has seen that
             //   for which the vote account has a record
-            let epoch_credits = if self.credits_observed < *prev_credits {
+            let earned_credits = if self.credits_observed < initial_epoch_credits {
                 // the staker observed the entire epoch
-                credits - prev_credits
-            } else if self.credits_observed < *credits {
+                final_epoch_credits - initial_epoch_credits
+            } else if self.credits_observed < final_epoch_credits {
                 // the staker registered sometime during the epoch, partial credit
-                credits - credits_observed
+                final_epoch_credits - new_credits_observed
             } else {
                 // the staker has already observed or been redeemed this epoch
                 //  or was activated after this epoch
                 0
             };
-
-            points += u128::from(self.delegation.stake(*epoch, stake_history))
-                * u128::from(epoch_credits);
+            let earned_credits = u128::from(earned_credits);
 
             // don't want to assume anything about order of the iterator...
-            credits_observed = credits_observed.max(*credits);
+            new_credits_observed = new_credits_observed.max(final_epoch_credits);
+
+            // finally calculate points for this epoch
+            points += stake * earned_credits;
         }
-        (points, credits_observed)
+
+        (points, new_credits_observed)
     }
 
     /// for a given stake and vote_state, calculate what distributions and what updates should be made
--- a/programs/vote/src/vote_state/mod.rs
+++ b/programs/vote/src/vote_state/mod.rs
@@ -30,8 +30,7 @@ pub const MAX_LOCKOUT_HISTORY: usize = 31;
 pub const INITIAL_LOCKOUT: usize = 2;
 
 // Maximum number of credits history to keep around
-//  smaller numbers makes
-pub const MAX_EPOCH_CREDITS_HISTORY: usize = 64;
+const MAX_EPOCH_CREDITS_HISTORY: usize = 64;
 
 #[frozen_abi(digest = "Ch2vVEwos2EjAVqSHCyJjnN2MNX1yrpapZTGhMSCjWUH")]
 #[derive(Serialize, Default, Deserialize, Debug, PartialEq, Eq, Clone, AbiExample)]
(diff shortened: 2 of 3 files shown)


mvines:
If you rebase this PR, it should get past `cargo audit`

programs/vote/src/vote_state/mod.rs
```diff
@@ -392,8 +391,7 @@ impl VoteState {
                 self.epoch_credits.last_mut().unwrap().0 = epoch;
             }
 
-            // if stakers do not claim before the epoch goes away they lose the
-            //  credits...
```
ryoqun, line 396: `claim-redeem` is past thing.

programs/vote/src/vote_state/mod.rs
```diff
@@ -30,8 +30,7 @@ pub const MAX_LOCKOUT_HISTORY: usize = 31;
 pub const INITIAL_LOCKOUT: usize = 2;
 
 // Maximum number of credits history to keep around
-//  smaller numbers makes
```
ryoqun, line 33: - incomplete comment
- `pub` isn't needed (at least for now)

programs/stake/src/stake_state.rs
```diff
@@ -407,37 +407,45 @@ impl Stake {
     ///   for credits_observed were the points paid
     pub fn calculate_points_and_credits(
         &self,
-        vote_state: &VoteState,
+        new_vote_state: &VoteState,
         stake_history: Option<&StakeHistory>,
     ) -> (u128, u64) {
-        if self.credits_observed >= vote_state.credits() {
+        // if there is no newer credits since observed, return no point
+        if new_vote_state.credits() <= self.credits_observed {
```
ryoqun, line 414: Added comment, and swapped the condition operands to match with the comment and read fluently (IMO).

programs/stake/src/stake_state.rs
```diff
@@ -407,37 +407,45 @@ impl Stake {
...
             return (0, 0);
         }
 
-        let mut credits_observed = self.credits_observed;
-        let mut points = 0u128;
-        for (epoch, credits, prev_credits) in vote_state.epoch_credits() {
+        let mut points = 0;
+        let mut new_credits_observed = self.credits_observed;
```
ryoqun, line 419: this is subtle, but let's match with the rhythm here: we're using (point, credits) tuple across this fn

ryoqun: also,  the intention of `point` (without `new_`) and `new_credits_observed` (with `new_`) is that point is isolated and independent for each inflation distribution. So, it's stateless. But `credits_observed` is continuously updated thing in stake accounts. So, it's state-full. So, only `credits_observed` can be said there is newer version and (current/older) version.

programs/stake/src/stake_state.rs
```diff
@@ -407,37 +407,45 @@ impl Stake {
...
+
+        for (epoch, final_epoch_credits, initial_epoch_credits) in
+            new_vote_state.epoch_credits().iter().copied()
```
ryoqun, line 422: `copied()` is to remove `*`s prefixed `credits` and `prev_credits`.

programs/stake/src/stake_state.rs
```diff
@@ -407,37 +407,45 @@ impl Stake {
...
-            points += u128::from(self.delegation.stake(*epoch, stake_history))
-                * u128::from(epoch_credits);
+            let earned_credits = u128::from(earned_credits);
 
             // don't want to assume anything about order of the iterator...
-            credits_observed = credits_observed.max(*credits);
+            new_credits_observed = new_credits_observed.max(final_epoch_credits);
+
+            // finally calculate points for this epoch
+            points += stake * earned_credits;
```
ryoqun, line 445: the main dish of this fn is put at the end, partly preparing for the next actual functional change pr.

programs/stake/src/stake_state.rs
```diff
@@ -407,37 +407,45 @@ impl Stake {
...
                 // the staker registered sometime during the epoch, partial credit
-                credits - credits_observed
+                final_epoch_credits - new_credits_observed
             } else {
                 // the staker has already observed or been redeemed this epoch
                 //  or was activated after this epoch
                 0
             };
-
-            points += u128::from(self.delegation.stake(*epoch, stake_history))
```
ryoqun, line 434: factor out `u128::from`s and stake calculation. let's make important definition (point calculation formula) shine by its own: https://github.com/solana-labs/solana/pull/13325/files#r515579738

programs/stake/src/stake_state.rs
ryoqun: yeah, I know this is against https://github.com/solana-labs/solana/pull/13325/files#r515579041 (`credits => points` here) and the general rule of putting looping-state-related code (maintaining max of new_credits_observed while looping) at the end to avoid intermixed state inside loop body.

programs/stake/src/stake_state.rs
```diff
@@ -407,37 +407,45 @@ impl Stake {
...
             return (0, 0);
         }
 
-        let mut credits_observed = self.credits_observed;
-        let mut points = 0u128;
-        for (epoch, credits, prev_credits) in vote_state.epoch_credits() {
+        let mut points = 0;
+        let mut new_credits_observed = self.credits_observed;
+
+        for (epoch, final_epoch_credits, initial_epoch_credits) in
```
ryoqun, line 421: the naming of `credits` and `prev_credits` here was a bit unhelpful in this context, imo.

what's important here is that they're together forming an interval because we're calculating overlaps between it and `self.credit_observed`.

So, I prefixed both of them with prefixes. (they used symmetrically in the logic but the one has a prefix and the other not; that's another point of confution, imo). `final_` and `initial_` is chosen over `last_/first_` and `end_/start_`:

- these epoch_credit interval's bounds are rather more rigid, not arbitrary, like they generally must not overlap with each other and properly ordered, etc. For example, this is not like lockout_intervals. So let's use somewhat formal-sounding wording here.
- it's less odd to see (final, initial) order than `(end, start)` because the `final` wording has more semantic significance than `initial`; usually `end` and `start` have equal balance. Otherwise, we have to change ABI to make `first, last` to happen...

ryoqun:
@mvines Could you review this? Now enginner talks about artistic aesthetic. xD These should be fairly safu. but I just want some sanity check since I've added another cleaning: https://github.com/solana-labs/solana/pull/13325/commits/76a7f8f7a6b3b83001755606d14c94bb1a40f2b1

mvines (approved):
Nice readability improvement!

programs/stake/src/stake_state.rs
```diff
@@ -407,37 +407,45 @@ impl Stake {
...
     pub fn calculate_points_and_credits(
         &self,
-        vote_state: &VoteState,
+        new_vote_state: &VoteState,
         stake_history: Option<&StakeHistory>,
     ) -> (u128, u64) {
-        if self.credits_observed >= vote_state.credits() {
+        // if there is no newer credits since observed, return no point
+        if new_vote_state.credits() <= self.credits_observed {
             return (0, 0);
```
ryoqun, line 415: ;) https://github.com/solana-labs/solana/pull/13836/files#r531357042

PR #13330, opened 2020-11-01
explorer: Serum DEX instruction full decoding and instruction cards

oJshua:
#### Problem
Serum and token swap instructions need to be fully parsed and available as instruction cards on the transactions page.

#### Summary of Changes
- fully parses Serum instructions 
- Serum instructions have their own unique instruction cards 

Fixes https://github.com/solana-labs/solana/issues/13171

### DIFF
--- a/explorer/src/components/account/TokenHistoryCard.tsx
+++ b/explorer/src/components/account/TokenHistoryCard.tsx
@@ -3,7 +3,6 @@ import {
   PublicKey,
   ConfirmedSignatureInfo,
   ParsedInstruction,
-  PartiallyDecodedInstruction,
 } from "@solana/web3.js";
 import { CacheEntry, FetchStatus } from "providers/cache";
 import {
--- a/explorer/src/components/account/TokenHistoryCard.tsx
+++ b/explorer/src/components/account/TokenHistoryCard.tsx
@@ -3,6 +3,7 @@ import {
   PublicKey,
   ConfirmedSignatureInfo,
   ParsedInstruction,
+  PartiallyDecodedInstruction,
 } from "@solana/web3.js";
 import { CacheEntry, FetchStatus } from "providers/cache";
 import {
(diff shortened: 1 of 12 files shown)


jstarry:
@oJshua do you mind posting some links / screenshots of the changes?

oJshua:
@jstarry sure!

matchesorders
Explorer-Solana (2)
consumeevents
Explorer-Solana (2)
Explorer-Solana (3)

I've been using this as a starting point to find Serum instructions: http://localhost:3000/address/CU3ACKhEJmtkTA2LtUpYjJuic2tA3Uw8hX6eZvwaNFyp/tokens

oJshua:
We'll have to get an updated mapping of the various markets at some point...

jstarry:
@oJshua looks like there are a few instructions that the serum-js sdk doesn't support yet: "Disable Market", "Sweep Fees", and "New Order v2". Is support for those in the works on the Serum side? Also, when the explorer comes across those it's going to report an error, can we avoid that since it's a known limitation?

It'd be nice if we moved towards generating ui cards (like how spl token instructions work) rather than having a view for each instruction. I think it would help cut down on maintenance overhead. Thoughts? Doesn't have to hold up this PR though

oJshua:
> @oJshua looks like there are a few instructions that the serum-js sdk doesn't support yet: "Disable Market", "Sweep Fees", and "New Order v2". Is support for those in the works on the Serum side?

I will find out the status on this. 

> Also, when the explorer comes across those it's going to report an error, can we avoid that since it's a known limitation?

Sure, I'll prevent that. 

> It'd be nice if we moved towards generating ui cards (like how spl token instructions work) rather than having a view for each instruction. I think it would help cut down on maintenance overhead. Thoughts? Doesn't have to hold up this PR though

Agreed! I'm looking at getting token swap cards mapped this week, so those would be a good candidate for generating these dynamically. After that's done, we can circle back and do the same for the Serum cards.

jstarry (approved):

PR #13338, opened 2020-11-01
uses thread-pool when handling push messages

behzadnouri:
#### Problem
From runtime profiles, the majority time of solana-listen thread:
https://github.com/solana-labs/solana/blob/55b0428ff/core/src/cluster_info.rs#L2720
is spent handling push messages. The code here:
https://github.com/solana-labs/solana/blob/55b0428ff/core/src/cluster_info.rs#L2272-L2364
may utilize the idle gossip thread-pool.

#### Summary of Changes
Updated the code to use rayon parallel iterators.

### DIFF
--- a/core/src/cluster_info.rs
+++ b/core/src/cluster_info.rs
@@ -68,6 +68,7 @@ use std::{
     cmp::min,
     collections::{hash_map::Entry, HashMap, HashSet},
     fmt,
+    iter::FromIterator,
     net::{IpAddr, Ipv4Addr, SocketAddr, TcpListener, UdpSocket},
     ops::{Deref, DerefMut},
     sync::atomic::{AtomicBool, AtomicU64, Ordering},
@@ -1606,16 +1607,19 @@ impl ClusterInfo {
         let (_, push_messages) = self
             .time_gossip_write_lock("new_push_requests", &self.stats.new_push_requests)
             .new_push_messages(self.drain_push_queue(), timestamp());
+        let push_messages: Vec<_> = {
+            let gossip =
+                self.time_gossip_read_lock("push_req_lookup", &self.stats.new_push_requests2);
+            push_messages
+                .into_iter()
+                .filter_map(|(pubkey, messages)| {
+                    let peer = gossip.crds.get_contact_info(&pubkey)?;
+                    Some((peer.gossip, messages))
+                })
+                .collect()
+        };
         let messages: Vec<_> = push_messages
             .into_iter()
-            .filter_map(|(peer, messages)| {
-                let peer_label = CrdsValueLabel::ContactInfo(peer);
-                self.time_gossip_read_lock("push_req_lookup", &self.stats.new_push_requests2)
-                    .crds
-                    .lookup(&peer_label)
-                    .and_then(CrdsValue::contact_info)
-                    .map(|p| (p.gossip, messages))
-            })
             .flat_map(|(peer, msgs)| {
                 Self::split_gossip_messages(msgs)
                     .into_iter()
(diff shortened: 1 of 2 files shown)


core/src/cluster_info.rs
```diff
@@ -2272,95 +2276,114 @@ impl ClusterInfo {
...
+        self.stats
+            .push_message_count
+            .add_relaxed(messages.len() as u64);
+        // Obtain shred versions of the origins.
+        let shred_versions: Vec<_> = {
+            let gossip = self.gossip.read().unwrap();
+            messages
+                .iter()
+                .map(|(from, _)| match gossip.crds.get_contact_info(from) {
+                    None => 0,
```
carllin, line 2296: I think it's more idiomatic to use combinators over a match here, i.e.

```
 messages.iter().map(|(from, _)|
                gossip
                    .crds
                    .get_contact_info(from)
                    .map_or(0, |info| info.shred_version)
```

core/src/crds.rs
```diff
@@ -160,6 +161,11 @@ impl Crds {
         self.table.get(label)
     }
 
+    pub fn get_contact_info(&self, pubkey: &Pubkey) -> Option<&ContactInfo> {
+        let label = CrdsValueLabel::ContactInfo(*pubkey);
+        self.table.get(&label)?.value.contact_info()
```
carllin, line 166: Can be simplified to `self.lookup(&label)?.contact_info()`

core/src/cluster_info.rs
```diff
@@ -2272,95 +2276,114 @@ impl ClusterInfo {
...
+                .collect()
+        };
+        // Filter out data if the origin has different shred version.
+        let self_shred_version = self.my_shred_version();
+        let num_crds_values: u64 = messages.iter().map(|(_, data)| data.len() as u64).sum();
+        let messages: Vec<_> = messages
+            .into_iter()
+            .zip(shred_versions)
+            .filter_map(|((from, mut crds_values), shred_version)| {
+                Self::filter_by_shred_version(
```
carllin, line 2308: Is this also worth tossing at the thread pool?

core/src/cluster_info.rs
```diff
@@ -2272,95 +2276,114 @@ impl ClusterInfo {
...
+        };
+        if prune_messages.is_empty() {
+            return;
         }
-        let mut packets = to_packets_with_destination(recycler.clone(), &rsp);
+        let mut packets = to_packets_with_destination(recycler.clone(), &prune_messages);
         self.stats
             .push_response_count
             .add_relaxed(packets.packets.len() as u64);
-        if !packets.is_empty() {
```
carllin, line 2349: just to clarify, this `is_empty` check no longer necessary because the

```if prune_messages.is_empty()```

above guarantees the returned `packets` it's not empty?

File: core/src/cluster_info.rs [hunk 5]
```diff
@@ -2272,95 +2276,114 @@ impl ClusterInfo {
...
+            .add_relaxed(num_crds_values - num_filtered_crds_values);
+        // Update crds values and obtain updated keys.
+        let updated_labels: Vec<_> = {
+            let mut gossip =
+                self.time_gossip_write_lock("process_push", &self.stats.process_push_message);
+            let now = timestamp();
+            messages
+                .into_iter()
+                .flat_map(|(from, crds_values)| {
+                    gossip.process_push_message(&from, crds_values, now)
```
carllin, line 2336: Out of curiosity, how expensive is this/rest of this function compared to the prunes that we were able to but in a thread pool below (i.e. what percentage of the `solana-listen` thread's compute? 

For future reference, if we ever want to break up these HashMaps, I've been using `DashMap` in the accounts changes which shards a HashMap for more concurrency: https://github.com/solana-labs/solana/pull/12126/files#diff-6431a3eec6e102bbcffb94387cda89efbe02002649745273fb575234c0018de4R17

core/src/cluster_info.rs
behzadnouri: yes

core/src/crds.rs
behzadnouri: I feel like the extra level of indirection is unnecessary here, and this is already a simple one-liner and maybe even more explicit.

core/src/cluster_info.rs
behzadnouri: From the runtime profiles, the majority time of this function is spent making the prune messages. There is a significant overhead with rayon parallel iterators, which is justified if the processing takes enough time. So, for the other pieces of this function, I would rather not use thread-pool for now.

core/src/cluster_info.rs
behzadnouri: Please see bellow comment.

core/src/cluster_info.rs
behzadnouri: I personally find the match syntax more readable here. It also avoids additional call-stacks (`map_or` plus the closure), and the rust formatter adds 2 additional lines.

carllin (approved):

PR #13345, opened 2020-11-02
Output more inflation calc details in ledger-tool

ryoqun:
#### Problem

There is no visibility into the actual inner working of calculation for inflation for scrutinization.

#### Summary of Changes

Peek the inflation calculation and report it via `ledger-tool capitalization --enable-inflation --warp-epoch +1`. Also support csv format.

For that purpose, introduce small generic logging facility all wired from Bank::new_from_parent to StakeState.

I chose this approach over the alternative (create another scaffold as a separate call graph to actually loop over and collect inflation internal metrics):
- Pros: we can detect _the actual implementation error in the code which are used in the production_, on top of general calculation error conflicting with the written documentation.
- Pros: implementing the alternative takes coding effort and another risk of introducing bugs.
- Pros: By use of `impl FnMut` we achieve this with zero runtime overhead. (to be really specific, only a single easily-predictable branch is incurred)
- Pros: the framework is very small and yet is very extensible and general enough to adapt future changes, IMO.
- Pros: this framework can be accommodated for data-parallelism, if the calculation logic should be changed to do so, anticipating for more large number of stake accounts
- Cons: this makes the inflation code cluttered to some extent for the need of passing `tracer` around. (But I think this is acceptable, considering other Pros)


runtime/src/bank.rs
```diff
@@ -1413,11 +1459,20 @@ impl Bank {
             let voters_account_pre_balance = vote_account.lamports;
 
             for (stake_pubkey, stake_account) in stake_group.iter_mut() {
+                // curry closure to add the contextual stake_pubkey
```
ryoqun, line 1462: fun fun functional thing. xD

runtime/src/bank.rs
```diff
@@ -752,6 +758,29 @@ impl Bank {
...
+            slot,
+            &mut None::<fn(&RewardCalcEvent)>,
+        )
+    }
+
+    pub fn new_from_parent_with_tracer(
+        parent: &Arc<Bank>,
+        collector_id: &Pubkey,
+        slot: Slot,
+        tracer: &mut Option<impl FnMut(&RewardCalcEvent)>,
```
ryoqun, line 773: I wanted to factor this ugly type signature out; but I couldn't mainly the limitation of `impl Trait`.

ryoqun:
@ericlwilliams I created a sample csvs for you: https://docs.google.com/spreadsheets/d/1N85HpuiYNNIlZ8quFHfWGhC8duE4NQMxVIP0-dD4CBg (made public because all of information is derived from our public testnet/mainnet-beta cluster).

ericlwilliams:
> @ericlwilliams I created a sample csvs for you: https://docs.google.com/spreadsheets/d/1N85HpuiYNNIlZ8quFHfWGhC8duE4NQMxVIP0-dD4CBg (made public because all of information is derived from our public testnet/mainnet-beta cluster).
Thanks @ryoqun !

programs/stake/src/stake_state.rs
```diff
@@ -2477,7 +2517,8 @@ mod tests {
                     points: 1
                 },
                 &vote_state,
-                None
+                None,
+                &mut None::<fn(&_)>,
```
ryoqun, line 2521: These turbo fish are needed because of limitation of rustc type inference.

ryoqun:
@mvines Could you review this? Or should I delegate the review to @t-nelson  or @CriesofCarrots because they're working with me as the things turned out. :)

programs/stake/src/stake_state.rs
```diff
@@ -31,6 +31,15 @@ pub enum StakeState {
     RewardsPool,
 }
 
+#[derive(Debug)]
+pub enum InflationPointCalcEvent {
```
mvines, line 35:
```suggestion
pub enum InflationPointCalculationEvent {
```

programs/stake/src/stake_state.rs
```diff
@@ -384,8 +394,9 @@ impl Stake {
         point_value: &PointValue,
         vote_state: &VoteState,
         stake_history: Option<&StakeHistory>,
+        tracer: &mut Option<impl FnMut(&InflationPointCalcEvent)>,
```
mvines, line 397: I'd prefer a more descriptive name for `tracer`: `inflation_point_calc_tracer`?

runtime/src/bank.rs
```diff
@@ -543,6 +543,12 @@ pub enum RewardType {
     Voting,
 }
 
+#[derive(Debug)]
+pub enum RewardCalcEvent<'a, 'b> {
```
mvines, line 547:
```suggestion
pub enum RewardCalculationEvent<'a, 'b> {
```

runtime/src/bank.rs
```diff
@@ -543,6 +543,12 @@ pub enum RewardType {
...
+    Fee, // still not wired
```
mvines, line 548: Remove until wired?

runtime/src/bank.rs
```diff
@@ -752,6 +758,29 @@ impl Bank {
...
     pub fn new_from_parent(parent: &Arc<Bank>, collector_id: &Pubkey, slot: Slot) -> Self {
+        Self::_new_from_parent(
+            parent,
+            collector_id,
+            slot,
+            &mut None::<fn(&RewardCalcEvent)>,
+        )
+    }
+
+    pub fn new_from_parent_with_tracer(
```
mvines, line 769: Same comment as above, I'd prefer a more descriptive name for `tracer`: `reward_calc_tracer`?

runtime/src/bank.rs
mvines, line 773: Why make `tracer` optional for `new_from_parent_with_tracer()?`

programs/stake/src/stake_state.rs
```diff
@@ -2386,7 +2423,8 @@ mod tests {
                     points: 1
                 },
                 &vote_state,
-                None
+                None,
+                &mut None::<fn(&_)>,
```
mvines, line 2427: could we have a default "null" tracer to avoid the Option.  Just thinking that a `&mut null_tracer` construct would look nicer than the borderline unreadable (a non Rust expert) `&mut None::<fn(&_)>` construct

mvines (commented):
> Or should I delegate the review to @t-nelson or @CriesofCarrots because they're working with me as the things turned out.  

Their feedback would be most welcome too

programs/stake/src/stake_state.rs
ryoqun: done 41c8d0bb11

programs/stake/src/stake_state.rs
ryoqun: done: 41c8d0bb11

runtime/src/bank.rs
ryoqun: done 41c8d0bb11

runtime/src/bank.rs
ryoqun: cf90494fdf

runtime/src/bank.rs
ryoqun: 41c8d0bb11

runtime/src/bank.rs
ryoqun: good point! ref: 3a4b6b39f2

programs/stake/src/stake_state.rs
ryoqun: ref: 707c276699 9712389566

ryoqun:
@mvines Phew, I think I addressed all comments. Thanks for careful reviews. :)

programs/stake/src/stake_state.rs
```diff
@@ -384,21 +398,28 @@ impl Stake {
         point_value: &PointValue,
         vote_state: &VoteState,
         stake_history: Option<&StakeHistory>,
+        inflation_point_calc_tracer: &mut Option<impl FnMut(&InflationPointCalculationEvent)>,
```
ryoqun, line 401: As a decent usecase of https://github.com/rust-lang/rust/issues/63063, I want to define a type alias for this....

mvines (approved):

PR #13352, opened 2020-11-02
Correctly handle the only root case

ryoqun:
#### Problem

this is the last known bug of tower.

this is just a brain dump so that I can focus on inflation.

#### Summary of Changes

TBD

Fixes #

### DIFF
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -1005,7 +1005,9 @@ impl Tower {
 
             checked_slot = Some(*slot_in_tower);
 
-            retain_flags_for_each_vote_in_reverse.push(anchored_slot.is_none());
+            let is_still_unanchored = anchored_slot.is_none();
+            // not anchored slots must be retained
+            retain_flags_for_each_vote_in_reverse.push(is_still_unanchored);
         }
 
         // Check for errors if not anchored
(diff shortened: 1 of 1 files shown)


core/src/consensus.rs
```diff
@@ -3136,6 +3144,21 @@ pub mod test {
...
+        tower.lockouts.root_slot = Some(42);
+        tower.lockouts.votes.push_back(Lockout::new(42));
+        let vote = Vote::new(vec![42], Hash::default());
+        tower.last_vote = vote;
+
+        let mut slot_history = SlotHistory::default();
+        slot_history.add(42);
+
+        let tower = tower.adjust_lockouts_after_replay(42, &slot_history);
+        assert_eq!(tower.unwrap().voted_slots(), [42]);
```
ryoqun, line 3159: also assert root

core/src/consensus.rs
```diff
@@ -3136,6 +3144,21 @@ pub mod test {
         assert_eq!(tower.unwrap().voted_slots(), [43, 44]);
     }
 
+    #[test]
+    fn test_adjust_lockouts_after_replay_vote_on_only_root() {
```
ryoqun, line 3148: also add future tower version of this test?

core/src/consensus.rs
```diff
@@ -1027,7 +1029,13 @@ impl Tower {
             retain_flags_for_each_vote_in_reverse.into_iter().rev();
 
         let original_votes_len = self.lockouts.votes.len();
-        self.initialize_lockouts(move |_| retain_flags_for_each_vote.next().unwrap());
+        let keep_the_only_root_vote_for_no_double_vote =
```
ryoqun, line 1032: comment here why this must be accepted.

ryoqun: because we start to vote on root....

also, it's very very rare to see this on the wild. I've just come up with this possibility while working on #12671.

carllin: We can also augment the check here https://github.com/solana-labs/solana/blob/master/core/src/consensus.rs#L454 to check that the vote slot > root so the `is_locked_out` check will fail

PR #13357, opened 2020-11-03
Fix stake split rent-exempt adjustment

CriesofCarrots:
#### Problem
Stake split doesn't properly handle the stake deduction from a Delegated original account when the new account has a balance less than the rent-exempt reserve. The original stake needs to be reduced by the entire balance transferred, except in the case of splitting the entire source stake, when it should simply be zeroed.

Stake split also errors on an attempt to split all of a source account to a new account that holds a balance. This is inconsistent with the behavior for a partial split, where an existing balance is acceptable and used to cover the rent-exempt reserve (see`test_split_with_rent()`).

#### Summary of Changes
- Fix math

Before/after examples added in comments.

This is WIP as I need to coordinate gating with @t-nelson 
cc: @ryoqun @mvines

### DIFF
--- a/programs/stake/src/stake_state.rs
+++ b/programs/stake/src/stake_state.rs
@@ -771,11 +771,12 @@ impl<'a> StakeAccount for KeyedAccount<'a> {
                     );
 
                     // verify enough lamports for rent in new split account
-                    if split.lamports()? + lamports < split_rent_exempt_reserve
+                    if lamports < split_rent_exempt_reserve.saturating_sub(split.lamports()?)
                     // verify full withdrawal can cover rent in new split account
                         || (lamports < split_rent_exempt_reserve && lamports == self.lamports()?)
                     // verify enough lamports left in previous stake and not full withdrawal
-                        || (lamports + meta.rent_exempt_reserve > self.lamports()? && lamports != self.lamports()?)
+                            || (lamports > self.lamports()? - meta.rent_exempt_reserve
+                            && lamports != self.lamports()?)
                     {
                         return Err(InstructionError::InsufficientFunds);
                     }
@@ -822,7 +823,8 @@ impl<'a> StakeAccount for KeyedAccount<'a> {
                     // enough lamports for rent in new stake
                     if lamports < split_rent_exempt_reserve
                     // verify enough lamports left in previous stake
-                        || (lamports + meta.rent_exempt_reserve > self.lamports()? && lamports != self.lamports()?)
+                        || (lamports + meta.rent_exempt_reserve > self.lamports()?
+                            && lamports != self.lamports()?)
                     {
                         return Err(InstructionError::InsufficientFunds);
                     }
--- a/programs/stake/src/stake_state.rs
+++ b/programs/stake/src/stake_state.rs
@@ -515,14 +515,18 @@ impl Stake {
         Ok(())
     }
 
-    fn split(&mut self, lamports: u64) -> Result<Self, StakeError> {
-        if lamports > self.delegation.stake {
+    fn split(
+        &mut self,
+        remaining_stake_delta: u64,
+        split_stake_amount: u64,
+    ) -> Result<Self, StakeError> {
+        if remaining_stake_delta > self.delegation.stake {
             return Err(StakeError::InsufficientStake);
         }
-        self.delegation.stake -= lamports;
+        self.delegation.stake -= remaining_stake_delta;
         let new = Self {
             delegation: Delegation {
-                stake: lamports,
+                stake: split_stake_amount,
                 ..self.delegation
             },
             ..*self
@@ -769,13 +773,31 @@ impl<'a> StakeAccount for KeyedAccount<'a> {
                         return Err(InstructionError::InsufficientFunds);
                     }
                     // split the stake, subtract rent_exempt_balance unless
-                    //  the destination account already has those lamports
-                    //  in place.
-                    // this could represent a small loss of staked lamports
-                    //  if the split account starts out with a zero balance
-                    let split_stake = stake.split(
-                        lamports - meta.rent_exempt_reserve.saturating_sub(split.lamports()?),
-                    )?;
+                    // the destination account already has those lamports
+                    // in place.
+                    // this means that the new stake account will have a stake equivalent to
+                    // lamports minus rent_exempt_reserve if it starts out with a zero balance
+                    let (remaining_stake_delta, split_stake_amount) = if lamports
+                        == self.lamports()?
+                    {
+                        // If split amount equals the full source stake, the new split stake must
+                        // equal the same amount, regardless of any current lamport balance in the
+                        // split account. Since split accounts retain the state of their source
+                        // account, this prevents any magic activation of stake by prefunding the
+                        // split account.
+                        (
+                            lamports - meta.rent_exempt_reserve,
+                            lamports - meta.rent_exempt_reserve,
+                        )
+                    } else {
+                        // Otherwise, the new split stake should reflect the entire split
+                        // requested, less any lamports needed to cover the rent-exempt reserve
+                        (
+                            lamports,
+                            lamports - meta.rent_exempt_reserve.saturating_sub(split.lamports()?),
+                        )
+                    };
+                    let split_stake = stake.split(remaining_stake_delta, split_stake_amount)?;
 
                     self.set_state(&StakeState::Stake(meta, stake))?;
                     split.set_state(&StakeState::Stake(meta, split_stake))?;
@@ -3049,7 +3071,7 @@ mod tests {
             split_stake_keyed_account.account.borrow_mut().lamports = 1_000;
             assert_eq!(
                 stake_keyed_account.split(
-                    stake_lamports - rent_exempt_reserve,
+                    stake_lamports - rent_exempt_reserve, // leave rent_exempt_reserve in original account
                     &split_stake_keyed_account,
                     &signers
                 ),
--- a/programs/stake/src/stake_state.rs
+++ b/programs/stake/src/stake_state.rs
@@ -3176,6 +3176,100 @@ mod tests {
     }
 
     #[test]
+    fn test_split_to_account_with_rent_exempt_reserve() {
+        let stake_pubkey = solana_sdk::pubkey::new_rand();
+        let stake_lamports = 42;
+        let rent_exempt_reserve = 10;
+
+        let split_stake_pubkey = solana_sdk::pubkey::new_rand();
+        let signers = vec![stake_pubkey].into_iter().collect();
+
+        let meta = Meta {
+            authorized: Authorized::auto(&stake_pubkey),
+            rent_exempt_reserve,
+            ..Meta::default()
+        };
+
+        let state = StakeState::Stake(
+            meta,
+            Stake::just_stake(stake_lamports - rent_exempt_reserve),
+        );
+        // Test various account prefunding, including empty, less than rent_exempt_reserve, exactly
+        // rent_exempt_reserve, and more than rent_exempt_reserve. The empty case is not covered in
+        // test_split, since that test uses a Meta with rent_exempt_reserve = 0
+        let split_lamport_balances = vec![0, 1, rent_exempt_reserve, rent_exempt_reserve + 1];
+        for initial_balance in split_lamport_balances {
+            let split_stake_account = Account::new_ref_data_with_space(
+                initial_balance,
+                &StakeState::Uninitialized,
+                std::mem::size_of::<StakeState>(),
+                &id(),
+            )
+            .expect("stake_account");
+
+            let split_stake_keyed_account =
+                KeyedAccount::new(&split_stake_pubkey, true, &split_stake_account);
+
+            let stake_account = Account::new_ref_data_with_space(
+                stake_lamports,
+                &state,
+                std::mem::size_of::<StakeState>(),
+                &id(),
+            )
+            .expect("stake_account");
+            let stake_keyed_account = KeyedAccount::new(&stake_pubkey, true, &stake_account);
+
+            // split more than available fails
+            assert_eq!(
+                stake_keyed_account.split(stake_lamports + 1, &split_stake_keyed_account, &signers),
+                Err(InstructionError::InsufficientFunds)
+            );
+
+            // should work
+            assert_eq!(
+                stake_keyed_account.split(stake_lamports / 2, &split_stake_keyed_account, &signers),
+                Ok(())
+            );
+            // no lamport leakage
+            assert_eq!(
+                stake_keyed_account.account.borrow().lamports
+                    + split_stake_keyed_account.account.borrow().lamports,
+                stake_lamports + initial_balance
+            );
+
+            if let StakeState::Stake(meta, stake) = state {
+                assert_eq!(
+                    Ok(StakeState::Stake(
+                        meta,
+                        Stake {
+                            delegation: Delegation {
+                                stake: stake_lamports / 2
+                                    - (rent_exempt_reserve.saturating_sub(initial_balance)),
+                                ..stake.delegation
+                            },
+                            ..stake
+                        }
+                    )),
+                    split_stake_keyed_account.state()
+                );
+                assert_eq!(
+                    Ok(StakeState::Stake(
+                        meta,
+                        Stake {
+                            delegation: Delegation {
+                                stake: stake_lamports / 2 - rent_exempt_reserve,
+                                ..stake.delegation
+                            },
+                            ..stake
+                        }
+                    )),
+                    stake_keyed_account.state()
+                );
+            }
+        }
+    }
+
+    #[test]
     fn test_split_100_percent_of_source() {
         let stake_pubkey = solana_sdk::pubkey::new_rand();
         let stake_lamports = 42;
@@ -3269,6 +3363,94 @@ mod tests {
 
             // reset
             stake_keyed_account.account.borrow_mut().lamports = stake_lamports;
+        }
+    }
+
+    #[test]
+    fn test_split_100_percent_of_source_to_account_with_lamports() {
+        let stake_pubkey = solana_sdk::pubkey::new_rand();
+        let stake_lamports = 42;
+        let rent_exempt_reserve = 10;
+
+        let split_stake_pubkey = solana_sdk::pubkey::new_rand();
+        let signers = vec![stake_pubkey].into_iter().collect();
+
+        let meta = Meta {
+            authorized: Authorized::auto(&stake_pubkey),
+            rent_exempt_reserve,
+            ..Meta::default()
+        };
+
+        let state = StakeState::Stake(
+            meta,
+            Stake::just_stake(stake_lamports - rent_exempt_reserve),
+        );
+        // Test various account prefunding, including empty, less than rent_exempt_reserve, exactly
+        // rent_exempt_reserve, and more than rent_exempt_reserve. Technically, the empty case is
+        // covered in test_split_100_percent_of_source, but included here as well for readability
+        let split_lamport_balances = vec![0, 1, rent_exempt_reserve, rent_exempt_reserve + 1];
+        for initial_balance in split_lamport_balances {
+            let split_stake_account = Account::new_ref_data_with_space(
+                initial_balance,
+                &StakeState::Uninitialized,
+                std::mem::size_of::<StakeState>(),
+                &id(),
+            )
+            .expect("stake_account");
+
+            let split_stake_keyed_account =
+                KeyedAccount::new(&split_stake_pubkey, true, &split_stake_account);
+
+            let stake_account = Account::new_ref_data_with_space(
+                stake_lamports,
+                &state,
+                std::mem::size_of::<StakeState>(),
+                &id(),
+            )
+            .expect("stake_account");
+            let stake_keyed_account = KeyedAccount::new(&stake_pubkey, true, &stake_account);
+
+            // split 100% over to dest
+            assert_eq!(
+                stake_keyed_account.split(stake_lamports, &split_stake_keyed_account, &signers),
+                Ok(())
+            );
+
+            // no lamport leakage
+            assert_eq!(
+                stake_keyed_account.account.borrow().lamports
+                    + split_stake_keyed_account.account.borrow().lamports,
+                stake_lamports + initial_balance
+            );
+
+            if let StakeState::Stake(meta, stake) = state {
+                assert_eq!(
+                    Ok(StakeState::Stake(
+                        meta,
+                        Stake {
+                            delegation: Delegation {
+                                stake: stake_lamports - rent_exempt_reserve,
+                                ..stake.delegation
+                            },
+                            ..stake
+                        }
+                    )),
+                    split_stake_keyed_account.state()
+                );
+                assert_eq!(
+                    Ok(StakeState::Stake(
+                        meta,
+                        Stake {
+                            delegation: Delegation {
+                                stake: 0,
+                                ..stake.delegation
+                            },
+                            ..stake
+                        }
+                    )),
+                    stake_keyed_account.state()
+                );
+            }
         }
     }
 
--- a/programs/stake/src/stake_state.rs
+++ b/programs/stake/src/stake_state.rs
@@ -770,8 +770,10 @@ impl<'a> StakeAccount for KeyedAccount<'a> {
                         split.data_len()? as u64,
                     );
 
-                    // verify enough lamports for rent in new stake with the split
+                    // verify enough lamports for rent in new split account
                     if split.lamports()? + lamports < split_rent_exempt_reserve
+                    // verify full withdrawal can cover rent in new split account
+                        || (lamports < split_rent_exempt_reserve && lamports == self.lamports()?)
                     // verify enough lamports left in previous stake and not full withdrawal
                         || (lamports + meta.rent_exempt_reserve > self.lamports()? && lamports != self.lamports()?)
                     {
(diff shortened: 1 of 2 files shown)


CriesofCarrots:
Splitting partial amount into empty account (default cli behavior)

Before:

```
$ solana stake-account stake.json
Balance: 0.1 SOL
Rent Exempt Reserve: 0.00228288 SOL
Delegated Stake: 0.09771712 SOL
Active Stake: 0.09771712 SOL
Delegated Vote Account Address: 2FPim6EJBTW3uwr36fAhK8BEGB22BjJuxkPbtMYAeVUv

$ solana split-stake stake.json split.json 0.05

Signature: 4u1xhY2NqvsoGQHqh65QvJpT3AVyBoyv2K7Bx6h3JbuJ8NLwMFAeHKCZdhSi8wNm5huYMvmphqHn1nYzs39SLYRw

$ solana stake-account stake.json
Balance: 0.05 SOL
Rent Exempt Reserve: 0.00228288 SOL
Delegated Stake: 0.05 SOL <--- Notice stake is not less the rent-exempt-reserve
Active Stake: 0.05 SOL
Delegated Vote Account Address: 2FPim6EJBTW3uwr36fAhK8BEGB22BjJuxkPbtMYAeVUv

$ solana stake-account split.json`
Balance: 0.05 SOL
Rent Exempt Reserve: 0.00228288 SOL
Delegated Stake: 0.04771712 SOL
Active Stake: 0.04771712 SOL
Delegated Vote Account Address: 2FPim6EJBTW3uwr36fAhK8BEGB22BjJuxkPbtMYAeVUv
```

After:

```
$ solana stake-account stake.json
Balance: 0.1 SOL
Rent Exempt Reserve: 0.00228288 SOL
Delegated Stake: 0.09771712 SOL
Active Stake: 0.09771712 SOL
Delegated Vote Account Address: GGohSWd9vwLDqfu32nXuzYM7B6W8N7baRSo5ph9CJsG9

$ solana split-stake stake.json split.json 0.05

Signature: 4oEubJWX4WcjDQaN2KtN2Li4TsZD3WRDeWfGGAg26FzWATKEbJt8Nc84zhfMDx78pY1oxgnGhQMsWDhB75MJTTSK

$ solana stake-account stake.json
Balance: 0.05 SOL
Rent Exempt Reserve: 0.00228288 SOL
Delegated Stake: 0.04771712 SOL
Active Stake: 0.04771712 SOL
Delegated Vote Account Address: GGohSWd9vwLDqfu32nXuzYM7B6W8N7baRSo5ph9CJsG9

$ solana stake-account split.json
Balance: 0.05 SOL
Rent Exempt Reserve: 0.00228288 SOL
Delegated Stake: 0.04771712 SOL
Active Stake: 0.04771712 SOL
Delegated Vote Account Address: GGohSWd9vwLDqfu32nXuzYM7B6W8N7baRSo5ph9CJsG9
```

CriesofCarrots:
Splitting complete stake account into account prefunded with rent-exempt reserve of 0.00228288 SOL (I had to edit cli to use this create-account instruction:
```
system_instruction::create_account(
            authorized_pubkey, // Sending 0, so any signer will suffice
            split_stake_pubkey,
            2282880,
            std::mem::size_of::<StakeState>() as u64,
            &id(),
        )
```

Before:

```
$ solana stake-account stake.json
Balance: 0.1 SOL
Rent Exempt Reserve: 0.00228288 SOL
Delegated Stake: 0.09771712 SOL
Active Stake: 0.09771712 SOL
Delegated Vote Account Address: JCcavabWfWZc9ykWb68Vjvpy6oH6nYWGrkwFdEnk4Z95

$ solana split-stake stake.json split.json 0.1
Error: RPC response error -32002: Transaction simulation failed: Error processing Instruction 1: custom program error: 0x4
```

After:

```
$ solana stake-account /Users/tyeraeulberg/.config/solana/stake.json`
Balance: 0.1 SOL
Rent Exempt Reserve: 0.00228288 SOL
Delegated Stake: 0.09771712 SOL
Active Stake: 0.09771712 SOL
Delegated Vote Account Address: FmM9tLoFrqAAS6K9MrMxDWT2o1RzpE5cv388iYJSMAm2
Stake Authority: Bbqg1M4YVVfbhEzwA9SpC9FhsaG83YMTYoR4a8oTDLX
Withdraw Authority: Bbqg1M4YVVfbhEzwA9SpC9FhsaG83YMTYoR4a8oTDLX

$ solana split-stake stake.json split.json 0.1

Signature: rVtvcTHJX6VYdYBPKyiX9rssVAB9SUSK6jNU28HSmwTMf9HJXVsrKRtrSCfVjbeam8V34CNFR8BmkDwkNPRHFVq

$ solana stake-account stake.json
Error: AccountNotFound: pubkey=CYRJWqiSjLitBAcRxPvWpgX3s5TvmN2SuRY3eEYypFvT

$ solana stake-account split.json
Balance: 0.10228288 SOL
Rent Exempt Reserve: 0.00228288 SOL
Delegated Stake: 0.09771712 SOL
Active Stake: 0.09771712 SOL
Delegated Vote Account Address: FmM9tLoFrqAAS6K9MrMxDWT2o1RzpE5cv388iYJSMAm2
```

CriesofCarrots:
@rwalker-com , would you be able to take a look at this and see if I'm missing anything? 🙏

rwalker-com (approved):
lgtm.  all I had to do was read tests:-)

programs/stake/src/stake_state.rs
```diff
@@ -766,13 +770,31 @@ impl<'a> StakeAccount for KeyedAccount<'a> {
...
-                        lamports - meta.rent_exempt_reserve.saturating_sub(split.lamports()?),
-                    )?;
+                    // the destination account already has those lamports
+                    // in place.
+                    // this means that the new stake account will have a stake equivalent to
+                    // lamports minus rent_exempt_reserve if it starts out with a zero balance
+                    let (remaining_stake_delta, split_stake_amount) = if lamports
+                        == self.lamports()?
+                    {
+                        // If split amount equals the full sounce stake, the new split stake must
```
t-nelson, line 780:
```suggestion
                        // If split amount equals the full source stake, the new split stake must
```

programs/stake/src/stake_state.rs
```diff
@@ -766,13 +770,31 @@ impl<'a> StakeAccount for KeyedAccount<'a> {
...
+                        // equal the same amount, regardless of any current lamport balance in the
+                        // split account. Since split accounts retain the state of their sournce
```
t-nelson, line 782:
```suggestion
                        // split account. Since split accounts retain the state of their source
```

t-nelson (approved):
LGTM!  Just a couple comment typoes

programs/stake/src/stake_state.rs
CriesofCarrots: 🤦‍♀️ How many wrong ways can I spell that word?

t-nelson: I believe company policy is to blame the Macbook Pro keyboard!

programs/stake/src/stake_state.rs
```diff
@@ -766,13 +770,31 @@ impl<'a> StakeAccount for KeyedAccount<'a> {
...
+                        // requested, less any lamports needed to cover the rent-exempt reserve
+                        (
+                            lamports,
+                            lamports - meta.rent_exempt_reserve.saturating_sub(split.lamports()?),
+                        )
+                    };
+                    let split_stake = stake.split(remaining_stake_delta, split_stake_amount)?;
 
                     self.set_state(&StakeState::Stake(meta, stake))?;
                     split.set_state(&StakeState::Stake(meta, split_stake))?;
```
ryoqun, line 800: @CriesofCarrots oops, I found another issue. Namely, the split account's rent-reserve can be set incorrectly. I'm still reading the code, but we could never safely `set_state` identical  `meta` to two stake accounts. We have to individually maintain rent-exempt reserves. another head-ache...

Very unfortunately, the are various sizes of stake accounts on the wild:

```
$ ./target/release/solana stake-account C75NM6yJMa4uAEfdUYnSmMHfVAUiPVd67rqTKaZdd5Z
Balance: 500000 SOL
Rent Exempt Reserve: 0.00228288 SOL
Delegated Stake: 499999.99771712 SOL
Active Stake: 499999.99771712 SOL
Delegated Vote Account Address: CCxSNvJogH6LWyoiEbG7JfcWybw2FqqCExs5GuemChGr
Stake Authority: so1b2w9fpMdqgzHR2UvW4iqkkc8nig6xTeBjU5HMbjG
Withdraw Authority: FdGYQdiRky8NZzN9wZtczTBcWLYYRXrJ3LMDhqDPn5rM


$ ./target/release/solana stake-account 13CJic1hdT9woLL7ctrKi9rBwBFoqjcF2BNnDbZceDnS
Balance: 25 SOL
Rent Exempt Reserve: 0.02878656 SOL
Delegated Stake: 24.97121344 SOL
Active Stake: 24.97121344 SOL
Delegated Vote Account Address: ateamaZDqNWDztxnVKZhRsp4ac53KvT1rVKyU5LnL6o
Stake Authority: 55c8bMTU311mv8KkAnTMtdGGKWAueAGhNUYdVfvjuvNy
Withdraw Authority: 55c8bMTU311mv8KkAnTMtdGGKWAueAGhNUYdVfvjuvNy
```

ryoqun:
```
$ ./target/release/solana account C75NM6yJMa4uAEfdUYnSmMHfVAUiPVd67rqTKaZdd5Z | head

Public Key: C75NM6yJMa4uAEfdUYnSmMHfVAUiPVd67rqTKaZdd5Z
Balance: 500000 SOL
Owner: Stake11111111111111111111111111111111111111
Executable: false
Rent Epoch: 107
Length: 200 (0xc8) bytes
0000:   02 00 00 00  80 d5 22 00  00 00 00 00  0d 03 10 73   ......"........s
0010:   44 03 0b f0  4d b2 8a 86  2c 2d 19 93  1f fb c5 e1   D...M...,-......
0020:   50 36 55 ed  9f 03 88 e8  d0 d7 98 bb  d9 4e 19 53   P6U..........N.S

$ ./target/release/solana account 13CJic1hdT9woLL7ctrKi9rBwBFoqjcF2BNnDbZceDnS | head

Public Key: 13CJic1hdT9woLL7ctrKi9rBwBFoqjcF2BNnDbZceDnS
Balance: 25 SOL
Owner: Stake11111111111111111111111111111111111111
Executable: false
Rent Epoch: 107
Length: 4008 (0xfa8) bytes
0000:   02 00 00 00  80 3f b7 01  00 00 00 00  3c 9c cb 27   .....?......<..'
0010:   f9 18 85 f0  46 c2 c1 29  91 56 ab 80  e1 37 8f 30   ....F..).V...7.0
0020:   76 d8 fd ff  ab d7 d0 4b  38 d7 fd ce  3c 9c cb 27   v......K8...<..'
```

ryoqun: That means my upcoming stake account normalization must also adjust rent-exempt reserves.

ryoqun: ref: https://discord.com/channels/428295358100013066/439194979856809985/771378185890693131

programs/stake/src/stake_state.rs
```diff
@@ -761,11 +761,22 @@ impl<'a> StakeAccount for KeyedAccount<'a> {
             match self.state()? {
                 StakeState::Stake(meta, mut stake) => {
                     meta.authorized.check(signers, StakeAuthorize::Staker)?;
+                    let original_lamports_per_byte_year = calculate_lamports_per_byte_year(
```
t-nelson, line 764: Do you anticipate this value being used otherwise?  If not, it might simplify things a bit to combine the two utility functions into one `fn calculate_split_rent_exempt_reserve(source: &KeyedAccount, split: &KeyedAccount) -> u64`

CriesofCarrots: No, I hope it won't be used otherwise. Fair point, will squash.

programs/stake/src/stake_state.rs
```diff
@@ -761,11 +761,22 @@ impl<'a> StakeAccount for KeyedAccount<'a> {
...
+                    // verify that the same rent_exempt_reserve is sufficient to make the new
+                    // stake account rent exempt needed because the new uninitialized may be any
+                    // size larger than std::mem::size_of::<StakeState>()
+                        || calculate_rent_exempt_reserve(
+                            split.data_len()? as u64,
+                            original_lamports_per_byte_year,
+                        ) > meta.rent_exempt_reserve
                     {
                         return Err(InstructionError::InsufficientFunds);
                     }
```
t-nelson, line 782: We need to use the split's `rent_exempt_reserve` below, no? (Sorry can't comment outside diff context)

CriesofCarrots: No, the rent_exempt_reserve needs to remain the same between the source and split accounts. We just don't want to allow that to put the split account in a non-exempt state

programs/stake/src/stake_state.rs
CriesofCarrots: Updated

programs/stake/src/stake_state.rs
CriesofCarrots: Mm, maybe we can actually, as long as the stake being split never increases as a result.

programs/stake/src/stake_state.rs
```diff
@@ -969,6 +1007,17 @@ pub fn calculate_points(
     }
 }
 
+// utility function, used by Split
```
ryoqun, line 1010: maybe add a comment saying the context: `doing here because not to break compatibility; newer stake program ABI can just use newly-required Rent sysvar` or the like.

programs/stake/src/stake_state.rs
```diff
@@ -969,6 +1007,17 @@ pub fn calculate_points(
...
+fn calculate_split_rent_exempt_reserve(
+    source_meta: &Meta,
+    source_data_len: u64,
+    split_data_len: u64,
+) -> u64 {
+    let lamports_per_byte_year =
+        source_meta.rent_exempt_reserve / (source_data_len + ACCOUNT_STORAGE_OVERHEAD);
+    lamports_per_byte_year * (split_data_len + ACCOUNT_STORAGE_OVERHEAD)
```
ryoqun, line 1018: Also, let's cross-ref this out-of-wild rent calculation was unavoidable at `Rent` in rent.rs. :)

programs/stake/src/stake_state.rs
CriesofCarrots: Added

programs/stake/src/stake_state.rs
```diff
@@ -969,6 +1011,19 @@ pub fn calculate_points(
     }
 }
 
+// utility function, used by Split
+//This emulates current Rent math in order to preserve backward compatibility. In the future, and
+//to support variable rent, the Split instruction should pass in the Rent sysvar instead.
+fn calculate_split_rent_exempt_reserve(
+    source_meta: &Meta,
```
ryoqun, line 1018: nits: `source_rent_exempt_reserve`?

programs/stake/src/stake_state.rs
CriesofCarrots: Where would you like to see that comment in rent.rs?

programs/stake/src/stake_state.rs
CriesofCarrots: Updated. No tests failed, so that's good. I still need to add a case for larger source to smaller split, splitting partial funds.

programs/stake/src/stake_state.rs
```diff
@@ -757,33 +761,71 @@ impl<'a> StakeAccount for KeyedAccount<'a> {
...
+                        (
+                            lamports - meta.rent_exempt_reserve,
+                            lamports - meta.rent_exempt_reserve,
+                        )
+                    } else {
+                        // Otherwise, the new split stake should reflect the entire split
+                        // requested, less any lamports needed to cover the rent-exempt reserve
+                        (
+                            lamports,
+                            lamports - meta.rent_exempt_reserve.saturating_sub(split.lamports()?),
```
ryoqun, line 803: I think this must be `lamports - split_rent_exempt_reserve.saturating_sub(split.lamports()?)`, right?

programs/stake/src/stake_state.rs
```diff
@@ -757,33 +761,71 @@ impl<'a> StakeAccount for KeyedAccount<'a> {
...
+                        == self.lamports()?
+                    {
+                        // If split amount equals the full source stake, the new split stake must
+                        // equal the same amount, regardless of any current lamport balance in the
+                        // split account. Since split accounts retain the state of their source
+                        // account, this prevents any magic activation of stake by prefunding the
+                        // split account.
+                        (
+                            lamports - meta.rent_exempt_reserve,
+                            lamports - meta.rent_exempt_reserve,
```
ryoqun, line 796: Also, this must be `lamports - split_rent_exempt_reserve.rent_exempt_reserve`, right?

programs/stake/src/stake_state.rs
```diff
@@ -757,33 +761,71 @@ impl<'a> StakeAccount for KeyedAccount<'a> {
...
                 StakeState::Stake(meta, mut stake) => {
                     meta.authorized.check(signers, StakeAuthorize::Staker)?;
+                    let split_rent_exempt_reserve = calculate_split_rent_exempt_reserve(
+                        meta.rent_exempt_reserve,
+                        self.data_len()? as u64,
+                        split.data_len()? as u64,
+                    );
 
                     // verify enough lamports for rent in new stake with the split
                     if split.lamports()? + lamports < meta.rent_exempt_reserve
```
ryoqun, line 771: instead of `split_rent_exempt_reserve > meta.rent_exempt_reserve` below, the condition can be tweaked to `split.lamports()? + lamports < split_rent_exempt_reserve`?

programs/stake/src/stake_state.rs
CriesofCarrots: No, definitely not. The amount of stake delegated should not be able to increase by splitting to a smaller-sized account.

programs/stake/src/stake_state.rs
CriesofCarrots: Ah, yes, correct.

programs/stake/src/stake_state.rs
```diff
@@ -757,33 +761,71 @@ impl<'a> StakeAccount for KeyedAccount<'a> {
...
-                     // verify enough lamports left in previous stake and not full withdrawal
+                    // verify enough lamports left in previous stake and not full withdrawal
                         || (lamports + meta.rent_exempt_reserve > self.lamports()? && lamports != self.lamports()?)
+                    // verify that the same rent_exempt_reserve is sufficient to make the new
+                    // stake account rent exempt needed because the new uninitialized may be any
+                    // size larger than std::mem::size_of::<StakeState>()
+                        || split_rent_exempt_reserve > meta.rent_exempt_reserve
```
ryoqun, line 777: this new error condition will make it fail to split smaller-sized stake account into bigger-sized stake account even if there is no inconsistency between `rent_exempt_reserve` and `data_len()`. That will be very hard to debug for users. If could, I'd like to avoid this by adjusting split_stake and split_rent_exempt_reserve...

programs/stake/src/stake_state.rs
ryoqun: Hhm. That's because of bad side-effect to the stake distribution for the current epoch? Maybe I'm lacking some precondition.

programs/stake/src/stake_state.rs
CriesofCarrots: Yes, my intent was to prevent all splits from smaller-sized stake accounts into bigger-sized ones, since we are otherwise addressing the causes of small stake accounts with incorrect larger `rent_exempt_reserve`s.

> If could, I'd like to avoid this by adjusting split_stake and split_rent_exempt_reserve...

I'm sorry, I don't understand what you mean. Unless you are referring to your comment above?

programs/stake/src/stake_state.rs
CriesofCarrots: Hmm, it looks like the max stake delta would be ~69 SOL. Is that small enough we don't care?

CriesofCarrots:
@t-nelson @ryoqun , once more if you would 🙏 
I reworked this such that:
- The Split account `Meta::rent_exempt_reserve` is calculated and set to the correct value for the new account, regardless of whether it is higher or lower than the original
- Splits that would leave the new account not rent exempt fail
- Split account lamports = stake + rent_exempt_reserve, *except* if the account is prefunded with more than the rent_exempt_reserve, to prevent the magic activation of stake via split (which could be up to u64::MAX)
- Otherwise, the max possible change in activated stake by splitting between 2 accounts of dissimilar sizes is ~69 SOL

@ryoqun , hoping for a little more guidance about where you'd like documentation in rent.rs 🙇‍♀️ . Otherwise, I think I've addressed all of your comments.

programs/stake/src/stake_state.rs
```diff
@@ -757,31 +761,61 @@ impl<'a> StakeAccount for KeyedAccount<'a> {
...
+                    // this means that the new stake account will have a stake equivalent to
+                    // lamports minus rent_exempt_reserve if it starts out with a zero balance
+                    let (remaining_stake_delta, split_stake_amount) = if lamports
+                        == self.lamports()?
+                    {
+                        // If split amount equals the full source stake, the new split stake must
+                        // equal the same amount, regardless of any current lamport balance in the
+                        // split account. Since split accounts retain the state of their source
+                        // account, this prevents any magic activation of stake by prefunding the
+                        // split account.
```
ryoqun, line 789: :+1:

programs/stake/src/stake_state.rs
```diff
@@ -757,31 +761,61 @@ impl<'a> StakeAccount for KeyedAccount<'a> {
...
+                        (
+                            lamports - meta.rent_exempt_reserve,
+                            lamports - split_rent_exempt_reserve,
```
ryoqun, line 792: Argurably hard to notice, but I think this causes overflow, allowing user to stake around `u64::max_size()` if all of following conditions are met by specifically creating crafted account state and transaction by malicious users:

1. `lamports` is identical as `source.lamports()`.
2. `split` account is bigger than source; so `split_rent_exempt_reserve` > `meta.rent_exempt_reserve`
3. `split` account is pre-funded so that `split.lamports()? + lamports >= split_rent_exempt_reserve`.

ryoqun: I think the best thing we could here is to reject such a case? Also, I'd like to add a test for this. (I know you already adding very extensive set of tests. Sorry for more trouble).

programs/stake/src/stake_state.rs
```diff
@@ -757,31 +761,61 @@ impl<'a> StakeAccount for KeyedAccount<'a> {
...
+                        self.data_len()? as u64,
+                        split.data_len()? as u64,
+                    );
 
                     // verify enough lamports for rent in new stake with the split
-                    if split.lamports()? + lamports < meta.rent_exempt_reserve
-                     // verify enough lamports left in previous stake and not full withdrawal
+                    if split.lamports()? + lamports < split_rent_exempt_reserve
+                    // verify enough lamports left in previous stake and not full withdrawal
                         || (lamports + meta.rent_exempt_reserve > self.lamports()? && lamports != self.lamports()?)
```
ryoqun, line 773: nits: I think these conditions are bit hard to read? How about swapping this inequality around `lamports` like this?

```rust
if lamports < split_rent_excempt_reserve + lamports
    || (lamports > self.lamports()? - meta.rent_excempt_reserver && ...)
```

programs/stake/src/stake_state.rs
```diff
@@ -757,31 +761,61 @@ impl<'a> StakeAccount for KeyedAccount<'a> {
...
+                        (
+                            lamports - meta.rent_exempt_reserve,
+                            lamports - split_rent_exempt_reserve,
+                        )
+                    } else {
+                        // Otherwise, the new split stake should reflect the entire split
+                        // requested, less any lamports needed to cover the rent-exempt reserve
+                        (
+                            lamports,
+                            lamports - split_rent_exempt_reserve.saturating_sub(split.lamports()?),
```
ryoqun, line 799: I think similar attack is possible here like this: https://github.com/solana-labs/solana/pull/13357/files#r517344371

ryoqun:
Still pondering on this. But I think this getting very solid with many new tests. Thanks for great work!

ryoqun:
> @ryoqun , hoping for a little more guidance about where you'd like documentation in rent.rs bowing_woman . Otherwise, I think I've addressed all of your comments.

maybe, this fn:

https://github.com/solana-labs/solana/blob/b0d1ae1d8bac917198b8130be0464a1166f1b1fe/sdk/program/src/rent.rs#L49

along like:

```
// note that stripped down version of this calcuration is reluctantly copied into calculate_split_rent_exempt_reserve in the stake program
```

programs/stake/src/stake_state.rs
CriesofCarrots: I think I see. I'll write a test to confirm. Yes, I think rejecting makes sense.

programs/stake/src/stake_state.rs
CriesofCarrots: This case is covered by the initial lamports check (more obvious with your rewriting: `lamports < split_rent_excempt_reserve.saturating_sub(split.lamports()?)`)

programs/stake/src/stake_state.rs
CriesofCarrots: ✅  Reframed all those checks in terms of `lamports` (and also let cargo fmt work on those lines)

programs/stake/src/stake_state.rs
CriesofCarrots: I included my test that demonstrates this: `test_split_to_larger_account_edge_case()`

programs/stake/src/stake_state.rs
CriesofCarrots: Please see `test_split_100_percent_of_source_to_larger_account_edge_case()`

programs/stake/src/stake_state.rs
```diff
@@ -3689,4 +4309,30 @@ mod tests {
         // Test another staking action
         assert_eq!(stake_keyed_account.deactivate(&clock, &new_signers), Ok(()));
     }
+
+    #[test]
+    fn test_calculate_lamports_per_byte_year() {
+        let rent_exempt_reserve = 2282880;
```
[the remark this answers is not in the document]
t-nelson: Probably in the above tests as well

t-nelson (commented):
Looking good!  I just have a couple test hardening recommendations

CriesofCarrots:
This will need rebasing on #13394 (make sure it goes in stake_state.rs, as opposed to the legacy mod)

programs/stake/src/stake_state.rs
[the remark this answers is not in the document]
CriesofCarrots: Probably a good idea, thanks!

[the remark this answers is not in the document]
CriesofCarrots: Done

programs/stake/src/stake_state.rs
```diff
@@ -3689,4 +4309,30 @@ mod tests {
...
+            rent_exempt_reserve
+        );
+
+        let larger_data = 4008;
+        let expected_rent_exempt_reserve = 28786560;
+        assert_eq!(
+            calculate_split_rent_exempt_reserve(rent_exempt_reserve, data_len, larger_data),
+            expected_rent_exempt_reserve
+        );
+
```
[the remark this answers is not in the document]
CriesofCarrots: Added

t-nelson (dismissed):
LGTM!

PR #13364, opened 2020-11-03
drops older gossip packets when load shedding

behzadnouri:
#### Problem
Gossip drops incoming packets when overloaded:
https://github.com/solana-labs/solana/blob/f6a73098a/core/src/cluster_info.rs#L2462-L2475
However newer packets are dropped in favor of the older ones.
This is probably not ideal as newer packets are more likely to contain
more recent data, so dropping them will keep the validator state
lagging.

#### Summary of Changes
The commit buffers the packet into a `VecDeque`.
When there are too many packets, it drops from the beginning of the `VecDeque`.

### DIFF
--- a/core/src/cluster_info.rs
+++ b/core/src/cluster_info.rs
@@ -2460,15 +2460,14 @@ impl ClusterInfo {
         let packets: Vec<_> = requests_receiver.recv_timeout(RECV_TIMEOUT)?.packets.into();
         let mut packets = VecDeque::from(packets);
         while let Ok(packet) = requests_receiver.try_recv() {
-            let num_packets = packets.len() + packet.packets.len();
-            if num_packets > MAX_GOSSIP_TRAFFIC {
-                let excess_count = usize::min(num_packets - MAX_GOSSIP_TRAFFIC, packets.len());
+            packets.extend(packet.packets.into_iter());
+            let excess_count = packets.len().saturating_sub(MAX_GOSSIP_TRAFFIC);
+            if excess_count > 0 {
+                packets.drain(0..excess_count);
                 self.stats
                     .gossip_packets_dropped_count
                     .add_relaxed(excess_count as u64);
-                packets.drain(0..excess_count);
             }
-            packets.extend(packet.packets.into_iter());
         }
         let (stakes, epoch_time_ms) = Self::get_stakes_and_epoch_time(bank_forks);
         // Using root_bank instead of working_bank here so that an enbaled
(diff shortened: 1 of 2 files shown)


core/src/cluster_info.rs
```diff
@@ -2456,24 +2456,20 @@ impl ClusterInfo {
...
-        while let Ok(more_reqs) = requests_receiver.try_recv() {
-            if num_requests >= MAX_GOSSIP_TRAFFIC {
-                continue;
+        const RECV_TIMEOUT: Duration = Duration::from_secs(1);
+        let packets: Vec<_> = requests_receiver.recv_timeout(RECV_TIMEOUT)?.packets.into();
+        let mut packets = VecDeque::from(packets);
+        while let Ok(packet) = requests_receiver.try_recv() {
+            let num_packets = packets.len() + packet.packets.len();
+            if num_packets > MAX_GOSSIP_TRAFFIC {
+                let excess_count = usize::min(MAX_GOSSIP_TRAFFIC - num_packets, packets.len());
```
carllin, line 2465: if `num_packets > MAX_GOSSIP_TRAFFIC` then `MAX_GOSSIP_TRAFFIC - num_packets` will overflow, should `excess_count` just be `num_packets - MAX_GOSSIP_TRAFFIC` here?

behzadnouri: oops, you are right. fixed

core/src/cluster_info.rs
```diff
@@ -2462,7 +2462,7 @@ impl ClusterInfo {
         while let Ok(packet) = requests_receiver.try_recv() {
             let num_packets = packets.len() + packet.packets.len();
             if num_packets > MAX_GOSSIP_TRAFFIC {
-                let excess_count = usize::min(MAX_GOSSIP_TRAFFIC - num_packets, packets.len());
+                let excess_count = usize::min(num_packets - MAX_GOSSIP_TRAFFIC, packets.len());
```
carllin, line 2465: Is there a reason to take the `min` here?

For instance if on the last iteration of this loop, num_packets = 400, MAX_GOSSIP_TRAFFIC = 100 and packets.len() = 1,  then we'll drop at most 1 packet and end up with `4x` the MAX_GOSSIP_TRAFFIC

behzadnouri: I updated the code to enforce the max-traffic more strictly. but to answer your question,  we cannot call `packets.drain` with range exceeding `packets.len`. So we needed a `min` there. I changed the code to call `drain` after `extend`, and so no need for `min` anymore. However the previous code was more similar to [current mainnet code](https://github.com/solana-labs/solana/blob/8838a55a1/core/src/cluster_info.rs#L2490-L2497), which allows the length overrun the max-gossip-traffic.
Either way I do not think it does matter much here, because the overshoot will be likely very small relatively.

carllin: I see, thanks!

carllin (approved):

PR #13365, opened 2020-11-03
Rbpf v0.2.0

Lichtso:
#### Problem
The original PR https://github.com/solana-labs/solana/pull/12951 had to be reverted because of a bug in the JIT code size estimation.

This PR is only for the general improvements and refactoring of RBPF.
The bug fixes, code to enable JIT and Interpreter testing, as well as JIT caching will come in separate PRs.

#### Summary of Changes
JIT is hard coded to be disabled, but can be overwritten manually in the code.

Fixes #

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5123,9 +5123,9 @@ dependencies = [
 
 [[package]]
 name = "solana_rbpf"
-version = "0.1.32"
+version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


jackcmay:
@Lichtso Were you able to verify that Michael's test case that found the jit but passes with this PR?

Lichtso:
@jackcmay The JIT is completely disabled, thus the bug should not occur. The fix is already on a branch in the RBPF repo and will be merged with a later PR here.

jackcmay (approved):

PR #13366, opened 2020-11-03
measures processing time of each kind of gossip packets

behzadnouri:
#### Problem
Gossip packets processing breakdown

#### Summary of Changes
Added metrics for measuring processing time of each kind of gossip packets.

### DIFF
--- a/core/src/cluster_info.rs
+++ b/core/src/cluster_info.rs
@@ -197,6 +197,16 @@ impl Counter {
     fn clear(&self) -> u64 {
         self.0.swap(0, Ordering::Relaxed)
     }
+    fn measure_task<F, R>(&self, task: F) -> R
+    where
+        F: FnOnce() -> R,
+    {
+        let now = Instant::now();
+        let out = task();
+        let micros = now.elapsed().as_micros();
+        self.0.fetch_add(micros as u64, Ordering::Relaxed);
+        out
+    }
 }
 
 #[derive(Default)]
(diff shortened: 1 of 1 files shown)


core/src/cluster_info.rs
```diff
@@ -196,6 +196,16 @@ impl Counter {
     fn clear(&self) -> u64 {
         self.0.swap(0, Ordering::Relaxed)
     }
+    fn measure_task<F, R>(&self, task: F) -> R
+    where
+        F: FnOnce() -> R,
+    {
+        let now = Instant::now();
```
carllin, line 203: nit: We've been shifting to using `Measure`  and `add_measure` like here: https://github.com/solana-labs/solana/blob/master/core/src/cluster_info.rs#L1162-L1177 over `Instant`

carllin (approved):

sakridge (approved):
2nd preference for Measure, but lgtm

core/src/cluster_info.rs
behzadnouri: yes, that is what I am using elsewhere, but here, some of these functions have multiple points of `return` so adding a `self.stats.abc.add_measure(&mut timer)` at each of those `return` points is a bit verbose and clumsy.
I have updated this pull-request with a new `ScopedTimer` to automatically track the execution time of the function with just one line addition.

carllin (approved):

core/src/cluster_info.rs
carllin: cool thanks!

Issue #13379, opened 2020-11-04
Inner instructions are duplicated for each transaction instruction


jstarry:
#### Problem
Instruction recorders are being reused for each message instruction which causes duplicate records of inner instructions for transactions with multiple instructions.

##### Explanation of bug:

`InstructionRecorder` has interior mutability
https://github.com/solana-labs/solana/blob/6d9ca0ae157a0f109987fd7802fcfcf19e532ed8/runtime/src/instruction_recorder.rs#L9-L12

`vec!` will clone the expression `n` times to populate the vector
https://github.com/solana-labs/solana/blob/6d9ca0ae157a0f109987fd7802fcfcf19e532ed8/runtime/src/bank.rs#L2439-L2446

#### Proposed Solution
- Fix the bug
- Ignore inner instructions that were stored in bigtable/blockstore prior to the fix

PR #13380, opened 2020-11-04
Fix duplicate records of inner instructions

jstarry:
#### Problem
Inner instructions are duplicately recorded for each instruction in a transaction due to an interior mutability bug.

#### Summary of Changes
- Fix bug and add test
- Deprecate bad inner instruction data in blockstore/bigtable
- Add new inner instruction field to metadata in blockstore/bigtable

Fixes https://github.com/solana-labs/solana/issues/13379

### DIFF
--- a/local-cluster/tests/local_cluster.rs
+++ b/local-cluster/tests/local_cluster.rs
@@ -1315,10 +1315,11 @@ fn test_fake_shreds_broadcast_leader() {
 fn test_faulty_node(faulty_node_type: BroadcastStageType) {
     solana_logger::setup();
     let num_nodes = 2;
-    let validator_config = ValidatorConfig::default();
     let mut error_validator_config = ValidatorConfig::default();
     error_validator_config.broadcast_stage_type = faulty_node_type;
-    let mut validator_configs = vec![validator_config; num_nodes - 1];
+    let mut validator_configs = Vec::with_capacity(num_nodes - 1);
+    validator_configs.resize_with(num_nodes - 1, || ValidatorConfig::default());
+
     // Push a faulty_bootstrap = vec![error_validator_config];
     validator_configs.insert(0, error_validator_config);
     let node_stakes = vec![300, 100];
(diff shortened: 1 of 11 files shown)


jstarry:
Tested locally against blockstore and emulated bigtable.. old inner instructions are no longer returned and new ones are not duplicated. Protobuf worked as expected!

local-cluster/tests/local_cluster.rs
```diff
@@ -1315,10 +1315,11 @@ fn test_fake_shreds_broadcast_leader() {
 fn test_faulty_node(faulty_node_type: BroadcastStageType) {
     solana_logger::setup();
     let num_nodes = 2;
-    let validator_config = ValidatorConfig::default();
     let mut error_validator_config = ValidatorConfig::default();
     error_validator_config.broadcast_stage_type = faulty_node_type;
-    let mut validator_configs = vec![validator_config; num_nodes - 1];
+    let mut validator_configs = Vec::with_capacity(num_nodes - 1);
+    validator_configs.resize_with(num_nodes - 1, ValidatorConfig::default);
```
jstarry, line 1321: This came up while I searched for other instances of the same bug. `ValidatorConfig` has interior mutability as well so I patched this up despite it not causing any issues for this test

mvines:
Can we avoid the `bad_inner_instructions` field?   Reuse the existing file and either purge the inner instructions out of bigtable for older slots, or just don't care about the invalid data.

jstarry:
Sure, we can a use a version field in the struct to ignore old bad data

mvines:
> Sure, we can a use a version field in the struct to ignore old bad data

Or not even?  We can just document in an issue that there's bad inner instruction metadata data for slots < N on mainnet-beta.  If/when somebody actually cares, we can write/run a custom purge on BigTable directly to remove the data.

jstarry:
Ok, that's fine. Whenever we update the warehouse nodes we should be sure to note the slot when this code is activated then.

mvines:
> Ok, that's fine. Whenever we update the warehouse nodes we should be sure to note the slot when this code is activated then.

Happy to load up bw1 with a "stable" channel build as soon as this PR hits v1.3

PR #13394, opened 2020-11-04
Feature-gate stake program

CriesofCarrots:
#### Problem
Upcoming stake-program fixes need to be feature-gated and released on testnet and mainnet-beta.

#### Summary of Changes
- Add legacy stake-program processor and handling
- Use feature_builtin to replace legacy stake program

Needs rebase on #13403

### DIFF
--- a/runtime/src/builtins.rs
+++ b/runtime/src/builtins.rs
@@ -39,8 +39,13 @@ pub enum ActivationType {
 }
 
 /// Builtin programs activated dynamically by feature
+///
 /// Note: If the feature_builtin is intended to replace another builtin program, it must have a new
 /// name.
+/// This is to enable the runtime to determine categorically whether the builtin update has
+/// occurred, and preserve idempotency in Bank::add_native_program across genesis, snapshot, and
+/// normal child Bank creation.
+/// https://github.com/solana-labs/solana/blob/84b139cc94b5be7c9e0c18c2ad91743231b85a0d/runtime/src/bank.rs#L1723
 fn feature_builtins() -> Vec<(Builtin, Pubkey, ActivationType)> {
     vec![
         (
--- a/runtime/src/builtins.rs
+++ b/runtime/src/builtins.rs
@@ -17,10 +17,12 @@ fn genesis_builtins() -> Vec<Builtin> {
             solana_vote_program::id(),
             solana_vote_program::vote_instruction::process_instruction,
         ),
+        // Remove legacy_stake_processor and move stake_instruction::process_instruction back to
+        // genesis_builtins around the v1.6 timeframe
         Builtin::new(
             "stake_program",
             solana_stake_program::id(),
-            solana_stake_program::stake_instruction::process_instruction,
+            solana_stake_program::legacy_stake_processor::process_instruction,
         ),
         Builtin::new(
             "config_program",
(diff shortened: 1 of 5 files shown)


runtime/src/bank.rs
```diff
@@ -4043,6 +4047,19 @@ impl Bank {
         }
     }
 
+    fn apply_stake_program_fix(&mut self) {
+        if let Some(mut account) = self.get_account(&solana_stake_program::id()) {
+            self.capitalization.fetch_sub(account.lamports, Relaxed);
+            account.lamports = 0;
```
mvines, line 4053: Why do we need to drain the account here?

runtime/src/bank.rs
```diff
@@ -3966,6 +3966,10 @@ impl Bank {
             self.apply_spl_token_v2_multisig_fix();
         }
 
+        if new_feature_activations.contains(&feature_set::stake_program_fix::id()) {
```
mvines, line 3969: Adding builtins is a trick business.  I think you should roll this into `ensure_feature_builtins` (current code breaks when a snapshot is loaded after the feature activation for example).

The new stake program could just be added to this list:
https://github.com/solana-labs/solana/blob/4b65e32f22ef0da600606a41bba0624e1f1cb2dd/runtime/src/builtins.rs#L34
and we might need to adjust some logic to allow for feature_builtins() to replace existing builtins, which I think is probably ok in general

runtime/src/bank.rs
CriesofCarrots: Otherwise add_builtin fails due to native-program overwrite protection: https://github.com/solana-labs/solana/blob/97284adabbad84da00ae04949a9db947848135ae/runtime/src/bank.rs#L9273

runtime/src/bank.rs
CriesofCarrots: > (current code breaks when a snapshot is loaded after the feature activation for example).

Do you mean the code in this pr breaks?
If so, do you mind pointing out how? It looks to me like it should work okay, since Bank::finish_init calls `self.apply_feature_activations(true);`

But allowing feature_builtins() to replace existing builtins does seem useful, so I'll move that direction.

mvines: A feature is only in the `new_feature_activations` once, when the Bank crosses the activation slot.    Call that slot N.   If that bank then creates a snapshot at slot N+1 and the validator restarts from it, `new_feature_activations` at snapshot restore will not contain `stake_program_fix` and `Bank::apply_stake_program_fix()` won't execute

CriesofCarrots: Ah, got it, hence `init_or_warp1. Thanks for pointing out `ensure_feature_builtins`!

t-nelson:
Any thoughts one replacing all of the math in the stake program with checked variants while we're replacing the program?  It might save us some trouble in the future

CriesofCarrots:
@mvines how does this look now?

mvines:
> @mvines how does this look now?

I think that looks good now.  My preference would be to break out the new bank mechanism from the rest of this PR.  That I think can be reviewed and landed first (@ryoqun probably should look at it as well).  Then the application of it to the stake account is a second PR

runtime/src/builtins.rs
```diff
@@ -30,16 +32,34 @@ fn genesis_builtins() -> Vec<Builtin> {
...
+                "secp256k1_program",
+                solana_sdk::secp256k1_program::id(),
+                solana_secp256k1_program::process_instruction,
+            ),
+            feature_set::secp256k1_program_enabled::id(),
+            ActivationType::NewProgram,
+        ),
+        (
+            Builtin::new(
+                "stake_program_v2",
```
ryoqun, line 55: Partly due to https://github.com/solana-labs/solana/pull/13403#discussion_r518078470, basically you can't use same name for the updated new instruction_processor.

Snapshot restoration must discern whether it's new or old only from AccountDB's contents....

CriesofCarrots:
@mvines this is ready for proper review. Thank you!

mvines (commented):
How should I review legacy_stake.stake.rs?

mvines:
> How should I review legacy_stake_stake.rs?

Specifically without a diff commit from the original stake_stake.rs, it's really hard to see what, if anything has changed

mvines:
(oops)

CriesofCarrots:
> > How should I review legacy_stake_stake.rs?
> 
> Specifically without a diff commit from the original stake_stake.rs, it's really hard to see what, if anything has changed

Fair! Let me make you a diff commit

runtime/src/builtins.rs
```diff
@@ -37,16 +39,29 @@ pub enum ActivationType {
 }
 
 /// Builtin programs activated dynamically by feature
+/// Note: If the feature_builtin is intended to replace another builtin program, it must have a new
+/// name.
```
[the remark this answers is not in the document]
CriesofCarrots: Done https://github.com/solana-labs/solana/pull/13394/commits/095098aad62c14f61e4a92f71a622ef6be4fed46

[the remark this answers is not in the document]
mvines: permalink or just reference `Bank::add_native_program()` 🙏

CriesofCarrots:
> > > How should I review legacy_stake_stake.rs?
> > 
> > 
> > Specifically without a diff commit from the original stake_stake.rs, it's really hard to see what, if anything has changed
> 
> Fair! Let me make you a diff commit

@mvines legacy_stake_state diff commit: https://github.com/solana-labs/solana/pull/13394/commits/41e3c5c340542983eca9296fec5f105c16f2419c

mvines (approved):

runtime/src/builtins.rs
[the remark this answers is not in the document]
CriesofCarrots: TIL github `y` shortcut

PR #13398, opened 2020-11-04
SPL Associated Token Account plumbing

mvines:
* Add explorer label
* Add program to `./fetch-spl.sh`


mvines:
TODO: Update program id

PR #13403, opened 2020-11-04
Allow feature builtins to overwrite existing builtins

CriesofCarrots:
#### Problem
Right now, there is no easy way to redeploy a native program (to fix issues, eg). One appealing approach would be to deploy the updated program using feature_builtins, but this requires a way to overwrite an active builtin.

#### Summary of Changes
- Add a method to enable overwriting existing builtins
- Use this method in `ensure_feature_builtins` to allow a feature_builtin to replace an existing builtin

Needed for #13394

### DIFF
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -10,7 +10,7 @@ use crate::{
     accounts_db::{ErrorCounters, SnapshotStorages},
     accounts_index::Ancestors,
     blockhash_queue::BlockhashQueue,
-    builtins,
+    builtins::{self, ActivationType},
     epoch_stakes::{EpochStakes, NodeVoteAccounts},
     instruction_recorder::InstructionRecorder,
     log_collector::LogCollector,
@@ -215,7 +215,7 @@ pub struct Builtins {
     pub genesis_builtins: Vec<Builtin>,
 
     /// Builtin programs activated dynamically by feature
-    pub feature_builtins: Vec<(Builtin, Pubkey)>,
+    pub feature_builtins: Vec<(Builtin, Pubkey, ActivationType)>,
 }
 
 const MAX_CACHED_EXECUTORS: usize = 100; // 10 MB assuming programs are around 100k
@@ -676,7 +676,7 @@ pub struct Bank {
     bpf_compute_budget: Option<BpfComputeBudget>,
 
     /// Builtin programs activated dynamically by feature
-    feature_builtins: Arc<Vec<(Builtin, Pubkey)>>,
+    feature_builtins: Arc<Vec<(Builtin, Pubkey, ActivationType)>>,
 
     /// Last time when the cluster info vote listener has synced with this bank
     pub last_vote_sync: AtomicU64,
@@ -1719,7 +1719,7 @@ impl Bank {
         }
     }
 
-    pub fn add_native_program(&self, name: &str, program_id: &Pubkey, force_overwrite: bool) {
+    pub fn add_native_program(&self, name: &str, program_id: &Pubkey, must_replace: bool) {
         let mut already_genuine_program_exists = false;
         if let Some(mut account) = self.get_account(&program_id) {
             already_genuine_program_exists = native_loader::check_id(&account.owner);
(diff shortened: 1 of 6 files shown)


mvines (approved):

runtime/src/bank.rs
```diff
@@ -1737,19 +1737,27 @@ impl Bank {
...
+                "There is no program account to replace"
+            );
+        } else if !already_genuine_program_exists {
             assert!(
                 !self.is_frozen(),
                 "Can't change frozen bank by adding not-existing new native program ({}, {}). \
                 Maybe, inconsistent program activation is detected on snapshot restore?",
                 name,
                 program_id
             );
```
ryoqun, line 1752: Could you move this assertion to reside immediately before `store_account`? This assertion is tightly associated to `store_account`.  We can safely `assert!(!self.is_frozen())` even if `must_replace` is `true`. In general, we must not `store_account` once frozen in every conceivable situation. But that kind of protection isn't easy to add inside `store_account` itself, unfortunately. So, we're manually doing this because these activations deal with primitive arrangements with potentinal misbehavior of wracking frozen (snapshot) banks incorrectly.

So, the change hunk could be changed to look like this:

```rust
if must_replace {
  assert!(already_genuine_program_exists, ...)
} else if already_genuine_program_exists {
  return;
}

assert!(!self.is_frozen(), ...);
self.store_account(...);

```

ryoqun:
Thanks for working on this! Good progress. In addition to `NewProgram`, I think this added `NewVersion` is decent usecase as well.

runtime/src/builtins.rs
```diff
@@ -30,15 +30,22 @@ fn genesis_builtins() -> Vec<Builtin> {
     ]
 }
 
+#[derive(Debug, Clone)]
+pub enum ActivationType {
```
ryoqun, line 34: oh, https://buildkite.com/solana-labs/solana/builds/34410#3d37c5dd-fddb-417c-8cb1-b6a514b0384b/139-1725

Sorry another my picky work is complaining to you again. ;)
Please add AbiExample/AbiEnumVisitor, this is needed for all types accessible from Bank.
These derives are like `Default` or `Debug` for the ABI check magic.

runtime/src/bank.rs
```diff
@@ -1737,19 +1737,27 @@ impl Bank {
...
-        if !already_genuine_program_exists {
+        if must_replace {
             assert!(
-                !self.is_frozen(),
-                "Can't change frozen bank by adding not-existing new native program ({}, {}). \
-                Maybe, inconsistent program activation is detected on snapshot restore?",
-                name,
-                program_id
+                already_genuine_program_exists,
+                "There is no program account to replace"
```
ryoqun, line 1743: nits: for helpful debbuging, could you add some `{}`s like the other `assert!`?

```rust
assert!(.., `There is no account to replace native program ({}, {}) with.`, name, program_id);
```

ryoqun (dismissed):
lgtm with a nit

Thanks for improving the native program initialization mechanism!

runtime/src/bank.rs
CriesofCarrots: Moving this assert breaks a bunch of snapshot tests once a `NewVersion` builtin is added. What do you recommend?

ryoqun: @CriesofCarrots Blame me, I completely forget that this fn must retain idempotent of AccountsDB; so more subtle adjustment was needed: https://github.com/solana-labs/solana/pull/13403/commits/2a9ff42055427c632d4eb2d81ea26b609a14f432 Hope this commit explains well. If still not sure, please say so. I'll update the in-source comment.

Sorry for confusing you because of my unreadable code. ;)

runtime/src/bank.rs
```diff
@@ -9334,13 +9359,25 @@ mod tests {
         assert_eq!(bank.get_account_modified_slot(&program_id).unwrap().1, slot);
 
         let mut bank = Arc::new(new_from_parent(&bank));
+        // When replacing native_program, name must change to disambiguate from repeated
+        // invocations.
```
CriesofCarrots, line 9363: I think we should put a comment about this in builtins.rs, since it's critical to know when constructing the name

CriesofCarrots: I'll sneak that into #13394

runtime/src/bank.rs
```diff
@@ -1715,16 +1715,15 @@ impl Bank {
...
-            already_genuine_program_exists = native_loader::check_id(&account.owner);
-
-            if !already_genuine_program_exists {
+    // NOTE: must hold idempotent for the same set of arguments
+    pub fn add_native_program(&self, name: &str, program_id: &Pubkey, must_replace: bool) {
+        let mut existing_genuine_program = self.get_account(&program_id);
+        if let Some(account) = &mut existing_genuine_program {
+            if !native_loader::check_id(&account.owner) {
                 // malicious account is pre-occupying at program_id
                 // forcibly burn and purge it
```
ryoqun, line 1728: well this logic changes introduced a regression where we could be tricked into skipping to create native program because `existing_genuine_program` can be set to a bad account and not reset and we enters this `match` arm returning from this `fn`: https://github.com/solana-labs/solana/pull/13403/files#diff-ed47b4a0198313377e091bb3957bbbc63d937805426d1b2b6de39d0a50d32a0cR1760

Fixed and test is imrproved at #13403.

runtime/src/bank.rs
```diff
@@ -1737,19 +1736,54 @@ impl Bank {
...
+        assert!(
+            !self.is_frozen(),
+            "Can't change frozen bank by adding not-existing new native program ({}, {}). \
+            Maybe, inconsistent program activation is detected on snapshot restore?",
+            name,
+            program_id
+        );
+
+        // Add a bogus executable native account, which will be loaded and ignored.
+        let account = native_loader::create_loadable_account(name);
```
ryoqun, line 1784: Also, when replacing, we must consider the balance of existing program account otherwise we would cause capitalization mismatch if the replaced native program isn't 1 lamports.

Fixed and test added at #13884.

PR #13409, opened 2020-11-05
Add cargo-test-bpf

mvines:
When testing a program crate, you typically need to build the crate as BPF and then the tests for the host.  The workflow for this is usually:
```
$ cargo build-bpf
$ bpf=1 cargo test ...
```
where `bpf=1` is a signal to the `solana-program-test` crate to load your program as BPF instead of for the host.   

But for a normal crate the workflow is simply `cargo test ...`.   In some of the SPL program crates, we've attempted to simulate the normal `cargo test` workflow by adding `build.rs` files that automatically run `cargo build-bpf`.  This is not nice for various reasons, such as:
1. You end up building the crate twice, for BPF and host
2. The `build.rs` requires a number of exceptions to avoid clippy and xargo
3. docs.rs obviously doesn't have `cargo build-bpf` so documentation generation on crate publishing doesn't work
4. The env var approach, `bpf=1`, is clunky

Introducing `cargo test-bpf`, which attempts to wrap up this workflow into a single command that mirrors what `cargo test` does for normal crates.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3763,6 +3763,7 @@ dependencies = [
 name = "solana-cargo-test-bpf"
 version = "1.5.0"
 dependencies = [
+ "cargo_metadata",
  "clap",
 ]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3760,6 +3760,13 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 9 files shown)


mvines:
(to draft, there are some bugs still...)

Issue #13296, opened 2020-10-30
explorer: Use consistent titles for SYSVAR accounts


oJshua:
#### Problem
Throughout Explorer, sysvar accounts are capitalized e.g. SYSVAR_SLOT_HASHES, SYSVAR_CLOCK. This should be the convention for the account pages as well. See: https://explorer.solana.com/address/SysvarS1otHashes111111111111111111111111111

#### Proposed Solution
Use suggested convention.

oJshua:
Per conversation with @mvines , we will actually switch sysvar names to 'Title Case' throughout Explorer.

PR #13428, opened 2020-11-05
explorer: Capitalize and underscore SYSVAR titles

oJshua:
#### Problem
Throughout Explorer, sysvar accounts are capitalized e.g. SYSVAR_SLOT_HASHES, SYSVAR_CLOCK. This should be the convention for the account pages as well. See: https://explorer.solana.com/address/SysvarS1otHashes111111111111111111111111111

#### Summary of Changes
Capitalizes and underscores sysvar titles.

Fixes https://github.com/solana-labs/solana/issues/13296

### DIFF
--- a/explorer/src/components/account/SysvarAccountSection.tsx
+++ b/explorer/src/components/account/SysvarAccountSection.tsx
@@ -105,7 +105,7 @@ function SysvarAccountRecentBlockhashesCard({
   return (
     <div className="card">
       <AccountHeader
-        title="Sysvar Recent Blockhashes"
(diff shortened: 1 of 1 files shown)


CriesofCarrots:
I can't remember where, but I think @mvines suggested sysvars be less shouty on the explorer. Could we use Title Case instead of screaming-snake?

oJshua:
I actually changed these to shouty in response to some nits @mvines passed my way. @mvines would you prefer full caps or Title Case? I can revert back if necessary.

mvines:
I Personally Far Prefer Title Case OVER_YELLING_SNAKE_CASE

Issue #10442, opened 2020-06-06
RPC account change notifications are not sent when an account is deleted


jstarry:
#### Problem
The RPC pubsub api allows subscribing to account changes in realtime. However, when account balance goes to 0, no notification is sent. This is a problem because clients with a live payer account need to know when an account's balance is depleted

#### Proposed Solution
Notify when account is deleted.

PR #13440, opened 2020-11-06
Send RPC notification when account is deleted

jstarry:
#### Problem
The RPC pubsub api allows subscribing to account changes in realtime. However, when account balance goes to 0, no notification is sent. This is a problem because clients with a live payer account need to know when an account's balance is depleted

#### Summary of Changes
Send a notification if an account is not found and last notified slot > 0

Fixes https://github.com/solana-labs/solana/issues/10442

### DIFF
--- a/core/src/rpc_subscriptions.rs
+++ b/core/src/rpc_subscriptions.rs
@@ -182,7 +182,7 @@ where
     K: Eq + Hash + Clone + Copy,
     S: Clone + Serialize,
     B: Fn(&Bank, &K) -> X,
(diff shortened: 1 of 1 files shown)


CriesofCarrots (approved):
Thanks, this lgtm!

Unfortunately, program subscriptions don't notify on account deletion either; I haven't thought of a clever/efficient way to mirror this fix. Any thoughts?

jstarry:
> Unfortunately, program subscriptions don't notify on account deletion either; I haven't thought of a clever/efficient way to mirror this fix. Any thoughts?

Nope, didn't think too hard about it though. It felt like a weird dev experience too because you would get random account notifications that won't have the owner set to the program you are subscribed to.

PR #13460, opened 2020-11-07
Clean up Delegation::stake_and_activating

ryoqun:
#### Problem

The staking calculation code is a bit hard to read (at least, for me).

#### Summary of Changes

Make it readable by introducing a common theme/doctrine of the processing and fix various consistency issues, which hampers reading.

(I'll add detailed commentary as pr comments)

Well, I think I found another bug where stake can be skewed in this code. So, this is a blocking pr.

Also, I make another follow-up cleaning pr for `stake_activating_and_deactivating`.

### DIFF
--- a/programs/stake/src/stake_state.rs
+++ b/programs/stake/src/stake_state.rs
@@ -263,54 +263,81 @@ impl Delegation {
         }
     }
 
-    fn stake_and_activating(&self, epoch: Epoch, history: Option<&StakeHistory>) -> (u64, u64) {
+    fn stake_and_activating(
+        &self,
+        target_epoch: Epoch,
+        history: Option<&StakeHistory>,
+    ) -> (u64, u64) {
         if self.is_bootstrap() {
+            // fully effective immediately
             (self.stake, 0)
-        } else if epoch == self.activation_epoch {
+        } else if target_epoch == self.activation_epoch {
+            // all is activating
             (0, self.stake)
-        } else if epoch < self.activation_epoch {
+        } else if target_epoch < self.activation_epoch {
+            // not yet enabled
             (0, 0)
-        } else if let Some((history, mut entry)) = history.and_then(|history| {
-            history
-                .get(&self.activation_epoch)
-                .map(|entry| (history, entry))
-        }) {
-            // && !is_bootstrap() && epoch > self.activation_epoch
-            let mut effective_stake = 0;
-            let mut next_epoch = self.activation_epoch;
+        } else if let Some((history, mut prev_epoch, mut prev_cluster_stake)) =
+            history.and_then(|history| {
+                history
+                    .get(&self.activation_epoch)
+                    .map(|cluster_stake_at_activation_epoch| {
+                        (
+                            history,
+                            self.activation_epoch,
+                            cluster_stake_at_activation_epoch,
+                        )
+                    })
+            })
+        {
+            // target_epoch > self.activation_epoch
 
-            // loop from my activation epoch until the current epoch
-            //   summing up my entitlement
+            // loop from my activation epoch until the target epoch summing up my entitlement
+            // current stake is updated using its previous epoch's cluster stake
+            let mut current_epoch;
+            let mut current_effective_stake = 0;
             loop {
-                if entry.activating == 0 {
+                current_epoch = prev_epoch + 1;
+                // if there is no activating stake at prev epoch, we should have been fully effective at this moment
+                if prev_cluster_stake.activating == 0 {
                     break;
                 }
+
                 // how much of the growth in stake this account is
                 //  entitled to take
-                let weight = (self.stake - effective_stake) as f64 / entry.activating as f64;
+                let remaining_activating_stake = self.stake - current_effective_stake;
+                let weight =
+                    remaining_activating_stake as f64 / prev_cluster_stake.activating as f64;
 
-                // portion of activating stake in this epoch I'm entitled to
-                effective_stake +=
-                    ((weight * entry.effective as f64 * self.warmup_cooldown_rate) as u64).max(1);
+                // portion of newly effective cluster stake I'm entitled to at current epoch
+                let newly_effective_cluster_stake =
+                    prev_cluster_stake.effective as f64 * self.warmup_cooldown_rate;
+                let newly_effective_stake =
+                    ((weight * newly_effective_cluster_stake) as u64).max(1);
 
-                if effective_stake >= self.stake {
-                    effective_stake = self.stake;
+                current_effective_stake += newly_effective_stake;
+                if current_effective_stake >= self.stake {
+                    current_effective_stake = self.stake;
                     break;
                 }
 
-                next_epoch += 1;
-                if next_epoch >= epoch || next_epoch >= self.deactivation_epoch {
+                if current_epoch >= target_epoch || current_epoch >= self.deactivation_epoch {
                     break;
                 }
-                if let Some(next_entry) = history.get(&next_epoch) {
-                    entry = next_entry;
+                if let Some(current_cluster_stake) = history.get(&current_epoch) {
+                    prev_epoch = current_epoch;
+                    prev_cluster_stake = current_cluster_stake;
                 } else {
                     break;
                 }
             }
-            (effective_stake, self.stake - effective_stake)
+
+            (
+                current_effective_stake,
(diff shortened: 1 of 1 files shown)


programs/stake/src/stake_state.rs
```diff
@@ -263,54 +263,81 @@ impl Delegation {
         }
     }
 
-    fn stake_and_activating(&self, epoch: Epoch, history: Option<&StakeHistory>) -> (u64, u64) {
```
ryoqun, line 266: let's qualify what kind of `epoch` it is.

programs/stake/src/stake_state.rs
```diff
@@ -263,54 +263,81 @@ impl Delegation {
...
             (self.stake, 0)
-        } else if epoch == self.activation_epoch {
+        } else if target_epoch == self.activation_epoch {
+            // all is activating
             (0, self.stake)
-        } else if epoch < self.activation_epoch {
+        } else if target_epoch < self.activation_epoch {
+            // not yet enabled
             (0, 0)
-        } else if let Some((history, mut entry)) = history.and_then(|history| {
```
ryoqun, line 273: entry is too generic. It doesn't add any information...

programs/stake/src/stake_state.rs
```diff
@@ -263,54 +263,81 @@ impl Delegation {
...
-            history
-                .get(&self.activation_epoch)
-                .map(|entry| (history, entry))
-        }) {
-            // && !is_bootstrap() && epoch > self.activation_epoch
```
ryoqun, line 278: Comments are generally for human; no need to spell out every exceptional conditionals: `!is_bootstrap()`

programs/stake/src/stake_state.rs
```diff
@@ -263,54 +263,81 @@ impl Delegation {
...
                 }
             }
-            (effective_stake, self.stake - effective_stake)
+
+            (
+                current_effective_stake,
+                self.stake - current_effective_stake,
+            )
         } else {
-            // no history or I've dropped out of history, so assume fully activated
```
ryoqun, line 313: let's make a clear distinction between `activated` /  `effective`.

programs/stake/src/stake_state.rs
```diff
@@ -263,54 +263,81 @@ impl Delegation {
...
+
+                current_effective_stake += newly_effective_stake;
+                if current_effective_stake >= self.stake {
+                    current_effective_stake = self.stake;
                     break;
                 }
 
-                next_epoch += 1;
-                if next_epoch >= epoch || next_epoch >= self.deactivation_epoch {
+                if current_epoch >= target_epoch || current_epoch >= self.deactivation_epoch {
```
ryoqun, line 324: IMO, this loop termination condition is really self-evident now.

programs/stake/src/stake_state.rs
```diff
@@ -263,54 +263,81 @@ impl Delegation {
...
                     break;
                 }
-                if let Some(next_entry) = history.get(&next_epoch) {
-                    entry = next_entry;
+                if let Some(current_cluster_stake) = history.get(&current_epoch) {
+                    prev_epoch = current_epoch;
+                    prev_cluster_stake = current_cluster_stake;
```
ryoqun, line 329: I grouped looping state mutation in this sole place this way. Also, introduced consistent and reasonable cognitive aid of `prev_`/`current_` prefixing.

programs/stake/src/stake_state.rs
```diff
@@ -263,54 +263,81 @@ impl Delegation {
...
+            // not yet enabled
             (0, 0)
-        } else if let Some((history, mut entry)) = history.and_then(|history| {
-            history
-                .get(&self.activation_epoch)
-                .map(|entry| (history, entry))
-        }) {
-            // && !is_bootstrap() && epoch > self.activation_epoch
-            let mut effective_stake = 0;
-            let mut next_epoch = self.activation_epoch;
```
ryoqun, line 280: I think `next_` was the root cause of confusion.

programs/stake/src/stake_state.rs
```diff
@@ -263,54 +263,81 @@ impl Delegation {
...
+                        )
+                    })
+            })
+        {
+            // target_epoch > self.activation_epoch
 
-            // loop from my activation epoch until the current epoch
-            //   summing up my entitlement
+            // loop from my activation epoch until the target epoch summing up my entitlement
+            // current stake is updated using its previous epoch's cluster stake
```
ryoqun, line 296: I'm here locally introducing a jargon called `cluster stake` for the name of variables of the `StakeHistoryEntry` type. I think the jargon better describes what it is than the straight snake_case version of the type name in this context.

programs/stake/src/stake_state.rs
ryoqun: Also moved the renamed equivalent of this into the preceding `if let`. The new code a bit ugly. But I wanted to group all loop variable initialization in one place.

programs/stake/src/stake_state.rs
```diff
@@ -263,54 +263,81 @@ impl Delegation {
...
                 }
 
-                next_epoch += 1;
-                if next_epoch >= epoch || next_epoch >= self.deactivation_epoch {
+                if current_epoch >= target_epoch || current_epoch >= self.deactivation_epoch {
                     break;
                 }
-                if let Some(next_entry) = history.get(&next_epoch) {
-                    entry = next_entry;
+                if let Some(current_cluster_stake) = history.get(&current_epoch) {
```
ryoqun, line 327: The whole various naming makes us to read this line like this even without the context:

> `history` is something which returns something historical (indexed by epoch) which is related to cluster-wide stake information for given epoch. Maybe some kind of totals of stakes. And we're getting new one for current_epoch and swap with prev for next iteration.

programs/stake/src/stake_state.rs
```diff
@@ -263,54 +263,81 @@ impl Delegation {
...
+        } else if target_epoch < self.activation_epoch {
+            // not yet enabled
             (0, 0)
-        } else if let Some((history, mut entry)) = history.and_then(|history| {
-            history
-                .get(&self.activation_epoch)
-                .map(|entry| (history, entry))
-        }) {
-            // && !is_bootstrap() && epoch > self.activation_epoch
-            let mut effective_stake = 0;
```
ryoqun, line 279: let's be explicit that this variable is  a mutable loop state  by prefixing some temporary adjective. (I picked `current_`).

programs/stake/src/stake_state.rs
```diff
@@ -263,54 +263,81 @@ impl Delegation {
...
                     break;
                 }
+
                 // how much of the growth in stake this account is
                 //  entitled to take
-                let weight = (self.stake - effective_stake) as f64 / entry.activating as f64;
-
-                // portion of activating stake in this epoch I'm entitled to
-                effective_stake +=
-                    ((weight * entry.effective as f64 * self.warmup_cooldown_rate) as u64).max(1);
```
ryoqun, line 294: let's break up important formula into smaller chunks so that we can chew on one by one.

programs/stake/src/stake_state.rs
ryoqun, line 296: And this is the succinct expression of the common theme/doctrine behind this looping, clearly expressed and strictly made consistent in the code as well.

programs/stake/src/stake_state.rs
```diff
@@ -263,54 +263,81 @@ impl Delegation {
...
+            })
+        {
+            // target_epoch > self.activation_epoch
 
-            // loop from my activation epoch until the current epoch
-            //   summing up my entitlement
+            // loop from my activation epoch until the target epoch summing up my entitlement
+            // current stake is updated using its previous epoch's cluster stake
+            let mut current_epoch;
+            let mut current_effective_stake = 0;
```
ryoqun, line 298: a bit of delicacy, but these `current_` grouped variables are rhymed with these `prev_`s: https://github.com/solana-labs/solana/pull/13460/files#diff-ab7997cf6770a6fae7b258e433b5dd8c33afa30431eab212552bc7f2b7b38db1R280

ryoqun:
@mvines In a similar spirit of #13325, could you enjoy reading this? :) if you're too busy, I'll ask others to review (CC: my :eyes:  on @CriesofCarrots or @t-nelson )

mvines (approved):
looks great

PR #11966, opened 2020-09-01
Ensure that the spl-token 2 native mint account is owned by the spl-token 2 program

mvines:
Workaround for https://github.com/solana-labs/solana-program-library/issues/374 until spl-token 3 is shipped

PR #13461, opened 2020-11-07
Rewrite stake accounts for clear migration

ryoqun:
#### Problem

~this is very much still draft.~ this is done job!

bugs fixed by #13357 and #13358 created many kinds of bad stake accounts.

#### Summary of Changes

~todo: add test.~ done

Fixes #

### DIFF
--- a/ledger-tool/src/main.rs
+++ b/ledger-tool/src/main.rs
@@ -2052,7 +2052,15 @@ fn main() {
                                     feature_account_balance,
                                 ),
                             );
+                            base_bank.store_account(
+                                &feature_set::rewrite_stake::id(),
+                                &feature::create_account(
+                                    &Feature { activated_at: None },
+                                    feature_account_balance,
+                                ),
+                            );
 
+                            let mut stole_failed_count = 0;
                             if base_bank
                                 .get_account(&feature_set::secp256k1_program_enabled::id())
                                 .is_some()
@@ -2064,6 +2072,21 @@ fn main() {
                                     &Account::default(),
                                 );
                             } else {
+                                stole_failed_count += 1;
+                            }
+
+                            if base_bank
+                                .get_account(&feature_set::instructions_sysvar_enabled::id())
+                                .is_some()
+                            {
+                                base_bank.store_account(
+                                    &feature_set::instructions_sysvar_enabled::id(),
+                                    &Account::default(),
+                                );
+                            } else {
+                                stole_failed_count += 1;
+                            }
+                            if stole_failed_count >= 1 {
                                 // we have no choice; maybe locally created blank cluster with
                                 // not-Development cluster type.
                                 let old_cap = base_bank.set_capitalization();
@@ -2073,7 +2096,10 @@ fn main() {
                                      requested: increasing {} from {} to {}",
                                     feature_account_balance, old_cap, new_cap,
                                 );
-                                assert_eq!(old_cap + feature_account_balance, new_cap);
+                                assert_eq!(
+                                    old_cap + feature_account_balance * stole_failed_count,
+                                    new_cap
+                                );
                             }
                         }
 
--- a/ledger-tool/src/main.rs
+++ b/ledger-tool/src/main.rs
@@ -2060,7 +2060,7 @@ fn main() {
                                 ),
                             );
 
-                            let mut stole_failed_count = 0;
+                            let mut store_failed_count = 0;
                             if base_bank
                                 .get_account(&feature_set::secp256k1_program_enabled::id())
                                 .is_some()
@@ -2072,7 +2072,7 @@ fn main() {
                                     &Account::default(),
                                 );
                             } else {
-                                stole_failed_count += 1;
+                                store_failed_count += 1;
                             }
 
                             if base_bank
@@ -2084,9 +2084,9 @@ fn main() {
                                     &Account::default(),
                                 );
                             } else {
-                                stole_failed_count += 1;
+                                store_failed_count += 1;
                             }
-                            if stole_failed_count >= 1 {
+                            if store_failed_count >= 1 {
                                 // we have no choice; maybe locally created blank cluster with
                                 // not-Development cluster type.
                                 let old_cap = base_bank.set_capitalization();
@@ -2097,7 +2097,7 @@ fn main() {
                                     feature_account_balance, old_cap, new_cap,
                                 );
                                 assert_eq!(
-                                    old_cap + feature_account_balance * stole_failed_count,
+                                    old_cap + feature_account_balance * store_failed_count,
                                     new_cap
                                 );
                             }
--- a/programs/stake/src/stake_state.rs
+++ b/programs/stake/src/stake_state.rs
@@ -166,6 +166,24 @@ impl Meta {
         }
         Ok(())
     }
+
+    pub fn reduce_overage_rent_exempt_reserve(
+        &mut self,
+        rent: &Rent,
+        data_len: usize,
+    ) -> Option<(u64, u64)> {
+        let corrected_rent_exempt_reserve = rent.minimum_balance(data_len);
+        if corrected_rent_exempt_reserve != self.rent_exempt_reserve {
+            // we forcibly update rent_excempt_reserve even if rent_exempt_reserve > account
+            // balance, hoping user might restore rent_exempt status by depositing.
+            // TODO: split is overflow safe?
+            let (old, new) = (self.rent_exempt_reserve, corrected_rent_exempt_reserve);
+            self.rent_exempt_reserve = corrected_rent_exempt_reserve;
+            Some((old, new))
+        } else {
+            None
+        }
+    }
 }
 
 #[derive(Debug, Serialize, Deserialize, PartialEq, Clone, Copy, AbiExample)]
@@ -687,6 +705,25 @@ impl Stake {
             Ok(())
         }
     }
+
+    fn reduce_overage(
+        &mut self,
+        account_balance: u64,
+        rent_exempt_balance: u64,
+    ) -> Option<(u64, u64)> {
+        // at this moment, balance >>> rent_excempt_balance (no overflow risk)
+        //   => no there could be possibility of non-rent-exempt stake accounts
+        // this will overwrite innocent deactivating and immeditealy withdrawn stake accounts as well is this ok?
+        let corrected_stake = account_balance.saturating_sub(rent_exempt_balance);
+        if self.delegation.stake > corrected_stake {
+            // this could result in creating 0 stake account..; is this safe?
+            let (old, new) = (self.delegation.stake, corrected_stake);
+            self.delegation.stake = corrected_stake;
+            Some((old, new))
+        } else {
+            None
+        }
+    }
 }
 
 pub trait StakeAccount {
@@ -1184,6 +1221,45 @@ fn calculate_split_rent_exempt_reserve(
     let lamports_per_byte_year =
         source_rent_exempt_reserve / (source_data_len + ACCOUNT_STORAGE_OVERHEAD);
     lamports_per_byte_year * (split_data_len + ACCOUNT_STORAGE_OVERHEAD)
+}
+
+pub type ReducedOverageStakeResult = (&'static str, (u64, u64), (u64, u64));
+
+pub fn reduce_overage_stakes(
+    stake_account: &mut Account,
+    rent: &Rent,
+) -> Result<ReducedOverageStakeResult, InstructionError> {
+    match stake_account.state()? {
+        StakeState::Initialized(mut meta) => {
+            let meta_result =
+                meta.reduce_overage_rent_exempt_reserve(rent, stake_account.data.len());
+
+            if meta_result.is_none() {
+                return Err(InstructionError::InvalidAccountData);
+            }
+
+            stake_account.set_state(&StakeState::Initialized(meta))?;
+            Ok(("initialized", meta_result.unwrap_or_default(), (0, 0)))
+        }
+        StakeState::Stake(mut meta, mut stake) => {
+            let meta_result =
+                meta.reduce_overage_rent_exempt_reserve(rent, stake_account.data.len());
+            let stake_result =
+                stake.reduce_overage(stake_account.lamports, meta.rent_exempt_reserve);
+
+            if meta_result.is_none() && stake_result.is_none() {
+                return Err(InstructionError::InvalidAccountData);
+            }
+
+            stake_account.set_state(&StakeState::Stake(meta, stake))?;
+            Ok((
+                "stake",
+                meta_result.unwrap_or_default(),
+                stake_result.unwrap_or_default(),
+            ))
+        }
+        _ => Err(InstructionError::InvalidAccountData),
+    }
 }
 
 // utility function, used by runtime::Stakes, tests
--- a/programs/stake/src/stake_state.rs
+++ b/programs/stake/src/stake_state.rs
@@ -167,16 +167,16 @@ impl Meta {
         Ok(())
     }
 
-    pub fn reduce_overage_rent_exempt_reserve(
+    pub fn rewrite_rent_exempt_reserve(
         &mut self,
         rent: &Rent,
         data_len: usize,
     ) -> Option<(u64, u64)> {
         let corrected_rent_exempt_reserve = rent.minimum_balance(data_len);
         if corrected_rent_exempt_reserve != self.rent_exempt_reserve {
-            // we forcibly update rent_excempt_reserve even if rent_exempt_reserve > account
-            // balance, hoping user might restore rent_exempt status by depositing.
-            // TODO: split is overflow safe?
+            // We forcibly update rent_excempt_reserve even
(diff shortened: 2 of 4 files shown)


programs/stake/src/stake_state.rs
```diff
@@ -153,6 +153,15 @@ impl Meta {
         }
         Ok(())
     }
+
+    pub fn reduce_overage_rent_exempt_reserve(&mut self, rent: &Rent, data_len: usize) {
+        let rent_exempt_reserve = rent.minimum_balance(data_len);
+        if rent_exempt_reserve != self.rent_exempt_reserve {
+            // we forcibly update rent_excempt_reset even if rent_exempt_reserve > account balance
+            // because user might restore rent_exempt status
```
ryoqun, line 161: wording: `in the hope`.

runtime/src/bank.rs
```diff
@@ -1255,6 +1266,31 @@ impl Bank {
...
     }
 
+    fn reduce_overage_stakes(&self) {
+        // is this really all of parsable accounts owned by stake11111?
+        self.cloned_stake_delegations()
+            .into_iter()
+            .for_each(|(stake_pubkey, delegation)| {
+                match (
+                    self.get_account(&stake_pubkey),
+                    self.get_account(&delegation.voter_pubkey),
```
ryoqun, line 1276: well doesn't vote at all...

programs/stake/src/stake_state.rs
```diff
@@ -560,6 +569,15 @@ impl Stake {
...
         }
     }
+
+    fn reduce_overage(&mut self, balance: u64, rent_exempt_balance: u64) {
+        // at this moment, balance >>> rent_excempt_balance (no overflow risk)
+        //   => no there could be possibility of non-rent-exempt stake accounts
+        let correct_stake = balance.saturating_sub(rent_exempt_balance);
+        if self.delegation.stake > correct_stake {
+            self.delegation.stake = correct_stake;
+        }
```
ryoqun, line 579: note that we can't the opposite thing of rewriting to stake more which has miss to be redelegated. We can't detect the good order (deposit/undelegate => delegate) from bad order (undelegate => delegate => deposit).

runtime/src/bank.rs
```diff
@@ -871,6 +870,21 @@ impl Bank {
         if parent.epoch() < new.epoch() {
             new.apply_feature_activations(false);
         }
+        // feature gate with the stake_program_v2!
+        // only run once for the entire lifetime of given cluster?
+        // add comment: remove me after a while v1.6?
+        // also only do at the start of epoch; every slot is too costly
+        if !new.lazy_rent_collection.load(Relaxed) {
```
ryoqun, line 877: I'm free-riding on `lazy_rent_collection` to pass on CI. needs proper care.

runtime/src/genesis_utils.rs
```diff
@@ -10,11 +11,15 @@ use solana_sdk::{
     system_program,
 };
 use solana_stake_program::stake_state;
+use solana_stake_program::stake_state::StakeState;
 use solana_vote_program::vote_state;
 use std::borrow::Borrow;
 
 // The default stake placed with the bootstrap validator
-pub const BOOTSTRAP_VALIDATOR_LAMPORTS: u64 = 42;
```
ryoqun, line 17: When checking the overall effect of this overage rewrite, I noticed that many tests are actually creating really small stake accounts under the rent-exempt threshold. I wanted to get rid of them because tests doesn't reflect reality and it's confusing. Also, to get good exposure, I wanted overage check always-on when testing.

After some dirties hacks, it turns out this is doable. (now ci passes).

ryoqun:
- [ ] add another feature gate specifically for this.

programs/stake/src/stake_state.rs
```diff
@@ -687,6 +705,21 @@ impl Stake {
             Ok(())
         }
     }
+
+    fn reduce_overage(&mut self, balance: u64, rent_exempt_balance: u64) -> Option<(u64, u64)> {
+        // at this moment, balance >>> rent_excempt_balance (no overflow risk)
+        //   => no there could be possibility of non-rent-exempt stake accounts
+        // this will overwrite innocent deactivating and immeditealy withdrawn stake accounts as well is this ok?
```
CriesofCarrots, line 712: I think the negative effect here is that the deactivating stake will potentially take longer to fully deactivate, correct?
That seems not terrible to me. The alternative (not correcting deactivating accounts) seems riskier to me.

programs/stake/src/stake_state.rs
```diff
@@ -166,6 +166,24 @@ impl Meta {
...
+    pub fn reduce_overage_rent_exempt_reserve(
+        &mut self,
+        rent: &Rent,
+        data_len: usize,
+    ) -> Option<(u64, u64)> {
+        let corrected_rent_exempt_reserve = rent.minimum_balance(data_len);
+        if corrected_rent_exempt_reserve != self.rent_exempt_reserve {
+            // we forcibly update rent_excempt_reserve even if rent_exempt_reserve > account
+            // balance, hoping user might restore rent_exempt status by depositing.
+            // TODO: split is overflow safe?
```
CriesofCarrots, line 179: What is your concern here?

programs/stake/src/stake_state.rs
```diff
@@ -687,6 +705,21 @@ impl Stake {
...
         }
     }
+
+    fn reduce_overage(&mut self, balance: u64, rent_exempt_balance: u64) -> Option<(u64, u64)> {
+        // at this moment, balance >>> rent_excempt_balance (no overflow risk)
+        //   => no there could be possibility of non-rent-exempt stake accounts
+        // this will overwrite innocent deactivating and immeditealy withdrawn stake accounts as well is this ok?
+        let corrected_stake = balance.saturating_sub(rent_exempt_balance);
+        if self.delegation.stake > corrected_stake {
+            // this could result in creating 0 stake account..; is this safe?
```
CriesofCarrots, line 715: I *think* as long as 0 stake generates 0 rewards, it should be safe. How were you thinking we'd check this?

runtime/src/bank.rs
```diff
@@ -1292,6 +1334,41 @@ impl Bank {
...
+                            format!("reduced overage stake: {}, {:?}", stake_pubkey, result,);
+                        info!("{}", message);
+                        datapoint_info!("stake_info", ("info", message, String));
+                        count += 1;
+                    }
+                }
+            });
+
+        let msg = format!(
+            "bank (slot: {}): reduce_overage_stakes: {} accounts rewritten",
```
CriesofCarrots, line 1360: I think a single datapoint that reports # delegations examined and # accounts rewritten would be useful.

programs/stake/src/stake_state.rs
ryoqun: Given large stake account split from small stake account, this rewrite logic could create a situation where `newly rewritten rent exempt reserve` > `stake account balance`.

In that case, further splitting the rewritten account could trigger integer overflow here, maybe?:

https://github.com/solana-labs/solana/blob/3827739708962b0e46cde6b604eea0a03292664c/programs/stake/src/stake_state.rs#L913

ryoqun:
> - [ ] add another feature gate specifically for this.

Also, I noticed that this calculation could be off without rewriting all `rent_exempt_reserve` and only enabling `stake_program_v2`. Because stake account size and rent exempt reserve isn't guaranteed to be in sync at the moment unfortunately, maybe?:

https://github.com/solana-labs/solana/blob/3827739708962b0e46cde6b604eea0a03292664c/programs/stake/src/stake_state.rs#L902-L906

programs/stake/src/stake_state.rs
ryoqun: yeah, I agree with rewriting those innocent accounts at once for all for reduced risk.

The exact repercussion is that stake account's user-visible `delegated stake` will suddenly reduced. I hope this doesn't cause any unneeded confusion. I mean, it should affect the sudden reduction of `delegated amount` in CLI and explorer, staking-enabled wallets for stake accounts which aren't affected by bug but rewritten by this.

> deactivating stake will potentially take longer to fully deactivate

Well, this is not true. Yeah, staking calc is very difficult to grasp only in mind. The conclusion is that the rate and overall needed duration (in epoch) of deactivation shouldn't be affected.

I wrote a simple thought experiment like this:

```
A user created 100 SOL stake account some epochs ago and warmed up 40 SOL (= effective stake) up to N-1 epoch 
Then the user deactivates and withdraws 60 SOL at N-1 epoch.
For simplicity, assume no other deactivating cluster stake, no rent exempt calculation

a: if the overage rewrite is run at the start of N epoch:
a.1 `delegation.stake` is reduced to 40 SOL.
a.2 recalculated effective stake would be 16 SOL at N-1 epoch (because delegation.stake is reduced by 2.5x from 100SOL to 40SOL)
a.3 new effective (= now deactivating) stake would be 12 SOL at N epoch (= 16 - (16 * 25%))
a.4 next effective (= deactivating) stake would be 9 SOL at N+1 epoch (= 12 - (12 * 25%))

b: if the overage rewrite isn't run at the start of N epoch:
b.1 `delegation.stake` is remained to be 100 SOL.
b.2 recalculated effective stake would remain 40SOL at N-1 epoch.
b.3 new effective (= now deactivating) stake would be 30 SOL at N epoch (= 40 - (40 * 25%)).
b.4 next effective (= deactivating) stake would be 22.5 SOL at N+1 epoch (= 30 - (30 * 25%))
```

runtime/src/genesis_utils.rs
ryoqun: > To confirm: only (3) really matters, but you're changing all of them for because it's easier, right?

Yeah, I didn't dig in at all. I just wanted to pass the CI to know my rewrite's full consequences using our test assets.



> Could you pull that change out into a separate PR? It's quite a bit of churn on this one.

Yeah, I'll create a separate pr.

ryoqun: or your recent #13627 improves the situation? :) I'll take a look at it, too.

programs/stake/src/stake_state.rs
CriesofCarrots: Fixing here: https://github.com/solana-labs/solana/pull/13649

programs/stake/src/stake_state.rs
ryoqun: yeah, I've changed the behavior here.

runtime/src/bank.rs
```diff
@@ -890,6 +902,35 @@ impl Bank {
...
+                // to exercise for this rarely touched code, do once per every epoch on
+                // less critical system than mainnet
+                self.epoch() > parent_epoch
+            }
+            ClusterType::MainnetBeta => {
+                // to avoid any potential risk of wrongly rewriting accounts in the future,
+                // only do this once on mainnet, sacrificing small risk of unknown
+                // bugs which again creates bad stake accounts..
+                Some(self.epoch()) == self.stake_program_v2_epoch()
+            }
```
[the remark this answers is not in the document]
ryoqun: Yeah, I've changed my mind. I like your thinking for consistency. I largely simplified the logic here.

programs/stake/src/stake_state.rs
ryoqun: > How were you thinking we'd check this?

I checked by use of :eyes:

Or, if you want extra safety, I think I can write tests for stake/inflation/split/merge/withdraw for 0 lamport accounts... A bit daunting work. ;)

runtime/src/bank.rs
ryoqun: yeah, good idea. :)

ledger-tool/src/main.rs
```diff
@@ -2052,7 +2052,15 @@ fn main() {
...
                             );
+                            base_bank.store_account(
+                                &feature_set::rewrite_stake::id(),
+                                &feature::create_account(
+                                    &Feature { activated_at: None },
+                                    feature_account_balance,
+                                ),
+                            );
 
+                            let mut stole_failed_count = 0;
```
CriesofCarrots, line 2063:
```suggestion
                            let mut store_failed_count = 0;
```

runtime/src/bank.rs
```diff
@@ -890,6 +900,27 @@ impl Bank {
         new
     }
 
+    fn enable_stake_rewrite(&self, parent_epoch: Epoch) -> bool {
+        if self.no_stake_rewrite.load(Relaxed) {
+            return false;
+        }
+
+        let activated_slot = if let Some(epoch) = self.rewrite_stake_slot() {
+            epoch
```
CriesofCarrots, line 909:
```suggestion
        let activated_slot = if let Some(slot) = self.rewrite_stake_slot() {
            slot
```

runtime/src/bank.rs
```diff
@@ -890,6 +900,27 @@ impl Bank {
...
+        };
+        if activated_slot == 0 {
+            return self.slot() == 0;
+        }
+        let (activated_epoch, _) = self.get_epoch_and_slot_index(activated_slot);
+
+        // to avoid any potential risk of wrongly rewriting accounts in the future,
+        // only do this once, taking small risk of unknown
+        // bugs which again creates bad stake accounts..
+        self.epoch() > parent_epoch && self.epoch() == activated_epoch
```
CriesofCarrots, line 921: I don't grok why we need to check epochs here. Wouldn't `self.slot() == activated_slot` be correct?

runtime/src/bank.rs
```diff
@@ -866,9 +870,15 @@ impl Bank {
...
-        if parent.epoch() < new.epoch() {
+        let parent_epoch = parent.epoch();
+        if parent_epoch < new.epoch() {
             new.apply_feature_activations(false);
         }
+
+        if new.enable_stake_rewrite(parent_epoch) {
+            // Remove me after a while around v1.6
+            new.rewrite_stakes();
+        }
```
CriesofCarrots, line 881: It's not immediately obvious to me why this can't be part of `apply_feature_activations`

CriesofCarrots (commented):
Fantastic, this seems really close! A couple nits, and I just have a couple quick questions about the feature activation (probably I'm missing something).

runtime/src/bank.rs
ryoqun: yeah, a bit of understanding of implied subtlety is needed. In short, `apply_feature_activations` is for proper state restoration of _non-AccountsDB `Bank` fields_. Here, we're rewriting AccountsDB state (stake accounts). `apply_feature_activation` is called from both normal child bank creation and both from snapshot restore.

I'll add a comment for this.

runtime/src/bank.rs
ryoqun: well, there is possibility of skipped slots. yeah hard to read correctly. ;)

CriesofCarrots: How could the activation slot be skipped? It's set in `new.apply_feature_activations() as `self.slot()` immediately before: https://github.com/solana-labs/solana/blob/acc40b1c9e54c7277f5369d20cd9f1e2fc88d1e7/runtime/src/bank.rs#L4116

runtime/src/bank.rs
CriesofCarrots: `reconfigure_token2_native_mint()` is in `apply_feature_activations`... is the difference there the capitalization adjustment?
Anyway, I think a comment would help, thanks!

runtime/src/bank.rs
ryoqun: yeah, i was wrong... fixed: https://github.com/solana-labs/solana/pull/13461/commits/4410a3599fa376e3eba82832c9222c8ab47798b3

runtime/src/bank.rs
ryoqun: hmm. you're right in this regard as well.. I've fixed https://github.com/solana-labs/solana/pull/13461/commits/a35d8ded5510834464f94ff5759a8e79c1eb1f50.

turned out we only needed special considerations for builtins.

CriesofCarrots: Nice! So much simpler :)

CriesofCarrots (approved):
Thanks for all the care taken on this. Lgtm!

ryoqun:
finally! I've finished to test this locally extensively with mainnet-beta snapshot. Merging this shortly.

Lichtso:
Can we remove the function `Bank::reconfigure_token2_native_mint()` now?

mvines:
> Can we remove the function `Bank::reconfigure_token2_native_mint()` now?

Did you mean to necropost here? Short answer is maybe! Worth exploring

Lichtso:
Yes, I meant to post. Just saw the function added in this PR still being around as I am working on the feature set transition for the `LoadedPrograms` cache.

ryoqun:
> Yes, I meant to post. Just saw the function added in this PR still being around as I am working on the feature set transition for the `LoadedPrograms` cache.

@Lichtso hmm? i think `reconfigure_token2_native_mint` is added by a different pr: https://github.com/solana-labs/solana/pull/11966#discussion_r1206137615, not this pr..

anyway, i also think it's okay to remove that function. (thanks for feature cleanup prs by the way)

PR #13463, opened 2020-11-07
Add back shredding broadcast stats

sakridge:
#### Problem

Lots of shredding time a bit of a mystery since stats are not implemented in a scalable way and thus disabled.

#### Summary of Changes

Accumulate serialize/gen-data/gen-coding stats for the slot and report them with the other broadcast stats.

Fixes #

### DIFF
--- a/ledger/src/shred.rs
+++ b/ledger/src/shred.rs
@@ -11,7 +11,7 @@ use rayon::{
     ThreadPool,
 };
 use serde::{Deserialize, Serialize};
-use solana_metrics::datapoint_debug;
+use solana_measure::measure::Measure;
 use solana_perf::packet::Packet;
 use solana_rayon_threadlimit::get_thread_count;
 use solana_sdk::{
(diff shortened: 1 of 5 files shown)


carllin (approved):

ledger/src/shred.rs
```diff
@@ -21,10 +21,38 @@ use solana_sdk::{
...
+}
+impl ProcessShredsStats {
+    pub fn update(&mut self, new_stats: &ProcessShredsStats) {
+        self.shredding_elapsed += new_stats.shredding_elapsed;
+        self.receive_elapsed += new_stats.receive_elapsed;
+        self.serialize_elapsed += new_stats.serialize_elapsed;
+        self.gen_data_elapsed += new_stats.gen_data_elapsed;
+        self.gen_coding_elapsed += new_stats.gen_coding_elapsed;
+        self.sign_coding_elapsed += new_stats.sign_coding_elapsed;
+        self.coding_send_elapsed += new_stats.gen_coding_elapsed;
```
behzadnouri, line 48: Looks like this is by mistake adding `gen_coding_elapsed` to `coding_send_elapsed`.
Putting a fix now.

PR #13464, opened 2020-11-07
Prevent scans from seeing root updates/clean

carllin:
#### Problem
Setting new roots during a scan can cause:

1)  Root updates add new entries to the accounts index, causing mixed results from old/newer roots
2) Cleanup to delete entries in the accounts index for old roots in the middle of a scan, causing some pubkeys to disappear from the scan

#### Summary of Changes
1) Track the max root at the beginning of the scan and only lookup roots < `max_root` so new roots don't get observed in the middle of the scan
2) Prevent cleanup from cleaning based on any updates to `roots` > `max_root`. This means for the following fork state:

`
0 (root) -> 1 (root) -> 2 (latest rooted bank at beginning of scan)
`

Then the following statement is guaranteed on scans of rooted  banks `0, 1, or 2`: 
`All non-zero lamport scanned pubkeys `P` should reflect the account state for that pubkey as of bank 2`

This is because for a pubkey that was observable from bank `2` to not be found by the scan, clean would have to detect an update to that key on a rooted slot > `2`, but we prevent clean on roots `> 2`.

It is indeterministic  whether zero-lamport accounts in `0, 1, 2` will show up in the scan though (depending on if a clean happens or not). I think this may be fine?

Edge cases and inconsistencies for scans on `non-rooted` banks will be handled in a follow-up PR

@CriesofCarrots fyi this series of PR's will address the scan inconsistencies we discussed 🙏 

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -1297,6 +1297,22 @@ impl AccountsDB {
         collector
     }
 
+    pub fn unchecked_scan_accounts<F, A>(&self, ancestors: &Ancestors, scan_func: F) -> A
+    where
+        F: Fn(&mut A, Option<(&Pubkey, Account, Slot)>),
+        A: Default,
+    {
+        let mut collector = A::default();
+        self.accounts_index
+            .unchecked_scan_accounts(ancestors, |pubkey, (account_info, slot)| {
+                let account_slot = self
+                    .get_account_from_storage(slot, account_info)
+                    .map(|account| (pubkey, account, slot));
+                scan_func(&mut collector, account_slot)
+            });
+        collector
+    }
+
     pub fn range_scan_accounts<F, A, R>(&self, ancestors: &Ancestors, range: R, scan_func: F) -> A
     where
         F: Fn(&mut A, Option<(&Pubkey, Account, Slot)>),
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -744,6 +744,14 @@ impl AccountsDB {
         // hold a lock to prevent slot shrinking from running because it might modify some rooted
         // slot storages which can not happen as long as we're cleaning accounts because we're also
         // modifying the rooted slot storages!
+        let max_clean_root = match (self.accounts_index.min_ongoing_scan_root(), max_clean_root) {
+            (None, None) => None,
+            (Some(min_scan_root), None) => Some(min_scan_root),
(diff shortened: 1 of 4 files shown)


runtime/src/accounts_index.rs
```diff
@@ -319,23 +384,27 @@ impl<T: 'static + Clone> AccountsIndex<T> {
         (reclaims, slot_list.is_empty())
     }
 
+    pub fn min_ongoing_scan_root(&self) -> Option<Slot> {
+        self.ongoing_scan_roots
+            .read()
+            .unwrap()
+            .keys()
+            .cloned()
+            .next()
```
sakridge, line 393: Can it be just `.keys().next().clone()` ? You only need the first key right? not all keys cloned.

carllin: yeah totally, fixed thanks!

runtime/src/accounts_db.rs
```diff
@@ -744,6 +744,14 @@ impl AccountsDB {
...
         // slot storages which can not happen as long as we're cleaning accounts because we're also
         // modifying the rooted slot storages!
+        let max_clean_root = match (self.accounts_index.min_ongoing_scan_root(), max_clean_root) {
+            (None, None) => None,
+            (Some(min_scan_root), None) => Some(min_scan_root),
+            (None, Some(max_clean_root)) => Some(max_clean_root),
+            (Some(min_scan_root), Some(max_clean_root)) => {
+                Some(std::cmp::min(min_scan_root, max_clean_root))
+            }
+        };
```
carllin, line 754: Commenting out this block should cause `test_store_scan_consistency` to fail

runtime/src/accounts_index.rs
```diff
@@ -194,15 +196,70 @@ impl<T: 'static + Clone> AccountsIndex<T> {
...
+            // `max_root()` grabs a lock while
+            // the `ongoing_scan_roots` lock is held,
+            // make sure inverse doesn't happen to avoid
+            // deadlock
+            let max_root = self.max_root();
+            *w_ongoing_scan_roots.entry(max_root).or_default() += 1;
+            max_root
+        };
+
+        self.do_scan_accounts(ancestors, func, range, Some(max_root));
```
carllin, line 226: Changing `Some(max_root) -> None`  should cause `test_store_scan_consistency` to fail

sakridge (approved):
lgtm

PR #13465, opened 2020-11-07
Fix avx check with newest nightly compiler

sakridge:
#### Problem

Compiler error with nightly about using target_feature on non-function.

#### Summary of Changes

Move target_feature code to a function.

Fixes #


t-nelson (approved):
LGTM!

PR #13477, opened 2020-11-09
Fix Bank accounts hash mismatch related to Clock::unix_timestamp

CriesofCarrots:
#### Problem
The `timestamp_bounding` feature broke testnet, as `Bank::ancestors` is not deterministic across nodes.
This filtering was intended to minimize the effect of delinquent vote accounts; however, since delinquent accounts shouldn't reflect much stake, and since timestamps are being bounded by PoH, this no longer seems necessary.

#### Summary of Changes
- Remove non-deterministic filtering
- Update docs
- Update feature key to re-enable once testnet is back up

### DIFF
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -1545,10 +1545,9 @@ impl Bank {
             .filter_map(|(pubkey, (_, account))| {
                 VoteState::from(&account).and_then(|state| {
                     let timestamp_slot = state.last_timestamp.slot;
-                    if (self
+                    if self
                         .feature_set
                         .is_active(&feature_set::timestamp_bounding::id())
-                        && self.ancestors.contains_key(&timestamp_slot))
                         || self.slot().checked_sub(timestamp_slot)?
                             <= DEPRECATED_TIMESTAMP_SLOT_RANGE as u64
                     {
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -1545,9 +1545,11 @@ impl Bank {
             .filter_map(|(pubkey, (_, account))| {
                 VoteState::from(&account).and_then(|state| {
                     let timestamp_slot = state.last_timestamp.slot;
-                    if self
+                    if (self
                         .feature_set
                         .is_active(&feature_set::timestamp_bounding::id())
+                        && self.slot().checked_sub(timestamp_slot)?
+                            <= self.epoch_schedule().slots_per_epoch)
                         || self.slot().checked_sub(timestamp_slot)?
                             <= DEPRECATED_TIMESTAMP_SLOT_RANGE as u64
                     {
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -4258,6 +4258,17 @@ pub fn goto_end_of_slot(bank: &mut Bank) {
     }
 }
 
+use solana_vote_program::vote_state::BlockTimestamp;
+use solana_vote_program::vote_state::VoteStateVersions;
+pub fn update_vote_account_timestamp(timestamp: BlockTimestamp, bank: &Bank, vote_pubkey: &Pubkey) {
+    let mut vote_account = bank.get_account(vote_pubkey).unwrap_or_default();
+    let mut vote_state = VoteState::from(&vote_account).unwrap_or_default();
+    vote_state.last_timestamp = timestamp;
+    let versioned = VoteStateVersions::Current(Box::new(vote_state));
+    VoteState::to(&versioned, &mut vote_account).unwrap();
+    bank.store_account(vote_pubkey, &vote_account);
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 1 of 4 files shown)


sakridge (approved):
lgtm

runtime/src/bank.rs
```diff
@@ -1548,7 +1548,8 @@ impl Bank {
                     if (self
                         .feature_set
                         .is_active(&feature_set::timestamp_bounding::id())
-                        && self.ancestors.contains_key(&timestamp_slot))
+                        && self.slot().checked_sub(timestamp_slot)?
+                            <= self.epoch_schedule().slots_per_epoch)
```
CriesofCarrots, line 1552: @aeyakovenko , I think that the vote accounts with stake beyond 99.5% will get swept up in the stake-weighted median calculation, so I've limited this to just filtering out anything older than an epoch.

runtime/src/bank.rs
```diff
@@ -4258,6 +4259,17 @@ pub fn goto_end_of_slot(bank: &mut Bank) {
     }
 }
 
+use solana_vote_program::vote_state::BlockTimestamp;
+use solana_vote_program::vote_state::VoteStateVersions;
+pub fn update_vote_account_timestamp(timestamp: BlockTimestamp, bank: &Bank, vote_pubkey: &Pubkey) {
```
carllin, line 4264: i think we can move this into the test module below (make the module public for the crate), and then from  tests in `bank_forks` that need the import, import `bank::tests::update_vote_account_timestamp,`

runtime/src/bank_forks.rs
```diff
@@ -378,4 +390,74 @@ mod tests {
...
+
+        let additional_timestamp_secs = 2;
+
+        let num_slots = slots_in_epoch + DEPRECATED_TIMESTAMP_SLOT_RANGE as u64 + 1;
+        for slot in 1..num_slots {
+            // Just after the epoch boundary, timestamp a vote that will shift
+            // Clock::unix_timestamp from Bank::unix_timestamp_from_genesis()
+            let update_timestamp_case = slot == slots_in_epoch;
+
+            let child1 = Bank::new_from_parent(&bank_forks0[slot - 1], &Pubkey::default(), slot);
```
carllin, line 421: nit: How about avoiding the code duplication for the `update_vote_account_timestamp` logic with:

```
let child1 = Bank::new_from_parent(&bank_forks0[slot - 1], &Pubkey::default(), slot);
            let child2 = Bank::new_from_parent(&bank_forks1[slot - 1], &Pubkey::default(), slot);
            if update_timestamp_case {
                for child in &[&child1, &child2] {
                    let recent_timestamp: UnixTimestamp = child.unix_timestamp_from_genesis();
                    update_vote_account_timestamp(
                        BlockTimestamp {
                            slot: child.slot(),
                            timestamp: recent_timestamp + additional_timestamp_secs,
                        },
                        child,
                        &voting_keypair.pubkey(),
                    );
                }
            }

            // Set root in bank_forks0 to truncate the ancestors history
            bank_forks0.insert(child1);
            bank_forks0.set_root(slot, &None, None);

            // Don't set root in bank_forks1 to keep the ancestors history
            bank_forks1.insert(child2);

```

runtime/src/bank_forks.rs
```diff
@@ -378,4 +390,74 @@ mod tests {
...
+        } = create_genesis_config_with_leader(10_000, &leader_keypair.pubkey(), 1_000);
+        let slots_in_epoch = 32;
+        genesis_config.epoch_schedule = EpochSchedule::new(slots_in_epoch);
+
+        let bank0 = Bank::new(&genesis_config);
+        let mut bank_forks0 = BankForks::new(bank0);
+
+        let bank1 = Bank::new(&genesis_config);
+        let mut bank_forks1 = BankForks::new(bank1);
+        bank_forks0.set_root(0, &None, None);
```
carllin, line 411: nit: move `set_root` after `let mut bank_forks0 = BankForks::new(bank0);`  be closer to the declaration site

runtime/src/bank_forks.rs
```diff
@@ -378,4 +390,74 @@ mod tests {
...
+                );
+            }
+            bank_forks1.insert(child2);
+        }
+        let child1 =
+            Bank::new_from_parent(&bank_forks0[num_slots - 1], &Pubkey::default(), num_slots);
+        let child2 =
+            Bank::new_from_parent(&bank_forks1[num_slots - 1], &Pubkey::default(), num_slots);
+        bank_forks1.set_root(num_slots - 1, &None, None);
+
```
carllin, line 455: instead of these new children, can we just reuse the last children that were constructed in the loop?

carllin (dismissed):
Thanks for the quick turnaround!

runtime/src/bank_forks.rs
CriesofCarrots: Done https://github.com/solana-labs/solana/pull/13477/commits/1721cb69af4c9f50fa98d638a7a9a15883914395

runtime/src/bank_forks.rs
CriesofCarrots: Done https://github.com/solana-labs/solana/pull/13477/commits/1721cb69af4c9f50fa98d638a7a9a15883914395

runtime/src/bank_forks.rs
CriesofCarrots: Much nicer! https://github.com/solana-labs/solana/pull/13477/commits/1721cb69af4c9f50fa98d638a7a9a15883914395

runtime/src/bank.rs
CriesofCarrots: https://github.com/solana-labs/solana/pull/13477/commits/1721cb69af4c9f50fa98d638a7a9a15883914395

runtime/src/bank.rs
aeyakovenko: good enough for now

aeyakovenko (approved):

PR #13479, opened 2020-11-09
Test for different ancestors with mismatch bank hash

sakridge:
#### Problem

No test for bank hash for when ancestor list is different from different squashing paths.

#### Summary of Changes

Test bank hash with two banks with different squash sequences.

Fixes #

### DIFF
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -9803,6 +9803,48 @@ mod tests {
     }
 
     #[test]
+    fn test_bank_hash_internal_state_different_squash() {
+        solana_logger::setup();
+        let (genesis_config, mint_keypair) = create_genesis_config(2_000);
+        let bank0 = Bank::new(&genesis_config);
+        let bank1 = Bank::new(&genesis_config);
+        let initial_state = bank0.hash_internal_state();
+        assert_eq!(bank1.hash_internal_state(), initial_state);
+
+        let pubkey = solana_sdk::pubkey::new_rand();
+        bank0.transfer(1_000, &mint_keypair, &pubkey).unwrap();
+        assert_ne!(bank0.hash_internal_state(), initial_state);
+        bank1.transfer(1_000, &mint_keypair, &pubkey).unwrap();
+        assert_eq!(bank0.hash_internal_state(), bank1.hash_internal_state());
+
+        // Create children with different squashing
+        // 0 <- 2 (s) <- 4 (s)
+        // and
+        // 1 <- 3 <- 5 (s)
+        // ensure that 4.bank_hash() == 5.bank_hash()
+        let bank0 = Arc::new(bank0);
+        let bank2 = new_from_parent(&bank0);
+        bank2.squash();
+        let bank2 = Arc::new(bank2);
+        let bank4 = new_from_parent(&bank2);
+        bank4.squash();
+
+        info!("done with 2 & 4, starting 3 & 5");
+
+        let bank1 = Arc::new(bank1);
+        let bank3 = new_from_parent(&bank1);
+        let bank3 = Arc::new(bank3);
+        let bank5 = new_from_parent(&bank3);
+        bank5.squash();
+        info!("bank5:*********************");
+        bank5.print_accounts_stats();
+        info!("bank4:*********************");
+        bank4.print_accounts_stats();
+        info!("done:*********************");
+        assert_eq!(bank5.hash_internal_state(), bank4.hash_internal_state());
+    }
+
+    #[test]
     fn test_get_timestamp_estimate() {
         let validator_vote_keypairs0 = ValidatorVoteKeypairs::new_rand();
         let validator_vote_keypairs1 = ValidatorVoteKeypairs::new_rand();
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -4202,6 +4202,17 @@ pub fn goto_end_of_slot(bank: &mut Bank) {
     }
 }
 
+use solana_vote_program::vote_state::BlockTimestamp;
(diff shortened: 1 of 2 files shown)


runtime/src/bank.rs
```diff
@@ -6962,6 +6962,35 @@ mod tests {
...
+        let pubkey = solana_sdk::pubkey::new_rand();
+        bank0.transfer(1_000, &mint_keypair, &pubkey).unwrap();
+        assert_ne!(bank0.hash_internal_state(), initial_state);
+        bank1.transfer(1_000, &mint_keypair, &pubkey).unwrap();
+        assert_eq!(bank0.hash_internal_state(), bank1.hash_internal_state());
+
+        // Create children with different squashing
+        // 0 <- 2 (s) <- 4 (s)
+        // and
+        // 1 <- 3 <- 5 (s)
```
carllin, line 6982: I think it would be more realistic to simulate squashes on the same fork (i.e. any two nodes who squash `A` and `B` should ensure they are ancestors/descendants of one another). We could set up a single fork and just squash at different points along that fork.

carllin: Maybe create the branch:
1) `0 -> 1 -> 2 ->3->4` (two copies of this fork)
2) One copy squashes at  each `0, 1, 2, 3` (right after each `new_from_parent`)
3) One copy only squashes at `3` (after the entire fork chain has been constructed)
4) Check the ancestors of `4` to make sure they're different, but the hash of `4` should be the same

sakridge: That's essentially what I did, just with 4 slots instead of 5 in your example.

sakridge: actually I think I need to do it with bank forks so that it prunes the older roots

CriesofCarrots: I think we'll also need to adjust the vote-account timestamp to hit the logic in question, and will probably need to move a few slots further away from the epoch boundary to not hit the PoH bounding condition.

I think if we add 10 banks before the different squashing behavior, something like this against bank 11 should work:
```
let recent_timestamp: UnixTimestamp = bank.unix_timestamp_from_genesis();
update_vote_account_timestamp(
    BlockTimestamp {
        slot: bank.slot(),
        timestamp: recent_timestamp + 1,
    },
    &bank,
    &validator_vote_keypairs.vote_keypair.pubkey(),
);
```

I think I can work this up and add to my fix PR. Thanks for showing the way, @sakridge and @carllin !

sakridge: @CriesofCarrots that's tricky, I might need your help here:
https://github.com/solana-labs/solana/pull/13479/files#diff-a0e99746f30ade88937b90a9e96a93e3e77c7eb6d7ff6683b3fdaa46cebcf718R411

CriesofCarrots: It is. I realized the forks need to be past the first epoch boundary, and the vote timestamp needs to be sufficiently in the past that it doesn't hit the feature-not-activated case.

Here's the version I came up with: https://github.com/solana-labs/solana/pull/13477/commits/4c9d84564f3f28be6d5a5dcf4ee1d383f335d616
It fails consistently on master, and seems to succeed consistently with the fix, but I'm still nervous about the fact that the test was flaky on v1.3 without the timestamp stuff.

CriesofCarrots: @sakridge Does the use of BankForks account for that flakiness?

sakridge: ok; great. Yes, the BankForks version seems to pass consistently for me. I agree that it is a bit worrying that it can fail.

CriesofCarrots: Great, I haven't been able to make it fail against v1.3 either.
I'll rebase on your BankForks version

sakridge: Looks like rent is collected in one case but not the other:

```
thread 'bank::tests::test_bank_hash_internal_state_different_squash' panicked at 'where?', runtime/src/accounts_db.rs:2613:17
stack backtrace:
   0: std::panicking::begin_panic
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/std/src/panicking.rs:497
   1: solana_runtime::accounts_db::AccountsDB::store
             at ./src/accounts_db.rs:2613
   2: solana_runtime::accounts::Accounts::store_slow
             at ./src/accounts.rs:560
   3: solana_runtime::bank::Bank::store_account
             at ./src/bank.rs:3356
   4: solana_runtime::bank::Bank::collect_rent_in_partition
             at ./src/bank.rs:2933
   5: solana_runtime::bank::Bank::collect_rent_eagerly
             at ./src/bank.rs:2880
   6: solana_runtime::bank::Bank::freeze
             at ./src/bank.rs:1586
   7: solana_runtime::bank::Bank::squash
             at ./src/bank.rs:1627
   8: solana_runtime::bank::tests::test_bank_hash_internal_state_different_squash
             at ./src/bank.rs:9826
   9: solana_runtime::bank::tests::test_bank_hash_internal_state_different_squash::{{closure}}
             at ./src/bank.rs:9806
  10: core::ops::function::FnOnce::call_once
```

@ryoqun

CriesofCarrots: @sakridge, here is the rebased test: https://github.com/solana-labs/solana/pull/13477/commits/fff4e9b3dc53b13551fa60b1768a6697d17a2dae

sakridge: I see, the problem is because slot 0 bank is `drop`'ed and thus cleaned up. Then when the rent collection comes, when it does a scan, it doesn't find the slot 0 account. The rent is only collected when the pubkey is in a certain range and it's random, that seems to cause the mismatch.

sakridge:
added in another PR

PR #13491, opened 2020-11-10
Fix signature access

sakridge:
#### Problem

Access on transaction which may have no signatures.

#### Summary of Changes

Fix access

Fixes #


carllin (dismissed):

core/src/completed_data_sets_service.rs
```diff
@@ -76,7 +74,51 @@ impl CompletedDataSetsService {
         Ok(())
     }
 
+    fn get_transaction_signatures(entries: Vec<Entry>) -> Vec<Signature> {
+        entries
+            .into_iter()
+            .flat_map(|e| {
+                e.transactions
+                    .into_iter()
+                    .filter_map(|t| t.signatures.get(0).cloned())
```
carllin, line 83: Maybe we can avoid the clone with `.filter_map(|mut t| t.signatures.drain(..).next())`

PR #10902, opened 2020-07-03
Rpc: Add getStakeActivation endpoint

CriesofCarrots:
#### Problem
`solana stake-account` returns lots of useful information about stake activation/deactivation for the current epoch. But it's pretty hard to calculate this information if you're not using the solana cli.

#### Summary of Changes
- Add `getStakeActivation` endpoint; it takes a stake account address (and optional past epoch and optional commitment) as input parameters and returns effective, activating, and deactivating stake amounts

Fixes #10885

PR #13497, opened 2020-11-10
cli: stake-account/stakes: Use GetStakeActivation

ryoqun:
#### Problem

CLI calculates stakes locally, which is very hard to properly feature-gate it. And it's not useful for fixing the stake calculation code.

#### Summary of Changes

Use aptly-provided `getStakeActivation` rpc endpoint.

based-on: #10902


cli/src/stake.rs
```diff
@@ -1512,9 +1513,17 @@ pub fn build_stake_state(
             stake,
         ) => {
             let current_epoch = clock.epoch;
-            let (active_stake, activating_stake, deactivating_stake) = stake
-                .delegation
-                .stake_activating_and_deactivating(current_epoch, Some(stake_history));
```
ryoqun, line 1517: :wave:

cli/src/stake.rs
```diff
@@ -1512,9 +1513,17 @@ pub fn build_stake_state(
...
+            let (active_stake, activating_stake, deactivating_stake) = match stake_activation.state
+            {
+                StakeActivationState::Active => (stake_activation.active, 0, 0),
+                StakeActivationState::Inactive => (0, 0, 0),
+                StakeActivationState::Activating => {
+                    (stake_activation.active, stake_activation.inactive, 0)
+                }
+                StakeActivationState::Deactivating => {
+                    (stake_activation.active, 0, stake_activation.inactive)
+                }
```
ryoqun, line 1525: shameless inverse conversion from RpcStakeActivation back to the original stake tuple (`(u64, u64, u64)`). If I had more time, I would fix CliStakeState itself... ;)

client/src/rpc_client.rs
```diff
@@ -474,6 +474,13 @@ impl RpcClient {
         self.get_epoch_info_with_commitment(CommitmentConfig::default())
     }
 
+    pub fn get_stake_activation(&self, pubkey: &Pubkey) -> ClientResult<RpcStakeActivation> {
```
ryoqun, line 477: I think adding new `fn`s is ok in terms of compatibility even with a patch release.

ryoqun:
@CriesofCarrots Could you quick review this? Thanks!

cli/src/cluster_query.rs
```diff
@@ -1339,20 +1335,16 @@ pub fn process_show_stakes(
...
     progress_bar.finish_and_clear();
 
-    let stake_history = from_account(&stake_history_account).ok_or_else(|| {
-        CliError::RpcRequestError("Failed to deserialize stake history".to_string())
-    })?;
-
     let mut stake_accounts: Vec<CliKeyedStakeState> = vec![];
     for (stake_pubkey, stake_account) in all_stake_accounts {
         if let Ok(stake_state) = stake_account.state() {
+            let stake_activation = rpc_client.get_stake_activation(&stake_pubkey)?;
```
CriesofCarrots, line 1347: Unfortunately, this doesn't quite work. If a stake account is undelegated, `getStakeActivation` returns a JSON-rpc error: `{"jsonrpc":"2.0","error":{"code":-32602,"message":"Invalid param: stake account has not been delegated"},"id":1}`
...which errors out here and breaks the loop (or in the case of `solana stake-account`, prevents the account info being shown).

A simple (naive) fix might be to write this as:
```suggestion
            let stake_activation = rpc_client.get_stake_activation(&stake_pubkey).unwrap_or(
              RpcStakeActivation {
                state: StakeActivationState::Inactive,
                active: 0,
                inactive: 0,
              }
            );
```
although that would disguise other types of errors.

CriesofCarrots:
This change makes `solana stakes` prohibitively expensive: ~3min on mainnet-beta currently
Trying a different approach!

ryoqun:
ref: similar change was incorporated into #13501

PR #13500, opened 2020-11-10
packs more crds-values in a single gossip packet

behzadnouri:
#### Problem
split_gossip_messages:
https://github.com/solana-labs/solana/blob/a97c04b40/core/src/cluster_info.rs#L1536-L1574
splits crds-values into chunks to fit into a gossip packet. However it is
using a global upper-bound for the header-size across all protocols:
https://github.com/solana-labs/solana/blob/a97c04b40/core/src/cluster_info.rs#L90-L93
This can be wasteful as the specific gossip protocol can have smaller
header than this upper-bound (e.g. Protocol::PushMessage is 170 bytes
smaller). Adding more crds-values in one gossip packet can avoid the
overheads of separate packets and reduce total number of bytes sent over
the wire.

#### Summary of Changes
This commit updates the splitting function to take a max-chunk-size
argument. At call-site, this value is set to the size of the protocol
which the values are sent over.


behzadnouri:
cluster_info_stats new_pash_requests_num_split_packets
On a gce cluster this commit decreases `new_push_requests_num` by ~5% (left chunk is this commit). Processing times also show a slight improvement.

core/src/cluster_info.rs
```diff
@@ -87,22 +88,26 @@ pub const GOSSIP_SLEEP_MILLIS: u64 = 100;
...
+// TODO: Update this to 26 once payload sizes are upgraded across fleet.
 pub const MAX_SNAPSHOT_HASHES: usize = 16;
 /// Number of bytes in the randomly generated token sent with ping messages.
 const GOSSIP_PING_TOKEN_SIZE: usize = 32;
 const GOSSIP_PING_CACHE_CAPACITY: usize = 16384;
 const GOSSIP_PING_CACHE_TTL: Duration = Duration::from_secs(640);
 
+lazy_static::lazy_static! {
+    /// Maximum serialized sized of crds-values in a Protocol::PushMessage.
+    static ref PUSH_MESSAGE_MAX_PAYLOAD_SIZE: u64 = PACKET_DATA_SIZE as u64
```
sakridge, line 107: We've seen some bad performance with lazy_static in the past, best to do just a fixed constant and then verify that with a test.

behzadnouri: good idea, done.

core/src/cluster_info.rs
```diff
@@ -1533,44 +1536,54 @@ impl ClusterInfo {
...
+        max_chunk_size: usize,
+        data_feed: I,
+    ) -> impl Iterator<Item = Vec<T>>
+    where
+        T: Serialize + Debug,
+        I: IntoIterator<Item = T>,
+    {
+        let mut data_feed = data_feed.into_iter().fuse();
+        let mut buffer = vec![];
+        let mut buffer_size = 0;
```
carllin, line 1554: maybe something more explicit like `serialized_buffer_size` to communicate this is not necessarily equal to `buffer.len()`

core/src/cluster_info.rs
```diff
@@ -1533,44 +1536,54 @@ impl ClusterInfo {
...
-                        payload_size = msg_size;
+    fn split_gossip_messages<I, T>(
+        max_chunk_size: usize,
+        data_feed: I,
+    ) -> impl Iterator<Item = Vec<T>>
+    where
+        T: Serialize + Debug,
+        I: IntoIterator<Item = T>,
+    {
+        let mut data_feed = data_feed.into_iter().fuse();
```
carllin, line 1552: is the fuse here just a precaution? Do we care about a more stringent guarantee that there are no more items in the iterator after the first `None` (i.e. something like a flag once a None is observed, and an assert in the below loop?)

carllin (approved):

PR #13513, opened 2020-11-10
Fix parsing CreateAccountWithSeed instructions

### LINKED ISSUE #13354: CreateAccountWithSeed parsing fails
#### Problem
SystemInstruction::CreateAccountWithSeed instructions are not being parsed successfully:
https://explorer.solana.com/tx/4aYb68NDK5bDwLufW9TUKawKS8Yfp59pzXRKVhXkq8F3LMXrMbG3wetR6a9k32oudE4U4KET6rMH5uykg9mv1Mbn

#### Proposed Solution
Figure out why and fix

CriesofCarrots:
#### Problem
Based on the CreateAccountWithSeed documentation, the system instruction parser expected those instructions to always load 3 accounts. However, these instructions can succeed with only two, eg https://explorer.solana.com/tx/4aYb68NDK5bDwLufW9TUKawKS8Yfp59pzXRKVhXkq8F3LMXrMbG3wetR6a9k32oudE4U4KET6rMH5uykg9mv1Mbn

This is because the base pubkey is provided in the instruction data, and is only required to be a signer in the accounts list, not a particular indexed account in the list. In other words, the funding key may serve as the base key of the derived address without being passed in as a 3rd account index, since it is already a signer.

#### Summary of Changes
- Fix system instruction parser to properly handle 2- and 3-account cases
- Update CreateAccountWithSeed documentation to indicate that the 3rd account is optional if base account == funding account
- Add CreateAccountWithSeed test so that both 2- and 3-account cases are illustrated

Fixes #13354

### DIFF
--- a/runtime/src/system_instruction_processor.rs
+++ b/runtime/src/system_instruction_processor.rs
@@ -480,6 +480,43 @@ mod tests {
     }
 
     #[test]
+    fn test_create_account_with_seed_separate_base_account() {
+        let new_owner = Pubkey::new(&[9; 32]);
+        let from = solana_sdk::pubkey::new_rand();
+        let base = solana_sdk::pubkey::new_rand();
+        let seed = "shiny pepper";
+        let to = Pubkey::create_with_seed(&base, seed, &new_owner).unwrap();
+
+        let from_account = Account::new_ref(100, 0, &system_program::id());
+        let to_account = Account::new_ref(0, 0, &Pubkey::default());
+        let base_account = Account::new_ref(0, 0, &Pubkey::default());
+
+        assert_eq!(
+            process_instruction(
+                &Pubkey::default(),
+                &[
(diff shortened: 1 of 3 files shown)


CriesofCarrots:
@mvines , can you give https://github.com/solana-labs/solana/pull/13513/commits/dde761810f2e509fa422456843b7084b86c8bb9f a quick peek to see if I've described the functionality in a way that makes sense?

mvines:
> @mvines , can you give [dde7618](https://github.com/solana-labs/solana/commit/dde761810f2e509fa422456843b7084b86c8bb9f) a quick peek to see if I've described the functionality in a way that makes sense?

Makes sense.  Took me a bit going through the system program to see how that actually worked :)

CriesofCarrots:
>  Took me a bit going through the system program to see how that actually worked :)

You and me both!

PR #13527, opened 2020-11-11
Discard pre hard fork persisted tower if hard-forking

ryoqun:
#### Problem

The tds relaunch failed. The theory is like this:

Substantial validator (~20-30%, rough guess from chronograf) has rooted newer slots than the hard fork slot (46765918) and kept the ledger. In that case, validator cannot vote because restored tower isn't adjusted for the hard fork and it prevents voting at all because of switch threshold failure. Finally this  makes liveness under supermajority.
voting node log sample:

```
[2020-11-11T08:02:10.893472532Z INFO  solana_core::validator] Waiting for 80% of activated stake at slot 46765918 to be in gossip...
[2020-11-11T08:02:10.781175975Z ERROR solana_core::consensus] For some reason, we're REPROCESSING slots which has already been voted and ROOTED by us; VOTING will be SUSPENDED UNTIL 46766871!
[2020-11-11T13:09:32.455506994Z INFO  solana_core::replay_stage] Waiting to switch vote to 46766495, resetting to slot None on same fork for now
[2020-11-11T13:09:33.192110752Z INFO  solana_core::replay_stage] Waiting to switch vote to 46766980, 
resetting to slot None on same fork for now
[2020-11-11T13:09:33.192139724Z INFO  solana_core::replay_stage] Couldn't vote on heaviest fork: 46766980, heaviest_fork_failures: [FailedSwitchThreshold(46766980)]
$ TZ=UTC ls -l ledger/tower-twP716sMjTFynrZRSCKkHhoeHaQEXnR2bzqDEzuxP81.bin 
-rw-r--r-- 1 sol sol 2360 Nov  9 04:17 ledger/tower-twP716sMjTFynrZRSCKkHhoeHaQEXnR2bzqDEzuxP81.bin # tower not saved after the restart at all.
```

#### Summary of Changes

Discard old persited tower if hard-forking.

- [x] Think more to ensure to avoid false positive/false negative...
- [x] Write a test.

Fixes #

### DIFF
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -1182,6 +1182,9 @@ pub enum TowerError {
 
     #[error("The tower is fatally inconsistent with blockstore: {0}")]
     FatallyInconsistent(&'static str),
+
+    #[error("The tower is useless because of new hard fork: {0}")]
+    HardFork(Slot),
 }
 
 impl TowerError {
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -646,7 +646,9 @@ impl Tower {
                     // By the time we reach here, any ancestors of the `last_vote`,
                     // should have been filtered out, as they all have a descendant,
                     // namely the `last_vote` itself.
-                    assert!(!last_vote_ancestors.contains(candidate_slot));
+                    if !self.is_stray_last_vote() {
+                        assert!(!last_vote_ancestors.contains(candidate_slot));
+                    }
 
                     // Evaluate which vote accounts in the bank are locked out
                     // in the interval candidate_slot..last_vote, which means
--- a/core/src/validator.rs
+++ b/core/src/validator.rs
@@ -723,6 +723,17 @@ fn post_process_restored_tower(
     restored_tower
         .and_then(|tower| {
             let root_bank = bank_forks.root_bank();
+            if let Some(wait_slot_for_supermajority) = config.wait_for_supermajority {
+                if root_bank.slot() == wait_slot_for_supermajority { // <= is this check too strict?
+                    // intentionally fail to restore tower; we're supposedly in a new hard fork; past
+                    // out-of-chain vote state doesn't make sense at all
+                    // what if --wait-for-supermajority again if the validator restarted?
+                    warn!("bla bla");
+                    //datapoint_warn!(...);
+                    return Err(crate::consensus::TowerError::HardFork(wait_slot_for_supermajority));
+                }
+            }
+
             let slot_history = root_bank.get_slot_history();
             tower.adjust_lockouts_after_replay(root_bank.slot(), &slot_history)
         })
--- a/core/src/validator.rs
+++ b/core/src/validator.rs
@@ -725,13 +725,24 @@ fn post_process_restored_tower(
     restored_tower
         .and_then(|tower| {
             let root_bank = bank_forks.root_bank();
+            let slot_history = root_bank.get_slot_history();
+            let tower = tower.adjust_lockouts_after_replay(root_bank.slot(), &slot_history);
+
             if let Some(wait_slot_for_supermajority) = config.wait_for_supermajority {
-                if root_bank.slot() == wait_slot_for_supermajority { // <= is this check too strict?
+                if root_bank.slot() == wait_slot_for_supermajority {
                     // intentionally fail to restore tower; we're supposedly in a new hard fork; past
                     // out-of-chain vote state doesn't make sense at all
                     // what if --wait-for-supermajority again if the validator restarted?
-                    warn!("bla bla");
-                    //datapoint_warn!(...);
+                    let message = format!("Hardfork is detected; discarding tower restoration result: {:?}", tower);
+                    datapoint_error!(
+                        "tower_error",
+                        (
+                            "error",
+                            message,
+                            String
+                        ),
+                    );
+                    error!("{}", message);
 
                     // unconditionally relax tower requirement so that we can always restore tower
                     // from root bank.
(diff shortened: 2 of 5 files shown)


ryoqun:
@carllin Could you review this? :)

core/src/validator.rs
```diff
@@ -720,9 +720,26 @@ fn post_process_restored_tower(
...
 ) -> Tower {
+    let mut should_require_tower = config.require_tower;
+
     restored_tower
         .and_then(|tower| {
             let root_bank = bank_forks.root_bank();
+            if let Some(wait_slot_for_supermajority) = config.wait_for_supermajority {
+                if root_bank.slot() == wait_slot_for_supermajority { // <= is this check too strict?
+                    // intentionally fail to restore tower; we're supposedly in a new hard fork; past
+                    // out-of-chain vote state doesn't make sense at all
```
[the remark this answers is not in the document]
sakridge: This is probably fine for now, but how about storing the shred version in the persistent vote file? I think that might be more robust, we can just discard it if the shred version has a mismatch.

core/src/validator.rs
```diff
@@ -720,9 +720,26 @@ fn post_process_restored_tower(
...
+                    // what if --wait-for-supermajority again if the validator restarted?
+                    warn!("bla bla");
```
sakridge, line 733: remove?

carllin:
Should the persisted tower be deleted in the hard fork case? I guess it's overwritten on the first vote? I'm imagining an edge case where somebody starts up with `wait_for_supermajority`, then shuts down before the validator votes, then starts up again without `wait_for_supermajority`

core/src/validator.rs
ryoqun: well added proper message.

core/src/validator.rs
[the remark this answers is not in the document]
ryoqun: yeah. makes sense.

local-cluster/src/cluster.rs
```diff
@@ -40,5 +40,15 @@ pub trait Cluster {
...
+    fn create_restart_context(
+        &mut self,
+        pubkey: &Pubkey,
+        cluster_validator_info: &mut ClusterValidatorInfo,
+    ) -> (solana_core::cluster_info::Node, Option<ContactInfo>);
+    fn restart_node_with_context(
+        cluster_validator_info: ClusterValidatorInfo,
+        restart_context: (solana_core::cluster_info::Node, Option<ContactInfo>),
+    ) -> ClusterValidatorInfo;
+    fn add_restarted_node(&mut self, pubkey: &Pubkey, cluster_validator_info: ClusterValidatorInfo);
```
ryoqun, line 52: some ugly api is needed to work around blocking `Validaotor::new`. ;)

local-cluster/tests/local_cluster.rs
```diff
@@ -1984,6 +1984,121 @@ fn test_future_tower_master_slave() {
...
+    };
+    let cluster = std::sync::Arc::new(std::sync::Mutex::new(LocalCluster::new(&config)));
+
+    let val_a_ledger_path = cluster.lock().unwrap().ledger_path(&validator_a_pubkey);
+
+    loop {
+        sleep(Duration::from_millis(100));
+
+        if let Some(root) = root_in_tower(&val_a_ledger_path, &validator_a_pubkey) {
+            if root >= 15 {
```
carllin, line 2029: Let's call this root `min_root`, and then the `hard_fork_slot` below can be `min_root - 5` to indicate the purpose is to be less than `min_root`

local-cluster/tests/local_cluster.rs
```diff
@@ -1984,6 +1984,122 @@ fn test_future_tower_master_slave() {
     do_test_future_tower(ClusterMode::MasterSlave);
 }
 
+#[test]
+#[serial]
+fn test_hard_fork() {
```
ryoqun, line 1989: there was no test or whatever for hard fork code.. hence this generic test name can be justified.

core/src/consensus.rs
```diff
@@ -646,7 +646,9 @@ impl Tower {
                     // By the time we reach here, any ancestors of the `last_vote`,
                     // should have been filtered out, as they all have a descendant,
                     // namely the `last_vote` itself.
-                    assert!(!last_vote_ancestors.contains(candidate_slot));
+                    if !self.is_stray_last_vote() {
```
carllin, line 649: Investigating why this is necessary, hopefully just a test issue 🙏

core/src/validator.rs
```diff
@@ -720,11 +720,38 @@ fn post_process_restored_tower(
...
             let slot_history = root_bank.get_slot_history();
-            tower.adjust_lockouts_after_replay(root_bank.slot(), &slot_history)
+            let tower = tower.adjust_lockouts_after_replay(root_bank.slot(), &slot_history);
+
+            if let Some(wait_slot_for_supermajority) = config.wait_for_supermajority {
+                if root_bank.slot() == wait_slot_for_supermajority {
+                    // intentionally fail to restore tower; we're supposedly in a new hard fork; past
+                    // out-of-chain vote state doesn't make sense at all
+                    // what if --wait-for-supermajority again if the validator restarted?
+                    let message = format!("Hardfork is detected; discarding tower restoration result: {:?}", tower);
```
ryoqun, line 736: oops! `Debug of tower `can be really big. it might not be wise to put in the datapoint_error!, only add `error!` below.

ryoqun: nits: include `wait_slot_for_supermajority` instead.

core/src/consensus.rs
carllin: @ryoqun, I commented this check out,  but haven't been able to repro the failure you described. The test seems to be passing fine on several different machines.

ryoqun:
superseded by #13536

PR #13537, opened 2020-11-11
Custom heap is BPF only

jackcmay:
#### Problem

Custom heaps only apply to BPF programs but its inclusion in `entrypoint!` means it also shows up when using `program-test` so programs must feature guard the use of `entrypoint!` from the outside

#### Summary of Changes

Add feature guard to the `entrypoint!` macro to avoid the conflict or doing this:
https://github.com/solana-labs/solana-program-library/blob/487ad2d2d7cb0b8a58f02a156a659e6c8a74f354/examples/rust/logging/src/entrypoint.rs#L3

Mentioned: https://github.com/solana-labs/solana-program-library/issues/829

Fixes #


joncinque:
This is an even better solution, thanks!

PR #13544, opened 2020-11-12
explorer: Add additional tokens and icons to registry

dr497:
#### Problem

Missing tokens to the explorer

#### Summary of Changes

Added missing tokens to the explorer including NFTs listed on [solible.com](https://solible.com)


oJshua:
Thanks @dr497 ! This looks pretty good. Could you run `npm run format:fix` and commit the results?

oJshua (changes requested):
Thanks @dr497 ! This looks pretty good. Could you run `npm run format:fix` and commit the results?

dr497:
> Thanks @dr497 ! This looks pretty good. Could you please run `npm run format:fix` and commit the results?

Done!

mvines:
@dr497 - are the private keys for each of these NFTs retained somewhere?   Short term adding them into the explorer like this is ok.  Longer term we'll be working on a naming service that'll allow the NFT details to be stored on-chain so this kind of pull request becomes unnecessarily.  To store the details on-chain, we're thinking that the private key corresponding to the each NTF mint address will be required

dr497:
> @dr497 - are the private keys for each of these NFTs retained somewhere?   Short term adding them into the explorer like this is ok.  Longer term we'll be working on a naming service that'll allow the NFT details to be stored on-chain so this kind of pull request becomes unnecessarily.  To store the details on-chain, we're thinking that the private key corresponding to the each NTF mint address will be required

Yes, they are retained in a safe place. Can't wait for the naming service!

PR #13553, opened 2020-11-12
indexes nodes' contact infos in crds table

behzadnouri:
#### Problem
In several places in gossip code, the entire crds table is scanned only
to filter out nodes' contact infos. Currently on mainnet, crds table is
of size ~70k, while there are only ~470 nodes. So the full table scan is
inefficient. Instead we may maintain an index of only nodes' contact
infos.

#### Summary of Changes
Indexed nodes' contact infos in crds table.

### DIFF
--- a/core/src/crds.rs
+++ b/core/src/crds.rs
@@ -242,8 +242,11 @@ impl Crds {
         if let CrdsData::ContactInfo(_) = value.value.data {
             assert!(self.nodes.swap_remove(&index));
         }
-        // The previously last element in the table is now moved to the
-        // 'index' position. Shards and nodes need to be updated accordingly.
+        // If index == self.table.len(), then the removed entry was the last
+        // entry in the table, in which case no other keys were modified.
+        // Otherwise, the previously last element in the table is now moved to
+        // the 'index' position; and so shards and nodes need to be updated
+        // accordingly.
         let size = self.table.len();
         if index < size {
             let value = self.table.index(index);
--- a/core/src/crds.rs
+++ b/core/src/crds.rs
@@ -26,12 +26,15 @@
 
 use crate::contact_info::ContactInfo;
 use crate::crds_shards::CrdsShards;
-use crate::crds_value::{CrdsValue, CrdsValueLabel};
+use crate::crds_value::{CrdsData, CrdsValue, CrdsValueLabel};
 use bincode::serialize;
 use indexmap::map::{Entry, IndexMap};
+use indexmap::set::IndexSet;
 use rayon::{prelude::*, ThreadPool};
 use solana_sdk::hash::{hash, Hash};
 use solana_sdk::pubkey::Pubkey;
+use solana_sdk::signature::Keypair;
+use solana_sdk::timing::timestamp;
 use std::cmp;
 use std::collections::HashMap;
 use std::ops::Index;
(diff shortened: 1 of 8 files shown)


behzadnouri:
cluster_info_stats2 repair_retransmit_peers_crds_peers
Left chunk is this commit, versus current code on the right.

core/src/crds.rs
```diff
@@ -44,6 +47,8 @@ pub struct Crds {
     pub table: IndexMap<CrdsValueLabel, VersionedCrdsValue>,
     pub num_inserts: usize,
     pub shards: CrdsShards,
+    // Indices of all crds values which are node ContactInfo.
+    nodes: IndexSet<usize>,
```
carllin, line 51: maybe a more explicit name like `contact_infos`?

core/src/crds.rs
```diff
@@ -166,6 +183,19 @@ impl Crds {
         self.table.get(&label)?.value.contact_info()
     }
 
+    /// Returns all entries which are ContactInfo.
+    pub fn get_nodes(&self) -> impl Iterator<Item = &VersionedCrdsValue> {
```
carllin, line 187: how about `get_versioned_contact_infos`

core/src/crds.rs
```diff
@@ -166,6 +183,19 @@ impl Crds {
...
+        self.nodes.iter().map(move |i| self.table.index(*i))
+    }
+
+    /// Returns ContactInfo of all known nodes.
+    pub fn get_nodes_contact_info(&self) -> impl Iterator<Item = &ContactInfo> {
```
carllin, line 192: how about just `get_contact_infos`

core/src/crds_gossip_pull.rs
```diff
@@ -242,10 +242,9 @@ impl CrdsGossipPull {
     ) -> Vec<(f32, &'a ContactInfo)> {
         let mut rng = rand::thread_rng();
         let active_cutoff = now.saturating_sub(PULL_ACTIVE_TIMEOUT_MS);
-        crds.table
-            .values()
+        crds.get_nodes()
```
carllin, line 245: Can we just use the `get_nodes_contact_info()` you defined above function to avoid the `value.value.contact_info().unwrap();`?

core/src/crds.rs
```diff
@@ -209,12 +239,20 @@ impl Crds {
...
         assert!(self.shards.remove(index, &value));
+        if let CrdsData::ContactInfo(_) = value.value.data {
+            assert!(self.nodes.swap_remove(&index));
+        }
         // The previously last element in the table is now moved to the
-        // 'index' position. Shards need to be updated accordingly.
-        if index < self.table.len() {
+        // 'index' position. Shards and nodes need to be updated accordingly.
+        let size = self.table.len();
+        if index < size {
```
carllin, line 248: The only time `index == size`  is if the key being removed is the "last"  element in the `table`. In those cases, the below `insert` is unnecessary, but isn't it still necessary to call the `remove` and `swap_remove` logic below?

carllin, line 248: For clarity, could we add a small comment that the only time `index == size` is if the `key` being removed is the last element, in which case there's no need to run the below logic because 1) No other keys were dispalaced, 2) The given `key` is already removed by the above logic

carllin (dismissed):

PR #13555, opened 2020-11-12
Fix potential undefined behavior

joshua-maros:
#### Problem
I'm the creator of the ouroboros crate. As Solana is the first big project to use it, I was curious to see where it was being used. Upon looking at the code I realized it was using the crate in a way I hadn't anticipated which could lead to undefined behavior. I've made an updated version of the crate (`0.5.1`) which solves these problems. This PR changes the code to use this version of the crate. I am planning to yank the older versions once this PR is accepted. To summarize the problem, my crate was not supposed to allow code to directly access fields inside a self-referencing struct, they are only supposed to be accessible through automatically generated helper functions to ensure safety. In writing the macro I had failed to account for fields explicitly marked as `pub`, which allowed code to access them in potentially unsafe ways.

#### Summary of Changes
- The version of `ouroboros` was bumped to `0.5.1`, an update which correctly prevents direct access of fields.
- The code was edited to access the fields of self-referencing structs only through the helper methods.
- The fields were made private again as there was no longer any need to make them public.
- I have checked to make sure `solana-runtime` compiles successfuly in a clean build on my machine.

### DIFF
--- a/runtime/src/accounts_index.rs
+++ b/runtime/src/accounts_index.rs
@@ -32,9 +32,9 @@ pub struct AccountMapEntryInner<T> {
 
 #[self_referencing]
 pub struct ReadAccountMapEntry<T: 'static> {
-    pub owned_entry: AccountMapEntry<T>,
+    owned_entry: AccountMapEntry<T>,
     #[borrows(owned_entry)]
-    pub slot_list_guard: RwLockReadGuard<'this, SlotList<T>>,
+    slot_list_guard: RwLockReadGuard<'this, SlotList<T>>,
 }
 
 impl<T: Clone> ReadAccountMapEntry<T> {
@@ -47,19 +47,19 @@ impl<T: Clone> ReadAccountMapEntry<T> {
     }
 
     pub fn slot_list(&self) -> &SlotList<T> {
-        &*self.slot_list_guard
+        &*self.borrow_slot_list_guard()
     }
 
     pub fn ref_count(&self) -> &AtomicU64 {
-        &self.owned_entry.ref_count
+        &self.borrow_owned_entry_contents().ref_count
     }
 }
 
 #[self_referencing]
 pub struct WriteAccountMapEntry<T: 'static> {
-    pub owned_entry: AccountMapEntry<T>,
+    owned_entry: AccountMapEntry<T>,
     #[borrows(owned_entry)]
-    pub slot_list_guard: RwLockWriteGuard<'this, SlotList<T>>,
+    slot_list_guard: RwLockWriteGuard<'this, SlotList<T>>,
 }
 
 impl<T: 'static + Clone> WriteAccountMapEntry<T> {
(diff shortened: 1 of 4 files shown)


ryoqun:
@joshua-maros Thanks for contributing! we're really glad that the upstream crate author cares downstream users. :)

We considered various alternatives like `rental` (unmaintained), ` owning_ref` (needed `unsafe` at our side), `reffers`. But your crate just worked. We so thank you for your open source project.

Actually, we're already kind of using this on our staging environment for a wekk or two, not on production. But we haven't seen odd errors, though.
A bit hard question, but do you have any idea for how often the undefined behavior will be triggered?

ryoqun:
@joshua-maros Also, is there any concerns for performance impact by using the correct way of accessing fields?

runtime/src/accounts_index.rs
```diff
@@ -72,15 +72,18 @@ impl<T: 'static + Clone> WriteAccountMapEntry<T> {
...
     pub fn slot_list(&mut self) -> &SlotList<T> {
-        &*self.slot_list_guard
+        &*self.borrow_slot_list_guard()
     }
 
-    pub fn slot_list_mut(&mut self) -> &mut SlotList<T> {
-        &mut *self.slot_list_guard
+    pub fn slot_list_mut<RT>(
+        &mut self,
+        user: impl for<'this> FnOnce(&mut RwLockWriteGuard<'this, SlotList<T>>) -> RT,
```
ryoqun, line 80: cool. you're the first to commit  hrtb in our repo. :)

ryoqun:
@joshua-maros Also, it seems that our CI is failing. Could you work on this?: https://buildkite.com/solana-labs/solana/builds/34845#194f5245-e21a-4346-9547-cbadb137cd5a/186-194

Also, I think this pr is in pretty good shape. I think we can merge this soon.

joshua-maros:
From doing the refactor, it looks like the code as it stood before this PR did not contain any undefined behavior, but someone editing the code could accidentally introduce UB. One of the things the macro does is replace `'this` with `'static` in the struct definition, so if someone uses one of the public fields incorrectly, they could end up with an invalid reference. 

I have not tested for performance impact, but there should be little to none. The actual implementation of the auto-generated functions looks something like this:
```rust
fn borrow_something(&self) -> &Thing {
    &self.thing
}
fn use_something_mut<Return>(&mut self, fun: impl FnOnce(&mut Thing) -> Return) -> Return {
    fun(&mut self.thing)
}
```
The compiler should easily be able to optimize these away, maintaining the same performace as directly accessing the fields. Their purpose is to just enforce extra borrow checker limitations at compile time so no illegal references are stored inside or outside the struct.

joshua-maros:
Looks like the CI was noting that a lock file in another part of the project was outdated. I ran the suggested command `./scripts/cargo-for-all-lock-files.sh tree`

ryoqun:
> Travis CI - Pull Request Expected — Waiting for status to be reported 

Ignoring this because this isn't strict requirement for this nature of changes.

ryoqun (commented):
lgtm!

ryoqun:
@joshua-maros Thanks for contributing us! I just merged this. :)

runtime/src/accounts_index.rs
```diff
@@ -32,9 +32,9 @@ pub struct AccountMapEntryInner<T> {
 
 #[self_referencing]
 pub struct ReadAccountMapEntry<T: 'static> {
-    pub owned_entry: AccountMapEntry<T>,
+    owned_entry: AccountMapEntry<T>,
```
carllin, line 35: @joshua-maros I was wondering why this wasn't working when I removed the `pub` 😄

carllin:
@joshua-maros thanks for the diligence, much appreciated!

mvines:
@joshua-maros - FYI that yanking the older crate like you did has been super disruptive over here.  Many builds are now broken. I really appreciate the PR though so thanks for that!

PR #13569, opened 2020-11-13
Check for overflow in rent partition calculation

carllin:
#### Problem
if start_index == end_index == partition_count - 1,  this calculation: https://github.com/solana-labs/solana/blob/master/runtime/src/bank.rs#L3043 can panic with overflow

#### Summary of Changes
Add a check similar to that of `end_index` for `start_index` overflow.
Fixes #


ryoqun:
@carllin oh. one of my proudly hardened code got beaten... Could you write a unit test? I have crafted a nice gadget to test overflow regardless of release/debug build: https://github.com/solana-labs/solana/blob/master/runtime/src/bank.rs#L5029-L5042

carllin:
@ryoqun is this safe to check into 1.3/1.4 without gating? seems like it can possibly cause divergent behavior if people are collecting rent on different partitions (overflow/vs not overflow)?

sdk/src/feature_set.rs
```diff
@@ -43,7 +43,7 @@ pub mod sha256_syscall_enabled {
 }
 
 pub mod no_overflow_rent_distribution {
-    solana_sdk::declare_id!("4kpdyrcj5jS47CZb2oJGfVxjYbsMm2Kx97gFyZrxxwXz");
+    solana_sdk::declare_id!("9UQYJwv7RXK3mQQzAaxXcYmdVfii3xAGFMkuYnzJUwMX");
```
carllin, line 46: @ryoqun  I repurposed this existing feature, hopefully that is ok!

runtime/src/bank.rs
```diff
@@ -3046,6 +3036,8 @@ impl Bank {
         let partition_width = (PREFIX_MAX - partition_count + 1) / partition_count + 1;
         let mut start_key_prefix = if start_index == 0 && end_index == 0 {
             0
+        } else if start_index + 1 == partition_count && is_fix_enabled {
```
ryoqun, line 3039: @carllin Hmm, I think you don't need gating for this because we can be proved not using overflowing `start_key_prefix`. Could you proof-read my following reasoning?:

1. When this `if` clause's condition ( `start_index + 1 == partition_count`) is met, `end_index` is `>= start_index` because of this assert `assert!(start_index <= end_index);` at line 3017.
2. Then, `end_key_prefix` is always set to `PREFIX_MAX` because `end_index + 1 == partition_count` should be always evaluated to true at line 3045 as well.
3. Also note that there is no case of `end_index + 1 > partition_count` because of `start_index < partition_count` and `end_index < partition_count` at lines 3018-3019
4. Finally, if `end_key_prefix == PREFIX`, we're overwriting `start_key_prefix` with `end_key_prefix` at line 3055.
5. So, there should be no observable case to use overflowing value of `start_key_prefix`.

Maybe, you're only encountered this bug with debug build. In that case, it'll immediately panic on the detection of overflowing value..

sdk/src/feature_set.rs
ryoqun: If I'm correct, I think you don't need feature gating for this, meaning less work and no feature activation ops. https://github.com/solana-labs/solana/pull/13569/files#r526635538

Also, I have concerns for repurposing this:
- this isn't still enabled on mb (but will be enabled soon).
- this will made replaying past slots by `ledger-tool` a bit harder. I think we should at least wait a month or so in that regard.

runtime/src/bank.rs
ryoqun: in other words, this is harmless unless a debug build validator is running on any live clusters, which isn't practice because of slow binary to begin with... ;)

runtime/src/bank.rs
```diff
@@ -5646,6 +5617,38 @@ pub(crate) mod tests {
                     0xff, 0xff, 0xff, 0xff, 0xff, 0xff
                 ])
         );
+
+        fn check_no_remainder(partition_count: u64) -> bool {
```
ryoqun, line 5621: nits: hmm, how about renaming this to `should_cause_overflow`, the comment is great but a bit hard to grasp what this is doing at quick glance.

runtime/src/bank.rs
```diff
@@ -5646,6 +5617,38 @@ pub(crate) mod tests {
...
+            // Check `result = (u64::max_value() + 1) / partition_count` is exact and
```
ryoqun, line 5622: nits: also rename from `result` to  `partition_width`?

runtime/src/bank.rs
```diff
@@ -5646,6 +5617,38 @@ pub(crate) mod tests {
...
+            // does not have a remainder.
+            // This way, `result * partition_count == (u64::max_value() + 1)`,
+            // so the test actually tests for overflow
+            (u64::max_value() - partition_count + 1) % partition_count == 0
```
ryoqun, line 5626: yeah, now I see `%`  works for overflow detection, what a clever way. :)

runtime/src/bank.rs
```diff
@@ -5646,6 +5617,38 @@ pub(crate) mod tests {
...
+        }
+
+        let max_exact = 64;
+        // Make sure `max` divides evenly when calculating `calculate_partition_width`
```
ryoqun, line 5630: nits: 

```
// Make sure `max_exact` divides evenly and eventually could have caused an overflow
```

runtime/src/bank.rs
```diff
@@ -5646,6 +5617,38 @@ pub(crate) mod tests {
...
+        assert!(check_no_remainder(max_exact));
+        // Make sure `max` doesn't divide evenly when calculating `calculate_partition_width`
```
ryoqun, line 5632: nits:

```
// Make sure `max_unexact` doesn't divide evenly and eventually could not have caused an overflow
```

ryoqun (approved):
tentative LGTM (after gating is reverted if my reasoning is correct)!

Really thanks for fixing my bug..

runtime/src/bank.rs
carllin: @ryoqun yeah, I think I agree. For 4, the code in question:

```
if start_index != 0 && start_index == end_index {
            // n..=n (n != 0): a noop pair across epochs without a gap under
            // multi_epoch_cycle, just nullify it.
            if end_key_prefix == PREFIX_MAX {
                start_key_prefix = end_key_prefix;
                start_pubkey = end_pubkey;
```

Then the only case `end_key_prefix != PREFIX` is if `start_index == 0`. But if `start_index ==0` and `start_index + 1 == partition_count`,  then `partition_count == 1`, and so b/c `end_index < partition_count`, then `end_index == 0`. 

This would mean the: `let mut start_key_prefix = if start_index == 0 && end_index == 0 {
            0
        }`
clause would execute, bypassing my newly added `else if`, so behavior is also unchanged.

Thanks for the detailed reasoning, I should have looked at it more carefully, thanks!

runtime/src/bank.rs
```diff
@@ -5646,6 +5648,38 @@ pub(crate) mod tests {
                     0xff, 0xff, 0xff, 0xff, 0xff, 0xff
                 ])
         );
+
+        fn should_cause_overflow(partition_count: u64) -> bool {
+            // Check `partition_width = (u64::max_value() + 1) / partition_count` is exact and
+            // does not have a remainder.
+            // This way, `result * partition_count == (u64::max_value() + 1)`,
```
ryoqun, line 5655: Oops, i wasn't explicit...

nits: `result` => `partition_width`

runtime/src/bank.rs
ryoqun: no problem! well, I haven't thought enough when writing the code, that's my bad to begin with... ;)

ryoqun (dismissed):
LGTM!

PR #13576, opened 2020-11-13
Small cleaning update_epoch_stakes

ryoqun:
#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -1222,22 +1222,24 @@ impl Bank {
                 epoch >= leader_schedule_epoch.saturating_sub(MAX_LEADER_SCHEDULE_STAKES)
             });
 
-            let vote_stakes: HashMap<_, _> = self
-                .stakes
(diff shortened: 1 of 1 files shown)


runtime/src/bank.rs
```diff
@@ -1222,22 +1222,24 @@ impl Bank {
                 epoch >= leader_schedule_epoch.saturating_sub(MAX_LEADER_SCHEDULE_STAKES)
             });
 
-            let vote_stakes: HashMap<_, _> = self
-                .stakes
-                .read()
-                .unwrap()
-                .vote_accounts()
-                .iter()
-                .map(|(epoch, (stake, _))| (*epoch, *stake))
```
ryoqun, line 1231: `epoch` is obviously misnomer here: https://github.com/solana-labs/solana/blob/a8a77614fe011b199ae0632011a5d9deec9c14f9/runtime/src/stakes.rs#L194

PR #13605, opened 2020-11-15
Add stake calculation tests with inflation/slashing

ryoqun:
#### Problem

There is no test for expected change of stakes after inflation and slashing with stable rate, which is fixed at the original activation/deactivation with stake_history at the time.

In other words, there is no test to change `self.stake` as far as I know... Inflation and slashing directly mutates this. So, this isn't static state for the entire lifecycle of stake account.. Current set of tests implicitly tells otherwise. There should be at least one such a test.

#### Summary of Changes

~~Add it before we break it. Still a bit dirty code I'm occupied with other fixes.~~ I mostly cleaned this up.

Fixes #

### DIFF
--- a/programs/stake/src/stake_state.rs
+++ b/programs/stake/src/stake_state.rs
@@ -2058,6 +2058,164 @@ mod tests {
     }
 
     #[test]
+    fn test_inflation_and_slashing_with_activating_stake() {
+        let cluster_stake = 1_000;
+        let delegated_stake = 700;
+
+        let mut stake = Delegation {
+            stake: delegated_stake,
+            activation_epoch: 0,
+            deactivation_epoch: 4,
+            ..Delegation::default()
+        };
+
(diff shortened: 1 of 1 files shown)


programs/stake/src/stake_state.rs
```diff
@@ -1803,6 +1803,164 @@ mod tests {
         }
     }
 
+    #[test]
+    fn test_inflation_and_slashing_with_activating_stake() {
```
ryoqun, line 1807: todo: add an accompanying test for the case of deactivating accounts.

ryoqun: well, this test already contains deactivation!

ryoqun:
@CriesofCarrots Could you glance at this at your convenient time?

CriesofCarrots (approved):
lgtm

PR #13614, opened 2020-11-16
run.sh: Use default client keypair for the faucet

mvines:
This is annoying:
```
$ cargo build
$ ./run.sh
$ solana balance
0 SOL
```
What, I need to `solana airdrop`?  Why.

Instead with this PR:
```
$ cargo build
$ ./run.sh
$ solana balance
500000000 SOL
```
🍪


mvines:
Failure here is just https://github.com/solana-labs/solana/pull/13342

PR #13615, opened 2020-11-16
Reject faked stake/vote accounts in stake mgmt.

ryoqun:
#### Problem

There is no `account.owner` check for referenced stake/vote accounts when delegating/splitting/merging stake accounts.

Because these referenced accounts are read-only, the runtime doesn't enforce the `account.owner` check. Users can pass any accounts of any owner as long as it deserializes into `StakeState` or `VoteState`, so we must check `account.owner` by ourselves.

This time I checked this trait for possible similar issues, but I found other uses of `KeyedAccount` without `owner` check is legitimate: https://github.com/solana-labs/solana/blob/6c5fb329b292c1abf7aaf62cf5558f5a9a1ca98b/programs/stake/src/stake_state.rs#L53-L106

#### Summary of Changes

Add guards and outright reject such transactions.

Also, this is feature-gated under `stake_program_v2`, implicitly.

Fixes #

### DIFF
--- a/programs/stake/src/stake_instruction.rs
+++ b/programs/stake/src/stake_instruction.rs
@@ -532,9 +532,18 @@ mod tests {
         sysvar::stake_history::StakeHistory,
     };
     use std::cell::RefCell;
+    use std::str::FromStr;
 
     fn create_default_account() -> RefCell<Account> {
         RefCell::new(Account::default())
+    }
+
+    fn bad_stake_pubkey() -> Pubkey {
+        Pubkey::from_str("BadStake11111111111111111111111111111111111").unwrap()
+    }
+
+    fn bad_vote_pubkey() -> Pubkey {
+        Pubkey::from_str("BadVote111111111111111111111111111111111111").unwrap()
     }
 
     fn process_instruction(instruction: &Instruction) -> Result<(), InstructionError> {
@@ -552,6 +561,14 @@ mod tests {
                     config::create_account(0, &config::Config::default())
                 } else if sysvar::rent::check_id(&meta.pubkey) {
                     account::create_account(&Rent::default(), 1)
+                } else if meta.pubkey == bad_stake_pubkey() {
+                    let mut account = Account::default();
+                    account.owner = id();
+                    account
+                } else if meta.pubkey == bad_vote_pubkey() {
+                    let mut account = Account::default();
+                    account.owner = solana_vote_program::id();
+                    account
                 } else {
                     Account::default()
                 })
@@ -599,14 +616,14 @@ mod tests {
                     &Pubkey::default(),
                     &Pubkey::default(),
                     100,
-                    &Pubkey::default()
+                    &bad_stake_pubkey(),
                 )[1]
             ),
             Err(InstructionError::InvalidAccountData),
         );
         assert_eq!(
             process_instruction(
-                &merge(&Pubkey::default(), &Pubkey::default(), &Pubkey::default(),)[0]
+                &merge(&Pubkey::default(), &bad_stake_pubkey(), &Pubkey::default(),)[0]
             ),
             Err(InstructionError::InvalidAccountData),
         );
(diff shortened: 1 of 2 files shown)


ryoqun:
@CriesofCarrots Could you review this before #13461 ?

@rwalker-com Hi, could you review this?

programs/stake/src/stake_instruction.rs
```diff
@@ -552,6 +561,14 @@ mod tests {
...
                 } else if sysvar::rent::check_id(&meta.pubkey) {
                     account::create_account(&Rent::default(), 1)
+                } else if meta.pubkey == bad_stake_pubkey() {
+                    let mut account = Account::default();
+                    account.owner = id();
+                    account
+                } else if meta.pubkey == bad_vote_pubkey() {
+                    let mut account = Account::default();
+                    account.owner = solana_vote_program::id();
+                    account
```
CriesofCarrots, line 571: These are for test cases where the owner should be correct but the account data can't be deserialized as StakeState/VoteState respectively, correct?
I found that a little confusing in the context of the rest of the PR, which deals with valid accounts with bad owners.
I think renaming the pubkey functions a little would help... I'll suggest above

programs/stake/src/stake_instruction.rs
```diff
@@ -532,11 +532,20 @@ mod tests {
...
         RefCell::new(Account::default())
     }
 
+    fn bad_stake_pubkey() -> Pubkey {
+        Pubkey::from_str("BadStake11111111111111111111111111111111111").unwrap()
+    }
+
+    fn bad_vote_pubkey() -> Pubkey {
+        Pubkey::from_str("BadVote111111111111111111111111111111111111").unwrap()
+    }
```
CriesofCarrots, line 547:
```suggestion
    fn invalid_stake_state_pubkey() -> Pubkey {
        Pubkey::from_str("BadStake11111111111111111111111111111111111").unwrap()
    }

    fn invalid_vote_state_pubkey() -> Pubkey {
        Pubkey::from_str("BadVote111111111111111111111111111111111111").unwrap()
    }
```

CriesofCarrots (approved):
Fix and process tests lgtm!
One nit to make stake_instruction test changes more clear.

programs/stake/src/stake_state.rs
```diff
@@ -821,6 +821,10 @@ impl<'a> StakeAccount for KeyedAccount<'a> {
         config: &Config,
         signers: &HashSet<Pubkey>,
     ) -> Result<(), InstructionError> {
+        if vote_account.owner()? != solana_vote_program::id() {
+            return Err(InstructionError::IncorrectProgramId);
+        }
```
ryoqun, line 826: @jackcmay just making sure; it seems that `IncorrectProgramId` isn't used much; but I think this is a perfect usecase of `IncorrectProgramId` for these kinds of checks, I guess.

programs/stake/src/stake_instruction.rs
ryoqun: thanks for the suggestion!: done: https://github.com/solana-labs/solana/pull/13615/commits/933dcfd465775638f63cc323227d0b40cd6d55d6

rwalker-com (approved):
lgtm

programs/stake/src/stake_state.rs
jackcmay: The description of this error is doesn't make much sense but the way you propose to use it here makes as much sense as any ;-)

ryoqun:
@CriesofCarrots @rob-solana Thanks for reviewing! I'm going to merge this shortly after now. I'm doing last-minute local test this.

programs/stake/src/stake_state.rs
CriesofCarrots: Fwiw, this seems perfectly consistent with how the error is used in the Vest program

ryoqun: yeah, I copied from that. :)

ryoqun:
> buildkite/solana/downstream-projects — Failed (exit status 101) 

Merging this despite the quoted build failure; I'm aware this is already being addressed and the failure isn't related to this pr.

PR #13616, opened 2020-11-16
Default preflight_commitment to confirmation commitment

mvines:
`RpcClient::send_and_confirm_transaction_with_spinner_and_commitment()` - the user cannot specify a preflight commitment set it to the user-supplied commitment instead of max, which is more likely to be what the user really wants

`sendAndConfirmTransaction()`/`sendAndConfirmRawTransaction()` - same

### DIFF
--- a/client/src/rpc_client.rs
+++ b/client/src/rpc_client.rs
@@ -1264,7 +1264,10 @@ impl RpcClient {
         self.send_and_confirm_transaction_with_spinner_and_config(
             transaction,
             commitment,
-            RpcSendTransactionConfig::default(),
+            RpcSendTransactionConfig {
+                preflight_commitment: Some(commitment.commitment),
+                ..RpcSendTransactionConfig::default()
+            },
         )
     }
 
--- a/web3.js/src/connection.js
+++ b/web3.js/src/connection.js
@@ -2933,15 +2933,10 @@ export class Connection {
     const skipPreflight = options && options.skipPreflight;
     const preflightCommitment = options && options.preflightCommitment;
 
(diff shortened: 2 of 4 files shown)


client/src/rpc_client.rs
```diff
@@ -1264,7 +1264,10 @@ impl RpcClient {
         self.send_and_confirm_transaction_with_spinner_and_config(
             transaction,
             commitment,
-            RpcSendTransactionConfig::default(),
+            RpcSendTransactionConfig {
+                preflight_commitment: Some(commitment.commitment),
```
paul-schaaf, line 1268: im not familar with the rust client. is it intentional to not do something like `options.preflightCommitment || options.commitment` here?

mvines: In this case the user can no way to provide `preflightCommitment` into the function.

mvines:
🤔 web3.js tests blew up on me

web3.js/src/connection.js
```diff
@@ -2933,15 +2933,10 @@ export class Connection {
     const skipPreflight = options && options.skipPreflight;
     const preflightCommitment = options && options.preflightCommitment;
 
-    if (skipPreflight && preflightCommitment) {
-      throw new Error(
-        'cannot set preflightCommitment when skipPreflight is enabled',
```
mvines, line 2938: @jstarry - I removed this code because it felt to opinionated to me (and it blew up on me).   I prefer the JSON RPC server to make the call about whether this is a problem or not (it currently doesn't care)

jstarry (approved):

PR #13623, opened 2020-11-16
Revert solana-tokens to RpcClient

garious:
#### Problem

solana-tokens broken on mainnet because BanksServer port isn't load balanced.

#### Summary of Changes

Revert to RpcClient


garious:
Thanks @CriesofCarrots!!

PR #13626, opened 2020-11-17
Fix assertion failure

carllin:
#### Problem
Assert fails to check additional case that could trigger shred completion updates

#### Summary of Changes
Fixes: https://github.com/solana-labs/solana-security/issues/14

Fixes #


sakridge (approved):
looks good, thanks

PR #13628, opened 2020-11-17
Prevent scans on unrooted slots from seeing partial clean

carllin:
#### Problem
The promised follow up to Follow up to https://github.com/solana-labs/solana/pull/13464 😄 

Scans on unrooted slots can  see edge cases like:

```
                        slot 0
                          |
                        slot 1
                      /        \
                 slot 2         |
                    |       slot 3 (max root)
            slot 4 (scan)
```
By the time the scan on slot 4 is called, slot 2 may already have been cleaned by a clean on slot 3, but slot 4 may not have been cleaned. Because the state in slot 2 would have been purged and is not saved in any roots, updates from slot 4 are observable, while updates from slot 2 aren't. In this case, a scan on slot 4 wouldn't accurately reflect the state when bank 4 was frozen. In cases like this, we default to a scan on the latest root `3`.

#### Summary of Changes
Detect these cases where an unrooted ancestor has been cleaned like the example above. In these cases, it means there was a bigger root that was set than the bank currently being cleaned (root 3 makes everything after 2 outdated in the example above), so instead just default to a scan on the latest root because it has more up-to-date information.

Fixes #

### DIFF
--- a/client/src/thin_client.rs
+++ b/client/src/thin_client.rs
@@ -561,11 +561,6 @@ impl AsyncClient for ThinClient {
         serialize_into(&mut wr, &transaction)
             .expect("serialize Transaction in pub fn transfer_signed");
         assert!(buf.len() < PACKET_DATA_SIZE);
-        info!(
-            "sending {} to {}",
-            transaction.signatures[0],
-            self.tpu_addr()
-        );
         self.transactions_socket
             .send_to(&buf[..], &self.tpu_addr())?;
         Ok(transaction.signatures[0])
--- a/client/src/thin_client.rs
+++ b/client/src/thin_client.rs
@@ -3,7 +3,7 @@
 //! messages to the network directly. The binary encoding of its messages are
 //! unstable and may change in future releases.
 
-use crate::{rpc_client::RpcClient, rpc_response::Response};
+use crate::{rpc_client::RpcClient, rpc_config::RpcProgramAccountsConfig, rpc_response::Response};
 use bincode::{serialize_into, serialized_size};
 use log::*;
 use solana_sdk::{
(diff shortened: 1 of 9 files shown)


runtime/src/accounts_index.rs
```diff
@@ -220,9 +221,124 @@ impl<T: 'static + Clone> AccountsIndex<T> {
...
+
+            By the time the scan on slot 4 is called, slot 2 may already have been
+            cleaned by a clean on slot 3, but slot 4 may not have been cleaned.
+            The state in slot 2 would have been purged and is not saved in any roots.
+            In this case, a scan on slot 4 wouldn't accurately reflect he state when bank 4
+            was frozen. In cases like this, we default to a scan on the latest roots by
+            removing all `ancestors`.
+            */
+            &empty
+        };
```
carllin, line 303: This is the fix, removing it will cause `test_store_scan_consistency_unrooted` to fail

client/src/thin_client.rs
```diff
@@ -555,6 +561,11 @@ impl AsyncClient for ThinClient {
         serialize_into(&mut wr, &transaction)
             .expect("serialize Transaction in pub fn transfer_signed");
         assert!(buf.len() < PACKET_DATA_SIZE);
+        info!(
+            "sending {} to {}",
```
sakridge, line 565: remove or `debug!`?

runtime/src/accounts_index.rs
```diff
@@ -220,9 +220,124 @@ impl<T: 'static + Clone> AccountsIndex<T> {
...
+
+        Diagram 2:
+
+                Build fork structure:
+                        slot 0
+                          |
+                        slot 1
+                    /            \
+        slot 2 (max_root)         |
+                |            slot 3 (potential newer max root)
```
sakridge, line 323: slot 2 is rooted, right? how could slot 3 be rooted which doesn't chain to slot 2?

runtime/src/bank.rs
```diff
@@ -3045,6 +3046,8 @@ impl Bank {
         let partition_width = (PREFIX_MAX - partition_count + 1) / partition_count + 1;
         let mut start_key_prefix = if start_index == 0 && end_index == 0 {
             0
+        } else if start_index + 1 == partition_count {
```
sakridge, line 3049: Change like this would need to be gated, right? It could change rent behavior. Also, is it specifically related or necessary for the scan fix?

carllin: ah sorry,  I should've mentioned that this PR is dependent on another PR: https://github.com/solana-labs/solana/pull/13569#issuecomment-727121225 going in first, where gating will be necessary.

runtime/src/accounts_index.rs
carllin: yeah you're right, updated and changed the `max_root` to be at slot 1, thanks!

sakridge (approved):

PR #13633, opened 2020-11-17
Add back BPF error logging

jackcmay:
#### Problem

Logging of the actual BPF was lost when stable logging was introduced

#### Summary of Changes

Add back the BPF specific error that led to the more generic `InstructionError`

Fixes #

### DIFF
--- a/programs/bpf_loader/src/lib.rs
+++ b/programs/bpf_loader/src/lib.rs
@@ -295,6 +295,12 @@ impl Executor for BPFExecutor {
                     }
                 }
                 Err(error) => {
+                    log!(
+                        logger,
+                        "Program {} vm error: {}",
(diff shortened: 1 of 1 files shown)


programs/bpf_loader/src/lib.rs
```diff
@@ -295,6 +295,12 @@ impl Executor for BPFExecutor {
                     }
                 }
                 Err(error) => {
+                    log!(
+                        logger,
+                        "Program {} vm error: {}",
```
mvines, line 300:
```suggestion
                        "Program {} BPF VM error: {}",
```

mvines:
@jackcmay - I hope my little nit is ok :)

jackcmay:
Yup, thanks @mvines !

PR #13638, opened 2020-11-17
Add RpcClient::new_with_commitment()

mvines:
Add the ability to override the default commitment in RpcClient (much like what solana-web3.js provides)

Example of the resulting client cleanup: https://github.com/solana-labs/solana-program-library/pull/853


mvines:
Failed in `accounts_hash_verifier::tests::test_max_hashes`, unrelated

PR #13646, opened 2020-11-17
Update Initialized split rent-exempt value

CriesofCarrots:
#### Problem
In the churn of fixing stake split, we missed one case: splitting a StakeState::Initialized account to a smaller StakeState::Initialized account. This operation needs to include the adjustment for the rent_exempt_reserve to be correct on the new account.

#### Summary of Changes
Add rent_exempt_reserve adjustment for stake-split Initialized


ryoqun (dismissed):
LGTM! Thanks for quickly working on this!

PR #13654, opened 2020-11-17
Fix fragile tests in prep of stake rewrite pr

ryoqun:
#### Problem

Existing tests are creating bad stake accounts which are less than `rent_exempt_reserve`, notably only `42` lamports... #13461 will forbid such accounts. So, fix tests first.

#### Summary of Changes

Not ideal, but just hacky mindless replacement: `BOOTSTRAP_VALIDATOR_LAMPORTS` => `bootstrap_validator_lamports()`

Fixes #

### DIFF
--- a/runtime/src/genesis_utils.rs
+++ b/runtime/src/genesis_utils.rs
@@ -1,3 +1,4 @@
+use log::*;
 use solana_sdk::{
     account::Account,
     feature::{self, Feature},
@@ -10,11 +11,15 @@ use solana_sdk::{
     system_program,
 };
 use solana_stake_program::stake_state;
+use solana_stake_program::stake_state::StakeState;
 use solana_vote_program::vote_state;
 use std::borrow::Borrow;
 
 // The default stake placed with the bootstrap validator
-pub const BOOTSTRAP_VALIDATOR_LAMPORTS: u64 = 42;
+// fun fact: we're very close to make this const fn.
+pub fn bootstrap_validator_lamports() -> u64 {
+    StakeState::get_rent_exempt_reserve(&Rent::default())
+}
 
 pub struct ValidatorVoteKeypairs {
     pub node_keypair: Keypair,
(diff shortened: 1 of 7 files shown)


runtime/src/genesis_utils.rs
```diff
@@ -114,13 +119,18 @@ pub fn create_genesis_config_with_leader(
     bootstrap_validator_pubkey: &Pubkey,
     bootstrap_validator_stake_lamports: u64,
 ) -> GenesisConfigInfo {
+    //mint_lamports > bootstrap_validator_stake_lamports,
```
CriesofCarrots, line 122: Should this be an `if ...` clause?

runtime/src/genesis_utils.rs
```diff
@@ -10,11 +11,15 @@ use solana_sdk::{
...
+use solana_stake_program::stake_state::StakeState;
 use solana_vote_program::vote_state;
 use std::borrow::Borrow;
 
 // The default stake placed with the bootstrap validator
-pub const BOOTSTRAP_VALIDATOR_LAMPORTS: u64 = 42;
+// fun fact: we're very close to make this const fn.
+pub fn bootstrap_validator_lamports() -> u64 {
+    StakeState::get_rent_exempt_reserve(&Rent::default())
+}
```
CriesofCarrots, line 22: I actually find this redefinition a little confusing, because it is only when `bootstrap_validator_stake_lamports` is too low that we will see the problem described in this PR.
A low `bootstrap_validator_lamports` may cause different problems (which I know intimately after https://github.com/solana-labs/solana/pull/13627 :) ); if it's tied to the rent-exempt-balance of the account I think we will be more prone to misuse.

What about defining this separately as `bootstrap_validator_stake_lamports()` and only replacing the stake-lamports values?

runtime/src/genesis_utils.rs
ryoqun: yeah. corrected!

runtime/src/genesis_utils.rs
ryoqun: thanks for great suggestion. Well, I didn't pay enough attention to the code... I think I've applied what you have suggested. :)

CriesofCarrots (approved):
Thanks for the separate PR for this, and for the polish! lgtm

ryoqun:
Strictly speaking, #13461 didn't needed this. But, anyway I merged this because it's nice to have. Also, backported to v1.4 because this is test only change (very low risk) and to avoid future backport churn.

PR #13671, opened 2020-11-18
Add new inflation feature-ids

CriesofCarrots:
#### Problem
We need plumbing for pico- and full-inflation features in order to enable on testnet and mainnet-beta

#### Summary of Changes
- Add new pico_inflation feature-id (generated by [feature-proposal cli](https://github.com/solana-labs/solana-program-library/tree/master/feature-proposal/cli))
- Add full_inflation feature-id (generated by feature-proposal cli)
- Remove inflation_kill_switch feature
- Add inflation-start calculation so that Inflation values are correct on activation
- Add full_inflation values

cc @mvines

### DIFF
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -1335,6 +1335,12 @@ impl Bank {
         (examined_count, rewritten_count)
     }
 
+    fn get_inflation_start_slot(&self) -> Slot {
+        self.feature_set
+            .activated_slot(&feature_set::full_inflation::id())
+            .unwrap_or(0)
+    }
+
     // update rewards based on the previous epoch
     fn update_rewards(
         &mut self,
@@ -1347,8 +1353,9 @@ impl Bank {
         // if I'm the first Bank in an epoch, count, claim, disburse rewards from Inflation
 
         // calculated as: prev_slot / (slots / year)
-        let slot_in_year =
-            (self.epoch_schedule.get_last_slot_in_epoch(prev_epoch)) as f64 / self.slots_per_year;
+        let num_slots = self.epoch_schedule.get_last_slot_in_epoch(prev_epoch)
+            - self.get_inflation_start_slot();
+        let slot_in_year = num_slots as f64 / self.slots_per_year;
 
         let epoch_duration_in_years = self.epoch_duration_in_years(prev_epoch);
 
@@ -10704,5 +10711,39 @@ pub(crate) mod tests {
         bank.store_account(&bootstrap_stake_pubkey, &bootstrap_stake_account);
 
         assert_eq!(bank.rewrite_stakes(), (1, 1));
+    }
+
+    #[test]
+    fn test_get_inflation_start_slot() {
+        let GenesisConfigInfo {
+            mut genesis_config, ..
+        } = create_genesis_config_with_leader(42, &solana_sdk::pubkey::new_rand(), 42);
+        genesis_config
+            .accounts
+            .remove(&feature_set::full_inflation::id())
+            .unwrap();
+        let bank = Bank::new(&genesis_config);
+
+        // Advance to slot 2
+        let mut bank = new_from_parent(&Arc::new(bank));
+        bank = new_from_parent(&Arc::new(bank));
+        assert_eq!(bank.get_inflation_start_slot(), 0);
+
+        // Request `full_inflation` activation
+        let full_inflation_activation_slot = 2;
+        bank.store_account(
+            &feature_set::full_inflation::id(),
+            &feature::create_account(
+                &Feature {
+                    activated_at: Some(full_inflation_activation_slot),
+                },
+                42,
+            ),
+        );
+        bank.compute_active_feature_set(true);
+        assert_eq!(
+            bank.get_inflation_start_slot(),
+            full_inflation_activation_slot
+        );
     }
 }
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -1350,6 +1350,17 @@ impl Bank {
             })
     }
 
+    fn get_inflation_num_slots(&self) -> u64 {
+        let inflation_activation_slot = self.get_inflation_start_slot();
+        // Normalize inflation_start to align with the start of rewards accrual.
+        let inflation_start_slot = self.epoch_schedule.get_first_slot_in_epoch(
+            self.epoch_schedule
+                .get_epoch(inflation_activation_slot)
+                .saturating_sub(1),
+        );
+        self.epoch_schedule.get_first_slot_in_epoch(self.epoch()) - inflation_start_slot
+    }
+
     // update rewards based on the previous epoch
     fn update_rewards(
         &mut self,
@@ -1361,9 +1372,8 @@ impl Bank {
         }
         // if I'm the first Bank in an epoch, count, claim, disburse rewards from Inflation
 
-        // calculated as: prev_slot / (slots / year)
-        let num_slots = self.epoch_schedule.get_last_slot_in_epoch(prev_epoch)
-            - self.get_inflation_start_slot();
+        // calculated as: num_slots / (slots / year)
+        let num_slots = self.get_inflation_num_slots();
         let slot_in_year = num_slots as f64 / self.slots_per_year;
 
         let epoch_duration_in_years = self.epoch_duration_in_years(prev_epoch);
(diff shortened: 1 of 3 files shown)


mvines (dismissed):

runtime/src/bank.rs
```diff
@@ -1305,8 +1320,9 @@ impl Bank {
         // if I'm the first Bank in an epoch, count, claim, disburse rewards from Inflation
 
         // calculated as: prev_slot / (slots / year)
```
ryoqun, line 1322: nits: update comment like this?: `prev_slot` => `num_slots`

runtime/src/bank.rs
```diff
@@ -1305,8 +1320,9 @@ impl Bank {
...
-        let slot_in_year =
-            (self.epoch_schedule.get_last_slot_in_epoch(prev_epoch)) as f64 / self.slots_per_year;
+        let num_slots = self.epoch_schedule.get_last_slot_in_epoch(prev_epoch)
```
ryoqun, line 1323: hmm, I think this could overflow at the initial inflation?

I think for the activated epoch, this is true: `self.get_inflation_start_slot()` (= start slot of new epoch) > `self.epoch_schedule.get_last_slot_in_epoch(prev_epoch)` (= end slot of old epoch)

runtime/src/bank.rs
```diff
@@ -4093,10 +4109,13 @@ impl Bank {
             self.rent_collector.rent.burn_percent = 50; // 50% rent burn
         }
 
-        if new_feature_activations.contains(&feature_set::inflation_kill_switch::id()) {
-            *self.inflation.write().unwrap() = Inflation::new_disabled();
-            self.fee_rate_governor.burn_percent = 100; // 100% fee burn
-            self.rent_collector.rent.burn_percent = 100; // 100% rent burn
+        if new_feature_activations.contains(&feature_set::full_inflation::id()) {
+            let mut full_inflation = Inflation::default();
```
ryoqun, line 4113: nits: I rather want to define `Inflation::full()`. well `ledger-tool` needs to access this for various simulation.

runtime/src/bank.rs
CriesofCarrots: Ah, thank you, I missed that case.
I think `saturating_sub` should be correct, right?

runtime/src/bank.rs
```diff
@@ -1304,9 +1319,12 @@ impl Bank {
...
         // if I'm the first Bank in an epoch, count, claim, disburse rewards from Inflation
 
-        // calculated as: prev_slot / (slots / year)
-        let slot_in_year =
-            (self.epoch_schedule.get_last_slot_in_epoch(prev_epoch)) as f64 / self.slots_per_year;
+        // calculated as: num_slots / (slots / year)
+        let num_slots = self
+            .epoch_schedule
+            .get_last_slot_in_epoch(prev_epoch)
+            .saturating_sub(self.get_inflation_start_slot());
```
ryoqun, line 1326: hmm, in this code, I think we don't pay rewards for the activation epoch...

I think the math should be like this?: `self.epoch_schedule.get_last_slot_in_epoch(prev_epoch) - self.epoch_schedule.get_last_slot_in_epoch(activation_epoch.saturating_sub(1))`

Anyway, this is good indicate we need to write a good unit test for this particular calculation by extracting this first...

such a test should check that `num_slots` is rougthly `320000`.

sdk/src/inflation.rs
```diff
@@ -69,6 +69,17 @@ impl Inflation {
         Self::new_fixed(0.0001) // 0.01% inflation
     }
 
+    pub fn full() -> Self {
+        Self {
+            initial: DEFAULT_INITIAL,
+            terminal: DEFAULT_TERMINAL,
+            taper: DEFAULT_TAPER,
```
ryoqun, line 76: @CriesofCarrots @ericlwilliams are we sure these are the final inflation parameters, maybe?: https://github.com/solana-labs/solana/blob/a1b16a583489bd6cd3dc71f2924d0fe2563d10e4/sdk/src/inflation.rs#L25-L27

runtime/src/bank.rs
ryoqun: also, we should be careful for genesis case...

sdk/src/inflation.rs
CriesofCarrots: I believe so: https://forums.solana.com/t/final-inflation-parameter-proposal-and-steps-to-implementation/977
But I certainly wouldn't mind @ericlwilliams 's eyes on this one more time before merge.

runtime/src/bank.rs
```diff
@@ -1293,6 +1293,33 @@ impl Bank {
...
+    fn get_inflation_num_slots(&self, prev_epoch: Epoch) -> u64 {
+        let inflation_activation_slot = self.get_inflation_start_slot();
+        // Normalize inflation_start to the last slot of the previous epoch to ensure that
+        // slot_in_year includes the complete epoch.
+        let inflation_start_slot = self.epoch_schedule.get_last_slot_in_epoch(
+            self.epoch_schedule
+                .get_epoch(inflation_activation_slot)
+                .saturating_sub(1),
+        );
+        self.epoch_schedule.get_last_slot_in_epoch(prev_epoch) - inflation_start_slot
```
ryoqun, line 1320: here is my paper simulation:

Say, we activated at 50112000 `Epoch: 116, Slots: 0/432000, [50112000..50544000)`.

as a ultimate goal, for the first inflation reward, we should reward for 2 days (= 432000 num_slots) worth duration.

to create the result, we can choose from...:

```
self.get_inflation_start_slot():50112000
activation_epoch: 116,
self.epoch_schedule.get_first_slot_in_epoch(activation_epoch.saturating_sub(1)): 49680000
self.epoch(): 116
prev_epoch: 115
self.epoch_schedule.get_last_slot_in_epoch(prev_epoch): 50111999
```

then finally this formulate returns `431999`:

```
let inflation_genesis_slot = self.epoch_schedule.get_first_slot_in_epoch(activation_epoch.saturating_sub(1)); // 49680000
let inflation_current_slot = self.epoch_schedule.get_last_slot_in_epoch(prev_epoch); // 50111999
let elapsed_slots = inflation_current_slot - inflation_genesis_slot; // => 431999
```

Then consider, next epoch `Epoch: 117, Slots: 0/432000, [50544000..50976000)`

```
self.get_inflation_start_slot():50112000 // should be same
activation_epoch: 116, // should be same
self.epoch_schedule.get_first_slot_in_epoch(activation_epoch.saturating_sub(1)): 49680000 // should be same
self.epoch(): 117
prev_epoch: 116
self.epoch_schedule.get_last_slot_in_epoch(prev_epoch): 50543999
```

then finally this formulate returns `863999`:

```
let inflation_genesis_slot = self.epoch_schedule.get_first_slot_in_epoch(activation_epoch.saturating_sub(1)); // 49680000
let inflation_current_slot = self.epoch_schedule.get_last_slot_in_epoch(prev_epoch); // 50543999
let elapsed_slots = inflation_current_slot - inflation_genesis_slot; // => 863999
```

and so on...

ryoqun: the basic idea is that we must start to the `taper` ticking down at the start of the epoch for which inflation is rewarded. so we should kind of need to shift an epoch for it.

ryoqun: btw, this adjustment wasn't needed if we're enabling inflation since genesis...
Because, we're starting in the middle and we want to reward at the activation_epoch for the duration `prev_epoch`, we have to be careful for the handling of these, I think...

sdk/src/inflation.rs
ericlwilliams: Yes! as long as the `DEFAULT_TAPER` is implemented as negative in the calcs

runtime/src/bank.rs
CriesofCarrots: This PR now aligns the taper with the start of rewards accrual.
However, I think that we don't really want to reward at the activation_epoch -- the activation should kick off accruals, and the reward should happen at the next epoch boundary.

runtime/src/bank.rs
```diff
@@ -1335,6 +1335,32 @@ impl Bank {
...
+
+    fn get_inflation_num_slots(&self) -> u64 {
+        let inflation_activation_slot = self.get_inflation_start_slot();
+        // Normalize inflation_start to align with the start of rewards accrual.
+        let inflation_start_slot = self.epoch_schedule.get_first_slot_in_epoch(
+            self.epoch_schedule
+                .get_epoch(inflation_activation_slot)
+                .saturating_sub(1),
+        );
+        self.epoch_schedule.get_first_slot_in_epoch(self.epoch()) - inflation_start_slot
```
ryoqun, line 1361: yeah, I like the new definition over the previous one (`self.epoch_schedule.get_last_slot_in_epoch(prev_epoch)`).

runtime/src/bank.rs
```diff
@@ -10702,4 +10728,140 @@ pub(crate) mod tests {
...
+                },
+                42,
+            ),
+        );
+        bank.compute_active_feature_set(true);
+        assert_eq!(bank.get_inflation_num_slots(), slots_per_epoch);
+        for _ in 0..slots_per_epoch {
+            bank = new_from_parent(&Arc::new(bank));
+        }
+        assert_eq!(bank.get_inflation_num_slots(), 2 * slots_per_epoch);
```
ryoqun, line 10846: as a one of this outcomes (https://github.com/solana-labs/solana/pull/13671/files#r527695008), the math got simpler, not needing off-by-one anymore.

ryoqun (approved):
lgtm!

PR #13673, opened 2020-11-18
SPL token balance in transaction metadata 

oJshua:
#### Problem
Transaction metadata does not include SPL Token balance changes.

#### Summary of Changes
- collects token balances of relevant accounts pre and post transaction
- stores deltas in associated transaction metadata 

Fixes https://github.com/solana-labs/solana/issues/12318


transaction-status/src/lib.rs
```diff
@@ -194,6 +227,16 @@ impl UiTransactionStatusMeta {
...
+            pre_token_balances: meta.pre_token_balances.map(|bal| {
+                bal.into_iter()
+                    .map(|bal| UiTransactionTokenBalance::parse(bal))
+                    .collect()
+            }),
+            post_token_balances: meta.post_token_balances.map(|bal| {
+                bal.into_iter()
+                    .map(|bal| UiTransactionTokenBalance::parse(bal))
+                    .collect()
+            }),
```
CriesofCarrots, line 239: Spell out `bal` in this section, imo

transaction-status/src/lib.rs
```diff
@@ -210,6 +253,26 @@ impl From<TransactionStatusMeta> for UiTransactionStatusMeta {
...
                 .map(|ixs| ixs.into_iter().map(|ix| ix.into()).collect()),
             log_messages: meta.log_messages,
+            pre_token_balances: meta.pre_token_balances.map(|balance| {
+                balance
+                    .into_iter()
+                    .map(|balance| UiTransactionTokenBalance {
+                        account_index: balance.account_index,
+                        mint: balance.mint,
+                        ui_token_amount: balance.ui_token_amount,
+                    })
```
CriesofCarrots, line 263: Could you impl From to make this slimmer/less repetitive?

core/src/transaction_status_service.rs
```diff
@@ -101,6 +109,36 @@ impl TransactionStatusService {
...
+                        .iter()
+                        .map(|balance| TransactionTokenBalance {
+                            account_index: balance.account_index as u8,
+                            mint: balance.mint.clone(),
+                            ui_token_amount: UiTokenAmount {
+                                ui_amount: balance.ui_token_amount.ui_amount,
+                                decimals: balance.ui_token_amount.decimals,
+                                amount: balance.ui_token_amount.amount.clone(),
+                            },
+                        })
```
CriesofCarrots, line 123: Could you impl From to make this slimmer/less repetitive?

transaction-status/src/lib.rs
```diff
@@ -128,6 +143,16 @@ impl UiInnerInstructions {
...
+impl UiTransactionTokenBalance {
+    fn parse(token_balance: TransactionTokenBalance) -> Self {
+        Self {
+            account_index: token_balance.account_index,
+            mint: token_balance.mint,
+            ui_token_amount: token_balance.ui_token_amount,
+        }
+    }
+}
+
```
CriesofCarrots, line 155: I don't think we need this special parse function, do we? I think the From implementations between UiTransactionTokenBalance and TransactionTokenBalance that I mention elsewhere should suffice?

runtime/Cargo.toml
```diff
@@ -46,6 +47,7 @@ solana-sdk = { path = "../sdk", version = "1.5.0" }
 solana-stake-program = { path = "../programs/stake", version = "1.5.0" }
 solana-vote-program = { path = "../programs/vote", version = "1.5.0" }
 solana-secp256k1-program = { path = "../programs/secp256k1", version = "1.5.0" }
+spl-token-v2-0 = { package = "spl-token", version = "=2.0.8" }
```
CriesofCarrots, line 50: Would prefer to avoid this dependency if possible. Let's move the new runtime functionality into a module in `solana-transaction-status`, and BankingStage and BlockstoreProcessor can call it from there. Both solana-core and solana-ledger both depend on transaction-status already

runtime/src/bank.rs
```diff
@@ -397,8 +400,35 @@ impl TransactionBalancesSet {
         }
     }
 }
+
 pub type TransactionBalances = Vec<Vec<u64>>;
 
+pub struct TransactionTokenBalance {
+    pub account_index: usize,
```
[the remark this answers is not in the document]
oJshua: Yes, I wasn't sure how to handle this situation so I went for something similar to InnerInstructions. Do you think that will be fine?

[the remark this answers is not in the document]
CriesofCarrots: Pity there isn't a nice protobuf analog for this. Yep, matching InnerInstructions is fine.

transaction-status/src/token_balances.rs
```diff
@@ -0,0 +1,115 @@
...
+            }
+        }
+
+        let mut transaction_balances: Vec<TransactionTokenBalance> = vec![];
+        for index in fetch_account_hash.keys() {
+            if let Some(account_id) = account_keys.get(*index as usize) {
+                if let Some(results) =
+                    collect_token_balance_from_account(&bank, account_id, &mut mint_decimals)
+                {
+                    let (mint, ui_token_amount) = results;
```
CriesofCarrots, line 79:
```suggestion
                if let Some((mint, ui_token_amount)) =
                    collect_token_balance_from_account(&bank, account_id, &mut mint_decimals)
                {
```

CriesofCarrots (commented):
Nice rework! Just a couple nits.
Backward-compat tests look fine?

runtime/src/bank.rs
```diff
@@ -26,6 +26,7 @@ use crate::{
 use byteorder::{ByteOrder, LittleEndian};
 use itertools::Itertools;
 use log::*;
+
```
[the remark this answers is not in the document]
oJshua: Oh thanks, I forgot to take one last pass

CriesofCarrots (approved):
R+ CI

PR #13689, opened 2020-11-19
Upgradeable loader

jackcmay:
#### Problem

Once finalized, programs are immutable and upgrading a program can be tedious and the mechanisms required are determined on a  case-by-case basis.  

Tracked by issue: #13694

#### Summary of Changes

Provide developers a mechanism to easily update programs on-chain

Fixes #


sdk/program/src/loader_upgradeable_instruction.rs
```diff
@@ -0,0 +1,93 @@
+//! Instruction types
+
+use crate::{
+    pubkey::Pubkey,
+};
+
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Clone)]
+pub enum UpgradeableLoaderInstruction {
```
mvines, line 8: Let's extend `LoaderInstruction` instead.   Loaders that aren't upgradable don't need to implement the new instructions tacked on the end

jackcmay: Not following, why would no-upgradeable loaders implemnt `UpgradeableLoaderInstruction`?  Also, what did you have in mind for `extend`?

sdk/program/src/bpf_loader_upgradeable.rs
```diff
@@ -0,0 +1,185 @@
...
+    loader_upgradeable_instruction::UpgradeableLoaderInstruction,
+    pubkey::Pubkey,
+    system_instruction, sysvar,
+};
+use bincode::serialized_size;
+
+crate::declare_id!("BPFLoaderUpgradeab1e11111111111111111111111");
+
+#[derive(Debug, Serialize, Deserialize, PartialEq, Clone, Copy, AbiExample)]
+pub enum UpgradableLoaderState {
```
mvines, line 22: How about `#[repr(u8)]`.  Save them bits

sdk/program/src/bpf_loader_upgradeable.rs
```diff
@@ -0,0 +1,185 @@
...
+    /// Account is not initialized.
+    Uninitialized,
+    /// An executable account.
+    Executable {
+        /// Public key of the code account.
+        code_pubkey: Pubkey,
```
mvines, line 28: nit:
```suggestion
        /// Address of the code account.
        code: Pubkey,
```

sdk/program/src/bpf_loader_upgradeable.rs
```diff
@@ -0,0 +1,185 @@
...
+) -> Result<Vec<Instruction>, InstructionError> {
+    Ok(vec![Instruction::new(
+        id(),
+        &UpgradeableLoaderInstruction::Write { offset, bytes },
+        vec![AccountMeta::new(*source_pubkey, false)],
+    )])
+}
+
+/// Returns the instructions required to deploy a program
+pub fn deploy(
```
mvines, line 92: We hope that once realloc comes in `max_program_size` will no longer be required, so how about reserving `deploy()` for that sunny day and calling this function:
```suggestion
pub fn deploy_with_max_program_size(
```

sdk/program/src/loader_upgradeable_instruction.rs
```diff
@@ -0,0 +1,75 @@
...
+    /// zero.
+    ///
+    /// # Account references
+    ///   0. [writable] The code account.
+    ///   1. [] The executable account.
+    ///   2. [] The source account where the program data has been written.
+    ///   4. [signer] The program's authority.
+    ///   5. [writable] The spills account.
+    ///   6. [] Rent sysvar
+    ///   7. [] Slot sysvar
```
mvines, line 62:
```suggestion
    ///   7. [] Clock sysvar
```

sdk/program/src/loader_upgradeable_instruction.rs
```diff
@@ -0,0 +1,75 @@
...
+    /// that creates the executable and code accounts. Otherwise another party
+    /// can acquire ownership of the uninitialized accounts.
+    ///
+    /// # Account references
+    ///   0. [writable] The code account.
+    ///   1. [writable] The executable account.
+    ///   2. [writable] The source account where the program data has been
+    ///      written.
+    ///   3. [] The program's authority.
+    ///   4. [] Rent sysvar
```
mvines, line 42: Need the Clock sysvar in here as well

sdk/program/src/loader_upgradeable_instruction.rs
```diff
@@ -0,0 +1,75 @@
...
+    ///   6. [] Rent sysvar
+    ///   7. [] Slot sysvar
+    Upgrade,
+
+    /// Sets a new program authority that is allowed to update the program.  To
+    /// permanently disable program updates set the authority to `None`.
+    ///
+    /// # Account references
+    ///   0. `[writable]` The code account to change the authority of.
+    ///   1. `[signer]` The current authority.
```
mvines, line 70: IMO pass the new authority as the second key.  None = no second key.
Less custom serde work for non-Rust clients

sdk/program/src/loader_upgradeable_instruction.rs
```diff
@@ -0,0 +1,75 @@
...
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Clone)]
+pub enum UpgradeableLoaderInstruction {
+    /// Write program data into a source account.
+    ///
+    /// A source account is an intermediary that once fully populated is used
+    /// with the `Finalize` instruction to populate the program's code account.
+    ///
+    /// # Account references
+    ///   0. [writable, signer] source account to write program data to.
+    Write {
```
mvines, line 14: How will `Write` know `0` is a source account and not a code or executable account?

sdk/program/src/bpf_loader_upgradeable.rs
```diff
@@ -0,0 +1,185 @@
...
+    Executable {
+        /// Public key of the code account.
+        code_pubkey: Pubkey,
+    },
+    // A code account.
+    Code {
+        /// Slot that the program was last modified.
+        slot: u64,
+        /// Program's authority.
+        authority: Option<Pubkey>,
```
mvines, line 35: nit: 
```suggestion
        /// Program's upgrade authority
        upgrade_authority: Option<Pubkey>,
```

sdk/program/src/loader_upgradeable_instruction.rs
```diff
@@ -0,0 +1,75 @@
...
+    ///     the program data and authority information.
+    ///
+    /// The `Deploy` instruction requires no signers and MUST be included within
+    /// the same Transaction as the system program's `CreateAccount` instruction
+    /// that creates the executable and code accounts. Otherwise another party
+    /// can acquire ownership of the uninitialized accounts.
+    ///
+    /// # Account references
+    ///   0. [writable] The code account.
+    ///   1. [writable] The executable account.
```
mvines, line 38: Document both accounts must be in the `UpgradableLoaderState::Uninitialized` state.

sdk/program/src/loader_upgradeable_instruction.rs
```diff
@@ -0,0 +1,75 @@
...
+    ///   2. [writable] The source account where the program data has been
+    ///      written.
+    ///   3. [] The program's authority.
```
mvines, line 41:
```suggestion
    ///   3. [] The program's upgrade authority.
```
For flexibility we should support `None` here.  I could imagine re-implementing `solana deploy` to use the upgradable loader in all cases if we can make it a superset of the existing loader

sdk/program/src/bpf_loader_upgradeable.rs
jackcmay: What distinction are you trying to make?

mvines: 1) It's an address, not a pubkey.  pubkey is one kind of address
2) `_pubkey`, or `_address` suffix seems unnecessary

sdk/program/src/loader_upgradeable_instruction.rs
jackcmay: Was originally only going to set it on upgrades, but yeah, more consistent.

sdk/program/src/loader_upgradeable_instruction.rs
jackcmay: Yeah, much better

sdk/program/src/loader_upgradeable_instruction.rs
jackcmay: Good point, will need an `initialize_source_account` instruction

jackcmay:
@mvines Invoking a program now requires that the invoker to include the code account as the first account in the instruction.  This isn't ideal for a developer's experience.  We could have the runtime attach the code account to the list of accounts passed to the loader and then the loader would eat it before passing the rest to the program's entrypoint.  Get's a little thorny because the loader would then have to a program signed CPI in order to do `CreateAccount` the code account.  Thoughts?

mvines:
> @mvines Invoking a program now requires that the invoker to include the code account as the first account in the instruction. This isn't ideal for a developer's experience. 

I'd definitely like to avoid this if at all possible.  

> Get's a little thorny because the loader would then have to a program signed CPI in order to do `CreateAccount` the code account. Thoughts?

Not sure I totally follow, not sure I need to though?

jackcmay:
Hiding the code account during invocation:
- Requires the runtime looking up the code account for the caller, either via derived address or by looking it up in the executable account's data.  This would be a special case keyed on if the loader is upgradeable
- Using a derived address for the code account would be ideal since it would allow us to totally hide the code account from the user (for both invocation, deploy, and upgrade), but that requires the bpf loader to `CreateAccount` since it needs to be the signer and this requires CPI from native programs.

sdk/program/src/bpf_loader_upgradeable.rs
```diff
@@ -0,0 +1,214 @@
...
+use bincode::serialized_size;
+
+crate::declare_id!("BPFLoaderUpgradeab1e11111111111111111111111");
+
+#[derive(Debug, Serialize, Deserialize, PartialEq, Clone, Copy, AbiExample)]
+pub enum UpgradableLoaderState {
+    /// Account is not initialized.
+    Uninitialized,
+    /// A source account.
+    Source,
```
mvines, line 26: nit: I think we need a better term than _source_ here.   eg, "Wait, this account holds source code?"

ELF account?

mvines:
> Hiding the code account during invocation:
> 
> * Requires the runtime looking up the code account for the caller, either via derived address or by looking it up in the executable account's data.  This would be a special case keyed on if the loader is upgradeable

Not too bad I guess.  Little sad that it breaks the current program loader model but oh well.

> * Using a derived address for the code account would be ideal since it would allow us to totally hide the code account from the user (for both invocation, deploy, and upgrade), but that requires the bpf loader to `CreateAccount` since it needs to be the signer and this requires CPI from native programs.

How would the code account be hidden here?   But requiring native CPI first mostly kills this option I think.

jackcmay:
I have the loader working with a derived `code` account via native CPI.  The code account address is derived from the executable and the loader now CPI's `CreateAccount` to create it (and to realloc it when that is supported).  Really cleans up a lot of stuff by not exposing the code account beyond the loader/helper functions.  It does change the construction of the the `loader1` accounts by inserting the code account into the chain (in accounts::load_executable_accounts`)

Changes are still local

jackcmay:
@mvines Pushed the latest wip changes, ignore the cli stuff, just an initial hacking.  Check out the bank.rs, message_processor.rs, and accounts.rs for runtime changes.  BPF loader also has most of the changes neccissary.

runtime/src/accounts.rs
```diff
@@ -260,6 +260,21 @@ impl Accounts {
 
             // add loader to chain
             let program_owner = program.owner;
+
+            if solana_sdk::bpf_loader_upgradeable::check_id(&program_owner) {
+                let (code_pubkey, _) = solana_sdk::pubkey::Pubkey::find_program_address(
```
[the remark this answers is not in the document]
jackcmay: We could get around this by adding the bump seeds to the executable account data, then it would just be a call to `create_program_address`

sdk/program/src/bpf_loader_upgradeable.rs
jackcmay: Would like to not be format specfic, how about "binary" account?

sdk/program/src/bpf_loader_upgradeable.rs
```diff
@@ -0,0 +1,218 @@
...
+crate::declare_id!("BPFLoaderUpgradeab1e11111111111111111111111");
+
+#[derive(Debug, Serialize, Deserialize, PartialEq, Clone, Copy, AbiExample)]
+pub enum UpgradeableLoaderState {
+    /// Account is not initialized.
+    Uninitialized,
+    /// A source account.
+    Source,
+    /// An executable account.
+    Executable {
```
mvines, line 28: This is the actual "Program" account right?  That is, what's given as a `program_id` by the users?
```suggestion
    /// A program account.
    Program {
```

sdk/program/src/bpf_loader_upgradeable.rs
```diff
@@ -0,0 +1,218 @@
...
+        /// Address of the code account.
+        code: Pubkey,
+    },
+    // A code account.
+    Code {
+        /// Slot that the program was last modified.
+        slot: u64,
+        /// Program's upgrade authority.
+        upgrade_authority: Option<Pubkey>,
+        // The raw executable code follows this serialized structure in the
```
mvines, line 38:
```suggestion
        // The program code follows this serialized structure in the
```

sdk/program/src/loader_upgradeable_instruction.rs
```diff
@@ -0,0 +1,104 @@
...
+    ///   1. [writable] The uninitialized ProgramData account.
+    ///   2. [writable] The uninitialized Program account.
+    ///   3. [writable] The Buffer account where the program data has been
+    ///      written.
+    ///   4. [] Rent sysvar.
+    ///   5. [] Clock sysvar.
+    ///   6. [] System program (`solana_sdk::system_program::id()`).
+    ///   7. [] The program's authority, optional, if omitted then the program
+    ///      will not be upgradeable.
+    DeployWithMaxDataLen {
```
mvines, line 69: Just call this `Deploy`?

programs/bpf/build.rs
```diff
@@ -37,17 +37,17 @@ fn rerun_if_changed(files: &[&str], directories: &[&str], excludes: &[&str]) {
 fn main() {
     let bpf_c = env::var("CARGO_FEATURE_BPF_C").is_ok();
     if bpf_c {
-        let install_dir =
-            "OUT_DIR=../target/".to_string() + &env::var("PROFILE").unwrap() + &"/bpf".to_string();
+        // let install_dir =
```
mvines, line 40: ?

programs/bpf_loader/src/syscalls.rs
```diff
@@ -1247,58 +1247,6 @@ impl<'a> SyscallObject<BPFError> for SyscallInvokeSignedC<'a> {
     }
 }
 
-fn verify_instruction<'a>(
```
mvines, line 1250: Seems like this refactoring could be a separate precursor PR

sdk/program/src/message.rs
```diff
@@ -280,6 +280,16 @@ impl Message {
         bincode::serialize(self).unwrap()
     }
 
+    pub fn program_id(&self, instruction_index: usize) -> Option<&Pubkey> {
```
mvines, line 283: These SDK changes could be in a precursor PR too

cli/src/cli.rs
```diff
@@ -42,7 +42,9 @@ use solana_faucet::faucet_mock::request_airdrop_transaction;
 use solana_rbpf::vm::{Config, Executable};
 use solana_remote_wallet::remote_wallet::RemoteWalletManager;
 use solana_sdk::{
+    account::Account,
```
mvines, line 45: Cli changes could be a separate PR too

sdk/program/src/loader_upgradeable_instruction.rs
jackcmay: Thought we were saving that for later when realloc was supported

programs/bpf/build.rs
jackcmay: Testing artifacts

sdk/program/src/loader_upgradeable_instruction.rs
mvines: When relloc is supported, I guess the max data len here essentially becomes ignored right?

programs/bpf_loader/src/syscalls.rs
jackcmay: Yeah, was thinking of breaking that and the BankClient changes out into separate ones, didn't seem too important but could at risk of slowing down getting this into 1.4

sdk/program/src/message.rs
jackcmay: Yup, same as above

cli/src/cli.rs
jackcmay: Yeah, nice to see the instruction definitions together with how a user would interact with them

programs/bpf_loader/src/syscalls.rs
mvines: Yeah I guess vote for small/quick-to-review PRs in general.   Details get lost in big PRs

jackcmay: Yup, did that a lot with the CPI stuff, tons a dependent little prs fell out of that.  Backporting a ton of little prs can create a long tail depending on how dependent they are on each other, lots of babysitting CI

sdk/program/src/loader_upgradeable_instruction.rs
jackcmay: Correct

jackcmay:
@mvines, dependent changes have been broken out and merged into master so hopefully this pr is now simpler to review.  Still have some "Upgrade" tests but looking forward to feedback.

sdk/program/src/loader_upgradeable_instruction.rs
mvines: I'm unclear about what you're thinking we'd change when realloc is supported.   A new `Deploy` instruction?  Or just keep `DeployWithMaxDataLen` but now `max_data_len` is just totally ignored?

sdk/program/src/loader_upgradeable_instruction.rs
```diff
@@ -0,0 +1,104 @@
...
+    ///   1. [] The Program account.
+    ///   2. [Writable] The Buffer account where the program data has been
+    ///      written.
+    ///   3. [writable] The spill account.
+    ///   4. [] Rent sysvar.
+    ///   5. [] Clock sysvar.
+    ///   6. [signer] The program's authority.
+    Upgrade,
+
+    /// Sets a new program authority that is allowed to update the program.  To
```
mvines, line 95:
```suggestion
    /// Sets a new authority that is allowed to upgrade the program.  To
```

sdk/program/src/loader_upgradeable_instruction.rs
```diff
@@ -0,0 +1,104 @@
...
+    /// permanently disable program updates omit the new authority.
+    ///
+    /// # Account references
+    ///   0. `[writable]` The ProgramData account to change the authority of.
+    ///   1. `[signer]` The current authority.
+    ///   2. `[]` The new authority, optional, if omitted then the program will
+    ///      not be upgradeable.
```
mvines, line 102:
```suggestion
    ///      no longer be upgradeable.
```

programs/bpf_loader/src/lib.rs
```diff
@@ -182,73 +282,336 @@ fn process_instruction_general(
...
+            let programdata_data_offset = UpgradeableLoaderState::programdata_data_offset()?;
+
+            if max_data_len < buffer_data_len {
+                log!(logger, "Max data length is too small to hold Buffer data");
+                return Err(InstructionError::AccountDataTooSmall);
+            }
+
+            // Create ProgramData account
+
+            let (derived_address, nonce) =
```
mvines, line 379:
```suggestion
            let (derived_address, bump_seed) =
```

programs/bpf_loader/src/lib.rs
```diff
@@ -182,73 +282,336 @@ fn process_instruction_general(
...
+                    1.max(
+                        rent.minimum_balance(UpgradeableLoaderState::programdata_len(
+                            max_data_len,
+                        )?),
+                    ),
+                    UpgradeableLoaderState::programdata_len(max_data_len)? as u64,
+                    program_id,
+                ),
+                &[payer, programdata, system],
+                &[&[program.unsigned_key().as_ref(), &[nonce]]], // TODO nonce+1
```
mvines, line 400: What does `// TODO nonce+1` mean?

programs/bpf_loader/src/lib.rs
```diff
@@ -182,73 +282,336 @@ fn process_instruction_general(
...
-                if program.data_len()? < offset + len {
-                    log!(
-                        logger,
-                        "Write overflow: {} < {}",
-                        program.data_len()?,
-                        offset + len
-                    );
-                    return Err(InstructionError::AccountDataTooSmall);
+                if authority.signer_key().is_none() {
+                    log!(logger, "Upgrade authority did not sign");
```
mvines, line 501: I like all the logging for the error conditions.  This is a good precedent for the native programs given how painful it currently is to decode custom error codes

sdk/program/src/loader_upgradeable_instruction.rs
```diff
@@ -0,0 +1,104 @@
...
+    ///
+    /// The Buffer account must contain sufficient lamports to fund the
+    /// ProgramData account to be rent-exempt, any additional lamports left over
+    /// will be transferred to the spill, account leaving the Buffer account
+    /// balance at zero.
+    ///
+    /// # Account references
+    ///   0. [writable] The ProgramData account.
+    ///   1. [] The Program account.
+    ///   2. [Writable] The Buffer account where the program data has been
```
jackcmay, line 87: Think the buffer account needs to be a signer during upgrade?

mvines: I _think_ it's unnecessary.  There could be a window between the time that the Upgrade instruction is signed vs. processed, that the owner of the Buffer account could modify the buffer.   But then why would the program authority sign an Upgrade instruction without knowing who the owner of the Buffer account was

cli/src/cli.rs
```diff
@@ -187,10 +189,24 @@ pub enum CliCommand {
...
         program_location: String,
-        address: Option<SignerIndex>,
+        buffer: Option<SignerIndex>,
         use_deprecated_loader: bool,
+        use_upgradeable_loader: bool,
+        upgrade_authority: Option<Pubkey>,
+        max_len: Option<usize>,
         allow_excessive_balance: bool,
     },
+    Upgrade {
```
mvines, line 199:
```suggestion
    ProgramUpgrade {
```

cli/src/cli.rs
```diff
@@ -187,10 +189,24 @@ pub enum CliCommand {
...
+        program_location: String,
+        program: Pubkey,
+        upgrade_authority: SignerIndex,
+        buffer: Option<SignerIndex>,
+    },
+    SetUpgradeAuthority {
```
mvines, line 205:
```suggestion
    SetProgramUpgradeAuthority {
```

cli/src/cli.rs
```diff
@@ -616,23 +632,77 @@ pub fn parse_command(
...
                     use_deprecated_loader: matches.is_present("use_deprecated_loader"),
+                    use_upgradeable_loader: matches.is_present("use_upgradeable_loader"),
+                    upgrade_authority,
+                    max_len,
                     allow_excessive_balance: matches.is_present("allow_excessive_balance"),
                 },
                 signers,
             })
         }
+        ("upgrade", Some(matches)) => {
```
mvines, line 658:
```suggestion
        ("upgrade-program", Some(matches)) => {
```

cli/src/cli.rs
```diff
@@ -616,23 +632,77 @@ pub fn parse_command(
...
+                command: CliCommand::Upgrade {
+                    program_location: matches.value_of("program_location").unwrap().to_string(),
+                    program,
+                    upgrade_authority,
+                    buffer,
+                },
+                signers,
+            })
+        }
+        ("set-upgrade-authority", Some(matches)) => {
```
mvines, line 685:
```suggestion
        ("set-program-upgrade-authority", Some(matches)) => {
```

cli/src/cli.rs
```diff
@@ -1168,16 +1238,194 @@ fn send_and_confirm_transactions_with_spinner<T: Signers>(
...
+        &program_data,
+        Some(|x| bpf_verifier::check(x, false)),
+        Config::default(),
+    )
+    .map_err(|err| CliError::DynamicProgramError(format!("ELF error: {}", err)))?;
+
+    Ok(program_data)
+}
+
+fn complete_partial_init(
```
mvines, line 1261: Unfortunately the program commands are in the general `cli.rs` instead of a separate file (for now I guess).  So while the commands are in `cli.rs`, I think `complete_partial_init()` needs a more program/loader specific name.

cli/src/cli.rs
```diff
@@ -1342,94 +1597,248 @@ fn do_process_deploy(
...
+    program_location: &str,
+    program: Pubkey,
+    upgrade_authority: SignerIndex,
+    buffer: Option<SignerIndex>,
+) -> ProcessResult {
+    const WORDS: usize = 12;
+    // Create ephemeral keypair to use for Buffer account, if not provided
+    let mnemonic = Mnemonic::new(MnemonicType::for_word_count(WORDS)?, Language::English);
+    let seed = Seed::new(&mnemonic, "");
+    let new_keypair = keypair_from_seed(seed.as_bytes())?;
```
mvines, line 1655: This is now duplicated between `process_deploy` and `progress_program_update()`, make a `fn`?

cli/src/cli.rs
```diff
@@ -1342,94 +1597,248 @@ fn do_process_deploy(
...
-        if num_required_signatures == 2 {
-            initial_transaction.try_sign(&signers, blockhash)?;
-        } else {
-            initial_transaction.try_sign(&[signers[0]], blockhash)?;
-        }
-        let result = rpc_client.send_and_confirm_transaction_with_spinner_and_config(
-            &initial_transaction,
-            config.commitment,
-            config.send_transaction_config,
+fn process_upgrade(
```
mvines, line 1643:
```suggestion
fn progress_program_update(
```

cli/src/cli.rs
```diff
@@ -1342,94 +1597,248 @@ fn do_process_deploy(
...
-            CliError::DynamicProgramError(format!("Program account allocation failed: {}", err))
-        })?;
     }
+    result
+}
 
-    let (blockhash, _, last_valid_slot) = rpc_client
-        .get_recent_blockhash_with_commitment(config.commitment)?
-        .value;
+fn do_process_upgrade(
```
mvines, line 1686:
```suggestion
fn do_process_program_upgrade(
```

cli/src/cli.rs
```diff
@@ -1342,94 +1597,248 @@ fn do_process_deploy(
...
+        upgrade_authority,
+    )?;
+
+    Ok(json!({
+        "programId": format!("{}", program),
+    })
+    .to_string())
+}
+
+fn process_set_upgrade_authority(
```
mvines, line 1796:
```suggestion
fn process_set_program_upgrade_authority(
```

sdk/program/src/loader_upgradeable_instruction.rs
jackcmay: Yeah, new `Deploy` instruction, I think that was your suggestion, I'm down for either.

programs/bpf_loader/src/lib.rs
jackcmay: I just removed that :-)

programs/bpf_loader/src/lib.rs
jackcmay: What do you think about providing even more information, like for example reporting the wrong balance or size values using `format!`?

mvines:
NB: when this is backported to v1.4, https://github.com/solana-labs/solana/pull/14147 will need to be backported as well

PR #13697, opened 2020-11-19
explorer: Add developer setting to enable customUrl param

jstarry:
#### Problem
It's difficult to keep the explorer pointed at a desired custom cluster node

#### Summary of Changes
Add a developer setting for enabling the "customUrl" url param.

Screen Shot 2020-11-19 at 9 36 46 PM

cc @mvines


bartosz-lipinski:
Looks awesome, this will be big help during development with custom nodes. What do you think about making some of that logic reusable library so external apps can have the same behavior for cluster/customUrl?

jstarry:
Yeah that'd be great. Maybe it could live in the sol-wallet-adapter?

PR #13699, opened 2020-11-19
Document RpcClient::get_account() gotcha

mvines:


CriesofCarrots:
I think that `ok_or_else` change was to return a user-readable error for non-existent, but I think it may have caused more trouble than it's worth 😕

mvines:
It can be fixed when we write the `async` verson of `RpcClient` :)

PR #13738, opened 2020-11-20
Add Bank feature queuing and use to separate inflation start and first rewards distribution

CriesofCarrots:
#### Problem
In current code, when inflation is activated the first rewards distribution happens immediately, with distributions based on the previous epoch. This is because of the order of operations at the epoch boundary: feature activations occur (which in the case of inflation, sets Inflation to non-zero), then rewards are distributed if Inflation is non-zero.

This feels weird. What we really want is for activation inflation to indicate the beginning of rewards *accrual*, which will be distributed at the end of the next epoch.

#### Summary of Changes
- Add a Bank mechanism for queuing feature activations to activate at the next epoch
- Add pico_inflation_rewards and full_inflation_rewards additional features
- When pico_inflation is activated, queue pico_inflation_rewards
- When full_inflation is activated, queue full_inflation_rewards
- Adjust `get_inflation_num_slots` since `get_inflation_start_slot` now aligns with taper (yay!)

### DIFF
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -38,7 +38,7 @@ use solana_sdk::{
     },
     epoch_info::EpochInfo,
     epoch_schedule::EpochSchedule,
-    feature,
+    feature::{self, Feature},
     feature_set::{self, FeatureSet},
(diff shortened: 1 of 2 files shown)


CriesofCarrots:
Not a need right now.

PR #13739, opened 2020-11-20
limits number of crds values returned when responding to pull requests

behzadnouri:
#### Problem
Crds values buffered when responding to pull-requests can be very large taking a lot of memory.

#### Summary of Changes
~~Added a hard-coded limit on the number of responses which may be buffered.~~
Added a limit for number of buffered crds values based on outbound data budget.

### DIFF
--- a/core/src/cluster_info.rs
+++ b/core/src/cluster_info.rs
@@ -109,7 +109,7 @@ const GOSSIP_PING_CACHE_CAPACITY: usize = 16384;
 const GOSSIP_PING_CACHE_TTL: Duration = Duration::from_secs(640);
 pub const DEFAULT_CONTACT_DEBUG_INTERVAL: u64 = 10_000;
 /// Limit number of crds values returned when responding to pull-requests.
-const PULL_REQUESTS_OUTPUT_SIZE_LIMIT: usize = 131_072;
+const PULL_REQUESTS_OUTPUT_SIZE_LIMIT: usize = 65_536;
 
 #[derive(Debug, PartialEq, Eq)]
 pub enum ClusterInfoError {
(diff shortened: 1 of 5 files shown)


core/src/crds_gossip_pull.rs
```diff
@@ -508,7 +511,10 @@ impl CrdsGossipPull {
                             Some(item.value.clone())
                         }
                     })
-                    .collect()
+                    .take(output_size_limit)
+                    .collect();
+                output_size_limit -= out.len();
```
carllin, line 516: Should we restructure this to early exit out of the iterator when `output_size_limit` hits zero to avoid doing the rest of the filters?

core/src/cluster_info.rs
```diff
@@ -108,6 +108,8 @@ const GOSSIP_PING_TOKEN_SIZE: usize = 32;
 const GOSSIP_PING_CACHE_CAPACITY: usize = 16384;
 const GOSSIP_PING_CACHE_TTL: Duration = Duration::from_secs(640);
 pub const DEFAULT_CONTACT_DEBUG_INTERVAL: u64 = 10_000;
+/// Limit number of crds values returned when responding to pull-requests.
+const PULL_REQUESTS_OUTPUT_SIZE_LIMIT: usize = 131_072;
```
carllin, line 112: Is this number based on the outbound budget/size of the network?

behzadnouri: This is based on max `gossip_pull_request-sent_requests` on current testnet and mainnet plus some margin. I realized it is a bit too high, so lowered it to 64k (2^16).

core/src/crds_gossip_pull.rs
behzadnouri: done

core/src/cluster_info.rs
carllin: ah ok, you think we should scale it based on the number of peers to accommodate network growth?

sakridge: We have a similar case in repair service with a dynamic limit:
https://github.com/solana-labs/solana/blob/5e2d38227fbf91320fcfdfff94b4444cbad3bfff/core/src/serve_repair.rs#L235

behzadnouri: updated the code to limit number of buffered crds values based on outbound data budget.

carllin (approved):

PR #13742, opened 2020-11-20
explorer: hotfix: customUrl effect is preventing redirects, add check

oJshua:
#### Problem
Developer customUrl flag is preventing redirects in certain situations.

#### Summary of Changes
Check if customUrl field is set before pushing to the history stack (preventing redirect). This *should* resolve situations where incoming links use deprecated URL patterns (e.g. /accounts/:address)


jstarry:
Thanks @oJshua!

PR #13750, opened 2020-11-22
processes pull-request callers only once per unique caller

behzadnouri:
#### Problem
process_pull_requests acquires a write lock on crds table to update
records timestamp for each pull-request caller:
https://github.com/solana-labs/solana/blob/3087c9049/core/src/crds_gossip_pull.rs#L287-L300
However, pull-requests overlap a lot in callers and this function ends
up doing a lot of redundant duplicate work.

#### Summary of Changes
This commit obtains unique callers before acquiring an exclusive lock on
crds table.

### DIFF
--- a/core/src/cluster_info.rs
+++ b/core/src/cluster_info.rs
@@ -2023,8 +2023,9 @@ impl ClusterInfo {
         feature_set: Option<&FeatureSet>,
     ) -> Packets {
         let mut time = Measure::start("handle_pull_requests");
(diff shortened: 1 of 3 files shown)


behzadnouri:
cluster_info_stats3 process_pull_requests_filter_current
Left chunk is this code, vs the current code on the right.

sakridge (approved):
nice, looks great

PR #13756, opened 2020-11-23
Fix Primordial stake check

sakridge:
#### Problem

`extraPrimordialStakes` is set as a scalar of number of extra staked nodes, but it checks for the array size of the variable to tell if the validators should not be staked `${#extraPrimordialStakes}`.

#### Summary of Changes

Change that to checking if the scalar equals 0.

Fixes #


t-nelson (approved):
Nice catch!  Thanks!

PR #13767, opened 2020-11-23
Fail stake Initialize if account data is the wrong size

CriesofCarrots:
#### Problem
The ability to create stake accounts that are larger than needed for StakeState is problematic due to the `rent_exempt_reserve`. For instance, if a user splits active stake into a much larger account, it presents the opportunity to magically deactivate stake up to the new rent_exempt_reserve without waiting the otherwise required cooldown period.

#### Summary of Changes
- Check account data on Initialize and fail the transaction if the size is incorrect. Technically, too-small accounts are already handled by the `StateMut::state()` method, but `!=` seemed more self-documenting than `>`


t-nelson (approved):
That was quick!  LGTM!

PR #13789, opened 2020-11-24
Validator CLI option to enable just-in-time compilation of BPF

Lichtso:
#### Problem
So far the JIT enable flag is hard coded to be disabled.

#### Summary of Changes
Adds and CLI option to the validator so that it can be manually controlled without rebuilding.

Fixes #

### DIFF
--- a/ledger/src/blockstore_processor.rs
+++ b/ledger/src/blockstore_processor.rs
@@ -343,8 +343,10 @@ pub fn process_blockstore(
         account_paths,
         &opts.frozen_accounts,
         opts.debug_keys.clone(),
-        Some(&crate::builtins::get(genesis_config.cluster_type)),
-        opts.bpf_jit,
+        Some(&crate::builtins::get(
+            genesis_config.cluster_type,
+            opts.bpf_jit,
+        )),
     );
     let bank0 = Arc::new(bank0);
     info!("processing ledger for slot 0...");
@@ -2832,7 +2834,7 @@ pub mod tests {
         genesis_config: &GenesisConfig,
         account_paths: Vec<PathBuf>,
     ) -> EpochSchedule {
-        let bank = Bank::new_with_paths(&genesis_config, account_paths, &[], None, None, false);
+        let bank = Bank::new_with_paths(&genesis_config, account_paths, &[], None, None);
         *bank.epoch_schedule()
     }
 
--- a/ledger/src/blockstore_processor.rs
+++ b/ledger/src/blockstore_processor.rs
@@ -311,7 +311,7 @@ pub type ProcessCallback = Arc<dyn Fn(&Bank) + Sync + Send>;
 
 #[derive(Default, Clone)]
 pub struct ProcessOptions {
-    pub bpf_jit_compilation: bool,
+    pub bpf_jit: bool,
     pub poh_verify: bool,
     pub full_leader_cache: bool,
     pub dev_halt_at_slot: Option<Slot>,
@@ -344,7 +344,7 @@ pub fn process_blockstore(
         &opts.frozen_accounts,
         opts.debug_keys.clone(),
         Some(&crate::builtins::get(genesis_config.cluster_type)),
-        opts.bpf_jit_compilation,
+        opts.bpf_jit,
     );
     let bank0 = Arc::new(bank0);
     info!("processing ledger for slot 0...");
--- a/ledger/src/blockstore_processor.rs
+++ b/ledger/src/blockstore_processor.rs
@@ -311,6 +311,7 @@ pub type ProcessCallback = Arc<dyn Fn(&Bank) + Sync + Send>;
 
 #[derive(Default, Clone)]
 pub struct ProcessOptions {
+    pub bpf_jit_compilation: bool,
     pub poh_verify: bool,
     pub full_leader_cache: bool,
     pub dev_halt_at_slot: Option<Slot>,
@@ -343,6 +344,7 @@ pub fn process_blockstore(
         &opts.frozen_accounts,
         opts.debug_keys.clone(),
         Some(&crate::builtins::get(genesis_config.cluster_type)),
+        opts.bpf_jit_compilation,
     );
     let bank0 = Arc::new(bank0);
     info!("processing ledger for slot 0...");
@@ -2830,7 +2832,7 @@ pub mod tests {
         genesis_config: &GenesisConfig,
         account_paths: Vec<PathBuf>,
     ) -> EpochSchedule {
-        let bank = Bank::new_with_paths(&genesis_config, account_paths, &[], None, None);
+        let bank = Bank::new_with_paths(&genesis_config, account_paths, &[], None, None, false);
         *bank.epoch_schedule()
     }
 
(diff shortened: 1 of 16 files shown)


validator/src/main.rs
```diff
@@ -1368,6 +1368,12 @@ pub fn main() {
                     "Mode to recovery the ledger db write ahead log."
                 ),
         )
+        .arg(
+            Arg::with_name("bpf_jit_compilation")
+                .long("enable-jit")
```
mvines, line 1373: minor naming nit:
```suggestion
            Arg::with_name("bpf_jit")
                .long("bpf-jit")
```

validator/src/main.rs
```diff
@@ -1510,6 +1516,7 @@ pub fn main() {
         poh_verify: !matches.is_present("skip_poh_verify"),
         debug_keys,
         contact_debug_interval,
+        bpf_jit_compilation: matches.is_present("bpf_jit_compilation"),
```
mvines, line 1519:
```suggestion
        bpf_jit: matches.is_present("bpf_jit"),
```

sdk/src/feature_set.rs
```diff
@@ -94,6 +94,10 @@ pub mod rewrite_stake {
     solana_sdk::declare_id!("6ap2eGy7wx5JmsWUmQ5sHwEWrFSDUxSti2k5Hbfv5BZG");
 }
 
+pub mod bpf_just_in_time_compilation {
+    solana_sdk::declare_id!("3bZhti6PMmsdkPxqPT8TPNzB3n9NYbKtnwqEpUDnFuPi");
```
mvines, line 98: Why do we need a feature for this?

Lichtso: The idea is that the cli flag is temporary for testing and that we will roll it out using a feature gate later. However, you are right that we don't need the feature right now, but it also does not hurt to have it already.

sakridge: Results should be bit-exact to the current code though, no?

mvines: I feel like we shouldn't need to feature gate this.  The BPF VM should behave the same with/without JIT, so it's a validator decision to enable/disable.  Initially the `--bpf-jit` argument would be required to enable, then once we're comfortable with it then that turns into a `--no-bpf-jit` argument to disable it

programs/bpf_loader/src/lib.rs
```diff
@@ -119,9 +119,7 @@ pub fn create_and_cache_executor(
         .map_err(|e| map_ebpf_error(invoke_context, e))?;
     executable.set_syscall_registry(syscall_registry);
 
-    if invoke_context.is_feature_active(&bpf_just_in_time_compilation::id())
-        && executable.jit_compile().is_err()
-    {
+    if invoke_context.use_bpf_jit() && executable.jit_compile().is_err() {
         return Err(BPFLoaderError::JustInTimeCompilationFailed.into());
```
mvines, line 123: I think we want to fallback to to the interpreted VM here.  Otherwise all JIT-enabled validators will fork from the rest

ledger/src/blockstore_processor.rs
```diff
@@ -344,6 +345,7 @@ pub fn process_blockstore(
         &opts.frozen_accounts,
         opts.debug_keys.clone(),
         Some(&crate::builtins::get(genesis_config.cluster_type)),
+        opts.bpf_jit,
```
mvines, line 348: I think we can avoid plumbing `bpf_jit` though the Bank and InvokeContext entirely.  

Pass `opts.bpf_jit` as a new parameter into `crate::builtins::get()`, then select either
https://github.com/solana-labs/solana/blob/008b56e535ae2d94906b0d90e78e256cd893b881/ledger/src/builtins.rs#L12
or a new `solana_bpf_loader_with_jit_program!()`

Lichtso: Don't the programs define which loader they need? In other words would the new loader work with the existing programs?

> Programs built for older loaders must use a matching entrypoint
> version.  An example is `bpf_loader_deprecated.rs` which requires
> `entrypoint_deprecated.rs`.

mvines: Ah sorry so I was suggesting that `solana_bpf_loader_program` and `solana_bpf_loader_with_jit_program` have the same program id, and only one is added to the builtins based on the state of the `bpf_jit` flag

programs/bpf_loader/src/lib.rs
Lichtso: Ok, so don't return an error but at least a log message?
Also, currently there is no way to check if an executable was compiled, so I would have to add that in RBPF, as there could be executables which can only be interpreted.

ledger/src/blockstore_processor.rs
Lichtso: I think I implemented it, but you should check if it is what you wanted.

programs/bpf_loader/src/lib.rs
mvines: > so don't return an error but at least a log message?

A metrics datapoint would be perfect

I think this part can be done as a follow-up PR though

programs/bpf_loader/src/lib.rs
```diff
@@ -153,11 +153,12 @@ pub fn create_vm<'a>(
     Ok(vm)
 }
 
-pub fn process_instruction(
+pub fn process_instruction_general(
```
mvines, line 156: Doesn't need to be `pub` anymore right?

programs/bpf_loader/src/lib.rs
```diff
@@ -214,6 +221,36 @@ pub fn process_instruction(
...
+    process_instruction_general(
+        program_id,
+        keyed_accounts,
+        instruction_data,
+        invoke_context,
+        false,
+    )
+}
+
+pub fn process_instruction_jit(
```
mvines, line 239: Doesn't need to be `pub` right?

mvines (approved):
Looks great, thanks!  I just have a couple minor nits but I think this is ready to land

programs/bpf_loader/src/lib.rs
Lichtso: Still needs it as it is used by a macro in a different file.

mvines: ah, yep. thanks

ledger/src/blockstore_processor.rs
jackcmay: @mvines Circling back on this, it's kinda funky to define two builtins in order to support a feature especially since we'll have to add two builtins for each new loader we add (like the upgradeable).  What do you think about passing feature set into the bank with an existing configuration (for example jit enabled or not) and then if a feature account enabled a specific feature that would override.  This would allow for future features to be communicated into the runtime from the command line

mvines: I think we should use a different mechanism than FeatureSet.   FeatureSet is supposed to be for things that would affect deterministic execution across all nodes in the cluster, and is currently based solely on on-chain state.

It's probably time to bundle these,
https://github.com/solana-labs/solana/blob/1d267eae6b82222027780280f8141710260c6123/runtime/src/bank.rs#L868-L870
into a `BankConfig` or `RuntimeConfig` struct that we can easily extend without an annoying plumbing effort on every new addition.  The JIT flag could then go in there?

jackcmay: Yeah, sgtm, a different mechanism is more appropriate.

PR #13790, opened 2020-11-24
Update master version to 1.5.0-pre.0 

mvines:
As a slight adjustment to our release versioning, the `-pre.X` suffix is retained through the stabilization process on testnet.  The pre-release suffix is dropped only when the release is finally suitable for use on mainnet-beta.  

This'll come into affect during the v1.5 stabilization.  However for illustration purposes, if v1.4 conformed to this scheme then the current `v1.4.10` release would be `v1.4.0-pre.10` and we would not ship `v1.4.0` until early December immediately before we upgrade mainnet-beta to v1.4

`./scripts/increment-cargo-version.sh` is now smart enough to figure out what to do without any arguments, provided that you run it from the correct branch.

This solves multiple problems:
* Avoids confusion from people who are not closely following along in Discord and wondering about why they should not use the latest v1.4 release on mainnet-beta
* Makes it clear that while on the beta channel, v1.4 is not yet stable and we might need to make breaking changes (from previous pre-releases)
* Makes it clear that the master branch is not stable


mvines:
Darn, needs rework.  The prerelease suffix is causing all kinds of trouble when building downstream projects

mvines:
☠️

PR #13795, opened 2020-11-25
caches vote-state de-serialized from vote accounts

behzadnouri:
#### Problem
Gossip and other places repeatedly de-serialize vote-state stored in
vote accounts. Ideally the first de-serialization should cache the
result.

#### Summary of Changes
This commit adds new VoteAccount type which lazily de-serializes
VoteState from Account data and caches the result internally.

Serialize and Deserialize traits are manually implemented to match
existing code. So, despite changes to frozen_abi, this commit should be
backward compatible.

### DIFF
--- a/runtime/src/vote_account.rs
+++ b/runtime/src/vote_account.rs
@@ -1,0 +1,181 @@
+use serde::de::{Deserialize, Deserializer};
+use serde::ser::{Serialize, Serializer};
+use solana_sdk::{account::Account, instruction::InstructionError};
+use solana_vote_program::vote_state::VoteState;
+use std::ops::Deref;
+use std::sync::{Arc, Once, RwLock, RwLockReadGuard};
+
+// The value here does not matter. It will be overwritten
+// at the first call to VoteAccount::vote_state().
+const INVALID_VOTE_STATE: Result<VoteState, InstructionError> =
+    Err(InstructionError::InvalidAccountData);
+
+#[derive(Clone, Debug, Default, PartialEq, AbiExample)]
+pub struct ArcVoteAccount(Arc<VoteAccount>);
+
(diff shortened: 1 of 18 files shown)


behzadnouri:
```bank-timestamp.get_timestamp_estimate_us```
bank-timestamp.get_timestamp_estimate_us_vote_account

```replay-loop-timing-stats.generate_new_bank_forks_elapsed```
replay-loop-timing-stats.generate_new_bank_forks_elapsed_vote_accounts

**Comment by sakridge** [comment]
> Left chunk is this code, vs current code on the right.
> `bank-timestamp.get_timestamp_estimate_us`
> bank-timestamp.get_timestamp_estimate_us_vote_account
> 
> `replay-loop-timing-stats.generate_new_bank_forks_elapsed`
> replay-loop-timing-stats.generate_new_bank_forks_elapsed_vote_accounts

That's great. I know from some previous profiling that vote state serialize/deserialize is a fairly significant chunk of time.

File: runtime/src/vote_account.rs [hunk 1]
```diff
@@ -0,0 +1,181 @@
...
+// at the first call to VoteAccount::vote_state().
+const INVALID_VOTE_STATE: Result<VoteState, InstructionError> =
+    Err(InstructionError::InvalidAccountData);
+
+#[derive(Clone, Debug, Default, PartialEq, AbiExample)]
+pub struct ArcVoteAccount(Arc<VoteAccount>);
+
+#[derive(Debug, AbiExample)]
+pub struct VoteAccount {
+    account: Account,
```
carllin, line 18: nice and hidden, this should keep it immutable 😃

carllin (approved):

sakridge (approved):

PR #13799, opened 2020-11-25
Nonce updates

sakridge:
#### Problem

Nonce cleanup

#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/accounts.rs
+++ b/runtime/src/accounts.rs
@@ -357,6 +357,23 @@ impl Accounts {
                         Err(e) => return (Err(e), hash_age_kind),
                     };
 
+                    // Update hash_age_kind with fee-subtracted accounts
+                    let hash_age_kind = hash_age_kind.map(|hash_age_kind| match hash_age_kind {
+                        HashAgeKind::Extant => HashAgeKind::Extant,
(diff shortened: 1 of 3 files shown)


ryoqun:
LGTM

t-nelson (approved):

runtime/src/accounts.rs
```diff
@@ -800,11 +841,16 @@ impl Accounts {
             }
             let (res, hash_age_kind) = &res[i];
             let maybe_nonce = match (res, hash_age_kind) {
-                (Ok(_), Some(HashAgeKind::DurableNonce(pubkey, acc))) => Some((pubkey, acc)),
+                (Ok(_), Some(HashAgeKind::DurableNonceFull(pubkey, acc, maybe_fee_account))) => {
+                    Some((pubkey, acc, maybe_fee_account))
```
ryoqun, line 845: til: it looks like this can be None. That's because we don't use `maybe_nonce` anymore if `res` is `Ok(_)`, considering we enabled `consistent recentblockhashes sysvar`.

Not that I'm asking you to do now with the follow up pr (#13868). just a note to solidify my understanding.

PR #13815, opened 2020-11-25
Use u64 behind the scenes for solana-tokens

CriesofCarrots:
#### Problem
Floating-point math inaccuracies can cause issues, particularly when working with 9-decimal numbers (as in the spl-feature-proposal use case).

#### Summary of Changes
- Store allocation amounts as u64
- Expect spl-token CSV to provide raw amounts (no decimals)
- Display amounts are still converted to floating point numbers, so there can be inaccuracies in the display. But all critical math (eg. transfer amounts, balance checks) uses integers.

Update of spl-feature-proposal to follow

### DIFF
--- a/tokens/src/commands.rs
+++ b/tokens/src/commands.rs
@@ -11,7 +11,9 @@ use indexmap::IndexMap;
 use indicatif::{ProgressBar, ProgressStyle};
 use pickledb::PickleDb;
 use serde::{Deserialize, Serialize};
-use solana_account_decoder::parse_token::{pubkey_from_spl_token_v2_0, spl_token_v2_0_pubkey};
+use solana_account_decoder::parse_token::{
+    pubkey_from_spl_token_v2_0, spl_token_v2_0_pubkey, token_amount_to_ui_amount,
+};
 use solana_client::{
     client_error::{ClientError, Result as ClientResult},
     rpc_client::RpcClient,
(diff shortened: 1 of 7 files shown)


tokens/src/commands.rs
```diff
@@ -817,16 +852,25 @@ pub fn test_process_distribute_stake_with_client(client: &RpcClient, sender_keyp
...
         .unwrap();
 
+    let expected_amount = sol_to_lamports(1000.0);
     let alice_pubkey = solana_sdk::pubkey::new_rand();
-    let allocation = Allocation {
-        recipient: alice_pubkey.to_string(),
-        amount: 1000.0,
-        lockup_date: "".to_string(),
-    };
+    // let allocation = Allocation {
```
mvines, line 857: Is this intention?

tokens/src/commands.rs
```diff
@@ -817,16 +852,25 @@ pub fn test_process_distribute_stake_with_client(client: &RpcClient, sender_keyp
...
-    wtr.serialize(&allocation).unwrap();
+    wtr.write_record(&["recipient", "amount", "lockup_date"])
+        .unwrap();
+    wtr.write_record(&[
+        alice_pubkey.to_string(),
+        lamports_to_sol(expected_amount).to_string(),
+        "".to_string(),
+    ])
+    .unwrap();
+    // wtr.serialize(&allocation).unwrap();
```
mvines, line 873: Is this intention?

mvines (approved):
lgtm overall!  Just those comments that maybe should be removed?

tokens/src/commands.rs
CriesofCarrots: Nope! Sec...

PR #13822, opened 2020-11-26
hidapi 1.2.4 is broken, lock version to 1.2.3

mvines:


CriesofCarrots:
In what way is it broken?

mvines:
macOS build blows up, breaking `cargo install spl-token-cli` and `cargo install solana-cli`.  We don't see the problem in the monorepo because Cargo.lock file specifies hidapi 1.2.3

mvines:
oh nice, hidapi 1.2.4 has been yanked

PR #13827, opened 2020-11-26
Prevent cleanup of zero lamport accounts past max_root

carllin:
#### Problem
`clean_accounts(max_root)`, currently can clean up zero-lamport account `A` in `max_root2 > max_root`, if `max_root2` is set  as a root before/during the `clean_accounts()` execution. 

For this to happen the zero-lamport account `A` in `max_root2` has to be in an AccountStorageEntry with only other zero-lamport accounts that also meet the zero-lamport delete qualifications. If this rare case does happen, it can lead to snapshot hash mismatch.

#### Summary of Changes
Extend `roots_and_ref_count()` to take a max root
Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -779,7 +779,8 @@ impl AccountsDB {
                         if account_info.lamports == 0 {
                             purges.insert(
                                 *pubkey,
-                                self.accounts_index.roots_and_ref_count(&locked_entry),
+                                self.accounts_index
+                                    .roots_and_ref_count(&locked_entry, max_clean_root),
                             );
                         }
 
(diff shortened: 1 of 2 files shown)


runtime/src/accounts_db.rs
```diff
@@ -5610,4 +5611,26 @@ pub mod tests {
...
+        db.store(0, &[(&account_key, &zero_lamport_account)]);
+        db.store(1, &[(&account_key, &zero_lamport_account)]);
+        db.add_root(0);
+        db.add_root(1);
+
+        // Only clean zero lamport accounts up to slot 0
+        db.clean_accounts(Some(0));
+
+        // Should still be able to find zero lamport account in slot 1
+        assert_eq!(
```
carllin, line 5631: Test fails here without the fix

sakridge (approved):
good catch

PR #13840, opened 2020-11-27
Fix undefined symbol "custom panic" in BPF test programs

Lichtso:
#### Problem
#13455 introduced undefined symbols (`custom_panic`) in the compiled BPF test programs.
These were caught by the JIT compiler, but not the interpreter because they are dead code.

#### Summary of Changes
Use the `entrypoint!` macro everywhere to implement the missing symbol.

Fixes #


jackcmay:
So the only issue is that some of the tests didn't define `custom_panic`? I want to keep these tests that have a raw entrypoint, I don't want to change these tests to the full entrypoint.  We can define a new macro that defines a default custom_panic and call that both in the entrypoint macro and directly in these tests.  Or just define entrypoint in these tests

jackcmay (approved):

PR #13841, opened 2020-11-28
Move instruction recording into MessageProcesor

mvines:
Inner instruction recording was way out in `programs/bpf_loader/src/syscalls.rs` when it really belongs in `runtime/src/message_processor.rs`.  This removes duplication across syscall implementations, avoids the need to recompile inner instructions, and ensures the inner instructions are fully validated before recording them.

### DIFF
--- a/runtime/src/message_processor.rs
+++ b/runtime/src/message_processor.rs
@@ -8,7 +8,7 @@ use solana_sdk::{
     account::Account,
     clock::Epoch,
     feature_set::{instructions_sysvar_enabled, FeatureSet},
-    instruction::{CompiledInstruction, Instruction, InstructionError},
+    instruction::{CompiledInstruction, InstructionError},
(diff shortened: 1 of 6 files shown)


runtime/src/message_processor.rs
```diff
@@ -500,6 +500,8 @@ impl MessageProcessor {
             // Verify the calling program hasn't misbehaved
             invoke_context.verify_and_update(message, instruction, accounts)?;
 
+            invoke_context.record_instruction(&instruction);
```
jstarry, line 503: Does the message in this function have the same account keys and ordering as the original message?

mvines: Not sure but I don't think it matters because at this point the instruction is compiled

mvines:
Probably won't continue down this path.  The lesser change seems to be to move `record_instruction()` after `translate_accounts()`, to ensure the program_id is checked

PR #13842, opened 2020-11-28
solana-validator: Add --maximum-local-snapshot-age argument

mvines:
If the validator has a snapshot that's less than 500 slots away from the highest snapshot available for download, the validator now uses its local snapshot.  500 slots is ~ 3.3 minutes, and I'm seeing a mainnet-beta snapshot download take about that time when not sitting on the GCP backbone.    So it's better to use that 3.3 minutes trying to catchup with what you already have.   

`--maximum-local-snapshot-age 0` will disable this functionality and restore the previous behaviour of unconditional snapshot downloading.

### DIFF
--- a/validator/src/main.rs
+++ b/validator/src/main.rs
@@ -640,6 +640,7 @@ fn rpc_bootstrap(
     bootstrap_config: RpcBootstrapConfig,
     no_port_check: bool,
     use_progress_bar: bool,
+    maximum_local_snapshot_age: Slot,
 ) {
     if !no_port_check {
         verify_reachable_ports(&node, cluster_entrypoint, &validator_config);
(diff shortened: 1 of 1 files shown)


sakridge:
> If the validator has a snapshot that's less than 500 slots away from the highest snapshot available for download, the validator now uses its local snapshot. 500 slots is ~ 3.3 minutes, and I'm seeing a mainnet-beta snapshot download take about that time when not sitting on the GCP backbone. So it's better to use that 3.3 minutes trying to catchup with what you already have.
> 
> `--maximum-local-snapshot-age 0` will disable this functionality and restore the previous behaviour of unconditional snapshot downloading.

I'm not sure the 500 slot math works out, but it's close enough. Local slots will be replayed at much faster than real time and catchup from the network should also be faster than real time. Need to compare that catchup time versus the download time, not the wall clock time of the slot.

mvines:
500 was just a first attempt at a default, I didn't want to make it too large to start with given the current default is effectively 0.   It's a little clunky how we also wait for snapshot hashes from trusted validators to show up on gossip before deciding not to use them if the local snapshot is new enough.   It would be better if the node just checked the current rooted slot over RPC, and based on that decided to continue along with it's local snapshot -- that was a larger refactoring than I wanted to take on right now though

PR #13848, opened 2020-11-29
Add extra checks to verify_and_update

jackcmay:
#### Problem

verify_and_update not checking all cases

#### Summary of Changes

Add more checks

Fixes #

### DIFF
--- a/runtime/src/message_processor.rs
+++ b/runtime/src/message_processor.rs
@@ -612,23 +612,25 @@ impl MessageProcessor {
         // Verify the per-account instruction results
         let (mut pre_sum, mut post_sum) = (0_u128, 0_u128);
         let mut work = |_unique_index: usize, account_index: usize| {
-            let key = &message.account_keys[account_index];
-            let account = &accounts[account_index];
-            // Find the matching PreAccount
-            for pre_account in pre_accounts.iter_mut() {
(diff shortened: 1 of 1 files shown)


runtime/src/message_processor.rs
```diff
@@ -634,6 +634,7 @@ impl MessageProcessor {
             Err(InstructionError::MissingAccount)
         };
         instruction.visit_each_account(&mut work)?;
+        work(0, instruction.program_id_index as usize)?;
```
jstarry, line 637: You're missing a call to work inside `create_pre_accounts` as well which will ensure that this checks the program id against the correct pre account.

Also, `unique_index` needs to be unique across all accounts. Right now, it reuses `0`.

Lastly, `unique_index` should be unique across all accounts, you're not accounting for a case where the program account itself exists in the account inputs.

jstarry: Never mind, I was looking at `verify`

jackcmay: - The pre accounts already has the program id since it must be passed in the outer instruction's meta
- unique_index isn't used by `verify_and_update`

jstarry: Agreed, thanks for confirming

jstarry (approved):

PR #13853, opened 2020-11-30
add zsh trouble shooting note to ledger docs

tomjohn1028:
#### Problem

First use of the Ledger wallet with the CLI running in **zsh** following the docs will error. There is great troubleshooting information but it may be missed on first glance if stepping through the documenation.
 
#### Summary of Changes

Added a foot note below the first code example to call attention to the troubleshooting documentation for **zsh** users.


mvines:
Thanks!

PR #13855, opened 2020-11-30
Explorer: add Raw instruction data to parsed instructions

### LINKED ISSUE #12758: Explorer: Raw instruction data is not available for parsed instructions
#### Problem
The explorer serves two main purposes. Transparency to end users and introspection for developers. The explorer has special handling for parsed instructions (stake, token, memo, etc) which is targeted to end users, but developers may wish to view the raw instruction data (in hex) for debugging

#### Proposed Solution
- Fetch base64 format of transaction when "Raw" toggle is selected for a transaction instruction

kevzettler:
#### Problem
The transaction Instruction UI  does not contain the raw instruction data. Users have requested access to the raw instruction data for debugging purposes. The current payload delivered to the UI only has a JSON representation of the instruction.

This issue was much deeper than expected. The transaction detail page is populated with a RPC call to `getParsedConfirmedTransaction` this RPC method is not documented on the [JSON RPC API](https://docs.solana.com/developing/clients/jsonrpc-api#json-rpc-api-reference) `getParsedConfirmedTransaction` only has a "parsed" JSON representation of the raw transaction instruction. I assume this was developed as an optimization for network bandwidth or something. 

The alternative RPC method `getConfirmedTransaction` includes the raw source data. There are[ some strange code paths around this feature](https://github.com/solana-labs/solana/blob/master/explorer/src/components/instruction/InstructionCard.tsx#L54-L62) There appears to be branching logic and display components for `RawParsedDetails` and `RawDetails`.  It's not clear to me when or if ever `RawDetails` gets rendered.  `RawParsedDetails` does not really render 'raw' details because the 'parsed' RPC payload never includes them. 

#### Summary of Changes
When the user clicks the "raw" button a request is made for the raw transaction details. This unfortunately fetches the entire raw transaction payload with all instructions. Luckily It is cached and all instructions now have their raw data sets.  Ideally I think a specific RPC method to fetch the raw TransactionInstruction records would be an optimization here. 

Adds a fetch action to the click event of the "raw" button on the
instruction UI. adds a `fetchRawTransaction` hook that is passed down to the
instruction UI components. Adds new `rawFetchTrigger` and `raw`
props passed to the instruction card components.

Fixes #12758

solana-raw

### DIFF
--- a/explorer/src/components/instruction/InstructionCard.tsx
+++ b/explorer/src/components/instruction/InstructionCard.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useContext } from "react";
 import {
   TransactionInstruction,
   SignatureResult,
@@ -6,6 +6,11 @@ import {
 } from "@solana/web3.js";
 import { RawDetails } from "./RawDetails";
 import { RawParsedDetails } from "./RawParsedDetails";
+import { SignatureContext } from "../../pages/TransactionDetailsPage";
+import {
+  useTransactionDetails,
+  useFetchRawTransaction,
+} from "providers/transactions/details";
 
 type InstructionProps = {
   title: string;
@@ -26,6 +31,22 @@ export function InstructionCard({
 }: InstructionProps) {
   const [resultClass] = ixResult(result, index);
   const [showRaw, setShowRaw] = React.useState(defaultRaw || false);
+  const signature = useContext(SignatureContext);
+  const details = useTransactionDetails(signature);
+  let raw: TransactionInstruction | undefined = undefined;
+  if (details) {
+    raw = details?.data?.raw?.transaction.instructions[index];
+  }
+  const fetchRaw = useFetchRawTransaction();
+  const fetchRawTrigger = () => fetchRaw(signature);
+
+  const rawClickHandler = () => {
+    if (!defaultRaw && !showRaw && !raw) {
+      fetchRawTrigger();
+    }
+
+    return setShowRaw((r) => !r);
+  };
 
   return (
     <div className="card">
@@ -42,7 +63,7 @@ export function InstructionCard({
           className={`btn btn-sm d-flex ${
             showRaw ? "btn-black active" : "btn-white"
           }`}
-          onClick={() => setShowRaw((r) => !r)}
+          onClick={rawClickHandler}
         >
           <span className="fe fe-code mr-1"></span>
           Raw
@@ -53,7 +74,7 @@ export function InstructionCard({
           <tbody className="list">
             {showRaw ? (
               "parsed" in ix ? (
-                <RawParsedDetails ix={ix} />
+                <RawParsedDetails ix={ix} raw={raw} />
               ) : (
                 <RawDetails ix={ix} />
               )
--- a/explorer/src/components/instruction/RawParsedDetails.tsx
+++ b/explorer/src/components/instruction/RawParsedDetails.tsx
@@ -1,8 +1,21 @@
 import React from "react";
-import { ParsedInstruction } from "@solana/web3.js";
+import { ParsedInstruction, TransactionInstruction } from "@solana/web3.js";
 import { Address } from "components/common/Address";
+import { wrap } from "utils";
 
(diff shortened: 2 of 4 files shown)


explorer/src/components/instruction/InstructionCard.tsx
```diff
@@ -23,10 +25,20 @@ export function InstructionCard({
   index,
   ix,
   defaultRaw,
+  fetchRawTrigger,
```
oJshua, line 28: Rather than passing a fetch trigger down, perhaps we could just set up useTransactionDetails and useFetchRawTransaction inside the InstructionCard itself. This would require passing down the signature either through props or establishing it in context (up to you, but props would be sufficient).

oJshua (dismissed):
Thanks for the contribution here; this is much needed! I just have one comment for a possible refactor.

kevzettler:
Hi @oJshua, Thanks for your review. I have revisited this and changed the data flow to use context. I used context because it was a huge chore to prop drill through all the `/components/instruction` files. Either way the functionality and value is the same to the end user.

oJshua (approved):
Excellent! This is perfect @kevzettler. Thanks again!

jstarry:
Thanks for the contribution @kevzettler!

> The transaction Instruction UI does not contain the raw instruction data. Users have requested access to the raw instruction data for debugging purposes. The current payload delivered to the UI only has a JSON representation of the instruction.

The transaction instruction UI does actually contain raw instruction data for instructions that our API does not have parsing support for. They look like this:
Screen Shot 2020-12-01 at 2 26 24 PM
http://explorer.solana.com/tx/5oqzcYbgqu5PDEHhXLNvtiSPLpmfmUfWJqcBRshgsTq7qyoGqaLKq4JrGHG1y5tMCaK52izsze9mNowd7jyjZjzG

---

> The transaction detail page is populated with a RPC call to getParsedConfirmedTransaction this RPC method is not documented on the JSON RPC API getParsedConfirmedTransaction only has a "parsed" JSON representation of the raw transaction instruction. I assume this was developed as an optimization for network bandwidth or something.

Close, but not quite. There is no undocumented `getParsedConfirmedTransaction` rpc method. If you take a look at the web3 sdk, the `getParsedConfirmedTransaction` sdk method is calling `getConfirmedTransaction` with the `jsonParsed` encoding.
https://github.com/solana-labs/solana/blob/6705b5a98c076ac08f3991bb8a6f9fcb280bf51e/web3.js/src/connection.js#L2591-L2597

Here's a snippet from the docs that describe this encoding param:
> <string> - (optional) encoding for the returned Transaction, either "json", "jsonParsed", "base58" (slow), or "base64". If parameter not provided, the default encoding is JSON.

https://docs.solana.com/developing/clients/jsonrpc-api#getconfirmedtransaction

---

> There are some strange code paths around this feature. There appears to be branching logic and display components for RawParsedDetails and RawDetails. It's not clear to me when or if ever RawDetails gets rendered. RawParsedDetails does not really render 'raw' details because the 'parsed' RPC payload never includes them.

Right, when `parsed` isn't returned (for instructions that the API doesn't know how to parse), this code path will be hit (which is what renders the example instruction I shared above)

https://github.com/solana-labs/solana/blob/6705b5a98c076ac08f3991bb8a6f9fcb280bf51e/explorer/src/components/instruction/InstructionCard.tsx#L75-L80

---

> When the user clicks the "raw" button a request is made for the raw transaction details. This unfortunately fetches the entire raw transaction payload with all instructions. Luckily It is cached and all instructions now have their raw data sets. Ideally I think a specific RPC method to fetch the raw TransactionInstruction records would be an optimization here.

This is totally fine. Transactions aren't that large and it's not unreasonable to think that a user will want to look at other raw instructions.

---

Code looks great but I would have liked to see reuse of / modifications to the existing `RawDetails` component to see a listing of account inputs and whether they are marked readonly / signer like we do for the existing raw instructions

Also, it would be nice if the account inputs section didn't switch to loading / failed when the raw transaction is fetched.


ezgif-6-10ac816d18fb

PR #13859, opened 2020-11-30
Add `flatten` to logs_subscribe commitment

joncinque:
#### Problem

Currently, commitment for `logs_subscribe` must be specified using the
non-flattened form in JSON, ie `{"commitment":{"commitment":"max}}`.

#### Summary of Changes

Bring commitment in line with documentation by specifying `flatten`

Fixes #


mvines:
thanks

PR #13869, opened 2020-11-30
Strengthen EpochSlots sanitization

mvines:


core/src/epoch_slots.rs
```diff
@@ -132,7 +141,7 @@ impl Uncompressed {
             if *s < self.first_slot {
                 return i;
             }
-            if *s - self.first_slot >= self.slots.capacity() {
+            if *s - self.first_slot >= self.slots.len() {
```
mvines, line 144: `Uncompressed::new()` will always create a bitvec with capacity equal to len, but it's much more correct to use len here

sakridge (approved):

PR #13871, opened 2020-11-30
Add msg! macro for program logging, deprecate info! macro

mvines:
This PR deprecates the `info!`  macro for program logging.  It's confusing because there are no other supported logging levels, and `info!` here does not map to the RUST_LOG validator "info" log level. 

Instead add a `msg!` macro that implements the same as `info!`, also provide a third form for generalized formatting.

This PR is currently just the `sdk/program` changes.  If this looks good I'll roll it out to the rest of the monorepo

### DIFF
--- a/sdk/program/src/log.rs
+++ b/sdk/program/src/log.rs
@@ -2,11 +2,8 @@
 
 use crate::account_info::AccountInfo;
 
-/// Prints a string
-/// There are two forms and are fast
-/// 1. Single string
-/// 2. 5 integers
 #[macro_export]
+#[deprecated(since = "1.4.14", note = "use `msg` macro instead")]
 macro_rules! info {
     ($msg:expr) => {
         $crate::log::sol_log($msg)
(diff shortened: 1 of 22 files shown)


sdk/program/src/log.rs
```diff
@@ -2,11 +2,11 @@
...
 use crate::account_info::AccountInfo;
 
-/// Prints a string
-/// There are two forms and are fast
-/// 1. Single string
-/// 2. 5 integers
 #[macro_export]
+#[deprecated(
+    since = "1.4.14",
+    note = "use msg macro instead"
```
jackcmay, line 8: "use `msg` macro instead"

sdk/program/src/log.rs
```diff
@@ -19,12 +19,37 @@ macro_rules! info {
...
+/// There are two fast forms:
+/// 1. Single string: `msg!("hi")`
+/// 2. 5 integers: `msg!(1, 2, 3, 4, 5)`
+///
+/// The third form is more generic and incurs a very large runtime overhead so it should be used
+/// with care:
+/// 3. Generalized format string: `msg("Hello: 1, 2, {}", "World", 3)`
+///
+#[macro_export]
+macro_rules! msg {
```
jackcmay, line 36: Does the implementaiton of `msg` differ from `info`, maybe just have `info` call `msg` to avoid duplication

mvines: Implementation doesn't differ, not sure there's really much to de-dup though and with duplication I'm 100% sure there's no backwards compatibility issue

sdk/program/src/log.rs
```diff
@@ -19,12 +19,37 @@ macro_rules! info {
...
+
+/// Print a message to the log
+///
+/// There are two fast forms:
+/// 1. Single string: `msg!("hi")`
+/// 2. 5 integers: `msg!(1, 2, 3, 4, 5)`
+///
+/// The third form is more generic and incurs a very large runtime overhead so it should be used
+/// with care:
+/// 3. Generalized format string: `msg("Hello: 1, 2, {}", "World", 3)`
```
joncinque, line 33: Should this comment instead be: `msg!("Hello {}: 1, 2, {}", "World", 3)` ?

mvines: Yep, thanks

PR #13879, opened 2020-12-01
solana-tokens: optimize PickleDb dumps

CriesofCarrots:
#### Problem
Currently, `solana-tokens` dumps PickleDb database information on every change, including after every distribution send. This can take a long time when there are a lot of transactions, which can disrupt transaction confirmation: eg. for a distribution set of ~500 transactions, the db dumps take long enough that the earliest transaction signatures are no longer in the node's status cache.

#### Summary of Changes
- Change PickleDb configuration to DumpUponRequest. Dump changes once after send loop and again after transaction confirmation loop. In the case of a mid-loop error, all intermediate changes should be written automatically, since PickleDb's drop implementation ensures dumping on destruction of the object.
- Add Ctrl-C handler to ensure all db changes are written on SIGINT and SIGTERM, even if mid-loop

### DIFF
--- a/tokens/src/commands.rs
+++ b/tokens/src/commands.rs
@@ -89,6 +89,8 @@ impl From<Vec<FundingSource>> for FundingSources {
     }
 }
 
+type StakeExtras = Vec<(Keypair, Option<DateTime<Utc>>)>;
+
 #[derive(thiserror::Error, Debug)]
 pub enum Error {
     #[error("I/O error")]
@@ -245,17 +247,16 @@ fn distribution_instructions(
     instructions
 }
 
-fn distribute_allocations(
+fn build_messages(
     client: &RpcClient,
     db: &mut PickleDb,
     allocations: &[Allocation],
     args: &DistributeTokensArgs,
     exit: Arc<AtomicBool>,
+    messages: &mut Vec<Message>,
+    stake_extras: &mut StakeExtras,
+    created_accounts: &mut u64,
 ) -> Result<(), Error> {
-    type StakeExtras = Vec<(Keypair, Option<DateTime<Utc>>)>;
-    let mut messages: Vec<Message> = vec![];
-    let mut stake_extras: StakeExtras = vec![];
-    let mut created_accounts = 0;
     for allocation in allocations.iter() {
         if exit.load(Ordering::SeqCst) {
             db.dump()?;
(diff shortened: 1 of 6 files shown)


CriesofCarrots:
@garious would appreciate your thoughts/concerns on this change. I'm trying to avoid needing `RpcClient::get_signature_statuses_with_history()`.

garious:
Why not pass db into `process_allocations` so that the handler can dump the db directly, without the AtomicBool indirection?

CriesofCarrots:
> Why not pass db into `process_allocations` so that the handler can dump the db directly, without the AtomicBool indirection?

I did actually start down that path, but would need to wrap db in `Arc` and some kind of lock to pass it into the handler; it looked like more churn in the db module than I was comfortable with. But I can work up that option, if you think it would be significantly better.

garious:
Probably better with the AtomicBool. How about some low-level tests for the new branches?

CriesofCarrots:
> How about some low-level tests for the new branches?

Yep! Adding now

CriesofCarrots:
@garious sorry for the delay. I ended up needing to split off helper functions to be able to target all the `dump` calls. There's a lot more that should probably be done in the way of unit-testing commands.rs, but I decided to stick within the scope of this pr.
Review probably easiest by commit.

tokens/src/commands.rs
```diff
@@ -296,20 +311,26 @@ fn distribute_allocations(
...
+    exit: Arc<AtomicBool>,
+    messages: Vec<Message>,
+    stake_extras: StakeExtras,
+) -> Result<(), Error> {
     for ((allocation, message), (new_stake_account_keypair, lockup_date)) in
         allocations.iter().zip(messages).zip(stake_extras)
     {
+        if exit.load(Ordering::SeqCst) {
+            db.dump()?;
+            #[cfg(not(test))]
```
garious, line 331: Why not return an error instead?

CriesofCarrots: Oh! That does seem nice.

CriesofCarrots: Updated

garious (approved):
Nice work, thank you!

CriesofCarrots:
CI failure because I neglected to rebase on: https://github.com/solana-labs/solana/pull/13896

PR #13914, opened 2020-12-02
Translate data length and owner as writable

jackcmay:
#### Problem

Account owner and data length may be updated by the runtime during invoke but the memory is not checked for rw privileges

#### Summary of Changes

Check for rw privileges

Fixes #

### DIFF
--- a/programs/bpf/Cargo.lock
+++ b/programs/bpf/Cargo.lock
@@ -2068,6 +2068,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "solana-bpf-rust-ro-modify"
+version = "1.5.0"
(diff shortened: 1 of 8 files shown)


bennofs:
Why does `translate_type_mut` even take an `AccessType` argument, shouldn't that always be `Store` for the `_mut` variant?

jackcmay:
@bennofs Good point, that should definitely be cleaned up, there might be some rust funkiness that led to it, will address in a cleanup pr.

jackcmay:
@bennofs @mrmaxmeier How about this: https://github.com/solana-labs/solana/pull/13921

PR #13919, opened 2020-12-02
samples repair peers using WeightedIndex

behzadnouri:
#### Problem
To output one random sample, weighted_best generates n random numbers:
https://github.com/solana-labs/solana/blob/f751a5d4e/core/src/weighted_shuffle.rs#L38-L63
WeightedIndex does so with only one random number:
https://github.com/rust-random/rand/blob/eb02f0e46/src/distributions/weighted_index.rs#L223-L240
Additionally, if the index is already constructed, it only does a total
of O(log(n)) amount of work; which can be achieved if RepairCache,
caches the weighted index:
https://github.com/solana-labs/solana/blob/f751a5d4e/core/src/serve_repair.rs#L83

Also, the repair-peers code can be reorganized to have fewer redundant
unlock-then-lock code.

#### Summary of Changes
* Replaced `weighted_best` with `WeightedIndex`.
* Reorganized some of the locking code when selecting repair peer.
* Tests show ~20% improvement in timings and less variations.

### DIFF
--- a/core/src/cluster_info.rs
+++ b/core/src/cluster_info.rs
@@ -1170,13 +1170,15 @@ impl ClusterInfo {
 
     /// all validators that have a valid tvu port and are on the same `shred_version`.
     pub fn tvu_peers(&self) -> Vec<ContactInfo> {
+        let self_pubkey = self.id();
+        let self_shred_version = self.my_shred_version();
         self.time_gossip_read_lock("tvu_peers", &self.stats.tvu_peers)
             .crds
             .get_nodes_contact_info()
-            .filter(|x| {
-                ContactInfo::is_valid_address(&x.tvu)
-                    && x.id != self.id()
-                    && x.shred_version == self.my_shred_version()
+            .filter(|node| {
+                node.id != self_pubkey
+                    && node.shred_version == self_shred_version
+                    && ContactInfo::is_valid_address(&node.tvu)
             })
             .cloned()
             .collect()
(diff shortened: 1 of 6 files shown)


core/src/cluster_slots.rs
```diff
@@ -106,28 +106,30 @@ impl ClusterSlots {
...
+                .map(|peer| {
+                    validator_stakes
+                        .get(&peer.id)
+                        .map(|node| node.total_stake)
+                        .unwrap_or(0)
+                        + 1
+                })
+                .collect()
+        };
+        let slot_peers = match self.lookup(slot) {
```
carllin, line 123: heh unrelated to this PR, but I'm realizing that this `lookup` is probably not that useful, especially since we cache the value the first time time we compute weights for this slot. Because repair usually runs on missing shreds in the middle of a slot, not many people will have the slot completed, so the `lookup()` here will return a `slot_peers` that's mostly empty.

carllin (approved):
Cool, I think the `WeightedIndex` was introduced to `rand` recently, a welcome improvement on our existing logic!

core/src/cluster_info.rs
```diff
@@ -1200,22 +1202,24 @@ impl ClusterInfo {
...
-                        .unwrap_or_else(|| /* fallback to legacy behavior */ true)
-                    }
-            })
-            .collect();
+        // self.tvu_peers() already filters on:
+        //   node.id != self.id() &&
+        //     node.shred_verion == self.my_shred_version()
+        let nodes = {
+            let gossip = self.gossip.read().unwrap();
+            self.tvu_peers()
```
jstarry, line 1210: @behzadnouri this can cause a deadlock on machines with write biased rwlocks if a writer gets queued for the rwlock before the read lock is acquired inside `self.tvu_peers()`.

carllin: hehe good catch! glad that local cluster debugging also reaped other benefits 🎉

behzadnouri: Thanks for catching this. Sent out https://github.com/solana-labs/solana/pull/13973

PR #13924, opened 2020-12-02
Cap number of program address seeds

jackcmay:
#### Problem

`create_program_address` has no cap on the number of seeds that can be passed which can lead to a large hash compute load

#### Summary of Changes

Cap the maximum number of seeds to a reasonable amount

Fixes #

### DIFF
--- a/programs/bpf_loader/src/syscalls.rs
+++ b/programs/bpf_loader/src/syscalls.rs
@@ -24,7 +24,7 @@ use solana_sdk::{
     message::Message,
     process_instruction::{stable_log, ComputeMeter, InvokeContext, Logger},
     program_error::ProgramError,
-    pubkey::{Pubkey, PubkeyError},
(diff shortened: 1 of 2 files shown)


jackcmay:
@jstarry @mvines FYI

mvines (approved):

programs/bpf_loader/src/syscalls.rs
```diff
@@ -921,6 +923,9 @@ impl<'a> SyscallInvokeSigned<'a> for SyscallInvokeSignedRust<'a> {
                 signers_seeds_len,
                 self.loader_id,
             )?;
+            if signers_seeds.len() > MAX_SEEDS {
```
jstarry, line 926: This isn't the right check. This is limiting the number of signers rather than the number of seeds for a given signer. We need a check below for `untranslated_seeds.len() > MAX_SEEDS`.

But we should also limit the number of signers.. probably to the the length of keyed accounts?

PR #13929, opened 2020-12-03
caches staked nodes computed from vote-accounts

behzadnouri:
#### Problem
Staked nodes are repeatedly recomputed on both gossip threads on each iteration:
https://github.com/solana-labs/solana/blob/52c2cbd63/core/src/cluster_info.rs#L1837
https://github.com/solana-labs/solana/blob/52c2cbd63/core/src/cluster_info.rs#L2610

#### Summary of Changes
Implemented caching.

### DIFF
--- a/core/src/cluster_info.rs
+++ b/core/src/cluster_info.rs
@@ -39,7 +39,6 @@ use itertools::Itertools;
 use rayon::prelude::*;
 use rayon::{ThreadPool, ThreadPoolBuilder};
 use serde::ser::Serialize;
-use solana_ledger::staking_utils;
 use solana_measure::measure::Measure;
 use solana_measure::thread_mem_usage;
 use solana_metrics::{inc_new_counter_debug, inc_new_counter_error};
@@ -1834,7 +1833,7 @@ impl ClusterInfo {
 
                     let stakes: HashMap<_, _> = match bank_forks {
                         Some(ref bank_forks) => {
-                            staking_utils::staked_nodes(&bank_forks.read().unwrap().working_bank())
+                            bank_forks.read().unwrap().working_bank().staked_nodes()
                         }
                         None => HashMap::new(),
                     };
@@ -2492,7 +2491,7 @@ impl ClusterInfo {
                 let epoch = bank.epoch();
                 let epoch_schedule = bank.epoch_schedule();
                 epoch_time_ms = epoch_schedule.get_slots_in_epoch(epoch) * DEFAULT_MS_PER_SLOT;
-                staking_utils::staked_nodes(&bank)
+                bank.staked_nodes()
             }
             None => {
                 inc_new_counter_info!("cluster_info-purge-no_working_bank", 1);
--- a/runtime/src/serde_snapshot/tests.rs
+++ b/runtime/src/serde_snapshot/tests.rs
@@ -261,7 +261,7 @@ mod test_bank_serialize {
 
     // These some what long test harness is required to freeze the ABI of
     // Bank's serialization due to versioned nature
-    #[frozen_abi(digest = "5NHt6PLRJPWJH9FUcweSsUWgN5hXMfXj1BduDrDHH73w")]
+    #[frozen_abi(digest = "Gv3em1cZt9cjQWepg8C5aaK95deyA1fifowRfmmTuoES")]
     #[derive(Serialize, AbiExample)]
     pub struct BankAbiTestWrapperFuture {
         #[serde(serialize_with = "wrapper_future")]
--- a/runtime/src/vote_account.rs
+++ b/runtime/src/vote_account.rs
@@ -26,17 +26,15 @@ pub struct VoteAccount {
     vote_state_once: Once,
 }
 
-#[derive(Debug, AbiExample, Deserialize, Serialize)]
+#[derive(Debug, AbiExample)]
 pub struct VoteAccounts {
     vote_accounts: HashMap<Pubkey, (u64 /*stake*/, ArcVoteAccount)>,
-    #[serde(skip)]
     staked_nodes: RwLock<
         HashMap<
             Pubkey, // VoteAccount.vote_state.node_pubkey.
             u64,    // Total stake across all vote-accounts.
         >,
     >,
-    #[serde(skip, default = "std::sync::Once::new")]
     staked_nodes_once: Once,
 }
 
@@ -267,6 +265,25 @@ impl FromIterator<(Pubkey, (u64 /*stake*/, ArcVoteAccount))> for VoteAccounts {
         I: IntoIterator<Item = (Pubkey, (u64, ArcVoteAccount))>,
     {
         Self::from(HashMap::from_iter(iter))
+    }
+}
+
+impl Serialize for VoteAccounts {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        self.vote_accounts.serialize(serializer)
+    }
+}
+
+impl<'de> Deserialize<'de> for VoteAccounts {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let vote_accounts = VoteAccountsHashMap::deserialize(deserializer)?;
+        Ok(Self::from(vote_accounts))
     }
 }
 
--- a/runtime/src/vote_account.rs
+++ b/runtime/src/vote_account.rs
@@ -1,9 +1,15 @@
 use serde::de::{Deserialize, Deserializer};
 use serde::ser::{Serialize, Serializer};
-use solana_sdk::{account::Account, instruction::InstructionError};
+use solana_sdk::{account::Account, instruction::InstructionError, pubkey::Pubkey};
 use solana_vote_program::vote_state::VoteState;
-use std::ops::Deref;
-use std::sync::{Arc, Once, RwLock, RwLockReadGuard};
+use std::{
+    borrow::Borrow,
+    cmp::Ordering,
+    collections::{hash_map::Entry, HashMap},
+    iter::FromIterator,
+    ops::Deref,
+    sync::{Arc, Once, RwLock, RwLockReadGuard},
+};
 
 // The value here does not matter. It will be overwritten
 // at the first call to VoteAccount::vote_state().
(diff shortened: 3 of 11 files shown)


behzadnouri:
staked_nodes

runtime/src/serde_snapshot/tests.rs
```diff
@@ -261,7 +261,7 @@ mod test_bank_serialize {
 
     // These some what long test harness is required to freeze the ABI of
     // Bank's serialization due to versioned nature
-    #[frozen_abi(digest = "8bNY87hccyDYCRar1gM3NSvpvtiUM3W3rGeJLJayz42e")]
+    #[frozen_abi(digest = "ByqDP7c1XGQSnrVsTCweFcFtyTG6fyohueFCXLBPEMeC")]
```
sakridge, line 264: Any backwards compatibility concerns from changing the serialization hash here? Or the fields are serde_skip?

behzadnouri: I am deriving Serialize/Deserialize manually for the new `VoteAccounts` type which I think will provide backward compatibility. I will tag you on the related section of the code to take a look.

runtime/src/vote_account.rs
```diff
@@ -92,16 +204,104 @@ impl PartialEq<VoteAccount> for VoteAccount {
...
+
+impl<'de> Deserialize<'de> for VoteAccounts {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let vote_accounts = VoteAccountsHashMap::deserialize(deserializer)?;
+        Ok(Self::from(vote_accounts))
+    }
+}
```
behzadnouri, line 288: @sakridge regarding the backward compatibility. This will serialize/deserialize as a
```
HashMap<Pubkey, (u64, ArcVoteAccount)>
```
which was is the current type in the `Stakes` struct:
https://github.com/solana-labs/solana/blob/79fb64687/runtime/src/stakes.rs#L13

carllin:
@behzadnouri I think we already have a `NodeIdToVoteAccounts `: https://github.com/solana-labs/solana/blob/master/runtime/src/epoch_stakes.rs#L6-L12 returned by: https://github.com/solana-labs/solana/blob/master/runtime/src/epoch_stakes.rs#L44-L46 that maps a node to its stake `https://github.com/solana-labs/solana/blob/master/runtime/src/epoch_stakes.rs#L12`. 

The only thing to be wary of is that if a node has zero stake, it won't show up in the map: https://github.com/solana-labs/solana/blob/master/runtime/src/epoch_stakes.rs#L86-L100, so anywhere where we currently use stakes to weight 0-staked keys, we may not find that key in the map and have to handle that.

behzadnouri:
> @behzadnouri I think we already have a `NodeIdToVoteAccounts `: https://github.com/solana-labs/solana/blob/master/runtime/src/epoch_stakes.rs#L6-L12 returned by: https://github.com/solana-labs/solana/blob/master/runtime/src/epoch_stakes.rs#L44-L46 that maps a node to its stake `https://github.com/solana-labs/solana/blob/master/runtime/src/epoch_stakes.rs#L12`.
> 
> The only thing to be wary of is that if a node has zero stake, it won't show up in the map: https://github.com/solana-labs/solana/blob/master/runtime/src/epoch_stakes.rs#L86-L100, so anywhere where we currently use stakes to weight 0-staked keys, we may not find that key in the map and have to handle that.

Here, we need to wrap vote-accounts so that when vote-accounts changes, the staked-nodes cache is updated accordingly (if it has already been populated). Please see the extra logic around insert/remove/add_stake/sub_stake. So `NodeIdToVoteAccounts` wouldn't be the right type here.

core/src/cluster_info.rs
```diff
@@ -1834,7 +1833,7 @@ impl ClusterInfo {
 
                     let stakes: HashMap<_, _> = match bank_forks {
                         Some(ref bank_forks) => {
-                            staking_utils::staked_nodes(&bank_forks.read().unwrap().working_bank())
+                            bank_forks.read().unwrap().working_bank().staked_nodes()
```
sakridge, line 1836: This is only somewhat related, but I think these should be looking at the rooted bank.

core/src/cluster_info.rs
```diff
@@ -2492,7 +2491,7 @@ impl ClusterInfo {
                 let epoch = bank.epoch();
                 let epoch_schedule = bank.epoch_schedule();
                 epoch_time_ms = epoch_schedule.get_slots_in_epoch(epoch) * DEFAULT_MS_PER_SLOT;
-                staking_utils::staked_nodes(&bank)
+                bank.staked_nodes()
```
sakridge, line 2494: Same here, rooted bank.

runtime/src/vote_account.rs
```diff
@@ -31,6 +49,100 @@ impl VoteAccount {
...
+
+    /// VoteState.node_pubkey of this vote-account.
+    fn node_pubkey(&self) -> Option<Pubkey> {
+        Some(self.vote_state().as_ref().ok()?.node_pubkey)
+    }
+}
+
+impl VoteAccounts {
+    pub fn staked_nodes(&self) -> HashMap<Pubkey, u64> {
+        self.staked_nodes_once.call_once(|| {
```
sakridge, line 61: Wait, so if `bank.staked_nodes()` is called before the bank is frozen, will `staked_nodes` return an incomplete view of the staked nodes if further transactions update the `bank.vote_accounts`?

behzadnouri: once `staked_nodes` is calculated, any update to `vote_accounts` also updates the `staked_nodes` cache. plz see `test_staked_nodes` bellow.

core/src/cluster_info.rs
behzadnouri: can we make that change in a separate commit?
just in case it has unknown consequences

core/src/cluster_info.rs
behzadnouri: again, i think would be better to make that change separately

core/src/cluster_info.rs
sakridge: Yes, thats' fine.

runtime/src/vote_account.rs
sakridge: I see, ok.

sakridge: So if the stake goes to zero, how does it work with the key being present in the staked_nodes? Is that consistent to how it was before?

behzadnouri: If stake goes to zero, the node is removed from staked nodes.

Previously, if there is a `vote-account` with _zero stake_, then `VoteAccount.vote_state.node_pubkey` would still show up in the staked-nodes hash-map but with zero stake. So this is a change from before.

But a consequence of previous logic is that _sometimes_ (but not always) zero staked nodes are counted as "staked":
https://github.com/solana-labs/solana/blob/9593ba909/core/src/cluster_info.rs#L2085-L2089
https://github.com/solana-labs/solana/blob/9593ba909/core/src/cluster_info.rs#L2049
https://github.com/solana-labs/solana/blob/9593ba909/core/src/crds_gossip_pull.rs#L525

But that seems more like a bug rather than the intended behavior. Other places do not make this distinction for zero-staked nodes whether they show up in the hash map or not. So, unless I am missing something, I think it is the right thing to exclude nodes with zero stake from staked-nodes hash-map.

Do you know of any place that counting zero-staked nodes as staked is actually intentional and desired behavior?

sakridge: Yea, I agree seems like a bug. I think it sounds ok.

sakridge (dismissed):
lgtm

runtime/src/vote_account.rs
carllin: @behzadnouri thanks for the thorough investigation! I think the same zero-lamport not showing up is true for places where there was a change from `staking_utils::staked_nodes_at_epoch() ->  epoch_staked_nodes()`, so it would be good to double check the zero-staked changes are safe as well.

carllin:
> > @behzadnouri I think we already have a `NodeIdToVoteAccounts `: https://github.com/solana-labs/solana/blob/master/runtime/src/epoch_stakes.rs#L6-L12 returned by: https://github.com/solana-labs/solana/blob/master/runtime/src/epoch_stakes.rs#L44-L46 that maps a node to its stake `https://github.com/solana-labs/solana/blob/master/runtime/src/epoch_stakes.rs#L12`.
> > The only thing to be wary of is that if a node has zero stake, it won't show up in the map: https://github.com/solana-labs/solana/blob/master/runtime/src/epoch_stakes.rs#L86-L100, so anywhere where we currently use stakes to weight 0-staked keys, we may not find that key in the map and have to handle that.
> 
> Here, we need to wrap vote-accounts so that when vote-accounts changes, the staked-nodes cache is updated accordingly (if it has already been populated). Please see the extra logic around insert/remove/add_stake/sub_stake. So `NodeIdToVoteAccounts` wouldn't be the right type here.

Got it, this is for tracking live stakes

runtime/src/vote_account.rs
carllin: From what I gather:

1) The change in broadcast stage: https://github.com/solana-labs/solana/pull/13929/files#diff-b9b3dfeefcf7a68b5072d1b4fa4beb2598e93ce3130856beab78607cc960c60cR308 should be safe because it ultimately calls down to: https://github.com/solana-labs/solana/blob/master/core/src/cluster_info.rs#L1254-L1256, which accounts for missing stakes

2) The change in retransmit: https://github.com/solana-labs/solana/pull/13929/files#diff-8d73ca7d46b0b2dba59642e0e183fbb6127c8459ade483df4ac729390519e6f6R248, like 1) above, also calls down to https://github.com/solana-labs/solana/blob/master/core/src/cluster_info.rs#L1254-L1256, which accounts for missing stakes

3) The change in leader schedule: https://github.com/solana-labs/solana/pull/13929/files#diff-e7124b3c7b7174e8cc0a9a8a62dc8fbd9e52de923c3133712eaffb96d6c8c414R10 would be scary if it caused the leader schedule to be computed differently, but should be ok if `WeightedIndex` treats zero weight as a no-op: https://github.com/solana-labs/solana/blob/master/ledger/src/leader_schedule.rs#L25, but this would be good to confirm.

behzadnouri: yes, weighted-index should ignore zero weights

carllin (dismissed):

PR #13933, opened 2020-12-03
Fix skip propagation check

carllin:
#### Problem
Non consecutive leader blocks like `1-> 3` are still within the same `NUM_CONSECUTIVE_LEADER_SLOTS` set of slots, but `maybe_start_leader` will still incur a propagation check on slot `1` before starting to generate block `3`. This may cause slot `3` to be skipped. 

#### Summary of Changes
Don't trigger the propagation check on blocks within the same `NUM_CONSECUTIVE_LEADER_SLOTS` set

Fixes #


sakridge (approved):
nice, looks good

PR #13944, opened 2020-12-03
Improved Transaction Forwarding

sakridge:
#### Problem

Transaction forwarding is fragile, single node can be down/behind/etc and retry at 5 seconds is too conservative.

#### Summary of Changes

Forward to multiple leaders and make it configurable
TODO:
- [x] de-dupe leaders when multiple leaders are specified
- [x] testing
- [x] Hold more transactions than 1 slot away in banking stage, maybe the range of 2-3 leader slots.

Fixes #

### DIFF
--- a/cli/src/cli.rs
+++ b/cli/src/cli.rs
@@ -522,6 +522,7 @@ impl CliConfig<'_> {
             commitment: CommitmentConfig::recent(),
             send_transaction_config: RpcSendTransactionConfig {
                 skip_preflight: true,
+                preflight_commitment: Some(CommitmentConfig::recent().commitment),
                 ..RpcSendTransactionConfig::default()
             },
             ..Self::default()
@@ -1406,6 +1407,7 @@ fn send_deploy_messages(
             config.commitment,
             RpcSendTransactionConfig {
                 skip_preflight: true,
+                preflight_commitment: Some(config.commitment.commitment),
                 ..RpcSendTransactionConfig::default()
             },
         )
@@ -1815,6 +1817,7 @@ fn process_set_program_upgrade_authority(
             config.commitment,
             RpcSendTransactionConfig {
                 skip_preflight: true,
+                preflight_commitment: Some(config.commitment.commitment),
                 ..RpcSendTransactionConfig::default()
             },
         )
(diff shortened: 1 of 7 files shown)


CriesofCarrots:
@sakridge is there any reason the 2 new params should not be plugged into `solana-validator` cli?
Also, I have a couple questions on BankingStage changes. Can we chat tomorrow?

CriesofCarrots:
@sakridge I think this may be ready to go. 

I did some testing against a GCE testnet today by blasting an rpc node with 10k transactions. The combo of the longer banking-stage hold and increased leader forwards dropped the number of slots it took for all transactions to land by about 15% (29 vs 36 slots). Just increasing the `rpc_send_transaction_leader_forward_count` appeared to reduce the number of retries done by the send-transaction-service, and reducing `rpc_send_transaction_retry_ms` cleared the pending transaction queue faster.

Are there any particular test cases you'd like to see?

sakridge:
> @sakridge I think this may be ready to go.
> 
> I did some testing against a GCE testnet today by blasting an rpc node with 10k transactions. The combo of the longer banking-stage hold and increased leader forwards dropped the number of slots it took for all transactions to land by about 15% (29 vs 36 slots). Just increasing the `rpc_send_transaction_leader_forward_count` appeared to reduce the number of retries done by the send-transaction-service, and reducing `rpc_send_transaction_retry_ms` cleared the pending transaction queue faster.
> 
> Are there any particular test cases you'd like to see?

Cool, those results sound good. Seems ready to go.

Further work could be adjusting the banking_stage buffers which could have an area where they do both a hold and forward and maybe some better pruning.

Something like:
If I'm leader within 0-2 slots => hold
3-7 slots => hold + forward
8+ slots => forward

CriesofCarrots:
> Further work could be adjusting the banking_stage buffers which could have an area where they do both a hold and forward and maybe some better pruning.
> 
> Something like:
> If I'm leader within 0-2 slots => hold
> 3-7 slots => hold + forward
> 8+ slots => forward

Ah cool, that's a nice idea. I'll merge this, and play with that and PR separately.

PR #13945, opened 2020-12-03
verifies that serialized size of transactions does not exceed packet data size

behzadnouri:
#### Problem
Serialized size of transactions can exceed PACKET_DATA_SIZE.

#### Summary of Changes
Check serialized size of transactions in `verify_transaction_signatures`.


sakridge (approved):
lgtm

jstarry:
looks good

PR #13946, opened 2020-12-03
Blockstore exact purge cleanup

CriesofCarrots:
#### Problem
Blockstore exact purge can't handle malformed transactions (no signatures)

#### Summary of Changes
- Properly check index and skip deletes, since transactions with no signatures cannot have related TransactionStatus or AddressSignature records


sakridge (approved):
looks good!

PR #13961, opened 2020-12-04
v1.3 persistent tower

CriesofCarrots:
https://github.com/solana-labs/solana/pull/10718 and follow-up for v1.3

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4128,6 +4128,7 @@ name = "solana-local-cluster"
 version = "1.3.24"
 dependencies = [
  "assert_matches",
+ "fs_extra",
  "gag",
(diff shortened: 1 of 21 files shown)


ryoqun:
@CriesofCarrots Thanks for working instead of me. Is this still needed and planned to be merged? As you're more aware of the root cause of the outage, I think this giant back-porting can be skipped, perhaps?

If needed, I'm happy to review this. At first glance, this looks mostly ok.

CriesofCarrots:
> I think this giant back-porting can be skipped, perhaps?

Yes, especially given the plans for v1.4, I think it can be skipped. @sakridge can you confirm?

PR #13976, opened 2020-12-06
Use LRU cache and ahash hash of shreds to filter duplicates

sakridge:
#### Problem

Just comparing slot/index filters duplicate shreds too aggressively.

#### Summary of Changes

Use LRU cache of blake3 hash of the shred data to filter duplicates.

Fixes #

### DIFF
--- a/core/src/shred_fetch_stage.rs
+++ b/core/src/shred_fetch_stage.rs
@@ -1,6 +1,10 @@
 //! The `shred_fetch_stage` pulls shreds from UDP sockets and sends it to a channel.
 
-use bv::BitVec;
+use ahash::AHasher;
+use lru::LruCache;
+use rand::{thread_rng, Rng};
+use std::hash::Hasher;
+
 use solana_ledger::blockstore::MAX_DATA_SHREDS_PER_SLOT;
 use solana_ledger::shred::{
     CODING_SHRED, DATA_SHRED, OFFSET_OF_SHRED_INDEX, OFFSET_OF_SHRED_SLOT, OFFSET_OF_SHRED_TYPE,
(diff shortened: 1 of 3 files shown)


core/src/shred_fetch_stage.rs
```diff
@@ -128,9 +126,8 @@ impl ShredFetchStage {
         let mut stats = ShredFetchStats::default();
 
         while let Some(mut p) = recvr.iter().next() {
-            if last_cleared.elapsed().as_millis() > 200 {
-                shreds_received.clear();
-                last_cleared = Instant::now();
+            if last_updated.elapsed().as_millis() > 400 {
```
carllin, line 129: 400 -> `DEFAULT_MS_PER_SLOT`?

core/src/shred_fetch_stage.rs
```diff
@@ -86,16 +87,13 @@ impl ShredFetchStage {
...
                     // Shred filter
-                    let slot_received =
-                        shreds_received
-                            .entry((slot, shred_type))
-                            .or_insert_with(|| {
-                                BitVec::new_fill(false, MAX_DATA_SHREDS_PER_SLOT as u64)
-                            });
-                    if !slot_received.get(index.into()) {
+                    let mut hasher = blake3::Hasher::new();
+                    hasher.update(&p.data[0..p.meta.size]);
```
carllin, line 91: is there a big perf implication here for computing a cryptographic hash like blake3 for every shred? 

Here we only care about detecting duplicates, so can we use a weaker hash by using something like hashbrown::HashSet (uses ahash): https://github.com/tkaitchuck/aHash

sakridge: Yes, but a potential bad actor can easily find collisions in those and I think blake3 is fairly close in performance. I can test it, but if it's close, I would rather have the safety of blake3.

carllin: @sakridge we can add a seed, i.e. generate a seed every 400 ms to add to the key. I think purported benchmarks were supposed to be 16-32x faster.

sakridge: blake3 can reject at 650k shreds/s per core, but ahash is faster.

carllin (approved):

PR #13992, opened 2020-12-07
Better dupe detection

sakridge:
#### Problem

Duplicate shred cases ignored.

#### Summary of Changes

Set in duplicate column and ignore on voting.

Fixes #

### DIFF
--- a/core/src/replay_stage.rs
+++ b/core/src/replay_stage.rs
@@ -989,27 +989,41 @@ impl ReplayStage {
             // errors related to the slot being purged
             let slot = bank.slot();
             warn!("Fatal replay error in slot: {}, err: {:?}", slot, err);
-            if let BlockstoreProcessorError::InvalidBlock(BlockError::InvalidTickCount) = err {
-                datapoint_info!(
-                    "replay-stage-mark_dead_slot",
-                    ("error", format!("error: {:?}", err), String),
-                    ("slot", slot, i64)
-                );
-            } else {
-                datapoint_error!(
-                    "replay-stage-mark_dead_slot",
-                    ("error", format!("error: {:?}", err), String),
-                    ("slot", slot, i64)
-                );
-            }
-            bank_progress.is_dead = true;
-            blockstore
-                .set_dead_slot(slot)
-                .expect("Failed to mark slot as dead in blockstore");
+            let is_serious = matches!(
+                err,
+                BlockstoreProcessorError::InvalidBlock(BlockError::InvalidTickCount)
+            );
+            Self::mark_dead_slot(blockstore, bank_progress, slot, &err, is_serious);
             err
         })?;
 
         Ok(tx_count)
+    }
+
+    fn mark_dead_slot(
+        blockstore: &Blockstore,
+        bank_progress: &mut ForkProgress,
+        slot: Slot,
+        err: &BlockstoreProcessorError,
+        is_serious: bool,
+    ) {
+        if is_serious {
+            datapoint_error!(
+                "replay-stage-mark_dead_slot",
+                ("error", format!("error: {:?}", err), String),
+                ("slot", slot, i64)
+            );
+        } else {
+            datapoint_info!(
+                "replay-stage-mark_dead_slot",
+                ("error", format!("error: {:?}", err), String),
+                ("slot", slot, i64)
+            );
+        }
+        bank_progress.is_dead = true;
+        blockstore
+            .set_dead_slot(slot)
+            .expect("Failed to mark slot as dead in blockstore");
     }
 
     #[allow(clippy::too_many_arguments)]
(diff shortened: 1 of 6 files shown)


lombokcryptoid (approved):

lombokcryptoid (approved):

lombokcryptoid (approved):

lombokcryptoid (approved):

core/src/replay_stage.rs
```diff
@@ -1337,23 +1337,37 @@ impl ReplayStage {
...
-                info!("bank frozen: {}", bank.slot());
-                bank.freeze();
-                heaviest_subtree_fork_choice
-                    .add_new_leaf_slot(bank.slot(), Some(bank.parent_slot()));
-                if let Some(sender) = bank_notification_sender {
-                    sender
-                        .send(BankNotification::Frozen(bank.clone()))
-                        .unwrap_or_else(|err| warn!("bank_notification_sender failed: {:?}", err));
-                }
+                if !blockstore.has_duplicate_shreds_in_slot(bank.slot()) {
```
carllin, line 1340: Do you think it's worth it to check this earlier in the replay of this slot here (maybe saves some time replaying this slot?): https://github.com/solana-labs/solana/pull/13992/files#diff-6d8458bb2e53158ac472a9ad4709e6a0a52b75d930c019013298e8acda133828R1286

On the flip side, it incurs the cost of the extra check while replaying, which wouldn't be much if our rocks wrapper actually supported a key check without loading the value: https://github.com/rust-rocksdb/rust-rocksdb/issues/439#issuecomment-740763958

core/src/replay_stage.rs
```diff
@@ -1337,23 +1337,37 @@ impl ReplayStage {
...
+                    }
 
-                Self::record_rewards(&bank, &rewards_recorder_sender);
+                    Self::record_rewards(&bank, &rewards_recorder_sender);
+                } else {
+                    bank_progress.is_dead = true;
+                    blockstore
+                        .set_dead_slot(bank.slot())
+                        .expect("Failed to mark slot as dead in blockstore");
+                    warn!(
```
carllin, line 1365: The other instance of marking slots dead in replay stage submits a metric like:
```
datapoint_error!(
                    "replay-stage-mark_dead_slot",
                    ("error", format!("error: {:?}", err), String),
                    ("slot", slot, i64)
                );
```

might be worth doing here as well

ledger/src/blockstore.rs
```diff
@@ -1051,6 +1051,26 @@ impl Blockstore {
             });
 
             if erasure_config != erasure_meta.config {
+                let coding_start = erasure_meta.first_coding_index;
+                let coding_end = coding_start + erasure_meta.size as u64;
```
carllin, line 1055: hmmm is this `erasure_meta.size` set anywhere? It defaults to 0, and I don't see `ErasureMeta::set_size` called anywhere. 

Maybe a test to see if this logic catches the duplicate and marks the duplicate column?

ledger/src/blockstore.rs
```diff
@@ -1051,6 +1051,26 @@ impl Blockstore {
...
+                let mut conflicting_shred = None;
+                for coding_index in coding_start..coding_end {
+                    if let Ok(Some(shred)) = self.get_coding_shred(slot, coding_index) {
+                        conflicting_shred = Some(shred);
```
carllin, line 1059: should we add a sanity check here that this `conflicting_shred` has a different `num_data_shreds()` or `num_coding_shreds` than the newly received `shred`?

ledger/src/blockstore.rs
```diff
@@ -1051,6 +1051,26 @@ impl Blockstore {
             });
 
             if erasure_config != erasure_meta.config {
+                let coding_start = erasure_meta.first_coding_index;
+                let coding_end = coding_start + erasure_meta.size as u64;
+                let mut conflicting_shred = None;
+                for coding_index in coding_start..coding_end {
+                    if let Ok(Some(shred)) = self.get_coding_shred(slot, coding_index) {
```
carllin, line 1058: Thinking about this more, this check will trigger a duplicate proof even if

Two *different* coding shreds from the same FEC set have *different* configs for one of (num_coding, num_data, first_coding_index), but those coding shreds aren't for the same index. 

I think what we actually want to check for is if two coding shreds have the same slot/index and different payloads. For instance, if a leader generated

C1, C2. for one set, and a second C2' with a different coding config, we would want to detect C2, C2' conflict.

sakridge: We can check both. Changing the coding config in the middle of an FEC set should be a violation, right?

ledger/src/blockstore.rs
```diff
@@ -1051,6 +1051,26 @@ impl Blockstore {
             });
 
             if erasure_config != erasure_meta.config {
+                let coding_start = erasure_meta.first_coding_index;
```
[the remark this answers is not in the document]
sakridge: ah, good catch

core/src/replay_stage.rs
sakridge: I think we can revisit it later, at least now with just checking once it shouldn't be much more expensive in the normal case.

ryoqun: @carllin @sakridge 

> On the flip side, it incurs the cost of the extra check while replaying, which might be more efficient once the rocks wrapper supports a key check without loading the value: rust-rocksdb/rust-rocksdb#439 (comment)

This is finally landed in upstream and available in our codebase starting from: #18904

... but it look like the new method incurs the possibility of false positive:

https://docs.rs/rocksdb/0.17.0/rocksdb/struct.DBWithThreadMode.html#method.key_may_exist_cf_opt:

> Returns false if the given key definitely doesn’t exist in the specified column family, otherwise returns true.

carllin: dang, well that false positive makes that functionality a whole lot less useful :|

PR #13995, opened 2020-12-07
Add local cluster tests that broadcast duplicate slots

jstarry:
#### Problem
Lacking local cluster tests for faulty nodes which sends duplicate shreds for complete blocks

#### Summary of Changes
- Added new "duplicate node" local cluster test which runs two nodes with the same keys (one gets setup after the other)
- Added a new broadcast stage for sending duplicate shreds:
  - Partitions the cluster based on stake
  - Does not send invalid entries (which would just cause the recipient to mark the slot as dead)
  - Wait to broadcast alternative version of a slot until all shreds can be sent at once (to avoid nodes getting duplicate shreds from repair)
  - Broadcasts duplicate shreds after a configurable slot delay

### DIFF
--- a/core/src/broadcast_stage/broadcast_duplicates_run.rs
+++ b/core/src/broadcast_stage/broadcast_duplicates_run.rs
@@ -1,0 +1,333 @@
+use super::broadcast_utils::ReceiveResults;
+use super::*;
+use log::*;
+use solana_ledger::entry::{create_ticks, Entry, EntrySlice};
+use solana_ledger::shred::Shredder;
+use solana_runtime::blockhash_queue::BlockhashQueue;
+use solana_sdk::clock::Slot;
+use solana_sdk::fee_calculator::FeeCalculator;
+use solana_sdk::hash::Hash;
+use solana_sdk::signature::{Keypair, Signer};
+use solana_sdk::transaction::Transaction;
+use std::collections::VecDeque;
+use std::sync::Mutex;
+
+// Queue which facilitates delivering shreds with a delay
+type DelayedQueue = VecDeque<(Option<Pubkey>, Option<Vec<Shred>>)>;
+
+#[derive(Clone)]
+pub(super) struct BroadcastDuplicatesRun {
+    config: BroadcastDuplicatesConfig,
+    // Local queue for broadcast to track which duplicate blockhashes we've sent
+    duplicate_queue: BlockhashQueue,
+    // Shared queue between broadcast and transmit threads
+    delayed_queue: Arc<Mutex<DelayedQueue>>,
+    // Buffer for duplicate entries
+    duplicate_entries_buffer: Vec<Entry>,
+    last_duplicate_entry_hash: Hash,
+    last_broadcast_slot: Slot,
+    next_shred_index: u32,
+    shred_version: u16,
+    keypair: Arc<Keypair>,
+}
+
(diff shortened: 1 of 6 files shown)


jstarry:
Test commands with log level
```
RUST_LOG=warn,solana_local_cluster=info RUST_BACKTRACE=1 cargo test --package solana-local-cluster --test local_cluster -- test_duplicate_node --exact --nocapture --ignored
RUST_LOG=warn,solana_local_cluster=info RUST_BACKTRACE=1 cargo test --package solana-local-cluster --test local_cluster -- test_fake_shreds_broadcast_leader --exact --nocapture --ignored
```

jstarry:
cc @carllin @sakridge

carllin:
@jstarry thanks!

jstarry:
@carllin we still want this to land after your changes right? If so, I'll keep it from getting going stale

carllin:
@jstarry yeah, there's a few more changes before this one can go in, but that's the goal!

jstarry:
Recreated to fix CI: https://github.com/solana-labs/solana/pull/16050

carllin:
Found one more edge case: https://github.com/carllin/solana/commit/e5cacb122cf9e65c4e1f3fc7b2347ffde44462af

Broadcasts need to handle interrupted slots, otherwise they'll start from the shred index of the previous interrupted block

local-cluster/tests/local_cluster.rs
```diff
@@ -2002,6 +2020,39 @@ fn test_faulty_node(faulty_node_type: BroadcastStageType) {
     cluster.check_for_new_roots(16, &"test_faulty_node");
 }
 
+#[test]
+#[serial]
+#[allow(unused_attributes)]
+#[ignore]
+// Run two validators with the same keys
+fn test_duplicate_node() {
```
carllin, line 2028: I think we can remove this test since running two validators with the same keys causes an autokill once they detect each other in gossip now

jstarry: couldn't a malicious user disable the autokill?

carllin: yeah, but since we don't disable that, this will kill the nodes and then the test will fail right?

carllin:
Tagging @AshwinSekar, these are some tests, specifically `test_duplicate_shreds_broadcast_leader` that tests a leader broadcasting two versions of the same block. Might be useful to integrate this into a test as well across multiple nodes

core/src/broadcast_stage/broadcast_duplicates_run.rs
```diff
@@ -0,0 +1,333 @@
...
+            }
+        }
+
+        let duplicate_recipients = Arc::new(duplicate_recipients);
+        let real_recipients = Arc::new(real_recipients);
+
+        let data_shreds = Arc::new(data_shreds);
+        blockstore_sender.send((data_shreds.clone(), None))?;
+
+        // 3) Start broadcast step
```
AshwinSekar, line 266: For my understanding, why do we send out the duplicate shreds first before the real recipients?

local-cluster/tests/local_cluster.rs
jstarry: Ah sure, removed! I'll look into adding that back in a follow up PR

carllin (approved):

PR #13996, opened 2020-12-07
Fix logsSubscribe

CriesofCarrots:
#### Problem
`logsSubscribe` docs say `commitment` is optional. It isn't. If you try the 2nd docs example, it fails expecting a tuple of length 2.
Also, docs example c+p returns a parse error due to missing comma.

#### Summary of Changes
Fix commitment handling to really be optional; add comma in docs


mvines (approved):
ty

PR #13884, opened 2020-12-01
Simple cap by including sysvars and native programs

ryoqun:
#### Problem

Simplify capitalization calculation by removing special casing.

#### Summary of Changes

Adjust cap while adding/updating sysvar and native program accounts correctly to simplify cap calculation.

~Also, this fixes a regression bug and a new bug introduced at #13403. In this sense, this pr is a follow up to #13403, #11750.~ (EDIT: this was done by separate pr)

PR #14008, opened 2020-12-08
Add tests for store_account_and_update_capitalization

ryoqun:
#### Problem

#13884 is too big. Let's split it.

#### Summary of Changes

Extracted pretty simple test additions and minor implementation modifications.

### DIFF
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -3693,14 +3693,18 @@ impl Bank {
     }
 
     #[cfg(test)]
-    fn add_account_and_update_capitalization(&self, pubkey: &Pubkey, new_account: &Account) {
+    fn store_account_and_update_capitalization(&self, pubkey: &Pubkey, new_account: &Account) {
         if let Some(old_account) = self.get_account(&pubkey) {
-            if new_account.lamports > old_account.lamports {
-                self.capitalization
-                    .fetch_add(new_account.lamports - old_account.lamports, Relaxed);
-            } else {
-                self.capitalization
-                    .fetch_sub(old_account.lamports - new_account.lamports, Relaxed);
+            match new_account.lamports.cmp(&old_account.lamports) {
+                std::cmp::Ordering::Greater => {
+                    self.capitalization
+                        .fetch_add(new_account.lamports - old_account.lamports, Relaxed);
+                }
+                std::cmp::Ordering::Less => {
+                    self.capitalization
+                        .fetch_sub(old_account.lamports - new_account.lamports, Relaxed);
+                }
+                std::cmp::Ordering::Equal => {}
             }
         } else {
             self.capitalization.fetch_add(new_account.lamports, Relaxed);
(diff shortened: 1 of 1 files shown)


runtime/src/bank.rs
```diff
@@ -3693,14 +3693,18 @@ impl Bank {
     }
 
     #[cfg(test)]
-    fn add_account_and_update_capitalization(&self, pubkey: &Pubkey, new_account: &Account) {
+    fn store_account_and_update_capitalization(&self, pubkey: &Pubkey, new_account: &Account) {
```
ryoqun, line 3696: Well, I renamed this to sound like a cousin to `store_account`. Introducing different verb `add_` was a bad idea...

runtime/src/bank.rs
```diff
@@ -5131,6 +5135,84 @@ pub(crate) mod tests {
...
+    fn assert_capitalization_diff(bank: &Bank, updater: impl Fn(), asserter: impl Fn(u64, u64)) {
+        let old = bank.capitalization();
+        updater();
+        let new = bank.capitalization();
+        asserter(old, new);
+        assert_eq!(bank.capitalization(), bank.calculate_capitalization());
+    }
+
+    #[test]
+    fn test_store_account_and_update_capitalization_missing() {
```
ryoqun, line 5147: classical textbookish non-overlapping exhaustive shortly-chucked set of unit tests for `store_account_and_update_capitalization`, which is newly promoted from `cfg(test)`.

runtime/src/bank.rs
```diff
@@ -5131,6 +5135,84 @@ pub(crate) mod tests {
...
+        let bank = Bank::new(&genesis_config);
+        let pubkey = mint_keypair.pubkey();
+
+        let new_lamports = 500;
+        let account = Account::new(new_lamports, 0, &system_program::id());
+
+        assert_capitalization_diff(
+            &bank,
+            || bank.store_account_and_update_capitalization(&pubkey, &account),
+            |old, new| assert_eq!(old + 100, new),
```
ryoqun, line 5176: trivia: Q: why do I this manual math hard code like this instead of `old + (new_lamports - old_lamports)` sometimes? Isn't a bad practice?

A: Well this is partly art of coding but these well nuanced hard coded math can catch really silly test failure. Say, I assigned same value to both `old_lamports` and `new_lamports` in a huste of refactoring by accident, then, `old + (new_lamports - old_lamports)` wouldn't detect that this test got broken meaning actually noting (including intended branch coverage in `store_account_and_update_capitalization`) is tested.

Also, note that these silly things could happen in the refactoring of `cfg(not(test))` code.

ryoqun:
@CriesofCarrots Could you review this at your convenient time?

lombokcryptoid (approved):

lombokcryptoid (approved):

lombokcryptoid (approved):

CriesofCarrots (approved):

PR #14014, opened 2020-12-08
Remove "Timezone For Log Messages" section

leoluk:
To my understanding, the Solana process always logs a UTC timestamp
anyway, so this is not needed. Plus, using PST on servers is really
weird from a systems engineering point of view, especially if you're
in a different time zone.


mvines (approved):
Thanks

PR #14018, opened 2020-12-08
checks for duplicate validator instances using gossip

behzadnouri:
#### Problem
Two running instances of the same validator is not good.

#### Summary of Changes
Added a new node-instance crds value which contains a random token and timestamp of when the instance was created. If a node sees itself with a different token value and more recent timestamp it will stop.

### DIFF
--- a/core/src/cluster_info.rs
+++ b/core/src/cluster_info.rs
@@ -2917,7 +2917,9 @@ impl ClusterInfo {
                                     self.id()
                                 );
                                 exit.store(true, Ordering::Relaxed);
-                                return;
+                                // TODO: Pass through ValidatorExit here so
+                                // that this will exit cleanly.
+                                std::process::exit(1);
                             }
                             _ => error!("gossip run_listen failed: {}", err),
                         }
--- a/core/src/cluster_info.rs
+++ b/core/src/cluster_info.rs
@@ -18,8 +18,8 @@ use crate::{
     crds_gossip_error::CrdsGossipError,
     crds_gossip_pull::{CrdsFilter, ProcessPullStats, CRDS_GOSSIP_PULL_CRDS_TIMEOUT_MS},
     crds_value::{
-        self, CrdsData, CrdsValue, CrdsValueLabel, EpochSlotsIndex, LowestSlot, SnapshotHash,
-        Version, Vote, MAX_WALLCLOCK,
+        self, CrdsData, CrdsValue, CrdsValueLabel, EpochSlotsIndex, LowestSlot, NodeInstance,
+        SnapshotHash, Version, Vote, MAX_WALLCLOCK,
     },
     data_budget::DataBudget,
     epoch_slots::EpochSlots,
@@ -300,6 +300,7 @@ pub struct ClusterInfo {
     socket: UdpSocket,
     local_message_pending_push_queue: RwLock<Vec<(CrdsValue, u64)>>,
     contact_debug_interval: u64,
+    instance: RwLock<NodeInstance>,
 }
 
 impl Default for ClusterInfo {
@@ -556,6 +557,7 @@ impl ClusterInfo {
             socket: UdpSocket::bind("0.0.0.0:0").unwrap(),
             local_message_pending_push_queue: RwLock::new(vec![]),
             contact_debug_interval: DEFAULT_CONTACT_DEBUG_INTERVAL,
+            instance: RwLock::new(NodeInstance::new(id, timestamp())),
         };
         {
             let mut gossip = me.gossip.write().unwrap();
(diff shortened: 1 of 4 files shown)


core/src/cluster_info.rs
```diff
@@ -614,16 +617,25 @@ impl ClusterInfo {
     ) {
         let now = timestamp();
         self.my_contact_info.write().unwrap().wallclock = now;
-        let entry =
-            CrdsValue::new_signed(CrdsData::ContactInfo(self.my_contact_info()), &self.keypair);
+        self.instance.write().unwrap().update_wallclock(now);
+        let entries: Vec<_> = vec![
+            CrdsData::ContactInfo(self.my_contact_info()),
+            CrdsData::NodeInstance(self.instance.read().unwrap().clone()),
```
mvines, line 623: Do we need to send `NodeInstance` continually?   `CrdsData::Version` is just added once at startup

behzadnouri: That is a good question. Is there a risk that the value is not propagated if it is only pushed once?
Something like:
* The network is partitioned during the push process.
    * or, the value is [becomes outdated and not pushed any further](https://github.com/solana-labs/solana/blob/20c0e8793/core/src/crds_gossip_push.rs#L178-L188).
* and, the value expires and is purged from crds table, before it is returned in pull-requests.
    * or, similarly, it is [not processed in pull request responses because it is too old](https://github.com/solana-labs/solana/blob/f6a7732ff/core/src/crds_gossip_pull.rs#L339-L345).

There is an [`update_record_timestamp`](https://github.com/solana-labs/solana/blob/f6a7732ff/core/src/crds.rs#L240-L251) which extends `local-timestamp` of the associated values if we hear from a contact-info (on some but not all gossip paths), which may stop the node from getting purged from crds table as long as its contact info is propagated, but seems like if the `wallclock` is too old, it will not get propagated through gossip.

core/src/cluster_info.rs
```diff
@@ -300,6 +300,7 @@ pub struct ClusterInfo {
     socket: UdpSocket,
     local_message_pending_push_queue: RwLock<Vec<(CrdsValue, u64)>>,
     contact_debug_interval: u64,
+    instance: RwLock<NodeInstance>,
```
sakridge, line 303: can `instance` just be the unique u64 value? Everything else can be constructed at send time, no? And not changing, so doesn't need a lock.

behzadnouri: We at least need:
* a timestamp of when the instance started, to stop the older one.
* and also, a randomly generated token, to distinguish the instances if both start at the same milli-second (e.g. programmatically)

Nevertheless, we do not need to update the wallclock here, so I can remove `RwLock`, and only update the wallclock on the value which is pushed through gossip.

core/src/cluster_info.rs
mvines: > Is there a risk that the value is not propagated if it is only pushed once?

That sounds like it.  I guess version has this same problem then.  Maybe not a big deal here though, you're now more knowledgable in this area than I am so your call :)

core/src/cluster_info.rs
mvines: I think if `instance` was a UTC timestamp at ms or even second accuracy we'd get 99.9% of the way there.   If two instances happen to start at exactly the same time then they both exit.

core/src/cluster_info.rs
behzadnouri: I am leaning towards sending these periodically, just in case they get dropped because of network partitions, message expiration, etc. It should not add too much overhead.

core/src/cluster_info.rs
behzadnouri: Current code no longer has `RwLock` so I guess should be good.

Alternative is to use `std::time::SystemTime::now()` as the timestamp of when the validator started (as opposed to milliseconds from `timestamp()`), and assume that no two validators can start at the exact same `SystemTime`. Then we no longer need the random `token`, and one single immutable `SystemTime` should suffice.

core/src/cluster_info.rs
mvines: sgtm

core/src/cluster_info.rs
mvines: wfm

mvines:
@behzadnouri - is this PR ready for another review now?

mvines (approved):

behzadnouri:
> @behzadnouri - is this PR ready for another review now?

yes, the changes where:
* adding `std::process::exit` to stop all threads (with a todo to do clean exit, but that requires bigger code change to pass around `ValidatorExit`).
* removing `RwLock` from `ClusterInfo.instance`, in favor of an immutable value. `wallclock` is updated on a clone before pushing the value over gossip.
* pushing `instance` early on in gossip along with `version`, in addition to periodic pushes with `ContactInfo` every ~7 seconds.

PR #14030, opened 2020-12-09
duplicate (slot, index) retransmit Shred filter

sakridge:
#### Problem

Duplicate shred slot/index combination could produce a lot of retransmit traffic.

#### Summary of Changes

Limit to retransmitting 2 shreds.

Fixes #

### DIFF
--- a/core/src/retransmit_stage.rs
+++ b/core/src/retransmit_stage.rs
@@ -1,5 +1,7 @@
 //! The `retransmit_stage` retransmits shreds between validators
 
+use crate::shred_fetch_stage::ShredFetchStage;
+use crate::shred_fetch_stage::ShredFetchStats;
 use crate::{
     cluster_info::{compute_retransmit_peers, ClusterInfo, DATA_PLANE_FANOUT},
     cluster_info_vote_listener::VerifiedVoteReceiver,
@@ -12,7 +14,10 @@ use crate::{
     result::{Error, Result},
     window_service::{should_retransmit_and_persist, WindowService},
 };
+use ahash::AHasher;
 use crossbeam_channel::Receiver;
+use lru::LruCache;
+use rand::{thread_rng, Rng};
 use solana_ledger::{
     blockstore::{Blockstore, CompletedSlotsReceiver},
     leader_schedule_cache::LeaderScheduleCache,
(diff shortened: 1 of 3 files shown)


core/src/retransmit_stage.rs
```diff
@@ -279,6 +287,23 @@ fn retransmit(
...
 
+            match ShredFetchStage::get_slot_index(packet, &mut ShredFetchStats::default()) {
+                Some(slot_index) => {
+                    let mut received = shreds_received.lock().unwrap();
+                    if let Some(count) = received.get(&slot_index) {
+                        if *count < MAX_DUPLICATE_COUNT {
+                            let new_count = *count + 1;
+                            received.put(slot_index, new_count);
+                        } else {
+                            drop(received);
```
behzadnouri, line 298: why need `drop(received);` here?

sakridge: We don't. I removed it.

core/src/retransmit_stage.rs
```diff
@@ -254,6 +262,7 @@ fn retransmit(
         w_epoch_stakes_cache.stakes_and_index = stakes_and_index;
         drop(w_epoch_stakes_cache);
         r_epoch_stakes_cache = epoch_stakes_cache.read().unwrap();
+        shreds_received.lock().unwrap().clear();
```
carllin, line 265: Wondering if we should change `now - last > 1000` here: https://github.com/solana-labs/solana/pull/14030/files#diff-8d73ca7d46b0b2dba59642e0e183fbb6127c8459ade483df4ac729390519e6f6R255 to `now - last > NUM_CONSECUTIVE_LEADER_SLOTS * DEFAULT_MS_PER_SLOT` because that's how often the propagation check may retransmit.

core/src/retransmit_stage.rs
```diff
@@ -279,6 +288,22 @@ fn retransmit(
                 continue;
             }
 
+            match ShredFetchStage::get_slot_index(packet, &mut ShredFetchStats::default()) {
```
carllin, line 291: Should this logic also account for shred hash so that we retransmit duplicates? i.e. if we retransmit the same shred twice, but then we see a duplicate, we probably want to retransmit the duplicate as well.

Maybe the logic should be, retransmit a single version based on hash of a shred slot/index twice, AND only retransmit at most 2 different hashes of the a shred with the same slot/index

sakridge: Shred fetch stage will filter on shred hash, so only unique shreds should come into here. Then this will allow 2 unique shreds.

core/src/retransmit_stage.rs
sakridge: Those timings aren't exact, it just needs to prevent it from creating a huge amount of unrestricted traffic.

Issue #13323, opened 2020-10-31
explorer: Token history table is not very user friendly


jstarry:
#### Problem
The token history table is very barebones and is hard to get much useful information out of. For instance, the history table lists all the transfers for each token but doesn't say what the amount was or if tokens were sent or received. After https://github.com/solana-labs/solana/pull/13320 it's possible to view swap instructions but it's not clear which tokens were swapped.

The table is very long for users with many different tokens, it would be nice to filter for the token I care about

Related issue for expanding inner instructions: https://github.com/solana-labs/solana/issues/11935

#### Proposed Solution
- Add filtering by token
- Show more instruction details inline
- Truncate signatures and token addresses that aren't in the registry yet

PR #14032, opened 2020-12-09
explorer: Token history dropdown filter

oJshua:
#### Problem
There previously was no way to filter by token on the token history card.

#### Summary of Changes
This adds a dropdown filter in the top right corner.

Fixes https://github.com/solana-labs/solana/issues/13323

### DIFF
--- a/explorer/src/components/account/TokenHistoryCard.tsx
+++ b/explorer/src/components/account/TokenHistoryCard.tsx
@@ -47,7 +47,7 @@ import { Link } from "react-router-dom";
 import { Location } from "history";
 import { useQuery } from "utils/url";
 
-const TRUNCATE_TOKEN_DISPLAY = 8;
+const TRUNCATE_TOKEN_DISPLAY = 10;
 
 type InstructionType = {
   name: string;
@@ -304,11 +304,11 @@ const FilterDropdown = ({ filter, toggle, show, tokens }: FilterProps) => {
     };
   };
 
-  const FILTER_OPTIONS: (TokenInfoWithPubkey | null)[] = [null];
+  const filterOptions: (TokenInfoWithPubkey | null)[] = [null];
   const nameLookup: { [mint: string]: string } = {};
 
   tokens.forEach((token) => {
-    FILTER_OPTIONS.push(token);
+    filterOptions.push(token);
     const pubkey = token.info.mint.toBase58();
     nameLookup[pubkey] = formatTokenDisplay(pubkey, cluster);
   });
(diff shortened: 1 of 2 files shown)


oJshua:
filter

http://localhost:3000/address/ErF6ptYb3tKzSZjLKBoTqYJPUkMQtPUeAGo16kGLdGVt/tokens?filter=6P3jBKNtALeFjA4isgXwBMK2sUAnRxnphFSo4ZQTeQyx

explorer/src/components/account/TokenHistoryCard.tsx
```diff
@@ -148,7 +171,20 @@ function TokenHistoryTable({ tokens }: { tokens: TokenInfoWithPubkey[] }) {
...
-  if (mintAndTxs.length === 0) {
+  const filtered = mintAndTxs.filter(({ mint }) => {
+    if (filter === "") {
+      return true;
+    }
+    return mint.toBase58() === filter;
+  });
+
+  React.useEffect(() => {
+    if (!fetching && filtered.length < 1 && !allFoundOldest) {
```
jstarry, line 182: I don't think `allFoundOldest` is appropriate here. Seems like it should be something like "filteredFoundOldest"

explorer/src/components/account/TokenHistoryCard.tsx
```diff
@@ -245,6 +287,73 @@ function TokenHistoryTable({ tokens }: { tokens: TokenInfoWithPubkey[] }) {
...
+    } else {
+      params.set("filter", filter);
+    }
+    return {
+      ...location,
+      search: params.toString(),
+    };
+  };
+
+  const filterOptions: (TokenInfoWithPubkey | null)[] = [null];
```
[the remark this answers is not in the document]
oJshua: I like that idea.

explorer/src/components/account/TokenHistoryCard.tsx
```diff
@@ -100,9 +126,24 @@ function TokenHistoryTable({ tokens }: { tokens: TokenInfoWithPubkey[] }) {
     return history?.data !== undefined;
   });
 
+  const filteredFoundOldest = React.useMemo(() => {
```
jstarry, line 129: I don't think we need to compute both `allFoundOldest` and `filteredFoundOldest`. How about filtering tokens early on? I think it would clean up a lot of the code

oJshua: Okay, I'll look at doing that. I introduced filteredFoundOldest based on one of your other comments, but I agree we can probably filter them earlier in the process.

oJshua: Thanks for the suggestion, I think I just needed to see this from a different perspective.

jstarry (approved):

PR #11750, opened 2020-08-20
Switch programs activation to whole-set based gating

jackcmay:
#### Problem

`get_entered_epoch_callback` isn't called when restoring from snapshots, which is too confusing and error-prone.
Also, we can't simply call it immediately after snapshot restoration because it expects to be called **exactly once at each epoch boundary**

`get_programs` and `get_builtins` returns delta set. i.e. _add these new additions of programs to the current available set at the given epoch_. This works nicely in the ideal world, where we're running the validator since genesis without ever restarting a perfect bug-free validator.

In reality, we must rely on snapshots 99.999% of time. When restoring from snapshots, the delta set doesn't work quite: we don't persist _the current available set_ (namely `bank.message_processor` is effectively `serde(skip)`).

#### Summary of Changes

So, just reflect the reality by making these functions snapshot-friendly by returning whole-set of available programs at the given epoch. And make it callable  from `finish_init()`, which is called after snapshot restoration.

Also, fix a bunch of other dangerous code along the way.

Also, this is intended to be back-port friendly; so the fix is intentionally not exhaustive. Still, `get_entered_epoch_callback` is a bit error-prone. Specifically, it must be **idempotent**. (We could solve this by artificially introducing some intermediate `struct` like `ScheduledBankFeatures` or the like instead of mind-opening way of passing `&mut Bank`).

Deprecates #11736

PR #14042, opened 2020-12-10
Tiny add_native_program bug fixes with cleanups

ryoqun:
#### Problem

the simpler capitalization pr got too big... #13884.

> Also, this fixes a regression bug and a new bug introduced at #13403. In this sense, this pr is a follow up to #13403, #11750.

Especially, this is extractable.

#### Summary of Changes

So, extract it! And bunch of tiny renames and preparation changes for this.

There is no need for gating because add_native_program is seldom touched on the live cluster.

### DIFF
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -10121,7 +10121,7 @@ pub(crate) mod tests {
     }
 
     #[test]
-    fn test_add_native_program_squated_while_not_replacing() {
+    fn test_add_native_program_squatted_while_not_replacing() {
         let (genesis_config, mint_keypair) = create_genesis_config(100_000);
         let bank = Bank::new(&genesis_config);
         let program_id = solana_sdk::pubkey::new_rand();
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -1244,7 +1244,7 @@ impl Bank {
         self.store_account(pubkey, &new_account);
     }
 
-    fn inherit_sysvar_account_balance(&self, old_account: &Option<Account>) -> u64 {
+    fn inherit_specially_retained_account_balance(&self, old_account: &Option<Account>) -> u64 {
         old_account.as_ref().map(|a| a.lamports).unwrap_or(1)
     }
 
@@ -1332,7 +1332,10 @@ impl Bank {
             unix_timestamp,
         };
         self.update_sysvar_account(&sysvar::clock::id(), |account| {
-            create_account(&clock, self.inherit_sysvar_account_balance(account))
+            create_account(
+                &clock,
+                self.inherit_specially_retained_account_balance(account),
+            )
         });
     }
 
@@ -1343,7 +1346,10 @@ impl Bank {
                 .map(|account| from_account::<SlotHistory>(&account).unwrap())
                 .unwrap_or_default();
             slot_history.add(self.slot());
-            create_account(&slot_history, self.inherit_sysvar_account_balance(account))
+            create_account(
+                &slot_history,
+                self.inherit_specially_retained_account_balance(account),
+            )
         });
     }
 
@@ -1354,7 +1360,10 @@ impl Bank {
                 .map(|account| from_account::<SlotHashes>(&account).unwrap())
                 .unwrap_or_default();
             slot_hashes.add(self.parent_slot, self.parent_hash);
-            create_account(&slot_hashes, self.inherit_sysvar_account_balance(account))
+            create_account(
+                &slot_hashes,
+                self.inherit_specially_retained_account_balance(account),
+            )
         });
     }
 
@@ -1398,7 +1407,7 @@ impl Bank {
         self.update_sysvar_account(&sysvar::fees::id(), |account| {
             create_account(
                 &sysvar::fees::Fees::new(&self.fee_calculator),
-                self.inherit_sysvar_account_balance(account),
+                self.inherit_specially_retained_account_balance(account),
             )
         });
     }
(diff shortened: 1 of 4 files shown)


runtime/src/bank.rs
```diff
@@ -1244,7 +1244,7 @@ impl Bank {
         self.store_account(pubkey, &new_account);
     }
 
-    fn inherit_sysvar_account_balance(&self, old_account: &Option<Account>) -> u64 {
+    fn inherit_specially_retained_account_balance(&self, old_account: &Option<Account>) -> u64 {
```
ryoqun, line 1247: rename prep for #13884

runtime/src/bank.rs
```diff
@@ -2026,13 +2039,16 @@ impl Bank {
                 // flush the Stakes cache
                 account.lamports = 0;
                 self.store_account(&program_id, &account);
+                None
```
ryoqun, line 2042: wrote a new test case specifically for this: (test_add_native_program_squatted_while_not_replacing)

runtime/src/bank.rs
```diff
@@ -2074,7 +2090,10 @@ impl Bank {
         );
 
         // Add a bogus executable native account, which will be loaded and ignored.
-        let account = native_loader::create_loadable_account(name);
+        let account = native_loader::create_loadable_account(
+            name,
+            self.inherit_specially_retained_account_balance(&existing_genuine_program),
```
ryoqun, line 2095: this new inherit_specially_retained_account_balance call is justified by this: (test_add_native_program_inherited_cap_while_replacing)

runtime/src/bank.rs
```diff
@@ -8523,13 +8591,34 @@ pub(crate) mod tests {
         assert!(bank.stakes.read().unwrap().vote_accounts().is_empty());
         assert!(bank.stakes.read().unwrap().stake_delegations().is_empty());
         assert_eq!(bank.calculate_capitalization(), bank.capitalization());
+        assert_eq!(
```
ryoqun, line 8594: these additional assertions to a test written at #11750 exposes a regression introduced at #13403.

runtime/src/bank.rs
```diff
@@ -7849,65 +7870,112 @@ pub(crate) mod tests {
 
         // First, initialize the clock sysvar
         let bank1 = Arc::new(Bank::new(&genesis_config));
-        bank1.update_sysvar_account(&dummy_clock_id, |optional_account| {
-            assert!(optional_account.is_none());
+        assert_capitalization_diff(
```
ryoqun, line 7873: best viewed with ignore whitespace mode in github

runtime/src/bank.rs
ryoqun: also improved existing test here: https://github.com/solana-labs/solana/pull/14042/files#r540367566

runtime/src/bank.rs
ryoqun: bug description (re-post from https://github.com/solana-labs/solana/pull/13403#discussion_r535955941):

 > Also, when replacing, we must consider the balance of existing program account otherwise we would cause capitalization mismatch if the replaced native program isn't 1 lamports.

ryoqun: Also, I later noticed that we can meaninglessly send to sysvars, but not to executable accounts. So, this can't happen on the wild. But, addressing and adding tests should never hurt us not much. :)

runtime/src/bank.rs
```diff
@@ -7849,65 +7870,112 @@ pub(crate) mod tests {
...
+                            slot: expected_previous_slot,
+                            ..Clock::default()
+                        },
+                        bank1.inherit_specially_retained_account_balance(optional_account),
+                    )
+                })
+            },
+            |old, new| {
+                // creating new sysvar twice in a slot shouldn't increment capitalization twice
+                assert_eq!(old, new);
```
ryoqun, line 7915: for example, this and that (https://github.com/solana-labs/solana/pull/14042/files#diff-ed47b4a0198313377e091bb3957bbbc63d937805426d1b2b6de39d0a50d32a0cR10071) writing style differences are somewhat intentional for easier diff of upcoming #13884 .

runtime/src/bank.rs
ryoqun: bug description (re-post from https://github.com/solana-labs/solana/pull/13403/files#r535953548):

>  well this logic changes introduced a regression where we could be tricked into skipping to create native program because `existing_genuine_program` can be set to a bad account and not reset and we enters this `match` arm returning from this `fn`: https://github.com/solana-labs/solana/pull/13403/files#diff-ed47b4a0198313377e091bb3957bbbc63d937805426d1b2b6de39d0a50d32a0cR1760

runtime/src/bank.rs
```diff
@@ -2014,9 +2023,13 @@ impl Bank {
 
     // NOTE: must hold idempotent for the same set of arguments
     pub fn add_native_program(&self, name: &str, program_id: &Pubkey, must_replace: bool) {
-        let mut existing_genuine_program = self.get_account(&program_id);
-        if let Some(account) = &mut existing_genuine_program {
-            if !native_loader::check_id(&account.owner) {
+        let existing_genuine_program = if let Some(mut account) = self.get_account(&program_id) {
+            // it's very unlikely to be squatted at program_id as non-system account because of burden to
+            // find victim's pubkey/hash. So, when account.owner is indeed native_loader's, it's
+            // safe to assume it's a genuine program.
```
ryoqun, line 2029: added some explanation.

runtime/src/bank.rs
ryoqun: in other words, we shouldn't keep around fake account as `Some(existing_genuine_program)`. It would make us skip native account creation (= `must_replace = false`) So, return None here.

Previously, it only cleared the bogus account...

ryoqun:
@CriesofCarrots Could you review this? thanks!

CriesofCarrots (approved):
Nice cleanup!

PR #14050, opened 2020-12-11
Check shred type in is_duplicate

sakridge:
#### Problem

Duplicate path doesn't check the shred type and looks only in the data column.

#### Summary of Changes

Check data or coding column for the right duplicate shred.

Fixes #


carllin (approved):
😃

PR #14062, opened 2020-12-11
Filter out stake and vote accounts with incorrect owners

CriesofCarrots:
#### Problem
Follow up to #13615
While the stake program now checks whether a vote account is correctly owned on Delegation, the runtime method to collect stake and vote accounts doesn't verify whether this is still the case.

#### Summary of Changes
Filter out invalid stake and vote accounts

### DIFF
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -1644,6 +1644,16 @@ impl Bank {
                             && (stake_account.owner != solana_stake_program::id()
                                 || vote_account.owner != solana_vote_program::id())
                         {
+                            datapoint_warn!(
+                                "bank-stake_delegation_accounts-invalid-account",
+                                ("slot", self.slot() as i64, i64),
+                                ("stake-address", format!("{:?}", stake_pubkey), String),
+                                (
+                                    "vote-address",
+                                    format!("{:?}", delegation.voter_pubkey),
+                                    String
+                                ),
+                            );
                             return;
                         }
                         let entry = accounts
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -1610,6 +1610,7 @@ impl Bank {
     ///  returns a map (has to be copied) of loaded
     ///   ( Vec<(staker info)> (voter account) ) keyed by voter pubkey
     ///
+    /// Filters out invalid pairs
     fn stake_delegation_accounts(
         &self,
         reward_calc_tracer: &mut Option<impl FnMut(&RewardCalculationEvent)>,
@@ -1627,19 +1628,24 @@ impl Bank {
                     self.get_account(&delegation.voter_pubkey),
                 ) {
                     (Some(stake_account), Some(vote_account)) => {
-                        let vote_owner = vote_account.owner;
-                        let entry = accounts
-                            .entry(delegation.voter_pubkey)
-                            .or_insert((Vec::new(), vote_account));
+                        // call tracer to catch any illegal data if any
                         if let Some(reward_calc_tracer) = reward_calc_tracer {
                             reward_calc_tracer(&RewardCalculationEvent::Staking(
                                 stake_pubkey,
                                 &InflationPointCalculationEvent::Delegation(
                                     *delegation,
-                                    vote_owner,
+                                    vote_account.owner,
                                 ),
                             ));
                         }
+                        if stake_account.owner != solana_stake_program::id()
+                            || vote_account.owner != solana_vote_program::id()
+                        {
+                            return;
+                        }
+                        let entry = accounts
+                            .entry(delegation.voter_pubkey)
+                            .or_insert((Vec::new(), vote_account));
                         entry.0.push((*stake_pubkey, stake_account));
                     }
                     (_, _) => {}
@@ -11350,7 +11356,7 @@ pub(crate) mod tests {
             &stake_account,
         );
 
-        // Accounts eligible for inclusion in rewards calculations must be valid
+        // Accounts must be valid stake and vote accounts
         let stake_delegation_accounts = bank.stake_delegation_accounts(&mut null_tracer());
         assert_eq!(stake_delegation_accounts.len(), 0);
     }
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -11343,19 +11343,28 @@ pub(crate) mod tests {
         let stake_delegation_accounts = bank.stake_delegation_accounts(&mut null_tracer());
         assert_eq!(stake_delegation_accounts.len(), 2);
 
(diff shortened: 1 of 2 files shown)


runtime/src/bank.rs
```diff
@@ -1627,19 +1628,27 @@ impl Bank {
...
+                                    vote_account.owner,
                                 ),
                             ));
                         }
+                        if self
+                            .feature_set
+                            .is_active(&feature_set::filter_stake_delegation_accounts::id())
+                            && (stake_account.owner != solana_stake_program::id()
+                                || vote_account.owner != solana_vote_program::id())
+                        {
```
CriesofCarrots, line 1646: @ryoqun what reporting do you think would be most helpful here?

runtime/src/bank.rs
```diff
@@ -1610,6 +1610,7 @@ impl Bank {
     ///  returns a map (has to be copied) of loaded
     ///   ( Vec<(staker info)> (voter account) ) keyed by voter pubkey
     ///
+    /// Filters out invalid pairs
```
ryoqun, line 1613: :+1:

runtime/src/bank.rs
```diff
@@ -11305,4 +11314,53 @@ pub(crate) mod tests {
...
+            .unwrap_or_default();
+        stake_account.owner = bogus_stake_program;
+        bank.store_account(
+            &validator_vote_keypairs1.stake_keypair.pubkey(),
+            &stake_account,
+        );
+
+        // Accounts must be valid stake and vote accounts
+        let stake_delegation_accounts = bank.stake_delegation_accounts(&mut null_tracer());
+        assert_eq!(stake_delegation_accounts.len(), 0);
```
ryoqun, line 11364: :+1:

runtime/src/bank.rs
ryoqun: hmm, I think just `error!()` suffices. At least this is more significant than `warn!`, imo.

CriesofCarrots: That seems a bit strong to me; it's not really a node or runtime error, is it? Would a metric datapoint reporting the slot and the address perhaps be more useful?

runtime/src/bank.rs
```diff
@@ -11305,4 +11314,53 @@ pub(crate) mod tests {
...
+        assert_eq!(stake_delegation_accounts.len(), 2);
+
+        // Modify staked vote account owner; a vote account owned by another program could be
+        // freely modified with malicious data
+        let bogus_vote_program = Pubkey::new_unique();
+        let mut vote_account = bank
+            .get_account(&validator_vote_keypairs0.vote_keypair.pubkey())
+            .unwrap_or_default();
+        vote_account.owner = bogus_vote_program;
+        bank.store_account(
```
ryoqun, line 11343: I think runtime enforces an intermediately write of account.lamports == 0 when chainging existing non-system accounts.
So, how about inserting this here?:

```rust
// simulate vote account removal via full withdrawal
bank.store_account(&validator_vote_keypairs0.vote_keypair.pubkey(), &Acount::default());
```

Then, the following assertion should be `1`: 

```rust
assert_eq!(bank.vote_accounts().len(), 1);
```

runtime/src/bank.rs
ryoqun: > That seems a bit strong to me; it's not really a node or runtime error, is it? Would a metric datapoint reporting the slot and the address perhaps be more useful?

Hhhm, yeah, I agree. After more thought, I think we can go with your suggested approach.

ryoqun:
@CriesofCarrots btw, did you tested this manually?:

Well, I have a secret sauce of easily creating fake account via the wracked `solana deploy` command:

``` bash
$ ./target/release/solana --url http://localhost:8899 account -o vote.data <foo bar pubkey json>
$ hexedit
$ ./target/release/solana.bad-deploy deploy --url http://localhost:8899/ vote.data <foo bar pubkey json>
```

```
diff --git a/cli/src/cli.rs b/cli/src/cli.rs
index aa7b69fb7c..e1ee8aa66b 100644
--- a/cli/src/cli.rs
+++ b/cli/src/cli.rs
@@ -1227,8 +1227,8 @@ fn do_process_deploy(
         CliError::DynamicProgramError(format!("Unable to read program file: {}", err))
     })?;
 
-    Executable::<BPFError>::from_elf(&program_data, Some(|x| bpf_verifier::check(x, true)))
-        .map_err(|err| CliError::DynamicProgramError(format!("ELF error: {}", err)))?;
+    //Executable::<BPFError>::from_elf(&program_data, Some(|x| bpf_verifier::check(x, true)))
+    //    .map_err(|err| CliError::DynamicProgramError(format!("ELF error: {}", err)))?;
 
     let loader_id = if use_deprecated_loader {
         bpf_loader_deprecated::id()

```

this is not strict requirement, though. :)

ryoqun (dismissed):
lgtm after metrics addition and small test tweaks!

Really thanks for taking this over. :)

ryoqun:
> Well, I have a secret sauce...

@CriesofCarrots And this for faster inflation:


```
SOLANA_RUN_SH_CLUSTER_TYPE=mainnet-beta/development SOLANA_RUN_SH_GENESIS_ARGS="--slots-per-epoch 128 --vote-commission-percentage=55" ./run.sh
```

runtime/src/bank.rs
CriesofCarrots: That is more realistic, but it doesn't change `bank.vote_accounts().len()`. (Confirmed with test run.)
Unless I'm missing something, I don't see that Stakes ever removes vote_accounts, regardless of owner or lamports balance.

ryoqun: well, sorry, my comment was partly wrong.

Account::default().owner isn't Vote1111; that's why it didn't change.

So, the correct code should be like this:

```
vote_account.lamports = 0;
// simulate vote account removal via full withdrawal
bank.store_account(&validator_vote_keypairs0.vote_keypair.pubkey(), &vote_account);

vote_account.lamports = original_lamports;
vote_account.owner = bogus_vote_program;

bank.store_account(&validator_vote_keypairs0.vote_keypair.pubkey(), &vote_account);
assert_eq!(bank.vote_accounts().len(), 1);

```

Well, Stake::store is hard to read without much aid; Please check this out: https://github.com/solana-labs/solana/pull/14065/files#r540752554 I still think it should remove `vote_accounts`.

CriesofCarrots: Oh thank you, I did miss that all owned vote accounts are removed and then potentially re-added.
Okay, great, your suggested code looks just like mine does now :)

CriesofCarrots:
@ryoqun thanks for the special sauce! It took a few more steps, but I was able to test this manually. Merging!

PR #14065, opened 2020-12-11
Add more comments in Stakes::store for its subtlety

ryoqun:
#### Problem

Stakes::store() does its elegance without much murmuring regarding the handling of proper stale state prevention for stakes and votes.

#### Summary of Changes

Make it chatty for peaceful future humanity.

no functional change.

Fixes #

### DIFF
--- a/runtime/src/stakes.rs
+++ b/runtime/src/stakes.rs
@@ -123,7 +123,13 @@ impl Stakes {
         fix_stake_deactivate: bool,
     ) -> Option<ArcVoteAccount> {
         if solana_vote_program::check_id(&account.owner) {
+            // unconditionally remove existing at first; there is no dependent calculated state for
+            // votes, not like stakes (stake codepath maintains calculated stake value grouped by
+            // delegated vote pubkey)
             let old = self.vote_accounts.remove(pubkey);
+            // when account is removed (lamports == 0), don't readd so that given
+            // `pubkey` can be used for any owner in the future, while not
+            // affecting Stakes.
             if account.lamports != 0 {
                 let stake = old.as_ref().map_or_else(
                     || {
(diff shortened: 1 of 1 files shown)


runtime/src/stakes.rs
```diff
@@ -123,7 +123,13 @@ impl Stakes {
         fix_stake_deactivate: bool,
     ) -> Option<ArcVoteAccount> {
         if solana_vote_program::check_id(&account.owner) {
+            // unconditionally remove existing at first; there is no dependent calculated state for
+            // votes, not like stakes (stake codepath maintains calculated stake value grouped by
+            // delegated vote pubkey)
             let old = self.vote_accounts.remove(pubkey);
```
ryoqun, line 129: here.

runtime/src/stakes.rs
```diff
@@ -182,12 +191,19 @@ impl Stakes {
...
+                // when account is removed (lamports == 0), remove it from Stakes as well
+                // so that given `pubkey` can be used for any owner in the future, while not
+                // affecting Stakes.
                 self.stake_delegations.remove(pubkey);
             } else if let Some(delegation) = delegation {
                 self.stake_delegations.insert(*pubkey, delegation);
             }
             None
         } else {
+            // there is no need to remove possibly existing Stakes cache entries with given
```
ryoqun, line 203: @CriesofCarrots `Account::default()` was hitting here in #14062.

ryoqun:
@bennofs I hope you glanced this code, considering #14062 finding. Could you check my sanity as a neutral outsider? There should be no hole anymore for Stakes with #14062 (@CriesofCarrots thanks for working on this) :)

runtime/src/stakes.rs
CriesofCarrots: 💯

PR #10099, opened 2020-05-18
Optimize stale slot shrinking for previously cleaned roots

ryoqun:
### Problem

There are several problems for eager rent collection and stale slot shrinking...

#### stale slot shrinking (perf. degradation)

Shrinking is causing trouble under load or benchmark.

First, shrinking (reading & writing accounts) a slot full of account updates isn't a light-weight background task in any way, even not _stale_ at all. But, that's constantly happening too often while benchmarking. That's partly because supposedly our bench is repeatedly updating same accounts from some limited (large) number of test account sets. So, that heavy slot is wrongly passing the existing _stale_ check too easily. (20% of accounts are outdated, meaning unused/empty/shrinkable). Yeah, that's plain bug when stale slot shrinking was introduced. Sorry...

Also, bench might not be representational real world work load. But, this still bothers us and indeed exposes the potential perf problem in stale slot shrinking. And, shrinking probably will be hurting the performance under real-world saturating tps situation.

#### eager rent collection (bloated snapshot)

The snapshots are too prone to grow when there are theoretically maximum number of alive `AppendVec`s. This is currently happening only on devnet, because there are so many rent-exempt accounts on devnet (1.5 million accounts; I counted from a sample recent snapshot). That huge number of AppendVec is a side-effect of the fact eager rent collection is spreading those accounts over 2 days worth of slots (an epoch on tds/mainnet-beta), so thinly (well, that's intended design...).

When, there is such large number of AppendVecs, the stale slot shrinking works poorly because the traversing algorithm is simplistic. It was originally exactly designed for the _stale_ slots. However, eager rent collection creates fresh AppendVecs periodically now on devnet at too fast pace, making the _stale_ assumption obsolete.

The algorithm is like this: grab all of rooted alive slot set from AccountsIndex at a time and loop over the set (without fetching new roots) until the end of the set at 100ms interval, and repeat the process infinitely. This means when there are 400K slots, it doesn't shrink newly- (and constantly- by eager rent collection) created slots for 11 hours.

Eager rent collection constantly creates 4MB `AppendVec`s with few rent-collected accounts, which should be shrunk but not done, causing bloated snapshot. Those few accounts prevents the AppendVec from being reclaimed quickly (usually reclaimed for the idling cluster before eager rent collection). Then, the outdated garbage sysvars in the AppendVec is nightmare for `bzip2`; 130kb of slightly changing bits for each `AppendVec`.

[...]

PR #14072, opened 2020-12-11
Shrink slots by sparseness of written data size

ryoqun:
#### Problem

Shrinker doesn't trigger and causes indefinite AppendVec leaks when there is consistent write-once only transactions...

Eventually, this causes max_mmap_count. Also, this makes snapshots contain very large number of sparse files; so snapshot is bloated and snapshot creation takes more processing.

#### Summary of Changes

Be more aggressive by also considering the bytes sparseness.

Also note that this is a quick and dirty band aid;  The root problem is soon be solved by @carllin 's different pr.

follow up: #10099

### DIFF
--- a/runtime/src/accounts_background_service.rs
+++ b/runtime/src/accounts_background_service.rs
@@ -162,10 +162,15 @@ impl AccountsBackgroundService {
                     assert!(last_cleaned_block_height <= snapshot_block_height);
                     last_cleaned_block_height = snapshot_block_height;
                 } else {
-                    consumed_budget = bank.process_stale_slot_with_budget(
-                        consumed_budget,
-                        SHRUNKEN_ACCOUNT_PER_INTERVAL,
-                    );
+                    // under sustained writes, shrink can lag behind so cap to
+                    // SHRUNKEN_ACCOUNT_PER_INTERVAL (which is based on INTERVAL_MS,
+                    // which in turn roughly asscociated block time)
+                    consumed_budget = bank
+                        .process_stale_slot_with_budget(
+                            consumed_budget,
+                            SHRUNKEN_ACCOUNT_PER_INTERVAL,
+                        )
+                        .min(SHRUNKEN_ACCOUNT_PER_INTERVAL);
 
                     if bank.block_height() - last_cleaned_block_height
                         > (CLEAN_INTERVAL_BLOCKS + thread_rng().gen_range(0, 10))
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -266,6 +266,14 @@ impl AccountStorageEntry {
         self.approx_store_count.load(Ordering::Relaxed)
     }
 
+    pub fn written_bytes(&self) -> u64 {
+        self.accounts.len() as u64
+    }
+
+    pub fn total_bytes(&self) -> u64 {
+        self.accounts.capacity()
+    }
+
     pub fn has_accounts(&self) -> bool {
         self.count() > 0
     }
@@ -1019,9 +1027,13 @@ impl AccountsDB {
                 let stores = stores_lock.read().unwrap();
                 let mut alive_count = 0;
                 let mut stored_count = 0;
+                let mut written_bytes = 0;
+                let mut total_bytes = 0;
                 for store in stores.values() {
                     alive_count += store.count();
                     stored_count += store.approx_stored_count();
+                    written_bytes += store.written_bytes();
+                    total_bytes += store.total_bytes();
                 }
                 if alive_count == stored_count && stores.values().len() == 1 {
                     trace!(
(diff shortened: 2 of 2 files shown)


runtime/src/accounts_db.rs
```diff
@@ -266,6 +266,21 @@ impl AccountStorageEntry {
         self.approx_store_count.load(Ordering::Relaxed)
     }
 
+    // .len()/.is_empty()/.capacity() is mimicing std::collection api
+    pub fn len(&self) -> u64 {
```
ryoqun, line 270: well, this is in interms of bytes; maybe, mimicing the api doesn't make sense; I'll rename later.

runtime/src/accounts_db.rs
```diff
@@ -1032,14 +1051,20 @@ impl AccountsDB {
...
-                        slot,
-                        alive_count,
-                        stored_count,
-                    );
-                    return 0;
+                } else if !forced {
+                    let sparse_by_count = (alive_count as f32 / stored_count as f32) <= 0.8;
+                    let sparse_by_space = (written_space as f32 / total_space as f32) <= 0.8;
+                    if !sparse_by_count && !sparse_by_space {
+                        trace!(
```
ryoqun, line 1058: hmm, should I change to `info!`; not sure.. time to bloat our INFO logs further...? ;)

runtime/src/hardened_unpack.rs
```diff
@@ -38,7 +38,7 @@ const MAX_SNAPSHOT_ARCHIVE_UNPACKED_APPARENT_SIZE: u64 = 64 * 1024 * 1024 * 1024
 // This is the actually consumed disk usage for sparse files
 const MAX_SNAPSHOT_ARCHIVE_UNPACKED_ACTUAL_SIZE: u64 = 4 * 1024 * 1024 * 1024 * 1024;
 
-const MAX_SNAPSHOT_ARCHIVE_UNPACKED_COUNT: u64 = 500_000;
+const MAX_SNAPSHOT_ARCHIVE_UNPACKED_COUNT: u64 = 5_000_000;
```
ryoqun, line 41: stolen from #14071

carllin:
cool, thanks for making the fix!

runtime/src/accounts_db.rs
ryoqun: yeah, I took the bitter path. ;)

ryoqun: yeah, I took the bitter path. ;)

runtime/src/hardened_unpack.rs
ryoqun: rebased

runtime/src/accounts_background_service.rs
```diff
@@ -162,10 +162,15 @@ impl AccountsBackgroundService {
                     assert!(last_cleaned_block_height <= snapshot_block_height);
                     last_cleaned_block_height = snapshot_block_height;
                 } else {
-                    consumed_budget = bank.process_stale_slot_with_budget(
-                        consumed_budget,
-                        SHRUNKEN_ACCOUNT_PER_INTERVAL,
-                    );
+                    // under sustained writes, shrink can lag behind so cap to
+                    // SHRUNKEN_ACCOUNT_PER_INTERVAL (which is based on INTERVAL_MS,
+                    // which in turn roughly asscociated block time)
```
ryoqun, line 167: @carllin In addition to our discuss yesterday, adjusting the budget code is also needed...

runtime/src/accounts_db.rs
```diff
@@ -1032,14 +1044,17 @@ impl AccountsDB {
...
                     return 0;
-                } else if (alive_count as f32 / stored_count as f32) >= 0.80 && !forced {
-                    trace!(
-                        "shrink_stale_slot ({}): not enough space to shrink: {} / {}",
-                        slot,
-                        alive_count,
-                        stored_count,
+                } else if !forced {
+                    let sparse_by_count = (alive_count as f32 / stored_count as f32) <= 0.8;
+                    let sparse_by_bytes = (written_bytes as f32 / total_bytes as f32) <= 0.8;
```
ryoqun, line 1049: the reason I intentionally didn't this at the first implementation is that I didn't care about unused mmaps. But, well, it turned out address space is scarce resource indeed... ;)

ryoqun:
here is side-by-side comparison (I restarted patched and unpatched validators at the same time):

note last restart (rightmost line):

BEFORE (BAD; LEAKING; very steep line)

image

AFTER (GOOD; NOT LEAKING; moderately steep line; gradual increase is expected while capped 420000)

image

ryoqun:
BEFORE (less frequent shrinking)

image

AFTER (more frequent and slightly increased shrink count due to more aggressive budgeting)

image

ryoqun:
snapshot size

BEFORE (more steep increase)

image


AFTER (less steep increase)

image

ryoqun:
@carllin Yeah, I've finished it. could you review this?

runtime/src/accounts_background_service.rs
```diff
@@ -162,10 +162,15 @@ impl AccountsBackgroundService {
...
+                    consumed_budget = bank
+                        .process_stale_slot_with_budget(
+                            consumed_budget,
+                            SHRUNKEN_ACCOUNT_PER_INTERVAL,
+                        )
+                        .min(SHRUNKEN_ACCOUNT_PER_INTERVAL);
```
carllin, line 173: Since:
1) `consumed_budget` is always at most `SHRUNKEN_ACCOUNT_PER_INTERVAL`, 
2) `process_stale_slot_with_budget` cleans only when `consumed_budget == 0`
3) `process_stale_slot_with_budget` always decreases `consumed_budget` by `SHRUNKEN_ACCOUNT_PER_INTERVAL`.

Then if there are a constant stream of shrinkable slots, can this `process_stale_slot_with_budget()`  be simplified to just shrink every 2 calls?

carllin (approved):

runtime/src/accounts_background_service.rs
ryoqun: @carllin hehe, too good eye. xD Yeah, your explanation is right. I've just minimized risk by not rewriting the budgeting logic....

PR #14083, opened 2020-12-11
Cleanup ledger builtins

jackcmay:
#### Problem

New BPF lead is enabled everywhere so don't block it from mainnet in genesis

#### Summary of Changes

Part of default genesis now, removed from features builtin

Fixes #


jackcmay:
@ryoqun This look correct to you?

ryoqun:
@jackcmay code looks harmless to me. If you can start mainnet-beta with this pr without problems, this pr is LGTM! Thanks for cleaning.

ryoqun:
also, sorry... it seems that I couldn't catch this broke testnet while reviewing (ref: #14147)

PR #14088, opened 2020-12-12
indexes crds values associated with a pubkey

behzadnouri:
#### Problem
record_labels returns all the possible labels for a record identified by
a pubkey, used in updating timestamp of crds values:
https://github.com/solana-labs/solana/blob/1792100e2/core/src/crds_value.rs#L560-L577
https://github.com/solana-labs/solana/blob/1792100e2/core/src/crds.rs#L240-L251
The code relies on CrdsValueLabel to be limited to a small deterministic
set of possible values for a fixed pubkey. As we expand crds values to
include duplicate shreds, this limits what the duplicate proofs can be
keyed by in the table.
In addition the computation of these labels is inefficient and will
become more so as duplicate shreds and more types of crds values are
added. An alternative is to maintain an index of all crds values
associated with a pubkey.

#### Summary of Changes
Added an index for crds values associated with a pubkey.

### DIFF
--- a/core/src/crds.rs
+++ b/core/src/crds.rs
@@ -36,8 +36,8 @@ use solana_sdk::pubkey::Pubkey;
 use solana_sdk::signature::Keypair;
 use solana_sdk::timing::timestamp;
 use std::cmp;
-use std::collections::HashMap;
-use std::ops::Index;
+use std::collections::{hash_map, HashMap};
+use std::ops::{Index, IndexMut};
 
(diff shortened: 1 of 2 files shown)


behzadnouri:
```cluster_info_stats2.process_pull_resp:```
records_index cluster_info_stats2 process_pull_resp
```cluster_info_stats3.process_pull_requests:```
records_index cluster_info_stats3 process_pull_requests

sakridge (approved):
lgtm

PR #14096, opened 2020-12-13
Add accounts cluster bench

sakridge:
#### Problem

No standalone program for creating many accounts

#### Summary of Changes

Add one

Fixes #

### DIFF
--- a/accounts-cluster-bench/src/main.rs
+++ b/accounts-cluster-bench/src/main.rs
@@ -1,3 +1,4 @@
+#![allow(clippy::integer_arithmetic)]
 use clap::{crate_description, crate_name, value_t, value_t_or_exit, App, Arg};
 use log::*;
 use rand::{thread_rng, Rng};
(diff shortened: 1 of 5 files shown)


accounts-cluster-bench/src/main.rs
```diff
@@ -0,0 +1,549 @@
...
+            id.pubkey(),
+        );
+
+        let (blockhash, _fee_calculator) = client.get_recent_blockhash().unwrap();
+        match request_airdrop_transaction(&faucet_addr, &id.pubkey(), airdrop_amount, blockhash) {
+            Ok(transaction) => {
+                let mut tries = 0;
+                loop {
+                    tries += 1;
+                    let signature = client.send_transaction(&transaction).unwrap();
```
CriesofCarrots, line 46: In practice, I'm seeing intermittent panics on this line on `Transaction simulation failed: The bank has seen this signature before`. Presumably the transaction sent in the previous loop iteration is actually landing due to send-tx-service retries.

CriesofCarrots: I'm going to try polling for SingleGossip commitment, which seems sufficient and less likely to hit the 15s timeout:
```
let result = client.poll_for_signature_with_commitment(&signature, CommitmentConfig::single_gossip());
```

PR #6799, opened 2019-11-07
Fix freeze and register_tick race

carllin:
#### Problem
https://github.com/solana-labs/solana/pull/6013 fixed the race, but was broken again in later changes.

#### Summary of Changes
Refix #6003 

Fixes #

PR #14101, opened 2020-12-14
Fix race between setting tick height and calculating accounts hash

carllin:
#### Problem
Freeze in ReplayStage only checks the tick height: https://github.com/solana-labs/solana/blob/master/core/src/replay_stage.rs#L1339-L1348), so it's possible that the the accounts diff hash is calculated after freeze, but before the block hash sysvar is updated. Like this:
```
let current_tick_height = self.tick_height.fetch_add(1, Relaxed) as u64;
<<< Freeze happens here>>>
if self.is_block_boundary(current_tick_height + 1) {
            w_blockhash_queue.register_hash(hash, &self.fee_calculator);
            if self.fix_recent_blockhashes_sysvar_delay() {
                self.update_recent_blockhashes_locked(&w_blockhash_queue);
            }
        }
```

#### Summary of Changes
Move blockhash registration before tick registration.

TODO: Testing and double checking subtleties in old related PR's that I may have forgotten about:

1) https://github.com/solana-labs/solana/pull/6013
2) https://github.com/solana-labs/solana/pull/6799

Fixes #

### DIFF
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -2277,9 +2277,6 @@ impl Bank {
         );
 
         inc_new_counter_debug!("bank-register_tick-registered", 1);
-        // Grab blockhash lock before incrementing tick height so that replay stage does
-        // not attempt to freeze after observing the last tick and before blockhash is
-        // updated
         let mut w_blockhash_queue = self.blockhash_queue.write().unwrap();
         if self.is_block_boundary(self.tick_height.load(Relaxed) + 1) {
             w_blockhash_queue.register_hash(hash, &self.fee_calculator);
@@ -2287,7 +2284,12 @@ impl Bank {
                 self.update_recent_blockhashes_locked(&w_blockhash_queue);
             }
         }
-        let current_tick_height = self.tick_height.fetch_add(1, Relaxed) as u64;
+        // ReplayStage will start computing the accounts delta hash when it
+        // detects the tick height has reached the boundary, so the system
+        // needs to guarantee all account updates for the slot have been
+        // committed before this tick height is incremented (like the blockhash
+        // sysvar above)
+        self.tick_height.fetch_add(1, Relaxed);
     }
 
     pub fn is_complete(&self) -> bool {
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -2281,13 +2281,13 @@ impl Bank {
         // not attempt to freeze after observing the last tick and before blockhash is
         // updated
         let mut w_blockhash_queue = self.blockhash_queue.write().unwrap();
-        let current_tick_height = self.tick_height.fetch_add(1, Relaxed) as u64;
-        if self.is_block_boundary(current_tick_height + 1) {
+        if self.is_block_boundary(self.tick_height.load(Relaxed) + 1) {
             w_blockhash_queue.register_hash(hash, &self.fee_calculator);
             if self.fix_recent_blockhashes_sysvar_delay() {
                 self.update_recent_blockhashes_locked(&w_blockhash_queue);
             }
         }
+        let current_tick_height = self.tick_height.fetch_add(1, Relaxed) as u64;
     }
 
     pub fn is_complete(&self) -> bool {
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -867,6 +867,8 @@ pub struct Bank {
     pub feature_set: Arc<FeatureSet>,
 
     pub drop_callback: RwLock<OptionalDropCallback>,
+
+    pub freeze_started: AtomicBool,
 }
 
 impl Default for BlockhashQueue {
@@ -1022,6 +1024,7 @@ impl Bank {
                     .as_ref()
                     .map(|drop_callback| drop_callback.clone_box()),
             )),
+            freeze_started: AtomicBool::new(false),
         };
 
         datapoint_info!(
@@ -1145,6 +1148,7 @@ impl Bank {
             transaction_log_collector: new(),
             feature_set: new(),
             drop_callback: RwLock::new(OptionalDropCallback(None)),
+            freeze_started: AtomicBool::new(fields.hash != Hash::default()),
         };
         bank.finish_init(genesis_config, additional_builtins);
 
@@ -1246,6 +1250,10 @@ impl Bank {
 
     pub fn is_frozen(&self) -> bool {
         *self.hash.read().unwrap() != Hash::default()
+    }
+
+    pub fn freeze_started(&self) -> bool {
+        self.freeze_started.load(Relaxed)
     }
 
     pub fn status_cache_ancestors(&self) -> Vec<u64> {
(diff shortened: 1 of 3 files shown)


ryoqun:
Thanks for fixing this!

(Also, I wonder why this problem happens so often at our internal testing than at any of live clusters? Naturally, this fix explain that?)

sakridge:
Nice, this seems to make sense. Is there an assert we can add on doing a store on a frozen bank that makes sense to catch this kind of thing earlier?

runtime/src/bank.rs
```diff
@@ -2268,13 +2268,13 @@ impl Bank {
...
         let mut w_blockhash_queue = self.blockhash_queue.write().unwrap();
-        let current_tick_height = self.tick_height.fetch_add(1, Relaxed) as u64;
-        if self.is_block_boundary(current_tick_height + 1) {
+        if self.is_block_boundary(self.tick_height.load(Relaxed) + 1) {
             w_blockhash_queue.register_hash(hash, &self.fee_calculator);
             if self.fix_recent_blockhashes_sysvar_delay() {
                 self.update_recent_blockhashes_locked(&w_blockhash_queue);
             }
         }
+        let current_tick_height = self.tick_height.fetch_add(1, Relaxed) as u64;
```
sakridge, line 2277: Maybe a comment about why this needs to be last in the function?

carllin:
> Nice, this seems to make sense. Is there an assert we can add on doing a store on a frozen bank that makes sense to catch this kind of thing earlier?

Added to AccountsDb to catch frozen stores, what do you think?

carllin:
Been running this on a validator against mainnet, and also a 5 node cluster with 5k tps, no issues thus far.

sakridge:
> > Nice, this seems to make sense. Is there an assert we can add on doing a store on a frozen bank that makes sense to catch this kind of thing earlier?
> 
> Added check: https://github.com/solana-labs/solana/pull/14101/files#diff-1090394420d51617f3233275c2b65ed706b35b53b115fe65f82c682af8134a6fR2653 to AccountsDb to catch frozen stores, what do you think?

How about just similar assert in `Bank::store_account` to this:
```
assert!(
    !self.is_frozen(),
    "commit_transactions() working on a frozen bank!"
);
```

carllin:
@sakridge, yeah I considered something similar, but the issue is `is_frozen() == true` is set only after `*hash = self.hash_internal_state();`.

This means  when the `let accounts_delta_hash = self.rc.accounts.bank_hash_info_at(self.slot());` inside `hash_internal_state()` is being computed, `is_frozen() == false`. Thus if the `store()` happens after the `accounts_delta_hash` computation and before the rest of the `hash_internal_state()` function finishes, the assert wouldn't trigger.

sakridge:
> @sakridge, yeah I considered something similar, but the issue is `is_frozen() == true` is set only after `*hash = self.hash_internal_state();`.
> 
> This means when the `let accounts_delta_hash = self.rc.accounts.bank_hash_info_at(self.slot());` inside `hash_internal_state()` is being computed, `is_frozen() == false`. Thus if the `store()` happens after the `accounts_delta_hash` computation and before the rest of the `hash_internal_state()` function finishes, the assert wouldn't trigger.

I see. `is_frozen: AtomicBool` in the bank?

carllin:
> > @sakridge, yeah I considered something similar, but the issue is `is_frozen() == true` is set only after `*hash = self.hash_internal_state();`.
> > This means when the `let accounts_delta_hash = self.rc.accounts.bank_hash_info_at(self.slot());` inside `hash_internal_state()` is being computed, `is_frozen() == false`. Thus if the `store()` happens after the `accounts_delta_hash` computation and before the rest of the `hash_internal_state()` function finishes, the assert wouldn't trigger.
> 
> I see. `is_frozen: AtomicBool` in the bank?

Hehe nope, it checks the blockhash like so:

```
pub fn is_frozen(&self) -> bool {
        *self.hash.read().unwrap() != Hash::default()
    }
```

I added it as a method on AccountsDb so it would catch it regardless of the storage entry point (i.e. bank.store() vs bank.accounts.store_accounts() in bank.commit_transactions())

sakridge:
> > > @sakridge, yeah I considered something similar, but the issue is `is_frozen() == true` is set only after `*hash = self.hash_internal_state();`.
> > > This means when the `let accounts_delta_hash = self.rc.accounts.bank_hash_info_at(self.slot());` inside `hash_internal_state()` is being computed, `is_frozen() == false`. Thus if the `store()` happens after the `accounts_delta_hash` computation and before the rest of the `hash_internal_state()` function finishes, the assert wouldn't trigger.
> > 
> > 
> > I see. `is_frozen: AtomicBool` in the bank?
> 
> Hehe nope, it checks the blockhash like so:
> 
> ```
> pub fn is_frozen(&self) -> bool {
>         *self.hash.read().unwrap() != Hash::default()
>     }
> ```
> 
> I added it as a method on AccountsDb so it would catch it regardless of the storage entry point (i.e. bank.store() vs bank.accounts.store_accounts() in bank.commit_transactions())

Yea, I get the trade-off. But it is propagating this notion of freezing down to an additional layer and also adding a whole other structure to take care of and prune vs. adding it in those two places which we do a store. Also taking additional locks for duplicate functionality.

I think it might work like this:

```
Bank {
  ..
  is_frozen: AtomicBool,
}
fn is_frozen(&self) -> bool { self.is_frozen.load(..) }
fn is_frozen_and_hashed(&self) -> bool { self.is_frozen() && self.hash.read().unwrap() != default() }
fn freeze(&self) {
    let mut hash = self.hash.write().unwrap();
    store_xyz();
    self.is_frozen.store(true, ..);
    *hash = self.hash_internal_state();
}
```

carllin:
@sakridge I followed your suggestion except I renamed `is_frozen -> freeze_started` and `is_frozen_and_hashed -> is_frozen` (i.e. is_frozen remains unchanged). This is because:

1) In your examples, `self.is_frozen() && self.hash.read().unwrap() != default()` is equivalent to `self.hash.read().unwrap() != default()`, because the  `is_frozen` AtomicBool must be set if `self.hash.read().unwrap() != default()`.
2) Defaulting to the original `is_frozen` seems safer than making the rename change because I am not sure if there are implicit dependencies on blockhash in a few places.

runtime/src/bank.rs
```diff
@@ -3770,6 +3790,7 @@ impl Bank {
     }
 
     pub fn store_account(&self, pubkey: &Pubkey, account: &Account) {
+        assert!(self.freeze_started());
```
carllin, line 3793: Second place we store, added check

runtime/src/bank.rs
```diff
@@ -3055,8 +3075,8 @@ impl Bank {
         signature_count: u64,
     ) -> TransactionResults {
         assert!(
-            !self.is_frozen(),
-            "commit_transactions() working on a frozen bank!"
+            !self.freeze_started(),
```
carllin, line 3078: First place we store, added check

runtime/src/bank.rs
```diff
@@ -1145,6 +1148,7 @@ impl Bank {
             transaction_log_collector: new(),
             feature_set: new(),
             drop_callback: RwLock::new(OptionalDropCallback(None)),
+            freeze_started: AtomicBool::new(fields.hash != Hash::default()),
```
carllin, line 1151: @ryoqun I think this is a safe change to the ABI, and safe functionality-wise

ryoqun: yeah, I think this is ok.

sakridge:
> 
> 
> @sakridge I followed your suggestion: [605f14a](https://github.com/solana-labs/solana/commit/605f14acc86f8ac0ff13e4e7d99756a1d09bb8c4) except I introduced a new ` freeze_started` flag (serves the same purpose as the `is_frozen` AtomicBool in your example), and `is_frozen_and_hashed -> is_frozen` (i.e. `is_frozen` remains unchanged). This is because:
> 
>     1. In your examples, `self.is_frozen() && self.hash.read().unwrap() != default()` is equivalent to `self.hash.read().unwrap() != default()`, because the  `is_frozen` AtomicBool must be set if `self.hash.read().unwrap() != default()`, so I think just having the name `is_frozen()` is sufficient
> 
>     2. Defaulting to the original `is_frozen` seems safer than making the rename change because I am not sure if there are implicit dependencies on blockhash in a few places.

Yes, perfect. Those names are better. Thanks. Key point was just that an atomic load is much cheaper than rwlock+ btree operation.

PR #14119, opened 2020-12-15
Remove regular expressions, speeds up untar

sakridge:
#### Problem

regex is very slow for tar file checking

#### Summary of Changes

Use custom loop, speeds up untar by 3x. 15s => 5s.

Fixes #

### DIFF
--- a/runtime/src/hardened_unpack.rs
+++ b/runtime/src/hardened_unpack.rs
@@ -1,6 +1,5 @@
 use bzip2::bufread::BzDecoder;
 use log::*;
-use regex::Regex;
 use solana_sdk::genesis_config::GenesisConfig;
 use std::{
     fs::{self, File},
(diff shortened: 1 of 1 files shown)


runtime/src/hardened_unpack.rs
```diff
@@ -167,24 +166,46 @@ pub fn unpack_snapshot<A: Read, P: AsRef<Path>>(
...
+                saw_numbers = false;
+                periods += 1;
+            } else {
+                return false;
+            }
+        } else {
+            saw_numbers = true;
+        }
+    }
+    saw_numbers
```
ryoqun, line 196: maybe this needs to be changed `periods == 1 && saw_numbers`?

I think "232323" incorrectly returns false.

runtime/src/hardened_unpack.rs
```diff
@@ -167,24 +166,46 @@ pub fn unpack_snapshot<A: Read, P: AsRef<Path>>(
     )
 }
 
-fn is_valid_snapshot_archive_entry(parts: &[&str], kind: tar::EntryType) -> bool {
-    let like_storage = Regex::new(r"^\d+\.\d+$").unwrap();
-    let like_slot = Regex::new(r"^\d+$").unwrap();
```
ryoqun, line 172: maybe, these regex isn't jit compiled? unfortunate. xD

ryoqun:
> Use custom loop, speeds up untar by 3x. 15s => 5s.

cool. thanks working on this!

runtime/src/hardened_unpack.rs
sakridge: yes, fixed and added test

runtime/src/hardened_unpack.rs
sakridge: Can't say I've ever come across a regex that is truly fast, but maybe :)

ryoqun (approved):
lgtm!

PR #14133, opened 2020-12-15
adds crds-value for broadcasting duplicate shreds through gossip

behzadnouri:
#### Problem
In gossip, the header overhead we get from:
https://github.com/solana-labs/solana/blob/de9ac43eb/core/src/cluster_info.rs#L434-L435
https://github.com/solana-labs/solana/blob/de9ac43eb/core/src/crds_value.rs#L31-L36
https://github.com/solana-labs/solana/blob/de9ac43eb/core/src/crds_value.rs#L73
already exceeds SIZE_OF_NONCE in shreds. We also need aditional
meta-data (wallclock, source pubkey, ...). Which means that given the
SHRED_PAYLOAD_SIZE, we cannot fit all these in PACKET_DATA_SIZE:
https://github.com/solana-labs/solana/blob/de9ac43eb/ledger/src/shred.rs#L80

On top of that, we need 2 shred payloads as the proof of duplicate. So
each DuplicateShred crds value includes only a chunk of the payload,
along with the meta-data to reconstruct the full payload from the chunks
on the receiving end.

#### Summary of Changes
added new crds-value for broadcasting duplicate shreds through gossip.


core/src/duplicate_shred.rs
```diff
@@ -0,0 +1,338 @@
...
+    wallclock: u64,
+    max_size: usize, // Maximum serialized size of each DuplicateShred.
+    encoder: F,
+) -> Result<impl Iterator<Item = DuplicateShred>, Error>
+where
+    F: FnOnce(Vec<u8>) -> Result<Vec<u8>, std::io::Error>,
+{
+    if proof.shred1 == proof.shred2 {
+        return Err(Error::InvalidDuplicateSlotProof);
+    }
```
carllin, line 106: is this necessary with the `check_shreds()` check below?

core/src/duplicate_shred.rs
```diff
@@ -0,0 +1,338 @@
...
+    #[error("shred type mismatch")]
+    ShredTypeMismatch,
+    #[error("slot mismatch")]
+    SlotMismatch,
+    #[error("value out of bounds")]
+    ValueOutOfBounds,
+}
+
+// Verifies that the two shreds can indicate duplicate proof.
+fn check_shreds(shred1: &Shred, shred2: &Shred) -> Result<(), Error> {
```
carllin, line 78: nit: rename to `check_shreds_are_duplicate`

core/src/duplicate_shred.rs
```diff
@@ -0,0 +1,338 @@
...
+    let shred1 = Shred::new_from_serialized_shred(proof.shred1.clone())?;
+    let shred2 = Shred::new_from_serialized_shred(proof.shred2.clone())?;
+    check_shreds(&shred1, &shred2)?;
+    let (slot, shred_index, shred_type) = (
+        shred1.slot(),
+        shred1.index(),
+        shred1.common_header.shred_type,
+    );
+    let data = bincode::serialize(proof)?;
+    let data = encoder(data).map_err(Error::EncodingError)?;
```
carllin, line 116: generally, I think we should avoid `map_err` as it squashes information inside the inner error. Can we wrap the `std::io::Error` and bubble it up here?

core/src/duplicate_shred.rs
```diff
@@ -0,0 +1,338 @@
...
+    let chunk_size = if DUPLICATE_SHRED_HEADER_SIZE < max_size {
+        max_size - DUPLICATE_SHRED_HEADER_SIZE
+    } else {
+        return Err(Error::InvalidSizeLimit);
+    };
+    let chunks: Vec<_> = data.chunks(chunk_size).map(Vec::from).collect();
+    let num_chunks = u8::try_from(chunks.len()).map_err(|_| Error::ValueOutOfBounds)?;
```
carllin, line 123: Would like to avoid `map_err`

core/src/duplicate_shred.rs
```diff
@@ -0,0 +1,338 @@
...
+                    return Err(Error::DataChunkMismatch);
+                }
+            }
+        }
+    }
+    if data.len() != num_chunks as usize {
+        return Err(Error::MissingDataChunk);
+    }
+    let data = (0..num_chunks).map(|k| data.remove(&k).unwrap());
+    let data = decoder(data.concat()).map_err(Error::DecodingError)?;
```
carllin, line 204: Would like to avoid `map_err` if possible, see: https://github.com/solana-labs/solana/pull/14133/files#r544905497

core/src/duplicate_shred.rs
```diff
@@ -0,0 +1,338 @@
...
+use std::{
+    collections::{hash_map::Entry, HashMap},
+    convert::TryFrom,
+};
+use thiserror::Error;
+
+const DUPLICATE_SHRED_HEADER_SIZE: usize = 63;
+
+#[derive(Clone, Debug, PartialEq, AbiExample, Deserialize, Serialize)]
+pub struct DuplicateShred {
```
carllin, line 21: nit: Rename to DuplicateShredChunk

behzadnouri: Leaning on brevity here, since the chunks are an implementation detail hidden inside the struct (i.e. they are not exposed outside this file). We can revisit later if there was ambiguity.

core/src/duplicate_shred.rs
carllin: ah I missed that it was already wrapped, that works!

core/src/duplicate_shred.rs
carllin: ah I missed that it was already wrapped, that works!

carllin (approved):

PR #14134, opened 2020-12-15
add proposal for porting GDB to rBPF

Sladuca:
#### Problem

[rpbf](https://github.com/solana-labs/rbpf) doesn't provide any debugging tools to debug BPF-interpreted programs, so for the most part the user can only use print statements.

#### Summary of Changes
add proposal for implementing GDB backend for solana's eBPF VM

### DIFF
--- a/docs/src/proposals/gdb.md
+++ b/docs/src/proposals/gdb.md
@@ -8,16 +8,18 @@ It would be awesome if we had the ability to attatch a debugger to BPF, ideally 
 
 ## Proposed Solution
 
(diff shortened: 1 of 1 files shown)


jackcmay:
@Sladuca This looks like a good start.  Thoughts on using lldb vs gdb?  Also, the addition of the debugger feature should have minimal or no effect on the performance of the VM when no debugger is attached.

Sladuca:
I didn't consider lldb, I'll spend some time looking into it as an alternative. As far as performance impact, perhaps it's a good idea to add `debug` as a feature flag on the VM so it stays entirely separate?

Sladuca:
It turns out that it's actually much easier to implement a remote target than an architecture definition. I think this is a better alternative because it can be used regardless of the context in which the VM is running. It also allows the user to attach to a VM instance running pretty much wherever they want without doing anything differently.

PR #14145, opened 2020-12-16
Add blockstore skipped api

CriesofCarrots:
#### Problem
The endpoints `getConfirmedBlock` and `getBlockTime` return null for skipped slots, but also for various blockstore errors. This makes `null` responses difficult for clients to interpret.

#### Summary of Changes
- Add `is_skipped` blockstore api; Note: this api cannot distinguish between slots skipped during consensus and slots that are missing from the ledger due to a ledger jump to a more-recent snapshot
- Check is_skipped in rpc and return a specific rpc error if true

### DIFF
--- a/ledger/src/blockstore.rs
+++ b/ledger/src/blockstore.rs
@@ -2649,6 +2649,21 @@ impl Blockstore {
         matches!(self.db.get::<cf::Root>(slot), Ok(Some(true)))
     }
 
+    /// Returns true if a slot is between the rooted slot bounds of the ledger, but has not itself
+    /// been rooted. This is either because the slot was skipped, or due to a gap in ledger data,
+    /// as when booting from a newer snapshot.
+    pub fn is_skipped(&self, slot: Slot) -> bool {
+        let lowest_root = self
+            .rooted_slot_iterator(0)
+            .ok()
+            .and_then(|mut iter| iter.next())
+            .unwrap_or_default();
+        match self.db.get::<cf::Root>(slot).ok().flatten() {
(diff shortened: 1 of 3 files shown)


ledger/src/blockstore.rs
```diff
@@ -2649,6 +2649,21 @@ impl Blockstore {
...
+    /// as when booting from a newer snapshot.
+    pub fn is_skipped(&self, slot: Slot) -> bool {
+        let lowest_root = self
+            .rooted_slot_iterator(0)
+            .ok()
+            .and_then(|mut iter| iter.next())
+            .unwrap_or_default();
+        match self.db.get::<cf::Root>(slot).ok().flatten() {
+            Some(_) => false,
+            None => slot < self.max_root() && slot > lowest_root,
```
carllin, line 2663: If `slot < lowest_root` does that necessarily mean the slot was cleaned up?

Thinking of an edge case like:

1) Lowest root is 5
2) We check confirmation on slot 6, finds null
3) We call `is_skipped`
4) Cleanup cleans up to slot 7
5) `is_skipped` finds slot < lowest_root, but returns false

CriesofCarrots: If you have a suggestion for handling this better, I'm open to it. But I don't think we mind this case too much, since it will be caught lower in the method and return slot cleaned up. The prime use case here is near the max-root.
I do need to think about whether we want to return this error for skipped slots in the range of BigTable, though...

carllin (approved):

PR #14147, opened 2020-12-16
Fix up upgradeable bpf loader activation

mvines:


ledger/src/builtins.rs
```diff
@@ -11,37 +11,33 @@ macro_rules! to_builtin {
...
+fn genesis_builtins(bpf_jit: bool) -> Vec<Builtin> {
+    vec![
         to_builtin!(solana_bpf_loader_deprecated_program!()),
         if bpf_jit {
             to_builtin!(solana_bpf_loader_program_with_jit!())
         } else {
             to_builtin!(solana_bpf_loader_program!())
         },
-    ];
-    if cluster_type != ClusterType::MainnetBeta {
```
mvines, line 23: `cluster_type != ClusterType::MainnetBeta` was what was breaking testnet specifically.   But more generally it's not necessary to introduce cluster-specific programs here at all.  The programs in `feature_builtins()` get enabled automatically when `ClusterType::Development` and for all other cluster types the feature must be first activated

ledger/src/builtins.rs
```diff
@@ -11,37 +11,33 @@ macro_rules! to_builtin {
...
-    builtins
+    ]
 }
 
 /// Builtin programs activated dynamically by feature
-fn feature_builtins() -> Vec<(Builtin, Pubkey, ActivationType)> {
+fn feature_builtins(bpf_jit: bool) -> Vec<(Builtin, Pubkey, ActivationType)> {
     vec![(
-        to_builtin!(solana_bpf_loader_upgradeable_program!()),
+        if bpf_jit {
```
mvines, line 28: "with_jit" was missing for `feature_builtins()` activation of the upgradeable loader too

jackcmay:
@mvines  Thoughts on how we catch this in CI next time?

mvines:
> @mvines Thoughts on how we catch this in CI next time?

Little tricky, we’d first create a snapshot from all four cluster types and then try to load each.  Basically an expensive integration test.

I really want nodes on mainnet/testnet that update nightly with the latest code on the edge/beta channels.  Looks like this’ll be a 2021 thing but that setup would catch bugs like this pretty quickly (in a day or two) as well

ryoqun:
> we’d first create a snapshot from all four cluster types and then try to load each. Basically an expensive integration test.

>  Looks like this’ll be a 2021 thing

yeah, this is what #12175 does.... I couldn't get there due to recent fires after fires. Currently, it's not nightly but at my whim... ;)

PR #14162, opened 2020-12-16
cargo-build-bpf: Create a random `-keypair.json` file alongside the program deploy artifact for easy upgrades

mvines:


sdk/src/signature.rs
```diff
@@ -333,9 +333,8 @@ pub fn read_keypair<R: Read>(reader: &mut R) -> Result<Keypair, Box<dyn error::E
     Ok(Keypair(dalek_keypair))
 }
 
-pub fn read_keypair_file(path: &str) -> Result<Keypair, Box<dyn error::Error>> {
-    assert!(path != "-");
-    let mut file = File::open(path.to_string())?;
+pub fn read_keypair_file<F: AsRef<Path>>(path: F) -> Result<Keypair, Box<dyn error::Error>> {
```
jackcmay, line 336: Break backward compatibility when we port to v1.4?

mvines: This shouldn't break backwards compat.  A `&str` is still accepted by the new signature, and the `assert!(path != "-");` is just dumb -- that business logic should be in the caller for this function

jackcmay (approved):

PR #14184, opened 2020-12-17
JSON RPC: Add transactionCount field to GetEpochInfo

mvines:
The `getTransactionCount` RPC API provides no context regarding the slot that the count was returned for, and adding context would break backwards compatibility.

Instead add a new transactionCount field to GetEpochInfo, which already includes a slot context


CriesofCarrots (approved):
Oh nice, the backward compat is pretty simple. Lgtm!

PR #14205, opened 2020-12-18
getBlockTime RPC method now falls back to BigTable in all cases

mvines:
We still want to fallback to BigTable if available when `blockstore.get_block_time()` returns `Ok(None)`


CriesofCarrots:
Do we need to make this change in `get_confirmed_block()` as well?

mvines:
> Do we need to make this change in `get_confirmed_block()` as well?

I checked, and no.  `get_confirmed_block()` just returns a solid `Err` in this case whereas `get_block_time` plays coy with that `Option`

PR #14208, opened 2020-12-19
Add extra shrink_store path and more aggressive shrinking

sakridge:
#### Problem

Accounts on tmpfs is fast, but takes a lot of space as accounts grow and are not optimally compacted.

#### Summary of Changes

Add a separate shrink file path and more agressively shrink slots with extra accounts space.

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -430,6 +430,8 @@ pub struct AccountsDB {
     /// Set of storage paths to pick from
     pub(crate) paths: Vec<PathBuf>,
 
+    pub shrink_paths: RwLock<Option<Vec<PathBuf>>>,
+
     /// Directory of paths this accounts_db needs to hold/remove
     temp_paths: Option<Vec<TempDir>>,
 
@@ -511,6 +513,7 @@ impl Default for AccountsDB {
             shrink_candidate_slots: Mutex::new(Vec::new()),
             write_version: AtomicU64::new(0),
             paths: vec![],
+            shrink_paths: RwLock::new(None),
             temp_paths: None,
             file_size: DEFAULT_FILE_SIZE,
             thread_pool: rayon::ThreadPoolBuilder::new()
@@ -554,6 +557,15 @@ impl AccountsDB {
             }
         }
         new
+    }
+
+    pub fn set_shrink_paths(&self, paths: Vec<PathBuf>) {
+        assert!(!paths.is_empty());
+        let mut shrink_paths = self.shrink_paths.write().unwrap();
+        for path in &paths {
+            std::fs::create_dir_all(path).expect("Create directory failed.");
+        }
+        *shrink_paths = Some(paths);
     }
 
     pub fn file_size(&self) -> u64 {
@@ -989,11 +1001,28 @@ impl AccountsDB {
     }
 
     fn shrink_stale_slot(&self, candidates: &mut MutexGuard<Vec<Slot>>) -> usize {
-        if let Some(slot) = self.do_next_shrink_slot(candidates) {
-            self.do_shrink_stale_slot(slot)
-        } else {
-            0
+        let mut shrunken_account_total = 0;
+        let mut shrunk_slot_count = 0;
+        let start = Instant::now();
+        let num_roots = self.accounts_index.num_roots();
+        loop {
+            if let Some(slot) = self.do_next_shrink_slot(candidates) {
+                shrunken_account_total += self.do_shrink_stale_slot(slot);
+            } else {
+                return 0;
+            }
+            if start.elapsed().as_millis() > 100 || shrunk_slot_count > num_roots / 10 {
+                debug!(
+                    "do_shrink_stale_slot: {} {} {}us",
+                    shrunk_slot_count,
+                    candidates.len(),
+                    start.elapsed().as_micros()
+                );
+                break;
+            }
+            shrunk_slot_count += 1;
         }
+        shrunken_account_total
     }
 
     // Reads all accounts in given slot's AppendVecs and filter only to alive,
@@ -1028,14 +1057,15 @@ impl AccountsDB {
                 } else if !forced {
                     let sparse_by_count = (alive_count as f32 / stored_count as f32) <= 0.8;
                     let sparse_by_bytes = (written_bytes as f32 / total_bytes as f32) <= 0.8;
-                    let skip_shrink = !sparse_by_count && !sparse_by_bytes;
-                    info!(
-                        "shrink_stale_slot ({}): skip_shrink: {} count: {}/{} byte: {}/{}",
-                        slot, skip_shrink, alive_count, stored_count, written_bytes, total_bytes,
-                    );
-                    if skip_shrink {
+                    let not_sparse = !sparse_by_count && !sparse_by_bytes;
+                    let too_small_to_shrink = total_bytes <= PAGE_SIZE;
+                    if not_sparse || too_small_to_shrink {
                         return 0;
                     }
+                    info!(
+                        "shrink_stale_slot ({}): not_sparse: {} count: {}/{} byte: {}/{}",
+                        slot, not_sparse, alive_count, stored_count, written_bytes, total_bytes,
+                    );
                 }
                 for store in stores.values() {
                     let mut start = 0;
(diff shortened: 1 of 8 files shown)


runtime/src/accounts_db.rs
```diff
@@ -989,11 +1001,28 @@ impl AccountsDB {
...
-        if let Some(slot) = self.do_next_shrink_slot(candidates) {
-            self.do_shrink_stale_slot(slot)
-        } else {
-            0
+        let mut total = 0;
+        let mut count = 0;
+        let start = Instant::now();
+        let num_roots = self.accounts_index.num_roots();
+        loop {
+            if let Some(slot) = self.do_next_shrink_slot(candidates) {
```
carllin, line 1009: I think I noticed at one point, that at startup from a snapshot, the `candidates` set was really large. Might be pertinent to check since we're dependent on going through all the candidates before looping back around

runtime/src/accounts_db.rs
```diff
@@ -989,11 +1001,28 @@ impl AccountsDB {
...
+                total += self.do_shrink_stale_slot(slot);
+            } else {
+                return 0;
+            }
+            if start.elapsed().as_millis() > 100 || count > num_roots / 10 {
```
carllin, line 1014: Before this change, I think @ryoqun can attest to this as well, but one of the dangers we've observed of shrinking too fast is that if the slot is not at the shrink threshold yet (< 80%), then we pop off the slot from the `candidates` set here, but not shrinking actually happens, and then that candidate potentially isn't retried for a while.

From my understanding, the idea here is to try to scan for the first set of shrinkable slots, otherwise we're processing 1/10th of the roots, and then the `do_next_shrink_slot` will re-add the roots again once the iteration finishes. I think this works because the only time a slot is removed from `roots` is through `clean_dead_slot`, in which case `purge_slots` will also have cleaned up all the storages.

carllin (approved):

ryoqun:
maybe, this pr could follow #14219's new convention and multiple argument support?

runtime/src/accounts_db.rs
ryoqun: I think @carllin is right.

Notice that original shrinking impl were intentionally designed to be less aggressive concerned about ramp-tps's bursted txes. (oh, good old days.) :)

We've now encountered the real tx load. So this aggressiveness makes sense now. Also, i think this aggressiveness can evade the shinking-not-retried-for-a-while issue.

runtime/src/accounts_db.rs
```diff
@@ -1029,13 +1058,13 @@ impl AccountsDB {
                     let sparse_by_count = (alive_count as f32 / stored_count as f32) <= 0.8;
                     let sparse_by_bytes = (written_bytes as f32 / total_bytes as f32) <= 0.8;
```
ryoqun, line 1059: I think we're better off skipping to shrink if total_bytes are around PAGE_SIZE or 2*PAGE_SIZE. Otherwise, we'll start to shrink small appendvecs too frequently for almost no win after this pr (yeah, I could have done at #14072):

(This is taken from mainnet-beta while shrinking old roots)

```
[2020-12-21T16:06:08.052216524Z INFO  solana_runtime::accounts_db] shrink_stale_slot (56363861): skip_shrink: false count: 3/6 byte: 1485/4096
[2020-12-21T16:06:08.310957737Z INFO  solana_runtime::accounts_db] shrink_stale_slot (56480921): skip_shrink: false count: 5/6 byte: 1821/4096
[2020-12-21T16:06:08.531950417Z INFO  solana_runtime::accounts_db] shrink_stale_slot (56456146): skip_shrink: false count: 25/28 byte: 8341/12288
[2020-12-21T16:06:08.738150058Z INFO  solana_runtime::accounts_db] shrink_stale_slot (56351547): skip_shrink: false count: 7/8 byte: 2653/4096
[2020-12-21T16:06:09.746224464Z INFO  solana_runtime::accounts_db] shrink_stale_slot (56199914): skip_shrink: false count: 7/8 byte: 2261/4096
[2020-12-21T16:06:09.846730032Z INFO  solana_runtime::accounts_db] shrink_stale_slot (56431829): skip_shrink: true count: 34/37 byte: 10573/12288
[2020-12-21T16:06:10.175186883Z INFO  solana_runtime::accounts_db] shrink_stale_slot (56487717): skip_shrink: true count: 44/48 byte: 14085/16384
[2020-12-21T16:06:10.633622902Z INFO  solana_runtime::accounts_db] shrink_stale_slot (56388477): skip_shrink: false count: 7/8 byte: 2429/4096
[2020-12-21T16:06:10.950872969Z INFO  solana_runtime::accounts_db] shrink_stale_slot (56222118): skip_shrink: false count: 4/5 byte: 1517/4096
[2020-12-21T16:06:11.505014070Z INFO  solana_runtime::accounts_db] shrink_stale_slot (56454911): skip_shrink: true count: 11/13 byte: 3781/4096
[2020-12-21T16:06:11.754149803Z INFO  solana_runtime::accounts_db] shrink_stale_slot (56535189): skip_shrink: false count: 16/17 byte: 5165/8192
[2020-12-21T16:06:11.959589737Z INFO  solana_runtime::accounts_db] shrink_stale_slot (56188165): skip_shrink: false count: 2/3 byte: 909/4096
[2020-12-21T16:06:12.376137359Z INFO  solana_runtime::accounts_db] shrink_stale_slot (56435798): skip_shrink: false count: 15/16 byte: 4693/8192
[2020-12-21T16:06:12.714198710Z INFO  solana_runtime::accounts_db] shrink_stale_slot (56334004): skip_shrink: false count: 4/5 byte: 1013/4096
[2020-12-21T16:06:13.515383761Z INFO  solana_runtime::accounts_db] shrink_stale_slot (56465584): skip_shrink: false count: 9/10 byte: 2869/4096
[2020-12-21T16:06:15.517786062Z INFO  solana_runtime::accounts_db] shrink_stale_slot (56350686): skip_shrink: false count: 6/9 byte: 2397/4096
[2020-12-21T16:06:16.418928396Z INFO  solana_runtime::accounts_db] shrink_stale_slot (56552396): skip_shrink: true count: 59/60 byte: 21069/24576
[2020-12-21T16:06:17.119784717Z INFO  solana_runtime::accounts_db] shrink_stale_slot (56361031): skip_shrink: false count: 6/7 byte: 1957/4096
[2020-12-21T16:06:17.220099106Z INFO  solana_runtime::accounts_db] shrink_stale_slot (56428561): skip_shrink: false count: 5/6 byte: 1821/4096
```

runtime/src/accounts_db.rs
```diff
@@ -989,11 +1001,28 @@ impl AccountsDB {
     }
 
     fn shrink_stale_slot(&self, candidates: &mut MutexGuard<Vec<Slot>>) -> usize {
-        if let Some(slot) = self.do_next_shrink_slot(candidates) {
-            self.do_shrink_stale_slot(slot)
-        } else {
-            0
+        let mut total = 0;
+        let mut count = 0;
```
ryoqun, line 1005: nits: `shrunken_slot_count`?

runtime/src/accounts_db.rs
```diff
@@ -989,11 +1001,28 @@ impl AccountsDB {
     }
 
     fn shrink_stale_slot(&self, candidates: &mut MutexGuard<Vec<Slot>>) -> usize {
-        if let Some(slot) = self.do_next_shrink_slot(candidates) {
-            self.do_shrink_stale_slot(slot)
-        } else {
-            0
+        let mut total = 0;
```
ryoqun, line 1004: nits: `shrunken_account_total`?

ryoqun (approved):
lgtm after https://github.com/solana-labs/solana/pull/14208/files#r546792467 is addressed (I think this is easy thing). (EDIT: oh, and this: https://github.com/solana-labs/solana/pull/14208#issuecomment-749017001)

mvines:
> maybe, this pr could follow #14219's new convention and multiple argument support?

Yes please.   Let's drop the comma support entirely actually for this new arg

runtime/src/accounts_db.rs
sakridge: That's the point of the check for `num_roots / 10` which should limit it to roughly 10% of the candidates set in a single run.

runtime/src/accounts_db.rs
sakridge: Yea, I did experiment with this as well. The effect is not large, but I can add that change.

PR #14210, opened 2020-12-19
Fix timestamp handling on ledger warp

CriesofCarrots:
#### Problem
Testnet was recently rebooted from a snapshot that had warped forward 2 epochs. On a warp, the new Bank doesn't have populated epoch-stakes, so the timestamp calculation falls back to the genesis-estimate. For testnet, this was 2 months in the past (although on a cluster that is proceeding at very close to poh time, this fallback would be in the future). The cluster is slowly correcting the timestamp, but with only 25% change allowed each epoch, this will take a while.

#### Summary of Changes
- Rewrite Bank unix_timestamp and epoch_start_timestamp on Bank::warp_from_parent() so these values are close to realistic. There will still be some variation from real time, depending on how close the warp is to the start of an epoch, but should be < 1 epoch's worth of time.
- Fix up `update_clock` so that unix_timestamp defaults to the ancestor timestamp when no recent vote timestamps are found (as in a big warp)
- Fix timestamp-correction metrics to report on warp slots, and report the correct genesis-estimate
- Warp the current timestamp on testnet forward, feature-gated and only for ClusterType::Testnet

### DIFF
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -1347,11 +1347,12 @@ impl Bank {
                 } else {
                     (EstimateType::Unbounded, None)
                 };
+
+            let ancestor_timestamp = self.clock().unix_timestamp;
             if let Some(timestamp_estimate) =
                 self.get_timestamp_estimate(estimate_type, epoch_start_timestamp)
             {
                 unix_timestamp = timestamp_estimate;
-                let ancestor_timestamp = self.clock().unix_timestamp;
                 if self
                     .feature_set
                     .is_active(&feature_set::timestamp_bounding::id())
@@ -1359,14 +1360,14 @@ impl Bank {
                 {
                     unix_timestamp = ancestor_timestamp;
                 }
-                datapoint_info!(
-                    "bank-timestamp-correction",
-                    ("slot", self.slot(), i64),
-                    ("from_genesis", self.unix_timestamp_from_genesis(), i64),
-                    ("corrected", timestamp_estimate, i64),
-                    ("ancestor_timestamp", ancestor_timestamp, i64),
-                );
             }
+            datapoint_info!(
+                "bank-timestamp-correction",
+                ("slot", self.slot(), i64),
+                ("from_genesis", self.unix_timestamp_from_genesis(), i64),
+                ("corrected", unix_timestamp, i64),
+                ("ancestor_timestamp", ancestor_timestamp, i64),
+            );
         }
         let epoch_start_timestamp = if self
             .feature_set
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -1369,7 +1369,7 @@ impl Bank {
                 ("ancestor_timestamp", ancestor_timestamp, i64),
             );
         }
-        let epoch_start_timestamp = if self
+        let mut epoch_start_timestamp = if self
             .feature_set
             .is_active(&feature_set::timestamp_bounding::id())
         {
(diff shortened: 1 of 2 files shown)


sdk/src/feature_set.rs
```diff
@@ -139,6 +143,7 @@ lazy_static! {
         (simple_capitalization::id(), "simple capitalization"),
         (bpf_loader_upgradeable_program::id(), "upgradeable bpf loader"),
         (try_find_program_address_syscall_enabled::id(), "add try_find_program_address syscall"),
+        (warp_testnet_timestamp::id(), "warp testnet timestamp to current #TODO"),
```
mvines, line 146:
```suggestion
        (warp_testnet_timestamp::id(), "warp testnet timestamp to current #14210"),
```

mvines (dismissed):

runtime/src/bank.rs
```diff
@@ -1075,10 +1075,22 @@ impl Bank {
...
+        let parent_timestamp = parent.clock().unix_timestamp;
         let mut new = Bank::new_from_parent(parent, collector_id, slot);
         new.apply_feature_activations(true);
         new.update_epoch_stakes(new.epoch_schedule().get_epoch(slot));
         new.tick_height.store(new.max_tick_height(), Relaxed);
+
+        let mut clock = new.clock();
+        clock.epoch_start_timestamp = parent_timestamp;
+        clock.unix_timestamp = parent_timestamp;
+        new.update_sysvar_account(&sysvar::clock::id(), |account| {
```
ryoqun, line 1087: I think you can just use this given `account` to create `Clock`? Or, you want to mangle the new state of Clock for `new` (bank)? Yeah, update_sysvar_account specifically gives the parent sysvar even if it's already updated at the `new` (bank).

runtime/src/bank.rs
```diff
@@ -1319,45 +1332,54 @@ impl Bank {
                     // needed for timestamp bounding, but isn't yet corrected for the activation slot
                     let epoch_start_timestamp = if self.slot() > timestamp_bounding_activation_slot
                     {
-                        let epoch = if let Some(epoch) = parent_epoch {
-                            epoch
+                        let warp_testnet_timestamp = self
+                            .feature_set
+                            .activated_slot(&feature_set::warp_testnet_timestamp::id());
+                        if warp_testnet_timestamp.is_some()
```
ryoqun, line 1338: nits: I think this can just be `if warp_testnet_timestamp == Some(self.slot())` with simpler condition and fewer `.unwrap()`.

runtime/src/bank.rs
```diff
@@ -1319,45 +1332,54 @@ impl Bank {
...
+                            && warp_testnet_timestamp.unwrap() == self.slot()
```
ryoqun, line 1339: nits: I think this can just be `if warp_testnet_timestamp == Some(self.slot())` with simpler condition and fewer `.unwrap()`.

runtime/src/bank.rs
ryoqun: oh, `update_clock` is already doing its update logic outside `update_sysvar_account`'s closure. So, I guess this is just following the style.

ryoqun:
@CriesOfCarrots I think this pr is safe from bank hash mismatches even if this is changing already feature-activated code.

CriesofCarrots:
> @CriesofCarrots I think this pr is safe from bank hash mismatches even if this is changing already feature-activated code.

@ryoqun super, thank you so much for confirming!

PR #14212, opened 2020-12-19
Mint index

carllin:
#### Problem
Scanning is slow on SPL tokens accounts with a filter for the mint is currently slow without an index
#### Summary of Changes
Introduce a mint index

Fixes #

### DIFF
--- a/core/src/rpc.rs
+++ b/core/src/rpc.rs
@@ -43,6 +43,7 @@ use solana_runtime::{
     bank::Bank,
     bank_forks::BankForks,
     commitment::{BlockCommitmentArray, BlockCommitmentCache, CommitmentSlots},
+    inline_spl_token_v2_0::{SPL_TOKEN_ACCOUNT_MINT_OFFSET, SPL_TOKEN_ACCOUNT_OWNER_OFFSET},
 };
 use solana_sdk::{
     account::Account,
@@ -52,7 +53,7 @@ use solana_sdk::{
     epoch_info::EpochInfo,
     epoch_schedule::EpochSchedule,
     hash::Hash,
-    pubkey::{Pubkey, PUBKEY_BYTES},
+    pubkey::Pubkey,
     signature::Signature,
     stake_history::StakeHistory,
     system_instruction,
@@ -1335,7 +1336,7 @@ impl JsonRpcRequestProcessor {
         ));
         // Filter on Owner address
         filters.push(RpcFilterType::Memcmp(Memcmp {
-            offset: 32,
+            offset: SPL_TOKEN_ACCOUNT_OWNER_OFFSET,
             bytes: MemcmpEncodedBytes::Binary(owner_key.to_string()),
             encoding: None,
         }));
@@ -1374,7 +1375,7 @@ impl JsonRpcRequestProcessor {
         ));
         // Filter on Mint address
         filters.push(RpcFilterType::Memcmp(Memcmp {
-            offset: 0,
+            offset: SPL_TOKEN_ACCOUNT_MINT_OFFSET,
             bytes: MemcmpEncodedBytes::Binary(mint_key.to_string()),
             encoding: None,
         }));
@@ -1501,7 +1502,7 @@ fn get_spl_token_owner_filter(program_id: &Pubkey, filters: &[RpcFilterType]) ->
         match filter {
             RpcFilterType::DataSize(size) => data_size_filter = Some(*size),
             RpcFilterType::Memcmp(Memcmp {
-                offset: PUBKEY_BYTES,
+                offset: SPL_TOKEN_ACCOUNT_OWNER_OFFSET,
                 bytes: MemcmpEncodedBytes::Binary(bytes),
                 ..
             }) => {
--- a/core/src/rpc.rs
+++ b/core/src/rpc.rs
@@ -1312,7 +1312,9 @@ impl JsonRpcRequestProcessor {
             .account_indexes
             .contains(&AccountIndex::ProgramId)
         {
-            bank.get_filtered_indexed_accounts(&IndexKey::ProgramId(*program_id), filter_closure)
+            bank.get_filtered_indexed_accounts(&IndexKey::ProgramId(*program_id), |account| {
+                account.owner == *program_id && filter_closure(account)
+            })
         } else {
             bank.get_filtered_program_accounts(program_id, filter_closure)
         }
@@ -1347,10 +1349,11 @@ impl JsonRpcRequestProcessor {
             .contains(&AccountIndex::SplTokenOwner)
         {
             bank.get_filtered_indexed_accounts(&IndexKey::SplTokenOwner(*owner_key), |account| {
-                filters.iter().all(|filter_type| match filter_type {
-                    RpcFilterType::DataSize(size) => account.data.len() as u64 == *size,
-                    RpcFilterType::Memcmp(compare) => compare.bytes_match(&account.data),
-                })
+                account.owner == spl_token_id_v2_0()
+                    && filters.iter().all(|filter_type| match filter_type {
+                        RpcFilterType::DataSize(size) => account.data.len() as u64 == *size,
+                        RpcFilterType::Memcmp(compare) => compare.bytes_match(&account.data),
+                    })
             })
         } else {
             self.get_filtered_program_accounts(bank, &spl_token_id_v2_0(), filters)
@@ -1385,10 +1388,11 @@ impl JsonRpcRequestProcessor {
             .contains(&AccountIndex::SplTokenMint)
         {
             bank.get_filtered_indexed_accounts(&IndexKey::SplTokenMint(*mint_key), |account| {
-                filters.iter().all(|filter_type| match filter_type {
-                    RpcFilterType::DataSize(size) => account.data.len() as u64 == *size,
-                    RpcFilterType::Memcmp(compare) => compare.bytes_match(&account.data),
-                })
+                account.owner == spl_token_id_v2_0()
+                    && filters.iter().all(|filter_type| match filter_type {
+                        RpcFilterType::DataSize(size) => account.data.len() as u64 == *size,
+                        RpcFilterType::Memcmp(compare) => compare.bytes_match(&account.data),
+                    })
             })
         } else {
             self.get_filtered_program_accounts(bank, &spl_token_id_v2_0(), filters)
--- a/core/src/rpc.rs
+++ b/core/src/rpc.rs
@@ -318,9 +318,10 @@ impl JsonRpcRequestProcessor {
         let keyed_accounts = get_filtered_program_accounts(&bank, program_id, filters);
         let result =
             if program_id == &spl_token_id_v2_0() && encoding == UiAccountEncoding::JsonParsed {
-                get_parsed_token_accounts(bank, keyed_accounts).collect()
+                get_parsed_token_accounts(bank, keyed_accounts.into_iter()).collect()
             } else {
                 keyed_accounts
+                    .into_iter()
                     .map(|(pubkey, account)| RpcKeyedAccount {
                         pubkey: pubkey.to_string(),
                         account: UiAccount::encode(
@@ -1155,18 +1156,9 @@ impl JsonRpcRequestProcessor {
                 "Invalid param: not a v2.0 Token mint".to_string(),
             ));
         }
-        let filters = vec![
-            // Filter on Mint address
-            RpcFilterType::Memcmp(Memcmp {
-                offset: 0,
-                bytes: MemcmpEncodedBytes::Binary(mint.to_string()),
-                encoding: None,
-            }),
-            // Filter on Token Account state
-            RpcFilterType::DataSize(TokenAccount::get_packed_len() as u64),
-        ];
         let mut token_balances: Vec<RpcTokenAccountBalance> =
-            get_filtered_program_accounts(&bank, &mint_owner, filters)
+            get_filtered_spl_token_accounts_by_mint(&bank, &mint, vec![])
+                .into_iter()
                 .map(|(address, account)| {
                     let amount = TokenAccount::unpack(&account.data)
                         .map(|account| account.amount)
@@ -1210,22 +1202,22 @@ impl JsonRpcRequestProcessor {
                 bytes: MemcmpEncodedBytes::Binary(owner.to_string()),
                 encoding: None,
             }),
-            // Filter on Token Account state
-            RpcFilterType::DataSize(TokenAccount::get_packed_len() as u64),
         ];
-        if let Some(mint) = mint {
-            // Optional filter on Mint address
-            filters.push(RpcFilterType::Memcmp(Memcmp {
-                offset: 0,
-                bytes: MemcmpEncodedBytes::Binary(mint.to_string()),
-                encoding: None,
-            }));
-        }
-        let keyed_accounts = get_filtered_program_accounts(&bank, &token_program_id, filters);
+        // Optional filter on Mint address, uses mint account index for scan
+        let keyed_accounts = if let Some(mint) = mint {
+            get_filtered_spl_token_accounts_by_mint(&bank, &mint, filters)
+        } else {
+            // Filter on Token Account state
+            filters.push(RpcFilterType::DataSize(
+                TokenAccount::get_packed_len() as u64
+            ));
+            get_filtered_program_accounts(&bank, &token_program_id, filters)
+        };
         let accounts = if encoding == UiAccountEncoding::JsonParsed {
-            get_parsed_token_accounts(bank.clone(), keyed_accounts).collect()
+            get_parsed_token_accounts(bank.clone(), keyed_accounts.into_iter()).collect()
         } else {
             keyed_accounts
+                .into_iter()
                 .map(|(pubkey, account)| RpcKeyedAccount {
                     pubkey: pubkey.to_string(),
                     account: UiAccount::encode(
@@ -1269,22 +1261,22 @@ impl JsonRpcRequestProcessor {
                 bytes: MemcmpEncodedBytes::Binary(delegate.to_string()),
                 encoding: None,
             }),
-            // Filter on Token Account state
-            RpcFilterType::DataSize(TokenAccount::get_packed_len() as u64),
         ];
-        if let Some(mint) = mint {
-            // Optional filter on Mint address
-            filters.push(RpcFilterType::Memcmp(Memcmp {
-                offset: 0,
-                bytes: MemcmpEncodedBytes::Binary(mint.to_string()),
-                encoding: None,
-            }));
-        }
-        let keyed_accounts = get_filtered_program_accounts(&bank, &token_program_id, filters);
+        // Optional filter on Mint address, uses mint account index for scan
+        let keyed_accounts = if let Some(mint) = mint {
+            get_filtered_spl_token_accounts_by_mint(&bank, &mint, filters)
+        } else {
+            // Filter on Token Account state
+            filters.push(RpcFilterType::DataSize(
+                TokenAccount::get_packed_len() as u64
+            ));
+            get_filtered_program_accounts(&bank, &token_program_id, filters)
+        };
         let accounts = if encoding == UiAccountEncoding::JsonParsed {
-            get_parsed_token_accounts(bank.clone(), keyed_accounts).collect()
+            get_parsed_token_accounts(bank.clone(), keyed_accounts.into_iter()).collect()
         } else {
             keyed_accounts
+                .into_iter()
                 .map(|(pubkey, account)| RpcKeyedAccount {
                     pubkey: pubkey.to_string(),
                     account: UiAccount::encode(
@@ -1402,14 +1394,27 @@ fn get_filtered_program_accounts(
     bank: &Arc<Bank>,
     program_id: &Pubkey,
     filters: Vec<RpcFilterType>,
-) -> impl Iterator<Item = (Pubkey, Account)> {
+) -> Vec<(Pubkey, Account)> {
     bank.get_filtered_program_accounts(&program_id, |account| {
         filters.iter().all(|filter_type| match filter_type {
             RpcFilterType::DataSize(size) => account.data.len() as u64 == *size,
             RpcFilterType::Memcmp(compare) => compare.bytes_match(&account.data),
         })
     })
-    .into_iter()
+}
+
+/// Get an iterator of spl-token accounts by mint address
+fn get_filtered_spl_token_accounts_by_mint(
+    bank: &Arc<Bank>,
+    mint_key: &Pubkey,
+    filters: Vec<RpcFilterType>,
+) -> Vec<(Pubkey, Account)> {
+    bank.get_filtered_spl_token_accounts_by_mint(&mint_key, |account| {
+        filters.iter().all(|filter_type| match filter_type {
+            RpcFilterType::DataSize(size) => account.data.len() as u64 == *size,
+            RpcFilterType::Memcmp(compare) => compare.bytes_match(&account.data),
+        })
+    })
 }
 
 pub(crate) fn get_parsed_token_account(
--- a/core/src/rpc.rs
+++ b/core/src/rpc.rs
@@ -111,7 +111,7 @@ pub struct JsonRpcConfig {
     pub enable_bigtable_ledger_storage: bool,
     pub enable_bigtable_ledger_upload: bool,
     pub max_multiple_accounts: Option<usize>,
-    pub account_indexes: Vec<AccountIndex>,
+    pub account_indexes: HashSet<AccountIndex>,
 }
 
 #[derive(Clone)]
--- a/core/src/rpc.rs
+++ b/core/src/rpc.rs
@@ -39,7 +39,7 @@ use solana_metrics::inc_new_counter_info;
 use solana_perf::packet::PACKET_DATA_SIZE;
 use solana_runtime::{
     accounts::AccountAddressFilter,
-    accounts_index::{IndexKey, IndexType},
+    accounts_index::{AccountIndex, IndexKey},
     bank::Bank,
     bank_forks::BankForks,
     commitment::{BlockCommitmentArray, BlockCommitmentCache, CommitmentSlots},
@@ -110,7 +110,7 @@ pub struct JsonRpcConfig {
     pub enable_bigtable_ledger_storage: bool,
     pub enable_bigtable_ledger_upload: bool,
     pub max_multiple_accounts: Option<usize>,
-    pub supported_indexes: Vec<IndexType>,
+    pub account_indexes: Vec<AccountIndex>,
 }
 
 #[derive(Clone)]
@@ -1308,8 +1308,8 @@ impl JsonRpcRequestProcessor {
         };
         if self
             .config
-            .supported_indexes
-            .contains(&IndexType::ProgramId)
+            .account_indexes
+            .contains(&AccountIndex::ProgramId)
         {
             bank.get_filtered_indexed_accounts(&IndexKey::ProgramId(*program_id), filter_closure)
         } else {
@@ -1342,10 +1342,10 @@ impl JsonRpcRequestProcessor {
 
         if self
             .config
-            .supported_indexes
-            .contains(&IndexType::TokenOwner)
+            .account_indexes
+            .contains(&AccountIndex::SplTokenOwner)
         {
-            bank.get_filtered_indexed_accounts(&IndexKey::TokenOwner(*owner_key), |account| {
+            bank.get_filtered_indexed_accounts(&IndexKey::SplTokenOwner(*owner_key), |account| {
                 filters.iter().all(|filter_type| match filter_type {
                     RpcFilterType::DataSize(size) => account.data.len() as u64 == *size,
                     RpcFilterType::Memcmp(compare) => compare.bytes_match(&account.data),
@@ -1378,8 +1378,12 @@ impl JsonRpcRequestProcessor {
             bytes: MemcmpEncodedBytes::Binary(mint_key.to_string()),
             encoding: None,
         }));
-        if self.config.supported_indexes.contains(&IndexType::Mint) {
-            bank.get_filtered_indexed_accounts(&IndexKey::Mint(*mint_key), |account| {
+        if self
+            .config
+            .account_indexes
+            .contains(&AccountIndex::SplTokenMint)
+        {
+            bank.get_filtered_indexed_accounts(&IndexKey::SplTokenMint(*mint_key), |account| {
                 filters.iter().all(|filter_type| match filter_type {
                     RpcFilterType::DataSize(size) => account.data.len() as u64 == *size,
                     RpcFilterType::Memcmp(compare) => compare.bytes_match(&account.data),
--- a/core/src/rpc.rs
+++ b/core/src/rpc.rs
@@ -1202,16 +1202,7 @@ impl JsonRpcRequestProcessor {
         check_slice_and_encoding(&encoding, data_slice_config.is_some())?;
         let (_, mint) = get_token_program_id_and_mint(&bank, token_account_filter)?;
 
-        let mut filters = vec![
-            // Filter on Owner address
-            RpcFilterType::Memcmp(Memcmp {
-                offset: 32,
-                bytes: MemcmpEncodedBytes::Binary(owner.to_string()),
-                encoding: None,
-            }),
-            // Filter on Token Account state
-            RpcFilterType::DataSize(TokenAccount::get_packed_len() as u64),
-        ];
+        let mut filters = vec![];
         if let Some(mint) = mint {
             // Optional filter on Mint address
             filters.push(RpcFilterType::Memcmp(Memcmp {
@@ -1416,8 +1407,24 @@ fn get_filtered_program_accounts(
 fn get_filtered_spl_token_accounts_by_owner(
     bank: &Arc<Bank>,
     owner_key: &Pubkey,
-    filters: Vec<RpcFilterType>,
+    mut filters: Vec<RpcFilterType>,
 ) -> Vec<(Pubkey, Account)> {
+    // The by-owner accounts index checks for Token Account state and Owner address on inclusion.
+    // However, due to the current AppendVec implementation, accounts may remain in the index after
+    // being wiped and reinitialized as something else. We include the redundant filters here to
+    // avoid returning these accounts.
+    //
+    // Filter on Token Account state
+    filters.push(RpcFilterType::DataSize(
+        TokenAccount::get_packed_len() as u64
+    ));
+    // Filter on Owner address
+    filters.push(RpcFilterType::Memcmp(Memcmp {
+        offset: 32,
+        bytes: MemcmpEncodedBytes::Binary(owner_key.to_string()),
+        encoding: None,
+    }));
+
     bank.get_filtered_indexed_accounts(&IndexKey::TokenOwner(*owner_key), |account| {
         filters.iter().all(|filter_type| match filter_type {
             RpcFilterType::DataSize(size) => account.data.len() as u64 == *size,
@@ -1430,8 +1437,24 @@ fn get_filtered_spl_token_accounts_by_owner(
 fn get_filtered_spl_token_accounts_by_mint(
     bank: &Arc<Bank>,
     mint_key: &Pubkey,
-    filters: Vec<RpcFilterType>,
+    mut filters: Vec<RpcFilterType>,
 ) -> Vec<(Pubkey, Account)> {
+    // The by-mint accounts index checks for Token Account state and Mint address on inclusion.
+    // However, due to the current AppendVec implementation, accounts may remain in the index after
+    // being wiped and reinitialized as something else. We include the redundant filters here to
+    // avoid returning these accounts.
+    //
+    // Filter on Token Account state
+    filters.push(RpcFilterType::DataSize(
+        TokenAccount::get_packed_len() as u64
+    ));
+    // Filter on Mint address
+    filters.push(RpcFilterType::Memcmp(Memcmp {
+        offset: 0,
+        bytes: MemcmpEncodedBytes::Binary(mint_key.to_string()),
+        encoding: None,
+    }));
+
     bank.get_filtered_indexed_accounts(&IndexKey::Mint(*mint_key), |account| {
         filters.iter().all(|filter_type| match filter_type {
             RpcFilterType::DataSize(size) => account.data.len() as u64 == *size,
--- a/core/src/rpc.rs
+++ b/core/src/rpc.rs
@@ -39,6 +39,7 @@ use solana_metrics::inc_new_counter_info;
 use solana_perf::packet::PACKET_DATA_SIZE;
 use solana_runtime::{
     accounts::AccountAddressFilter,
+    accounts_index::IndexKey,
     bank::Bank,
     bank_forks::BankForks,
     commitment::{BlockCommitmentArray, BlockCommitmentCache, CommitmentSlots},
@@ -1202,17 +1203,19 @@ impl JsonRpcRequestProcessor {
                 bytes: MemcmpEncodedBytes::Binary(owner.to_string()),
                 encoding: None,
             }),
-        ];
-        // Optional filter on Mint address, uses mint account index for scan
-        let keyed_accounts = if let Some(mint) = mint {
-            get_filtered_spl_token_accounts_by_mint(&bank, &mint, filters)
-        } else {
             // Filter on Token Account state
-            filters.push(RpcFilterType::DataSize(
-                TokenAccount::get_packed_len() as u64
-            ));
-            get_filtered_program_accounts(&bank, &token_program_id, filters)
-        };
+            RpcFilterType::DataSize(TokenAccount::get_packed_len() as u64),
+        ];
+        if let Some(mint) = mint {
+            // Optional filter on Mint address
+            filters.push(RpcFilterType::Memcmp(Memcmp {
+                offset: 0,
+                bytes: MemcmpEncodedBytes::Binary(mint.to_string()),
+                encoding: None,
+            }));
+        }
+
+        let keyed_accounts = get_filtered_spl_token_accounts_by_owner(&bank, owner, filters);
         let accounts = if encoding == UiAccountEncoding::JsonParsed {
             get_parsed_token_accounts(bank.clone(), keyed_accounts.into_iter()).collect()
         } else {
@@ -1403,13 +1406,27 @@ fn get_filtered_program_accounts(
     })
 }
 
+/// Get an iterator of spl-token accounts by owner address
+fn get_filtered_spl_token_accounts_by_owner(
+    bank: &Arc<Bank>,
+    owner_key: &Pubkey,
+    filters: Vec<RpcFilterType>,
+) -> Vec<(Pubkey, Account)> {
+    bank.get_filtered_indexed_accounts(&IndexKey::TokenOwner(*owner_key), |account| {
+        filters.iter().all(|filter_type| match filter_type {
+            RpcFilterType::DataSize(size) => account.data.len() as u64 == *size,
+            RpcFilterType::Memcmp(compare) => compare.bytes_match(&account.data),
+        })
+    })
+}
+
 /// Get an iterator of spl-token accounts by mint address
 fn get_filtered_spl_token_accounts_by_mint(
     bank: &Arc<Bank>,
     mint_key: &Pubkey,
     filters: Vec<RpcFilterType>,
 ) -> Vec<(Pubkey, Account)> {
-    bank.get_filtered_spl_token_accounts_by_mint(&mint_key, |account| {
+    bank.get_filtered_indexed_accounts(&IndexKey::Mint(*mint_key), |account| {
         filters.iter().all(|filter_type| match filter_type {
             RpcFilterType::DataSize(size) => account.data.len() as u64 == *size,
             RpcFilterType::Memcmp(compare) => compare.bytes_match(&account.data),
--- a/core/src/rpc.rs
+++ b/core/src/rpc.rs
@@ -52,7 +52,7 @@ use solana_sdk::{
     epoch_info::EpochInfo,
     epoch_schedule::EpochSchedule,
     hash::Hash,
-    pubkey::Pubkey,
+    pubkey::{Pubkey, PUBKEY_BYTES},
     signature::Signature,
     stake_history::StakeHistory,
     system_instruction,
@@ -316,7 +316,13 @@ impl JsonRpcRequestProcessor {
         let encoding = config.encoding.unwrap_or(UiAccountEncoding::Binary);
         let data_slice_config = config.data_slice;
         check_slice_and_encoding(&encoding, data_slice_config.is_some())?;
-        let keyed_accounts = get_filtered_program_accounts(&bank, program_id, filters);
+        let keyed_accounts = {
+            if let Some(owner) = get_spl_token_owner_filter(program_id, &filters) {
+                get_filtered_spl_token_accounts_by_owner(&bank, &owner, filters)
+            } else {
+                get_filtered_program_accounts(&bank, program_id, filters)
+            }
+        };
         let result =
             if program_id == &spl_token_id_v2_0() && encoding == UiAccountEncoding::JsonParsed {
                 get_parsed_token_accounts(bank, keyed_accounts.into_iter()).collect()
@@ -1194,7 +1200,7 @@ impl JsonRpcRequestProcessor {
         let encoding = config.encoding.unwrap_or(UiAccountEncoding::Binary);
         let data_slice_config = config.data_slice;
         check_slice_and_encoding(&encoding, data_slice_config.is_some())?;
-        let (token_program_id, mint) = get_token_program_id_and_mint(&bank, token_account_filter)?;
+        let (_, mint) = get_token_program_id_and_mint(&bank, token_account_filter)?;
 
         let mut filters = vec![
             // Filter on Owner address
@@ -1432,6 +1438,34 @@ fn get_filtered_spl_token_accounts_by_mint(
             RpcFilterType::Memcmp(compare) => compare.bytes_match(&account.data),
         })
     })
+}
+
+fn get_spl_token_owner_filter(program_id: &Pubkey, filters: &[RpcFilterType]) -> Option<Pubkey> {
+    if program_id != &spl_token_id_v2_0() {
+        return None;
+    }
+    let mut data_size_filter: Option<u64> = None;
+    let mut owner_key: Option<Pubkey> = None;
+    for filter in filters {
+        match filter {
+            RpcFilterType::DataSize(size) => data_size_filter = Some(*size),
+            RpcFilterType::Memcmp(Memcmp {
+                offset: PUBKEY_BYTES,
+                bytes: MemcmpEncodedBytes::Binary(bytes),
+                ..
+            }) => {
+                if let Ok(key) = Pubkey::from_str(bytes) {
+                    owner_key = Some(key)
+                }
+            }
+            _ => {}
+        }
+    }
+    if data_size_filter == Some(TokenAccount::get_packed_len() as u64) {
+        owner_key
+    } else {
+        None
+    }
 }
 
 pub(crate) fn get_parsed_token_account(
@@ -5802,6 +5836,54 @@ pub mod tests {
                 }
             })
         );
+    }
+
+    #[test]
+    fn test_get_spl_token_owner_filter() {
+        let owner = Pubkey::new_unique();
+        assert_eq!(
+            get_spl_token_owner_filter(
+                &Pubkey::from_str("TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA").unwrap(),
+                &[
+                    RpcFilterType::Memcmp(Memcmp {
+                        offset: 32,
+                        bytes: MemcmpEncodedBytes::Binary(owner.to_string()),
+                        encoding: None
+                    }),
+                    RpcFilterType::DataSize(165)
+                ],
+            )
+            .unwrap(),
+            owner
+        );
+
+        // Filtering on mint instead of owner
+        assert!(get_spl_token_owner_filter(
+            &Pubkey::from_str("TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA").unwrap(),
+            &[
+                RpcFilterType::Memcmp(Memcmp {
+                    offset: 0,
+                    bytes: MemcmpEncodedBytes::Binary(owner.to_string()),
+                    encoding: None
+                }),
+                RpcFilterType::DataSize(165)
+            ],
+        )
+        .is_none());
+
+        // Wrong program id
+        assert!(get_spl_token_owner_filter(
+            &Pubkey::new_unique(),
+            &[
+                RpcFilterType::Memcmp(Memcmp {
+                    offset: 32,
+                    bytes: MemcmpEncodedBytes::Binary(owner.to_string()),
+                    encoding: None
+                }),
+                RpcFilterType::DataSize(165)
+            ],
+        )
+        .is_none());
     }
 
     #[test]
--- a/core/src/rpc.rs
+++ b/core/src/rpc.rs
@@ -1403,14 +1403,13 @@ fn get_filtered_program_accounts(
     program_id: &Pubkey,
     filters: Vec<RpcFilterType>,
 ) -> impl Iterator<Item = (Pubkey, Account)> {
-    bank.get_program_accounts(&program_id)
-        .into_iter()
-        .filter(move |(_, account)| {
-            filters.iter().all(|filter_type| match filter_type {
-                RpcFilterType::DataSize(size) => account.data.len() as u64 == *size,
-                RpcFilterType::Memcmp(compare) => compare.bytes_match(&account.data),
-            })
+    bank.get_filtered_program_accounts(&program_id, |account| {
+        filters.iter().all(|filter_type| match filter_type {
+            RpcFilterType::DataSize(size) => account.data.len() as u64 == *size,
+            RpcFilterType::Memcmp(compare) => compare.bytes_match(&account.data),
         })
+    })
+    .into_iter()
 }
 
 pub(crate) fn get_parsed_token_account(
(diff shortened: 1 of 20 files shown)


runtime/src/accounts_index.rs
```diff
@@ -366,6 +383,8 @@ impl<T: 'static + Clone> AccountsIndex<T> {
         F: FnMut(&Pubkey, (&T, Slot)),
         R: RangeBounds<Pubkey>,
     {
+        // TODO: expand to use mint index to find the `pubkey_list` below more efficiently
+        // instead of scanning the entire range
```
carllin, line 387: @CriesofCarrots  have to expand this method to leverage the new index for scans that query by the mint

sdk/src/pubkey.rs
```diff
@@ -1,9 +1,10 @@
+pub const PUBKEY_BYTES: usize = 32;
 pub use solana_program::pubkey::*;
```
mvines, line 2: nit: `PUBKEY_BYTES` can live in `solana_program::pubkey`

runtime/src/accounts_index.rs
```diff
@@ -581,6 +622,36 @@ impl<T: 'static + Clone> AccountsIndex<T> {
...
+                        self.mint_index
+                            .index
+                            .entry(mint_pubkey)
+                            .or_insert(DashSet::new())
+                            .downgrade()
+                    });
+                mint_account_keys.insert(*pubkey);
+            }
+
+            // TODO: handle if pubkey switches mint within the same slot
```
CriesofCarrots, line 646: I don't think this is possible? In order to switch a mints, a token account would have to close (lamports zeroed) and be removed before being reinitialized with the new mint.

runtime/src/accounts_index.rs
```diff
@@ -581,6 +622,36 @@ impl<T: 'static + Clone> AccountsIndex<T> {
...
     }
 
+    pub fn update_secondary_indexes(
+        &self,
+        slot: Slot,
+        pubkey: &Pubkey,
+        account_owner: &Pubkey,
+        account_data: &[u8],
+    ) {
+        if *account_owner == inline_spl_token_v2_0::id() && account_data.len() >= PUBKEY_BYTES {
```
CriesofCarrots, line 632: I think we also need to check that `account_data.len() == spl_token_v2_0::state::Account::LEN`, otherwise this will incorrectly index Mint and Multisig accounts as well

runtime/src/accounts_index.rs
```diff
@@ -181,9 +187,20 @@ impl<'a, T: 'static + Clone> Iterator for AccountsIndexIterator<'a, T> {
...
 }
 
+#[derive(Debug, Default)]
+pub struct MintIndex {
+    index: DashMap<Pubkey, MintIndexEntry>,
+    // Map from pubkey to the mint it's stored in, used for cleanup.
+    // Alternative is to store Option<Pubkey> in each AccountInfo in the
+    // AccountsIndex if something is an SPL account with a mint, but then
+    // every AccountInfo would have to allocate `Option<Pubkey>`
+    accounts_to_mints: DashMap<Pubkey, DashSet<Pubkey>>,
```
CriesofCarrots, line 197: A token account can only be associated with one mint at a time. Do we need the DashSet here?

CriesofCarrots: As discussed over DM, we do need a DashSet, as this spans all updates, not a particular point in time.

runtime/src/accounts_index.rs
CriesofCarrots: Ah, this is possible with serial transactions in the same slot.

runtime/src/accounts_db.rs
```diff
@@ -2896,33 +2925,34 @@ impl AccountsDB {
...
                 );
 
             let mut accounts_map: HashMap<Pubkey, Vec<(u64, AccountInfo)>> = HashMap::new();
-            for accumulator_entry in accumulator.iter() {
+            for accumulator_entry in accumulator.into_iter() {
                 for (pubkey, storage_entry) in accumulator_entry {
-                    let entry = accounts_map.entry(*pubkey).or_insert_with(Vec::new);
-                    entry.extend(storage_entry.iter().cloned());
+                    let entry = accounts_map.entry(pubkey).or_insert_with(Vec::new);
+                    entry.extend(storage_entry.into_iter());
```
ryoqun, line 2941: less, clones. :)

runtime/src/accounts_index.rs
```diff
@@ -581,6 +685,53 @@ impl<T: 'static + Clone> AccountsIndex<T> {
...
+                            .entry(*pubkey)
+                            .or_insert(DashSet::new())
+                            .downgrade()
+                    });
+
+                if account_to_mint.insert(mint_key) && account_to_mint.len() > 1 {
+                    datapoint_info!("key_multiple_mints", ("pubkey", pubkey.to_string(), String),);
+                }
+            }
+        }
```
ryoqun, line 732: maybe, purge entries if `account.lamports == 0` to avoid cache poisoning?

runtime/src/accounts_db.rs
```diff
@@ -2458,6 +2482,11 @@ impl AccountsDB {
             let pubkey = pubkey_account.0;
             self.accounts_index
                 .upsert(slot, pubkey, info, &mut reclaims);
+            self.accounts_index.update_secondary_indexes(
```
ryoqun, line 2485: maybe, we need to consider forks (unrooted ones could be abondoned) much like AccountsIndex?

ryoqun: hmm, I think this is already handled?, so my above comment is wrong:  https://github.com/solana-labs/solana/pull/14212/files?diff=split&w=1#diff-db6927be895a2b4d140c54b84392a861b4535e91afc466c5e93c243c6bf2b28aR443

carllin: hehe yeah, I had a much more complicated version that took those into account, but moved to this simpler fix for the immediate band-aid

runtime/src/accounts_index.rs
```diff
@@ -581,6 +685,53 @@ impl<T: 'static + Clone> AccountsIndex<T> {
...
+                    .unwrap_or_else(|| {
+                        self.mint_index
+                            .accounts_to_mints
+                            .entry(*pubkey)
+                            .or_insert(DashSet::new())
+                            .downgrade()
+                    });
+
+                if account_to_mint.insert(mint_key) && account_to_mint.len() > 1 {
+                    datapoint_info!("key_multiple_mints", ("pubkey", pubkey.to_string(), String),);
```
ryoqun, line 729: nits: too many INFO logs,  maybe?

runtime/src/accounts_index.rs
```diff
@@ -181,9 +192,20 @@ impl<'a, T: 'static + Clone> Iterator for AccountsIndexIterator<'a, T> {
...
 
+#[derive(Debug, Default)]
+pub struct MintIndex {
+    index: DashMap<Pubkey, MintIndexEntry>,
+    // Map from pubkey to the mint it's stored in, used for cleanup.
+    // Alternative is to store Option<Pubkey> in each AccountInfo in the
+    // AccountsIndex if something is an SPL account with a mint, but then
+    // every AccountInfo would have to allocate `Option<Pubkey>`
+    accounts_to_mints: DashMap<Pubkey, DashSet<Pubkey>>,
+}
```
ryoqun, line 203: Given, 20M~50M spl-token accounts, this new secondary index could consume 2G~5G RAM considering 100 bytes per entry (=account) as a rough estimate with safe buffer?

sakridge:
Can we make this extra indexing optional.

core/src/rpc.rs
```diff
@@ -1155,18 +1156,9 @@ impl JsonRpcRequestProcessor {
                 "Invalid param: not a v2.0 Token mint".to_string(),
             ));
         }
-        let filters = vec![
```
ryoqun, line 1158: here

runtime/src/accounts_index.rs
```diff
@@ -376,6 +412,43 @@ impl<T: 'static + Clone> AccountsIndex<T> {
...
+                .iter()
+                .map(|pubkey_ref| *pubkey_ref)
+                .collect()
+        };
+
+        for pubkey in mint_keys {
+            // Maybe these reads from the AccountsIndex can be batched everytime it
+            // grabs the read lock as well...
+            if let Some((list_r, index)) = self.get(&pubkey, Some(ancestors), max_root) {
+                func(
```
ryoqun, line 444: I guess we need similar check as the original code had there: https://github.com/solana-labs/solana/pull/14212/files?diff=split&w=1#r546409102.

The reason is that mint_index could be stale (maybe?) if the account is closed and used for accounts which isn't owned by mint.

runtime/src/accounts_index.rs
ryoqun: yeah, I think this is true; @carllin how about briefly commenting about this in the source code?

runtime/src/accounts_db.rs
```diff
@@ -2896,33 +2925,34 @@ impl AccountsDB {
                         let entry = accum
                             .entry(stored_account.meta.pubkey)
                             .or_insert_with(Vec::new);
+                        self.accounts_index.update_secondary_indexes(
```
ryoqun, line 2928: well, secure snapshot isn't still a thing. But, I think this is newly susceptible to this secondary index poisoning? Or at least, unbounded memory growth?

So, we should check actual aliveness of each entries in the mint index comparing against the accounts index at last stage of `generate_index()`.

ryoqun: Or we can just add TODO comment.

Also, I think we can construct mint index at last of `generate_index()` using `account_index` after the whole account hash check, maybe?

ryoqun:
> Can we make this extra indexing optional.

+1

@carllin roughly reviewed this.

runtime/src/accounts_index.rs
carllin: I've updated in the latest commit https://github.com/solana-labs/solana/pull/14212/commits/f4f906a9531d1ff9565996cd8848c712a58d9f68 to track slots in the index, so `purge_keys()` and `purge_keys_exact()` should take care of zero-lamport accounts similar to the regular index path.

runtime/src/accounts_db.rs
carllin: @ryoqun updated to track the slots

runtime/src/accounts_index.rs
carllin: Yup that's a good catch! if you have a pubkey with a mint in slot 0, and then it's updated to a non-owned account in slot 1 in the same fork, then the initial code would still return the slot 0 entry even when there's a scan done on slot 1. The original check has been re-added and @CriesofCarrots has added a test here as well: https://github.com/solana-labs/solana/pull/14212/files#diff-ed47b4a0198313377e091bb3957bbbc63d937805426d1b2b6de39d0a50d32a0cR8667

runtime/src/accounts_index.rs
```diff
@@ -181,9 +201,169 @@ impl<'a, T: 'static + Clone> Iterator for AccountsIndexIterator<'a, T> {
...
+
+        // Check if the entry for `inner_key` in the reverse index is empty
+        // and can be removed
+        let needs_remove = {
+            if let Some(slots_to_remove) = slots_to_remove {
+                self.reverse_index
+                    .get(inner_key)
+                    .map(|slots_map| {
+                        for slot in slots_to_remove.iter() {
+                            if let Some(removed) = slots_map.value().remove(slot) {
```
carllin, line 312: @ryoqun  I think this is safe because when `slots_to_remove == None` is passhere:  https://github.com/solana-labs/solana/blob/f4f906a9531d1ff9565996cd8848c712a58d9f68/runtime/src/accounts_index.rs#L669, so it shouldn't clear any other thread's writes to this slot's secondary index, the accounts index lock for the slot is held here: https://github.com/solana-labs/solana/blob/f4f906a9531d1ff9565996cd8848c712a58d9f68/runtime/src/accounts_index.rs#L666

runtime/src/accounts_index.rs
```diff
@@ -181,9 +201,172 @@ impl<'a, T: 'static + Clone> Iterator for AccountsIndexIterator<'a, T> {
...
```
carllin, line 315: @ryoqun, this should be safe because when `slots_to_remove == None` is passed here:  https://github.com/solana-labs/solana/blob/51ebca2a6cd2e17879d7bbae28a2250746c356fd/runtime/src/accounts_index.rs#L676, the accounts index lock for the pubkey is held here: https://github.com/solana-labs/solana/blob/51ebca2a6cd2e17879d7bbae28a2250746c356fd/runtime/src/accounts_index.rs#L668, so it shouldn't clear any other thread's writes to this pubkey's index.

runtime/src/accounts_index.rs
```diff
@@ -181,9 +201,172 @@ impl<'a, T: 'static + Clone> Iterator for AccountsIndexIterator<'a, T> {
...
+                    for (slot, key) in removed_slot_map {
+                        key_to_removed_slots.entry(key).or_default().push(slot);
+                    }
+                }
+                // We just removed the key, no need to remove it again
+                false
+            }
+        };
+
+        if needs_remove {
```
carllin, line 336: @ryoqun would be good to have you  double check this type of delete only if empty logic, it shows up quite a few times in this PR, sorry!

The get called from `purge`, `purge_exact()` and `clean_rooted_entries`. The write lock for the account pubkey in the AccountsIndex is always held for these, so in the mapping

Mint pubkey -> Account pubkey -> Slots, the operations on the account and slot clearing should be safe, the only concurrent access is on the Mint pubkey, but I think the double checking empty before removing here: https://github.com/solana-labs/solana/blob/51ebca2a6cd2e17879d7bbae28a2250746c356fd/runtime/src/accounts_index.rs#L289-L290 should cover that as well

core/src/rpc.rs
```diff
@@ -1299,6 +1292,103 @@ impl JsonRpcRequestProcessor {
...
+    fn get_filtered_spl_token_accounts_by_owner(
+        &self,
+        bank: &Arc<Bank>,
+        owner_key: &Pubkey,
+        mut filters: Vec<RpcFilterType>,
+    ) -> Vec<(Pubkey, Account)> {
+        // The by-owner accounts index checks for Token Account state and Owner address on inclusion.
+        // However, due to the current AppendVec implementation, accounts may remain in the index after
+        // being wiped and reinitialized as something else. We include the redundant filters here to
+        // avoid returning these accounts.
```
ryoqun, line 1330: :+1:

core/src/rpc.rs
```diff
@@ -1299,6 +1292,103 @@ impl JsonRpcRequestProcessor {
...
+        filters.push(RpcFilterType::Memcmp(Memcmp {
+            offset: 0,
+            bytes: MemcmpEncodedBytes::Binary(mint_key.to_string()),
+            encoding: None,
+        }));
+        if self.config.supported_indexes.contains(&IndexType::Mint) {
+            bank.get_filtered_indexed_accounts(&IndexKey::Mint(*mint_key), |account| {
+                filters.iter().all(|filter_type| match filter_type {
+                    RpcFilterType::DataSize(size) => account.data.len() as u64 == *size,
+                    RpcFilterType::Memcmp(compare) => compare.bytes_match(&account.data),
```
ryoqun, line 1385: Well, bytes_match does `bs58::decode()`.. Meaning we waste cpu cycles by rather heavy decoding for the number of returned accounts? I wonder how this could be next bottoleneck after the introduction of indexes..

If not too much hassle, could you decode when we construct `RpcFilterType::Memcmp`?

Admittedly, that's a bit unfortunate we must do this extra filter according to this comment: https://github.com/solana-labs/solana/pull/14212/files#diff-cf429ddb403b689f476ba2ee33dc2f4b8a509e93f5b5f901f38874a21cc65959R1366

ryoqun: Also, this could improve DX abit. That's because we're squashing base58 decoding errors inside `bytes_match` into brutal `false` (not hit)..

core/src/rpc.rs
```diff
@@ -1299,6 +1292,103 @@ impl JsonRpcRequestProcessor {
...
+        // being wiped and reinitialized as something else. We include the redundant filters here to
+        // avoid returning these accounts.
+        //
+        // Filter on Token Account state
+        filters.push(RpcFilterType::DataSize(
+            TokenAccount::get_packed_len() as u64
+        ));
+        // Filter on Owner address
+        filters.push(RpcFilterType::Memcmp(Memcmp {
+            offset: 32,
```
ryoqun, line 1338: nits: could live in inline_spl_token_v2_0?

core/src/rpc.rs
```diff
@@ -1397,20 +1487,32 @@ fn get_encoded_account(
...
+    if program_id != &spl_token_id_v2_0() {
+        return None;
+    }
+    let mut data_size_filter: Option<u64> = None;
+    let mut owner_key: Option<Pubkey> = None;
+    for filter in filters {
+        match filter {
+            RpcFilterType::DataSize(size) => data_size_filter = Some(*size),
+            RpcFilterType::Memcmp(Memcmp {
+                offset: PUBKEY_BYTES,
```
ryoqun, line 1500: maybe related to https://github.com/solana-labs/solana/pull/14212/files#r548046729,

this is rather SPL_TOKEN_ACCOUNT_OWNER_OFFSET or something? I mean, this is depending on particular data layout of spl-token account, not the size of Pubkey.

validator/src/main.rs
```diff
@@ -1438,6 +1448,22 @@ pub fn main() {
...
     let contact_debug_interval = value_t_or_exit!(matches, "contact_debug_interval", u64);
 
+    let mut supported_indexes: Vec<IndexType> = vec![];
+    for index in matches
+        .values_of("indexes")
+        .unwrap_or_default()
+        .map(|value| match value {
+            "program-id" => IndexType::ProgramId,
+            "token-owner" => IndexType::TokenOwner,
+            "token-mint" => IndexType::Mint,
```
ryoqun, line 1458: nits: I love naming consistency, so I humbly request this:

```rust
            "program-id" => IndexType::ProgramId,
            "spl-token-mint" => IndexType::SplTokenMint,
            "spl-token-owner" => IndexType::SplTokenOwner,
```

Notice the subtlety of reordering to align with byte layout of spl-token account. :)

core/src/rpc.rs
```diff
@@ -109,6 +110,7 @@ pub struct JsonRpcConfig {
     pub enable_bigtable_ledger_storage: bool,
     pub enable_bigtable_ledger_upload: bool,
     pub max_multiple_accounts: Option<usize>,
+    pub supported_indexes: Vec<IndexType>,
```
mvines, line 113:
```suggestion
    pub account_indexes: Vec<AccountIndex>,
```

core/src/rpc.rs
```diff
@@ -1299,6 +1292,103 @@ impl JsonRpcRequestProcessor {
...
+        // being wiped and reinitialized as something else. We include the redundant filters here to
+        // avoid returning these accounts.
+        //
+        // Filter on Token Account state
+        filters.push(RpcFilterType::DataSize(
+            TokenAccount::get_packed_len() as u64
+        ));
+        // Filter on Mint address
+        filters.push(RpcFilterType::Memcmp(Memcmp {
+            offset: 0,
```
ryoqun, line 1377: related to this, https://github.com/solana-labs/solana/pull/14212/files#r548052293

this is rather SPL_TOKEN_ACCOUNT_MINT_OFFSET or something?

runtime/src/accounts_index.rs
```diff
@@ -23,6 +30,26 @@ pub type RefCount = u64;
...
+enum ScanTypes<R: RangeBounds<Pubkey>> {
+    Standard(Option<R>),
+    Indexed(IndexKey),
+}
+
+#[derive(Debug, Clone, Copy)]
+pub enum IndexKey {
+    ProgramId(Pubkey),
+    TokenOwner(Pubkey),
+    Mint(Pubkey),
```
ryoqun, line 44: heh, what a coincidence where all of our current secondary indexes are all indexed by `Pubkey`s. :)

core/src/rpc.rs
mvines: Or if renaming `IndexType` is too much it would be good to only import `accounts_index` and qualify this and other usage to `accounts_index::IndexType`.

validator/src/main.rs
```diff
@@ -1361,6 +1362,15 @@ pub fn main() {
                 .takes_value(false)
                 .help("Use the just-in-time compiler instead of the interpreter for BPF."),
         )
+        .arg(
+            Arg::with_name("indexes")
+                .long("index")
```
mvines, line 1367:
```suggestion
            Arg::with_name("account_indexes")
                .long("account-index")
```
"index" is a little too generic

validator/src/main.rs
```diff
@@ -1438,6 +1448,22 @@ pub fn main() {
...
+        .values_of("indexes")
+        .unwrap_or_default()
+        .map(|value| match value {
+            "program-id" => IndexType::ProgramId,
+            "token-owner" => IndexType::TokenOwner,
+            "token-mint" => IndexType::Mint,
+            _ => unreachable!(),
+        })
+    {
+        if !supported_indexes.contains(&index) {
```
mvines, line 1462: Just use a HashSet to avoid this block?

core/src/validator.rs
```diff
@@ -108,6 +109,7 @@ pub struct ValidatorConfig {
     pub bpf_jit: bool,
     pub send_transaction_retry_ms: u64,
     pub send_transaction_leader_forward_count: u64,
+    pub supported_indexes: Vec<IndexType>,
```
mvines, line 112: Another example of where `supported_indexes` is a little too generic:
```suggestion
    pub account_indexes: Vec<accounts_index::IndexType>,
```

runtime/src/accounts_index.rs
mvines, line 44: Make these names SPL token specific to avoid future confusion:
```suggestion
    SplTokenOwner(Pubkey),
    SplTokenMint(Pubkey),
```

runtime/src/accounts_index.rs
```diff
@@ -23,6 +30,26 @@ pub type RefCount = u64;
...
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum IndexType {
+    ProgramId,
+    TokenOwner,
+    Mint,
```
mvines, line 51:
```suggestion
    SplTokenOwner,
    SplTokenMint,
```

runtime/src/accounts_index.rs
```diff
@@ -181,9 +208,198 @@ impl<'a, T: 'static + Clone> Iterator for AccountsIndexIterator<'a, T> {
...
+        }
+    }
+}
+
 #[derive(Debug, Default)]
 pub struct AccountsIndex<T> {
     pub account_maps: RwLock<AccountMap<Pubkey, AccountMapEntry<T>>>,
+    program_id_index: SecondaryIndex,
+    token_owner_index: SecondaryIndex,
+    mint_index: SecondaryIndex,
```
mvines, line 402:
```suggestion
    spl_token_owner_index: SecondaryIndex,
    spl_token_mint_index: SecondaryIndex,
```

runtime/src/accounts_index.rs
```diff
@@ -181,9 +208,198 @@ impl<'a, T: 'static + Clone> Iterator for AccountsIndexIterator<'a, T> {
...
+        if is_key_empty {
+            if let dashmap::mapref::entry::Entry::Occupied(key_entry) = self.index.entry(*key) {
+                if key_entry.get().is_empty() {
+                    key_entry.remove();
+                }
+            }
+        }
+    }
+
+    // Specifying `slots_to_remove` == None will only remove keys for those specific slots
```
ryoqun, line 326: typo: `None` => `Some`?

core/src/rpc.rs
garywang, line 1330: This means we also have to check that the program ID still matches, right?

runtime/src/accounts_index.rs
```diff
@@ -181,9 +208,198 @@ impl<'a, T: 'static + Clone> Iterator for AccountsIndexIterator<'a, T> {
...
+                                    .or_default()
+                                    .push(*slot);
+                            }
+                        }
+                        w_slots_map.is_empty()
+                    })
+                    .unwrap_or(false)
+            } else {
+                if let Some((_, removed_slot_map)) = self.reverse_index.remove(inner_key) {
+                    for (slot, key) in removed_slot_map.into_inner().unwrap().into_iter() {
```
ryoqun, line 360: nits: rename to `removed_key` to align with the code in the above `if`-then clause.

runtime/src/accounts_index.rs
```diff
@@ -181,9 +208,198 @@ impl<'a, T: 'static + Clone> Iterator for AccountsIndexIterator<'a, T> {
...
 
+#[derive(Debug, Default)]
+struct SecondaryIndex {
+    // Map from index keys to index values
+    index: DashMap<Pubkey, SecondaryIndexEntry>,
+    // Map from index values back to index keys, used for cleanup.
+    // Alternative is to store Option<Pubkey> in each AccountInfo in the
+    // AccountsIndex if something is an SPL account with a mint, but then
+    // every AccountInfo would have to allocate `Option<Pubkey>`
+    reverse_index: DashMap<Pubkey, RwLock<HashMap<Slot, Pubkey>>>,
```
ryoqun, line 219: nits: define `type SecondaryReverseIndexEntry`?

core/src/rpc.rs
CriesofCarrots: 💯, now that these pathways are split in accounts.rs. Thank you, good catch!

core/src/rpc.rs
CriesofCarrots: No problem to rename!

runtime/src/accounts_index.rs
```diff
@@ -23,6 +30,26 @@ pub type RefCount = u64;
 pub type AccountMap<K, V> = BTreeMap<K, V>;
 
 type AccountMapEntry<T> = Arc<AccountMapEntryInner<T>>;
+type SecondaryIndexEntry = DashMap<Pubkey, RwLock<HashSet<Slot>>>;
+
+enum ScanTypes<R: RangeBounds<Pubkey>> {
+    Standard(Option<R>),
```
ryoqun, line 36: nits: rename to `Full{Scan}`. well, I think it's clearer than the Standard of Scan

runtime/src/accounts_index.rs
```diff
@@ -181,9 +208,198 @@ impl<'a, T: 'static + Clone> Iterator for AccountsIndexIterator<'a, T> {
...
+    index: DashMap<Pubkey, SecondaryIndexEntry>,
+    // Map from index values back to index keys, used for cleanup.
+    // Alternative is to store Option<Pubkey> in each AccountInfo in the
+    // AccountsIndex if something is an SPL account with a mint, but then
+    // every AccountInfo would have to allocate `Option<Pubkey>`
+    reverse_index: DashMap<Pubkey, RwLock<HashMap<Slot, Pubkey>>>,
+}
+
+impl SecondaryIndex {
+    fn insert(&self, key: &Pubkey, inner_key: &Pubkey, slot: Slot) {
```
ryoqun, line 223: I'll later review this code block in full....

runtime/src/accounts_index.rs
CriesofCarrots: Done in https://github.com/solana-labs/solana/pull/14212/commits/d204a08432c3936cd1450cf16e1a67236e758461

runtime/src/accounts_index.rs
CriesofCarrots: Done in https://github.com/solana-labs/solana/pull/14212/commits/d204a08432c3936cd1450cf16e1a67236e758461

runtime/src/accounts_index.rs
CriesofCarrots: Done in https://github.com/solana-labs/solana/pull/14212/commits/d204a08432c3936cd1450cf16e1a67236e758461

core/src/validator.rs
CriesofCarrots: Done in https://github.com/solana-labs/solana/pull/14212/commits/d204a08432c3936cd1450cf16e1a67236e758461

validator/src/main.rs
CriesofCarrots: Done in https://github.com/solana-labs/solana/pull/14212/commits/d204a08432c3936cd1450cf16e1a67236e758461

core/src/rpc.rs
CriesofCarrots: Done in https://github.com/solana-labs/solana/pull/14212/commits/d204a08432c3936cd1450cf16e1a67236e758461

validator/src/main.rs
CriesofCarrots: Done in https://github.com/solana-labs/solana/pull/14212/commits/d204a08432c3936cd1450cf16e1a67236e758461

core/src/rpc.rs
CriesofCarrots: Done in https://github.com/solana-labs/solana/pull/14212/commits/d607d1aa5da54b078f1eb27bcfe1752df515072e

core/src/rpc.rs
CriesofCarrots: Done in https://github.com/solana-labs/solana/pull/14212/commits/d607d1aa5da54b078f1eb27bcfe1752df515072e

core/src/rpc.rs
CriesofCarrots: Done in https://github.com/solana-labs/solana/pull/14212/commits/d607d1aa5da54b078f1eb27bcfe1752df515072e

core/src/rpc.rs
CriesofCarrots: That's a really nice idea! I think I'd prefer to do that in a small follow-up PR

validator/src/main.rs
CriesofCarrots: Done in https://github.com/solana-labs/solana/pull/14212/commits/75eb65a7825c7d115d887a4a4e4a81a83f96f614

core/src/rpc.rs
CriesofCarrots: Done in https://github.com/solana-labs/solana/pull/14212/commits/0f3c1299ae01163835777c3f45abd6fb3049e606

runtime/src/serde_snapshot.rs
```diff
@@ -260,12 +264,14 @@ fn reconstruct_accountsdb_from_fields<E, P>(
...
     stream_append_vecs_path: P,
     cluster_type: &ClusterType,
+    account_indexes: HashSet<AccountIndex>,
 ) -> Result<AccountsDB, Error>
 where
     E: Into<AccountStorageEntry>,
     P: AsRef<Path>,
 {
     let mut accounts_db = AccountsDB::new(account_paths.to_vec(), cluster_type);
+    accounts_db.account_indexes = account_indexes;
```
CriesofCarrots, line 274:
```suggestion
    let mut accounts_db = AccountsDB::new_with_indexes(account_paths.to_vec(), cluster_type, account_indexes);
```

runtime/src/accounts_index.rs
carllin: ah good catch, thanks!

runtime/src/accounts_index.rs
carllin: Renamed to UnIndexed, because sometimes it's a range scan, not necessarily a full scan

runtime/src/accounts_db.rs
ryoqun: @carllin ah, this security concern is maybe fixed now because purge_secondary_indexes_by_inner_key is correctly wired in the primary index codepath, right?

runtime/src/accounts_db.rs
```diff
@@ -2896,33 +2940,36 @@ impl AccountsDB {
...
                     account_infos.sort_by(|a, b| a.0.cmp(&b.0));
                     for (_, account_info) in account_infos {
-                        self.accounts_index.upsert(
-                            *slot,
-                            pubkey,
-                            account_info.clone(),
-                            &mut _reclaims,
-                        );
+                        self.accounts_index
+                            .upsert(*slot, &pubkey, account_info, &mut _reclaims);
```
ryoqun, line 2972: how about moving `update_secondary_indexes` https://github.com/solana-labs/solana/pull/14212/files#diff-1090394420d51617f3233275c2b65ed706b35b53b115fe65f82c682af8134a6fR2943 here just align with `AccountsDB::update_index`? Or is there any reason, we must do `update_secondary_indexes` earlier than this `upsert` for the snapshot (generate_index) case?

Moreover, I think we could even move the invocation of `update_secondary_indexes` inside `upsert` itself. So, this will clarify that we're updating primary index and secondary index at the same time.

runtime/src/accounts_index.rs
ryoqun: it looks like this if body isn't tested at all?: https://codecov.io/gh/solana-labs/solana/pull/14212/diff#D1-376 I'd like a one test at least. :)

image

runtime/src/accounts_index.rs
```diff
@@ -181,9 +211,201 @@ impl<'a, T: 'static + Clone> Iterator for AccountsIndexIterator<'a, T> {
...
+        }
+    }
+
+    // Specifying `slots_to_remove` == Some will only remove keys for those specific slots
+    // found for the `inner_key` in the reverse index. Otherwise, passing `None`
+    // will  remove all keys that are found for the `inner_key` in the reverse index.
+
+    // Note passing `None` is dangerous unless you're sure there's no other competing threads
+    // writing updates to the index for this Pubkey at the same time!
+    fn remove_by_inner_key(&self, inner_key: &Pubkey, slots_to_remove: Option<&HashSet<Slot>>) {
```
ryoqun, line 335: how about renaming this just to `fn remove`?

I sensed that this fn is at same level conceptually like `fn insert` and `fn get` (i.e. directly called from others). Alternatively, rename to `{insert/get/remove}_by_inner_key`?

I usually like verbose name for internal only fn and terse name for internal+external fn.

runtime/src/accounts_index.rs
```diff
@@ -181,9 +211,201 @@ impl<'a, T: 'static + Clone> Iterator for AccountsIndexIterator<'a, T> {
...
+                    }
+                }
+                w_slot_set.is_empty()
+            } else {
+                false
+            };
+
+            // Check if `key` is empty
+            let is_key_empty = if is_inner_key_empty {
+                if let dashmap::mapref::entry::Entry::Occupied(inner_key_entry) =
```
ryoqun, line 297: nits: let's use `use`? (pun intended)

runtime/src/accounts_index.rs
```diff
@@ -181,9 +211,201 @@ impl<'a, T: 'static + Clone> Iterator for AccountsIndexIterator<'a, T> {
...
+                slots_map.write().unwrap().insert(slot, *key)
+            } else {
+                None
+            }
+        };
+
+        if let Some(prev_key) = prev_key {
+            // If the inner key was moved to a different primary key, remove
+            // the previous index entry.
+            if prev_key != *key {
```
ryoqun, line 272: I think this check can be proofed to be unnecessary? i.e. if `prev_key`  is Some, it's always `prev_key` != `*key`.

runtime/src/accounts_index.rs
```diff
@@ -181,9 +211,201 @@ impl<'a, T: 'static + Clone> Iterator for AccountsIndexIterator<'a, T> {
...
+                false
+            };
+            is_key_empty
+        } else {
+            false
+        };
+
+        // Delete the `key` if the set of inner keys is empty
+        if is_key_empty {
+            if let dashmap::mapref::entry::Entry::Occupied(key_entry) = self.index.entry(*key) {
```
ryoqun, line 321: nits: let's use `use`?

runtime/src/accounts_index.rs
```diff
@@ -181,9 +211,201 @@ impl<'a, T: 'static + Clone> Iterator for AccountsIndexIterator<'a, T> {
...
+    // writing updates to the index for this Pubkey at the same time!
+    fn remove_by_inner_key(&self, inner_key: &Pubkey, slots_to_remove: Option<&HashSet<Slot>>) {
+        // Save off which keys in `self.index` had slots removed so we can remove them
+        // after we purge the reverse index
+        let mut key_to_removed_slots: HashMap<Pubkey, Vec<Slot>> = HashMap::new();
+
+        // Check if the entry for `inner_key` in the reverse index is empty
+        // and can be removed
+        let needs_remove = {
+            if let Some(slots_to_remove) = slots_to_remove {
```
ryoqun, line 343: here

runtime/src/accounts_index.rs
```diff
@@ -611,22 +943,40 @@ impl<T: 'static + Clone> AccountsIndex<T> {
         }
     }
 
+    fn purge_secondary_indexes_by_inner_key(
+        &self,
+        inner_key: &Pubkey,
+        slots_to_remove: Option<&HashSet<Slot>>,
+    ) {
+        self.program_id_index
```
ryoqun, line 951: maybe bail out early if we aren't indexing like indexing path to avoid this meaningless loop over N where number of slots (3x empty lookup on every account write)?: https://github.com/solana-labs/solana/pull/14212/files#r548632636

runtime/src/accounts_index.rs
```diff
@@ -581,6 +881,38 @@ impl<T: 'static + Clone> AccountsIndex<T> {
...
+        &self,
+        pubkey: &Pubkey,
+        slot: Slot,
+        account_owner: &Pubkey,
+        account_data: &[u8],
+        account_indexes: &HashSet<AccountIndex>,
+    ) {
+        if account_indexes.contains(&AccountIndex::ProgramId) {
+            self.program_id_index.insert(account_owner, pubkey, slot);
+        }
```
ryoqun, line 894: maybe add comment like this?

``` rust
// we skip updating even if given pubkey's account is reset to Account::Default due to `lamports == 0`.
// Skipping means not updating secondary index to mark the account as missing.
// This doesn't introduce false positives (= wrong index hit) because reset accounts has already
// previous updates at least, otherwise it would be purged altogether via purge_keys_exact in `fn clean_accounts`,
// including the secondary indexes.
```

runtime/src/accounts_db.rs
```diff
@@ -2458,6 +2495,13 @@ impl AccountsDB {
             let pubkey = pubkey_account.0;
             self.accounts_index
```
ryoqun, line 2496: nits: maybe add comment here like this?

```rust
// we don't atomically update both primary index and secondary index together.
// This certainly creates small time window with inconsistent state across the two indexes.
// But, this is acceptable because:
//  - we don't care about strict consistent view at any given moment of time because only working 
//     bank's slot will suffer from this. we expects rooted slot is used more often
//  - secondary index is always used while being dependant on the primary index by means of
//      post-filtering via the primary index, not used by its own. So, what accounts_index sees is
//      enough for us to regard as the source of truth.
```

runtime/src/accounts_index.rs
```diff
@@ -181,9 +211,201 @@ impl<'a, T: 'static + Clone> Iterator for AccountsIndexIterator<'a, T> {
...
+            };
+            is_key_empty
+        } else {
+            false
+        };
+
+        // Delete the `key` if the set of inner keys is empty
+        if is_key_empty {
+            if let dashmap::mapref::entry::Entry::Occupied(key_entry) = self.index.entry(*key) {
+                if key_entry.get().is_empty() {
```
ryoqun, line 322: maybe add comment like this considering your comment(https://github.com/solana-labs/solana/pull/14212/files#r547081588)?:

```
// Other threads might interleaved by having written to this `key`,
// so double-check again for its emptiness 
```

runtime/src/accounts_index.rs
ryoqun: > Mint pubkey -> Account pubkey -> Slots, the operations on the account and slot clearing should be safe, the only concurrent access is on the Mint pubkey, but I think the double checking empty before removing here:
> 
> https://github.com/solana-labs/solana/blob/51ebca2a6cd2e17879d7bbae28a2250746c356fd/runtime/src/accounts_index.rs#L289-L290
> should cover that as well

Yeah, I think so too.

runtime/src/accounts_index.rs
ryoqun: done. hooray! it was super dense review session.

core/src/rpc.rs
```diff
@@ -1299,6 +1293,111 @@ impl JsonRpcRequestProcessor {
...
+
+    /// Get an iterator of spl-token accounts by owner address
+    fn get_filtered_spl_token_accounts_by_owner(
+        &self,
+        bank: &Arc<Bank>,
+        owner_key: &Pubkey,
+        mut filters: Vec<RpcFilterType>,
+    ) -> Vec<(Pubkey, Account)> {
+        // The by-owner accounts index checks for Token Account state and Owner address on inclusion.
+        // However, due to the current AppendVec implementation, accounts may remain in the index after
```
ryoqun, line 1331: nits: AppendVec => AccountsDB? I think this limitation derives from the overall AccountsDB implementation details.

core/src/rpc.rs
```diff
@@ -1299,6 +1293,111 @@ impl JsonRpcRequestProcessor {
...
+
+    /// Get an iterator of spl-token accounts by mint address
+    fn get_filtered_spl_token_accounts_by_mint(
+        &self,
+        bank: &Arc<Bank>,
+        mint_key: &Pubkey,
+        mut filters: Vec<RpcFilterType>,
+    ) -> Vec<(Pubkey, Account)> {
+        // The by-mint accounts index checks for Token Account state and Mint address on inclusion.
+        // However, due to the current AppendVec implementation, accounts may remain in the index after
```
ryoqun, line 1371: nits: AppendVec => AccountsDB? I think this limitation derives from the overall AccountsDB implementation details.

ryoqun (approved):
lgtm after all comments are resolved. (should be straightforward)

this pr is very great work and you're so fast to create such as secondary indexing. I'm really impressed. :)

runtime/src/accounts_db.rs
carllin: @ryoqun that's a good point, and one I was thinking about as well when initially implementing it. I think ideally the `update_secondary_indexes ` is called inside `upsert`, but my primary challenge came from the `generate_index` function  Note in that function, here is the update to the secondary index: https://github.com/solana-labs/solana/blob/f4ff936fb8ff64ebec752fc28c0cafde3879f16b/runtime/src/accounts_db.rs#L2943-L2949, and here is the update to the original index: https://github.com/solana-labs/solana/blob/f4ff936fb8ff64ebec752fc28c0cafde3879f16b/runtime/src/accounts_db.rs#L2972

Concerns:
1) `update_secondary_indexes` requires the account data, which is only available as a reference in the scan here, https://github.com/solana-labs/solana/blob/f4ff936fb8ff64ebec752fc28c0cafde3879f16b/runtime/src/accounts_db.rs#L2943-L2949.
So we would either have to clone out this account data so that the later call to `upsert` can be passed that account data, or:
2) We move upsert inside the scan there. Functionally this should be equivalent, as long as `upsert` is thread safe for the `par_iter` inside `scan_account_storage_inner`. I wasn't 100% sure this was the case, so I decided to mitigate risk and not do that here.

core/src/rpc.rs
carllin: agreed!

runtime/src/accounts_index.rs
carllin: agreed, more documentation the better in this code path!

runtime/src/accounts_db.rs
carllin: That's a really good point. Also scans are only supported on `frozen` banks, so if somebody is trying to do a scan on a working bank, it is not required to be consistent.

runtime/src/accounts_index.rs
carllin: @ryoqun I think the false positives warning is helpful, but the justification needs a little tweaking. This is my thoughts:

```
// Note because of the below check below on the account data length, when an
        // account hits zero lamports and is reset to Account::Default, then we skip
        // the below updates to the secondary indexes.
        //
        // Skipping means not updating secondary index to mark the account as missing.
        // This doesn't introduce false positives during a scan because the caller to scan
        // provides the ancestors to check. So even if a zero-lamport account is not yet
        // removed from the secondary index, the scan function will:
        // 1) consult the primary index via `get(&pubkey, Some(ancestors), max_root)`
        // and find the zero-lamport version
        // 2) When the fetch from storage occurs, it will return nothing because the account
        // is default in that storage entry
```

runtime/src/accounts_db.rs
carllin: Edit: I think calling the `update_secondary_indexes` where I have it right now is indeed not safe because it doesn't respect the `stored_account.meta.write_version`, i.e. this sort https://github.com/solana-labs/solana/blob/f4ff936fb8ff64ebec752fc28c0cafde3879f16b/runtime/src/accounts_db.rs#L2969 is critical in knowing how to order the updates to the index. Ordering is important because later updates will pop off earlier ones in the same slot.

runtime/src/accounts_index.rs
carllin: @ryoqun I wanted to specify a difference between removing using the key from the `reverse_index == account key` vs removing using the primary secondary key (mint, program_id, token owner).

runtime/src/accounts_index.rs
carllin: So the fundamental question is:

1) 
```
let should_insert = {
                // Most of the time, key should already exist and match
                // the one in the update
                if let Some(existing_key) = slots_map.read().unwrap().get(&slot) {
                    existing_key != key
                } else {
                    // If there is no key yet, then insert
                    true
                }
            };
```

2) 
```
if should_insert {
    slots_map.write().unwrap().insert(slot, *key)
}
```

Between the check in 1) and the write in 2) is it possible another thread already wrote the same `slot` to `slots_map`?

In the current implementation of AccountsDb, because a pubkey is exclusively updated by one thread per slot due to the Accounts locks, this may be true. But if ever in the future this changes, this may not hold. For instance if somebody was to try and introduce a bug like this: https://github.com/solana-labs/solana/pull/14212#discussion_r548774161 where you concurrently insert multiple versions of a pubkey from the *same* slot into the index at once 😉 

What do you think? Leave it out because it's currently unnecessary?

runtime/src/accounts_index.rs
carllin: @ryoqun do you mean switch the indexes to Option types (instantiate based on validator args) and don't call these functions if those indexes are `None`?

runtime/src/accounts_db.rs
carllin: @ryoqun yeah it should be bounded by the accounts_index now

runtime/src/accounts_db.rs
carllin: Updated here: https://github.com/solana-labs/solana/pull/14212/commits/84d82f5ecd3da227f9963275a159d44a37960e17

runtime/src/accounts_db.rs
```diff
@@ -2456,8 +2493,26 @@ impl AccountsDB {
...
+            // This certainly creates small time window with inconsistent state across the two indexes.
+            // However, this is acceptable because:
+            //
+            //  - A strict consistent view at any given moment of time is not necessary, because the only
+            //  use case for the secondary index is `scan`, and `scans` are only supported/require consistency
+            //  on frozen banks, and this inconsistency is only possible on working banks.
+            //
+            //  - The secondary index is never consulted as primary source of truth for gets/stores.
+            //  So, what the accounts_index sees alone is sufficient as a source of truth for other non-scan
+            //  account operations.
```
ryoqun, line 2506: I'm seeing your great rewrite to move inside `upsert`. :) I think this block comment as well is moved there because we don't see two indexes here anymore.

runtime/src/accounts_db.rs
```diff
@@ -2881,52 +2938,68 @@ impl AccountsDB {
...
-
-            let mut accounts_map: HashMap<Pubkey, Vec<(u64, AccountInfo)>> = HashMap::new();
-            for accumulator_entry in accumulator.iter() {
-                for (pubkey, storage_entry) in accumulator_entry {
-                    let entry = accounts_map.entry(*pubkey).or_insert_with(Vec::new);
-                    entry.extend(storage_entry.iter().cloned());
-                }
-            }
-
+            let accounts_map: AccountsMap = DashMap::new();
```
ryoqun, line 2941: nits: `let accounts_map = AccountsMap::new();` doesn't work?

runtime/src/accounts_db.rs
```diff
@@ -2881,52 +2938,68 @@ impl AccountsDB {
...
+            });
             // Need to restore indexes even with older write versions which may
             // be shielding other accounts. When they are then purged, the
             // original non-shielded account value will be visible when the account
             // is restored from the append-vec
-            if !accumulator.is_empty() {
+            if !accounts_map.is_empty() {
                 let mut _reclaims: Vec<(u64, AccountInfo)> = vec![];
-                for (pubkey, account_infos) in accounts_map.iter_mut() {
-                    account_infos.sort_by(|a, b| a.0.cmp(&b.0));
```
ryoqun, line 2918: nice rewrite; I see the reason this `.sort` is alternatively accomplished just by using `BTreeMap`. :)

runtime/src/accounts_index.rs
```diff
@@ -581,19 +879,67 @@ impl<T: 'static + Clone> AccountsIndex<T> {
         max_root
     }
 
+    pub fn update_secondary_indexes(
```
ryoqun, line 882: nits: no need to be `pub`? :)

runtime/src/accounts_index.rs
ryoqun: yeah, perfect rewrite. :) Thanks for polishing my jottings. :)

runtime/src/accounts_index.rs
ryoqun: @carllin I meant simpler ones like this (similar to `update_secondary_indexes`):

```rust
if account_indexes.contains(&AccountIndex::ProgramId) {
    self.program_id_index.remove_by_inner_key(inner_key, slots_to_remove);
}
if account_indexes.contains(&AccountIndex::SplTokenMint) {
    self.spl_token_mint.remove_by_inner_key(inner_key, slots_to_remove);
}
```

runtime/src/accounts_index.rs
ryoqun: ok, you're right. :)
let's retain the extra check adding some commentary like this:

```rust
// Interleaved other thread's writes between 
// `should_insert = { ... slots_map.get(...) ... }` and  `prev_key = { ... slots_map.insert(...) ... }`
// isn't possible due to current AccountsIndex's pubkey-per-thread exclusive-locking,
// but check here just in case for future more relaxed implementation
if prev_key != *key {
```

runtime/src/accounts_index.rs
ryoqun: :eyes:

runtime/src/accounts_index.rs
```diff
@@ -581,19 +879,67 @@ impl<T: 'static + Clone> AccountsIndex<T> {
...
+        // the below updates to the secondary indexes.
+        //
+        // Skipping means not updating secondary index to mark the account as missing.
+        // This doesn't introduce false positives during a scan because the caller to scan
+        // provides the ancestors to check. So even if a zero-lamport account is not yet
+        // removed from the secondary index, the scan function will:
+        // 1) consult the primary index via `get(&pubkey, Some(ancestors), max_root)`
+        // and find the zero-lamport version
+        // 2) When the fetch from storage occurs, it will return nothing because the account
+        // is default in that storage entry
```
ryoqun, line 904: Well, could you double check? As far as I re-read the code lightly, the following would be more accurate:


```rust
// 2) When the fetch from storage occurs, it will return Account::Default (as persisted tombstone for snapshots). But then
// it'll ultimately filtered out by post-process redundant filters like at get_filtered_spl_token_accounts_by_owner.
```

core/src/rpc.rs
```diff
@@ -1299,6 +1293,111 @@ impl JsonRpcRequestProcessor {
...
+    /// Get an iterator of spl-token accounts by owner address
+    fn get_filtered_spl_token_accounts_by_owner(
+        &self,
+        bank: &Arc<Bank>,
+        owner_key: &Pubkey,
+        mut filters: Vec<RpcFilterType>,
+    ) -> Vec<(Pubkey, Account)> {
+        // The by-owner accounts index checks for Token Account state and Owner address on inclusion.
+        // However, due to the current AccountsDB implementation, accounts may remain in the index after
+        // being wiped and reinitialized as something else. We include the redundant filters here to
```
ryoqun, line 1332: cross reference this subtlety https://github.com/solana-labs/solana/pull/14212/files#r548884667 like this?

```
..., accounts may be zero-lamport Account::Default() or remain in the index after 
being wiped and reinitialized as something else.
```

core/src/rpc.rs
```diff
@@ -1299,6 +1293,111 @@ impl JsonRpcRequestProcessor {
...
+    /// Get an iterator of spl-token accounts by mint address
+    fn get_filtered_spl_token_accounts_by_mint(
+        &self,
+        bank: &Arc<Bank>,
+        mint_key: &Pubkey,
+        mut filters: Vec<RpcFilterType>,
+    ) -> Vec<(Pubkey, Account)> {
+        // The by-mint accounts index checks for Token Account state and Mint address on inclusion.
+        // However, due to the current AccountsDB implementation, accounts may remain in the index after
+        // being wiped and reinitialized as something else. We include the redundant filters here to
```
ryoqun, line 1372: same as https://github.com/solana-labs/solana/pull/14212/files#r548884965

ryoqun (approved):
LGTM (again) This time only with nits. :)

Really I appreciate your last polish-ups.

runtime/src/accounts_index.rs
carllin: added!

runtime/src/accounts_index.rs
carllin: ah yeah, you're right, for some reason I thought `load` in AccountsDb would return `None` for zero-lamport accounts

core/src/rpc.rs
carllin: done!

core/src/rpc.rs
carllin: done!

runtime/src/accounts_index.rs
carllin: agreed, good catch

runtime/src/accounts_db.rs
carllin: ah yup, updated!

runtime/src/accounts_index.rs
carllin: @ryoqun fixed a bug and added tests that should cover this case now: https://github.com/solana-labs/solana/pull/14212/commits/5a46407e73480eed5039d87ef1ae638bbcf6dd77

runtime/src/accounts_index.rs
```diff
@@ -581,19 +883,79 @@ impl<T: 'static + Clone> AccountsIndex<T> {
...
+        // However, this is acceptable because:
+        //
+        //  - A strict consistent view at any given moment of time is not necessary, because the only
+        //  use case for the secondary index is `scan`, and `scans` are only supported/require consistency
+        //  on frozen banks, and this inconsistency is only possible on working banks.
+        //
+        //  - The secondary index is never consulted as primary source of truth for gets/stores.
+        //  So, what the accounts_index sees alone is sufficient as a source of truth for other non-scan
+        //  account operations.
         w_account_entry.update(slot, account_info, reclaims);
```
ryoqun, line 957: nits: maybe we can `drop(w_account_entry)`? I'd prefer `{...}` over `drop()` for readability and maintainability, though.

runtime/src/accounts_index.rs
```diff
@@ -181,9 +211,205 @@ impl<'a, T: 'static + Clone> Iterator for AccountsIndexIterator<'a, T> {
...
+                    }
+                }
+                // We just removed the key, no need to remove it again
+                false
+            }
+        };
+
+        if needs_remove {
+            if let Occupied(slot_map) = self.reverse_index.entry(*inner_key) {
+                if slot_map.get().read().unwrap().is_empty() {
```
ryoqun, line 383: nits: how about commenting this as well?
```
                // Other threads may have interleaved writes to this `inner_key`,
                // so double-check again for its emptiness
```

runtime/src/accounts_index.rs
```diff
@@ -735,13 +1136,31 @@ impl<T: 'static + Clone> AccountsIndex<T> {
     pub fn uncleaned_roots_len(&self) -> usize {
         self.roots_tracker.read().unwrap().uncleaned_roots.len()
     }
+
+    #[cfg(test)]
+    // filter any rooted entries and return them along with a bool that indicates
+    // if this account has no more entries.
+    pub fn purge_roots(&self, pubkey: &Pubkey) -> (SlotList<T>, bool) {
```
ryoqun, line 1143: :+1: for demoting this into `#[cfg(test)]`.

ryoqun: nits: also, simple comment (this doesn't update secondary indexes atm) could save us for lengthy debbuging when we promote back into `#[cfg(not(test))]` again in the distant future.

ryoqun (approved):
re-re-lgtm with nits

Thanks for paying attentions into very little details.

runtime/src/accounts_index.rs
carllin: hmm that's a good point, should be no need to hold the lock.

runtime/src/accounts_index.rs
```diff
@@ -181,9 +211,208 @@ impl<'a, T: 'static + Clone> Iterator for AccountsIndexIterator<'a, T> {
...
+                    }
+                }
+                // We just removed the key, no need to remove it again
+                false
+            }
+        };
+
+        if needs_remove {
+            // Other threads may have interleaved writes to this `inner_key`, between
+            // reeleasing the `self.reverse_index.get(inner_key)` lock and now,
```
CriesofCarrots, line 383:
```suggestion
            // releasing the `self.reverse_index.get(inner_key)` lock and now,
```

CriesofCarrots (dismissed):

soulfulAdam (approved):

PR #14213, opened 2020-12-19
Stream RPC snapshot downloads

### LINKED ISSUE #14211: Snapshot serving causes excessive memory usage
We had to temporarily bump RAM on the mainnet-beta boot nodes to handle a high load of snapshot serving traffic.  ~5 snapshots at second where being served across all boot nodes, at ~1.3GB per snapshot.  

Areas to investigate:
* [ ] Why does serving the same snapshot file concurrently cause 50GB+ of RAM usage.  Is the code serving snapshots doing something silly like loading the archive into RAM entirely?   At 1.3GB per snapshot archive, this would certainly add up if so
* [ ] We probably should rate limit snapshot downloading.  This probably should go into the validator's RPC code itself, as a haproxy-front end won't be as convenient here
* [ ] Some log data, or other metrics about the number of concurrent snapshots being downloaded might be useful
* [ ] Once the issue is resolved, the RAM on the mainnet-beta boot nodes should be restored to normal.

mvines:
Snapshot archives are too big to read into memory before serving them, stream them.  This PR reduces mainnet-beta boot node RAM usage by a solid 100GB.

Also add a metric datapoint for snapshot/genesis downloads
Freebie: uncompressed snapshot downloads now work. 

Fixes #14211


sakridge:
sweet

core/src/rpc_service.rs
```diff
@@ -129,10 +136,13 @@ impl RpcRequestMiddleware {
             response: Box::new(
                 tokio_fs_01::file::File::open(filename)
                     .and_then(|file| {
-                        let buf: Vec<u8> = Vec::new();
-                        tokio_io_01::io::read_to_end(file, buf)
```
ryoqun, line 133: cool

core/src/rpc_service.rs
```diff
@@ -129,10 +136,13 @@ impl RpcRequestMiddleware {
...
-                            .and_then(|item| Ok(hyper::Response::new(item.1.into())))
-                            .or_else(|_| Ok(RpcRequestMiddleware::internal_server_error()))
+                        use tokio_codec_01::{BytesCodec, FramedRead};
+
+                        let stream = FramedRead::new(file, BytesCodec::new())
+                            .map(tokio_01_bytes::BytesMut::freeze);
+                        let body = hyper::Body::wrap_stream(stream);
```
ryoqun, line 143: cool. :)

PR #10719, opened 2020-06-19
Remove unused StakeLockout::lockout

ryoqun:
#### Problem

StakeLockout::lockout isn't used anymore and it does unnecessary computation to mange the field and uses extra memory.

Also, the naming is outdated as a result.

#### Summary of Changes

Remove the field and tidy up the namings of types and fields.

This PR should contain no functional change.

Part of #10718

PR #14217, opened 2020-12-20
improves performance in replay-stage

behzadnouri:
#### Problem
bank::vote_accounts returns a hash-map which is slow to iterate, but all uses
only require an iterator:
https://github.com/solana-labs/solana/blob/b3dc98856/runtime/src/bank.rs#L4300-L4306
Similarly, calculate_stake_weighted_timestamp takes a hash-map whereas it only
requires an iterator:
https://github.com/solana-labs/solana/blob/b3dc98856/sdk/src/stake_weighted_timestamp.rs#L21-L28

#### Summary of Changes
* removed redundant hash-maps in favor of more cache friendly iterators. 
* minor performance improvements in Tower::collect_vote_lockouts.

### DIFF
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -216,6 +216,7 @@ impl Tower {
     where
         F: IntoIterator<Item = (Pubkey, (u64, ArcVoteAccount))>,
     {
+        let mut vote_slots = HashSet::new();
         let mut voted_stakes = HashMap::new();
         let mut total_stake = 0;
         let mut bank_weight = 0;
@@ -278,7 +279,7 @@ impl Tower {
 
             for vote in &vote_state.votes {
                 bank_weight += vote.lockout() as u128 * voted_stake as u128;
-                Self::populate_ancestor_voted_stakes(&mut voted_stakes, &vote, ancestors);
+                vote_slots.insert(vote.slot);
             }
 
             if start_root != vote_state.root_slot {
@@ -289,7 +290,7 @@ impl Tower {
                     };
                     trace!("ROOT: {}", vote.slot);
                     bank_weight += vote.lockout() as u128 * voted_stake as u128;
-                    Self::populate_ancestor_voted_stakes(&mut voted_stakes, &vote, ancestors);
+                    vote_slots.insert(vote.slot);
                 }
             }
             if let Some(root) = vote_state.root_slot {
@@ -298,7 +299,7 @@ impl Tower {
                     slot: root,
                 };
                 bank_weight += vote.lockout() as u128 * voted_stake as u128;
-                Self::populate_ancestor_voted_stakes(&mut voted_stakes, &vote, ancestors);
+                vote_slots.insert(vote.slot);
             }
 
             // The last vote in the vote stack is a simulated vote on bank_slot, which
@@ -326,6 +327,9 @@ impl Tower {
             total_stake += voted_stake;
         }
 
+        // TODO: populate_ancestor_voted_stakes only adds zeros. Comment why
+        // that is necessary (if so).
+        Self::populate_ancestor_voted_stakes(&mut voted_stakes, vote_slots, ancestors);
         ComputedBankState {
             voted_stakes,
             total_stake,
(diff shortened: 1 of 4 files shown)


behzadnouri:
Left chunk is this commit, vs current code on the right:
```replay-loop-timing-stats.compute_bank_stats_elapsed```
replay-loop-timing-stats.compute_bank_stats_elapsed
```bank-timestamp.get_timestamp_estimate_us```
bank-timestamp.get_timestamp_estimate_us

File: core/src/consensus.rs [hunk 1]
```diff
@@ -326,6 +327,9 @@ impl Tower {
             total_stake += voted_stake;
         }
 
+        // TODO: populate_ancestor_voted_stakes only adds zeros. Comment why
+        // that is necessary (if so).
+        Self::populate_ancestor_voted_stakes(&mut voted_stakes, vote_slots, ancestors);
```
carllin, line 332: This is a good catch, this function actually used to update an outdated `lockouts` field which was used to compute the `bank_weight`, which used to be the basis of our fork choice rule. That functionality, however, became outdated by our new fork choice rule and was removed here: https://github.com/solana-labs/solana/pull/10719/files#diff-0b654209e209924bfc70e8dd4c84d0979e36b6ca76e85a5899ba4572521a8170L551-L552

But I am not 100% sure this is safe to remove. Let's take at the two places `voted_stakes` is being used in `replay_stage`:

1) `confirm_forks`: https://github.com/solana-labs/solana/blob/master/core/src/replay_stage.rs#L363

This calls down to `tower.is_slot_confirmed`, which checks:

```
voted_stakes
            .get(&slot)
            .map(|stake| (*stake as f64 / total_stake as f64) > self.threshold_size)
            .unwrap_or(false)
```

So inserting 0 there for non-entries should not change behavior, given ` self.threshold_size > 0`


2) `check_vote_stake_threshold`: https://github.com/solana-labs/solana/blob/master/core/src/replay_stage.rs#L1473. This one is trickier, mostly due to this line: https://github.com/solana-labs/solana/blame/73586c1aadcb9ea6337fc4bdff6b5d75dc030417/core/src/consensus.rs#L511. So if an entry is missing, it will return false where before it may have returned true. If we remove that:

`if let Some(fork_stake) = stake_lockouts.get(&vote.slot) {`

with 

`let fork_stake = stake_lockouts.get(&vote.slot).unwrap_or(0)`

then, we need to consider any potential new cases in today's logic where we could return a different result; in this case, I think it would only be if `vote.slot`:

a) did not exist in `stake_lockouts`
b) satisfied this condition: https://github.com/solana-labs/solana/blame/73586c1aadcb9ea6337fc4bdff6b5d75dc030417/core/src/consensus.rs#L522

in which case we would return true whereas we would have returned false before. However, for b) to be true, there needs to be some `old_vote` in the *local tower* where `old_vote.slot == vote.slot`, which by a) means `old_vote.slot` is also not in
`stake_lockouts`, which means none of the descendants of old_vote.slot` were processed by any calls to `populate_ancestor_voted_stakes()` in `collect_vote_lockouts()`.  

Because this block: https://github.com/solana-labs/solana/blame/master/core/src/consensus.rs#L277-L281, which simulates a vote of `bank_slot`, we know always adds all the ancestors (up to root) of this slot: https://github.com/solana-labs/solana/blame/master/core/src/consensus.rs#L734 are present in `voted_stakes`, so the only way `old_vote.slot` is not in
`stake_lockouts` is if `old_vote.slot` and the slot we're trying to vote on are on different forks, i.e.

```
                     root
        /                             \
old_vote                     slot we're checking
```

Hence we can replicate this check without the `vote_stakes` if we pass in `ancestors` here and change that check from 

```
if let Some(fork_stake) = voted_stakes.get(&vote.slot) {
```

to 

```
if ancestors.get(slot).unwrap().contains(&vote.slot)
```

Then it should be equivalent.


Going one step further you'll notice that indeed all the logic related to `bank_weight` here can be removed, including all the `bank_weight +=` blocks as the `bank_weight` field is never used anywhere anymore

carllin (approved):

core/src/consensus.rs
behzadnouri: thank you for adding. this is very helpful

PR #14244, opened 2020-12-22
ClusterInfo: Update entrypoint contact info even when shred version adoption is not requested

mvines:
A different spin on https://github.com/solana-labs/solana/pull/14243

ClusterInfo's entrypoint contact info is now updated in all cases, rather than only when shred version adoption is requested.  This brings uniformity and reduces the number of runtime combinations.   I'm still not 100% confident that updating the entrypoint contact info is required at all, but at least now it's clear that this behavior is not actually related to shred version adoption.

### DIFF
--- a/core/src/cluster_info.rs
+++ b/core/src/cluster_info.rs
@@ -1731,19 +1731,21 @@ impl ClusterInfo {
         Ok(())
     }
 
-    fn handle_adopt_shred_version(self: &Arc<Self>, adopt_shred_version: &mut bool) {
-        // Adopt the entrypoint's `shred_version` if ours is unset
-        if *adopt_shred_version {
-            // If gossip was given an entrypoint, look up the ContactInfo by the given
-            // entrypoint gossip adddress
-            let gossip_addr = self.entrypoint.read().unwrap().as_ref().map(|e| e.gossip);
+    fn process_entrypoint(self: &Arc<Self>, entrypoint_processed: &mut bool) {
+        if *entrypoint_processed {
+            return;
+        }
 
-            if let Some(gossip_addr) = gossip_addr {
-                // If a pull from the entrypoint was successful, it should exist in the crds table
-                let entrypoint = self.lookup_contact_info_by_gossip_addr(&gossip_addr);
-                if let Some(entrypoint) = entrypoint {
+        let gossip_addr = self.entrypoint.read().unwrap().as_ref().map(|e| e.gossip);
+        if let Some(gossip_addr) = gossip_addr {
+            // If a pull from the entrypoint was successful it should exist in the CRDS table
+            let entrypoint = self.lookup_contact_info_by_gossip_addr(&gossip_addr);
+
+            if let Some(entrypoint) = entrypoint {
+                // Adopt the entrypoint's `shred_version` if ours is unset
+                if self.my_shred_version() == 0 {
                     if entrypoint.shred_version == 0 {
-                        info!("Unable to adopt entrypoint's shred version");
+                        warn!("Unable to adopt entrypoint shred version of 0");
                     } else {
                         info!(
                             "Setting shred version to {:?} from entrypoint {:?}",
@@ -1756,11 +1758,18 @@ impl ClusterInfo {
                             .unwrap()
                             .set_shred_version(entrypoint.shred_version);
                         self.insert_self();
-                        *self.entrypoint.write().unwrap() = Some(entrypoint);
-                        *adopt_shred_version = false;
+                        *entrypoint_processed = true;
                     }
+                } else {
+                    *entrypoint_processed = true;
                 }
+
+                // Update the entrypoint's id so future entrypoint pulls correctly reference it
+                *self.entrypoint.write().unwrap() = Some(entrypoint);
             }
+        } else {
+            // No entrypoint specified.  Nothing more to process
+            *entrypoint_processed = true;
         }
     }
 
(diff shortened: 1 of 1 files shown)


core/src/cluster_info.rs
```diff
@@ -1629,13 +1629,19 @@ impl ClusterInfo {
...
+        }
+
         self.append_entrypoint_to_pulls(thread_pool, &mut pulls);
         self.stats
             .new_pull_requests_count
             .add_relaxed(pulls.len() as u64);
         // There are at most 2 unique peers here: The randomly
         // selected pull peer, and possibly also the entrypoint.
-        let peers: Vec<Pubkey> = pulls.iter().map(|(peer, _, _, _)| *peer).dedup().collect();
+        let peers: Vec<Pubkey> = pulls.iter().map(|(peer, _, _, _)| *peer).collect();
```
mvines, line 1644: The above `debug_assert!()` helps point out how this `.dedup()` is not required.  `append_entrypoint_to_pulls()` only adds the entrypoint to `pulls` if it's not in the current CRDS table, which means that the entrypoint will never be in `pulls` twice.  So there's nothing to dedup

core/src/cluster_info.rs
```diff
@@ -1731,19 +1737,21 @@ impl ClusterInfo {
         Ok(())
     }
 
-    fn handle_adopt_shred_version(self: &Arc<Self>, adopt_shred_version: &mut bool) {
-        // Adopt the entrypoint's `shred_version` if ours is unset
-        if *adopt_shred_version {
-            // If gossip was given an entrypoint, look up the ContactInfo by the given
-            // entrypoint gossip adddress
-            let gossip_addr = self.entrypoint.read().unwrap().as_ref().map(|e| e.gossip);
+    fn process_entrypoint(self: &Arc<Self>, entrypoint_processed: &mut bool) {
```
mvines, line 1740: handle_adopt_shred_version -> process_entrypoint, because this function now updates the entrypoint's id even when not shred adopting.

This function will get slightly more involved when multiple entrypoints are supported, when it'll adopt the shred of the first entrypoint found and then update the id for all entrypoints as they are discovered.

core/src/cluster_info.rs
```diff
@@ -1756,11 +1764,16 @@ impl ClusterInfo {
...
                             .set_shred_version(entrypoint.shred_version);
                         self.insert_self();
-                        *self.entrypoint.write().unwrap() = Some(entrypoint);
-                        *adopt_shred_version = false;
                     }
                 }
+
+                // Update the entrypoint's id so future entrypoint pulls correctly reference it
+                *self.entrypoint.write().unwrap() = Some(entrypoint);
+                *entrypoint_processed = true;
```
behzadnouri, line 1772: Wouldn't this prevent `shred_version` to ever get updated, if on the first pass `entrypoint.shred_version == 0`?

mvines: I couldn't think of many good reasons why entrypoint.shred_version could ever be 0 on the first pass but not the second pass.   Only reason I can think of is if the entrypoint itself is restarted.

mvines: I guess that's reason enough 👍🏼

core/src/cluster_info.rs
mvines: Got it, thank you!

core/src/cluster_info.rs
```diff
@@ -1731,19 +1731,21 @@ impl ClusterInfo {
...
+        if let Some(gossip_addr) = gossip_addr {
+            // If a pull from the entrypoint was successful it should exist in the CRDS table
+            let entrypoint = self.lookup_contact_info_by_gossip_addr(&gossip_addr);
+
+            if let Some(entrypoint) = entrypoint {
+                // Adopt the entrypoint's `shred_version` if ours is unset
+                if self.my_shred_version() == 0 {
                     if entrypoint.shred_version == 0 {
-                        info!("Unable to adopt entrypoint's shred version");
+                        warn!("Unable to adopt entrypoint shred version of 0");
```
mvines, line 1748: entrypoint_processed now remains `false` if the entrypoint shred version is 0 somehow (like a poorly configured entrypoint) and this node wants to adopt its shred version, so we'll continue to try to adopt

core/src/cluster_info.rs
```diff
@@ -4225,12 +4234,69 @@ mod tests {
...
+
+        // Adopt the entrypoint's gossiped contact info and verify
+        let mut entrypoint_processed = false;
+        ClusterInfo::process_entrypoint(&cluster_info, &mut entrypoint_processed);
+        assert_eq!(
+            cluster_info.entrypoint.read().unwrap().as_ref().unwrap(),
+            &gossiped_entrypoint_info
+        );
+        assert!(!entrypoint_processed); // <--- entrypoint processing incomplete because shred adoption still pending
+        assert_eq!(cluster_info.my_shred_version(), 0); // <-- shred version still 0
```
mvines, line 4269: Added a test for the adoption retry case

mvines:
@behzadnouri - I think this PR is ready for another review!

core/src/cluster_info.rs
```diff
@@ -1756,11 +1758,18 @@ impl ClusterInfo {
...
-                        *self.entrypoint.write().unwrap() = Some(entrypoint);
-                        *adopt_shred_version = false;
+                        *entrypoint_processed = true;
                     }
+                } else {
+                    *entrypoint_processed = true;
                 }
+
+                // Update the entrypoint's id so future entrypoint pulls correctly reference it
+                *self.entrypoint.write().unwrap() = Some(entrypoint);
```
behzadnouri, line 1768: Unrelated to this pull-request (as it seems pre-existing logic), but seems like we are expecting the entry-point to never update its contact-info (and in particular its gossip address).
Is that a reasonable assumption? or, should we make an effort to sync `.entrypoint` if we get an updated contact-info?

behzadnouri (approved):

core/src/cluster_info.rs
mvines: I think it's a reasonable assumption since the only time that changes is during a network restart (assuming the entrypoint is configured correctly)

mvines:
I'm going to roll this PR into https://github.com/solana-labs/solana/pull/14256

PR #14256, opened 2020-12-22
validator: Multiple --entrypoint support

### LINKED ISSUE #5438: Support multiple validator --entrypoint arguments?
dyn.com had a temporary outage and nodes couldn't get on the cluster because the DNS entrypoint name wouldn't resolve.

We could mitigate by supporting multiple `--entrypoint` arguments to `solana-validator`, that the validator tries to connect to in sequence.  Eg:
```bash
$ solana-validator --entrypoint tds.solana.com:8899 --entrypoint 34.83.86.137:8899 ...
```

mvines:
Fixes #5438

### DIFF
--- a/core/src/cluster_info.rs
+++ b/core/src/cluster_info.rs
@@ -1731,19 +1731,21 @@ impl ClusterInfo {
         Ok(())
     }
 
-    fn handle_adopt_shred_version(&self, adopt_shred_version: &mut bool) {
-        // Adopt the entrypoint's `shred_version` if ours is unset
-        if *adopt_shred_version {
-            // If gossip was given an entrypoint, look up the ContactInfo by the given
-            // entrypoint gossip adddress
-            let gossip_addr = self.entrypoint.read().unwrap().as_ref().map(|e| e.gossip);
+    fn process_entrypoint(&self, entrypoint_processed: &mut bool) {
+        if *entrypoint_processed {
+            return;
+        }
 
-            if let Some(gossip_addr) = gossip_addr {
-                // If a pull from the entrypoint was successful, it should exist in the crds table
-                let entrypoint = self.lookup_contact_info_by_gossip_addr(&gossip_addr);
-                if let Some(entrypoint) = entrypoint {
+        let gossip_addr = self.entrypoint.read().unwrap().as_ref().map(|e| e.gossip);
+        if let Some(gossip_addr) = gossip_addr {
+            // If a pull from the entrypoint was successful it should exist in the CRDS table
+            let entrypoint = self.lookup_contact_info_by_gossip_addr(&gossip_addr);
+
+            if let Some(entrypoint) = entrypoint {
+                // Adopt the entrypoint's `shred_version` if ours is unset
+                if self.my_shred_version() == 0 {
                     if entrypoint.shred_version == 0 {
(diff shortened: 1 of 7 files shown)


mvines:
I've added the boot nodes as entrypoints with DNS entries, so one can do this now:
```
solana-validator ...
    --entrypoint entrypoint.mainnet-beta.solana.com:8001 \
    --entrypoint entrypoint2.mainnet-beta.solana.com:8001 \
    --entrypoint entrypoint3.mainnet-beta.solana.com:8001 \
    --entrypoint entrypoint4.mainnet-beta.solana.com:8001 \
    --entrypoint entrypoint5.mainnet-beta.solana.com:8001 \
```

validator/src/main.rs
```diff
@@ -3,7 +3,7 @@ use clap::{
     ArgMatches,
```
mvines, line 3: `validator/src/main.rs` will randomly run through all the provided entrypoints and only hard fail if they are all unreachable

core/src/cluster_info.rs
```diff
@@ -1499,52 +1504,49 @@ impl ClusterInfo {
...
     ) {
-        let pull_from_entrypoint = {
-            let mut w_entrypoint = self.entrypoint.write().unwrap();
-            if let Some(ref mut entrypoint) = &mut *w_entrypoint {
+        let entrypoint_id_and_gossip = {
+            let mut entrypoints = self.entrypoints.write().unwrap();
+            if entrypoints.is_empty() {
+                None
+            } else {
+                let i = thread_rng().gen_range(0, entrypoints.len());
```
mvines, line 1512: This is the main logic change in `core/src/cluster_info.rs`.   Rather than **the** entrypoint, every time we run though `append_entrypoint_to_pulls()` randomly pick **one** of the entrypoints and check if they need a pull using the same logic as before

core/src/cluster_info.rs
```diff
@@ -1729,37 +1731,51 @@ impl ClusterInfo {
         Ok(())
     }
 
-    fn handle_adopt_shred_version(self: &Arc<Self>, adopt_shred_version: &mut bool) {
-        // Adopt the entrypoint's `shred_version` if ours is unset
-        if *adopt_shred_version {
-            // If gossip was given an entrypoint, look up the ContactInfo by the given
-            // entrypoint gossip adddress
-            let gossip_addr = self.entrypoint.read().unwrap().as_ref().map(|e| e.gossip);
+    fn process_entrypoints(self: &Arc<Self>, entrypoints_processed: &mut bool) {
```
mvines, line 1734: Changes here:
* adopt shred version from the first non-zero entrypoint shred version
* update the entrypoint.id for all entrypoints.  Don't set `entrypoints_processed = true` until they've all been updated

docs/src/clusters.md
```diff
@@ -147,7 +147,11 @@ $ solana-validator \
     --rpc-port 8899 \
     --private-rpc \
     --dynamic-port-range 8000-8010 \
-    --entrypoint mainnet-beta.solana.com:8001 \
```
mvines, line 150: `mainnet-beta.solana.com` still exists, I'm just de-documenting it for consistency

behzadnouri (approved):

PR #14259, opened 2020-12-22
Revert "Add shrink paths (#14208)"

CriesofCarrots:
This reverts commit 9466ad3c1f11fb90df6015d6c910f0e89747a553.

#### Problem
CI coverage jobs are timing out regularly. It appears that after #14208  , `AccountsDB::test_shrink_and_clean` is taking a really long time, or sometimes not completing at all. This loop is not exiting: https://github.com/solana-labs/solana/blob/7042f11791371f9c6d05c76f1b3bfa89dafaa295/runtime/src/accounts_db.rs#L1023

### DIFF
--- a/core/src/validator.rs
+++ b/core/src/validator.rs
@@ -80,7 +80,6 @@ pub struct ValidatorConfig {
     pub expected_shred_version: Option<u16>,
     pub voting_disabled: bool,
     pub account_paths: Vec<PathBuf>,
-    pub account_shrink_paths: Option<Vec<PathBuf>>,
     pub rpc_config: JsonRpcConfig,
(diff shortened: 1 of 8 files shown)


mvines:
Ack, we have nodes using `--account-shrink-path` now.  Is the plan to bring it back before the next v1.4 release or is it going away permanently?

CriesofCarrots:
> Ack, we have nodes using `--account-shrink-path` now. Is the plan to bring it back before the next v1.4 release or is it going away permanently?

Yes, definitely. Just reverting until @sakridge can fix. That said, I'll try ignoring the problematic test instead. Hopefully the issue arising in the test (and only on CI machines, afaict) is not one that will affect running nodes.

PR #14266, opened 2020-12-23
Allow setting max poor block producer percentage

danpaul000:
#### Problem
Recent mainnet instability is often causing more than 20% of block producers to fail the threshold, halting all bonus stake from moving for good or bad block producers.

#### Summary of Changes
Allow setting the block production cutoff on the CLI to remove larger amount of stake if needed.  This is mostly a band-aid to keep the current stake bot running until @t-nelson finishes the work on the new delegation program, but will be useful right away for the currently deployed and often failing stake bot.

### DIFF
--- a/stake-o-matic/src/main.rs
+++ b/stake-o-matic/src/main.rs
@@ -64,7 +64,7 @@ struct Config {
     delinquent_grace_slot_distance: u64,
 
     /// Don't ever unstake more than this percentage of the cluster at one time
-    max_poor_block_productor_percentage: usize,
+    max_poor_block_producer_percentage: usize,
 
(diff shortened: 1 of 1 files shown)


stake-o-matic/src/main.rs
```diff
@@ -64,7 +64,7 @@ struct Config {
     delinquent_grace_slot_distance: u64,
 
     /// Don't ever unstake more than this percentage of the cluster at one time
-    max_poor_block_productor_percentage: usize,
+    max_poor_block_producer_percentage: usize,
```
mvines, line 67: 🤦🏼

mvines (approved):
lgtm!

PR #14272, opened 2020-12-23
indexes votes in crds table

behzadnouri:
#### Problem
ClusterInfo::get_votes does a full crds table scan to filter out votes:
https://github.com/solana-labs/solana/blob/29221904b/core/src/cluster_info.rs#L1013-L1034
This will become more inefficient as we are adding more crds values in gossip (duplicate shreds, root, ...).

#### Summary of Changes
Added an index for votes in crds table


behzadnouri:
Left chunk is this commit vs current code on the right.
`cluster_info_stats.get_votes`:
votes_index cluster_info_stats get_votes

carllin (approved):

PR #14276, opened 2020-12-23
Don't use caller passed executable account

jackcmay:
#### Problem

CPI takes the executable account to run from the caller which requires the runtime to validate that account.  

#### Summary of Changes

The runtime already has the executable account so use it.  This change limits the attack surface of CPI calls.

Fixes #

### DIFF
--- a/programs/bpf/c/src/invoke/invoke.c
+++ b/programs/bpf/c/src/invoke/invoke.c
@@ -291,7 +291,8 @@ extern uint64_t entrypoint(const uint8_t *input) {
   case TEST_EMPTY_ACCOUNTS_SLICE: {
     sol_log("Empty accounts slice");
 
-    SolAccountMeta arguments[] = {};
+    SolAccountMeta arguments[] = {
+        {accounts[INVOKED_ARGUMENT_INDEX].key, false, false}};
     uint8_t data[] = {};
     const SolInstruction instruction = {accounts[INVOKED_PROGRAM_INDEX].key,
                                         arguments, SOL_ARRAY_SIZE(arguments),
--- a/programs/bpf/rust/invoke/src/lib.rs
+++ b/programs/bpf/rust/invoke/src/lib.rs
@@ -385,7 +385,11 @@ fn process_instruction(
         }
         TEST_EMPTY_ACCOUNTS_SLICE => {
             msg!("Empty accounts slice");
-            let instruction = create_instruction(*accounts[INVOKED_PROGRAM_INDEX].key, &[], vec![]);
+            let instruction = create_instruction(
+                *accounts[INVOKED_PROGRAM_INDEX].key,
+                &[(accounts[INVOKED_ARGUMENT_INDEX].key, false, false)],
+                vec![],
+            );
             invoke(&instruction, &[])?;
         }
         TEST_CAP_SEEDS => {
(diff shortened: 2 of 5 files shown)


jstarry:
I think there's a undesirable side effect of this change..

Say you have Program A calling Program B. After this change, Program A has no way of knowing which programs Program B will call because Program B is free to call any programs that were passed to the root instruction.

jstarry:
Another thing.. if a program wants to call itself, does its program account need to be passed as an instruction account input? Seems like the intent of this PR is to answer "no" since we don't use caller passed executable accounts anyways. But it looks like it's still required because `invoke_context.get_account(..)` will not find the program account in `pre_accounts` or `account_deps`

jackcmay:
For the first point, yes, program b is free to call any program in the instruction, having the executables being passed via the program was not originally intentional so these changes fix that.  Not knowing what a program that you call does is very risky no matter what.  If we think we should be explicit about what programs a CPI'd program is free to call then we could add checks for that

For the second point from a tx/instruction perspective, yes, the program's executable account still needs to be part of the instruction.  The tx/instruction definition doesn't handle CPI very well and we eventually need new definitions that do.

jstarry:
> Not knowing what a program that you call does is very risky no matter what

Yes, this is the real issue at hand. Programs should only ever call hard coded / allowed list of programs 👍 

> Yes, the program's executable account still needs to be part of the instruction. The tx/instruction definition doesn't handle CPI very well and we eventually need new definitions that do.

I was actually thinking it would be better if the executable account doesn't need to be part of the instruction accounts in order to be self-called. Couldn't we stick the called program account inside `account_deps`?

jstarry:
Also, should this change be feature gated?

jackcmay:
Feature gated because of folks who might have attempted to modify an executable account?  (they can still pass whatever they want, executables will just be ignored)

By self-called you mean that since the executable corresponding to the program_id in the instruction is already loaded we should just use that, but that it no available because it is not in the instruction accounts or dep accounts?

jstarry:
> Feature gated because of folks who might have attempted to modify an executable account?

Because runtime behavior will change. A CPI will no longer need to be passed the executable account when invoked and nodes on an older version would fail the transaction if that was the case.

> By self-called you mean that since the executable corresponding to the program_id in the instruction is already loaded we should just use that, but that it not available because it is not in the instruction accounts or dep accounts?

Yes, exactly

jackcmay:
Yes on the gating, unlikely, but a good measure

I think the self calling question it is an interesting one,  on one hand not requiring the self-called program account to be passed would be inconsistent with the current practice that all programs that will be CPI'd be specified of the instruction's accounts.  Aka, if a program wants to call itself it needs to include itself in the instruction.  This would be akin to a new tx format that specifies as part of the instruction what programs will be CPI'd.

jstarry:
> Yes on the gating, unlikely, but a good measure

Yeah, I just was thinking that it would be nice to have some peace of mind there. I will probably be updating the spl lending program to no longer pass program info in the invoke call and it would be nice to know when I could deploy that change to mainnet.

> I think the self calling question it is an interesting one, on one hand not requiring the self-called program account to be passed would be inconsistent with the current practice that all programs that will be CPI'd be specified of the instruction's accounts.

Good point, I agree that a new tx format would help here. We could potentially expose the list of executable accounts via a sysvar too if we don't want to change the program entrypoint

PR #14285, opened 2020-12-24
uses ValidatorExit to stop when observing duplicate instances

behzadnouri:
#### Problem
`ClusterInfo::listen` does `std::process::exit` when it identifies duplicate instances:
https://github.com/solana-labs/solana/blob/87eb924d2/core/src/cluster_info.rs#L2940

#### Summary of Changes
* Passed through `ValidatorExit` instead to stop the validator.
* Added test coverage.

### DIFF
--- a/core/src/cluster_info.rs
+++ b/core/src/cluster_info.rs
@@ -25,6 +25,7 @@ use crate::{
     epoch_slots::EpochSlots,
     ping_pong::{self, PingCache, Pong},
     result::{Error, Result},
+    validator::ValidatorExit,
     weighted_shuffle::weighted_shuffle,
 };
 
@@ -2899,6 +2900,7 @@ impl ClusterInfo {
         requests_receiver: PacketReceiver,
         response_sender: PacketSender,
         exit: &Arc<AtomicBool>,
+        validator_exit: Arc<RwLock<Option<ValidatorExit>>>,
     ) -> JoinHandle<()> {
         let exit = exit.clone();
         let recycler = PacketsRecycler::default();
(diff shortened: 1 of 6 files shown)


behzadnouri:
@mvines @sakridge  
This code is to avoid `std::process::exit` when identifying duplicate instances and also add some test coverage.

* Using `ValidatorExit` it looks like it takes ~2-3min since observing the duplicate instance until the validator exits. Is that lag too long?
* From my testing, most of the threads stop after like 10 seconds. But some linger on: 
```
rocksdb:high0, ..., rocksdb:high2
rocksdb:low0,  ..., rocksdb:low11
sigverify_0,   ..., sigverify_5
sigverify_shred
solana-gossip
solana-snapshot
solana-validato
tokio-runtime-w
```
Linux thread names can only be 15 characters long, but we are using longer names, so some of the names are truncated. (Also _probably_ `solana-gossip` is [the thread emitting metrics](https://github.com/solana-labs/solana/blob/87eb924d2/metrics/src/metrics.rs#L148) and not [the actual gossip thread](https://github.com/solana-labs/solana/blob/87eb924d2/core/src/cluster_info.rs#L1822)).

mvines:
> Using ValidatorExit it looks like it takes ~2-3 minutes since observing the duplicate instance until the validator exits. Is that lag too long?

That feels a little long.  We could use ValidatorExit for the additional test coverage but keep the process::exit(1) when running as `solana-validator` to force an exit as quickly as possible

behzadnouri:
> That feels a little long. We could use ValidatorExit for the additional test coverage but keep the process::exit(1) when running as `solana-validator` to force an exit as quickly as possible

It seems like I cannot use the
```
    #[cfg(not(test))]
    std::process::exit(1);

    #[cfg(test)]
    // use validator-exit instead ...
```
trick here, because it is always `#[cfg(not(test))]` if it is an integration test or the test is in a different crate (e.g. `local_cluster` vs `core` in here).

PR #14289, opened 2020-12-26
Persist gossip contact info

mvines:
Cold starts for gossip can take a while, and node IP addresses really don't change often.  Save gossip contact info for faster validator restarts

### DIFF
--- a/core/src/cluster_info.rs
+++ b/core/src/cluster_info.rs
@@ -69,8 +69,11 @@ use std::{
     cmp::min,
     collections::{hash_map::Entry, HashMap, HashSet, VecDeque},
     fmt::{self, Debug},
+    fs::{self, File},
+    io::BufReader,
     net::{IpAddr, Ipv4Addr, SocketAddr, TcpListener, UdpSocket},
     ops::{Deref, DerefMut},
+    path::{Path, PathBuf},
     sync::atomic::{AtomicBool, AtomicU64, Ordering},
     sync::{Arc, RwLock, RwLockReadGuard, RwLockWriteGuard},
     thread::{sleep, Builder, JoinHandle},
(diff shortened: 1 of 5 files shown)


core/src/cluster_info.rs
```diff
@@ -653,6 +663,117 @@ impl ClusterInfo {
...
+                .unwrap()
+                .iter()
+                .map(|contact_info| contact_info.gossip)
+                .collect::<HashSet<_>>();
+
+            gossip
+                .crds
+                .get_nodes()
+                .filter_map(|v| {
+                    if let Some(contact_info) = v.value.contact_info() {
```
behzadnouri, line 681: maybe just 
```
let contact_info = v.value.contact_info().unwrap()
```
here.

core/src/cluster_info.rs
```diff
@@ -653,6 +663,117 @@ impl ClusterInfo {
...
+        };
+
+        info!(
+            "Loaded contact info for {} nodes from {}",
+            nodes.len(),
+            filename.display()
+        );
+        let now = timestamp();
+        for node in nodes {
+            if let Err(err) = self.gossip.write().unwrap().crds.insert(node, now) {
```
behzadnouri, line 771: maybe obtain this lock only once for the entire loop

behzadnouri (dismissed):

core/src/cluster_info.rs
mvines: Ah, good idea. Done

core/src/cluster_info.rs
mvines: Ok.  Generally unwraps like this make me a little nervous, we've had crashes due to bad unwraps before.  But I guess in this case `get_nodes_contact_info()` already calls `get_nodes()` and explicitly panic on `None` so an unwrap here isn't any worse

behzadnouri: Agree about avoiding unwrap in general, but in this case it is an invariant that *has to be always the case*.

PR #14305, opened 2020-12-28
Remove Testnet-specific old code

ryoqun:
#### Problem

This isn't needed at all now that feature (`4kpdyrcj5jS47CZb2oJGfVxjYbsMm2Kx97gFyZrxxwXz no overflow rent distribution`) is activated across all live clusters.

#### Summary of Changes

Remove this. And catch any capitalization mismatch sooner if ever we introduce such a bug sooner on the testnet.

follow-up: #12112 (this is the basis for the chosen reviewer)


t-nelson (approved):
:wastebasket:

PR #14331, opened 2020-12-29
Re-enable shrink test

### LINKED ISSUE #14301: Unignore test_shrink_and_clean
#### Problem
`AccountsDB::test_shrink_and_clean` is currently ignored (#14261 ). After #14208 , the test began taking a really long time, or sometimes not completing at all, timing out CI. This loop is not exiting: https://github.com/solana-labs/solana/blob/7042f11791371f9c6d05c76f1b3bfa89dafaa295/runtime/src/accounts_db.rs#L1023
cc @sakridge 

#### Proposed Solution
Debug, fix, unignore

sakridge:
#### Problem

Accounts shrink integration test was disabled because it runs for too long in coverage.

#### Summary of Changes

Move longer-running accounts tests to integration side.

Fixes #14301

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -590,7 +590,6 @@ impl AccountsDB {
         self.file_size
     }
 
-    #[cfg(test)]
     pub fn new_single() -> Self {
         AccountsDB {
             min_num_stores: 0,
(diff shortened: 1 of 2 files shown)


CriesofCarrots (approved):
lgtm
Doesn't seem to have lengthened stable tests much; I wonder if test concurrency was part of the issue.

sakridge:
> lgtm
> Doesn't seem to have lengthened stable tests much; I wonder if test concurrency was part of the issue.

I also lowered the amount of data in the test from 1000 roots to 100, but it's not really a unit-test type of test.

PR #14346, opened 2020-12-30
Speed up UDP reachable port checks

mvines:
The `solana-validator` port check takes ~15 seconds, reduce to 1-2 seconds.   The delay was due to unnecessarily waiting on UDP receive timeouts

### DIFF
--- a/net-utils/src/lib.rs
+++ b/net-utils/src/lib.rs
@@ -1,13 +1,17 @@
 //! The `net_utils` module assists with networking
-use log::*;
-use rand::{thread_rng, Rng};
-use socket2::{Domain, SockAddr, Socket, Type};
-use std::collections::{BTreeMap, BTreeSet};
-use std::io::{self, Read, Write};
-use std::net::{IpAddr, SocketAddr, TcpListener, TcpStream, ToSocketAddrs, UdpSocket};
-use std::sync::mpsc::channel;
-use std::time::Duration;
-use url::Url;
+use {
+    log::*,
+    rand::{thread_rng, Rng},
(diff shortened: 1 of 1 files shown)


net-utils/src/lib.rs
```diff
@@ -204,21 +208,38 @@ fn do_verify_reachable_ports(
...
+
                         let original_read_timeout = udp_socket.read_timeout().unwrap();
-                        udp_socket.set_read_timeout(Some(timeout)).unwrap();
-                        let recv_result = udp_socket.recv(&mut buf);
-                        debug!(
-                            "Waited for incoming datagram on udp/{}: {:?}",
-                            port, recv_result
-                        );
+                        udp_socket
+                            .set_read_timeout(Some(Duration::from_millis(250)))
```
mvines, line 222: Instead of the full `timeout`, set a much shorter receive timeout for each socket and check multiple times a second for it the port is now reachable (reminder: there are multiple open sockets for ports and the response from the IP echo server can be received on any of them)

net-utils/src/lib.rs
```diff
@@ -204,21 +208,38 @@ fn do_verify_reachable_ports(
...
+                            }
+
+                            let recv_result = udp_socket.recv(&mut [0; 1]);
+                            debug!(
+                                "Waited for incoming datagram on udp/{}: {:?}",
+                                port, recv_result
+                            );
+
+                            if recv_result.is_ok() {
+                                reachable_ports.write().unwrap().insert(port);
```
mvines, line 238: The socket that receives the response records the port as reachable, the threads for the other sockets for the same port will exit in at most 250ms.

PR #14349, opened 2020-12-30
adds root crds value to gossip

behzadnouri:
#### Problem
Following [this plan](https://discord.com/channels/428295358100013066/738253477331075162/784888854628139008)

#### Summary of Changes
Added a new crds value to broadcast current root.

### DIFF
--- a/core/src/cluster_info.rs
+++ b/core/src/cluster_info.rs
@@ -429,7 +429,7 @@ pub fn make_accounts_hashes_message(
 type Ping = ping_pong::Ping<[u8; GOSSIP_PING_TOKEN_SIZE]>;
 
 // TODO These messages should go through the gpu pipeline for spam filtering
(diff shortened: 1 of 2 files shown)


carllin:
@behzadnouri, let's actually hold off on this change until we get the rest of the logic implemented, and are able to:

1) Detect duplicates proofs/vote hashes in gossip
2) See entire network read the proofs, dump any bad versions of the slot and recover.

This logic to broadcast the root in gossip is part of the attempt to remove unnecessary proofs ASAP, but I think there are some kinks to iron out in that step (primarily if we remove the proofs in gossip, how will the validators catching up/in the minority know which is the correct version of the slot?), so for v1 let's assume the proofs will be available for as long as there aren't other duplicates overriding those slots in gossip (so less than the 100 proof limit).

PR #14362, opened 2020-12-30
adds an optional encoding for serializing duplicate shreds

behzadnouri:
#### Problem
The payload for `DuplicateShred` is larger than what can fit in one packet.
Currently this is split up into multiple chunks, each with its own added
overhead, Plus the overhead when these are wrapped in gossip messages: 
https://github.com/solana-labs/solana/blob/cb8ba739a/core/src/duplicate_shred.rs#L32-L36
Compressing the payload may reduce the number of chunks needed to send one
complete duplicate slot proof over gossip.

#### Summary of Changes
* Added an optional encoding for duplicate shred payloads.
* Since decoding is done in a separate thread than gossip, this should not slow
  down gossip.
* The test shows ~35% reduction in payload size of duplicate shreds.

### DIFF
--- a/core/src/encode.rs
+++ b/core/src/encode.rs
@@ -1,0 +1,171 @@
+use bincode::Options as _;
+use flate2::{
+    read::{DeflateDecoder, GzDecoder, ZlibDecoder},
+    write::{DeflateEncoder, GzEncoder, ZlibEncoder},
+    Compression,
+};
+use serde::{de::DeserializeOwned, Serialize};
+use std::io::{Read, Write};
+use thiserror::Error;
+
+#[derive(Deserialize, Serialize)]
+enum Encoded {
+    Bincode(Vec<u8>),
+    Deflate(Vec<u8>),
+    Gzip(Vec<u8>),
+    Zlib(Vec<u8>),
+    Zstd(Vec<u8>),
+}
+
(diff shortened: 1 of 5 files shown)


core/src/encode.rs
```diff
@@ -0,0 +1,171 @@
...
+}
+
+pub fn decode<T: DeserializeOwned>(bytes: &[u8]) -> Result<T, Error> {
+    let encoded = bincode::options().deserialize_from(bytes)?;
+    let decoder: Box<dyn Read> = match &encoded {
+        Encoded::Bincode(bytes) => Box::new(&bytes[..]),
+        Encoded::Deflate(bytes) => Box::new(DeflateDecoder::new(&bytes[..])),
+        Encoded::Gzip(bytes) => Box::new(GzDecoder::new(&bytes[..])),
+        Encoded::Zlib(bytes) => Box::new(ZlibDecoder::new(&bytes[..])),
+        Encoded::Zstd(bytes) => Box::new(zstd::stream::read::Decoder::new(&bytes[..])?),
```
sakridge, line 93: Are these decoders hardened against zip-bomb-like behavior?

behzadnouri: I raised that question to the upstream package; and meanwhile, added a `limit` here for number of bytes decoded.

PR #14429, opened 2021-01-05
Lower recycle store count

### LINKED ISSUE #14410: v1.5.1 validator can't keep up with mainnet
There's a regression on v1.5.1 that is preventing a validator running this release from keeping up with mainnet.  The general pattern is:
1. The validator catches up normally
2. Leave the validator running and after an hour or less it'll start falling behind.   Leave it overnight and it's even further behind
The same machine setup works fine with a v1.4.20 validator install.

sakridge:
#### Problem

Too many cached stores can cause swap usage which is detrimental to account store times.

#### Summary of Changes

Lower recycle store count.

Fixes #14410


sakridge:
This seems to help beta node keep up with mainnet overnight:
recycle-stores

Still a bit of a mystery why a copy into a recycled store would be more expensive with ramdisk, but this will be superceded by the accounts caching change anyway.

behzadnouri (approved):

Issue #14414, opened 2021-01-04
The term "singleGossip" is unnecessarily confusing for DX


mvines:
Any dev that learns about "singleGossip" (`CommitmentLevel::SingleGossip` in code, `singleGossip` on the cli/web3.js) will be immediately confused.  Contrast with `CommitmentLevel::max`, which is immediately obvious, and `CommitmentLevel::recent` while not great is at least guessable.     Worse, `CommitmentLevel::Single` is deprecated/discouraged making the singleGossip term even more confusing.

We can clean this up by creating a new `CommitmentLevel::Optimistic` (`optimistic` in the cli) that is mapped to singleGossip.  Then de-document "single" and "singleGossip", and flag CommitmentLevel::Single and CommitmentLevel::SingleGossip as [deprecated](https://doc.rust-lang.org/edition-guide/rust-2018/the-compiler/an-attribute-for-deprecation.html) on use.

PR #14430, opened 2021-01-05
Use optimistic confirmation in getSignatureStatuses, and various downstream client methods

CriesofCarrots:
#### Problem
`getSignatureStatuses` is our recommended api for confirming transactions, but it doesn't convey any information about optimistic confirmation.

#### Summary of Changes
- Add `confirmationStatus` field to TransactionStatus and populate as `recent/optimistic/max`. Adds in check of the latest optimistically-confirmed bank (which we already have in RPC)
- Use `confirmationStatus` in various client methods to offer more-precise determination of transaction confirmation for the requested commitment

Follow-up work (as part of #14414 ): enable use of singleGossip commitment for solana-cli operations, like `solana transfer`, which are currently locked to max

### DIFF
--- a/banks-interface/src/lib.rs
+++ b/banks-interface/src/lib.rs
@@ -12,9 +12,9 @@ use solana_sdk::{
 
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub enum TransactionConfirmationStatus {
-    Recent,
-    Optimistic,
-    Max,
+    Processed,
+    OptimisticallyConfirmed,
+    Finalized,
 }
 
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
--- a/banks-interface/src/lib.rs
+++ b/banks-interface/src/lib.rs
@@ -15,6 +15,7 @@ pub struct TransactionStatus {
     pub slot: Slot,
     pub confirmations: Option<usize>, // None = rooted
     pub err: Option<TransactionError>,
+    pub optimistically_confirmed: Option<bool>,
 }
 
 #[tarpc::service]
--- a/banks-interface/src/lib.rs
+++ b/banks-interface/src/lib.rs
@@ -11,11 +11,18 @@ use solana_sdk::{
 };
 
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub enum TransactionConfirmationStatus {
+    Recent,
+    Optimistic,
+    Max,
+}
+
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub struct TransactionStatus {
     pub slot: Slot,
     pub confirmations: Option<usize>, // None = rooted
     pub err: Option<TransactionError>,
-    pub optimistically_confirmed: Option<bool>,
+    pub confirmation_status: Option<TransactionConfirmationStatus>,
 }
 
 #[tarpc::service]
--- a/banks-interface/src/lib.rs
+++ b/banks-interface/src/lib.rs
@@ -13,7 +13,7 @@ use solana_sdk::{
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub enum TransactionConfirmationStatus {
     Processed,
-    OptimisticallyConfirmed,
+    Confirmed,
     Finalized,
 }
 
--- a/banks-server/src/banks_server.rs
+++ b/banks-server/src/banks_server.rs
@@ -189,11 +189,11 @@ impl Banks for BanksServer {
             confirmations,
             err: status.err(),
             confirmation_status: if confirmations.is_none() {
-                Some(TransactionConfirmationStatus::Max)
+                Some(TransactionConfirmationStatus::Finalized)
             } else if optimistically_confirmed.is_some() {
-                Some(TransactionConfirmationStatus::Optimistic)
+                Some(TransactionConfirmationStatus::OptimisticallyConfirmed)
             } else {
-                Some(TransactionConfirmationStatus::Recent)
+                Some(TransactionConfirmationStatus::Processed)
             },
         })
     }
--- a/banks-server/src/banks_server.rs
+++ b/banks-server/src/banks_server.rs
@@ -191,7 +191,7 @@ impl Banks for BanksServer {
             confirmation_status: if confirmations.is_none() {
                 Some(TransactionConfirmationStatus::Finalized)
             } else if optimistically_confirmed.is_some() {
-                Some(TransactionConfirmationStatus::OptimisticallyConfirmed)
+                Some(TransactionConfirmationStatus::Confirmed)
             } else {
                 Some(TransactionConfirmationStatus::Processed)
             },
--- a/banks-server/src/banks_server.rs
+++ b/banks-server/src/banks_server.rs
@@ -169,6 +169,11 @@ impl Banks for BanksServer {
         let (slot, status) = bank.get_signature_status_slot(&signature)?;
         let r_block_commitment_cache = self.block_commitment_cache.read().unwrap();
 
+        let optimistically_confirmed_bank = self.bank(CommitmentLevel::SingleGossip);
+        let optimistically_confirmed = optimistically_confirmed_bank
+            .get_signature_status_slot(&signature)
+            .map(|_| true);
+
         let confirmations = if r_block_commitment_cache.root() >= slot
             && r_block_commitment_cache.highest_confirmed_root() >= slot
         {
@@ -182,6 +187,7 @@ impl Banks for BanksServer {
             slot,
             confirmations,
             err: status.err(),
+            optimistically_confirmed,
         })
     }
 
--- a/banks-server/src/banks_server.rs
+++ b/banks-server/src/banks_server.rs
@@ -4,7 +4,9 @@ use futures::{
     future,
     prelude::stream::{self, StreamExt},
 };
-use solana_banks_interface::{Banks, BanksRequest, BanksResponse, TransactionStatus};
+use solana_banks_interface::{
+    Banks, BanksRequest, BanksResponse, TransactionConfirmationStatus, TransactionStatus,
+};
 use solana_runtime::{bank::Bank, bank_forks::BankForks, commitment::BlockCommitmentCache};
 use solana_sdk::{
     account::Account,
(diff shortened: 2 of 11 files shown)


mvines (commented):
@jstarry - does the getSignatureStatus API addition look ok to you?  `"optimisticallyConfirmed"` feels a little wordy to me but I don't have a better suggestion at the moment.

Rest lgtm!

docs/src/developing/clients/jsonrpc-api.md
```diff
@@ -1945,6 +1945,7 @@ An array of:
   - `slot: <u64>` - The slot the transaction was processed
   - `confirmations: <usize | null>` - Number of blocks since signature confirmation, null if rooted, as well as finalized by a supermajority of the cluster
   - `err: <object | null>` - Error if transaction failed, null if transaction succeeded. [TransactionError definitions](https://github.com/solana-labs/solana/blob/master/sdk/src/transaction.rs#L24)
+  - `optimisticallyConfirmed: <bool | null>` - Whether the transaction has been optimistically confirmed by the node. See [singleGossip Commitment](jsonrpc-api.md#configuring-state-commitment)
```
jstarry, line 1948:
```suggestion
  - `optimisticallyConfirmed: <bool | null>` - Whether the transaction has been optimistically confirmed by the cluster. See [singleGossip Commitment](jsonrpc-api.md#configuring-state-commitment)
```

banks-interface/src/lib.rs
```diff
@@ -15,6 +15,7 @@ pub struct TransactionStatus {
     pub slot: Slot,
     pub confirmations: Option<usize>, // None = rooted
     pub err: Option<TransactionError>,
+    pub optimistically_confirmed: Option<bool>,
```
jstarry, line 18: Why option wrapped?

jstarry (commented):
How about `confirmation_status` instead and use an enum of "unconfirmed", "optimistically confirmed", and "finalized"? We already represent "finalized" with `confirmations: null` but I always felt that was a bit of a confusing API and it would be nice to use a single field to determine what the client should do

banks-interface/src/lib.rs
CriesofCarrots: Easy backward compatibility

CriesofCarrots:
> How about `confirmation_status` instead and use an enum of "unconfirmed", "optimistically confirmed", and "finalized"?

That's a nice idea. @mvines , work for you?

banks-interface/src/lib.rs
```diff
@@ -10,11 +10,19 @@ use solana_sdk::{
     transaction::{self, Transaction, TransactionError},
 };
 
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub enum TransactionConfirmationStatus {
+    Unconfirmed,
+    OptimisticallyConfirmed,
+    Finalized,
```
mvines, line 17: How about Max to align with `CommitmentLevel`:
```suggestion
    Max,
```

banks-interface/src/lib.rs
```diff
@@ -10,11 +10,19 @@ use solana_sdk::{
     transaction::{self, Transaction, TransactionError},
 };
 
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub enum TransactionConfirmationStatus {
+    Unconfirmed,
+    OptimisticallyConfirmed,
```
mvines, line 16: Am I being to optimistic here?
```suggestion
    Optimistic,
```

banks-interface/src/lib.rs
```diff
@@ -10,11 +10,19 @@ use solana_sdk::{
     transaction::{self, Transaction, TransactionError},
 };
 
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub enum TransactionConfirmationStatus {
+    Unconfirmed,
```
mvines, line 15: More nits: would "unknown" make more sense?   Or by this point do we know for certain that the transaction does in fact exist?
```suggestion
    Unknown,
```

CriesofCarrots: Yes, we do know it exists by this point, as the signature is found in the recent bank.

banks-interface/src/lib.rs
CriesofCarrots: 🤣 
Does make sense, though, with the new field name.

banks-interface/src/lib.rs
mvines: Oh wait, so this could be `Recent` then?

CriesofCarrots: Sure!
Recent, Max, Optimistic... I guess this sets us up nicely for https://github.com/solana-labs/solana/issues/14414

CriesofCarrots:
Not yet sure what's up with these solana-web3 live tests, but I do see the same failures locally without this PR

banks-server/src/banks_server.rs
```diff
@@ -180,6 +186,13 @@ impl Banks for BanksServer {
             slot,
             confirmations,
             err: status.err(),
+            confirmation_status: if confirmations.is_none() {
+                Some(TransactionConfirmationStatus::Max)
```
jstarry, line 190: It doesn't look like the bank server logic matches rpc server logic so I don't think we can set this as `Max` quite yet. We don't have this check: https://github.com/solana-labs/solana/blob/d2b0fd973f91bb0c57f51e7762dfa592344e158a/core/src/rpc.rs#L889

banks-interface/src/lib.rs
jstarry: I prefer finalized for clarity, could we change commitment from max to finalized instead since we already planning to modify commitment level?

banks-interface/src/lib.rs
jstarry: What do we think clients will use `Recent` for? Is it more or less an indication that the transaction was processed?

mvines: iirc, Serum uses Recent for most queries actually for a snappy UI (at the risk of a later correction to the UI)

CriesofCarrots: Yes, it would indicates that the transaction landed.

banks-server/src/banks_server.rs
CriesofCarrots: Looks like we could update BanksServer to use Max, since it has access to the BlockCommitmentCache. The current behavior already has this issue (not matching RpcClient), since it sets confirmations to None when a transaction is rooted.

My main intent was to keep the BanksServer and RpcClient interfaces the same, although I suppose that isn't strictly necessary. We could do with BanksServer whatever best suits ProgramTest, since that is the only downstream user.

banks-interface/src/lib.rs
jstarry: Could we name it "Landed" or "Processed" then? Recent doesn't mean much to a developer I think

banks-server/src/banks_server.rs
jstarry: Yup, keeping the interfaces the same sounds great, just wanted to call out the behavior difference. Could be a follow up fix

CriesofCarrots: Perfect, I will update BanksServer in a follow-up.

banks-interface/src/lib.rs
CriesofCarrots: @jstarry would you feel any better about Recent/Max if this field were renamed to `commitmentStatus`?

jstarry: Nope, I just don't think Recent/Max are dev friendly and it seems like the right time to change them

CriesofCarrots:
I updated this to match the outcome of yesterday's terminology discussion. I also stripped out the web3.js changes, to PR separately and not bork the automated release process.

Follow-up work:
- Update banks-server transaction-status logic to match Rpc
- Switch Commitment recent->processed, max->finalized as part of https://github.com/solana-labs/solana/issues/14414

banks-interface/src/lib.rs
```diff
@@ -10,11 +10,19 @@ use solana_sdk::{
     transaction::{self, Transaction, TransactionError},
 };
 
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub enum TransactionConfirmationStatus {
+    Processed,
+    OptimisticallyConfirmed,
```
mvines, line 16: I'm still stuck on the number of letters in "OptimisticallyConfirmed".   8 syllables! 
Just "Confirmed"?  "Settled" was also proposed in Discord.

CriesofCarrots: Oh, I like "Confirmed". This usage seems consistent with our documented definition of confirmed, and we can document the guarantees of "optimistic confirmation" elsewhere.

I'm not too clear on what "Settled" means in this context. The payment-processing meaning doesn't seem to apply.

mvines: I took these new names for a spin in some `solana-test-validator` output updates that I'm working on, it felt nice:
```
$ solana-test-validator
Mint address: vines1vzrYbzLMRdu58ou5XTby4qAqVRLmqo36NKPTg
Ledger location: test-ledger
Log: test-ledger/validator.log
Identity: FXTGZfrgFeCgFXuFRQziNyZcnXftz56wfQLghfzqpK6j
Version: 1.6.0
JSON RPC URL: http://127.0.0.1:8899
JSON RPC PubSub Websocket: ws://127.0.0.1:8900/
Gossip Address: 127.0.0.1:1024
TPU Address: 127.0.0.1:53170
Faucet Address: 127.0.0.1:9900
⠤ 00:00:46 | Processed Slot: 93 | Confirmed Slot: 92 | Finalized Slot: 61 | Snapshot Slot: 0 | Transactions: 93 | ◎499.999540000
```


Last line: **Processed Slot: 93 | Confirmed Slot: 92 | Finalized Slot: 61**

banks-interface/src/lib.rs
jstarry: I love that these words are all the same length

jstarry (dismissed):
Looks great! I think another nice follow up would be to change the `confirmations` to cluster confirmations to line up with the new "confirmed" confirmation status.

CriesofCarrots:
> Looks great! I think another nice follow up would be to change the `confirmations` to cluster confirmations to line up with the new "confirmed" confirmation status.

I'll look into this. I assume you mean swapping from the current BlockCommitmentCache-based cluster confirmations to something based on optimistically confirmed banks?

jstarry:
> I'll look into this. I assume you mean swapping from the current BlockCommitmentCache-based cluster confirmations to something based on optimistically confirmed banks?

Yes, exactly.

PR #14437, opened 2021-01-05
explorer: add load button for largest accounts card

oJshua:
#### Problem
Loading the largest accounts has performance issues and until we can properly fix, we will introduce a load button.

#### Summary of Changes
Adds a load button to Largest Accounts card on Supply page.


mvines (approved):
Thanks

PR #14438, opened 2021-01-05
patches crds vote-index assignment bug

behzadnouri:
#### Problem
If tower is full, old votes are evicted from the front of the deque:
https://github.com/solana-labs/solana/blob/2074e407c/programs/vote/src/vote_state/mod.rs#L367-L373
whereas recent votes if expire are evicted from the back:
https://github.com/solana-labs/solana/blob/2074e407c/programs/vote/src/vote_state/mod.rs#L529-L537

As a result, from a single tower_index scalar, we cannot infer which crds-vote
should be overwritten:
https://github.com/solana-labs/solana/blob/2074e407c/core/src/crds_value.rs#L576

In addition there is an off by one bug in the existing code. tower_index is
bounded by MAX_LOCKOUT_HISTORY - 1:
https://github.com/solana-labs/solana/blob/2074e407c/core/src/consensus.rs#L382
So, it is at most 30, whereas MAX_VOTES is 32:
https://github.com/solana-labs/solana/blob/2074e407c/core/src/crds_value.rs#L29
Which means that this branch is never taken:
https://github.com/solana-labs/solana/blob/2074e407c/core/src/crds_value.rs#L590-L593
so crds table alwasys keeps 29 **oldest** votes by wallclock, and then
only overrides the 30st one each time. (i.e a tally of only two most
recent votes).

#### Summary of Changes
* Pass current state of tower to push-vote.
* Find which crds vote is evicted from the tower, and override that one.

### DIFF
--- a/core/src/cluster_info.rs
+++ b/core/src/cluster_info.rs
@@ -1085,8 +1085,9 @@ impl ClusterInfo {
         // deque, or recent vote which has expired before getting enough
         // confirmations.
         // If all votes are still in the tower, add a new vote-index. If more
-        // than one vote is evicted, the one with most recent wallclock is
-        // returned because that had got fewer confirmations.
+        // than one vote is evicted, the oldest one by wallclock is returned in
+        // order to allow more recent votes more time to propagate through
+        // gossip.
         // TODO: When there are more than one vote evicted from the tower, only
         // one crds vote is overwritten here. Decide what to do with the rest.
         let mut num_crds_votes = 0;
@@ -1117,7 +1118,7 @@ impl ClusterInfo {
                         _ => panic!("this should not happen!"),
                     }
                 })
-                .max() // Take the evicted vote with the most recent wallclock.
+                .min() // Boot the oldest evicted vote by wallclock.
                 .map(|(_ /*wallclock*/, ix)| ix)
         };
         let vote_index = vote_index.unwrap_or(num_crds_votes);
--- a/core/src/cluster_info.rs
+++ b/core/src/cluster_info.rs
@@ -3718,6 +3718,81 @@ mod tests {
         assert_eq!(max_ts, new_max_ts);
     }
 
+    fn new_vote_transaction<R: Rng>(rng: &mut R, slots: Vec<Slot>) -> Transaction {
+        let vote = Vote::new(slots, solana_sdk::hash::new_rand(rng));
+        let ix = vote_instruction::vote(
+            &Pubkey::new_unique(), // vote_pubkey
+            &Pubkey::new_unique(), // authorized_voter_pubkey
+            vote,
+        );
+        Transaction::new_with_payer(
+            &[ix], // instructions
+            None,  // payer
+        )
+    }
+
+    #[test]
+    fn test_push_votes_with_tower() {
+        let get_vote_slots = |cluster_info: &ClusterInfo, now| -> Vec<Slot> {
+            let (labels, _, _) = cluster_info.get_votes(now);
+            let gossip = cluster_info.gossip.read().unwrap();
+            let mut vote_slots = HashSet::new();
+            for label in labels {
+                match &gossip.crds.lookup(&label).unwrap().data {
+                    CrdsData::Vote(_, vote) => {
+                        assert!(vote_slots.insert(vote.slot().unwrap()));
+                    }
+                    _ => panic!("this should not happen!"),
+                }
+            }
+            vote_slots.into_iter().collect()
+        };
+        let mut rng = rand::thread_rng();
+        let now = timestamp();
+        let keys = Keypair::new();
+        let contact_info = ContactInfo::new_localhost(&keys.pubkey(), 0);
+        let cluster_info = ClusterInfo::new_with_invalid_keypair(contact_info);
+        let mut tower = Vec::new();
+        for k in 0..MAX_LOCKOUT_HISTORY {
+            let slot = k as Slot;
+            tower.push(slot);
+            let vote = new_vote_transaction(&mut rng, vec![slot]);
+            cluster_info.push_vote(&tower, vote);
+        }
+        let vote_slots = get_vote_slots(&cluster_info, now);
+        assert_eq!(vote_slots.len(), MAX_LOCKOUT_HISTORY);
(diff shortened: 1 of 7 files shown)


core/src/cluster_info.rs
```diff
@@ -1102,22 +1103,59 @@ impl ClusterInfo {
...
-            let r_gossip =
+        // Find a crds vote which is evicted from the tower, and recycle its
+        // vote-index. This can be either an old vote which is popped off the
+        // deque, or recent vote which has expired before getting enough
+        // confirmations.
+        // If all votes are still in the tower, add a new vote-index. If more
+        // than one vote is evicted, the one with most recent wallclock is
+        // returned because that had got fewer confirmations.
+        // TODO: When there are more than one vote evicted from the tower, only
+        // one crds vote is overwritten here. Decide what to do with the rest.
```
carllin, line 1117: I think it's ok to keep the others in the tower. Votes spending more time in gossip gives a better chance for observers to consume them for things like optimistic confirmation

core/src/cluster_info.rs
```diff
@@ -1102,22 +1103,59 @@ impl ClusterInfo {
...
+                    num_crds_votes += 1;
+                    match &vote.data {
+                        CrdsData::Vote(_, vote) if should_evict_vote(vote) => {
+                            Some((vote.wallclock, ix))
+                        }
+                        CrdsData::Vote(_, _) => None,
+                        _ => panic!("this should not happen!"),
+                    }
+                })
+                .max() // Take the evicted vote with the most recent wallclock.
```
carllin, line 1146: hmmm thinking more on this, I'm a bit worried about evicting the most recent wallclock because of the implications for optimistic confirmation which relies on giving those votes enough time to propagate through the rest of the network.  Please see: https://github.com/solana-labs/solana/blob/master/core/src/cluster_info_vote_listener.rs#L578-L596

So where before, if we had two forks

```
1, 2, 3, 4, 5, 6
     |
     20
```
Let's say the order of events is:

1)  the validator first voted on slots 1, 2, 3, 4 ,5, 6 on the first fork
2) The lock on slot 3 is `16`, so they can't vote on fork 2 until slot 20
3) The validator now votes on slot 20

In the original code, slot 20 would pop off the vote for slot 3 (which already has been sitting in gossip for 17 slots), giving the later votes 4, 5, 6 to propagate. In the new code, this vote for 20 would immediately pop off slot `6`, which hasn't had as long to propagate.

behzadnouri: > In the new code, this vote for 20 would immediately pop off slot 6, which hasn't had as long to propagate.

So, one thing I do not understand is if a validator has switched to a different fork, why should we want to propagate a recently expired vote (with no or only few confirmations) from an older fork which this validator has already switched off?
Shouldn't we instead favor propagating votes with more confirmations?

> In this respect, I think it may be safer to stay as close to the original code to place the vote based on its position in the tower.

The current code is actually giving less time to votes to propagate. It keeps 29 oldest votes by wallclock, and then
only overrides the 30st one each time. (i.e a tally of only two most recent votes):
https://github.com/solana-labs/solana/blob/85fdbd87b/core/src/crds_value.rs#L612-L617

So once it hits `MAX_VOTES`, a vote is kicked out as soon as there are only 2 more recent votes.

carllin: >The current code is actually giving less time to votes to propagate. It keeps 29 oldest votes by wallclock, and then
only overrides the 30st one each time. (i.e a tally of only two most recent votes):
https://github.com/solana-labs/solana/blob/85fdbd87b/core/src/crds_value.rs#L612-L617

Right that's assuming we keep pushing to the stack in such a way that the stack is growing right? i.e. the tower index is always `30` on each update. But in the above example, let's say the the tower index on the vote `6` is `N`. Then after vote `20`, because `3, 4, 5, 6` are all removed from the tower, the tower index for the vote `20` update is now `N - 3`, replacing vote `3` in the stack. In the new code we would remove slot `6`.

>So, one thing I do not understand is if a validator has switched to a different fork, why should we want to propagate a recently expired vote (with no or only few confirmations) from an older fork which this validator has already switched off?
Shouldn't we instead favor propagating votes with more confirmations?

Yeah, so the consideration here is for the optimistic confirmation code here: https://github.com/solana-labs/solana/blob/master/core/src/cluster_info_vote_listener.rs#L578-L596, which only counts a vote toward a slot `S`'s optimistic confirmation if the *last* slot in the vote ==  `S`. Thus the reason to keep around an expired vote is to give the last vote maximal time to propagate, just in case the previous fork was the major fork. 

So in the example above, slots `3, 4, 5, 6` were all removed, so we're just deciding if it's better to remove `3` or `6`. My feeling is because the vote for `6` potentially carries more information (it includes earlier missing votes between 3 and 6 that may not have landed, and is for a more specific fork than just `3`).  It may not make a big difference in the general case, but it differs from the behavior today. Maybe @aeyakovenko can comment if there are additional subtlties here.

behzadnouri: updated the code to boot the oldest vote by wallclock

core/src/cluster_info.rs
```diff
@@ -1076,22 +1077,60 @@ impl ClusterInfo {
...
+                    num_crds_votes += 1;
+                    match &vote.data {
+                        CrdsData::Vote(_, vote) if should_evict_vote(vote) => {
+                            Some((vote.wallclock, ix))
+                        }
+                        CrdsData::Vote(_, _) => None,
+                        _ => panic!("this should not happen!"),
+                    }
+                })
+                .min() // Boot the oldest evicted vote by wallclock.
```
carllin, line 1121: Can we add a test that:
1)  When we have the MAX_LOCKOUT_HISTORY votes, pushing a new vote actually pops off the first one (confirm behavior of the current bug is gone).
2) When multiple votes are up for eviction, the oldest one is evicted.

behzadnouri: added more test coverage

> When we have the MAX_LOCKOUT_HISTORY votes, pushing a new vote actually pops off the first one

It will pop off whatever is not in the `tower: &[Slot]` argument. It can be the first or it can be the last.

carllin (approved):

PR #14439, opened 2021-01-05
explorer: display block time on cluster stats

oJshua:
#### Problem
Cluster stats should display the current block time.ie, every time we update the current slot,  include a call to getBlockTime, and display the current time (same local time format used for locked stake accounts).

#### Summary of Changes
Fetches blocktime when the absolute slot changes. Displayed with proper format on cluster stats page.


oJshua:
Explorer-Solana

mvines (approved):

PR #14440, opened 2021-01-05
consolidate constants related to time for future refactoring

jeffwashington:
#### Problem
Multiple places in the code perform calculations time per slot time in similar ways.

#### Summary of Changes
This change cleans up constant usage and consolidates calculations. This makes it more clear that the code is trying to use time per slot.

The concepts of:
default seconds per slot
default milliseconds per slot
are used instead of a calculation based on
ticks per slot
and
ticks per second.

Fixes #


sakridge (approved):
lgtm

PR #14447, opened 2021-01-05
Add fixed tick rate adjustment

sakridge:
#### Problem

PoH a bit too slow after update to hit target tick rate.

#### Summary of Changes

Adjust the target tick time down by 50ms.

Fixes #

### DIFF
--- a/core/src/poh_service.rs
+++ b/core/src/poh_service.rs
@@ -21,6 +21,8 @@ pub const NUM_HASHES_PER_BATCH: u64 = 1;
 
 pub const DEFAULT_PINNED_CPU_CORE: usize = 0;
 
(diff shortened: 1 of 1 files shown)


core/src/poh_service.rs
```diff
@@ -21,6 +21,8 @@ pub const NUM_HASHES_PER_BATCH: u64 = 1;
 
 pub const DEFAULT_PINNED_CPU_CORE: usize = 0;
 
+const TARGET_TICK_ADJUSTMENT_NS: u64 = 50_000_000;
```
CriesofCarrots, line 24:
```suggestion
const TARGET_SLOT_ADJUSTMENT_NS: u64 = 50_000_000;
```

CriesofCarrots (approved):
lgtm

PR #14448, opened 2021-01-05
Fix downstream build

CriesofCarrots:
#### Problem
Downstream build broken

#### Summary of Changes
Bump serum-dex pegged commit. Build failures seem to be due to borsh conflicts with the serum fork.


CriesofCarrots:
I'm merging ahead of green to waste less dev time. This shouldn't affect the local-cluster job.

PR #14449, opened 2021-01-05
Enable program upgrades via CPI

jackcmay:
#### Problem

All access to loaders via CPI is blocked but for multisig program upgrades, the `Upgrade` instruction must be performed from the governance program.

#### Summary of Changes

Open up CPI to allow upgradeable loader `Upgrade` instructions

Fixes #

### DIFF
--- a/programs/bpf/tests/programs.rs
+++ b/programs/bpf/tests/programs.rs
@@ -21,7 +21,7 @@ use solana_runtime::{
 };
 use solana_sdk::{
     account::Account,
-    bpf_loader, bpf_loader_deprecated,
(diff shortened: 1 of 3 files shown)


jstarry:
Seems useful to allow setting a new authority as well

jackcmay:
Yeah, talked to the wormhole folks about that, they said they didn't need it.  I'm going to leave the hole small unless we get feedback that it needs to be bigger.

mvines:
Was there a v1.4 backport for this as well @jackcmay?

jackcmay:
No, I held a few changes out of 1.4, can do more of those tomorrow.

PR #14467, opened 2021-01-06
limits number of crds values associated with a pubkey

behzadnouri:
#### Problem
Some crds values are not bounded for a fixed pubkey,
and a node can bloat crds table by pushing too many records.

#### Summary of Changes
Added a limit on the number of crds values associated with a pubkey.
When the limit is reached older values are purged from the table.


behzadnouri:
This is to limit current spike in crds table size:
crds_table_size

sakridge:
> This is to limit current spike in crds table size:

So, is it that there are a ton of NodeInstance lablels?

behzadnouri:
> So, is it that there are a ton of NodeInstance lablels?

Currently that is the only type in v1.4 that can keep growing in count regardless of the number of nodes. But I am not sure why there should be a ton of those.

Nevertheless, we are going to need this bound anyways, because otherwise a node intentionally or because of a bug can push unlimited records to crds table, and those never get purges as long as the node is alive.

core/src/crds.rs
```diff
@@ -275,22 +280,52 @@ impl Crds {
...
-                .with_min_len(1024)
-                .filter_map(|(k, v)| {
-                    let timeout = timeouts.get(&k.pubkey()).unwrap_or(&default_timeout);
-                    if v.local_timestamp.saturating_add(*timeout) <= now {
-                        Some(k.clone())
-                    } else {
+        let evict = |pubkey, index: &IndexSet<usize>| {
+            let timeout = *timeouts.get(pubkey).unwrap_or(&default_timeout);
+            let mut old_labels = Vec::new();
+            let mut buffer: Vec<_> = index
```
sakridge, line 297: Took me a while to read this. There might be another name for `buffer` which is more descriptive. And maybe some comments either here and/or the function signature about "Find old or extra labels" and "Prune extra labels from the same pubkey".

behzadnouri: Added more comments and type annotations.
The `buffer` is literally buffering indices to be partitioned. I added comment for clarity but if you have suggestion for a better name I will take it.

core/src/crds.rs
```diff
@@ -275,22 +280,57 @@ impl Crds {
...
-                    let timeout = timeouts.get(&k.pubkey()).unwrap_or(&default_timeout);
-                    if v.local_timestamp.saturating_add(*timeout) <= now {
-                        Some(k.clone())
-                    } else {
+        // Given an index of all crd values associated with a pubkey,
+        // returns crds labels of old values to be evicted.
+        let evict = |pubkey, index: &IndexSet<usize>| {
+            let timeout = *timeouts.get(pubkey).unwrap_or(&default_timeout);
+            let mut old_labels = Vec::new();
+            // Buffer of crds values to be evicted based on their wallclock.
```
sakridge, line 299: `old_labels` is a collection of crds values to be evicted right?

`buffer` is a collection of crds values which are new but in a label which is potentially unlimited per pubkey. From the comments I don't get any notion of this.

Maybe `recent_unlimited_labels` as the name.

core/src/crds_value.rs
```diff
@@ -428,6 +428,23 @@ impl CrdsValueLabel {
...
+            CrdsValueLabel::ContactInfo(_) => Some(1),
+            CrdsValueLabel::Vote(_, _) => Some(MAX_VOTES as usize),
+            CrdsValueLabel::LowestSlot(_) => Some(1),
+            CrdsValueLabel::SnapshotHashes(_) => Some(1),
+            CrdsValueLabel::EpochSlots(_, _) => Some(MAX_EPOCH_SLOTS as usize),
+            CrdsValueLabel::AccountsHashes(_) => Some(1),
+            CrdsValueLabel::LegacyVersion(_) => Some(1),
+            CrdsValueLabel::Version(_) => Some(1),
+            CrdsValueLabel::NodeInstance(_, _) => None,
+            CrdsValueLabel::DuplicateShred(_, _) => None,
```
sakridge, line 445: Can we not inherently limit these though? Can NodeInstance be limited to 2 unique values per pubkey?
And DuplicateShred could have a fixed limit, maybe around 50-100 shred entries per pubkey.

behzadnouri: With node-instance (which is causing trouble right now), it is very tricky. Because if there are 2 instances of the same node, and they pick the same index, then they will override each other's `NodeInstance` in the crds table, making it less likely for one to see the other (because every 3rd node can only have `NodeInstance` of one of these two in its crds table at each moment). That was the reason the token was added to the label in https://github.com/solana-labs/solana/pull/14037

With the two instances operating independently (and unaware of each other), it is not trivial how each can choose a different index not overriding the other's instance in the table.

core/src/crds.rs
behzadnouri: done

core/src/crds_value.rs
sakridge: How would they pick the same index, isn't it random u64?

behzadnouri: The `token` is u64, and that is what it is included in the label, effectively making `NodeInstances` per pubkey unlimited:
https://github.com/solana-labs/solana/blob/30038a884/core/src/crds_value.rs#L395

What I was saying above is that it is not trivial to change:
```
NodeInstance(Pubkey, u64 /*token*/),
```
to
```
type NodeInstanceIndex = u8;
const MAX_NODE_INSTANCE_INDEX: NodeInstanceIndex = 2;

enum CrdsValueLabel {
  ... 
  NodeInstance(Pubkey, NodeInstanceIndex),
}
```
like what is done for `VoteIndex`:
https://github.com/solana-labs/solana/blob/30038a884/core/src/crds_value.rs#L28-L32

Because instances may pick the same `NodeInstanceIndex` overriding each others in gossip.

sakridge: Right, you would still have something like the custom logic above to clean it up because the label itself does not limit the number of values inherently. I guess what I'm saying instead of having this 512 limit, it can be even stricter for NodeInstance, just the two latest unique labels need to be present.

behzadnouri: > I guess what I'm saying instead of having this 512 limit, it can be even stricter for NodeInstance, just the two latest unique labels need to be present.

That is a fair point, but I tried to keep the logic here simple for now, to first stop the crds table unnecessarily getting bigger.

But, the actual bug here is that nodes keep generating new node-instances, which I am trying to debug why.

sakridge: Ok, yea that's fine. thanks.

sakridge (approved):

PR #14468, opened 2021-01-06
Variable hashes per tick plumbing

jeffwashington:
#### Problem
Create the plumbing to allow hashes_per_tick to change from a constant value.

#### Summary of Changes
Bank contains the state which is the correct hashes_per_tick.

Fixes #

### DIFF
--- a/core/src/poh_service.rs
+++ b/core/src/poh_service.rs
@@ -24,6 +24,9 @@ pub const DEFAULT_PINNED_CPU_CORE: usize = 0;
 const TARGET_SLOT_ADJUSTMENT_NS: u64 = 50_000_000;
 
 impl PohService {
+    // Note that poh_config is a snapshot at this function call.
+    // At runtime, the rest of the system may transition to use a new poh_config instance.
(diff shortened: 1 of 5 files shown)


jeffwashington:
@carllin further testing shows this isn't quite ready yet.

core/src/poh_service.rs
```diff
@@ -24,6 +24,9 @@ pub const DEFAULT_PINNED_CPU_CORE: usize = 0;
 const TARGET_SLOT_ADJUSTMENT_NS: u64 = 50_000_000;
 
 impl PohService {
+    // Note that poh_config is a snapshot at this function call.
+    // At runtime, the rest of the system may transition to use a new poh_config instance.
```
sakridge, line 28: Think we should consider just removing the Arc and passing cloned versions to these services. The Arc makes it look like it's shared and somewhat immutable, but it's not really the case. May obviate the need for this comment.

jeffwashington:
not ready - needs some more work.

PR #14482, opened 2021-01-08
Add buffer authority to upgradeable loader

jackcmay:
#### Problem

When upgrading a program via a multisig governance program, the participants need to know that the data used in the upgrade is immutable before they sign.  The current upgradeable loader does not enforce any signing requirement on the buffer account.

#### Summary of Changes

- Add an authority to the buffer account that once set to `None` leaves the buffer account immutable
- Update the CLI to load buffer accounts, set buffer account authorities, and enable `deploy` to optionally take a buffer account instead of a program location

Fixes #

### DIFF
--- a/cli/src/cli.rs
+++ b/cli/src/cli.rs
@@ -2665,7 +2665,7 @@ mod tests {
         let program_id = json
             .as_object()
             .unwrap()
-            .get("programId")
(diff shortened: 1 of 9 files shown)


jackcmay:
@mvines fyi, try out the cli if you get a chance

PR #14487, opened 2021-01-08
Various snapshot-related code clean up

mvines:
Minor code improvements and refactoring that I ran into while walking through the snapshot code
* Avoid excessive mkdir-ing of accounts directories
* Replace wrong `symlink_dir` usage with `symlink_file`
* Limit TempDir surface area
* Rename AccountsPackage::root to AccountsPackage::slot for more clarity
* Rename CompressionType to ArchiveFormat.   This makes room for other snapshot archive types that aren't tarball based  or aren't full snapshots


mvines:
(going to backport this to v1.4 as well, to hopefully prevent future merge conflicts until we EOL v1.4)

PR #14488, opened 2021-01-08
Suppress cargo audit failure for `difference` crate

mvines:
There's no newer crate to upgrade to yet


mvines:
Also need to bump smallvec

PR #14494, opened 2021-01-08
chore: bpf-loader test, don't match specific error text

CriesofCarrots:
#### Problem
As of https://github.com/solana-labs/solana/pull/14286, Loader doesn't skip preflight checks. That means that the `failedLoad` in `bpf-loader.test.js` will more often than not fail on a preflight simulation, rather than the confirmation. This fails the expect, which is looking for the specific thrown confirmation error, and subsequently all the following tests.

#### Summary of Changes
Don't match specific error text in expect case


CriesofCarrots:
There's still something flaky in `stake-program.test.js - live staking actions`, but I haven't spotted it yet

PR #14514, opened 2021-01-09
Update kin.svg

dillontking:
Replace Kin token logo SVG to cropped "coin" style icon rather than full brand logo, improving visibility particularly at lower resolutions in explorer UI.

@blahartinger


mvines:
Thanks

PR #14516, opened 2021-01-09
Avoid tmp snapshot backlog in SnapshotPackagerService under high load

mvines:
It's common to see `ledger/snapshot/tmp-snapshot-*` directories piling up as a snapshot archive is being created.   This occurs because the snapshot packager is busy making an archive while new snapshot packages are piling up in the channel.   However once the current archive is built, snapshot packager discards all but the latest package in the channel.

Instead rework the snapshot packager to pull packages off the channel as quickly as they come in

### DIFF
--- a/core/src/accounts_hash_verifier.rs
+++ b/core/src/accounts_hash_verifier.rs
@@ -4,10 +4,11 @@
 // hash on gossip. Monitor gossip for messages from validators in the --trusted-validators
 // set and halt the node if a mismatch is detected.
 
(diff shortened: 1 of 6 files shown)


sakridge:
Could we remove the original channel and just use the mutex from the other side of the snapshot_package_sender?

mvines:
> Could we remove the original channel and just use the mutex from the other side of the snapshot_package_sender?

Ah, nice idea.  Latest commit in the PR now does this

mvines:
@sakridge - do you care to review this again with the channel removal changes?

sakridge (approved):

PR #14517, opened 2021-01-09
Clean accounts dirs earlier

sakridge:
#### Problem

Accounts files can take significant resources when mapped into memory causing problems for validator startup.

#### Summary of Changes

Clean the accounts paths earlier in the validator startup.

Fixes #

### DIFF
--- a/core/src/validator.rs
+++ b/core/src/validator.rs
@@ -42,7 +42,6 @@ use solana_ledger::{
     leader_schedule_cache::LeaderScheduleCache,
     poh::compute_hash_time_ns,
 };
-use solana_measure::measure::Measure;
 use solana_metrics::datapoint_info;
 use solana_runtime::{
     accounts_index::AccountIndex,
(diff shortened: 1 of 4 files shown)


core/src/validator.rs
```diff
@@ -285,19 +284,6 @@ impl Validator {
             }
         }
 
-        info!("Cleaning accounts paths..");
```
mvines, line 288: Should we keep this cleanup here as well for other users of `Validator`, such as `solana-test-validator` and `local-cluster`?

sakridge: `local-cluster` doesn't really need it although better to test that path there than not, I'm not sure about `solana-test-validator`.

The other option is move all the snapshot download stuff into `Validator::new` and then re-order it there.

mvines: `solana-test-validator` probably doesn't strictly need it I guess, the accounts are generally expected to be small here since it's just a dev tool.   Moving the snapshot stuff into Validator::new would be nice, probably a ton of churn with low ROI at this point.  One day I'd love to Validator::new to live in `validator/` at which point download refactoring would be much easier.

sakridge: Ok, I'll close for now, it was to mitigate validators having to rm -rf this directory on startup manually.

Issue #2436, opened 2019-01-15
Slow locks in Rust standard library


garious:
#### Problem

Rust's default thread synchronization libraries are a bit slow and we're starting to workaround that slowness in ways that don't offer a clear path forward. For example, in the Bank, rather than wrapping each account with a RwLock, we maintain a separate `HashSet<Pubkey>` of all "locked" accounts on the side.

#### Proposed Solution

First try the parking_lot library as a drop-in replacement. If it goes well, consider replacing that `HashSet<Pubkey>` with `RwLock<Account>`.

https://docs.rs/parking_lot/0.7.1/parking_lot/

sambley:
@garious, @pgarg66, tried moving over RwLock and Mutex to parking lot but the results are worse for our case. The average TPS went down by 77% after moving over to use RwLock/Mutex from parking_lot.

From the raw benchmark numbers, write locks are much slower and read lock / mutex operations are faster. 

- Running with 10 writer threads and 10 reader threads
parking_lot::RwLock  - [write]    109.170 kHz [read]     55.892 kHz
std::sync::RwLock    - [write]      7.401 kHz [read]   4430.701 kHz

       name         |    average     |     median     |    std.dev.   
parking_lot::Mutex   |     27.596 kHz |     27.513 kHz |      0.368 kHz
std::sync::Mutex     |     53.418 kHz |     53.395 kHz |      0.351 kHz

Most likely the write lock threads in our use case are not getting scheduled fast enough which probably is contributing significantly to the processing of transactions.

garious:
Thanks for the analysis @sambley. If no other thoughts on how to replace that `HashSet<Pubkey>` with something more elegant, can you unassign this issue? Plenty of other fish to fry.

PR #14521, opened 2021-01-10
Use parking_lot in highly-contended locks

ryoqun:
My 30 min take to revisit https://github.com/solana-labs/solana/issues/2436

#### Problem

In my understanding, std locks is generally under-performing under the sake of broader platform support.

Also, we have now lock for each account address. It's getting not ignorable for lock memory overhead.

#### Summary of Changes

Prove it.

Fixes #

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2546,9 +2546,9 @@ dependencies = [
 
 [[package]]
 name = "parking_lot"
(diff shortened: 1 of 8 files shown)


ryoqun:
well, sadly I cannot see any positive perf results...

PR #14547, opened 2021-01-12
Use leader_forward_count for tx retries too

CriesofCarrots:
#### Problem
The send-transaction-service can be configured to send each transaction to more than one upcoming leader to increase the chances of the transaction landing. However, when a transaction is retried, it is only sent to one. This was my oversight.

#### Summary of Changes
- Use leader_forward_count for tx retries too, increasing the chances of a transaction landing sooner

### DIFF
--- a/core/src/send_transaction_service.rs
+++ b/core/src/send_transaction_service.rs
@@ -196,6 +196,7 @@ impl SendTransactionService {
                             &tpu_address,
                             &mut transactions,
                             &leader_info,
+                            leader_forward_count,
                         );
                     }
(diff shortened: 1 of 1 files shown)


CriesofCarrots:
@sakridge post-merge comments welcome if you have any concerns about this. Looked pretty innocuous on my testnet runs.

sakridge:
> @sakridge post-merge comments welcome if you have any concerns about this. Looked pretty innocuous on my testnet runs.

Ok, yea. seemed like it could work better this way.

PR #14553, opened 2021-01-13
validator: Add --rpc-threads argument

mvines:


mvines:
@t-nelson - yep, we're really doing this :).  Preliminary testing on bv5 shows I can use `--rpc-threads 512` and still have a bit of headroom, and the timeouts do appear to be the tokio threadpool not being able to pick up new work.

PR #14565, opened 2021-01-13
patches bug in turbine's neighbors computation

behzadnouri:
#### Problem
Removing local node's index early from the set here:
https://github.com/solana-labs/solana/blob/e1b59ded4/core/src/retransmit_stage.rs#L346
distorts the order of nodes depending on which node is computing the
turbine fan-out tree, and results in incorrect neighbors computation.

#### Summary of Changes
* Remove self _after_ computing turbine neighbors.


sakridge (approved):
lgtm

PR #14566, opened 2021-01-13
stake-o-matic: Add --max-commission argument, default 100%

mvines:
Validators with a commission greater than this amount get unstaked.


t-nelson (approved):
LGTM!

danpaul000 (approved):
Thanks!

PR #14578, opened 2021-01-14
Use parking_lot in highly-contended locks 2

ryoqun:
(this is retake of #14521)

My 30 min take to revisit https://github.com/solana-labs/solana/issues/2436

#### Problem

In my understanding, std locks is generally under-performing under the sake of broader platform support.

Also, we have now lock for each account address. It's getting not ignorable for lock memory overhead.

bench-tps didn't show any noticiable improvement (what I did at #14521); but, I noticed accounts clean had decent improvement:

Dominant part (`accounts_scan`) of `clean_accounts()` is reduced from ~6.2s to ~3.6s (~40% reduction) for mainnet-beta.

BEFORE
```
[2021-01-14T07:48:36.138645068Z INFO  solana_metrics::metrics] datapoint: clean_accounts accounts_scan=6225743i clean_old_rooted=8353i store_counts=217420i purge_filter=35229i calc_deps=100237i reclaims=261i
[2021-01-14T07:48:59.286480276Z INFO  solana_metrics::metrics] datapoint: clean_accounts accounts_scan=6208862i clean_old_rooted=28683i store_counts=217170i purge_filter=34473i calc_deps=100716i reclaims=780i
[2021-01-14T07:49:22.233762757Z INFO  solana_metrics::metrics] datapoint: clean_accounts accounts_scan=6209687i clean_old_rooted=22747i store_counts=218517i purge_filter=35076i calc_deps=102270i reclaims=739i
[2021-01-14T07:49:44.504727997Z INFO  solana_metrics::metrics] datapoint: clean_accounts accounts_scan=6218259i clean_old_rooted=23315i store_counts=216620i purge_filter=34327i calc_deps=100805i reclaims=563i
[2021-01-14T07:50:05.647851603Z INFO  solana_metrics::metrics] datapoint: clean_accounts accounts_scan=6210132i clean_old_rooted=22951i store_counts=214930i purge_filter=34004i calc_deps=100984i reclaims=430i
[2021-01-14T07:50:26.660171909Z INFO  solana_metrics::metrics] datapoint: clean_accounts accounts_scan=6209848i clean_old_rooted=14793i store_counts=216184i purge_filter=34297i calc_deps=100606i reclaims=330i
[2021-01-14T07:50:48.947896826Z INFO  solana_metrics::metrics] datapoint: clean_accounts accounts_scan=6279651i clean_old_rooted=16856i store_counts=238793i purge_filter=37990i calc_deps=110786i reclaims=492i
[...]

PR #14592, opened 2021-01-14
Partial accounts clean (v1.4)

sakridge:
#### Problem

Full accounts clean scan takes a long time

#### Summary of Changes

Keep track of zero lamport and delta account updates to drive the clean instead of full accounts scan.

Fixes #


sakridge:
@ryoqun @carllin 
This puts the clean down to ~1 second vs. the 10s it is now. zero-lamport accounts is around 200k. delta keys smaller. Snapshots and accounts sizes seem to be roughly the same.

Not sure I have the locking right yet, but hows the concept? I think we can drive the `uncleaned_keys` from the cache entries on master, but that's not on 1.4.

ryoqun:
This looks nice! I'll review later.

runtime/src/accounts_db.rs
```diff
@@ -773,15 +780,29 @@ impl AccountsDB {
...
+                }
+            }
+        }
+        let delta_key_count = pubkeys.len();
+        let zero_lamport_count = {
+            let zero_lamport_accounts = self.zero_lamport_accounts.read().unwrap();
+            pubkeys.extend(zero_lamport_accounts.iter());
+            zero_lamport_accounts.len()
+        };
+        pubkeys.sort_unstable();
```
carllin, line 801: Could use a `BTreeSet<Slot>` here to avoid the sort and dedup

runtime/src/accounts_db.rs
```diff
@@ -773,15 +780,29 @@ impl AccountsDB {
...
 
+        let mut key_create = Measure::start("key_create");
+        let mut pubkeys: Vec<Pubkey> = Vec::new();
+        {
+            let mut uncleaned_keys = self.uncleaned_keys.lock().unwrap();
+            let max_slot = max_clean_root.unwrap_or(u64::MAX);
+            let slots: Vec<Slot> = uncleaned_keys.keys().cloned().collect();
+            for slot in slots {
+                if slot < max_slot {
+                    pubkeys.extend(uncleaned_keys.remove(&slot).unwrap());
```
carllin, line 791: In the old code, we only care about the pubkeys for rooted slots, so we could check filter these slots by the `AccountsIndex.roots_tracker.uncleaned_roots` set as well. One thing to note is that @ryoqun  noticed quite a lot of contention on that `uncleaned_roots` lock already: https://github.com/solana-labs/solana/pull/14578#issuecomment-760061156, so it would be good to extract it into a `DashSet`

runtime/src/accounts_db.rs
```diff
@@ -773,15 +780,29 @@ impl AccountsDB {
...
         self.report_store_stats();
 
+        let mut key_create = Measure::start("key_create");
+        let mut pubkeys: Vec<Pubkey> = Vec::new();
+        {
+            let mut uncleaned_keys = self.uncleaned_keys.lock().unwrap();
+            let max_slot = max_clean_root.unwrap_or(u64::MAX);
+            let slots: Vec<Slot> = uncleaned_keys.keys().cloned().collect();
+            for slot in slots {
+                if slot < max_slot {
```
carllin, line 790: In the old code, we only care about the pubkeys for rooted slots, so we could check filter these slots by the `AccountsIndex.roots_tracker.uncleaned_roots` set as well. One thing to note is that @ryoqun  noticed quite a lot of contention on that `uncleaned_roots` lock already: https://github.com/solana-labs/solana/pull/14578#issuecomment-760061156, so it would be good to extract it into a `DashSet`

runtime/src/accounts_db.rs
```diff
@@ -419,6 +419,10 @@ pub struct AccountsDB {
     /// Keeps tracks of index into AppendVec on a per slot basis
     pub accounts_index: AccountsIndex<AccountInfo>,
 
+    pub uncleaned_keys: Mutex<HashMap<Slot, Vec<Pubkey>>>,
```
carllin, line 422: I think we can also remove dead/unrooted slots from `uncleaned_keys` in `purge_slots()` as part of slot cleanup when slots are dropped via `Bank::drop()`,

runtime/src/accounts_db.rs
```diff
@@ -419,6 +419,10 @@ pub struct AccountsDB {
...
+
+    pub zero_lamport_accounts: RwLock<HashSet<Pubkey>>,
```
carllin, line 424: I think converting these to `DashMap` and `DashSet` respectively may not be a bad idea, @ryoqun was seeing a lot of contention on locks between replay/ and the `self.thread_pool_clean.install` in `clean`

carllin:
I'm slightly worried about the memory efficiency of  the `pub uncleaned_keys: Mutex<HashMap<Slot, Vec<Pubkey>>>,` if the slots touch a lot of redundant accounts. I feel like ideally we would have something like a `pub uncleaned_keys: Mutex<HashMap<Pubkey, Vec<Pubkey>>>` tracking what slots a pubkey was touched in. 

1. Mark a pubkey in the map on Accounts flush if `S` is a root (should be majority of cases)
2. Otherwise keep around the `HashMap<Slot, Vec<Pubkey>>` for the slot and aggregate it into the main `uncleaned_keys` when `S` is rooted, or remove it when `S` is cleaned by `purge_slots`

sakridge:
> 
> 
> I'm slightly worried about the memory efficiency of the `pub uncleaned_keys: Mutex<HashMap<Slot, Vec<Pubkey>>>,` if the slots touch a lot of redundant accounts. I feel like alternatively we could have something like a `pub uncleaned_keys: Mutex<HashMap<Pubkey, Vec<Pubkey>>>` tracking what slots a pubkey was touched in.
> 
>     1. Mark a pubkey in the map on Accounts flush if `S` is a root (should be majority of cases)
> 
>     2. Otherwise on unrooted flush(random selection + cache overwhelmed cases) keep around the `HashMap<Slot, Vec<Pubkey>>` for the slot and aggregate it into the main `uncleaned_keys` when `S` is rooted, or if `S` is unrooted, remove it when `S` is cleaned by `purge_slots`

The Vec<Pubkey> will be only for the unique ones touched in the slot. That's because it's constructed by the `update_delta_hash` function which finds the latest updates for all the keys in a given slot.

runtime/src/accounts_db.rs
sakridge: Yea, but the critical sections of these is much smaller and it's only in the case of 0-lamport stores/updates which is much rarer than the generic case of taking locks on the index.

carllin:
> > I'm slightly worried about the memory efficiency of the `pub uncleaned_keys: Mutex<HashMap<Slot, Vec<Pubkey>>>,` if the slots touch a lot of redundant accounts. I feel like alternatively we could have something like a `pub uncleaned_keys: Mutex<HashMap<Pubkey, Vec<Pubkey>>>` tracking what slots a pubkey was touched in.
> > ```
> > 1. Mark a pubkey in the map on Accounts flush if `S` is a root (should be majority of cases)
> > 
> > 2. Otherwise on unrooted flush(random selection + cache overwhelmed cases) keep around the `HashMap<Slot, Vec<Pubkey>>` for the slot and aggregate it into the main `uncleaned_keys` when `S` is rooted, or if `S` is unrooted, remove it when `S` is cleaned by `purge_slots`
> > ```
> 
> The `Vec<Pubkey>` will be only for the unique ones touched in the slot. That's because it's constructed by the `update_delta_hash` function which finds only the latest updates for all the keys in a given slot.

Oh I was talking about between slots, i.e. it 10-20k of the same accounts are touched in multiple slots, might add up quickly

sakridge:
> Oh I was talking about between slots, i.e. it 10-20k of the same accounts are touched in multiple slots, might add up quickly

I see. I don't really want to do a lot of work on rooting since that is in a somewhat critical path.

This structure is also only for 1.4 where we don't have the caching code. I think the cache itself will be able to track these outstanding uncleaned keys so there won't be any additional memory pressure. Or when we flush the cache, we combine down the key set as you said.

I can remove it when slot is cleaned by purge_slots.

runtime/src/accounts_db.rs
sakridge: I wrote it with BTreeSet, it's slower overall and spending time constructing the btree inside the locks creates more contetion on those locks.

vec is nice that constructing the unsorted one is very cheap and sort/dedup are pretty cheap and done outside the lock.
```
vec: 23ms sort + 8ms clean + 1.5ms delta = 32ms total
scan: 227ms

btree: 0ms sort + 60ms clean + 11ms delta = 71ms total
scan: 228ms
```
Scan is a similar speed in both cases.

runtime/src/accounts_db.rs
sakridge: ok, used dashset/dashmap

runtime/src/accounts_db.rs
```diff
@@ -803,6 +882,14 @@ impl AccountsDB {
                                     );
                                 }
 
+                                let has_zero_lamport_accounts = locked_entry
+                                    .slot_list()
+                                    .iter()
+                                    .any(|(_slot, account_info)| account_info.lamports == 0);
```
ryoqun, line 888: nits: `.any((_slot, account_info) account_info.is_zero_lamport())`

runtime/src/accounts_index.rs
```diff
@@ -218,9 +223,10 @@ pub struct AccountsIndex<T> {
     spl_token_owner_index: SecondaryIndex<RwLockSecondaryIndexEntry>,
     roots_tracker: RwLock<RootsTracker>,
     ongoing_scan_roots: RwLock<BTreeMap<Slot, u64>>,
+    pub zero_lamport_accounts: DashSet<Pubkey>,
```
ryoqun, line 226: nits: `zero_lamport_pubkeys`?

ryoqun: nits: also I'd like to remove `pub` by defining a getter called `fn zero_lamport_pubkeys()`.

runtime/src/accounts_db.rs
```diff
@@ -803,6 +882,14 @@ impl AccountsDB {
...
+                                if !has_zero_lamport_accounts {
+                                    self.accounts_index.remove_zero_lamport(pubkey);
```
ryoqun, line 890: nits: how about to name this `fn` to read better in this actual callsite like `self.accounts_index.forget_as_zero_lamport()` or `self.accounts_index.unmark_as_zero_lamport()`?

runtime/src/accounts_db.rs
```diff
@@ -753,6 +776,66 @@ impl AccountsDB {
         (reclaims, dead_keys)
     }
 
+    // Construct a vec of the keys needed for cleaning from uncleaned_keys and
+    // zero_lamport_keys sets.
+    fn get_clean_keys(
```
ryoqun, line 781: nits: this code is the meat of this pr? so, `get_` sounds too humble. how about `fn construct_`/`fn create_`/`fn calculate_`.

runtime/src/accounts_db.rs
```diff
@@ -457,6 +473,11 @@ pub struct AccountsDB {
     pub cluster_type: Option<ClusterType>,
 
     pub account_indexes: HashSet<AccountIndex>,
+
+    /// Set of unique keys per slot which is used
```
ryoqun, line 477: nits: `per updated slot`

runtime/src/accounts_db.rs
```diff
@@ -457,6 +473,11 @@ pub struct AccountsDB {
...
+    /// to drive clean_accounts
+    /// Generated by get_accounts_delta_hash
+    uncleaned_keys: DashMap<Slot, Vec<Pubkey>>,
```
ryoqun, line 480: nits: `uncleaned_pubkeys`

runtime/src/accounts_db.rs
```diff
@@ -753,6 +776,66 @@ impl AccountsDB {
         (reclaims, dead_keys)
     }
 
+    // Construct a vec of the keys needed for cleaning from uncleaned_keys and
+    // zero_lamport_keys sets.
```
ryoqun, line 780: Maybe add comment like this?

```rust
// Such a vec of the pubkeys is a combined set of all of alive zero lamport pubkeys
// and all of pubkeys updated since last clean.
```

runtime/src/accounts_db.rs
ryoqun: or `clean_candidate_pubkeys`?

runtime/src/accounts_db.rs
ryoqun: or: `uncleaned_slot_pubkeys` or `uncleaned_pubkeys_by_slot`. So, I can accept `key` as local varialbles... ;)

runtime/src/accounts_db.rs
```diff
@@ -803,6 +882,14 @@ impl AccountsDB {
                                     );
                                 }
 
+                                let has_zero_lamport_accounts = locked_entry
+                                    .slot_list()
```
ryoqun, line 886: nits: how about reducing `.slot_list()` from twice to once by defining `let slot_list = ...`  to be used also at line 876?

runtime/src/accounts_db.rs
```diff
@@ -803,6 +882,14 @@ impl AccountsDB {
                                     );
                                 }
 
+                                let has_zero_lamport_accounts = locked_entry
```
ryoqun, line 885: the reason of looping here is that you need to check all entries regardless `max_clean_root`? how about commenting the subtlety?

runtime/src/accounts_index.rs
```diff
@@ -768,13 +774,20 @@ impl<T: 'static + Clone> AccountsIndex<T> {
             //  - The secondary index is never consulted as primary source of truth for gets/stores.
             //  So, what the accounts_index sees alone is sufficient as a source of truth for other non-scan
             //  account operations.
+            if account_info.is_zero_lamport() {
+                self.zero_lamport_accounts.insert(*pubkey);
+            }
```
ryoqun, line 779: strictly spekaing, this new code must be inside `w_account_enty` lock? It can't be moved out to before getting `w_account_entry`?

runtime/src/accounts_db.rs
```diff
@@ -753,6 +776,66 @@ impl AccountsDB {
...
+                })
+                .collect()
+        };
+        key_create.stop();
+        timings.key_create += key_create.as_us();
+        self.thread_pool_clean.install(|| {
+            delta_keys.par_iter().for_each(|keys| {
+                for key in keys {
+                    pubkeys.insert(*key);
+                }
```
ryoqun, line 825: maybe, can we use `pubkeys.extend()` for possible bulk write?

runtime/src/accounts_db.rs
```diff
@@ -753,6 +776,66 @@ impl AccountsDB {
...
+    fn get_clean_keys(
+        &self,
+        max_clean_root: Option<Slot>,
+        timings: &mut CleanKeyTimings,
+    ) -> Vec<Pubkey> {
+        let mut zero_lamport_key = Measure::start("zero_lamport_key");
+        let pubkeys = self.accounts_index.zero_lamport_accounts.clone();
+        timings.zero_lamport_count = pubkeys.len() as u64;
+        zero_lamport_key.stop();
+        timings.zero_lamport_key += zero_lamport_key.as_us();
```
carllin, line 790: nit: rename to `clone_zero_lamport_key`

runtime/src/accounts_db.rs
```diff
@@ -753,6 +776,66 @@ impl AccountsDB {
...
+                        Some(*slot)
+                    } else {
+                        None
+                    }
+                })
+                .collect();
+            slots
+                .into_iter()
+                .filter_map(|slot| {
+                    let keys = self.uncleaned_keys.remove(&slot).unwrap().1;
```
carllin, line 810: For some future proofing, I think we should move the `unwrap()`call to inside the `if self.accounts_index.is_root(slot)` call below. 

This is just in case in the future we decide to move the handling of `Bank::drop() -> purge_slots()` logic into a separate thread than the AccountsBackgroundService (if that starts getting backed up).

And then we can add the following comment for why the unwrap is safe:

"This function is called from `clean_accounts()`, so no roots from the `accounts_index` can be purged in between when we filtered for roots above and now. Thus because `purge_slots()` only removes unrooted slots, and `clean_accounts()` is the only logic that purges rooted slots, then this slot must still exist in the map."

runtime/src/accounts_db.rs
carllin: Maybe a comment here as well that this check is sufficient, because any "outdated" zero-lamport keys, i.e. keys that became zero-lamport but then received a non-zero value at a later slot, will eventually be purged from the index by the call to `clean_old_rooted_accounts` below, and then this check will succeed on the next clean.

runtime/src/accounts_db.rs
ryoqun: well, I tested this locally. seems I can't do this.

ryoqun (approved):
LGTM with nits

Very clever implementation! I learned a lot. :)

runtime/src/accounts_db.rs
sakridge: Yes, the zero_lamport_pubkey index is meant to contain *all* keys which may have a 0-lamport update in them, even if it is rooted or not, because it may become rooted in the future. This would then prune any entries which do not have a 0-lamport update because it was removed in some other clean or update operation.

runtime/src/accounts_index.rs
sakridge: Well, that's a good question. I was thinking it needs to be in at least a read-lock on the entry. We don't take a read-lock on the update flow, so I'm not sure it makes sense to add another lock just to update this set.

It's not great to do more work in the write critical section, but I think zero-lamport store case is relatively un-common.

sakridge: If I do it above outside the lock, then there is a race between the clean thread and the insert thread.
1. insert thread inserts the key in to zero_lamport_keys, but doesn't get into the write lock
2. clean thread grabs the read lock on the entry
3. clean thread sees no zero-lamport updates in the list and thus removes key from zero_lamport_keys
4. clean thread releases read lock
5. insert thread grabs write lock
6. insert thread inserts the 0-lamport update
7. State is bad since there is a zero lamport update for the key but it doesn't exist in the zero_lamport_keys set.

runtime/src/accounts_db.rs
```diff
@@ -753,6 +779,76 @@ impl AccountsDB {
         (reclaims, dead_keys)
     }
 
+    // Construct a vec of pubkeys for cleaning from:
+    //   uncleaned_pubkeys - the delta set of updated pubkeys in rooted slots from the last clean
+    //   zero_lamport_pubkeys - set of all alive pubkeys containing 0-lamport updates
```
ryoqun, line 784: :100:

runtime/src/accounts_db.rs
```diff
@@ -803,6 +896,16 @@ impl AccountsDB {
                                     );
                                 }
 
+                                // prune zero_lamport_pubkey set which should contain all 0-lamport
+                                // keys whether rooted or not. A 0-lamport update may become rooted
+                                // in the future.
```
ryoqun, line 901: this

ryoqun (approved):
re-LGTM with no nits

PR #14594, opened 2021-01-15
Fix program-test's CPI support

jackcmay:
#### Problem

Program-tests does not support making cross-program invocations
- Account construction doesn't match the message
- Invoke context is not passed along with pre-accounts, etc...
- Logging does not match

#### Summary of Changes

Fix the above issues through with the side effect of using unsafe code in the test framework.

Fixes #

### DIFF
--- a/program-test/src/lib.rs
+++ b/program-test/src/lib.rs
@@ -20,8 +20,9 @@ use {
     solana_sdk::{
         account::Account,
         keyed_account::KeyedAccount,
-        process_instruction::BpfComputeBudget,
-        process_instruction::{InvokeContext, MockInvokeContext, ProcessInstructionWithContext},
+        process_instruction::{
+            stable_log, BpfComputeBudget, InvokeContext, ProcessInstructionWithContext,
+        },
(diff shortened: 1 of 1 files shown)


program-test/src/lib.rs
```diff
@@ -64,7 +66,21 @@ pub fn to_instruction_error(error: ProgramError) -> InstructionError {
 }
 
 thread_local! {
-    static INVOKE_CONTEXT:RefCell<Rc<MockInvokeContext>> = RefCell::new(Rc::new(MockInvokeContext::default()));
+    static INVOKE_CONTEXT:RefCell<Option<(usize, usize)>> = RefCell::new(None);
+}
+fn set_invoke_context(new: &mut dyn InvokeContext) {
+    INVOKE_CONTEXT.with(|invoke_context| {
+        if unsafe { invoke_context.replace(Some(transmute::<_, (usize, usize)>(new))) }.is_some() {
+            panic!("Overriting invoke context!")
```
mvines, line 74:
```suggestion
            panic!("Overwriting invoke context!")
```

PR #14595, opened 2021-01-15
More generic accounts purge functions

carllin:
#### Problem
`AccountsDb::purge_keys_exact() -> AccountsDb::purge_exact() -> AccountsDb::purge_secondary_indexes_by_inner_key()` requires a `HashSet` when sometimes a slot is all that is necessary

purge_slot_cache_keys() clones `dead_slot ` HashSet every time, only needs an iterator

#### Summary of Changes
Support more generic parameters, remove clones.

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -24,6 +24,7 @@ use crate::{
         AccountIndex, AccountsIndex, Ancestors, IndexKey, IsCached, SlotList, SlotSlice,
     },
     append_vec::{AppendVec, StoredAccountMeta, StoredMeta},
+    contains::Contains,
 };
 use blake3::traits::digest::Digest;
 use dashmap::DashMap;
@@ -893,17 +894,20 @@ impl AccountsDB {
         }
     }
 
-    fn purge_keys_exact(
-        &self,
-        pubkey_to_slot_set: Vec<(Pubkey, HashSet<Slot>)>,
-    ) -> Vec<(u64, AccountInfo)> {
+    fn purge_keys_exact<'a, C: 'a>(
+        &'a self,
+        pubkey_to_slot_set: &'a [(Pubkey, C)],
+    ) -> Vec<(u64, AccountInfo)>
+    where
+        C: Contains<'a, Slot>,
+    {
         let mut reclaims = Vec::new();
         let mut dead_keys = Vec::new();
 
         for (pubkey, slots_set) in pubkey_to_slot_set {
             let is_empty = self.accounts_index.purge_exact(
                 &pubkey,
-                &slots_set,
+                slots_set,
                 &mut reclaims,
                 &self.account_indexes,
             );
@@ -1079,11 +1083,18 @@ impl AccountsDB {
         let pubkey_to_slot_set: Vec<_> = purges
             .into_iter()
             .map(|(key, (slots_list, _ref_count))| {
-                (key, slots_list.into_iter().map(|(slot, _)| slot).collect())
+                (
+                    key,
+                    slots_list
+                        .into_iter()
+                        .map(|(slot, _)| slot)
+                        .collect::<HashSet<Slot>>(),
+                )
             })
             .collect();
 
-        let reclaims = self.purge_keys_exact(pubkey_to_slot_set);
+        let reclaims = self.purge_keys_exact(&pubkey_to_slot_set);
+
         self.handle_reclaims(&reclaims, None, false, None);
 
         reclaims_time.stop();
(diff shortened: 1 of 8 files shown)


runtime/src/contains.rs
```diff
@@ -0,0 +1,49 @@
...
+    fn contains(&self, key: &T) -> bool;
+    fn contains_iter(&'a self) -> Self::Iter;
+}
+
+impl<'a, T: 'a + Eq + Hash, U: 'a> Contains<'a, T> for HashMap<T, U> {
+    type Item = &'a T;
+    type Iter = std::collections::hash_map::Keys<'a, T, U>;
+
+    fn contains(&self, key: &T) -> bool {
+        self.contains_key(key)
```
ryoqun, line 20: how about using fully qualified syntax like this for consisntent like `<HashSet<T>>::contains(self, key)`?

runtime/src/contains.rs
```diff
@@ -0,0 +1,49 @@
...
+        self.iter()
+    }
+}
+
+impl<'a, T: 'a + Eq + Hash + Copy> Contains<'a, T> for T {
+    type Item = &'a T;
+    type Iter = std::iter::Once<&'a T>;
+
+    fn contains(&self, key: &T) -> bool {
+        *key == *self
```
ryoqun, line 44: nits: iirc, we can just remove `*`s from the both operands?

runtime/src/contains.rs
```diff
@@ -0,0 +1,49 @@
+use std::{
+    borrow::Borrow,
+    cmp::Eq,
+    collections::{HashMap, HashSet},
+    hash::Hash,
+};
+
+pub trait Contains<'a, T: Eq + Hash> {
```
ryoqun, line 8: nice gadget. :)

runtime/src/accounts_db.rs
```diff
@@ -2385,22 +2396,26 @@ impl AccountsDB {
         );
     }
 
-    fn purge_slot_cache_keys(&self, dead_slot: Slot, slot_cache: SlotCache) {
+    fn purge_slot_cache_keys(
+        &self,
+        dead_slot: Slot,
+        slot_cache: SlotCache,
+    ) -> Vec<(u64, AccountInfo)> {
```
ryoqun, line 2403: no need to return? (maybe incorrect rebase cleanup because I added late-comment at the AccountsDB caching pr).

runtime/src/accounts_db.rs
```diff
@@ -3491,12 +3506,13 @@ impl AccountsDB {
             self.accounts_index.unref_from_storage(&pubkey);
         }
 
-        for slot in dead_slots.iter() {
+        let dead_slots_iter_ = dead_slots_iter.clone();
```
ryoqun, line 3509: nits: I think this is `.clone()`-ing the dead_slots itself, not the iterator; how about this?: https://github.com/solana-labs/solana/commit/ba2cc5c87845e9e0138ced8615bc30341040ab75

Anyway, `dead_slots` shouldn't that big...
 

```patch
index 9fe57b9a47..0c1808a4cd 100644
@@ -3494,8 +3494,8 @@ impl AccountsDB {
     }
 
     fn finalize_dead_slot_removal<'a>(
-        &'a self,
-        dead_slots_iter: impl IntoIterator<Item = &'a Slot> + Clone,
+        &self,
+        dead_slots_iter: impl Iterator<Item = &'a Slot> + Clone,
         purged_slot_pubkeys: HashSet<(Slot, Pubkey)>,
         mut purged_account_slots: Option<&mut AccountSlots>,
     ) {
@@ -3506,13 +3506,13 @@ impl AccountsDB {
             self.accounts_index.unref_from_storage(&pubkey);
         }
 
-        let dead_slots_iter_ = dead_slots_iter.clone();
-        for slot in dead_slots_iter {
+        for slot in dead_slots_iter.clone() {
             self.accounts_index.clean_dead_slot(*slot);
         }
+
         {
             let mut bank_hashes = self.bank_hashes.write().unwrap();
-            for slot in dead_slots_iter_ {
+            for slot in dead_slots_iter {
                 bank_hashes.remove(slot);
             }
         }
@@ -3549,7 +3549,11 @@ impl AccountsDB {
                     })
             })
         };
-        self.finalize_dead_slot_removal(dead_slots, purged_slot_pubkeys, purged_account_slots);
+        self.finalize_dead_slot_removal(
+            dead_slots.iter(),
+            purged_slot_pubkeys,
+            purged_account_slots,
+        );
         measure.stop();
         inc_new_counter_info!("clean_stored_dead_slots-ms", measure.as_ms() as usize);
     }

```

ryoqun (dismissed):
LGTM with nits

runtime/src/accounts_db.rs
carllin: ah true, thanks!

PR #14626, opened 2021-01-16
Track account writable deescalation

### LINKED ISSUE #13908: Account readwrite deescalation is not persisted in instruction execution context
#### Problem
If a program passes a readwrite account as readonly to an inner instruction, then any writes should cause an error. Since the runtime doesn't track changes to write deescalation, it allows those instructions to succeed but will not return those state changes to the caller program due to this check:

https://github.com/solana-labs/solana/blob/10c81a2448c1db49e010afbbed94ac48209cba48/programs/bpf_loader/src/syscalls.rs#L1422

#### Proposed Solution
Temporarily deescalate `is_writable` in `PreAccounts` for inner instructions.

jackcmay:
#### Problem

If a program passes a readwrite account as readonly to an inner instruction, then any writes should cause an error. Since the runtime doesn't track changes to write deescalation, it allows those instructions to succeed but will not return those state changes to the caller program

#### Summary of Changes

Track the writable flag through the call flow

Fixes #13908


jackcmay:
@mvines Looks like SPL Feature Proposal breaks with these changes.  Is it doing trying to write to accounts that have been passed as read-only?

jackcmay:
TODO: This also needs to be gated

runtime/src/message_processor.rs
```diff
@@ -577,6 +587,11 @@ impl MessageProcessor {
             .iter()
             .map(|seeds| Pubkey::create_program_address(&seeds, caller_program_id))
             .collect::<Result<Vec<_>, solana_sdk::pubkey::PubkeyError>>()?;
+        let mut caller_privileges = keyed_accounts
+            .iter()
+            .map(|keyed_account| keyed_account.is_writable())
+            .collect::<Vec<bool>>();
+        caller_privileges.insert(0, false);
```
jstarry, line 594: @jackcmay why insert `false` here? And isn't `caller_privileges` supposed to be aligned with `accounts` rather than `keyed_accounts`?

jackcmay: IIRC it holds the space of the program, which isn't part of the `keyed_accounts` and the program is not writeable

jstarry: Got it, it doesn't look like that's the case anymore on latest master so I have a fix here: https://github.com/solana-labs/solana/pull/18616

PR #14631, opened 2021-01-16
feature gates turbine retransmit peers patch

behzadnouri:
#### Problem
https://github.com/solana-labs/solana/pull/14565 needs to be feature-gated before it is enabled on mainnet,
otherwise the cluster will be out of sync on how they compute turbine retransmit peers.

#### Summary of Changes
added a feature gate for the patch.

### DIFF
--- a/core/src/retransmit_stage.rs
+++ b/core/src/retransmit_stage.rs
@@ -252,7 +252,7 @@ fn enable_turbine_retransmit_peers_patch(shred_slot: Slot, root_bank: &Bank) -> 
             let epoch_schedule = root_bank.epoch_schedule();
             let feature_epoch = epoch_schedule.get_epoch(feature_slot);
             let shred_epoch = epoch_schedule.get_epoch(shred_slot);
-            feature_epoch + 1 < shred_epoch
+            feature_epoch < shred_epoch
         }
     }
 }
--- a/core/src/retransmit_stage.rs
+++ b/core/src/retransmit_stage.rs
@@ -37,7 +37,7 @@ use std::{
     collections::hash_set::HashSet,
     collections::{BTreeMap, HashMap},
     net::UdpSocket,
-    ops::DerefMut,
+    ops::{Deref, DerefMut},
     sync::atomic::{AtomicBool, AtomicU64, Ordering},
     sync::mpsc::channel,
     sync::mpsc::RecvTimeoutError,
@@ -49,7 +49,6 @@ use std::{
 
 const MAX_DUPLICATE_COUNT: usize = 2;
 const DEFAULT_LRU_SIZE: usize = 10_000;
-const BANKS_LRU_CACHE_CAP: usize = 100;
 
 // Limit a given thread to consume about this many packets so that
 // it doesn't pull up too much work.
@@ -243,34 +242,17 @@ fn check_if_already_received(
 }
 
 // Returns true if turbine retransmit peers patch (#14565) is enabled.
-fn enable_turbine_retransmit_peers_patch(
-    shred_slot: Slot,
-    bank_forks: &RwLock<BankForks>,
-    cache: &mut LruCache<Slot, Arc<Bank>>,
-) -> bool {
-    let bank: Arc<Bank> = match cache.get(&shred_slot) {
-        Some(bank) => bank.clone(),
-        None => {
-            let bank_forks = bank_forks.read().unwrap();
-            match bank_forks.get(shred_slot) {
-                Some(bank) => {
-                    cache.put(shred_slot, bank.clone());
-                    bank.clone()
-                }
-                None => bank_forks.root_bank(),
-            }
-        }
-    };
-    let feature_slot = bank
+fn enable_turbine_retransmit_peers_patch(shred_slot: Slot, root_bank: &Bank) -> bool {
+    let feature_slot = root_bank
         .feature_set
         .activated_slot(&feature_set::turbine_retransmit_peers_patch::id());
     match feature_slot {
         None => false,
         Some(feature_slot) => {
-            let epoch_schedule = bank.epoch_schedule();
+            let epoch_schedule = root_bank.epoch_schedule();
             let feature_epoch = epoch_schedule.get_epoch(feature_slot);
             let shred_epoch = epoch_schedule.get_epoch(shred_slot);
-            feature_epoch < shred_epoch
+            feature_epoch + 1 < shred_epoch
         }
     }
 }
@@ -287,7 +269,6 @@ fn retransmit(
     epoch_stakes_cache: &RwLock<EpochStakesCache>,
     last_peer_update: &AtomicU64,
     shreds_received: &Mutex<ShredFilterAndHasher>,
-    banks_cache: &mut LruCache<Slot, Arc<Bank>>,
 ) -> Result<()> {
     let timer = Duration::new(1, 0);
     let r_lock = r.lock().unwrap();
@@ -305,7 +286,10 @@ fn retransmit(
     drop(r_lock);
 
     let mut epoch_fetch = Measure::start("retransmit_epoch_fetch");
-    let r_bank = bank_forks.read().unwrap().working_bank();
+    let (r_bank, root_bank) = {
+        let bank_forks = bank_forks.read().unwrap();
+        (bank_forks.working_bank(), bank_forks.root_bank())
+    };
     let bank_epoch = r_bank.get_leader_schedule_epoch(r_bank.slot());
     epoch_fetch.stop();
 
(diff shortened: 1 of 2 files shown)


core/src/retransmit_stage.rs
```diff
@@ -209,49 +213,69 @@ pub type ShredFilter = LruCache<(Slot, u32, bool), Vec<u64>>;
...
+            Some(shred.0)
+        }
+    }
+}
 
-            false
+// Returns true if turbine retransmit peers patch (#14565) is enabled.
+fn enable_turbine_retransmit_peers_patch(bank_forks: &RwLock<BankForks>, shred_slot: Slot) -> bool {
+    let bank = {
+        let bank_forks = bank_forks.read().unwrap();
```
carllin, line 247: instead of grabbing this `bank_forks` lock everytime, could we save the `feature_slot` and clone out the `epoch_schedule` to be used for `get_epoch()`

behzadnouri: the constraint is that we need to look up the bank for the `shred_slot`, where the `shred_slot` may change from one packet to the other.
I updated the code by adding a local cache for `slot -> bank` mapping to avoid locking `bank_forks` each time.

carllin: got it, the bank needs to be continually updated to check for feature updates. I think we can do a little better than checking for a bank every slot though, since features are only updated at epoch boundaries: https://github.com/solana-labs/solana/blob/master/runtime/src/bank.rs#L1073.  

So If we clone out the `epoch_schedule` on startup: https://github.com/solana-labs/solana/pull/14631/files#diff-8d73ca7d46b0b2dba59642e0e183fbb6127c8459ade483df4ac729390519e6f6R266 which doesn't change from bank to bank, then we can detect when a `shred_slot` is in a new epoch and then only grab a new bank then.

behzadnouri: The concern with relying on:
* the `epoch_schedule` not changing from bank to bank.
* features getting activated only at epoch boundaries.

is that above two are some internal implementation details in another crate. They are not part of the api, and could easily change by someone unaware of the code here. So it opens door to bugs in the future.

This function locks `bank_forks` once per *unique* shred-slot but besides that it is already O(1). I feel like we at least need some evidence that this is going to be a bottleneck, before optimizing it further by relying on internals of another crate.

core/src/retransmit_stage.rs
```diff
@@ -330,19 +362,22 @@ fn retransmit(
...
-            let (my_index, shuffled_stakes_and_index) = ClusterInfo::shuffle_peers_and_index(
+            let (my_index, mut shuffled_stakes_and_index) = ClusterInfo::shuffle_peers_and_index(
                 &my_id,
                 &r_epoch_stakes_cache.peers,
                 &r_epoch_stakes_cache.stakes_and_index,
                 packet.meta.seed,
             );
             peers_len = cmp::max(peers_len, shuffled_stakes_and_index.len());
+            // Until the patch is activated, do the old buggy thing.
+            if !enable_turbine_retransmit_peers_patch(shred_slot, bank_forks, &mut banks_cache) {
```
sakridge, line 378: How is banks_cached pruned of old banks?

behzadnouri: It is a local variable to `fn retransmit`, so it is dropped at function exit.
And its size is bounded by `MAX_PACKET_BATCH_SIZE`.

sakridge: I see. I'm not sure the batches will be all that big, the cache may not be very effective at this level.

core/src/retransmit_stage.rs
```diff
@@ -209,49 +213,76 @@ pub type ShredFilter = LruCache<(Slot, u32, bool), Vec<u64>>;
...
+        .entry(shred_slot)
+        .or_insert_with(|| {
+            let bank_forks = bank_forks.read().unwrap();
+            match bank_forks.get(shred_slot) {
+                Some(bank) => bank.clone(),
+                None => bank_forks.root_bank(),
+            }
+        })
+        .clone();
+    let feature_slot = bank
```
mvines, line 260: On testnet, this feature is effectively active now right?  If so then this appears to turn the feature off for testnet nodes as they update to this patch, so I think we'd want to either add a check for !Testnet here unfortunately

behzadnouri: so i was thinking not to backport this to testnet,
or when backporting modify it so that it is effectively a no-op

mvines: > so i was thinking not to backport this to testnet,

What does that mean, not backport to the v1.5 branch?  That doesn't work since the 1.5 branch has to work with both mainnet-beta and testnet

core/src/retransmit_stage.rs
behzadnouri: I switched to an `LruCache` preserved between calls to `fn retransmit(...)`.

One question is that, what to do if `bank_forks.get(shred_slot)` returns `None`:
* I am using `bank_forks.root_bank()` as the fallback. Would that make sense?
* Should it cache that fallback bank for the `shred_slot`?
I chose not to cache that, because following lookups will still use the fallback bank even if `bank_forks.get(shred_slot)` no longer is `None`. This is probably safer approach, but can potentially cause more locking on `bank_forks`.

core/src/retransmit_stage.rs
behzadnouri: I see.

How about just rolling it out to testnet and taking the risk that it **might** break testnet during the upgrade and we may need a restart?
The `DATA_PLANE_FANOUT` is 200 and so there is a good chance that the cluster keeps working with or without https://github.com/solana-labs/solana/pull/14565 during the course of upgrade (because there will be only 2 layers, and there will be a spanning tree nevertheless).

The advantage is that we can test dynamically enabling the feature on testnet before doing so on mainnet.

mvines: Sure.  Restarting testnet is always a good exercise :)

core/src/retransmit_stage.rs
sakridge: I think you would want to find a bank with the same epoch as the `shred.slot`. `working_bank` would probably be more likely, because `root_bank` will trail the epoch start by some amount which is takes to root the banks coming into the epoch.

But to be accurate, you could save a bank polled from `working_bank` when the epoch changes and then have both epoch N, N+1 banks to use to compute it.

behzadnouri: I am not sure I understand your suggestion. Can you please clarify?

Would it make sense to go through all banks in bank_forks:
https://github.com/solana-labs/solana/blob/d63b2baf0/runtime/src/bank_forks.rs#L46
to find one with the same `epoch` as `shred.slot`?
And, if so, then
* Which one should it be used if there are multiple matches?
* What to do if there are no matches?

sakridge: > Would it make sense to go through all banks in bank_forks?

Yea, that's an option.

> Which one should it used if there are multiple matches?

It shouldn't matter, any bank in the epoch should give the same answer, first one it finds.

> What to do if there are no matches?

Using `root_bank` should be ok then.

core/src/retransmit_stage.rs
```diff
@@ -305,7 +286,10 @@ fn retransmit(
     drop(r_lock);
 
     let mut epoch_fetch = Measure::start("retransmit_epoch_fetch");
-    let r_bank = bank_forks.read().unwrap().working_bank();
+    let (r_bank, root_bank) = {
```
carllin, line 289: nit: rename `r_bank` to `working_bank`

core/src/retransmit_stage.rs
```diff
@@ -243,34 +242,17 @@ fn check_if_already_received(
...
         .activated_slot(&feature_set::turbine_retransmit_peers_patch::id());
     match feature_slot {
         None => false,
         Some(feature_slot) => {
-            let epoch_schedule = bank.epoch_schedule();
+            let epoch_schedule = root_bank.epoch_schedule();
             let feature_epoch = epoch_schedule.get_epoch(feature_slot);
             let shred_epoch = epoch_schedule.get_epoch(shred_slot);
-            feature_epoch < shred_epoch
+            feature_epoch + 1 < shred_epoch
```
carllin, line 255: Sorry, I think I read the discord comment wrong, I think  this should be a `<=`, or change `feature_epoch + 1` to `feature_epoch + 2`.


So if the feature activates on epoch `N`, then  we want to start returning `true` here on epoch  `N + 1`

carllin, line 255: Sorry, I think I read the discord comment wrong, I think  this should be a `<=`, or change `feature_epoch + 1` to `feature_epoch + 2`.

So if the feature activates on epoch `N`, then  we want to start returning `true` here on epoch  `N + 1`

sakridge (approved):

core/src/retransmit_stage.rs
carllin, line 255: Sorry, I think I read the discord comment wrong, I think  this should be a `<=`

So if the feature activates on epoch `N`, then  we want to start returning `true` here on epoch  `N + 1`

behzadnouri: updated to `feature_epoch < shred_epoch`

carllin (approved):

PR #14650, opened 2021-01-18
modify merge to use faster hash lookup and replace/insert

jeffwashington:
#### Problem
Speed up hash calculations.

#### Summary of Changes
HashMap has entries api designed for faster lookup and replace/insert.

Fixes #


jeffwashington:
this code was deleted by carl, so this pr is no longer necessary.

jeffwashington:
@carllin - fyi, we arrived at a similar conclusion that the entry api is a faster way to do hashmap things.

carllin:
@jeffwashington, hehe, awesome!

PR #14596, opened 2021-01-15
Clean accounts cache before flush

carllin:
#### Problem
Accounts cache flushes a lot of outdated keys from earlier roots that have already been updated by later roots, which bloats AppendVecs

#### Summary of Changes
Note, builds on top of some refactoring done here: https://github.com/solana-labs/solana/pull/14595

Remove outdated updates from earlier roots

Results: Significantly less bloat in AppendVecs, which means much smaller shrink and flush times (note before 18:00 is with this change, after is without these changes)

Screen Shot 2021-01-14 at 6 31 57 PM
Screen Shot 2021-01-14 at 6 31 45 PM

Can view the number of bytes/accounts removed from the cache here: 

https://metrics.solana.com:8888/sources/3/chronograf/data-explorer?query=SELECT%20mean%28%22account_bytes_saved%22%29%20AS%20%22mean_account_bytes_saved%22%2C%20mean%28%22num_accounts_saved%22%29%20AS%20%22mean_num_accounts_saved%22%20FROM%20%22mainnet-beta%22.%22autogen%22.%22accounts_db-flush_accounts_cache%22%20WHERE%20time%20%3E%20now%28%29%20-%2012h%20GROUP%20BY%20time%28%3Ainterval%3A%29%20FILL%28null%29
Fixes #

PR #14652, opened 2021-01-19
Partial accounts clean

sakridge:
#### Problem

Full accounts scan clean takes >10s and growing with account size

#### Summary of Changes

Clean less keys by tracking the two cases:
* Touched keys per slot in uncleaned_keys derived from
accounts delta hash operation.
* Set of keys with any zero-lamport updates.

Fixes #


runtime/src/accounts_db.rs
```diff
@@ -617,6 +633,15 @@ pub struct AccountsDB {
...
     pub caching_enabled: bool,
+
+    /// Set of unique keys per slot which is used
+    /// to drive clean_accounts
+    /// Generated by get_accounts_delta_hash
+    uncleaned_keys: DashMap<Slot, Vec<Pubkey>>,
+
+    /// Squashed down uncleaned_keys and the highest slot it
+    /// is uncleaned at
+    squashed_uncleaned_keys: RwLock<(HashSet<Pubkey>, u64)>,
```
ryoqun, line 644: nits: `Slot`?

runtime/src/accounts_db.rs
```diff
@@ -2454,6 +2589,7 @@ impl AccountsDB {
             // It should not be possible that a slot is neither in the cache or storage. Even in
             // a slot with all ticks, `Bank::new_from_parent()` immediately stores some sysvars
             // on bank creation.
+            self.uncleaned_keys.remove(remove_slot);
```
ryoqun, line 2592: nits: how about adding blank line above to make it clear the above comment isn't explaining this line?

runtime/src/accounts_db.rs
```diff
@@ -921,6 +949,96 @@ impl AccountsDB {
...
 
+    fn squash_uncleaned_keys(&self, max_slot: Slot) {
+        let (keys, mut slots) = self.collect_uncleaned_keys_to_slot(max_slot);
+        if slots.is_empty() {
+            return;
+        }
+
+        slots.sort_unstable();
+        let max_slot = *slots.last().unwrap();
+        let mut squashed_keys = self.squashed_uncleaned_keys.write().unwrap();
```
ryoqun, line 960: how about destructuring?

```rust
let (mut pubkeys, root) =  self.squashed_uncleaned_keys.write().unwrap();
```

runtime/src/accounts_db.rs
```diff
@@ -921,6 +949,96 @@ impl AccountsDB {
...
+        for key in keys.into_iter().flatten() {
+            squashed_keys.0.insert(key);
```
ryoqun, line 962: extend can be used?

runtime/src/accounts_db.rs
```diff
@@ -921,6 +949,96 @@ impl AccountsDB {
...
+    ) -> Vec<Pubkey> {
+        let mut zero_lamport_key = Measure::start("zero_lamport_key");
+        let mut pubkeys = self.accounts_index.zero_lamport_accounts.clone();
+        timings.zero_lamport_count = pubkeys.len() as u64;
+        zero_lamport_key.stop();
+        timings.zero_lamport_key += zero_lamport_key.as_us();
+
+        let mut key_create = Measure::start("key_create");
+        let max_slot = max_clean_root.unwrap_or_else(|| self.accounts_index.max_root());
+        let delta_keys: Vec<_> = { self.collect_uncleaned_keys_to_slot(max_slot).0 };
```
ryoqun, line 1011: maybe just like this without `{...}`?

```rust
let (delta_keys, _slots) = self.collect_uncleaned_keys_to_slot(max_slot);
```

runtime/src/accounts_db.rs
```diff
@@ -921,6 +949,96 @@ impl AccountsDB {
...
+                } else {
+                    None
+                }
+            })
+            .collect();
+        (
+            slots
+                .iter()
+                .filter_map(|slot| {
+                    let keys = self.uncleaned_keys.remove(slot).unwrap().1;
```
ryoqun, line 984: ok, lack of drain_filter and avoidance of `.clone()` causing this involved coding... ;)

runtime/src/accounts_db.rs
```diff
@@ -921,6 +949,96 @@ impl AccountsDB {
...
+                    if self.accounts_index.is_root(*slot) {
+                        Some(keys)
+                    } else {
+                        None
+                    }
+                })
+                .collect(),
+            slots,
```
ryoqun, line 992: how about returning only `real_max_slot` while maintaining it at above loops?
- less clustter api
- no need to do `sort_unstable()` at the caller site.

runtime/src/accounts_db.rs
```diff
@@ -921,6 +949,96 @@ impl AccountsDB {
...
+        self.thread_pool_clean.install(|| {
+            delta_keys.par_iter().for_each(|keys| {
+                for key in keys {
+                    pubkeys.insert(*key);
+                }
+            });
+        });
+
+        {
+            let mut squashed_uncleaned_keys = self.squashed_uncleaned_keys.write().unwrap();
```
ryoqun, line 1024: nits: `let (mut squashed_uncleaned_keys, squashed_root) = ...`

ryoqun (dismissed):
LGTM with nits

Very clever implementation! I learned a lot. :)

runtime/src/accounts_db.rs
```diff
@@ -981,6 +1096,14 @@ impl AccountsDB {
                                     );
                                 }
 
+                                let has_zero_lamport_accounts = locked_entry
```
ryoqun, line 1099: nits: carry over this comment? https://github.com/solana-labs/solana/pull/14592/files#r562004860

Yeah, what a burden to maintain 2 branches... Thanks for doing so.

runtime/src/accounts_db.rs
```diff
@@ -931,6 +960,96 @@ impl AccountsDB {
...
+                    }
+                })
+                .collect(),
+            max_slot_in_uncleaned_pubkeys,
+        )
+    }
+
+    // Construct a vec of the keys needed for cleaning from uncleaned_keys and
+    // zero_lamport_keys sets.
+    fn get_clean_keys(
```
ryoqun, line 1006: nits: also this name change and some comments are missing?

runtime/src/accounts_db.rs
sakridge: yes, added.

runtime/src/accounts_db.rs
```diff
@@ -2862,6 +3005,8 @@ impl AccountsDB {
             self.accounts_cache.set_max_flush_root(*root);
         }
 
+        self.squash_uncleaned_pubkeys(self.accounts_cache.fetch_max_flush_root());
```
ryoqun, line 3008: for the record, I even considered for the case of no `--accounts-db-caching-enabled`. it seems to work because squashing is optional. :)

sakridge: Yea, squashing is an optional memory-saving optimization.

runtime/src/accounts_db.rs
```diff
@@ -981,6 +1100,17 @@ impl AccountsDB {
                                     );
                                 }
 
+                                // prune zero_lamport_pubkey set which should contain all 0-lamport
+                                // keys whether rooted or not. A 0-lamport update may become rooted
+                                // in the future.
+                                let has_zero_lamport_accounts = locked_entry
+                                    .slot_list()
```
ryoqun, line 1107: nits: also I'd like to have the `let slot_list = ...` nicety for our prosperity. :)

runtime/src/accounts_db.rs
sakridge: I didn't get it to work. The `write().unwrap()` returns RwLockGuard which I don't think can be directly destructured like that. Maybe there is another syntax, but you're kind of operating on the reference to the lock guard, so you can't copy items out to work on them.

runtime/src/accounts_db.rs
ryoqun: same as this: https://github.com/solana-labs/solana/pull/14652#discussion_r562024379

ryoqun (approved):
re-LGTM with only 1 nit

runtime/src/accounts_db.rs
```diff
@@ -931,6 +960,100 @@ impl AccountsDB {
...
+            delta_keys.par_iter().for_each(|keys| {
+                for key in keys {
+                    pubkeys.insert(*key);
+                }
+            });
+        });
+
+        {
+            let mut squashed_uncleaned_pubkeys = self.squashed_uncleaned_pubkeys.write().unwrap();
+            if squashed_uncleaned_pubkeys.1 <= max_slot {
```
carllin, line 1038: @ryoqun, @sakridge A few thoughts I had while rebasing https://github.com/solana-labs/solana/pull/14596. 

1) Since `max_slot` is constrained by the current scan root, but the cache is always flushing all roots (so `squashed_uncleaned_pubkeys.1` is progressing regardless of the scan root) , is it possible that continuous scans could constantly trigger this condition and prevent clean from happening()? I was thinking maybe we need like a `Pubkey -> Slot` map to see which pubkeys are available for cleaning.

2) Since `squashed_uncleaned_pubkeys` is almost always is *only* updated on `force_flush_accounts_cache()` of the roots, which happens before clean_accounts, is there any reason to squash in the flush instead of here? 

One  reason I could think of is in the exceptional case where `flush_accounts_cache_if_needed()`  flushes the roots because more than  the limit number of slots in the cache, and we might want to condense memory usage there. But then of course problem 1) above is still an issue.

3) https://github.com/solana-labs/solana/pull/14652/files#diff-1090394420d51617f3233275c2b65ed706b35b53b115fe65f82c682af8134a6fR3567-R3570 I think this may also be able to be moved from blocking `freeze()` into `flush_accounts_cache()` as well, since we know the pubkeys at flushing time as well, and can do some redundant cleaning based on later slots in the cache (can be added to my PR: https://github.com/solana-labs/solana/pull/14596).

sakridge: 1. Yea, I think it's possible it prevents it. Another option is just to be more conservative with what root it squashes to. Maybe squash in batches of 10-50 slots.
2. Well if we only squash right before `clean_accounts`, we don't need `squashed_uncleaned_pubkeys` at all. We just coalesce from the `uncleaned_pubkeys`. The squash from `flush_accounts_if_needed` was to lower the memory usage of the dirty key set by coalescing multiple slots.
3. Yea, that sounds good.

carllin:
@sakridge @ryoqun hmmm seems like this PR , commit c89f5e28b7833fc04e951bcff6df7a0b098976a4 on 1.5 broke the clean and the snapshots are blowing up in size:

Before this PR (900MB - 1GB):
Screen Shot 2021-01-22 at 1 45 16 AM

After this PR (starts around 3.5 GB, then continues to grow aggressively in size):
Screen Shot 2021-01-22 at 1 45 25 AM

ryoqun:
> @sakridge @ryoqun hmmm seems like this PR , commit [c89f5e2](https://github.com/solana-labs/solana/commit/c89f5e28b7833fc04e951bcff6df7a0b098976a4) on 1.5 broke the clean and the snapshots are blowing up in size:
> 
> Before this PR (900MB - 1GB):
> Screen Shot 2021-01-22 at 1 45 16 AM
> 
> After this PR (starts around 3.5 GB, then continues to grow aggressively in size):
> Screen Shot 2021-01-22 at 1 45 25 AM

@carllin Oh, sad... Thanks for spotting this! Considering normally mainnet-beta snapshot isn't around 1G without `--accounts-db-caching-enabled`, this bloat is happening with caching enabled?

mvines:
Bloat appears to be happening with caching enabled.  I'm confirming now but if so we should probably back out this PR for now, if it can't be fixed quickly

PR #14653, opened 2021-01-19
Prevent the invoke and upgrade of programs in the same tx batch

jackcmay:
#### Problem

Upgradeable programs cannot be invoked and upgraded in the same transaction batch but the programs are not locked as writeable

#### Summary of Changes

Identify upgrade instructions in the transaction batch and reject any single transaction that attempts to both upgrade and invoke and also lock any programs in the batch that are being upgraded.

Fixes #

### DIFF
--- a/programs/bpf_loader/src/lib.rs
+++ b/programs/bpf_loader/src/lib.rs
@@ -26,7 +26,7 @@ use solana_sdk::{
     bpf_loader_upgradeable::{self, UpgradeableLoaderState},
     clock::Clock,
     entrypoint::SUCCESS,
-    feature_set::bpf_compute_budget_balancing,
+    feature_set::{bpf_compute_budget_balancing, prevent_upgrade_and_invoke},
     instruction::InstructionError,
     keyed_account::{from_keyed_account, next_keyed_account, KeyedAccount},
     loader_instruction::LoaderInstruction,
@@ -469,6 +469,12 @@ fn process_loader_upgradeable_instruction(
             if !program.executable()? {
                 log!(logger, "Program account not executable");
                 return Err(InstructionError::AccountNotExecutable);
+            }
+            if !program.is_writable()
+                && invoke_context.is_feature_active(&prevent_upgrade_and_invoke::id())
+            {
+                log!(logger, "Program account not writeable");
+                return Err(InstructionError::InvalidArgument);
             }
             if &program.owner()? != program_id {
                 log!(logger, "Program account not owned by loader");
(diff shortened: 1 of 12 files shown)


mvines:
@jackcmay - could we just require the program account be writable here, and enforce that in Upgrade instruction handling?
https://github.com/solana-labs/solana/blob/58487c636050cc51a8e8dfe68102063e37e48d51/sdk/program/src/loader_upgradeable_instruction.rs#L89

jackcmay:
Yeah, that would probably work too, more efficient but weird 😜

mvines:
> Yeah, that would probably work too, more efficient but weird 😜

Why weird?   Touching the core runtime to do this is pretty scary.   Requiring a write lock the way I propose avoids special cases

jackcmay:
It's weird in the first place that we are pulling an account not specified in the tx so there's that.  But requiring an executable to be loaded as writeable is non-intuitive and being enforced by the loader moves some of the knowledge away from accounts load/lock code.  Don't get me wrong, I think it's an acceptable solution and comes with less overhead.

sakridge:
> 
> 
> > Yeah, that would probably work too, more efficient but weird 😜
> 
> Why weird? Touching the core runtime to do this is pretty scary. Requiring a write lock the way I propose avoids special cases

Yea. I like this idea to just require it as writable.

bennofs (approved):
Looks fine to me. I agree that loading an account not specified in the TX sounds scary though, I'll think a little bit more about that.

jackcmay:
@mvines @sakridge Ready for re-review

programs/bpf_loader/src/lib.rs
```diff
@@ -470,6 +470,10 @@ fn process_loader_upgradeable_instruction(
                 log!(logger, "Program account not executable");
                 return Err(InstructionError::AccountNotExecutable);
             }
+            if !program.is_writable() {
```
mvines, line 473: We'll need to gate this.  Otherwise there's a window between 33%-66% of stake upgrade where you can trivially break consensus by upgrading a program with an older cli

mvines (commented):
I love there's no runtime/ changes anymore.  Looks good except for the need for a feature gate :(

sdk/src/feature_set.rs
```diff
@@ -179,6 +183,7 @@ lazy_static! {
         (abort_on_all_cpi_failures::id(), "Abort on all CPI failures"),
         (use_loaded_executables::id(), "Use loaded executable accounts"),
         (turbine_retransmit_peers_patch::id(), "turbine retransmit peers patch #14631"),
+        (prevent_upgrade_and_invoke::id(), "Prevent upgrade and invoke in same tx batch"),
```
mvines, line 186: maybe this?  nbd though
```suggestion
        (prevent_upgrade_and_invoke::id(), "require writeable program account during upgrade #14653"),
```

sdk/program/src/loader_upgradeable_instruction.rs
```diff
@@ -86,7 +86,7 @@ pub enum UpgradeableLoaderInstruction {
     ///
     /// # Account references
     ///   0. [writable] The ProgramData account.
-    ///   1. [] The Program account.
+    ///   1. [writable] The Program account.
     ///   2. [Writable] The Buffer account where the program data has been
```
mvines, line 90:
```suggestion
    ///   2. [writable] The Buffer account where the program data has been
```

mvines (approved):

sakridge (approved):
lgtm

PR #14657, opened 2021-01-19
Configure Bigtable's timeout, enabling by default

ryoqun:
#### Problem

There is two known indefinitely-processing places which can called from the RPC service codepath (getConfirmedBlocks)

#### Summary of Changes

Introduce timeouts.

### DIFF
--- a/core/src/rpc.rs
+++ b/core/src/rpc.rs
@@ -116,7 +116,7 @@ pub struct JsonRpcConfig {
     pub max_multiple_accounts: Option<usize>,
     pub account_indexes: HashSet<AccountIndex>,
     pub rpc_threads: usize,
-    pub rpc_timeout: Option<Duration>,
+    pub rpc_bigtable_timeout: Option<Duration>,
 }
 
 #[derive(Clone)]
@@ -740,7 +740,12 @@ impl JsonRpcRequestProcessor {
                         bigtable_blocks.retain(|&slot| slot <= end_slot);
                         bigtable_blocks
                     })
-                    .unwrap_or_else(|_| vec![]));
+                    .map_err(|_| {
+                        Error::invalid_params(
+                            "BigTable query failed (maybe timeout due to too large range?)"
+                                .to_string(),
+                        )
+                    })?);
             }
         }
 
--- a/core/src/rpc.rs
+++ b/core/src/rpc.rs
@@ -81,6 +81,7 @@ use std::{
         mpsc::{channel, Receiver, Sender},
         Arc, Mutex, RwLock,
     },
+    time::Duration,
 };
 use tokio::runtime;
 
@@ -115,6 +116,7 @@ pub struct JsonRpcConfig {
     pub max_multiple_accounts: Option<usize>,
     pub account_indexes: HashSet<AccountIndex>,
     pub rpc_threads: usize,
+    pub rpc_timeout: Option<Duration>,
 }
 
 #[derive(Clone)]
--- a/storage-bigtable/src/bigtable.rs
+++ b/storage-bigtable/src/bigtable.rs
@@ -6,6 +6,7 @@ use crate::{
     root_ca_certificate,
 };
 use log::*;
+use std::time::Duration;
 use thiserror::Error;
 use tonic::{metadata::MetadataValue, transport::ClientTlsConfig, Request};
 
@@ -106,7 +107,11 @@ impl BigTableConnection {
     ///
     /// The BIGTABLE_EMULATOR_HOST environment variable is also respected.
     ///
-    pub async fn new(instance_name: &str, read_only: bool) -> Result<Self> {
+    pub async fn new(
+        instance_name: &str,
+        read_only: bool,
+        timeout: Option<Duration>,
+    ) -> Result<Self> {
         match std::env::var("BIGTABLE_EMULATOR_HOST") {
             Ok(endpoint) => {
                 info!("Connecting to bigtable emulator at {}", endpoint);
@@ -135,19 +140,33 @@ impl BigTableConnection {
                     instance_name
                 );
 
+                let endpoint = {
+                    let endpoint =
+                        tonic::transport::Channel::from_static("https://bigtable.googleapis.com")
+                            .tls_config(
+                            ClientTlsConfig::new()
+                                .ca_certificate(
+                                    root_ca_certificate::load().map_err(Error::CertificateError)?,
+                                )
+                                .domain_name("bigtable.googleapis.com"),
+                        )?;
+
+                    if let Some(timeout) = timeout {
+                        if read_only {
+                            // apply timeout only when thre is no write queries!
+                            endpoint.timeout(timeout)
+                        } else {
+                            warn!("BigTableConnection's requested timeout configuration is ignored because it's not read-only.");
+                            endpoint
+                        }
+                    } else {
+                        endpoint
+                    }
+                };
+
                 Ok(Self {
                     access_token: Some(access_token),
-                    channel: tonic::transport::Channel::from_static(
-                        "https://bigtable.googleapis.com",
-                    )
-                    .tls_config(
-                        ClientTlsConfig::new()
-                            .ca_certificate(
-                                root_ca_certificate::load().map_err(Error::CertificateError)?,
-                            )
-                            .domain_name("bigtable.googleapis.com"),
-                    )?
-                    .connect_lazy()?,
+                    channel: endpoint.connect_lazy()?,
                     table_prefix,
                 })
             }
--- a/storage-bigtable/src/bigtable.rs
+++ b/storage-bigtable/src/bigtable.rs
@@ -152,13 +152,7 @@ impl BigTableConnection {
                         )?;
 
                     if let Some(timeout) = timeout {
-                        if read_only {
-                            // apply timeout only when thre is no write queries!
-                            endpoint.timeout(timeout)
-                        } else {
-                            warn!("BigTableConnection's requested timeout configuration is ignored because it's not read-only.");
-                            endpoint
-                        }
+                        endpoint.timeout(timeout)
                     } else {
                         endpoint
                     }
--- a/storage-bigtable/src/bigtable.rs
+++ b/storage-bigtable/src/bigtable.rs
@@ -6,7 +6,7 @@ use crate::{
     root_ca_certificate,
 };
 use log::*;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 use thiserror::Error;
 use tonic::{metadata::MetadataValue, transport::ClientTlsConfig, Request};
 
@@ -69,6 +69,9 @@ pub enum Error {
 
     #[error("RPC error: {0}")]
     RpcError(tonic::Status),
+
+    #[error("Timeout error")]
+    TimeoutError,
 }
 
 impl std::convert::From<std::io::Error> for Error {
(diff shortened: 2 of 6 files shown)


ryoqun:
Hopefully, this is the last timeout issue.

ryoqun:
good news and bad news:

timeout is working but shouldn't return empty.... we're close. :)

```
$ date && curl -v -H 'Content-Type: application/json' -d '{"jsonrpc":"2.0","id":1,"method":"getConfirmedBlocks","params":[50000000,50040001]}' http://localhost:8899 ; date
Tue Jan 19 04:15:35 UTC 2021
*   Trying 127.0.0.1:8899...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8899 (#0)
> POST / HTTP/1.1
> Host: localhost:8899
> User-Agent: curl/7.68.0
> Accept: */*
> Content-Type: application/json
> Content-Length: 83
> 
* upload completely sent off: 83 out of 83 bytes
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< content-type: application/json; charset=utf-8
< content-length: 37
< date: Tue, 19 Jan 2021 04:15:35 GMT
< 
{"jsonrpc":"2.0","result":[],"id":1}
* Connection #0 to host localhost left intact
Tue Jan 19 04:16:05 UTC 2021

```

ryoqun:
@mvines could you take a quite early preview on this? I want to sneak this into the next v1.4 version to fix all of timeout issues.

ryoqun:
cc: @CriesofCarrots

validator/src/main.rs
```diff
@@ -1287,6 +1287,14 @@ pub fn main() {
                 .default_value(&default_rpc_threads)
                 .help("Number of threads to use for servicing RPC requests"),
         )
+        .arg(
+            Arg::with_name("rpc_timeout")
```
mvines, line 1291: I'm ok calling this `--rpc-bigtable-timeout`, and then plumbing `rpc_bigtable_timeout` throughout this PR, that's really what this arg is after all

validator/src/main.rs
```diff
@@ -1287,6 +1287,14 @@ pub fn main() {
...
+                .long("rpc-timeout")
+                .value_name("NUMBER")
+                .validator(is_parsable::<u64>)
+                .takes_value(true)
```
mvines, line 1295: Default to "30" seconds?

storage-bigtable/src/bigtable.rs
```diff
@@ -135,19 +140,33 @@ impl BigTableConnection {
...
+                                )
+                                .domain_name("bigtable.googleapis.com"),
+                        )?;
+
+                    if let Some(timeout) = timeout {
+                        if read_only {
+                            // apply timeout only when thre is no write queries!
+                            endpoint.timeout(timeout)
+                        } else {
+                            warn!("BigTableConnection's requested timeout configuration is ignored because it's not read-only.");
```
mvines, line 159: I think we can also apply the timeout when read-write and configure `--rpc-bigtable-timeout` higher on nodes with read-write access if necessary (I suspect it won't be necessary though)

mvines:
> good news and bad news:
> 
> timeout is working but shouldn't return empty.... we're close. :)
> 
> ```
> $ date && curl -v -H 'Content-Type: application/json' -d '{"jsonrpc":"2.0","id":1,"method":"getConfirmedBlocks","params":[50000000,50040001]}' http://localhost:8899 ; date
> Tue Jan 19 04:15:35 UTC 2021
> *   Trying 127.0.0.1:8899...
> * TCP_NODELAY set
> * Connected to localhost (127.0.0.1) port 8899 (#0)
> > POST / HTTP/1.1
> > Host: localhost:8899
> > User-Agent: curl/7.68.0
> > Accept: */*
> > Content-Type: application/json
> > Content-Length: 83
> > 
> * upload completely sent off: 83 out of 83 bytes
> * Mark bundle as not supporting multiuse
> < HTTP/1.1 200 OK
> < content-type: application/json; charset=utf-8
> < content-length: 37
> < date: Tue, 19 Jan 2021 04:15:35 GMT
> < 
> {"jsonrpc":"2.0","result":[],"id":1}
> * Connection #0 to host localhost left intact
> Tue Jan 19 04:16:05 UTC 2021
> ```

Ah, yeah would be really nice to return a temporary server error code on a timeout

validator/src/main.rs
ryoqun: done at https://github.com/solana-labs/solana/pull/14657/commits/2b73a6c1ce44ba1511c66fcd6be431d6999f46f4

validator/src/main.rs
ryoqun: :+1: 

done https://github.com/solana-labs/solana/pull/14657/commits/2b73a6c1ce44ba1511c66fcd6be431d6999f46f4

Originally this flag is also used for the blockstore code-path, not only the bigtable codepath of getConfirmedBlocks. I wanted to avoid incompatible change of timing out at 30 secs by default.  But, later I noticed that bigtable is actually causing trouble. I reverted the changes for the blockstore's codepath.

So, assuming bigtable's user is only us, I think it's safe to change the behavior now. :)

storage-bigtable/src/bigtable.rs
ryoqun: kk, it's assuring for me to hear that from you. I'm lightly worried how long it takes to write to bigtables in general and general retry handling in the write path. I'm new to this area of code. ;)

core/src/rpc.rs
```diff
@@ -738,7 +740,12 @@ impl JsonRpcRequestProcessor {
                         bigtable_blocks.retain(|&slot| slot <= end_slot);
                         bigtable_blocks
                     })
-                    .unwrap_or_else(|_| vec![]));
+                    .map_err(|_| {
```
ryoqun, line 743: well, I'm actively censoring not to embed arbitrary bigtable error in the http response. :)

validator/src/main.rs
```diff
@@ -1287,6 +1287,15 @@ pub fn main() {
                 .default_value(&default_rpc_threads)
                 .help("Number of threads to use for servicing RPC requests"),
         )
+        .arg(
+            Arg::with_name("rpc_bigtable_timeout")
+                .long("rpc-bigtable-timeout")
+                .value_name("NUMBER")
```
mvines, line 1293:
```suggestion
                .value_name("SECONDS")
```

validator/src/main.rs
```diff
@@ -1287,6 +1287,15 @@ pub fn main() {
...
+                .validator(is_parsable::<u64>)
+                .takes_value(true)
+                .default_value("30")
+                .help("Number of seconds to time out rpc requests backed by BigTable"),
```
mvines, line 1297:
```suggestion
                .help("Number of seconds before timing out RPC requests backed by BigTable"),
```

mvines (approved):
lgtm, just a couple nits

ryoqun:
@mvines Thanks for quick review!

> Ah, yeah would be really nice to return a temporary server error code on a timeout

Well, I chose client error code (4xx):

- Actually, server isn't at fault; client is too greedy by requesting too large range for practical single request. Clients should just paginate?
- Maybe, we don't intend to fix this in near future (For me, 500s is telling I gotta fix it.... lol)
- I want 500s to disappear on the gcp load balancer graph.

mvines:
> @mvines Thanks for quick review!
> 
> > Ah, yeah would be really nice to return a temporary server error code on a timeout
> 
> Well, I chose client error code (4xx):
> 
> * Actually, server isn't at fault; client is too greedy by requesting too large range for practical single request. Clients should just paginate?
> * Maybe, we don't intend to fix this in near future (For me, 500s is telling I gotta fix it.... lol)
> * I want 500s to disappear on the gcp load balancer graph.

Yeah maybe 408 or 413

core/src/rpc.rs
```diff
@@ -738,7 +740,12 @@ impl JsonRpcRequestProcessor {
...
+                        Error::invalid_params(
```
ryoqun, line 744: Oops, this decent error doesn't translate to bad http reqs to my surprise... I'll just leave this as-is... ;)

```
$ date && curl -v -H 'Content-Type: application/json' -d '{"jsonrpc":"2.0","id":1,"method":"getConfirmedBlocks","params":[50000000,50040001]}' http://localhost:8899 ; date
Tue Jan 19 06:37:03 UTC 2021
*   Trying 127.0.0.1:8899...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8899 (#0)
> POST / HTTP/1.1
> Host: localhost:8899
> User-Agent: curl/7.68.0
> Accept: */*
> Content-Type: application/json
> Content-Length: 83
> 
* upload completely sent off: 83 out of 83 bytes
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< content-type: application/json; charset=utf-8
< content-length: 123
< date: Tue, 19 Jan 2021 06:37:03 GMT
< 
{"jsonrpc":"2.0","error":{"code":-32602,"message":"BigTable query failed (maybe timeout due to too large range?)"},"id":1}
* Connection #0 to host localhost left intact
Tue Jan 19 06:37:33 UTC 2021

```

ryoqun:
ugh, not sure about regression, but I'm still seeing the entire rpc stall even after 30secs..... I'll investigate this.

ryoqun:
The sad news is that timeout doesn't work when there are other active concurrent reqs to bigtable....

It seems that it ignores the endpoint-specific timeout value?

Possible culprits are tonic or grpc itself?

ugggg...

ryoqun:
we want to do like this; but this is gated by tokio 1.0. T_T

https://github.com/hyperium/tonic/compare/v0.3.1...v0.4.0#diff-3ae4162baf9b9d96773d8a767be78d98dd26d9c5b6555c2c751685a03e37a789R15

storage-bigtable/src/bigtable.rs
```diff
@@ -240,7 +262,14 @@ impl BigTable {
         let mut cell_timestamp = 0;
         let mut cell_value = vec![];
         let mut cell_version_ok = true;
+        let started = Instant::now();
+
         while let Some(res) = rrr.message().await? {
+            if let Some(timeout) = self.timeout {
+                if Instant::now().duration_since(started) > timeout {
+                    return Err(Error::TimeoutError);
+                }
```
ryoqun, line 271: I added this new timeout as well.

ryoqun:
> The sad news is that timeout doesn't work when there are other active concurrent reqs to bigtable....
> 
> It seems that it ignores the endpoint-specific timeout value?
> 
> Possible culprits are tonic or grpc itself?
> 
> ugggg...



> we want to do like this; but this is gated by tokio 1.0. T_T
> 
> [hyperium/tonic@v0.3.1...v0.4.0diff-3ae4162baf9b9d96773d8a767be78d98dd26d9c5b6555c2c751685a03e37a789R15](https://github.com/hyperium/tonic/compare/v0.3.1...v0.4.0diff-3ae4162baf9b9d96773d8a767be78d98dd26d9c5b6555c2c751685a03e37a789R15)

Well, I was wrong, frankly...... The original timeout in tonic is applied only to the first response. it seems that bigtable's response chunked/streamed. And I added another timeout.

This time, I'm really getting consistent result. Namely, when the thread pool is full, validator is stuck, but after while it'll start to return requests.

ryoqun:
I didn't introduced much of  branches... but,

>  The diff coverage is `2.9%`.

Seems like bigtable is dangerous place. :)

PR #14673, opened 2021-01-19
Cli: default to single gossip

### LINKED ISSUE #14589: Default solana cli to singleGossip
eg. `solana airdrop` should be FAST

CriesofCarrots:
#### Problem
The Cli waits for any transaction to be finalized, but single-gossip commitment is sufficient for most actions.

#### Summary of Changes
- Default cli to single gossip by initializing RpcClient commitment: https://github.com/solana-labs/solana/pull/14673/commits/7ce63605a90b7099ec051ebfc79104e94ebc4c3d
- Simplify rpc-client calls to use RpcClient's configured commitment; remaining `_with_commitment` calls are because of the different return signatures for get_recent_blockhash and get_account. (Incidentally, there is a lot more work that could be done here, especially the methods in `cli/src/checks.rs` and `client/src/nonce_utils.rs`. But this seemed like enough churn for one PR)

Fixes #14589

### DIFF
--- a/cli/src/cli.rs
+++ b/cli/src/cli.rs
@@ -522,7 +522,7 @@ impl Default for CliConfig<'_> {
             rpc_timeout: Duration::from_secs(u64::from_str(DEFAULT_RPC_TIMEOUT_SECONDS).unwrap()),
             verbose: false,
             output_format: OutputFormat::Display,
-            commitment: CommitmentConfig::default(),
+            commitment: CommitmentConfig::single_gossip(),
             send_transaction_config: RpcSendTransactionConfig::default(),
             address_labels: HashMap::new(),
         }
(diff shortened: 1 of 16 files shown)


CriesofCarrots:
@mvines , I'm wondering about adding `commitment_arg` to all the methods that send transactions, or promoting `commitment_arg` to a global variable, so that users can use max commitment across the board if they want to. Thoughts?

mvines:
> @mvines , I'm wondering about adding `commitment_arg` to all the methods that send transactions, or promoting `commitment_arg` to a global variable, so that users can use max commitment across the board if they want to. Thoughts?

promoting `commitment_arg` to global seems nice since `--url` is already global

CriesofCarrots:
> promoting `commitment_arg` to global seems nice since `--url` is already global

Great. I'll do that in a follow-up PR

mvines (dismissed):
🎉

PR #14684, opened 2021-01-20
Cli: promote commitment to a global arg + config.yml

CriesofCarrots:
#### Problem
Commitment is only configurable for certain cli commands; commands that submit transactions are notable exceptions.

#### Summary of Changes
- Promote commitment to a global arg so it may be configured for all commands. Note: a handful of commands in `cluster_query.rs`, including `solana ping`, were previous set with default commitment of `recent`. With this change, all commands will default to `singleGossip`.
- Add commitment to `solana config ... ` options
- Also fix a couple issues with verbosity I came across in the course of this work

Fixes https://github.com/solana-labs/solana/pull/14673#issuecomment-763060706


cli-config/src/config.rs
```diff
@@ -20,6 +20,8 @@ pub struct Config {
 
     #[serde(default)]
     pub address_labels: HashMap<String, String>,
+
+    pub commitment: String,
```
mvines, line 24: I'm worried this'll cause the cli config to be overwritten with the default when moving between a cli that has/doesn't have this field.    Adding something like `#[serde(default = "singleGossip")]` should prevent this though

mvines (approved):
lgtm overall, just worried about retaining the cli config when upgrading to a version with this patch in it

cli-config/src/config.rs
CriesofCarrots: Oh, good call, thx! Looks like `#[serde(default)]` is sufficient, due to default implementation https://github.com/solana-labs/solana/pull/14684/files#diff-f589ded5548e89f8e7153569e86178c754a3644b9c06d75dca1af23034c1b390R46

PR #14699, opened 2021-01-20
broadcasts duplicate shreds through gossip

behzadnouri:
#### Problem
Following up on [this plan](https://discord.com/channels/428295358100013066/738253477331075162/784888854628139008).

#### Summary of Changes
when window service finds duplicate shreds, they are broadcasted through gossip.

### DIFF
--- a/core/src/duplicate_shred.rs
+++ b/core/src/duplicate_shred.rs
@@ -18,6 +18,9 @@ use thiserror::Error;
 
 const DUPLICATE_SHRED_HEADER_SIZE: usize = 63;
 
+pub(crate) type DuplicateShredIndex = u16;
+pub(crate) const MAX_DUPLICATE_SHREDS: DuplicateShredIndex = 512;
+
 /// Function returning leader at a given slot.
 pub trait LeaderScheduleFn: FnOnce(Slot) -> Option<Pubkey> {}
 impl<F> LeaderScheduleFn for F where F: FnOnce(Slot) -> Option<Pubkey> {}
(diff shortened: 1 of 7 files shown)


core/src/duplicate_shred.rs
```diff
@@ -114,24 +105,65 @@ fn check_shreds(
 pub fn from_duplicate_slot_proof(
```
carllin, line 105: Is this function used anywhere now that we have `from_shred()`?

core/src/duplicate_shred.rs
```diff
@@ -100,12 +90,13 @@ fn check_shreds(
...
         Err(Error::InvalidDuplicateShreds)
     } else {
-        let slot_leader = leader(shred1.slot()).ok_or(Error::UnknownSlotLeader)?;
-        if !shred1.verify(&slot_leader) || !shred2.verify(&slot_leader) {
-            Err(Error::InvalidSignature)
-        } else {
-            Ok(())
+        if let Some(leader_schedule) = leader_schedule {
+            let slot_leader = leader_schedule(shred1.slot()).ok_or(Error::UnknownSlotLeader)?;
+            if !shred1.verify(&slot_leader) || !shred2.verify(&slot_leader) {
```
carllin, line 95: I think this signature verification may be unnecessary since:

1) All incoming shreds to window are verified by signature verification by the SigVerifyStage: https://github.com/ryoqun/solana/blob/master/core/src/sigverify_stage.rs#L89
2) All data shreds recovered from erasure are verified before the duplicate check in insert: https://github.com/solana-labs/solana/blob/master/ledger/src/blockstore.rs#L845-L855
3) We don't check for duplicates on recovered coding shreds

carllin (approved):

core/src/duplicate_shred.rs
behzadnouri: no. I am going to delete it once this code is finalized.

core/src/duplicate_shred.rs
behzadnouri: I see.
Currently this check is disabled since the argument passed in is `None`. I will delete this entirely if there was no use-case for verifying shreds here. But, I guess we probably need to verify duplicate-shreds obtained over gossip.

PR #14702, opened 2021-01-20
Add generalized voting process to enable full inflation

mvines:
#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -1600,13 +1600,18 @@ impl Bank {
     // `pico_inflation` be enabled 2nd, the incorrect start slot provided here should have no
     // effect on the inflation calculation.
     fn get_inflation_start_slot(&self) -> Slot {
-        self.feature_set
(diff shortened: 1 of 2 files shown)


runtime/src/bank.rs
```diff
@@ -4669,7 +4674,8 @@ impl Bank {
             self.rent_collector.rent.burn_percent = 50; // 50% rent burn
         }
 
-        if new_feature_activations.contains(&feature_set::full_inflation::id()) {
+        if !new_feature_activations.is_disjoint(&self.feature_set.full_inflation_features_enabled())
```
mvines, line 4677: The if condition change here means that the block could be run multiple times, but I think that's ok

mvines:
@CriesofCarrots - does this PR still look good to you?  I've made no changes since you looked at it earlier while it was in Draft

CriesofCarrots (approved):
lgtm!

PR #14710, opened 2021-01-20
speed up merkle root calculation

jeffwashington:
#### Problem
merkle root calculation and capitalization are noticeably expensive.

#### Summary of Changes
Remove some array copies and combine computations.

Using new generic function and cap for 10M hashes on colo machine:
compute_merkle_root_and_capitalization(us) avg: 1900825.7
compute_merkle_root_legacy(us) avg: 3535954.1
see merkle-root-bench
~54% the runtime

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -3249,7 +3249,8 @@ impl AccountsDB {
         );
     }
 
-    pub fn compute_merkle_root(hashes: Vec<(Pubkey, Hash, u64)>, fanout: usize) -> Hash {
+    // TODO: remove pub. bench uses it for now.
+    pub fn compute_merkle_root_legacy(hashes: Vec<(Pubkey, Hash, u64)>, fanout: usize) -> Hash {
         let hashes: Vec<_> = hashes
             .into_iter()
             .map(|(_pubkey, hash, _lamports)| hash)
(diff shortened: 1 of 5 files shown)


runtime/src/accounts_db.rs
```diff
@@ -3068,7 +3068,8 @@ impl AccountsDB {
         );
     }
 
-    pub fn compute_merkle_root(hashes: Vec<(Pubkey, Hash, u64)>, fanout: usize) -> Hash {
+    // TODO: remove pub. bench uses it for now.
+    pub fn compute_merkle_root_legacy(hashes: Vec<(Pubkey, Hash, u64)>, fanout: usize) -> Hash {
```
sakridge, line 3072: don't think we need to keep the legacy version when we have the new one

jeffwashington: @sakridge and I discussed. I will redo the testing into constants and remove the legacy function in a subsequent pr.

runtime/src/accounts_db.rs
```diff
@@ -3097,69 +3098,128 @@ impl AccountsDB {
...
+                }
+
+                (
+                    hasher.result(),
+                    Self::checked_cast_for_capitalization(this_sum),
+                )
+            })
+            .collect();
+        time.stop();
+        info!("hashing {} {}", total_hashes, time);
```
sakridge, line 3163: can we put as `debug` level?

Issue #14729, opened 2021-01-21
explorer: Transfer with seed needs an instruction details card


oJshua:
#### Problem
In Explorer, there doesn't exist an instruction card covering the Transfer With Seed operation. 

#### Proposed Solution
Implement 'transfer with seed' card.

PR #14730, opened 2021-01-21
explorer: introduce transfer with seed instruction card

oJshua:
#### Problem
In Explorer, there doesn't exist an instruction card covering the Transfer With Seed operation. 

#### Summary of Changes
Adds Transfer With Seed Instruction card

Fixes https://github.com/solana-labs/solana/issues/14729


oJshua:
transferwseed

CriesofCarrots (approved):
Thanks for the quick work!

PR #14745, opened 2021-01-21
Add ic_msg()/ic_logger_msg() macros

mvines:
The `log!()` macro in programs/bpf_loader/src/lib.rs is generically useful for other native programs.  Surface it in `sdk/src/process_instruction.rs` for use by the rest of the tree with some changes:
1. `log!()` is overloaded, so that name doesn't work generically
2. We use `msg!()` for BPF programs, so that's a starting point.  However `msg!()` relies on global state for its context, and we don't have that luxury in native programs (without introducing thread-local storage 😱)
3. Based on (1) and (2), I arrived at `ic_msg!(invoke_context,...)`.  It behaves like `msg!()` but with context
4. `ic_logger_msg()` is also provided for code that is already holding a `Logger`

### DIFF
--- a/programs/bpf_loader/src/lib.rs
+++ b/programs/bpf_loader/src/lib.rs
@@ -27,6 +27,7 @@ use solana_sdk::{
     clock::Clock,
     entrypoint::SUCCESS,
     feature_set::{bpf_compute_budget_balancing, prevent_upgrade_and_invoke},
+    ic_logger_msg, ic_msg,
     instruction::InstructionError,
     keyed_account::{from_keyed_account, next_keyed_account, KeyedAccount},
     loader_instruction::LoaderInstruction,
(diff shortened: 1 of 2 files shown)


mvines:
@jackcmay - wdyt?

jackcmay:
Cool, I was just thinking about this last night.  I'd like to msgify the message_processor/cpi call path to help folks debug execution issues.

This looks good, I do wonder if we need `ic_logger_msg`, these are mostly for non-critical path messages (errors mostly) so trying to save the `borrow_mut()` might not be important.

mvines:
> This looks good, I do wonder if we need `ic_logger_msg`, these are mostly for non-critical path messages (errors mostly) so trying to save the `borrow_mut()` might not be important.

I mostly did it just to avoid more churn in the bpf loader but I'm game to remove `ic_logger_msg` entirely if you agree.  I was not really concerned about the possible perf impact

jackcmay:
Yeah, I think that would be cleaner, might run into ownership/lock issues, if you do I'm fine with `ic_logger_msg` too.

mvines:
> Yeah, I think that would be cleaner, might run into ownership/lock issues, if you do I'm fine with `ic_logger_msg` too.

Yep,  I ran into a `cannot borrow `*invoke_context` as immutable because it is also borrowed as mutable` pretty quickly.  I'll keep `ic_logger_msg!()` around

PR #14746, opened 2021-01-21
Upgrade sha2 to 0.9.3

sakridge:
#### Problem

sha2 crate older and slower than could be.

#### Summary of Changes

Upgrade to the latest.

Fixes #

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -643,6 +643,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "cpuid-bool"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8aebca1129a03dc6dc2b127edd729435bbc4a37e1d5f4d7513165089ceb02634"
+
+[[package]]
 name = "crc32fast"
 version = "1.2.0"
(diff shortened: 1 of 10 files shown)


ryoqun:
@sakridge maybe this concern is kinda now irrelevant because we do sleep-injected poh? https://github.com/solana-labs/solana/pull/13381#issuecomment-723358980
So, this just save some cycles in the poh thread? Sounds good savings. :)

sakridge:
> 
> 
> @sakridge maybe this concern is kinda now irrelevant because we do sleep-injected poh? [#13381 (comment)](https://github.com/solana-labs/solana/pull/13381#issuecomment-723358980)
> So, this just save some cycles in the poh thread? Sounds good savings. :)

Exactly right :-)

sakridge:
> 
> 
> > @sakridge maybe this concern is kinda now irrelevant because we do sleep-injected poh? [#13381 (comment)](https://github.com/solana-labs/solana/pull/13381#issuecomment-723358980)
> > So, this just save some cycles in the poh thread? Sounds good savings. :)
> 
> Exactly right :-)

It also saves on PoH verify, transaction and accounts merkle hash construction.

PR #14762, opened 2021-01-22
Rpc: Add custom error for BigTable data not found

CriesofCarrots:
#### Problem
Rpc apis that query BigTable do not distinguish between BigTable errors and successful queries that do not find data; they return `null` for all these cases. This is problematic in the case of `getConfirmedBlock`; if a block is in range of Blockstore, the api returns a specific error for skipped blocks and `null` for read errors, and users are used to distinguishing these cases. But for an old slot, this pattern doesn't hold.

#### Summary of Changes
- Expose specific data NotFound errors from storage-bigtable
- Return a custom rpc error when BigTable block is not found

### DIFF
--- a/client/src/rpc_custom_error.rs
+++ b/client/src/rpc_custom_error.rs
@@ -12,7 +12,7 @@ pub const JSON_RPC_SERVER_ERROR_NODE_UNHEALTHLY: i64 = -32005;
 pub const JSON_RPC_SERVER_ERROR_TRANSACTION_PRECOMPILE_VERIFICATION_FAILURE: i64 = -32006;
 pub const JSON_RPC_SERVER_ERROR_SLOT_SKIPPED: i64 = -32007;
 pub const JSON_RPC_SERVER_ERROR_NO_SNAPSHOT: i64 = -32008;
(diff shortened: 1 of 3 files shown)


client/src/rpc_custom_error.rs
```diff
@@ -106,6 +110,11 @@ impl From<RpcCustomError> for Error {
                 message: "No snapshot".to_string(),
                 data: None,
             },
+            RpcCustomError::BigTableSlotSkipped { slot } => Self {
+                code: ErrorCode::ServerError(JSON_RPC_SERVER_ERROR_BIGTABLE_SLOT_SKIPPED),
+                message: format!("Slot {} was skipped, or missing in BigTable", slot),
```
CriesofCarrots, line 115: @mvines how do you feel about this wording?
I originally made this error more generic so it could work for blocks or signatures, but users seem happy with `null` responses for signatures.

mvines: Let's not reveal BigTable to the caller, that's in implementation detail.  PersistentLedgerStorage or LongTermStorage perhaps?   ArWeave/PostreSQL are possible other backends that might be plugged in one day

mvines (approved):

PR #14763, opened 2021-01-22
Aggregate purge and shrink metrics

carllin:
#### Problem
1) Shrink + purge metrics can be spammy when a lot of slots are cleaned up at once
2) Also noticed that `purge_slots()` and `purge_removed_slots_from_store()` shared basically the same code 😝 
 
#### Summary of Changes
1) Aggregate purge-related logic per call to `clean_accounts()` under `clean_accounts_stats`
2) Aggregate Bank::drop() purge into separate `external_purge_slots_stats`
3) Aggregate shrink metrics and report every second
4) Extract shared body of `purge_slots()` and `purge_removed_slots_from_store()` into common function
 `do_purge_slots_from_cache_and_store()`

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -691,7 +691,8 @@ struct PurgeStats {
     safety_checks_elapsed: AtomicU64,
     remove_storages_elapsed: AtomicU64,
     drop_storage_entries_elapsed: AtomicU64,
-    num_slots_removed: AtomicUsize,
+    num_cached_slots_removed: AtomicUsize,
+    num_stored_slots_removed: AtomicUsize,
     total_removed_storage_entries: AtomicUsize,
     total_removed_bytes: AtomicU64,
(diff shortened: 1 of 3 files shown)


runtime/src/accounts_db.rs
```diff
@@ -2515,12 +2769,16 @@ impl AccountsDB {
                 }
                 all_removed_slot_storages.push(slot_removed_storages.clone());
             }
+
+            // Remove any delta pubkey set if existing.
+            self.uncleaned_pubkeys.remove(remove_slot);
```
carllin, line 2774: new behavior is that a slot `S` can now also be removed from this set by the `clean_old_rooted_accounts() -> handle_reclaims() -> process_dead_slots() -> purge_storage_slots()` chain. I think this should be safe however because if a slot is removed here, that means `clean_accounts()` already has seen another set of slot(s) that have modified all the keys in `S`.

sakridge (dismissed):
thanks!

PR #14771, opened 2021-01-22
explorer: Add new token swap instructions and program id

joncinque:
#### Problem

The new token swap has new instructions and program id, so the explorer doesn't recognize it.

#### Summary of Changes

Add them!  I tested this locally and everything looked good

cc @bartosz-lipinski


oJshua (approved):
Perfect! Thanks!

PR #14772, opened 2021-01-22
remove legacy merkle root

jeffwashington:
#### Problem
Remove legacy code. Add expected values for test.

#### Summary of Changes
See problem.

Fixes #


merkle-root-bench/src/main.rs
```diff
@@ -28,38 +28,25 @@ fn main() {
...
+        })
+        .collect();
 
-    let len = elapsed.len();
-    for x in 0..iterations {
-        println!(
-            "compute_merkle_root_and_capitalization(us),{},legacy(us),{}",
-            elapsed[x], elapsed_legacy[x]
-        );
+    for result in elapsed.clone() {
```
sakridge, line 45: you don't need clone here:
```
for result in &elapsed {
    print(result);
}
```

sakridge (approved):
lgtm

PR #14781, opened 2021-01-22
removes redundant epoch stakes cache in retransmit

behzadnouri:
#### Problem
Following d6d76219b, staked nodes computed from vote accounts are
already cached in runtime::Stakes, so the caching in retransmit_stage is
redundant.

Removing the cache here, would simplify follow up changes to turbine
retransmit peer computations.

#### Summary of Changes

Fixes #


behzadnouri:
This is [`retransmit-stage.epoch_cache_update`](https://github.com/solana-labs/solana/blob/491b05975/core/src/retransmit_stage.rs#L297-L333) which shows effectively there is no change (left side is this code).
retransmit_stage epoch_cache_update

sakridge (approved):

PR #14782, opened 2021-01-22
explorer: Add timestamps to transaction history

oJshua:
#### Problem
In Explorer, transaction history doesn't show the newly available block times alongside transaction signatures. 

#### Summary of Changes
Display blocktime column if available.

### DIFF
--- a/explorer/src/components/account/TransactionHistoryCard.tsx
+++ b/explorer/src/components/account/TransactionHistoryCard.tsx
@@ -7,6 +7,7 @@ import { Signature } from "components/common/Signature";
 import { ErrorCard } from "components/common/ErrorCard";
 import { LoadingCard } from "components/common/LoadingCard";
 import { Slot } from "components/common/Slot";
+import { displayTimestampUtc } from "utils/date";
 
 export function TransactionHistoryCard({ pubkey }: { pubkey: PublicKey }) {
   const address = pubkey.toBase58();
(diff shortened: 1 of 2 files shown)


explorer/src/components/account/TransactionHistoryCard.tsx
```diff
@@ -75,6 +77,12 @@ export function TransactionHistoryCard({ pubkey }: { pubkey: PublicKey }) {
             <Slot slot={slot} link />
           </td>
 
+          {hasTimestamps && (
+            <td className="text-muted">
+              {blockTime ? displayTimestampUtc(blockTime * 1000) : "---"}
```
jstarry, line 82: Screen Shot 2021-01-26 at 5 43 58 PM

It's not clear from the UI that this is a UTC timestamp

jstarry (approved):
Looks good after the time zone fix

explorer/src/components/account/TransactionHistoryCard.tsx
oJshua: Agreed -- this actually stems from "unlock day" where we decided to remove the mention of UTC when displaying cluster times. @mvines this should be safe/okay to add back in? Side effect: UTC would also be displayed on the cluster stats info on the front page.

mvines: Yeah I'd prefer an explicit timezone (UTC wfm) everywhere now that the cluster time drift that was causing confusion in the past is resolved.

oJshua: On the transaction details page, we're displaying these as local time. I will follow suit here!

oJshua: @mvines sounds good. I'll update that in this PR.

Issue #14045, opened 2020-12-10
explorer: Display block rewards


oJshua:
#### Problem
Explorer does not display block rewards (https://explorer.solana.com/block/51932256). This is available from RPC and displayed in the CLI. 

#### Proposed Solution
Get block rewards from RPC and display on block page.

PR #14786, opened 2021-01-22
explorer: Add block rewards to block details page

oJshua:
#### Problem
Explorer does not display block rewards (https://explorer.solana.com/block/51932256). This is available from RPC and displayed in the CLI. 

#### Summary of Changes
- Add block rewards to the block details page

Fixes https://github.com/solana-labs/solana/issues/14045

### DIFF
--- a/explorer/src/components/block/BlockRewardsCard.tsx
+++ b/explorer/src/components/block/BlockRewardsCard.tsx
@@ -1,0 +1,59 @@
+import React from "react";
+import { lamportsToSolString } from "utils";
+import { ConfirmedBlock, PublicKey } from "@solana/web3.js";
+import { Address } from "components/common/Address";
+
+export function BlockRewardsCard({ block }: { block: ConfirmedBlock }) {
+  if (block.rewards.length < 1) {
+    return null;
+  }
+
+  return (
+    <div className="card">
+      <div className="card-header align-items-center">
+        <h3 className="card-header-title">Block Rewards</h3>
+      </div>
+
+      <div className="table-responsive mb-0">
+        <table className="table table-sm table-nowrap card-table">
+          <thead>
+            <tr>
+              <th className="text-muted">Address</th>
+              <th className="text-muted">Type</th>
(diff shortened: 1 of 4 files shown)


oJshua:
blockrewards

```
$ solana block 61904263|grep %

  E1SFkvPjU31xWMcvgnX6vhGvfRvb1zXvHFkqmQNEGZKK        fee        ◎0.002107500     ◎393.940422295          0.000534982%

```

explorer/src/components/block/BlockRewardsCard.tsx
```diff
@@ -0,0 +1,59 @@
...
+              let percentChange;
+              if (reward.postBalance !== null && reward.postBalance !== 0) {
+                percentChange = (
+                  (Math.abs(reward.lamports) /
+                    (reward.postBalance - reward.lamports)) *
+                  100
+                ).toFixed(9);
+              }
+              return (
+                <tr>
```
jstarry, line 39: You'll want a key here

jstarry (dismissed):
Nice! Are transaction fees the only reward type right now?

mvines:
> Nice! Are transaction fees the only reward type right now?

Rent, Voting and Staking rewards as well.

explorer/src/components/block/BlockRewardsCard.tsx
oJshua: Thanks!

jstarry:
> > Nice! Are transaction fees the only reward type right now?
> 
> Rent, Voting and Staking rewards as well.

Ah, just on epoch boundaries. @oJshua do you mind paginating new epoch blocks? https://explorer.solana.com/block/62208000 I think we should have a page for each epoch as well at some point

oJshua:
> Ah, just on epoch boundaries. @oJshua do you mind paginating new epoch blocks? https://explorer.solana.com/block/62208000 I think we should have a page for each epoch as well at some point

Oh yes, thanks for pointing that out. I'm on it.

PR #14792, opened 2021-01-23
Speed up generate_index

sakridge:
#### Problem

`generate_index` uses slower parallel data structures but doesn't manage to do anything parallel so is slower than needs to be.

#### Summary of Changes

* Use HashMap instead of DashMap
* Add an insert_update function to the accounts_index which doesn't check the read lock beforehand.
* Take account size from sanitize and re-use it for approx_store_count and also to size the hashmap to store the new updates.

Improves generate_index from ~55 seconds to 8 on a mainnet-beta snapshot.

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -522,7 +522,10 @@ impl AccountStorageEntry {
     }
 
     pub fn set_file<P: AsRef<Path>>(&mut self, path: P) -> IOResult<()> {
-        self.accounts.set_file(path)
+        let num_accounts = self.accounts.set_file(path)?;
+        self.approx_store_count
+            .store(num_accounts, Ordering::Relaxed);
+        Ok(())
     }
 
     pub fn get_relative_path(&self) -> Option<PathBuf> {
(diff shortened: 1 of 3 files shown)


runtime/src/accounts_index.rs
```diff
@@ -769,6 +779,26 @@ impl<T: 'static + Clone + IsCached + ZeroLamport> AccountsIndex<T> {
         }
     }
 
+    pub fn insert_new(
```
carllin, line 782: nit: how about renaming `insert_new() -> insert_new_if_missing()`, and `insert_new_entry() -> insert_new_entry_if_missing()`. 

From the name I thought at first this might not be safe because if it was replacing the pubkey entry every time, the ref count would be incorrect, but looks like the `w_account_entry.update()` below should cover that.

runtime/src/accounts_db.rs
```diff
@@ -4141,51 +4144,45 @@ impl AccountsDB {
...
-                                .is_none()
-                        );
-                    })
-                });
+            let num_accounts = storage_maps
+                .iter()
+                .map(|storage| storage.approx_stored_count())
+                .sum();
+            let mut accounts_map: AccountsMap = AccountsMap::with_capacity(num_accounts);
+            storage_maps.iter().for_each(|storage| {
```
carllin, line 4169: Ah dang, yeah there's only one of these storage entries per slot now!

carllin (approved):

runtime/src/accounts_index.rs
sakridge: Yea. It really is still upsert. Functionally the same, it just doesn't do the initial `get_account_write_entry()` that tries for the key with a read lock before taking the write lock. From `perf`, doing the search in the btree for the key seems to be quite expensive, so when inserting a lot of new keys, doing it twice slows it down a lot and there is no downside of taking `write` if it's single threaded.

PR #14798, opened 2021-01-23
Add security best practice sections

ryoqun:
#### Problem

There are two known security-related recommended practices when developing Solana's programs.

#### Summary of Changes

Document it properly as they're found.

Fixes #

### DIFF
--- a/docs/src/developing/programming-model/transactions.md
+++ b/docs/src/developing/programming-model/transactions.md
@@ -171,6 +171,33 @@ that this method only supports fixed sized types. Token utilizes the
 trait to encode/decode instruction data for both token instructions as well as
 token account states.
 
+### Multiple instructions in a single transaction
+
+A transaction can contain any number of arbitrary instructions up to the
+serialized transaction's max size (currently `PACKET_DATA_SIZE = 1232`) or
+computational budget limit via CPI. This means a malicious user can craft
+various irregular transactions, like a transaction containing your program's
+de-initialize instructions _many times_ inside a single transaction.
+
+This means your program should be hardened properly for any instruction
+sequences.
+
+Note that this hardening should address this subtle one: resetting account
+data to uninitialized state explicitly, not blindly relying on runtime's
+usual deletion behavior for accounts with no lamports at the end of the
+transaction.
+
+That's because even after one of your instructions resets account's lamports
+to 0 for account deletion, the runtime doesn't immediately reset the account's
+data for subsequent instructions when executing the transaction (On the other
+hand, the runtime does reset such account's data for subsequent transactions).
+Thus, it's possible for your program to be tricked to do certain actions more
+than allowed even after intended termination of the account, if not
+hardened properly.
+
+This behavior difference between intra-transaction and inter-transaction
+is intentional for the shared memory program's workings.
+
 ## Signatures
 
 Each transaction explicitly lists all account public keys referenced by the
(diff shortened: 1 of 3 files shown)


ryoqun:
@jackcmay Could you lightly check this? Especially, the placement of the new sections. :)

jackcmay:
Sure thing, will have time this weekend

docs/src/developing/programming-model/transactions.md
```diff
@@ -171,6 +171,33 @@ that this method only supports fixed sized types. Token utilizes the
 trait to encode/decode instruction data for both token instructions as well as
 token account states.
 
+### Multiple instructions in a single transaction
+
+A transaction can contain any number of arbitrary instructions up to the
```
jackcmay, line 176: This section could more clearly describe the runtime behavior of account deinitialization and what programs should do to harden themselves.  The last two paragraphs give good info about inter vs intra which is really the crux of what the programs need to be aware of.

jackcmay:
@ryoqun I pushed some recommendations, what do you think?

ryoqun:
@jackcmay thanks for polishing this! Now the text looks more solid. :)

docs/src/developing/programming-model/transactions.md
```diff
@@ -171,6 +171,26 @@ that this method only supports fixed sized types. Token utilizes the
...
+One not so obvious example is account deinitialization.  Some programs may
+attempt to deinitialize an account by setting its lamports to zero, with the
+assumption that the runtime will delete the account.  This assumption may be
+valid between transactions, but it is not between instructions or cross-program
+invocations.  To harden against this, the program should also explicitly zero out the
+account's data.
+
+An example of where this could be a problem is if a token program, upon
+transferring the token out of an account, sets the account's lamports to zero,
+assuming it will be deleted by the runtime.  If the program does zero out the
```
ryoqun, line 190: Maybe, `If the program does zero out` should be `If the program doesn't zero out`?

I think you just forgot to add `not` because a malicious user *couldn't* trail if the account has been zeroed-out.

ryoqun: @jackcmay it looks like @jstarry (thanks!) fixed this for us..: https://github.com/solana-labs/solana/commit/3eea88a3a84b278efc8d980f60c3b0d69dad0cbb

```patch
diff --git a/docs/src/developing/programming-model/transactions.md b/docs/src/developing/programming-model/transactions.md
index ce376ddaa5f..6fe4093360f 100644
--- a/docs/src/developing/programming-model/transactions.md
+++ b/docs/src/developing/programming-model/transactions.md
@@ -187,7 +187,7 @@ account's data.
 
 An example of where this could be a problem is if a token program, upon
 transferring the token out of an account, sets the account's lamports to zero,
-assuming it will be deleted by the runtime.  If the program does zero out the
+assuming it will be deleted by the runtime.  If the program does not zero out the
 account's data, a malicious user could trail this instruction with another that
 transfers the tokens a second time.
```

Could you quickly check this?. Also, how about backporting to v1.5 this as well?

jackcmay: @ryoqun @jstarry pushed to v1.5: https://github.com/solana-labs/solana/commit/eacf9209f778dc0d7844f1e726be47a23651b4f0

jstarry: Ah sorry for forgetting the backport!

PR #14806, opened 2021-01-23
Reduce ~2 GBs mem by avoiding another overalloc.

ryoqun:
#### Problem

Currently, `ClusterInfoVoteListener` over-allocates (unrecycled) `Packets` by factor of 128. Ultimately, this results in approximately ~2G consistent unneeded memory use  both on testnet and mainnet-beta, which is not easily swappable because of quick create and destroy.

##### Details

Firstly, about the the overallocation:

`ClusterInfoVoteListener` calls `to_packets_chunked` and in turn it calls `Packets::Default()`, which surprisingly creates PinnedVec with capacity of `128`. (I have a draft cleaning pr for this)

https://github.com/solana-labs/solana/blob/cbffab7850e3c3d1545c688da76ee636306b16fa/core/src/cluster_info_vote_listener.rs#L344

https://github.com/solana-labs/solana/blob/cbffab7850e3c3d1545c688da76ee636306b16fa/perf/src/packet.rs#L61-L64

https://github.com/solana-labs/solana/blob/cbffab7850e3c3d1545c688da76ee636306b16fa/perf/src/packet.rs#L21-L22

https://github.com/solana-labs/solana/blob/cbffab7850e3c3d1545c688da76ee636306b16fa/perf/src/packet.rs#L11

Then, `Packet`'s memory size is a `1304`:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=adfaa40101d596d9e372f96b20f1a9df

And finally, `VerifiedVotePakcets` retains those `Packets` keyed by `CrdsValueLabel::Vote(VoteIndex, Pubkey)`:

https://github.com/solana-labs/solana/blob/cbffab7850e3c3d1545c688da76ee636306b16fa/core/src/verified_vote_packets.rs#L9

https://github.com/solana-labs/solana/blob/cbffab7850e3c3d1545c688da76ee636306b16fa/core/src/verified_vote_packets.rs#L28-L32

Also, combined with the fact that there are 300~400 validators both on testnet and mainnet-beta, and x32 votes per validator, we allocate 2G mem, only for 15M worth data (x 128):

```
[4] pry(main)> 1304 * 32 * 400 * 128 / (1024 * 1024)
=> 2037
[5] pry(main)> 1304 * 32 * 400 / (1024 * 1024)
=> 15
```

And this is found by anallyzing the following `heaptrack` info:

```
1.99GB leaked over 259982 calls from
  alloc::alloc::alloc::h42bff26f59c33bd6
[...]

PR #14821, opened 2021-01-25
Remove potentially too costly Packets::default()

ryoqun:
#### Problem

I needed some cleanup while creating #14806 to ensure there is no similar issue.

The root problem is that current api is too easy to misuse. This follow-up pr is to fix the root cause and never reoccur similar issue in the future.

#### Summary of Changes

- Removed `Packets::default()`. As far as I read the code, this not need to be `Default`. By convention, `Default::default()` should be cheap but `Pakcets` violates that by over-allocating silently.
  - Also, implementing `Default` generally opens unbounded call-graph. So, I want to avoid it unless there is clear benefit/necessity to define `Default`.
- Then, _identical_ functionality is renamed to `Packets::for_test_and_bench` to aptly indicate the purpose. New name is intentionally verbose and indicates this shouldn't be used in the production code.
  - Also checked that all current call-sites are tests and benches. (Yeah, I did this boring work once for all... lol)
- Demoted `to_packets` to `#cfg(test)`,  another source of potential cause of overallocation. It seems that production code doesn't actually need it as well.

This pr's cleanup motto is that let's make the ultimate call-site decide the allocation behavior by explicit call of `to_packets_chunked`.

This pr won't be back-ported to v1.4 intentionally. Maybe not worth do it.

Also, there is no functional change.

Fixes #

### DIFF
--- a/bench-streamer/src/main.rs
+++ b/bench-streamer/src/main.rs
@@ -13,7 +13,7 @@ use std::time::SystemTime;
 
 fn producer(addr: &SocketAddr, exit: Arc<AtomicBool>) -> JoinHandle<()> {
     let send = UdpSocket::bind("0.0.0.0:0").unwrap();
-    let mut msgs = Packets::for_bench();
+    let mut msgs = Packets::default();
     msgs.packets.resize(10, Packet::default());
     for w in msgs.packets.iter_mut() {
         w.meta.size = PACKET_DATA_SIZE;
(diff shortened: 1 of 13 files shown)


sakridge:
How about just making `Packets::default` do a `PinnedVec::default` which just does a `Vec::new()`. Only reason to do `with_capacity` is performance, but if it's just used in the test context, then we don't really need that. I'm a bit mixed on the rename to `test_or_bench`.

I don't think the benchmarks do much allocating in the timing sections. In general it generates the packet vector up front, then runs the code on that.

ryoqun:
> How about just making `Packets::default` do a `PinnedVec::default` which just does a `Vec::new()`. Only reason to do `with_capacity` is performance, but if it's just used in the test context, then we don't really need that. I'm a bit mixed on the rename to `test_or_bench`.
> 
> I don't think the benchmarks do much allocating in the timing sections. In general it generates the packet vector up front, then runs the code on that.

@sakridge Thanks for review! That makes sense. I reverted most of changes? How does this pr look now?

sakridge (approved):
lgtm

PR #14822, opened 2021-01-25
Clean up VerifiedVotePackets

ryoqun:
#### Problem

Another half-finished cleanup created while investigating #14806 is now polished up.

This was especially needed for human's eye liveliness analysis of VerifiedVotePackets and its fields... lol

#### Summary of Changes

There is no functional change.

Fixes #

### DIFF
--- a/core/src/verified_vote_packets.rs
+++ b/core/src/verified_vote_packets.rs
@@ -3,20 +3,13 @@ use crate::{
     result::Result,
 };
 use solana_perf::packet::Packets;
-use std::{collections::HashMap, ops::Deref, time::Duration};
+use std::{collections::HashMap, time::Duration};
 
 #[derive(Default)]
 pub struct VerifiedVotePackets(HashMap<CrdsValueLabel, (u64, Packets)>);
 
-impl Deref for VerifiedVotePackets {
-    type Target = HashMap<CrdsValueLabel, (u64, Packets)>;
-    fn deref(&self) -> &Self::Target {
-        &self.0
-    }
-}
-
 impl VerifiedVotePackets {
-    pub fn get_and_process_vote_packets(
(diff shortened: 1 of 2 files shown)


core/src/verified_vote_packets.rs
```diff
@@ -3,20 +3,13 @@ use crate::{
...
-impl Deref for VerifiedVotePackets {
-    type Target = HashMap<CrdsValueLabel, (u64, Packets)>;
-    fn deref(&self) -> &Self::Target {
-        &self.0
-    }
-}
-
 impl VerifiedVotePackets {
-    pub fn get_and_process_vote_packets(
+    pub fn receive_and_process_vote_packets(
```
ryoqun, line 12: nits: `get_` => `receive_`.

core/src/verified_vote_packets.rs
```diff
@@ -3,20 +3,13 @@ use crate::{
     result::Result,
 };
 use solana_perf::packet::Packets;
-use std::{collections::HashMap, ops::Deref, time::Duration};
+use std::{collections::HashMap, time::Duration};
 
 #[derive(Default)]
 pub struct VerifiedVotePackets(HashMap<CrdsValueLabel, (u64, Packets)>);
 
-impl Deref for VerifiedVotePackets {
```
ryoqun, line 11: Well, it turns out that we don't need wholesale `Deref` for `VerifiedVotePackets`, only modest `#[cfg(test)]` simpler getter called `get_vote_packets`.

core/src/verified_vote_packets.rs
ryoqun: Also, I wanted to define real getter called `get_vote_packets` and wanted to avoid the same `get_` for vastly different tasks. :)

carllin (approved):

PR #14835, opened 2021-01-25
camelCase json field labels

### LINKED ISSUE #14830: solana program deploy prints different json schema
#### Problem
When you use cli command `solana program deploy` you will get a print out like `{"ProgramId":"EKf8..."}`if it's a new deployment and  `{"programId":"EKf8..."}` otherwise. This is annoying for those of us parsing it directly using jq in the command line. 

#### Proposed Solution
They should both have key 'programId'

CriesofCarrots:
#### Problem
`solana program deploy` returns json, but isn't consistent with camel-casing field labels.

#### Summary of Changes
- Fix it
In the future, consider fixing these commands to use OutputFormat to be consistent with other cli commands, instead of json all the time.

Fixes #14830

cc @jackcmay

### DIFF
--- a/cli/src/program.rs
+++ b/cli/src/program.rs
@@ -1144,12 +1144,12 @@ fn do_process_program_write_and_deploy(
 
     if let Some(program_signer) = program_signer {
         Ok(json!({
-            "ProgramId": format!("{}", program_signer.pubkey()),
+            "programId": format!("{}", program_signer.pubkey()),
         })
         .to_string())
     } else {
         Ok(json!({
-            "Buffer": format!("{}", buffer_pubkey),
+            "buffer": format!("{}", buffer_pubkey),
         })
         .to_string())
     }


jackcmay:
@CriesofCarrots There are also tests that rely on parsing this json output in the /tests folder that need to be updated

CriesofCarrots:
> @CriesofCarrots There are also tests that rely on parsing this json output in the /tests folder that need to be updated

Thanks @jackcmay . Oof, I just discovered `option_pubkey_to_string()` as well, so I'm going to close this and come back at it later to implement OutputFormat instead.

jackcmay:
@CriesofCarrots  I can make the updates one I understand what needs to be done.  What is the purpose of OutputFormat as it relates to `Program`?

CriesofCarrots:
> @CriesofCarrots I can make the updates one I understand what needs to be done. What is the purpose of OutputFormat as it relates to `Program`?

@jackcmay The purpose of OutputFormat generally is to return human-readable vs json data from solana-cli requests, depending on `--output` parameter. `solana deploy` was an outlier in that it returned json in all cases.
The way we're doing it now is creating a cli-specific return type struct that impl Display, and returning `config.output_format.formatted_string(&<STRUCT>)` from the process fn.
That's obviously a little bit of a hassle in this case, since the Program comprises 6(?) different return types of only contain one field each. I was going to ponder a bit and see if I can think of a slicker model for these cases.

I don't mind handling this and getting your review. I don't think it's super time-sensitive; I just thought it was a quick text update I could bang out.

Issue #14834, opened 2021-01-25
API nodes running the v1.5 branch are generating snapshots with MismatchedTotalLamports errors


mvines:
On ba1, see the `~/ledger.bad-snapshot-for-62313975` directory:
* Search for `MismatchedTotalLamports` in solana-validator.log 
* STR: `solana-ledger-tool -l ~/ledger.bad-snapshot-for-62313975 verify`

On ba2, same as `ba1` but use the `~/ledger.bad-snapshot-for-62248998` directory

ba1/ba2 was running https://github.com/solana-labs/solana/commit/b48dd58fda74729bd50e5f4d16f7c1ff987b7a3b at the time

Interestingly bv4 and bvb are running the same code and may not be seeing mismatches

sakridge:
Accounts caching enabled cc @carllin

mvines:
I've upgraded `bw3` to the tip of v1.5.  The warehouse nodes have automatic snapshot sanity checks every hour, so I'm hoping we'll see an error appear from `bw3` if this is a non-API node problem.

sakridge:
Snapshot at 62313975 created from starting at 62313875 and playing forward to ..975 which seems to verify:

`sakridge@lemond:~/solana/mainnet-beta-snapshot-issue-1-25-2021`

sakridge:
small ledger from 62313875 to 62313975:
```
sol@ba1:~/ledger.bad-snapshot-for-62313975/rocksdb2.tar.zst
```

sakridge:
cc @ryoqun

ryoqun:
this is the truth:

```
[2021-01-25T11:01:10.744412746Z INFO  solana_runtime::bank] bank frozen: 62313975 hash: 4CLpzFHdGw63SnWWv6aEqqPuwK19Kq4FRPfymdteKBP9 accounts_delta: 4YvcemZvpzWxB3AzAhUxGPS46fvXcShTnqibPZvCEtcV signature_count: 553 last_blockhash: 5BszKHkYhyKRgt77DLYrwgEgH6u5fAnqKtd61PnHkBXb capitalization: 488587430684744678
```

this is difference:

```
[2021-01-25T12:26:23.286289635Z WARN  solana_runtime::accounts] verify_bank_hash failed: MismatchedTotalLamports(488587317027074222, 488587430684744678)
[...]

PR #14852, opened 2021-01-26
Snapshots missing slots from accounts cache clean optimization

carllin:
#### Problem
https://github.com/solana-labs/solana/issues/14834#issuecomment-767360854

#### Summary of Changes
Flush everything up to the `max_flush_root` so that snapshots aren't missing slots

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -4434,7 +4434,7 @@ pub mod tests {
     use std::{
         iter::FromIterator,
         str::FromStr,
-        thread::{sleep, Builder, JoinHandle},
+        thread::{self, sleep, Builder, JoinHandle},
         time::Duration,
     };
 
@@ -7462,11 +7462,23 @@ pub mod tests {
             .is_none());
     }
 
+    struct ScanTracker {
+        t_scan: JoinHandle<()>,
+        exit: Arc<AtomicBool>,
+    }
+
+    impl ScanTracker {
+        fn exit(self) -> thread::Result<()> {
+            self.exit.store(true, Ordering::Relaxed);
+            self.t_scan.join()
+        }
+    }
+
     fn setup_scan(
         db: Arc<AccountsDB>,
         scan_ancestors: Arc<Ancestors>,
         stall_key: Pubkey,
-    ) -> (JoinHandle<()>, Arc<AtomicBool>) {
+    ) -> ScanTracker {
         let exit = Arc::new(AtomicBool::new(false));
         let exit_ = exit.clone();
         let ready = Arc::new(AtomicBool::new(false));
@@ -7480,8 +7492,6 @@ pub mod tests {
                     |_collector: &mut Vec<(Pubkey, Account)>, maybe_account| {
                         ready_.store(true, Ordering::Relaxed);
                         if let Some((pubkey, _, _)) = maybe_account {
-                            // Do the wait on account_key2, because clean is happening
-                            // on account_key1's index and we don't want to block the clean.
                             if *pubkey == stall_key {
                                 loop {
                                     if exit_.load(Ordering::Relaxed) {
@@ -7502,7 +7512,7 @@ pub mod tests {
             sleep(Duration::from_millis(10));
         }
 
-        (t_scan, exit)
+        ScanTracker { t_scan, exit }
     }
 
     #[test]
@@ -7537,7 +7547,7 @@ pub mod tests {
         let max_scan_root = 0;
         db.add_root(max_scan_root);
         let scan_ancestors: Arc<Ancestors> = Arc::new(vec![(0, 1), (1, 1)].into_iter().collect());
-        let (t_scan, exit) = setup_scan(db.clone(), scan_ancestors.clone(), account_key2);
+        let scan_tracker = setup_scan(db.clone(), scan_ancestors.clone(), account_key2);
 
         // Add a new root 2
         let new_root = 2;
@@ -7574,8 +7584,7 @@ pub mod tests {
 
         // When the scan is over, clean should not panic and should not purge something
         // still in the cache.
-        exit.store(true, Ordering::Relaxed);
-        t_scan.join().unwrap();
+        scan_tracker.exit().unwrap();
         db.clean_accounts(None);
         let account = db
             .do_load(&scan_ancestors, &account_key, Some(max_scan_root))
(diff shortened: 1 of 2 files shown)


carllin:
Currently running fix on top of 1.5 on `ba2`

ryoqun:
oh, so good finding. I feel guilty for not catching this while reviewing...

runtime/src/accounts_db.rs
```diff
@@ -3162,7 +3172,7 @@ impl AccountsDB {
     fn flush_slot_cache(
         &self,
         slot: Slot,
-        mut should_flush_f: Option<&mut impl FnMut(&Pubkey, &Account) -> bool>,
+        mut should_flush_f: Option<&mut impl FnMut(&Slot, &Pubkey, &Account) -> bool>,
```
ryoqun, line 3175: nits: Because `Slot` is `Copy` and `u64`, I don't think we need to use `&` for it. But if you want consistency for all args to be `&`, I don't say this strongly. :)

carllin:
> oh, so good finding. I feel guilty for not catching this while reviewing...

No worries, it was buried under so many layers of calls xD

runtime/src/accounts_db.rs
```diff
@@ -3110,9 +3116,13 @@ impl AccountsDB {
         // If `should_clean` is None, then`should_flush_f` is also None, which will cause
         // `flush_slot_cache` to flush all accounts to storage without cleaning any accounts.
         let mut should_flush_f = should_clean.map(|(account_bytes_saved, num_accounts_saved)| {
-            move |&pubkey: &Pubkey, account: &Account| {
+            move |slot: &Slot, &pubkey: &Pubkey, account: &Account| {
                 use std::collections::hash_map::Entry::{Occupied, Vacant};
-
+                if let Some(max_clean_root) = max_clean_root {
+                    if *slot > max_clean_root {
```
carllin, line 3122: @ryoqun Alternatively instead of checking this here on every account, I could move the check to here: https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_db.rs#L3136-L3139 outside of the `should_flush_f`, and pass `None` for `should_flush_f` when `root > max_clean_root` but the logic would be more spread out and harder to track. What do you think?

ryoqun:
Now that we're hurt, how about adding another safe-guard here?:

https://github.com/solana-labs/solana/blob/e57b9c3b020a910951640a42ecbd57069fcf3898/runtime/src/bank.rs#L3901-L3904



```rust
self.rc 
    .accounts 
    .accounts_db 
    .flush_accounts_cache(true, Some(self.slot()));
// Forced cache flushing MUST flush everything upto and including the self.slot() at least.
// That's because:
// - `self.slot()` must be root at this point.
// - this (root) slot itself cannot be skipped by pre-flush cleaning because this is the
//   highest slot of cleaning set. And each bank has at least 1 account updates due
//   to sysvar maintenance.
// - `At least` means max_flush_root could be larger than self.slot() if there are
//   `> MAX_CACHE_SLOT` cached and rooted slots.
// - otherwise, this would cause missing storages in the snapshot
assert!(self.slot() <= self.rc.accounts.accounts_db.max_flush_root());
```

Also, edit my jotted-down comments as you like, if you like the math proof style. :)
Or, this could be placed in `AccountsBackgroundService`?

Also, I'd like to add similar assertions `fn get_snapshot_storages()` with `accounts_db_cachding_enabled` flag guard. Defence in depth. :)

runtime/src/accounts_db.rs
```diff
@@ -2996,27 +2996,31 @@ impl AccountsDB {
...
     }
 
-    // `force_flush` flushes all the cached roots `<= max_clean_root`. It also then
+    // `force_flush` flushes all the cached roots `<= max_flush_root`. It also then
     // flushes:
     // 1) Any remaining roots if there are > MAX_CACHE_SLOTS remaining slots in the cache,
     // 2) It there are still > MAX_CACHE_SLOTS remaining slots in the cache, the excess
     // unrooted slots
-    pub fn flush_accounts_cache(&self, force_flush: bool, max_clean_root: Option<Slot>) {
+    pub fn flush_accounts_cache(&self, force_flush: bool, max_flush_root: Option<Slot>) {
```
ryoqun, line 3004: how about renaming `max_flush_root` *argument* to `requested_flush_root` across series of callgraph while keeping `AccountsCache::max_flush_root` *field*?

That's because this variable is more like similar to `SnapshotRequest` and I want to differentiate argument `max_flush_root` and `AccountsCache::max_flush_root`.

Or any other name is fine as long as it indicates we must do something up to the `Slot` regarding flushing.

I know `max_flush_root` is the first one to pop up in mind because of the already existing field name... But, engineer’s job is to give proper names to differentiate hard-to-understand workings of code. ;)

runtime/src/accounts_db.rs
```diff
@@ -3040,11 +3044,11 @@ impl AccountsDB {
         let old_slots = self.accounts_cache.find_older_frozen_slots(MAX_CACHE_SLOTS);
         let excess_slot_count = old_slots.len();
         let mut unflushable_unrooted_slot_count = 0;
-        let max_flushed_root = self.accounts_cache.fetch_max_flush_root();
+        let max_previously_flushed_root = self.accounts_cache.fetch_max_flush_root();
```
ryoqun, line 3047: nits: I think `old_max_flush_root` or `prev_max_flush_root` follows the convention of coding (`old_` or `prev_` prefix) to keep old value in advance of possible mutation of the value.

Mixing `_previously_` in middle sounds something more convoluted is about to happen.

runtime/src/accounts_db.rs
```diff
@@ -3078,30 +3082,32 @@ impl AccountsDB {
...
+        max_flush_root: Option<Slot>,
         should_clean: Option<(&mut usize, &mut usize)>,
     ) -> (usize, usize) {
+        let mut max_clean_root = None;
         if should_clean.is_some() {
-            max_flush_root = self.max_clean_root(max_flush_root);
+            // If there is a long running scan going on, this could prevent any cleaning
+            // based on updates from slots > `max_clean_root`.
+            max_clean_root = self.max_clean_root(max_flush_root);
         }
```
ryoqun, line 3108: nits: 

```rust
let max_clean_root = should_clean.map(|_| {
    // If there is a long running scan going on, this could prevent any cleaning
    // based on updates from slots > `max_clean_root`.
    self.max_clean_root(max_flush_root);
});
```

(shorting code and reducing `mut` is important to reduce reader's cognitive load.)

ryoqun: Also this code clearly reads/states that `max_clean_root` is defined at front and set in stone by narrowing `requested_flush_root` down via the `.max_clean_root()` filter _only if_ we `should_clean`.

runtime/src/accounts_db.rs
ryoqun: As a conclusion, I don't have strong opinion on this. But there is my personal idea:

I understand your concern that logics are being spread. At the same time, I'm also worried that there could be many things in one place. This is hard call.

But separating might be more preferable in this case:
- `slot > max_clean_root` check here is sort of belong to the looping construct over `cached_roots` conceptually, suggesting the below `for` is more proper home for it. That's because given `slot` is always same inside a single iteration.
- if we move this check out of this closure, we can define `max_clean_root` immediately before the `for` loop, reducing being more spread out.
- Although cpu branch prediction/cpu cache nicely work in this case, it's best to hoist redundant code outside inner hot loop. If we urged to remain this code as-is, I think one-line note is needed not to confuse future readers of this code.

runtime/src/accounts_db.rs
```diff
@@ -3078,30 +3082,32 @@ impl AccountsDB {
...
 
-        // If there is a long running scan going on, this could prevent any cleaning
-        // past `max_flush_root`.
+        // Always flush up to `max_flush_root`, which is necessary for things like snapshotting.
         let cached_roots: BTreeSet<Slot> = self.accounts_cache.clear_roots(max_flush_root);
```
ryoqun, line 3111: not related to this pr, but I noticed this `cached_roots` can also be moved immediately before the below `for` loop.

ryoqun:
by the way, I think this bug tells us that we should add another manual test step of running a validator which is constantly accessed by scanning rpc methods... lol

ryoqun (dismissed):
lgtm after all my mostly nits comments are addressed in some way. :)

Super quick debugging and actual fixing in proper way.

sakridge:
Nice find!

runtime/src/accounts_db.rs
carllin: agreed, updated!

runtime/src/accounts_db.rs
carllin: With your suggestion to switch `max_flush_root -> requested_flush_root`, this can go back to being `max_flushed_root` without ambiguity

runtime/src/accounts_db.rs
carllin: ah derp, yeah, thanks for cleaning this up

runtime/src/accounts_db.rs
carllin: added it to the `AccountsBackgroundService` closer to the snapshotting logic

runtime/src/accounts_db.rs
carllin: removed `Slot` entirely

sakridge (approved):

ryoqun (approved):
LGTM!

Thanks for general cleaning up.

PR #14856, opened 2021-01-26
cli now supports a custodian for stake authorize operations

mvines:
Part of #14587


mvines:
@t-nelson: can I hand off the ledger support for this to you?  My plan is to roll out all the client support as quickly as possible, but only land the feature to require a custodian in v1.5.  Then flip on the feature only after we're sure that all clients support it (Ledger being the last one most likely)

t-nelson (approved):
LGTM!  I'll handle ledger-app updates

PR #14877, opened 2021-01-27
explorer: introduce Token Balances on transaction details page

oJshua:
#### Problem
There is currently not a way to view token balance changes from the Transaction Details page.

#### Summary of Changes
Adds a "Token Balances" section on relevant transaction detail pages.

### DIFF
--- a/explorer/src/pages/TransactionDetailsPage.tsx
+++ b/explorer/src/pages/TransactionDetailsPage.tsx
@@ -440,8 +440,10 @@ function generateTokenBalanceRows(
         let delta;
 
         if (accountIndex in preBalanceMap) {
-          delta =
-            normalizeTokenDelta(uiTokenAmount, preBalanceMap[accountIndex].uiTokenAmount);
+          delta = normalizeTokenDelta(
+            uiTokenAmount,
+            preBalanceMap[accountIndex].uiTokenAmount
+          );
         } else {
           delta = uiTokenAmount.uiAmount;
         }
--- a/explorer/src/pages/TransactionDetailsPage.tsx
+++ b/explorer/src/pages/TransactionDetailsPage.tsx
@@ -45,6 +45,7 @@ import { isSerumInstruction } from "components/instruction/serum/types";
 import { MemoDetailsCard } from "components/instruction/MemoDetailsCard";
 import { TokenRegistry } from "tokenRegistry";
 import { BigNumber } from "bignumber.js";
+import { BalanceDelta } from "components/common/BalanceDelta";
 
 const AUTO_REFRESH_INTERVAL = 2000;
 const ZERO_CONFIRMATION_BAILOUT = 5;
@@ -358,24 +359,16 @@ function AccountsCard({
     const post = meta.postBalances[index];
     const pubkey = account.pubkey;
     const key = account.pubkey.toBase58();
-    const renderChange = () => {
-      const change = post - pre;
-      if (change === 0) return "";
-      const sols = lamportsToSolString(change);
-      if (change > 0) {
-        return <span className="badge badge-soft-success">+{sols}</span>;
-      } else if (change < 0) {
-        return <span className="badge badge-soft-warning">-{sols}</span>;
-      }
-      return <span className="badge badge-soft-secondary">0</span>;
-    };
+    const delta = new BigNumber(post).minus(new BigNumber(pre));
 
     return (
       <tr key={key}>
         <td>
           <Address pubkey={pubkey} link />
         </td>
-        <td>{renderChange()}</td>
+        <td>
+          <BalanceDelta delta={delta} isSol />
+        </td>
         <td>{lamportsToSolString(post)}</td>
         <td>
           {index === 0 && (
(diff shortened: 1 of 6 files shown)


oJshua:
deltasexample

oJshua:
intialize

explorer/src/pages/TransactionDetailsPage.tsx
```diff
@@ -410,6 +415,129 @@ function AccountsCard({
...
+
+  const accountRows = rows.map(({ account, delta, balance, mint }) => {
+    const key = account.toBase58();
+    const renderChange = () => {
+      if (delta > 0) {
+        return <span className="badge badge-soft-success">+{delta}</span>;
+      } else if (delta < 0) {
+        return <span className="badge badge-soft-warning">{delta}</span>;
+      }
+      return <span className="badge badge-soft-secondary">{delta}</span>;
```
jstarry, line 498: Nice, do you mind adding the `0` state to how we display SOL delta as well? Maybe create a component that can be shared as well

explorer/src/pages/TransactionDetailsPage.tsx
```diff
@@ -410,6 +415,129 @@ function AccountsCard({
...
+  postTokenBalances: TokenBalance[],
+  accounts: ParsedMessageAccount[]
+): TokenBalanceRow[] {
+  let preBalanceMap: { [index: number]: TokenBalance } = {};
+
+  preTokenBalances.forEach(
+    (balance) => (preBalanceMap[balance.accountIndex] = balance)
+  );
+
+  return postTokenBalances
```
jstarry, line 437: Will `postTokenBalances` contain tokens that were completely spent during a transaction (post balance of 0)?

explorer/src/pages/TransactionDetailsPage.tsx
```diff
@@ -410,6 +415,129 @@ function AccountsCard({
...
+    .map(
+      ({ uiTokenAmount, accountIndex, mint }): TokenBalanceRow => {
+        let delta;
+
+        if (accountIndex in preBalanceMap) {
+          delta = normalizeTokenDelta(
+            uiTokenAmount,
+            preBalanceMap[accountIndex].uiTokenAmount
```
jstarry, line 445: Unlikely to happen but it's possible that the mint of a token changed during a transaction (it could be closed and re-initialized). Can you make sure that the mint matches before computing the delta?

jstarry (commented):
Nice, how have you been testing these changes? Did you setup your own api node?

explorer/src/pages/TransactionDetailsPage.tsx
```diff
@@ -410,6 +415,129 @@ function AccountsCard({
...
+          );
```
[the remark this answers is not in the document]
oJshua: Yes, I started out doing that but there was some strange JS precision stuff going on, which was adding decimals. I would normally use math.js in that situation, but this way didn't involve that dependency. 

Explorer-Solana (3)

explorer/src/pages/TransactionDetailsPage.tsx
oJshua: Good call.

explorer/src/pages/TransactionDetailsPage.tsx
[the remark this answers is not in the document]
jstarry: It would be nice to have a math library that properly handles precision actually.

Looks like [mathjs is built on decimal.js](https://mathjs.org/docs/datatypes/bignumbers.html)

Maybe we should go with bignumber.js? Oyster lending has been using bn.js but I don't think it's as nice for displaying precise values as strings.

https://dev.to/fvictorio/a-comparison-of-bignumber-libraries-in-javascript-2gc5
> bignumber.js and decimal.js are similar, the main difference is that bignumber.js expresses its precision in terms of decimals (appropriate for financial applications, for example) and decimal.js does it in terms of significant digits (better for scientific applications). That's why decimal.js is not a good choice for arbitrary integer arithmetic1.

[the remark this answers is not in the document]
oJshua: Cool, I'll give bignumber a try.

explorer/src/pages/TransactionDetailsPage.tsx
oJshua: Sure thing. What would be the expected behavior in this kind of situation?

explorer/src/pages/TransactionDetailsPage.tsx
oJshua: Yes it will.

explorer/src/pages/TransactionDetailsPage.tsx
jstarry: Probably flatten into two separate rows, one for each mint

oJshua: Since we're reflecting the post balance of an account, maybe it would be better to just throw out the pre balance with the former mint? So a single row?

jstarry: Isn't the intent to show the delta? If I have balance A in token X to balance B in token Y I would expect to see a line item for a full withdrawal of A for token X and then a positive delta of B in token Y

oJshua: Yep -- that makes complete sense. I wasn't grokking that that could happen in one transaction. Thanks for the clarification.

oJshua: Explorer-Solana (6)

oJshua:
> Nice, how have you been testing these changes? Did you setup your own api node?

We have one instance of 1.5 running behind the load balancer. I was pointing directly to that.

jstarry (approved):

PR #14876, opened 2021-01-27
Check hash on stores

sakridge:
#### Problem

Test PR to check snapshot storage integrity with hashing

#### Summary of Changes

Calculate a hash for all stores in a snapshot on snapshot.

Fixes #

PR #14879, opened 2021-01-27
calculate hash from store instead of index

jeffwashington:
#### Problem
The number of accounts are growing.
There is value in moving operations out of the critical path. Some operations like clean and shrink are not able to complete effectively.
One such operation is calculating the hash of a bank.

#### Summary of Changes

Move the hash calculation out of accounts_background_service and into accounts_hash_verifier. The calculation also avoids using the index, reducing contention.
Add a command line option to calculate the hash using the index in the previous location and verify the hash calculation in ahv without using the index.

Fixes #

### DIFF
--- a/core/src/accounts_hash_verifier.rs
+++ b/core/src/accounts_hash_verifier.rs
@@ -8,6 +8,8 @@ use crate::{
     cluster_info::{ClusterInfo, MAX_SNAPSHOT_HASHES},
     snapshot_packager_service::PendingSnapshotPackage,
 };
+use solana_measure::measure::Measure;
+use solana_runtime::accounts_db::AccountsDB;
 use solana_runtime::snapshot_package::{AccountsPackage, AccountsPackageReceiver};
 use solana_sdk::{clock::Slot, hash::Hash, pubkey::Pubkey};
 use std::collections::{HashMap, HashSet};
@@ -83,6 +85,21 @@ impl AccountsHashVerifier {
         fault_injection_rate_slots: u64,
         snapshot_interval_slots: u64,
     ) {
+        let mut time = Measure::start("hash");
+
+        let simple_capitalization_enabled = true; // ??? TODO
+        let hash = AccountsDB::calculate_accounts_hash_using_stores_only(
+            accounts_package.storages.clone(),
+            simple_capitalization_enabled,
+        );
+        time.stop();
+
+        datapoint_info!(
+            "accounts_hash_verifier",
+            ("calculate_hash", time.as_us(), i64),
+        );
+        assert_eq!(hash.0, accounts_package.hash); // TODO: don't calculate hash elsewhere
+
         if fault_injection_rate_slots != 0
             && accounts_package.slot % fault_injection_rate_slots == 0
         {
--- a/core/src/accounts_hash_verifier.rs
+++ b/core/src/accounts_hash_verifier.rs
@@ -8,7 +8,9 @@ use crate::{
     cluster_info::{ClusterInfo, MAX_SNAPSHOT_HASHES},
     snapshot_packager_service::PendingSnapshotPackage,
 };
-use solana_runtime::snapshot_package::{AccountsPackagePre, AccountsPackageReceiver};
+use solana_runtime::snapshot_package::{
+    AccountsPackage, AccountsPackagePre, AccountsPackageReceiver,
+};
 use solana_sdk::{clock::Slot, hash::Hash, pubkey::Pubkey};
 use std::collections::{HashMap, HashSet};
 use std::{
@@ -49,7 +51,7 @@ impl AccountsHashVerifier {
 
                     match accounts_package_receiver.recv_timeout(Duration::from_secs(1)) {
                         Ok(accounts_package) => {
-                            Self::process_accounts_package(
+                            Self::process_accounts_package_pre(
                                 accounts_package,
                                 &cluster_info,
                                 &trusted_validators,
@@ -72,7 +74,7 @@ impl AccountsHashVerifier {
         }
     }
 
-    fn process_accounts_package(
+    fn process_accounts_package_pre(
         accounts_package: AccountsPackagePre,
         cluster_info: &ClusterInfo,
         trusted_validators: &Option<HashSet<Pubkey>>,
@@ -83,7 +85,32 @@ impl AccountsHashVerifier {
         fault_injection_rate_slots: u64,
         snapshot_interval_slots: u64,
     ) {
-        let accounts_package = solana_runtime::snapshot_utils::process_accounts_package_pre(accounts_package);
+        let accounts_package =
+            solana_runtime::snapshot_utils::process_accounts_package_pre(accounts_package);
+        Self::process_accounts_package(
+            accounts_package,
+            cluster_info,
+            trusted_validators,
+            halt_on_trusted_validator_accounts_hash_mismatch,
+            pending_snapshot_package,
+            hashes,
+            exit,
+            fault_injection_rate_slots,
+            snapshot_interval_slots,
+        );
+    }
+
+    fn process_accounts_package(
+        accounts_package: AccountsPackage,
+        cluster_info: &ClusterInfo,
+        trusted_validators: &Option<HashSet<Pubkey>>,
+        halt_on_trusted_validator_accounts_hash_mismatch: bool,
+        pending_snapshot_package: &Option<PendingSnapshotPackage>,
+        hashes: &mut Vec<(Slot, Hash)>,
+        exit: &Arc<AtomicBool>,
+        fault_injection_rate_slots: u64,
+        snapshot_interval_slots: u64,
+    ) {
         let hash = accounts_package.hash;
         if fault_injection_rate_slots != 0
             && accounts_package.slot % fault_injection_rate_slots == 0
--- a/core/src/accounts_hash_verifier.rs
+++ b/core/src/accounts_hash_verifier.rs
@@ -8,9 +8,7 @@ use crate::{
     cluster_info::{ClusterInfo, MAX_SNAPSHOT_HASHES},
     snapshot_packager_service::PendingSnapshotPackage,
 };
-use solana_measure::measure::Measure;
-use solana_runtime::accounts_db::AccountsDB;
-use solana_runtime::snapshot_package::{AccountsPackage, AccountsPackageReceiver};
+use solana_runtime::snapshot_package::{AccountsPackagePre, AccountsPackageReceiver};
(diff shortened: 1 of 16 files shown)


core/src/accounts_hash_verifier.rs
```diff
@@ -83,6 +85,21 @@ impl AccountsHashVerifier {
         fault_injection_rate_slots: u64,
         snapshot_interval_slots: u64,
     ) {
+        let mut time = Measure::start("hash");
+
+        let simple_capitalization_enabled = true; // ??? TODO
```
sakridge, line 90: could we just forward this through in the `AccountsPackage` struct?

jeffwashington: Yes. I wasn’t sure which structs could be changed and if that was better or to look it up somehow.

runtime/src/accounts_db.rs
```diff
@@ -277,6 +285,18 @@ impl<'a> LoadedAccount<'a> {
             },
         }
     }
+
+    pub fn lamports(&self) -> u64 {
+        match self {
+            LoadedAccount::Stored(stored_account_meta) => {
+                stored_account_meta.clone_account().lamports
```
sakridge, line 292: `stored_account_meta.account_meta.lamports`

core/src/accounts_hash_verifier.rs
jeffwashington: doh. turns out we don't need it anyway. We are not interested in calculating lamports anyway.

runtime/src/accounts_db.rs
```diff
@@ -3875,15 +3899,236 @@ impl AccountsDB {
...
+                let key = public_key;
+                let source_item = (
+                    version,
+                    *loaded_account.loaded_hash(),
+                    balance,
+                    raw_lamports,
+                );
+                match map.entry(*key) {
+                    Occupied(mut dest_item) => {
+                        if dest_item.get_mut().version() <= source_item.version() {
```
sakridge, line 4105: I think just comparing the version is not correct actually. We need to use the version to disambiguate within a slot, but, then pick the highest slot update from that set.

jeffwashington: I could imagine this to be true. I have been wanting to rework this. Hopefully that is the missing key to newly inconsistent results.

jeffwashington:
depends on #14876

runtime/src/accounts_db.rs
```diff
@@ -3716,9 +3735,14 @@ impl AccountsDB {
...
         slot: Slot,
         debug: bool,
+        unstable_sort: bool,
     ) -> ((Hash, u64), (Measure, Measure)) {
         let mut sort_time = Measure::start("sort");
-        Self::sort_hashes_by_pubkey(&mut hashes);
+        if unstable_sort {
+            hashes.par_sort_unstable_by(|a, b| a.0.cmp(&b.0));
+        } else {
+            Self::sort_hashes_by_pubkey(&mut hashes);
```
sakridge, line 3744: I think all callers have unique pubkeys so stable sort wouldn't make much difference here. Can we just use unstable for all?

runtime/src/accounts_db.rs
```diff
@@ -3875,15 +3899,241 @@ impl AccountsDB {
...
+            )
+        } else {
+            self.calculate_accounts_hash(slot, ancestors, false, simple_capitalization_enabled)
+                .unwrap()
+        }
+    }
+
+    pub fn update_accounts_hash_with_store_option(
+        &self,
+        use_store: bool,
```
sakridge, line 4037: The naming `use_store` and `with_store_option` doesn't make sense to me. Both methods use stores to fetch the hash data, just one uses the index to find which store to look at, the other scans the stores to find which to use.

`with_index` or `use_index` or `with_accounts_index` might be better options. Or it could be `use_only_stores` I guess.

runtime/src/accounts_db.rs
jeffwashington: results are equivalent, but performance is not. When called from the normal code path, I think the pub keys may be mostly or completely sorted since they came from the index. In that case, unstable sort was 3x slower than the stable sort we were using prior to my changes.
However, for the completely unsorted pubkeys that come from searching the stores, the stable sort was ~2x slower than the unstable sort. I read the docs for unstable and stable and it indicated you could get better sorting performance in some conditions with unstable sort, and I did see that.

jeffwashington: Reworking the pipeline of store searching through sort is one of the perf optimizations that I am aiming to work on next. In the meantime, if I use stable sort or unstable, the performance of something will get much worse.

runtime/src/accounts_db.rs
jeffwashington: without_index is my proposal. Ok?

runtime/src/accounts_db.rs
jeffwashington: and, yes, all known, current callers should have unique pub keys.

runtime/src/accounts_db.rs
sakridge: sounds good

runtime/src/accounts_db.rs
sakridge: I see, we should at least add a comment about that because you would expect unstable sort to be strictly better than stable sort.

And it's consistent with randomized keys?

jeffwashington:
@sakridge , a few tests are failing like the log snippets below. These tests pass locally. Can you please point me to how to understand this failure?

thread 'tests::V1_2_0_Development::test_bank_forks_snapshot_n' panicked at 'called `Result::unwrap()` on an `Err` value: IO(Os { code: 2, kind: NotFound, message: "No such file or directory" })', core/tests/snapshots.rs:167:10

stack backtrace:

thread 'tests::V1_2_0_Devnet::test_bank_forks_snapshot_n' panicked at 'called `Result::unwrap()` on an `Err` value: IO(Os { code: 2, kind: NotFound, message: "No such file or directory" })', core/tests/snapshots.rs:167:10

thread 'tests::V1_2_0_Testnet::test_bank_forks_snapshot_n' panicked at 'called `Result::unwrap()` on an `Err` value: IO(Os { code: 2, kind: NotFound, message: "No such file or directory" })', core/tests/snapshots.rs:167:10

thread 'tests::V1_2_0_MainnetBeta::test_bank_forks_snapshot_n' panicked at 'called `Result::unwrap()` on an `Err` value: IO(Os { code: 2, kind: NotFound, message: "No such file or directory" })', core/tests/snapshots.rs:167:10

runtime/src/accounts_db.rs
jeffwashington: [Here's](https://docs.rs/rayon/1.5.0/rayon/slice/trait.ParallelSliceMut.html#method.par_sort_unstable_by) where I got the idea.
"Unstable is generally faster than stable sorting, except in a few special cases, e.g. when the slice consists of several concatenated sorted sequences."

So, it seems we may want to prefer unstable in this case. However, switching the existing code path to stable resulted in measurably worse performance. I intend to refactor the pipeline of how we calculate a hash based on store data only (for better performance) once this pr goes in. I believe this sort change may disappear at that point. In the meantime, we end up with worse performance in one of the two cases. Here are some numbers that aren't quite what I'd want to compare, but they are perhaps useful:
sort time, unstable, not using index, in ahv threads: 989341us, 14518391 hashes
sort time, stable, using index, in normal validator threads(?): sort=34271us, hash_total=13550803

In my earlier experiments, using unstable sort on data that was indexed resulted in hash times that were much higher that 34ms.

runtime/src/accounts_db.rs
```diff
@@ -3875,15 +3899,241 @@ impl AccountsDB {
...
+                    dest_item.insert(found_item);
+                }
+            }
+            Vacant(v) => {
+                v.insert(found_item);
+            }
+        };
+    }
+
+    fn scan_slot_using_snapshot(
```
sakridge, line 4088: It's not scanning a slot. It's scanning a set of stores which may be from multiple slots.

`scan_snapshot_stores` for the name, maybe.

runtime/src/accounts_db.rs
```diff
@@ -3875,15 +3899,241 @@ impl AccountsDB {
...
+
+                let key = public_key;
+                let source_item = (
+                    version,
+                    *loaded_account.loaded_hash(),
+                    balance,
+                    raw_lamports,
+                    slot,
+                );
+                Self::handle_one_loaded_account(key, source_item, &map);
```
sakridge, line 4118: all the re-assignments of `key` make it a bit hard to follow. can we just pass `loaded_account.pubkey()` into `handle_one_loaded_account` directly?

runtime/src/accounts_db.rs
```diff
@@ -3875,15 +3899,241 @@ impl AccountsDB {
...
 
+    fn handle_one_loaded_account(
+        key: &Pubkey,
+        found_item: CalculateHashIntermediate,
+        map: &DashMap<Pubkey, CalculateHashIntermediate>,
+    ) {
+        match map.entry(*key) {
+            Occupied(mut dest_item) => {
+                let contents = dest_item.get();
+                if contents.4 < found_item.4
```
sakridge, line 4075: If `CalculateHashIntermediate` was a struct and not a tuple, this would read better as `found_item.slot` and `contents.slot`

runtime/src/accounts_db.rs
sakridge: Ok, let's add comment. Reading the code it's a bit weird and not obviously done for performance or functionality.

core/tests/snapshots.rs
```diff
@@ -218,6 +218,7 @@ mod tests {
             if slot % set_root_interval == 0 || slot == last_slot - 1 {
                 // set_root should send a snapshot request
                 bank_forks.set_root(bank.slot(), &request_sender, None);
+                bank.update_accounts_hash();
```
jeffwashington, line 221: @sakridge , this is required since we are no longer setting the hash on the bank. We used to do it [here](https://github.com/solana-labs/solana/blob/cc8be7869be5c352b381a3a75b4274d93a951841/runtime/src/accounts_background_service.rs#L91).
We get a default hash (111...) later [here](https://github.com/solana-labs/solana/blob/cc8be7869be5c352b381a3a75b4274d93a951841/core/tests/snapshots.rs#L157), if we don't set the hash in the bank sometime.
I'm not confident in the right way to fix these tests.

jeffwashington: we could calculate not using the index at some point. That would perhaps be more like the run-time behavior going forward. Perhaps it does not matter for these tests and this is sufficient.

jeffwashington:
submitted elsewhere

PR #14880, opened 2021-01-27
Don't reset store accounts from accounts clean

sakridge:
#### Problem

Stores can be reset while snapshots are being created.

#### Summary of Changes

Check if the store accounts are being removed from a clean on a rooted store, then don't reset the accounts if that is the case.

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -6738,15 +6738,8 @@ pub mod tests {
             let mut slot = 1;
             let slot_orig = slot;
             db.store_cached(slot, &[(&pubkey, &account)]);
-            //db.mark_slot_frozen(slot);
             assert_load_account(&db, slot_orig, pubkey, lamports);
             db.add_root(slot);
-            // ??? maybe db.flush_accounts_cache(true, Some(slot));
-            /*
-            let storage_entry = get_account_entry(&db);
-            storage_entry.set_status(AccountStorageStatus::Full);
-            */
-            //let store_id = storage_entry.append_vec_id();
             assert_load_account(&db, slot_orig, pubkey, lamports);
             db.flush_rooted_accounts_cache(Some(slot_orig), None);
             let snapshot = db.get_snapshot_storages(slot_orig);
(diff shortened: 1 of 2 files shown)


runtime/src/accounts_db.rs
```diff
@@ -1487,7 +1487,7 @@ impl AccountsDB {
 
         let reclaims = self.purge_keys_exact(&pubkey_to_slot_set);
 
-        self.handle_reclaims(&reclaims, None, false, None);
+        self.handle_reclaims(&reclaims, None, false, None, true);
```
sakridge, line 1490: I think it should be false.

runtime/src/accounts_db.rs
```diff
@@ -1833,7 +1840,7 @@ impl AccountsDB {
                 );
                 return 0;
             }
-            self.do_shrink_slot_stores(slot, stores.iter());
+            self.do_shrink_slot_stores(slot, stores.iter(), false);
```
sakridge, line 1843: This doesn't need the argument. It can be set to `false` in the call to `store_accounts_custom` inside `do_shrink_slot_stores`

runtime/src/accounts_db.rs
```diff
@@ -1552,8 +1553,13 @@ impl AccountsDB {
             } else {
                 (None, None)
             };
-        let dead_slots =
-            self.remove_dead_accounts(reclaims, expected_single_dead_slot, reclaimed_offsets);
+        let dead_slots = self.remove_dead_accounts(
+            reclaims,
+            expected_single_dead_slot,
+            reclaimed_offsets,
+            no_dead_slot,
```
sakridge, line 1560: Don't need to add `no_dead_slot` if it doesn't exist already.

jeffwashington:
@sakridge, I do not know the correct behavior for this now-failing test with the changes in this PR. Can you please help me understand what this test should now expect?

---- accounts_db::tests::test_account_grow stdout ----
  | thread 'accounts_db::tests::test_account_grow' panicked at 'assertion failed: `(left == right)`
  | left: `Full`,
  | right: `Available`', runtime/src/accounts_db.rs:5319:17

runtime/src/accounts_db.rs
```diff
@@ -6728,38 +6728,75 @@ pub mod tests {
     #[test]
     fn test_storage_remove_account_with_reset_then_access() {
```
jeffwashington, line 6729: @carllin , @sakridge 
I will need help getting this test to do the right thing. There is still too much I don't understand about caching, clean, etc. Stephen's goal was to test:
a case where since we don't reset the accounts, that we might look at the store and it was reset before, but now it has data.

In this scenario, I can't get the remove_account call to get hit from the clean (or shrink). I apologize - this code is really messy. I have been trying a lot of different approaches and looking at 'similar' tests. I think we shouldn't hold up this PR further for me to learn all I need to learn to get this tested right. We can talk Monday.

jeffwashington: I can confirm the test is now hitting the code path in remove_account for the clean. I am not sure which shrink @sakridge meant with this slack:
"we need a test that does a shrink then tests that it can still access the account data, and also for a clean" I'll try other variations of shrink tomorrow. shrink_all_slots isn't currently hitting remove_account as far as I can tell.

jeffwashington: I merged changes from carl's test and mine. I now confirm hitting the remove_account case with both shrink and clean.

jeffwashington:
This pr has cherry-picked #14987 and likely depends on that PR.

runtime/src/accounts_db.rs
```diff
@@ -4396,7 +4396,7 @@ impl AccountsDB {
             None::<StorageFinder>,
             None::<Box<dyn Iterator<Item = u64>>>,
             is_cached_store,
-            false,
+            true,
```
jeffwashington, line 4399: @sakridge , I think you made this change locally yesterday in discussion with carl and me. Can you please confirm? This test was failing because of this.
test_account_grow

jeffwashington: I had sent you a [github comment](https://github.com/solana-labs/solana/pull/14880#issuecomment-770334958) on this test.

sakridge: I'll take a look.

sakridge: I think the test is just flaky.

jeffwashington: Summary: this is a correct change.

jeffwashington:
submitted from #15022

PR #14884, opened 2021-01-27
keygen: Add num-threads argument for grind

joncinque:
#### Problem

Grinding vanity keypairs is fun to run in the background while doing other things, but grind defaults to `num_cpus::get()`, which can make it impossible to do anything else.

#### Summary of Changes

Add the "num-threads" argument to the grind command, defaulting to `num_cpus::get()`

Fixes #

### DIFF
--- a/keygen/src/keygen.rs
+++ b/keygen/src/keygen.rs
@@ -1,9 +1,10 @@
 use bip39::{Language, Mnemonic, MnemonicType, Seed};
 use clap::{
-    crate_description, crate_name, value_t, values_t_or_exit, App, AppSettings, Arg, ArgMatches,
-    SubCommand,
+    crate_description, crate_name, value_t, value_t_or_exit, values_t_or_exit, App, AppSettings,
+    Arg, ArgMatches, SubCommand,
 };
 use solana_clap_utils::{
+    input_validators::is_parsable,
     keypair::{
         keypair_from_seed_phrase, prompt_passphrase, signer_from_path,
         SKIP_SEED_PHRASE_VALIDATION_ARG,
@@ -207,6 +208,7 @@ fn grind_parse_args(
 }
 
 fn main() -> Result<(), Box<dyn error::Error>> {
+    let default_num_threads = num_cpus::get().to_string();
     let matches = App::new(crate_name!())
         .about(crate_description!())
         .version(solana_version::version!())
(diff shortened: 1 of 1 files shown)


keygen/src/keygen.rs
```diff
@@ -335,6 +336,13 @@ fn main() -> Result<(), Box<dyn error::Error>> {
                         .multiple(true)
                         .validator(grind_validator_starts_and_ends_with)
                         .help("Saves specified number of keypairs whos public key starts and ends with the indicated perfix and suffix\nExample: --starts-and-ends-with sol:ana:4\nPREFIX and SUFFIX type is Base58\nCOUNT type is u64"),
+                )
+                .arg(
+                    Arg::with_name("num_threads")
+                        .long("num-threads")
+                        .value_name("NUMBER")
+                        .takes_value(true)
```
mvines, line 344: Add `.validator(is_parsable::<usize>)` for better error handling when the user says `--num-threads BLAH`

keygen/src/keygen.rs
```diff
@@ -335,6 +336,13 @@ fn main() -> Result<(), Box<dyn error::Error>> {
...
+                        .help("Specify the number of grind threads, defaults to the number of CPUs on the machine"),
```
mvines, line 345: If you use `.default_value()` then the ", defaults to the number of CPUs on the machine" text can be removed and the cli help will show the actual default number to the user
```rust
let default_num_cpus = num_cpus::get().to_string();
...
.default_value(&default_num_cpus)
```

keygen/src/keygen.rs
```diff
@@ -532,11 +540,15 @@ fn do_main(matches: &ArgMatches<'_>) -> Result<(), Box<dyn error::Error>> {
                 exit(1);
             }
 
+            let num_threads =
+                value_t!(matches.value_of("num_threads"), usize).unwrap_or(num_cpus::get());
```
mvines, line 544: With `.default_value()` and `.validator()` now in place, it's safe to use `value_t_or_exit!()` here since the value of `matches.value_of("num_threads")` is now fully sanitized by clap
```suggestion
                value_t_or_exit!(matches.value_of("num_threads"), usize);
```

joncinque:
Thanks for the feedback!

mvines (approved):

PR #14886, opened 2021-01-27
explorer: Introduce scam registry and flag on account pages

oJshua:
#### Problem
We've identified a scam account and there will be more to come.

#### Summary of Changes
Introduce scam account registry and add specific account.


oJshua:
image

There is a fake Telegram group and they are asking for donations on https://explorer.solana.com/address/GACpXND1SSfTSQMmqGuFvGwXB3jGEYBDRGNzmLfTYwSP?cluster=mainnet-beta

PR #14891, opened 2021-01-27
Remove serial_test_derive dependency

palfrey:
`serial_test` since [0.3](https://github.com/palfrey/serial_test/releases/tag/v0.3.0) includes `serial_test_derive` and lets you use the macro without needing to directly include `serial_test_derive`


CriesofCarrots:
Thank you!

PR #14894, opened 2021-01-28
explorer: Update flagged account message

oJshua:
#### Problem
Message should be updated to reflect community decision and include solana-flagged-accounts repository.

#### Summary of Changes
Update messaging and add solana-flagged-repository.


oJshua:
Scratch this, we're going to use the raw github way / github pages @bartosz-lipinski suggested

PR #14902, opened 2021-01-28
cli: Improve reliability of program deploys

jstarry:
#### Problem
Deploys often fail

#### Summary of Changes
- Don't bail on rpc request timeouts
- Don't resign and send new transactions until slot expires
- Resend already sent transactions without resigning to see if any more will land
- Update the tpu port periodically
- Throttle transaction sending to 100 TPS to ensure more transactions land

#### Notes
- I tested against mainnet by writing a 900kb program to a program buffer account (about 1900 transactions)
- Still need to figure out if we can speed up "get_signature_statuses" because it is _really_ slow
- Maybe we should check buffer write success rate by checking the account data periodically rather than requesting hundreds of transaction statuses

Fixes #

### DIFF
--- a/cli/src/program.rs
+++ b/cli/src/program.rs
@@ -40,8 +40,16 @@ use solana_sdk::{
 };
 use solana_transaction_status::TransactionConfirmationStatus;
 use std::{
-    cmp::min, collections::HashMap, error, fs::File, io::Read, net::UdpSocket, path::PathBuf,
-    sync::Arc, thread::sleep, time::Duration,
+    cmp::min,
+    collections::HashMap,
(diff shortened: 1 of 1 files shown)


jstarry:
@CriesofCarrots CI is green now!

CriesofCarrots (approved):
Nice, this lgtm! Tested well for me locally and against testnet. I've also been considering sending the write transactions to 1 or more additional future leaders. Wdyt?

jstarry:
> I've also been considering sending the write transactions to 1 or more additional future leaders. Wdyt?

Sounds great!

PR #14908, opened 2021-01-28
program-test: Expose bank task to fix fuzzing

### LINKED ISSUE #14707: program-test in fuzzing environment blows up /tmp
#### Problem

Using program-test with honggfuzz creates many new directories of the form `/tmp/.tmp*` ie `/tmp/.tmp0BaGeJ`, with files that contain all sorts of garbage data and strings, .ie:

```
$w#OC4wOC4wOC4OC4wOC4w OCUUUUUUUU5w_ES333333334_Tw_S4Cw4_4/4w8
                                                              Cc15W?oBq$sAs!
q$qEEPw 0EO!0Ņ  g w 0COb&@k{
{zq   {_y*{_x-"-%--
Oyaa^p] yy-g-aw4W       qw4Wx{j_&g w b8ya{_)qg w U)yay{gg       Oyy=yrW?oqW?    OOym!h_yWP-
gW?ohy_!rg  e1{zm&y!    yiW?o{*Ua\WOhw  _Ovyg4OW-qO!UW  a`ya_p/C/%5 w Cw 6/eg w F/g w /$@w `w e/#g w 0w %{g g w O@{    Some <= kindcodeRentKindNonePermissionDeniedTransferWithSeed) when slicing `AddrNotAvailableAllocateWithSeed (bytes NotFoundTransferTimedOutAllocateindex out of bounds: the len is Error: Associated address does not match seed derivationTransfer  lamports to the associated token accountAllocate space for the associated token accountAssign the associated token account to the SPL Token programInitialize the associated token accountTried to shrink to a larger capacity/Users/mvines/.cache/solana/v0.12/rust-bpf-sysroot/src/liballoc/raw_vec.rsinternal error: entered unreachable codelamports_per_byte_yearexemption_thresholdburn_percentalready mutably borrowedstruct Rent with 3 elements/Users/mvines/.cache/solana/v0.12/rust-bpf-sysroot/src/libcore/slice/mod.rsassertion failed: `(left == right)`
  left: ``,
 right: ``: destination and source slices have different lengthsa Display implementation returned an error unexpectedly, expected invalid length /Users/mvines/.cache/solana/v0.12/rust-bpf-sysroot/src/liballoc/raw_vec.rsinternal error: entered unreachable code/Users/mvines/.cache/solana/v0.12/rust-bpf-sysroot/src/libcore/slice/mod.rsassertion failed: `(left == right)`
```

It fills up 1GB in a few seconds until blowing out the filesystem.

#### Proposed Solution

Investigate and see!  I tried disabling all logs, and that had no effect.

joncinque:
#### Problem

Program-test creates a running task to register ticks on the bank, but does not expose this task.  For most cases, this works great, but in a fuzzing environment, not closing the thread properly blows up `/tmp` with all sorts of directories like `/tmp/.tmp0000`, perhaps related to rocksdb?  I'm not too familiar with that side of it.

#### Summary of Changes

Expose the running task on the bank through a new return type, `ProgramTestState`, and allow it to be stopped on `drop`.  This changes the API on `ProgramTest.start()`, which seems reasonable considering most of these are tests users.  If needed, I can spin off a new / old function and mark this as deprecated.

One thing to note is that we can't simply destructure this as:
```
let ProgramTestState { mut banks_client, payer, last_blockhash, .. } = program_test.start().await;
```
Because the task gets dropped at that point.

The other nice part of this approach is that we can start to support a lot of functions that have been asked for by devs.  My next idea is to add a `warp_to_slot` on `ProgramTestState`.  We'll just need to keep track of the `BankForks` and use `Bank::warp_from_parent`.

This approach gets much more consistent performance from fuzzing as well.  For the test, we can't easily use the `fuzz!` macro because then we need to build with `cargo hfuzz build`, so instead it's simulating what a fuzz run would look like with arbitrary `&[u8]` data.

Fixes #14707

### DIFF
--- a/program-test/src/lib.rs
+++ b/program-test/src/lib.rs
@@ -579,11 +579,7 @@ impl ProgramTest {
         }
     }
 
-    /// Start the test client
-    ///
-    /// Returns a `BanksClient` interface into the test environment as well as a payer `Keypair`
-    /// with SOL for sending transactions
-    pub async fn start(self) -> ProgramTestState {
+    fn setup_bank(&self) -> (Arc<RwLock<BankForks>>, Keypair, Hash, GenesisConfig) {
         {
             use std::sync::Once;
             static ONCE: Once = Once::new();
@@ -624,7 +620,7 @@ impl ProgramTest {
         }
 
         // User-supplied additional builtins
-        for builtin in self.builtins {
+        for builtin in self.builtins.iter() {
             bank.add_builtin(
                 &builtin.name,
                 builtin.id,
@@ -632,7 +628,7 @@ impl ProgramTest {
             );
         }
 
-        for (address, account) in self.accounts {
+        for (address, account) in self.accounts.iter() {
             if bank.get_account(&address).is_some() {
                 info!("Overriding account at {}", address);
             }
@@ -645,16 +641,49 @@ impl ProgramTest {
                 ..BpfComputeBudget::default()
             }));
         }
-
         let bank = setup_fee_calculator(bank);
         let last_blockhash = bank.last_blockhash();
         let bank_forks = Arc::new(RwLock::new(BankForks::new(bank)));
+
+        (bank_forks, payer, last_blockhash, genesis_config)
+    }
+
+    pub async fn start(self) -> (BanksClient, Keypair, Hash) {
+        let (bank_forks, payer, last_blockhash, genesis_config) = self.setup_bank();
         let transport = start_local_server(&bank_forks).await;
         let banks_client = start_client(transport)
             .await
             .unwrap_or_else(|err| panic!("Failed to start banks client: {}", err));
 
-        ProgramTestState::new(
+        // Run a simulated PohService to provide the client with new blockhashes.  New blockhashes
+        // are required when sending multiple otherwise identical transactions in series from a
+        // test
+        tokio::spawn(async move {
+            loop {
+                bank_forks
+                    .read()
+                    .unwrap()
+                    .working_bank()
+                    .register_tick(&Hash::new_unique());
+                tokio::time::sleep(genesis_config.poh_config.target_tick_duration).await;
+            }
+        });
+
+        (banks_client, payer, last_blockhash)
+    }
+
+    /// Start the test client
+    ///
+    /// Returns a `BanksClient` interface into the test environment as well as a payer `Keypair`
+    /// with SOL for sending transactions
+    pub async fn start_with_context(self) -> ProgramTestContext {
+        let (bank_forks, payer, last_blockhash, genesis_config) = self.setup_bank();
+        let transport = start_local_server(&bank_forks).await;
+        let banks_client = start_client(transport)
+            .await
+            .unwrap_or_else(|err| panic!("Failed to start banks client: {}", err));
+
+        ProgramTestContext::new(
             bank_forks,
             banks_client,
             payer,
@@ -705,24 +734,21 @@ struct DroppableTask<T>(Sender<()>, JoinHandle<T>);
 
 impl<T> Drop for DroppableTask<T> {
     fn drop(&mut self) {
-        match self.0.try_send(()) {
-            Ok(_) => {}
-            Err(e) => {
-                debug!("Error cancelling task: {:?}", e);
-            }
+        if let Err(e) = self.0.try_send(()) {
+            debug!("Error cancelling task: {:?}", e);
         }
     }
 }
 
-pub struct ProgramTestState {
+pub struct ProgramTestContext {
     pub banks_client: BanksClient,
     pub payer: Keypair,
     pub last_blockhash: Hash,
     _bank_task: DroppableTask<()>,
 }
 
-impl ProgramTestState {
-    pub fn new(
+impl ProgramTestContext {
+    fn new(
         bank_forks: Arc<RwLock<BankForks>>,
         banks_client: BanksClient,
         payer: Keypair,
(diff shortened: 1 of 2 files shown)


program-test/src/lib.rs
```diff
@@ -698,3 +700,62 @@ impl ProgramTestBanksClientExt for BanksClient {
...
+
+pub struct ProgramTestState {
+    pub banks_client: BanksClient,
+    pub payer: Keypair,
+    pub last_blockhash: Hash,
+    _bank_task: DroppableTask<()>,
+}
+
+impl ProgramTestState {
+    pub fn new(
```
mvines, line 725: I think we can limit scope here
```suggestion
    fn new(
```

program-test/src/lib.rs
```diff
@@ -698,3 +700,62 @@ impl ProgramTestBanksClientExt for BanksClient {
...
+        match self.0.try_send(()) {
+            Ok(_) => {}
+            Err(e) => {
+                debug!("Error cancelling task: {:?}", e);
+            }
+        }
+    }
+}
+
+pub struct ProgramTestState {
```
mvines, line 717: nit: I think I fancy "context" more:
```suggestion
pub struct ProgramTestContext {
```

program-test/src/lib.rs
```diff
@@ -539,7 +583,7 @@ impl ProgramTest {
     ///
     /// Returns a `BanksClient` interface into the test environment as well as a payer `Keypair`
     /// with SOL for sending transactions
-    pub async fn start(self) -> (BanksClient, Keypair, Hash) {
+    pub async fn start(self) -> ProgramTestState {
```
mvines, line 586: I'm definitely concerned about changing this signature and blowing up existing users.  I think we need to keep it for existing users and future users that just want to keep it simple.

Some ideas for a new method name for this signature:
1. `enter()` -- you enter the ProgramTextContext
2. `run()` -- confusing, do I "start" or "run"
3. `start_with_context()` -- maybe my favourite of the three, consistent with `start()` and the Rust norms of adding "with" methods that enhance the functionality of the base method

mvines (commented):
I like the direction.  We needed a place to hook up a clock/slot manipulation interface and this provides such a location

program-test/src/lib.rs
joncinque: Good call, I was thinking someone might want to hack on it from the outside, but this'll be safer

program-test/src/lib.rs
joncinque: Works for me!

program-test/src/lib.rs
```diff
@@ -698,3 +729,59 @@ impl ProgramTestBanksClientExt for BanksClient {
...
+    fn new(
+        bank_forks: Arc<RwLock<BankForks>>,
+        banks_client: BanksClient,
+        payer: Keypair,
+        last_blockhash: Hash,
+        genesis_config: GenesisConfig,
+    ) -> Self {
+        // Run a simulated PohService to provide the client with new blockhashes.  New blockhashes
+        // are required when sending multiple otherwise identical transactions in series from a
+        // test
```
mvines, line 760: This behaviour is important for the user to know about and it won't occur if the user uses `start()`.  How about a the function header comment for `start_with_context()` that mentions this and how `start()` doesn't have it.   This can then for the basis for additional new features that are only supported by `start_with_context()`.  Then in v1.6 timeframe we could even consider marking `start()` as deprecated

mvines (dismissed):
nice work

program-test/src/lib.rs
joncinque: I'm not sure I understand -- lines 658-670 start up the ticking task for `start()`, so do you want to get rid of that to differentiate with `start_with_context()`?

mvines: Ah, sorry got confused there for a moment.  My comment makes no sense, withdrawn!

program-test/src/lib.rs
```diff
@@ -698,3 +729,59 @@ impl ProgramTestBanksClientExt for BanksClient {
...
+        let target_tick_duration = genesis_config.poh_config.target_tick_duration;
+        let (tx, mut rx) = mpsc::channel(1);
```
mvines, line 762:
```suggestion
        let (sender, mut receiver) = mpsc::channel(1);
```
🙏🏼.   Most people unfortunately like to use tx to mean transaction in the code base, except when it means something else like in this case

program-test/src/lib.rs
```diff
@@ -698,3 +729,59 @@ impl ProgramTestBanksClientExt for BanksClient {
...
+        let bank_task = DroppableTask(
+            tx,
+            tokio::spawn(async move {
+                loop {
+                    if rx.try_recv().is_ok() {
```
mvines, line 767: An AtomicBool is generally used in other parts of the code for this kind of exit pattern.   I'm not opposed to a channel if you want to keep it though.

Here's one example of using the atomic bool:
https://github.com/solana-labs/solana/blob/d6873b82ab61071eb87f4c18b2f40cd24d44bbc3/core/src/transaction_status_service.rs#L33

program-test/src/lib.rs
joncinque: Ah, right, on it

program-test/src/lib.rs
joncinque: That was actually my initial approach, but then I thought a message would be clearer.  I'll change it to an AtomicBool to keep things consistent in the codebase

PR #7774, opened 2020-01-13
Fix Turbine making Port based decisions

sagar-solana:
#### Problem

Turbine chooses whether or not to forward a packet to `neighbors` and `children` or just `children` based on which port the packet arrived on. Turbine also drops any packets that arrive on the repair port.

This port based filtering is a problem because it can easily be spoofed and cause selective propagation of packets through a cluster

#### Summary of Changes

A Validator will now always perform a "wide"(to neighbors and children) retransmit if it was on the critical path for a packet. 

Validators can easily check if they are "on the critical path" for a given packet (since the Turbine tree is shuffled per packet) by looking at their index after the shuffle. If their Index is a multiple of the Turbine Fanout, that means they must be on the critical path.

First of many partial fixes for #6672

Issue #6594, opened 2019-10-28
Retransmit recovered shreds through turbine


pgarg66:
#### Problem
A shred that is recovered through erasure codes, is not retransmitted through turbine. If the shred was originally sent to the node from slot leader (and it got dropped), likely the whole cluster is missing it. This can lead to repairs that could be avoided if the node retransmitted the shred thru turbine.

#### Proposed Solution
When a node recovers a data shred, it should check if the shred was sent to it from the leader. If so, it should retransmit the shred thru turbine.

jon-chuang:
I am interested in this. Can start in 6 weeks.

behzadnouri:
Closed by https://github.com/solana-labs/solana/pull/19233

PR #14910, opened 2021-01-28
makes turbine peer computation consistent between broadcast and retransmit

behzadnouri:
#### Problem
`get_broadcast_peers` is using `tvu_peers`:
https://github.com/solana-labs/solana/blob/84e52b606/core/src/broadcast_stage.rs#L362-L370
which is potentially inconsistent with `retransmit_peers`:
https://github.com/solana-labs/solana/blob/84e52b606/core/src/cluster_info.rs#L1332-L1345

Also, the leader does not include its own contact-info when broadcasting
shreds:
https://github.com/solana-labs/solana/blob/84e52b606/core/src/cluster_info.rs#L1324
but on the retransmit side, slot leader is removed only _after_ neighbors and
children are computed:
https://github.com/solana-labs/solana/blob/84e52b606/core/src/retransmit_stage.rs#L383-L384
So the turbine broadcast tree is different between the two stages.

Additionally, broadcast is done to only one node in the first layer:
https://github.com/solana-labs/solana/blob/84e52b606/core/src/broadcast_stage.rs#L388-L395
which is a single point of failure should this fail.

#### Summary of Changes
* Removed retransmit_peers. Broadcast and retransmit stages will use tvu_peers
  consistently.
* Retransmit stage removes slot leader _before_ computing children and
  neighbors.
* ~Broadcast stage forwards the shred to all nodes in the first layer in
  addition to the first node.~

### DIFF
--- a/core/src/broadcast_stage.rs
+++ b/core/src/broadcast_stage.rs
@@ -7,9 +7,9 @@ use self::{
 };
 use crate::contact_info::ContactInfo;
 use crate::crds_gossip_pull::CRDS_GOSSIP_PULL_CRDS_TIMEOUT_MS;
-use crate::weighted_shuffle::weighted_best;
+use crate::weighted_shuffle::weighted_shuffle;
 use crate::{
-    cluster_info::{ClusterInfo, ClusterInfoError},
+    cluster_info::{compute_retransmit_peers, ClusterInfo, ClusterInfoError, DATA_PLANE_FANOUT},
     poh_recorder::WorkingBankEntry,
     result::{Error, Result},
 };
@@ -23,7 +23,7 @@ use solana_metrics::{inc_new_counter_error, inc_new_counter_info};
 use solana_runtime::bank::Bank;
 use solana_sdk::timing::timestamp;
 use solana_sdk::{clock::Slot, pubkey::Pubkey};
-use solana_streamer::sendmmsg::send_mmsg;
+use solana_streamer::sendmmsg::multicast;
 use std::sync::atomic::AtomicU64;
 use std::{
     collections::HashMap,
@@ -373,7 +373,7 @@ pub fn get_broadcast_peers(
 /// # Remarks
 pub fn broadcast_shreds(
     s: &UdpSocket,
-    shreds: &Arc<Vec<Shred>>,
+    shreds: &[Shred],
     peers_and_stakes: &[(u64, usize)],
     peers: &[ContactInfo],
     last_datapoint_submit: &Arc<AtomicU64>,
@@ -385,25 +385,41 @@ pub fn broadcast_shreds(
         return Ok(());
     }
     let mut shred_select = Measure::start("shred_select");
+    let stakes: Vec<_> = peers_and_stakes.iter().map(|(stake, _)| *stake).collect();
     let packets: Vec<_> = shreds
         .iter()
         .map(|shred| {
-            let broadcast_index = weighted_best(&peers_and_stakes, shred.seed());
-
-            (&shred.payload, &peers[broadcast_index].tvu)
+            let index: Vec<_> = weighted_shuffle(&stakes, shred.seed())
+                .into_iter()
+                .map(|i| peers_and_stakes[i].1)
+                .collect();
+            // Broadcast to the first node.
+            // Forward to the rest of the first layer.
+            let (neighbors, _) = compute_retransmit_peers(DATA_PLANE_FANOUT, 0, &index);
+            let anchor = &peers[neighbors[0]];
+            let forwards = neighbors
+                .into_iter()
+                .skip(1)
+                .map(|i| &peers[i].tvu_forwards)
+                .filter(|addr| ContactInfo::is_valid_address(addr));
+            let addrs: Vec<_> = std::iter::once(&anchor.tvu).chain(forwards).collect();
+            (&shred.payload, addrs)
         })
         .collect();
     shred_select.stop();
     transmit_stats.shred_select += shred_select.as_us();
 
-    let mut sent = 0;
     let mut send_mmsg_time = Measure::start("send_mmsg");
-    while sent < packets.len() {
-        match send_mmsg(s, &packets[sent..]) {
-            Ok(n) => sent += n,
-            Err(e) => {
-                return Err(Error::Io(e));
-            }
+    for (payload, addrs) in packets {
+        let mut sent = 0;
+        while sent < addrs.len() {
+            sent += match multicast(s, payload, &addrs[sent..]) {
+                Ok(count) => count,
+                Err(err) => {
+                    error!("failed to broadcast shred: {}", err);
+                    return Err(Error::from(err));
+                }
+            };
         }
     }
     send_mmsg_time.stop();
--- a/core/src/broadcast_stage.rs
+++ b/core/src/broadcast_stage.rs
@@ -7,9 +7,9 @@ use self::{
 };
 use crate::contact_info::ContactInfo;
 use crate::crds_gossip_pull::CRDS_GOSSIP_PULL_CRDS_TIMEOUT_MS;
-use crate::weighted_shuffle::weighted_shuffle;
+use crate::weighted_shuffle::weighted_best;
 use crate::{
-    cluster_info::{compute_retransmit_peers, ClusterInfo, ClusterInfoError, DATA_PLANE_FANOUT},
+    cluster_info::{ClusterInfo, ClusterInfoError},
     poh_recorder::WorkingBankEntry,
     result::{Error, Result},
 };
@@ -23,7 +23,7 @@ use solana_metrics::{inc_new_counter_error, inc_new_counter_info};
 use solana_runtime::bank::Bank;
 use solana_sdk::timing::timestamp;
 use solana_sdk::{clock::Slot, pubkey::Pubkey};
-use solana_streamer::sendmmsg::multicast;
+use solana_streamer::sendmmsg::send_mmsg;
 use std::sync::atomic::AtomicU64;
 use std::{
     collections::HashMap,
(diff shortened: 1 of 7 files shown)


behzadnouri:
These are gce cluster metrics with 20 nodes and fanout of 3.
The left chunk is the master code, as opposed to this code on the right.
There are some savings on the repair side:

`serve_repair-total_packets`:
serve_repair-total_packets

`serve_repair-repair-timing::send-repairs-elapsed`:
serve_repair-repair-timing send-repairs-elapsed



But on broadcast and retransmit, timing metrics and total number of packets will increase because in the broadcast stage we are forwarding to all nodes in the first layer not just only the first node. Also existing bug effectively is not transmitting packets to the whole cluster, so removing that bug inevitably increases number of packets and elapsed time. 

`broadcast-transmit-shreds-stats::transmit_elapsed`:
broadcast-transmit-shreds-stats transmit_elapsed

`retransmit-stage::total_packets`:
retransmit-stage total_packets

core/src/broadcast_stage.rs
```diff
@@ -385,25 +385,41 @@ pub fn broadcast_shreds(
...
+            // Broadcast to the first node.
+            // Forward to the rest of the first layer.
+            let (neighbors, _) = compute_retransmit_peers(DATA_PLANE_FANOUT, 0, &index);
+            let anchor = &peers[neighbors[0]];
+            let forwards = neighbors
+                .into_iter()
+                .skip(1)
+                .map(|i| &peers[i].tvu_forwards)
+                .filter(|addr| ContactInfo::is_valid_address(addr));
+            let addrs: Vec<_> = std::iter::once(&anchor.tvu).chain(forwards).collect();
```
sakridge, line 405: hm. so the leader sends to the whole first layer?

behzadnouri: yes, and I think this is consistent with the intended design in:
https://docs.solana.com/cluster/turbine-block-propagation#layer-and-neighborhood-structure
> The current leader makes its initial broadcasts to at most DATA_PLANE_FANOUT nodes.

sakridge: Yes, but it does not send every shred to every node in the first layer. The first design is that it should do a round-robin of the nodes in the first layer with each shred, then those nodes transfer within themselves. Otherwise the leader has 200x (fanout times) the upload bandwidth requirement to send all the block data.

behzadnouri: > Yes, but it does not send every shred to every node in the first layer.

Yes, the current code does not do that. And I have mentioned in PR description this is a change I am making, which to me sounds more consistent with the intended design in:
https://docs.solana.com/cluster/turbine-block-propagation#layer-and-neighborhood-structure

Currently the leader broadcast the shred to exactly one node and that node can be any node, including an unstaked node (and the packet may be lost or the node may go down, or turbine peers computations may be inconsistent between the leader and the node which receives the packet). So my understanding is that this is very fragile and does not seem to have enough redundancy. If that single packet is lost only the leader will have the shred in the cluster.

> then those nodes transfer within themselves

The documentation suggests that that is the intended design. But the implementation is not doing that. Only the first node in each neighborhood is forwarding the shred to other nodes in that neighborhood. Other nodes in the neighborhood do not send packets to each other.

> Otherwise the leader has 200x (fanout times) the upload bandwidth requirement to send all the block data.

Yes, that is added cost of this PR. But already that single node which obtains the shred from the leader does that 200x fanout:
https://github.com/solana-labs/solana/blob/33dc63ccb/core/src/retransmit_stage.rs#L387-L388

I am not happy with the 200x fanout cost for the leader either. Perhaps the `DATA_PLANE_FANOUT` could be smaller. Or for now the leader could forward to a sample of nodes in the first layer rather than all.

sakridge: No it's not the intended design. Erasure coding makes up for the fact that a single packet could get lost to a down node.

So if there are 200 shreds in a slot and fanout=200. The leader broadcasts a single shred to each node in layer 0, then each node in layer 0 broadcasts to the other nodes in layer 0. The leader would send 200 shreds total and each of the nodes in the first layer also sends 199 shreds (and maybe 1 to lower layers). So everyone has sent about the same amount and has optimal upload bandwidth usage. If the leader sends to all 200, then it will have sent many more shreds than everyone else (200 *200).

sakridge: To me, that documentation is vague and says: "The current leader makes its initial broadcasts to at most DATA_PLANE_FANOUT nodes." Which does not say exactly how the leader splits the broadcast between those 1st layer nodes. Just that the initial broadcasts should be to that set. Where does it say the leader broadcasts all data to all nodes?

This one is more specific:
https://medium.com/solana-labs/turbine-solanas-block-propagation-protocol-solves-the-scalability-trilemma-2ddba46a51db

"Our solution to this problem, Turbine, borrows heavily from BitTorrent — although a few major technical details differentiate the two. Turbine is optimized for streaming, and transmits data using UDP only, and implements a random path per packet through the network as leaders (block producers) stream their data. The leader breaks the block into packets up to 64KB in size. For a 128MB block, the leader produces 2,000 64KB packets, and transmits each packet to a different validator."

behzadnouri: This is what the current implementation is doing:
* The leader sends the shred to the first node in the first layer: https://github.com/solana-labs/solana/blob/33dc63ccb/core/src/broadcast_stage.rs#L390-L394
* Each node always broadcasts to all its children (200 fanout): https://github.com/solana-labs/solana/blob/33dc63ccb/core/src/retransmit_stage.rs#L386-L391
* The first node in each neighborhood on each layer also broadcasts to all its neighbor in its layer (so an additional 199 fanout): https://github.com/solana-labs/solana/blob/33dc63ccb/core/src/retransmit_stage.rs#L387
* No other two nodes in the same layer or neighborhood send packets to each other.

So, effectively:
* The leader has fanout of: 1
* First node of each neighborhood in each layer has fanout of: 200 + 199 (children + neighbors)
* All other nodes have fanout of: 200 (children)

sakridge: I see. It's only 200 children if there are 200 neighborhoods in the next lower level right?

behzadnouri: yes

sakridge: Does the broadcast go to only the 1st layer? I thought it weights all tvu_peers? Is it filtered somewhere?

behzadnouri: Not sure I understood the question.
Currently, the leader sends the shred only to the first node in the first layer: https://github.com/solana-labs/solana/blob/33dc63ccb/core/src/broadcast_stage.rs#L390-L394

From there it is re-transmitted to the rest of cluster.

sakridge: It sends to 1 random node in `peers_and_stakes` but doesn't that come from `get_broadcast_peers` here: https://github.com/solana-labs/solana/blob/33dc63ccbdd685e9b3abe35640fd6a563dac81e9/core/src/broadcast_stage.rs#L366-L369

That looks at `cluster_info.tvu_peers()` which is all nodes in the whole network with a valid TVU port right? So, it will not be the subset of nodes which are just in layer 1.

behzadnouri: Right, but `shred.seed()` is fixing how nodes are ordered in the turbine retransmit tree, and `wighted_best` here:
https://github.com/solana-labs/solana/blob/33dc63ccb/core/src/broadcast_stage.rs#L390-L394
is effectively returning the very first node in the first layer.
It is equivalent of `shuffle_peers_and_index` in retransmit:
https://github.com/solana-labs/solana/blob/33dc63ccb/core/src/retransmit_stage.rs#L341
and then taking the very first node.

sakridge: Got it. Thanks.

behzadnouri: Can we have a decision here?

As I described above: https://github.com/solana-labs/solana/pull/14910#discussion_r567443834
current code is passing the onus of broadcasting from leader to the first node in the first layer (which is an extra hop and a single point of failure with no redundancy). See https://github.com/solana-labs/solana/issues/6594 for example:
> If the shred was originally sent to the node from slot leader (and it got dropped), likely **the whole cluster is missing it**.

I do not see how that is beneficial enough to do so. All nodes have a potential fanout of at least 200 whereas the leader is only 1.

That being said, the part that is broadcasting from leader to other nodes in the first layer, can be left out of this pr. The primary fix here is excluding the leader from peers computation in retransmit-stage to make it consistent with broadcast stage. You can use [turbine-tree-dump-branch](https://github.com/behzadnouri/solana/commit/turbine-tree-dump) which samples the turbine tree at broadcast and retransmit stage and shows that for the same value of the seed the two stages do **not** order the nodes in the same way. This is probably one of the reasons https://github.com/solana-labs/solana/pull/7774 got stuck.

sakridge: I don't think this strategy of broadcasting to everyone in layer 1 is feasible in the real network with fanout of 200. It works fine in a small network with nodes on the same 10gbps switch as tested, but 200x more upload bandwidth for the leader will be huge and way less efficient. We can scale the duplication with higher erasure rate. With erasure coding not any one data packet is a single point of failure, because there are multiple coding shreds sent that could recover the shred lost from that pathway.

> That being said, the part that is broadcasting from leader to other nodes in the first layer, can be left out of this pr

Ok, yes that sounds good.

behzadnouri: Reverted the part broadcasting from leader to first layer

core/src/retransmit_stage.rs
```diff
@@ -252,6 +252,35 @@ fn enable_turbine_retransmit_peers_patch(shred_slot: Slot, root_bank: &Bank) ->
     }
 }
 
+// Drops shred slot leader from retransmit peers.
+// TODO: decide which bank should be used here.
+// TODO: confirm that this should use shred.slot() as opposed to
+// packet.meta.slot.
```
sakridge, line 258: They should be the same value. I think `packet.meta.slot` was an optimization to not deserialize the slot again out of the packet.

sakridge (approved):
Looks good to me. Do we need a cutoff? I guess we can test on testnet.

core/src/retransmit_stage.rs
behzadnouri: thanks, removed the todo comment

behzadnouri:
> Looks good to me. Do we need a cutoff? I guess we can test on testnet.

The broadcast stage is not impacted, and the retransmit was excluding leader after peers computation now it is before. So, I am leaning to just test this on testnet.

behzadnouri:
Redoing the gce test since the broadcast part to first layer nodes was reverted.

This again is a gce cluster of 20 nodes and fanout of 3. The order is reversed though, i.e. the left chunk is this code, versus the current master code on the right.

`serve_repair-total_packets`:
serve_repair-total_packets

`serve_repair-repair-timing::send-repairs-elapsed`:
serve_repair-repair-timing send-repairs-elapsed

`retransmit-stage::total_packets`: More packets presumably because existing bug is effectively dropping packets because of wrong turbine tree calculation:
retransmit-stage total_packets

sakridge:
@behzadnouri Really promising results! Should we backport to 1.6 to give it a try on testnet?

behzadnouri:
> @behzadnouri Really promising results! Should we backport to 1.6 to give it a try on testnet?

yes, backporting.

behzadnouri:
**TDS** metrics up until the crash:

`serve_repair-total_packets`: fewer packets for repair.
serve_repair-total_packets

`serve_repair-repair::shred-count`: fewer shreds are repaired.
serve_repair-repair shred-count

`retransmit-stage::total_packets`: more packets reaching the cluster through turbine re-transmit.
retransmit-stage total_packets

`retransmit-slot-num-packets::num_shreds`: more shreds are broadcasted to the cluster through turbine re-transmit:
retransmit-slot-num-packets num_shreds

behzadnouri:
This patch is now live on **mainnet** as of 1.5.19 which was released ~3 days ago.
Similar improvements: fewer shreds have gone through repair, and more packets are re-transmitted through cluster:

`serve_repair-repair::repair_total` and `shred_count`: fewer shreds need repair
serve_repair-repair

`retransmit-stage::total_packets`: more packets reaching the cluster through turbine re-transmit.
retransmit-stage total_packets

carllin:
> This patch is now live on **mainnet** as of 1.5.19 which was released ~3 days ago.
> Similar improvements: fewer shreds have gone through repair, and more packets are re-transmitted through cluster:
> 
> `serve_repair-repair::repair_total` and `shred_count`: fewer shreds need repair
> serve_repair-repair
> 
> `retransmit-stage::total_packets`: more packets reaching the cluster through turbine re-transmit.
> retransmit-stage total_packets

Thanks for the update, awesome stuff! Glad those metrics are coming in handy 😉

PR #14912, opened 2021-01-28
Surface faucet start failures to the user of solana-test-validator

mvines:
When there's an existing user of port 9900, `solana-test-validator` provides no help as to what went wrong.  Instead surface the error so the user gets a hint about what went wrong

Fixes https://github.com/solana-labs/example-helloworld/issues/141

#### Before:
```
$ solana-test-validator -r 
Mint address: vines1vzrYbzLMRdu58ou5XTby4qAqVRLmqo36NKPTg
Ledger location: test-ledger
Log: test-ledger/validator.log
$
```
❓ 

#### After:
```
$ solana-test-validator 
Mint address: vines1vzrYbzLMRdu58ou5XTby4qAqVRLmqo36NKPTg
Ledger location: test-ledger
Log: test-ledger/validator.log
Error: failed to start faucet: Unable to bind faucet to 0.0.0.0:9900, check the address is not already in use: Address already in use (os error 48)
$
```
💡

### DIFF
--- a/bench-exchange/tests/bench_exchange.rs
+++ b/bench-exchange/tests/bench_exchange.rs
@@ -5,7 +5,7 @@ use solana_core::validator::ValidatorConfig;
 use solana_exchange_program::exchange_processor::process_instruction;
 use solana_exchange_program::id;
 use solana_exchange_program::solana_exchange_program;
-use solana_faucet::faucet::run_local_faucet;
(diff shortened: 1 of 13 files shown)


CriesofCarrots:
Ohai, I just ran into this! Thx for the fix.

PR #14914, opened 2021-01-29
Do not download and execute binaries via HTTP

leoluk:
This fixes a couple of trivial remote code execution opportunities.


mvines (approved):
Thanks!  That s is kinda important

PR #14922, opened 2021-01-29
Add validator flag to opt in to cpi and logs storage

CriesofCarrots:
#### Problem
Some api nodes are experiencing stalls while rocksdb compaction is running. Compaction of the TransactionStatus columns seems to take a particularly long time, and seems related to the extremely large sizes of some transaction meta, esp logs.

#### Summary of Changes
As a short-term fix, add a parameter to `solana-validator` to allow nodes to opt-in to storing the transaction meta that isn't critical for exchange-/custody-like services, namely: transaction logs and CPI inner instructions.
Without this parameter, TransactionStatus data should be significantly smaller for many transactions.

### DIFF
--- a/ledger/src/blockstore_processor.rs
+++ b/ledger/src/blockstore_processor.rs
@@ -1093,8 +1093,8 @@ pub struct TransactionStatusBatch {
     pub statuses: Vec<TransactionExecutionResult>,
     pub balances: TransactionBalancesSet,
     pub token_balances: TransactionTokenBalancesSet,
-    pub inner_instructions: Vec<Option<InnerInstructionsList>>,
-    pub transaction_logs: Vec<TransactionLogMessages>,
+    pub inner_instructions: Option<Vec<Option<InnerInstructionsList>>>,
+    pub transaction_logs: Option<Vec<TransactionLogMessages>>,
 }
 
 #[derive(Clone)]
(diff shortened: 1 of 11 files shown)


CriesofCarrots:
@sakridge does v1.5 seem sufficient for the users you've been supporting?

validator/src/main.rs
```diff
@@ -829,7 +829,7 @@ pub fn main() {
     let default_rpc_threads = num_cpus::get().to_string();
 
     let matches = App::new(crate_name!()).about(crate_description!())
-        .version(solana_version::version!())
+        // .version(solana_version::version!())
```
mvines, line 832: 👋🏼

CriesofCarrots: 🤕 Thank you!

ledger/src/blockstore_processor.rs
```diff
@@ -1106,21 +1110,28 @@ pub fn send_transaction_status_batch(
...
-        iteration_order,
-        statuses,
-        balances,
-        token_balances,
-        inner_instructions,
-        transaction_logs,
-    }) {
+    if !transaction_status_sender.enable_cpi_and_log_storage {
+        inner_instructions = inner_instructions.iter().map(|_| None).collect();
+        transaction_logs = transaction_logs.iter().map(|_| vec![]).collect();
```
carllin, line 1120: I was curious to see if there was way to avoid these `collect()`'s. This was the best I could come up with: https://github.com/carllin/solana/commit/4bde604f1f4e633e2d8c0ed30e0372b396ae1362

CriesofCarrots: Oh hey, that's not bad! I tried something similar, but didn't think to use an Option for the `send()`. Thx!

sakridge:
> 
> 
> @sakridge does v1.5 seem sufficient for the users you've been supporting?

Yea, 1.5 should be fine.

sakridge (approved):
Looks good to me. Thanks for the quick work. I'm a bit curious how much data it saves.

My guess is it should be fine to make it the new default to not collect this information, but we should make sure it's in the release notes.

carllin (approved):

PR #14923, opened 2021-01-29
Buffer authority must match upgrade authority for deploys and upgrades

jackcmay:
#### Problem

Buffer account can be recreated after its authority is marked `None`

#### Summary of Changes

- Enforce that buffer authority matches upgrade authority during deploys / upgrades
- Deploys now require an upgrade authority to be specified as a signer
- Buffer state still has an optional authority but restrict everyone from setting it to `None`
- Remove the tooling around buffer write `--final`

The new CLI is also compatible with the existing upgradeable loader and the work flow is the same except:
- Cannot set the buffer's authority to `None` via --final (using either `write-buffer` or `set-buffer-authority`
- When deploying cannot specify the upgrade authority by public key, need to provide a keypair or use the default keypair

Fixes #

### DIFF
--- a/programs/bpf_loader/src/lib.rs
+++ b/programs/bpf_loader/src/lib.rs
@@ -299,15 +299,26 @@ fn process_loader_upgradeable_instruction(
     match limited_deserialize(instruction_data)? {
         UpgradeableLoaderInstruction::InitializeBuffer => {
             let buffer = next_keyed_account(account_iter)?;
-            let authority = next_keyed_account(account_iter)?;
 
             if UpgradeableLoaderState::Uninitialized != buffer.state()? {
                 ic_logger_msg!(logger, "Buffer account already initialized");
                 return Err(InstructionError::AccountAlreadyInitialized);
             }
-            buffer.set_state(&UpgradeableLoaderState::Buffer {
-                authority_address: Some(*authority.unsigned_key()),
-            })?;
+
+            if invoke_context.is_feature_active(&matching_buffer_upgrade_authorities::id()) {
+                let authority = next_keyed_account(account_iter)?;
+
+                buffer.set_state(&UpgradeableLoaderState::Buffer {
+                    authority_address: Some(*authority.unsigned_key()),
+                })?;
+            } else {
+                let authority = next_keyed_account(account_iter)
+                    .ok()
+                    .map(|account| account.unsigned_key());
+                buffer.set_state(&UpgradeableLoaderState::Buffer {
+                    authority_address: authority.cloned(),
+                })?;
+            }
         }
         UpgradeableLoaderInstruction::Write { offset, bytes } => {
             let buffer = next_keyed_account(account_iter)?;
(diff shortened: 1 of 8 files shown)


sdk/program/src/bpf_loader_upgradeable.rs
```diff
@@ -103,29 +99,32 @@ pub fn create_buffer(
             UpgradeableLoaderState::buffer_len(program_len)? as u64,
             &id(),
         ),
-        Instruction::new(id(), &UpgradeableLoaderInstruction::InitializeBuffer, metas),
+        Instruction::new(
+            id(),
+            &UpgradeableLoaderInstruction::InitializeBuffer,
+            vec![
+                AccountMeta::new(*buffer_address, false),
+                AccountMeta::new(*authority_address, false),
```
jstarry, line 107:
```suggestion
                AccountMeta::new_readonly(*authority_address, false),
```

sdk/program/src/bpf_loader_upgradeable.rs
```diff
@@ -103,29 +99,32 @@ pub fn create_buffer(
...
-    if let Some(authority_address) = authority_address {
-        metas[1] = AccountMeta::new(*authority_address, true);
-    }
     Instruction::new(
         id(),
         &UpgradeableLoaderInstruction::Write { offset, bytes },
-        metas,
+        vec![
+            AccountMeta::new(*buffer_address, false),
+            AccountMeta::new(*authority_address, true),
```
jstarry, line 126:
```suggestion
            AccountMeta::new_readonly(*authority_address, true),
```

runtime/src/loader_utils.rs
```diff
@@ -89,14 +91,14 @@ pub fn load_buffer_account<T: Client>(
...
                 &buffer_pubkey,
-                None,
+                &buffer_authority_pubkey,
                 offset,
                 chunk.to_vec(),
             )],
             Some(&from_keypair.pubkey()),
         );
         bank_client
-            .send_and_confirm_message(&[from_keypair, &buffer_keypair], message)
```
jstarry, line 99: Oh did the bank client not fail if you try to sign with an unneeded keypair?

programs/bpf_loader/src/lib.rs
```diff
@@ -1927,7 +1981,50 @@ mod tests {
...
-                .send_and_confirm_message(&[&mint_keypair, &program_keypair], message)
+                .send_and_confirm_message(
+                    &[&mint_keypair, &program_keypair, &upgrade_authority_keypair],
+                    message
+                )
+                .unwrap_err()
+                .unwrap()
+        );
+
+        // Mismatched buffer and program authority
```
jstarry, line 1992: Could you also add a test that upgrades/deploys fail when buffer authority is `None`?

jstarry (dismissed):
Looks good!

runtime/src/loader_utils.rs
jackcmay: Writes require the buffer authority, it was just that the buffer authority was the buffer_keypair.

PR #14934, opened 2021-01-29
Prevent bricked install when ^C is pressed during archive extraction

mvines:
Hitting ^C while solana-install update is extracting the new channel archive can brick your install. Fix it

cc: https://github.com/solana-labs/solana/pull/14919

### DIFF
--- a/install/src/command.rs
+++ b/install/src/command.rs
@@ -158,13 +158,22 @@ fn extract_release_archive(
     let progress_bar = new_spinner_progress_bar();
     progress_bar.set_message(&format!("{}Extracting...", PACKAGE));
 
-    let _ = fs::remove_dir_all(extract_dir);
-    fs::create_dir_all(extract_dir)?;
+    if extract_dir.exists() {
+        let _ = fs::remove_dir_all(&extract_dir);
+    }
+
+    let tmp_extract_dir = extract_dir.with_file_name("tmp-extract");
+    if tmp_extract_dir.exists() {
+        let _ = fs::remove_dir_all(&tmp_extract_dir);
+    }
+    fs::create_dir_all(&tmp_extract_dir)?;
 
     let tar_bz2 = File::open(archive)?;
     let tar = BzDecoder::new(BufReader::new(tar_bz2));
     let mut release = Archive::new(tar);
-    release.unpack(extract_dir)?;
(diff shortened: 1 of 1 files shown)


install/src/command.rs
```diff
@@ -158,13 +158,22 @@ fn extract_release_archive(
...
+    }
+    fs::create_dir_all(&tmp_extract_dir)?;
 
     let tar_bz2 = File::open(archive)?;
     let tar = BzDecoder::new(BufReader::new(tar_bz2));
     let mut release = Archive::new(tar);
-    release.unpack(extract_dir)?;
+    release.unpack(&tmp_extract_dir)?;
+
+    fs::rename(&tmp_extract_dir, extract_dir)?;
```
mvines, line 176: 💥

install/src/command.rs
```diff
@@ -786,44 +795,59 @@ pub fn update(config_file: &str) -> Result<bool, String> {
             ExplicitRelease::Semver(release_semver) => {
                 let download_url = github_release_download_url(release_semver);
                 let release_dir = config.release_dir(&release_semver);
-                let download_url = if release_dir.join(".ok").exists() {
+                let download_url = if release_dir.exists() {
```
mvines, line 798: No more ".ok" junk needed because the release dir now only exists once extraction has fully completed

install/src/command.rs
```diff
@@ -786,44 +795,59 @@ pub fn update(config_file: &str) -> Result<bool, String> {
...
-                let release_dir = config.release_dir(&release_channel);
+                let version_url = release_channel_version_url(release_channel);
+
+                let (_temp_dir, temp_file, _temp_archive_sha256) =
+                    download_to_temp(&version_url, None)
+                        .map_err(|err| format!("Unable to download {}: {}", version_url, err))?;
+
+                let update_release_version = load_release_version(&temp_file)?;
+
+                let release_id = format!("{}-{}", release_channel, update_release_version.commit);
```
mvines, line 817: Stop reusing the channel name for the install directory.  Instead use "channel-commit".  In addition to avoiding the ^C issue, this gives us a backdoor to revert to an older channel build if you just upgraded to broken one (by messing around manually in `~/.local/share/solana/install/`.   The old/new commit is also printed to help with this

PR #14940, opened 2021-01-29
Only store the shred payload

sakridge:
#### Problem

We store the whole shred payload in the database, but only the actual payload is important

#### Summary of Changes

Only store the shred data.

Fixes #


sakridge:
#16602 fixed it

PR #14943, opened 2021-01-30
solana-validator: Add "init" subcommand

mvines:
There's no nice way to bootstrap a validator's ledger directory then exit.   This is a problem when running a validator in `--no-snapshot-fetch` mode (which is really best in general), but then you need to reinitialize the ledger directory due to some reason (fresh install, you fall 100k slots behind).  

The current workflow is restart the validator without `--no-snapshot-fetch` and watch the log as it downloads a new snapshot, then ^C at the right time, and restart the validator normally.   This is clunky.

Instead run `solana-validator <all normal args except for --no-snapshot-fetch> init` and the validator will exit itself at the right time.  No ^C required.

### DIFF
--- a/validator/src/main.rs
+++ b/validator/src/main.rs
@@ -1,6 +1,6 @@
 use clap::{
-    crate_description, crate_name, value_t, value_t_or_exit, values_t, values_t_or_exit, App, Arg,
-    ArgMatches,
+    crate_description, crate_name, value_t, value_t_or_exit, values_t, values_t_or_exit, App,
+    AppSettings, Arg, ArgMatches, SubCommand,
 };
 use log::*;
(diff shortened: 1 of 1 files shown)


mvines:
@ryoqun - if I recall correctly, you had originally suggested this functionally.  Does this align with what you were thinking of?

validator/src/main.rs
```diff
@@ -760,54 +766,6 @@ fn rpc_bootstrap(
     }
 }
 
-#[allow(clippy::too_many_arguments)]
```
ryoqun, line 763: bonus for removing this somewhat dubious thin wrapper fn. :)

ryoqun (approved):
LGTM!

Yeah, this is exactly what I wanted. :)

PR #14944, opened 2021-01-30
Nam request for Inflation Candidate nomination

nampdn:
#### Problem
Submit my nomination candidate request

#### Summary of Changes
Add pairs of nomination public keys


mvines (approved):
PR looks great.  I like how you ground out the `Nam..` keypair :)

PR #14954, opened 2021-01-31
fix test check

jeffwashington:
#### Problem
Test code has a bug.
#### Summary of Changes
Fix bug in test code, fix tests that then fail.
Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -5816,8 +5816,8 @@ pub mod tests {
         expected_lamports: u64,
     ) {
         let ancestors = vec![(slot, 0)].into_iter().collect();
-        let (account, slot) = accounts.load_slow(&ancestors, &pubkey).unwrap();
-        assert_eq!((account.lamports, slot), (expected_lamports, slot));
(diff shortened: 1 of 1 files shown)


jeffwashington:
@sakridge , I think this is a bug in the test code. The specified slot parameter looks like it was intended to be compared, but it isn't being compared.
Assuming this was a bug in the test code, there are a few failing tests that I do not yet know how to verify as being correct.

sakridge:
I think the original is correct actually.

runtime/src/accounts_db.rs
```diff
@@ -5816,8 +5816,8 @@ pub mod tests {
         expected_lamports: u64,
     ) {
         let ancestors = vec![(slot, 0)].into_iter().collect();
-        let (account, slot) = accounts.load_slow(&ancestors, &pubkey).unwrap();
-        assert_eq!((account.lamports, slot), (expected_lamports, slot));
```
jeffwashington, line 5820: @sakridge, Is the comparison of slot against itself here solely to cause the assert to print the slot at which it was found on failure?

PR #14957, opened 2021-02-01
Don't load all accounts into memory for capitalization check

carllin:
#### Problem
On Kin simulation cluster, trying to create a new snapshot through ledger tool took almost 4 hours, due to the capitalization check at the end of `do_process_blockstore_from_root()` pulling all the 80M+ accounts into memory, hitting swap.

The `update_accounts_hash() -> calculate_accounts_hash()` in `bank_to_snapshot_archive()` might also be running into a similar problem, taking almost 50 mins (faster because there's more parallel threads), but  that's not addressed in this PR as its not as flagrantly bad (not pulling the entire accounts structure in, just the hashes).

#### Summary of Changes
1) Accumulate capitalization iteratively in scan.
2) Remove cloning of accounts in scan.

@CriesofCarrots  fyi if you're trying to make any custom snapshots on the Kin cluster.

Fixes #

### DIFF
--- a/runtime/src/accounts.rs
+++ b/runtime/src/accounts.rs
@@ -547,19 +547,26 @@ impl Accounts {
         ancestors: &Ancestors,
         simple_capitalization_enabled: bool,
     ) -> u64 {
-        let balances =
-            self.load_all_unchecked(ancestors)
-                .into_iter()
-                .map(|(_pubkey, account, _slot)| {
-                    AccountsDB::account_balance_for_capitalization(
+        self.accounts_db.unchecked_scan_accounts(
+            ancestors,
+            |total_capitalization: &mut u64, some_account_tuple| {
+                if let Some((_pubkey, account, _slot)) =
+                    some_account_tuple.filter(|(_, account, _)| Self::is_loadable(account))
+                {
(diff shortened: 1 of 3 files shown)


carllin:
Hmmm, this actually doesn't seem to be the biggest issue, seems like just loading the AccountsIndex from a snapshot blows out the 64 GB limit: 

Screen Shot 2021-01-31 at 7 50 42 PM

ryoqun:
> Im guessing the AccountsIndex vectors may be overallocated, b/c based on my calculations, each entry if there is only one account update should be roughly 80 million * 48 bytes/entry ~= 3GB

@carllin Well, I think we should also consider various smart pointers/locks. IIRC, a single AccountsIndex entry also incurs 1 Arc (which is around 24byte), 1 RwLock (~70byte). (I assumed your 48 came from AccountInfo + ref_count.)
Also, BTreeMap overhead (at least 8 bytes). Off course, the AccountsIndex slot list can still be overallocated, though. :)

Also, I think `heaptrack` can easily tell why it's taking whopping 64GB in this case.

t-nelson (approved):
Seems like this one outgrew it's title a bit :slightly_smiling_face: 

LGTM!

PR #14978, opened 2021-02-01
program-test: Add ability to warp to the future

joncinque:
#### Problem

Program-test is very useful for end-to-end testing many situations, but it doesn't have the ability to warp to a slot in the future to test clock-dependent programs.

#### Summary of Changes

Add it in!  There's a bit of extra plumbing required to sync up all parts of banks-server, namely giving ProgramTest access to BankForks and BlockCommitmentCache.

The main test added is just looking at the clock, but we can add more tests for things like stake rewards and rent collection.  I prefer to do it in a separate PR.

There's a change to clear the invoke context at the end of each instruction processing, and I'm not totally sure of the impact that might have elsewhere.  It might break someone adding two of their own programs to ProgramTest and CPI'ing from one to the other.  Note that it was required to get the warp test working.  The fuzz test also triggers the issue in a simpler context, which is why it's updated in the same PR.  Alternatively, I can change that logic to just always set the invoke context, and not panic if it's changed.

cc @mvines

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4773,6 +4773,7 @@ dependencies = [
  "solana-program 1.6.0",
  "solana-runtime",
  "solana-sdk",
+ "thiserror",
  "tokio 0.3.5",
 ]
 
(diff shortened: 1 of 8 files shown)


joncinque:
I did some more testing with CPIs, and program-test needs to be fixed for both situations, so I'll split this into one PR for the fixes and another for the warping specifically

PR #14980, opened 2021-02-01
streamline calculate_accounts_hash

jeffwashington:
#### Problem
calculate_accounts_hash uses the index to iterate accounts. The results are thus already sorted in pubkey ascending order, which is what merkle root calculations expect. So, no need to sort, and no need to copy around pubkeys.

#### Summary of Changes
simplify calculate_accounts_hash to pass around less data and execute more quickly.

Fixes #


jeffwashington:
running on colo machine connected to mainnet beta.
old code: 5.7s
this code: 4.9s
lower memory usage

sakridge (approved):
looks good

PR #14981, opened 2021-02-01
Cli: some moniker follow-up

CriesofCarrots:
#### Problem
The cli url monikers are awesome! But if you try to run `solana config set -u d` or similar, you get scary panics next time you try to use the cli (`thread 'main' panicked at 'no nonempty setting', cli/src/cli.rs:450:14`)

Also, `solana logs -u d` doesn't work with a passed moniker. It actually attempts to connect to whatever you have set in your config, instead of `wss://devnet.solana.com`.

#### Summary of Changes
- Enable monikers in `solana config set`
- Fixup websocket compute to handle monikers properly

### DIFF
--- a/cli/src/cli.rs
+++ b/cli/src/cli.rs
@@ -471,11 +471,15 @@ impl CliConfig<'_> {
             (SettingType::Explicit, websocket_cfg_url.to_string()),
             (
                 SettingType::Computed,
-                solana_cli_config::Config::compute_websocket_url(json_rpc_cmd_url),
+                solana_cli_config::Config::compute_websocket_url(&normalize_to_url_if_moniker(
+                    json_rpc_cmd_url,
+                )),
             ),
             (
                 SettingType::Computed,
-                solana_cli_config::Config::compute_websocket_url(json_rpc_cfg_url),
+                solana_cli_config::Config::compute_websocket_url(&normalize_to_url_if_moniker(
+                    json_rpc_cfg_url,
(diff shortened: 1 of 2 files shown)


cli/src/cli.rs
```diff
@@ -471,11 +471,15 @@ impl CliConfig<'_> {
...
+                solana_cli_config::Config::compute_websocket_url(&normalize_to_url_if_moniker(
+                    json_rpc_cmd_url,
+                )),
             ),
             (
                 SettingType::Computed,
-                solana_cli_config::Config::compute_websocket_url(json_rpc_cfg_url),
+                solana_cli_config::Config::compute_websocket_url(&normalize_to_url_if_moniker(
+                    json_rpc_cfg_url,
+                )),
```
CriesofCarrots, line 482: This change enables users to store `json_rpc_url: d` or similar in the config file and parse properly (without those scary panics), so technically https://github.com/solana-labs/solana/pull/14981/commits/c3ab909d882b1c0a75cc9315eb80a799188713e0 isn't required. However, I prefer storing the verbose url in the config file instead of the moniker.

CriesofCarrots:
I don't think this should be objectionable, but post-merge comments welcome, @ryoqun !

ryoqun:
> I don't think this should be objectionable, but post-merge comments welcome, @ryoqun !

yeah! no objection. Thanks for love to the monikers :)

PR #14987, opened 2021-02-02
Fix pubkey refcount for shrink + clean

carllin:
#### Problem
Shrink doesn't unref accounts that it removes. For example:

1. Create two keys, `pubkey1`, `pubkey2` in slot 0
2. Updates `pubkey1` slot 1
3. Run clean + shrink on slot 0 so that it now only has 1 of those keys in the AppendVec
4. Updates the other key from slot 0 in slot 2
5. Run clean -> remove slot 0  via `process_dead_slots()`-> unref
6. Expect ref count to be 1 for `pubkey1`, but it's still 2.

#### Summary of Changes
Encode the above into a test.
@jeffwashington added a test here to illustrate setting up a clean + shrink. The bug is the ref count issue we discussed in the morning
Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -50,7 +50,7 @@ use solana_vote_program::vote_state::MAX_LOCKOUT_HISTORY;
 use std::{
     borrow::Cow,
     boxed::Box,
-    collections::{BTreeMap, BTreeSet, HashMap, HashSet},
+    collections::{hash_map::Entry, BTreeMap, BTreeSet, HashMap, HashSet},
     convert::{TryFrom, TryInto},
     io::{Error as IOError, Result as IOResult},
     ops::RangeBounds,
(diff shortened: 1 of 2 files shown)


jeffwashington:
I can't claim to fully understand all of this, but it looks reasonable. I imagine I will take the test you wrote and re-adapt the tests I had written. Is that the idea?

carllin:
> I can't claim to fully understand all of this, but it looks reasonable. I imagine I will take the test you wrote and re-adapt the tests I had written. Is that the idea?

Yeah, hopefully this illustrates the setup of a clean + shrink, but feel free to reach out with any additional clarifications!

runtime/src/accounts_db.rs
```diff
@@ -1636,10 +1641,17 @@ impl AccountsDB {
                     )| {
                         if let Some((locked_entry, _)) = self.accounts_index.get(pubkey, None, None)
                         {
-                            locked_entry
+                            let is_alive = locked_entry
                                 .slot_list()
                                 .iter()
-                                .any(|(_slot, i)| i.store_id == *store_id && i.offset == *offset)
+                                .any(|(_slot, i)| i.store_id == *store_id && i.offset == *offset);
+                            if !is_alive {
```
sakridge, line 1648: Does it apply to the other path? Or is it because this can operate on sub-slot stores?

runtime/src/accounts_db.rs
```diff
@@ -1636,10 +1641,17 @@ impl AccountsDB {
...
+                                // If the account index entry was removed, that means no other
+                                // AppendVec's for this slot contains this pubkey, so it's safe
+                                // to unref
+                                locked_entry.unref()
```
sakridge, line 1652: This comment confused me.

How about this:
This pubkey was found in the storage, but no longer exists in the index. It would have had a ref to the storage from the initial store, but it will not exist in the re-written slot. Unref it to keep the index consistent with removing the storage entries.

What if there are two dead entries in the slot for the same pubkey? Would they be unref'ed twice?

PR #14993, opened 2021-02-02
[web3] feat: add support for browser es modules

jstarry:
#### Problem
- web3.js doesn't provide a browser compatible esmodule
- web3.js doesn't bundle the buffer shim into the iife module
- keccak is not browser friendly (relies on `'stream'`)

#### Summary of Changes
- provide browser es module
- provide unminified browser script bundle
- replace keccak dependency with js-sha3
- remove unnecessary dependencies
- update babel dependencies / config
- fix `Buffer` shim for browser bundles
- remove `AgentManager` from browser bundles with tree shaking
- add browser tests (migrate from jest to mocha)
- cleanup flow types
- cleanup package.json

Fixes https://github.com/solana-labs/solana-web3.js/issues/1009


jstarry:
Tested on:
- browser with iife bundle
- minimal angular app
- explorer
- break client/server

jstarry:
I checked this pretty thoroughly but fairly likely this will break something, @mvines or @oJshua want to take a look over this?

oJshua:
> * keccak is not browser friendly (relies on `'stream'`)

Oh, that's interesting -- project page said it supported browser. `js-sha3` is probably a better choice. I'll have a closer look!

jstarry:
> Oh, that's interesting -- project page said it supported browser

Yeah, I thought so too.. Maybe they assume devs bring their own polyfill for stream? I'm not really sure

jstarry:
Ok, definitely going to need browser tests to have more confidence in the browser bundles we're providing. I think it's worth waiting for that before merging.

PR #14997, opened 2021-02-02
program-test: Set context without panic

joncinque:
#### Problem

Program-test currently has two issues due to the thread_local InvokeContext getting set twice:
* a transaction with multiple instructions
* a CPI from one program-test program into another

#### Summary of Changes

Remove the panic and add tests for multiple instructions and for CPI

Fixes #


jackcmay (approved):

joncinque:
Closing and reopening for CI

PR #15009, opened 2021-02-02
Cleanup v1 shrink path

sakridge:
#### Problem

v1 shrink functions are confusing because no explanation of v1 is or what the difference is between non-v1.

#### Summary of Changes

* Move legacy v1 functions to another impl block
* Move cfg(test) functions to own impl block
* add some comments.

Fixes #


carllin (commented):
Awesome, thanks!

PR #15013, opened 2021-02-02
Bump tonic, prost, tarpc, tokio

### LINKED ISSUE #10187: Update tokio from 0.1.x to 1.x with async fn

CriesofCarrots:
#### Problem
Really old tokio is used throughout the code base, blocking upgrades of other dependencies.

#### Summary of Changes
- Update version of tonic and various related dependencies. Unfortunately, since there is no published version of `goauth` on tokio v0.3, tokio must be bumped to v1.0

Todo:
- [x] Rebase on #15018 and see what kind of tokio pinning is still necessary for jsonrpc
- [x] Update remaining tokio in code base to v1.0 - edit: only tarpc-related updates remaining

Fixes #10187
...as much as possible atm. We're dependent on the jsonrpc crates to eliminate tokio v0.2 entirely

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -76,8 +76,9 @@ solana-vote-program = { path = "../programs/vote", version = "1.6.0" }
 spl-token-v2-0 = { package = "spl-token", version = "=3.0.1", features = ["no-entrypoint"] }
 tempfile = "3.1.0"
 thiserror = "1.0"
-tokio = { version = "0.2", features = ["full"] }
-tokio-util = { version = "0.2", features = ["codec"] }
+tokio = { version = "0.3", features = ["full"] }
+tokio_02 = { version = "0.2", package = "tokio", features = ["full"] }
+tokio-util = { version = "0.3", features = ["codec"] }
 solana-rayon-threadlimit = { path = "../rayon-threadlimit", version = "1.6.0" }
 trees = "0.2.1"
 
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -76,7 +76,7 @@ solana-vote-program = { path = "../programs/vote", version = "1.6.0" }
 spl-token-v2-0 = { package = "spl-token", version = "=3.0.1", features = ["no-entrypoint"] }
 tempfile = "3.1.0"
 thiserror = "1.0"
-tokio = { version = "0.3", features = ["full"] }
+tokio = { version = "1.1", features = ["full"] }
 tokio_02 = { version = "0.2", package = "tokio", features = ["full"] }
 tokio-util = { version = "0.3", features = ["codec"] }
 solana-rayon-threadlimit = { path = "../rayon-threadlimit", version = "1.6.0" }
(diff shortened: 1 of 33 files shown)


core/Cargo.toml
```diff
@@ -76,8 +76,9 @@ solana-vote-program = { path = "../programs/vote", version = "1.6.0" }
 spl-token-v2-0 = { package = "spl-token", version = "=3.0.1", features = ["no-entrypoint"] }
 tempfile = "3.1.0"
 thiserror = "1.0"
-tokio = { version = "0.2", features = ["full"] }
-tokio-util = { version = "0.2", features = ["codec"] }
+tokio = { version = "1.1", features = ["full"] }
```
ryoqun, line 79: :100:

CriesofCarrots:
This is ready for a proper look. I tested BigTable, rpc, and faucet with both TestValidator and multinode-demo scripts, and all seems well.

@mvines , downstream projects will need to update to tokio v1 to use solana-program-test with these changes (presumably 1.6). Does it seem reasonable to update `build-downstream-projects.sh` to bump the tokio version for CI?

mvines:
> Edit, looks like this needs to be in `patch.crates-io.sh`/`update-solana-dependencies.sh` in solana-program-library. Does reading the solana tokio version and bumping it there seem reasonable?

Will we need to adapt the SPL repo to any of the tokio API changes from 0.3 to 1.1?  We could just add a feature flag in SPL to build against tokio 0.3 or 1.1, then select a tokio from `build-downstream-projects.sh`

This looks awesome overall.  I find the continual API changes in tokio very frustrating

CriesofCarrots:
> Will we need to adapt the SPL repo to any of the tokio API changes from 0.3 to 1.1? We could just add a feature flag in SPL to build against tokio 0.3 or 1.1, then select a tokio from `build-downstream-projects.sh`

No, right now we don't need to adapt to any api changes. Our uses of tokio in SPL repo are all consistent between v0.3 and v1.
Tokio in SPL repo just needs to be consistent with the version in solana-program-test. Having looked at the scripts more closely, I think it makes sense to be in `patch.crates-io.sh`, since that's for any patch to local solana monorepo. I'll pr that and see what you think.

> This looks awesome overall. I find the continual API changes in tokio very frustrating

💯 💯

CriesofCarrots:
@mvines Ci == green; wdyt?

mvines (approved):

PR #15017, opened 2021-02-02
cli: add query command to solana program

jackcmay:
#### Summary of Changes

Add `solana program get-info` command to the cli that looks up and returns information about an upgradeable program or an upgradeable buffer account.

Fixes #

### DIFF
--- a/cli-output/src/cli_output.rs
+++ b/cli-output/src/cli_output.rs
@@ -1528,6 +1528,100 @@ impl fmt::Display for CliProgramAuthority {
     }
 }
 
+#[derive(Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct CliUpgradeableProgram {
+    pub program_id: String,
+    pub program_executable: bool,
+    pub program_lamports: u64,
+    pub programdata_address: String,
+    pub programdata_lamports: u64,
+    pub programdata_authority: String,
+    pub programdata_slot: u64,
+    pub programdata_data_len: usize,
+    pub programdata_program_len: usize,
+    pub use_lamports_unit: bool,
+}
+impl QuietDisplay for CliUpgradeableProgram {}
+impl VerboseDisplay for CliUpgradeableProgram {}
+impl fmt::Display for CliUpgradeableProgram {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        writeln!(f)?;
+        writeln_name_value(f, "Program Address:", &self.program_id)?;
+        writeln_name_value(f, "  Executable:", &self.program_executable.to_string())?;
+        writeln_name_value(
+            f,
+            "  Balance:",
+            &build_balance_message(self.program_lamports, self.use_lamports_unit, true),
+        )?;
+        writeln_name_value(f, "  ProgramData Address:", &self.programdata_address)?;
+        writeln_name_value(f, "    Authority:", &self.programdata_authority)?;
+        writeln_name_value(
+            f,
+            "    Balance:",
+            &build_balance_message(self.programdata_lamports, self.use_lamports_unit, true),
+        )?;
+        writeln_name_value(
+            f,
+            "    Total Length:",
+            &format!(
+                "{:?} ({:#x?}) bytes",
+                self.programdata_data_len, self.programdata_data_len
+            ),
+        )?;
+        writeln_name_value(
+            f,
+            "    Program Length:",
+            &format!(
+                "{:?} ({:#x?}) bytes",
+                self.programdata_program_len, self.programdata_program_len
+            ),
+        )?;
+        writeln_name_value(
+            f,
+            "    Last Upgraded In Slot:",
+            &self.programdata_slot.to_string(),
+        )?;
+        Ok(())
+    }
+}
+
+#[derive(Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct CliUpgradeableBuffer {
+    pub address: String,
+    pub lamports: u64,
+    pub authority: String,
+    pub data_len: usize,
+    pub program_len: usize,
+    pub use_lamports_unit: bool,
+}
+impl QuietDisplay for CliUpgradeableBuffer {}
+impl VerboseDisplay for CliUpgradeableBuffer {}
+impl fmt::Display for CliUpgradeableBuffer {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        writeln!(f)?;
+        writeln_name_value(f, "Buffer Address:", &self.address)?;
+        writeln_name_value(f, "  Authority:", &self.authority)?;
+        writeln_name_value(
+            f,
+            "  Balance:",
+            &build_balance_message(self.lamports, self.use_lamports_unit, true),
+        )?;
+        writeln_name_value(
+            f,
+            "  Total Length:",
+            &format!("{:?} ({:#x?}) bytes", self.data_len, self.data_len),
+        )?;
+        writeln_name_value(
+            f,
+            "  Program Length:",
+            &format!("{:?} ({:#x?}) bytes", self.program_len, self.program_len),
+        )?;
+        Ok(())
+    }
+}
+
 pub fn return_signers(
     tx: &Transaction,
     output_format: &OutputFormat,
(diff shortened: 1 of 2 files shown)


jackcmay:
@CriesofCarrots This command supports solana_cli_output, could you look it over to make sure it looks good?

cli-output/src/cli_output.rs
```diff
@@ -1528,6 +1528,100 @@ impl fmt::Display for CliProgramAuthority {
...
+    pub program_id: String,
+    pub program_executable: bool,
+    pub program_lamports: u64,
+    pub programdata_address: String,
+    pub programdata_lamports: u64,
+    pub programdata_authority: String,
+    pub programdata_slot: u64,
+    pub programdata_data_len: usize,
+    pub programdata_program_len: usize,
+    pub use_lamports_unit: bool,
```
CriesofCarrots, line 1543: Tag this field with `#[serde(skip_serializing)]` so it doesn't show up in the json struct

cli-output/src/cli_output.rs
```diff
@@ -1528,6 +1528,100 @@ impl fmt::Display for CliProgramAuthority {
...
+pub struct CliUpgradeableProgram {
+    pub program_id: String,
+    pub program_executable: bool,
+    pub program_lamports: u64,
+    pub programdata_address: String,
+    pub programdata_lamports: u64,
+    pub programdata_authority: String,
+    pub programdata_slot: u64,
+    pub programdata_data_len: usize,
+    pub programdata_program_len: usize,
```
CriesofCarrots, line 1542:
```suggestion
    pub executable: bool,
    pub lamports: u64,
    pub program_data_address: String,
    pub program_data_lamports: u64,
    pub program_data_authority: String,
    pub program_data_slot: u64,
    pub program_data_len: usize,
    pub program_len: usize,
```

cli-output/src/cli_output.rs
```diff
@@ -1528,6 +1528,100 @@ impl fmt::Display for CliProgramAuthority {
...
+
+#[derive(Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct CliUpgradeableBuffer {
+    pub address: String,
+    pub lamports: u64,
+    pub authority: String,
+    pub data_len: usize,
+    pub program_len: usize,
+    pub use_lamports_unit: bool,
```
CriesofCarrots, line 1597: Ditto re `#[serde(skip_serializing)]`

cli/src/program.rs
```diff
@@ -266,6 +270,24 @@ impl ProgramSubCommands for App<'_, '_> {
                                 .help("Public key of the account to query")
                         ),
                 )
+                .subcommand(
+                    SubCommand::with_name("get-info")
```
CriesofCarrots, line 274: There was a big push a while ago to strip out all the `get-`s from cli subcommands. I know it has already snuck back in here with "get-authority", but it would be nice to avoid introducing more. Can this just be "info"?
I think "get-authority" can just be "authority" as well (although we can leave "get-authority" as an alias, if you think anyone is using it already.

cli/src/program.rs
```diff
@@ -266,6 +270,24 @@ impl ProgramSubCommands for App<'_, '_> {
...
+                        .about("Display a buffer or program")
+                        .arg(
+                            Arg::with_name("account")
+                                .index(1)
+                                .value_name("ACCOUNT_ADDRESS")
+                                .takes_value(true)
+                                .required(true)
+                                .help("Public key of the buffer or program to query")
```
CriesofCarrots, line 282:
```suggestion
                                .help("Address of the buffer or program to query")
```

cli/src/program.rs
```diff
@@ -266,6 +270,24 @@ impl ProgramSubCommands for App<'_, '_> {
                                 .help("Public key of the account to query")
```
CriesofCarrots, line 270:
```suggestion
                                .help("Address of the account to query")
```

cli-output/src/cli_output.rs
CriesofCarrots: These work a lot better in camel-case json

CriesofCarrots: I wonder if `program_data_slot` ought to be more descriptive. `last_upgrade_slot`? (does it need the program_data prefix?)

cli/src/program.rs
```diff
@@ -937,6 +970,82 @@ fn process_get_authority(
...
+                                program_id: account_pubkey.to_string(),
+                                program_executable: account.executable,
+                                program_lamports: account.lamports,
+                                programdata_address: programdata_address.to_string(),
+                                programdata_lamports: programdata_account.lamports,
+                                programdata_authority: upgrade_authority_address
+                                    .map(|pubkey| pubkey.to_string())
+                                    .unwrap_or_else(|| "none".to_string()),
+                                programdata_slot: slot,
+                                programdata_data_len: programdata_account.data.len(),
```
CriesofCarrots, line 1009: Is this data needed? Might be slightly less confusing to only have one length reported, and I can get this info by calling `solana account <programdata_address>`

cli-output/src/cli_output.rs
```diff
@@ -1528,6 +1528,100 @@ impl fmt::Display for CliProgramAuthority {
...
+    pub programdata_program_len: usize,
+    pub use_lamports_unit: bool,
+}
+impl QuietDisplay for CliUpgradeableProgram {}
+impl VerboseDisplay for CliUpgradeableProgram {}
+impl fmt::Display for CliUpgradeableProgram {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        writeln!(f)?;
+        writeln_name_value(f, "Program Address:", &self.program_id)?;
+        writeln_name_value(f, "  Executable:", &self.program_executable.to_string())?;
```
CriesofCarrots, line 1551: Nitpicky, but I think everything from here down in can have one less indent.

cli-output/src/cli_output.rs
```diff
@@ -1528,6 +1528,100 @@ impl fmt::Display for CliProgramAuthority {
...
+    pub program_len: usize,
+    pub use_lamports_unit: bool,
+}
+impl QuietDisplay for CliUpgradeableBuffer {}
+impl VerboseDisplay for CliUpgradeableBuffer {}
+impl fmt::Display for CliUpgradeableBuffer {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        writeln!(f)?;
+        writeln_name_value(f, "Buffer Address:", &self.address)?;
+        writeln_name_value(f, "  Authority:", &self.authority)?;
```
CriesofCarrots, line 1605: Same nit about indents

cli/src/program.rs
```diff
@@ -937,6 +970,82 @@ fn process_get_authority(
...
+            {
+                Ok(config
+                    .output_format
+                    .formatted_string(&CliUpgradeableBuffer {
+                        address: account_pubkey.to_string(),
+                        lamports: account.lamports,
+                        authority: authority_address
+                            .map(|pubkey| pubkey.to_string())
+                            .unwrap_or_else(|| "none".to_string()),
+                        data_len: account.data.len(),
```
CriesofCarrots, line 1033: Same question about showing 2 lengths.

cli/src/program.rs
jackcmay: Yup, I'm replacing `get-info` with `show` and removing `get-authority`

CriesofCarrots: \o/ awesome!

cli/src/program.rs
jackcmay: My concern here is we print the lamports and someone might very rent-exempt but use the wrong length.  We could remove all the `Account` items from here

cli-output/src/cli_output.rs
jackcmay: Maybe no indents at all, we don't don't really have to reflect the internal structures

CriesofCarrots: Works for me!

cli/src/program.rs
CriesofCarrots: Do we not require the account be rent exempt on initialization?
But yeah, just removing the Account items would be fine.

PR #15018, opened 2021-02-02
Upgrade jsonrpc crates to v17.0.0

CriesofCarrots:
#### Problem
Our jsonrpc crates are out of date; the update requires some rewrite, as the tokio dependency moves from v0.1 to v0.2

#### Summary of Changes
Update crates and make necessary edits.

Toward #10187

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -891,6 +891,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "derive_more"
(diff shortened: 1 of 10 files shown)


mvines (approved):
Thanks for finishing this off for me!   Let's keep this on master only and ship in v1.6 since we want to keep v1.5 nice and stable now

PR #15022, opened 2021-02-02
Don't reset accounts if the remove_account comes from a clean

sakridge:
#### Problem

Two problems with resetting the accounts backing store from `remove_account`:

* When async hashing comes the store may be in-use with snapshot hash calculation and if the store is reset, then doing `store.accounts(0)` will return an empty vec.
* When unrefing the store, if there is a reset, then accounts will not be unref'ed correctly in `finalize_dead_slot_removal` and thus not cleared from the index.

#### Summary of Changes

Add a new flag `reset_accounts` and only set `true` in the store_accounts path which is operating on non-frozen slots. Shrink and clean operate on only rooted slots and will set `reset_accounts=false` to keep the store contents from being reset.

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -493,11 +493,15 @@ impl AccountStorageEntry {
         }
     }
 
-    fn remove_account(&self, num_bytes: usize) -> usize {
+    pub fn all_accounts(&self) -> Vec<StoredAccountMeta> {
+        self.accounts.accounts(0)
+    }
+
+    fn remove_account(&self, num_bytes: usize, reset_accounts: bool) -> usize {
         let mut count_and_status = self.count_and_status.write().unwrap();
         let (mut count, mut status) = *count_and_status;
 
-        if count == 1 && status == AccountStorageStatus::Full {
+        if count == 1 && status == AccountStorageStatus::Full && reset_accounts {
             // this case arises when we remove the last account from the
             //  storage, but we've learned from previous write attempts that
             //  the storage is full
@@ -1109,8 +1113,19 @@ impl AccountsDB {
         inc_new_counter_info!("clean-old-root-par-clean-ms", clean_rooted.as_ms() as usize);
 
         let mut measure = Measure::start("clean_old_root_reclaims");
+
+        // Don't reset from clean, since the pubkeys in those stores may need to be unref'ed
+        // and those stores may be used for background hashing.
+        let reset_accounts = false;
+
         let mut reclaim_result = (HashMap::new(), HashMap::new());
-        self.handle_reclaims(&reclaims, None, false, Some(&mut reclaim_result));
+        self.handle_reclaims(
+            &reclaims,
+            None,
+            false,
+            Some(&mut reclaim_result),
+            reset_accounts,
+        );
         measure.stop();
         debug!("{} {}", clean_rooted, measure);
         inc_new_counter_info!("clean-old-root-reclaim-ms", measure.as_ms() as usize);
@@ -1486,7 +1501,10 @@ impl AccountsDB {
 
         let reclaims = self.purge_keys_exact(&pubkey_to_slot_set);
 
-        self.handle_reclaims(&reclaims, None, false, None);
+        // Don't reset from clean, since the pubkeys in those stores may need to be unref'ed
+        // and those stores may be used for background hashing.
+        let reset_accounts = false;
+        self.handle_reclaims(&reclaims, None, false, None, reset_accounts);
 
         reclaims_time.stop();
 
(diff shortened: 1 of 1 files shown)


sakridge:
Migrate from #14880

jeffwashington:
👍

runtime/src/accounts_db.rs
```diff
@@ -493,11 +493,15 @@ impl AccountStorageEntry {
         }
     }
 
-    fn remove_account(&self, num_bytes: usize) -> usize {
+    pub fn accounts(&self, offset: usize) -> Vec<StoredAccountMeta> {
+        self.accounts.accounts(offset)
+    }
+
+    fn remove_account(&self, num_bytes: usize, reset_accounts: bool) -> usize {
```
ryoqun, line 500: nits: well, I always bothered about naming of `reset` in general. Somewhat confusingly, `accounts` here doesn't mean some kind of collections of `Accounts`, but rather specifically referring to the `.accounts` field, which in turn I think should be named just like `.append_vec`.

how about these renames along line of `rewind_append_vec`, `truncate_store`, `reset_storage`, `clear_accounts`?

runtime/src/accounts_db.rs
```diff
@@ -1721,6 +1749,7 @@ impl AccountsDB {
                 Some(Box::new(move |_, _| shrunken_store.clone())),
                 Some(Box::new(write_versions.into_iter())),
                 false,
+                false, /* reset_accounts */
```
ryoqun, line 1752: here.

runtime/src/accounts_db.rs
ryoqun: For example, maybe future me must wonder what `reset_accounts` mean while reading this code residing at the way-higher level in call-graph wise:

https://github.com/solana-labs/solana/pull/15022/files#r569300084

runtime/src/accounts_db.rs
```diff
@@ -493,11 +493,15 @@ impl AccountStorageEntry {
         }
     }
 
-    fn remove_account(&self, num_bytes: usize) -> usize {
+    pub fn accounts(&self, offset: usize) -> Vec<StoredAccountMeta> {
```
ryoqun, line 496: :+1: 

nits: how about omitting the `offset` and renaming to `all_accounts`? I always wonder what `0` means in the `...accounts(0)` when I read the code. Also, it seems that all of current caller are passing `0`. ;)

runtime/src/accounts_db.rs
```diff
@@ -2609,9 +2638,10 @@ impl AccountsDB {
             }
         }
 
-        // 1) Remove old bank hash from self.bank_hashes
```
ryoqun, line 2612: Thanks for policing stale comment. :)

runtime/src/accounts_db.rs
```diff
@@ -4015,13 +4047,22 @@ impl AccountsDB {
         hashes: &[Hash],
         is_cached_store: bool,
     ) {
+        // This path comes from a store to a non-frozen slot.
+        // If a store is dead here, then a newer update for
+        // each pubkey in the store must exist in another
+        // store in the slot. Thus it is safe to reset the store and
+        // re-use it for future stores. The pubkey ref counts should still
```
ryoqun, line 4054: nits: `re-use it for future store operations` (to disambuiate between store `SlotStore`/`AccountStorage` struct and the  `store_account_default()` operation.

runtime/src/accounts_db.rs
```diff
@@ -4015,13 +4047,22 @@ impl AccountsDB {
...
+        // hold just 1 ref from this slot.
```
ryoqun, line 4055: Nice comment and reasoning. :)

runtime/src/accounts_db.rs
```diff
@@ -4015,13 +4047,22 @@ impl AccountsDB {
         hashes: &[Hash],
         is_cached_store: bool,
     ) {
+        // This path comes from a store to a non-frozen slot.
```
carllin, line 4050: heh , maybe we should call this `store_accounts_non_frozen()` to make sure nobody calls this/or is wary of calling this from another path in the future.

carllin (approved):

runtime/src/accounts_db.rs
```diff
@@ -6769,6 +6816,135 @@ pub mod tests {
         assert_load_account(&accounts, current_slot, purged_pubkey2, 0);
     }
 
+    fn do_full_clean_refcount(store1_first: bool, store_size: u64) {
+        let pubkey1 = Pubkey::from_str("My11111111111111111111111111111111111111111").unwrap();
+        let pubkey2 = Pubkey::from_str("My22211111111111111111111111111111111111111").unwrap();
+        let pubkey3 = Pubkey::from_str("My33311111111111111111111111111111111111111").unwrap();
```
ryoqun, line 6822: yeah, this way saves our sanity from `pubkey::new_rand()`. xD

runtime/src/accounts_db.rs
```diff
@@ -3123,6 +3153,7 @@ impl AccountsDB {
                     Some(Box::new(move |_, _| flushed_store.clone())),
                     None,
                     false,
+                    false,
```
ryoqun, line 3156: nits: how about defining another thin wrapper for humans like `store_accounts_by_replacing`, which just forwards args by appending `false, false` at last:

```rust
fn store_accounts_by_replacing<'a>(
    &'a self,
    slot: Slot,
    accounts: &[(&Pubkey, &Account)],
    hashes: &[Hash],
    storage_finder: Option<StorageFinder<'a>>,
    write_version_producer: Option<Box<dyn Iterator<Item = u64>>>,
) -> StoreAccountsTiming {
    self.store_accounts_custom(..., false, false)
}
```

This is because:
- hard to grasp with increasing number of arguments.
- similar to this(https://github.com/solana-labs/solana/pull/15022/files#r569329203), let's make intended code-path-dependent assumptions explicit by naming things clearly.
- it seems that all of callers currently passing `false, false` indicating this is common arguments
- imo, `store_accounts**_by_replacing**` should be decent/helpful name to be called from the flushing and shrinking (v1/v2) paths. or `store_existing_accounts_again()`?
- lastly, I think `{false,true}, /* foo_bar */` is a good sign to refactor code itself.... ;)

ryoqun:
maybe backporting to v1.5?

runtime/src/accounts_db.rs
```diff
@@ -3688,7 +3720,7 @@ impl AccountsDB {
                     "AccountDB::accounts_index corrupted. Storage pointed to: {}, expected: {}, should only point to one slot",
                     store.slot(), *slot
                 );
-                let count = store.remove_account(account_info.stored_size);
+                let count = store.remove_account(account_info.stored_size, reset_accounts);
```
ryoqun, line 3723: @sakridge well, I pondered some time for the possibility where dead slot's stores are recycled too prematurely. If we could allow that, it's still possible to `.reset()` when using via background-hashing. In conclusion, it seems that this isn't possible, right? how about mentioning at one of new comments in some way?

Well, AccountsDB is quite complex. I wonder now that we have AccountsCaching, recycling code could be a worth thing to maintain...?

ryoqun: Hhh, I rather now think the bad recycle could happen in fact? Could you write a simple test for this?

sakridge: Yes, the AccountStorageEntry can be recycled while the hashing is in-progress, but it will not be taken out because the Arc ref will be >1 as from this check:

https://github.com/solana-labs/solana/blob/971c222cf7618559761bbdad63256a803171da85/runtime/src/accounts_db.rs#L2153

The AccountsPackage and the recycler will both hold a reference to the store.

ryoqun: oh, thanks for the explanation. makes sense.

ryoqun:
I wonder how much this cloud be leaking, relating to #14366 

If there is zero lamport updates to an account, this bug could prevent storages with older updates of the account from being removed correctly via wrong result at `calc_delete_dependencies()`?

Also, normally (not relating to zero lamport accounts), once slots got dead, its resources are removed previously? (except the slight skew of ref_count)?

sakridge:
> I wonder how much this cloud be leaking, relating to #14366
> 
> If there is zero lamport updates to an account, this bug could prevent storages with older updates of the account from being removed correctly via wrong result at `calc_delete_dependencies()`?
> 
> Also, normally (not relating to zero lamport accounts), once slots got dead, its resources are removed previously? (except the slight skew of ref_count)?

Yea I had the same idea along with Carl’s shrink fix to the ref counts as well.

ryoqun (approved):
lgtm with nits.

Nice find. :)

runtime/src/accounts_db.rs
sakridge: Yea. I think I'll do this in a follow-up.

runtime/src/accounts_db.rs
sakridge: Yes. Did this.

PR #15034, opened 2021-02-03
calculate hash from store instead of index

jeffwashington:
Problem
The number of accounts are growing.
There is value in moving operations out of the critical path. Some operations like clean and shrink are not able to complete effectively.
One such operation is calculating the hash of a bank.

Summary of Changes
Move the hash calculation out of accounts_background_service and into accounts_hash_verifier. The calculation also avoids using the index, reducing contention.
Add a command line option to calculate the hash using the index in the previous location and verify the hash calculation in ahv without using the index.

Fixes #

### DIFF
--- a/accounts-bench/src/main.rs
+++ b/accounts-bench/src/main.rs
@@ -1,3 +1,5 @@
+#[macro_use]
+extern crate log;
 use clap::{crate_description, crate_name, value_t, App, Arg};
 use rayon::prelude::*;
 use solana_measure::measure::Measure;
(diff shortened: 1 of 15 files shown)


jeffwashington:
original draft pr #14879

jeffwashington:
@sakridge this now contains the restoration of calculating hash in accounts bg svc. This is the first pr to go in.

jeffwashington:
Note that this had to be reworked at the last minute. Snapshot storages contain serialized banks and accountdb that had a default/missing hash value. We will be working to modify the persistence order since the goal is to know the hash later.

jeffwashington:
@ryoqun , @carllin - fyi. Apologies, I should've had you review it first.

PR #15050, opened 2021-02-03
removes pubkey references

behzadnouri:
#### Problem
The pubkeys are retained until the next epoch:
https://github.com/solana-labs/solana/blob/8376781ec/core/src/replay_stage.rs#L1867-L1869
plus the overhead of `Rc` itself.
So not clear these are saving any memory over the plain pubkeys.
And they are not necessary for the code logic either,
but they have a negative impact on the runtime.

#### Summary of Changes
Removed `PubkeyReferences`

### DIFF
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -1,7 +1,4 @@
-use crate::{
-    progress_map::{LockoutIntervals, ProgressMap},
-    pubkey_references::PubkeyReferences,
-};
+use crate::progress_map::{LockoutIntervals, ProgressMap};
 use chrono::prelude::*;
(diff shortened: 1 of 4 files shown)


behzadnouri:
Left chunk is after removing the pubkey references:

`replay-loop-timing-stats::compute_bank_stats_elapsed`:
rm_pubkey_refs replay-loop-timing-stats compute_bank_stats_elapsed

carllin (approved):

carllin:
Thanks for the cleanup! Can also probably be removed from `cluster_info_vote_listener` and `cluster_slots.rs`

behzadnouri:
> Thanks for the cleanup! Can also probably be removed from `cluster_info_vote_listener` and `cluster_slots.rs`

yes, I will send you another pull request for that

PR #15058, opened 2021-02-03
blockstore: Revert hard nofile limit back to 500000

mvines:
894b412aef bumped the hard limit up to 700000.  This will cause every v1.5 validator that had previously followed our guidance to set nofile to 500000 to fail to start up.   Too harsh!


t-nelson (approved):
Whoops!  I meant to roll that one back.  Thanks for catching

PR #15075, opened 2021-02-04
Add program deployment docs

jackcmay:
#### Summary of Changes

Fixes #


docs/src/cli/deploy-a-program.md
```diff
@@ -0,0 +1,217 @@
...
+and then later grows substantially, the redeployment may fail.  To avoid this,
+specify a `max_len` that is at least the size (in bytes) that the program is
+expected to become (plus some wiggle room).
+
+```bash
+solana program deploy --max-len 200000 <PROGRAM_FILEPATH>
+```
+
+Note that program accounts are required to be
+[rent-excempt](developing/programming-model/accounts.md#rent-exemption), and the
```
CriesofCarrots, line 110: rent-excempt > rent-exempt

docs/src/cli/deploy-a-program.md
```diff
@@ -0,0 +1,217 @@
...
+./path-to-program/program.so
+./path-to-program/program-keypair.json
+```
+
+### Showing a program account
+
+To get information about a deployed program:
+
+```bash
+solana show <ACCOUNT_ADDRESS>
```
CriesofCarrots, line 64: solana program show

docs/src/cli/deploy-a-program.md
```diff
@@ -0,0 +1,217 @@
...
+```bash
+solana program deploy --program-id <KEYPAIR_FILEPATH> <PROGRAM_FILEPATH>
+```
+
+If the program id is not specified on the command line the tools will first look
+for a keypair file matching the `<PROGRAM_FILEPATH>`, or internally generate a
+new keypair.
+
+A matching program/keypair file is generated automatically by the program build
+tools and looks like:
```
CriesofCarrots, line 52: Suggest:

A matching program keypair file is in the same directory as the program's shared object, and named `<PROGRAM_NAME>-keypair.json`. Matching program keypairs are generated automatically by the program build tools:

docs/src/cli/deploy-a-program.md
```diff
@@ -0,0 +1,217 @@
...
+redeployments will be to the same program address.
+
+The command looks the same as the deployment command:
+
+```bash
+solana program deploy <PROGRAM_FILEPATH>
+```
+
+By default, programs are deployed to accounts that are twice the size of the
+original deployment.  Doing so leaves room for further redeployments.  But, if
```
CriesofCarrots, line 99: Doing so leaves room for further redeployments >
Doing so leaves room for program growth in future redeployments

(Otherwise, it sort of sounds like future redeployments are written after the original in the same account.)

docs/src/cli/deploy-a-program.md
```diff
@@ -0,0 +1,217 @@
...
+
+Or after deployment and specifying the current authority:
+
+```bash
+solana program set-upgrade-authority <PROGRAM_ADDRESS> --upgrade-authority <UPGRADE_AUTHORITY_SIGNER> --new-upgrade-authority <NEW_UPGRADE_AUTHORITY>
+```
+
+### Immutable programs
+
+A program can be marked immutable, which prevents all further redeployments by
```
CriesofCarrots, line 141: A program can be marked immutable, which prevents all further redeployments, by

docs/src/cli/deploy-a-program.md
```diff
@@ -0,0 +1,217 @@
...
+$ cp original.so extended.so
+$ truncate -r dump.so extended.so
+$ sha256sum extended.so dump.so
+```
+
+### Using an intermediary Buffer account
+
+Instead of deploying directly to the program account, the program can be written
+to an intermediary buffer account.  Doing so can be useful for things like
+multi-entity governed programs.
```
CriesofCarrots, line 181: I still don't really understand when and why I would use a buffer account. I assume it's something about separating the write to the chain (and possible updates) from the actual activation. Perhaps just half a sentence to expand on how deployment of a multi-entity governed program works?

docs/src/cli/deploy-a-program.md
```diff
@@ -0,0 +1,217 @@
...
+
+```bash
+solana program deploy --program-id <PROGRAM_ADDRESS> --buffer <BUFFER_ADDRESS>
+```
+
+Note, the buffer's authority must match the program's upgrade authority.
+
+Buffers also support `show` and `dump` just like programs do.
+
+### Non-redeployable programs
```
CriesofCarrots, line 207: Non-upgradeable ?

docs/sidebars.js
```diff
@@ -53,7 +53,8 @@ module.exports = {
 	    "cli/choose-a-cluster",
 	    "cli/transfer-tokens",
 	    "cli/delegate-stake",
-	    "cli/manage-stake-accounts",
+            "cli/manage-stake-accounts",
+            "cli/deploy-a-program",
```
CriesofCarrots, line 57: Some extra indents here

docs/src/cli/deploy-a-program.md
jackcmay: Haven't used "upgrade" or "upgradeable" in the entire doc, didn't want to introduce that terminology here.

docs/src/cli/deploy-a-program.md
```diff
@@ -0,0 +1,217 @@
...
+solana show <ACCOUNT_ADDRESS>
+```
+
+An example output looks like:
+
+```bash
+Program Id: 3KS2k14CmtnuVv2fvYcvdrNgC94Y11WETBpMUGgXyWZL
+ProgramData Address: EHsACWBhgmw8iq5dmUZzTA1esRqcTognhKNHUkPi4q4g
+Authority: FwoGJNUaJN2zfVEex9BB11Dqb3NJKy3e9oY3KTh9XzCU
+Last Upgraded In Slot: 63890568
```
CriesofCarrots, line 73: Last Deployed In Slot

docs/src/cli/deploy-a-program.md
```diff
@@ -0,0 +1,217 @@
...
+### Non-redeployable programs
+
+Using `solana program ...` utilizes Solana's upgradeable loader, but there is
+another way to deploy programs using the original on-chain loader.
+
+```bash
+solana deploy <PROGRAM_FILEPATH>
+```
+
+Programs deployed with `solana deploy ...` are not upgradeable and are not
```
CriesofCarrots, line 216: are not upgradeable > cannot be redeployed to the same address and are not
?

docs/src/cli/deploy-a-program.md
CriesofCarrots: Hmm, you do use the word upgradeable twice in this section below. I see your point, though (and added a suggestion for one of those uses).
`Non-redeployable` is a bear, though, and not entirely accurate, since any program can be redeployed to a different address.

docs/sidebars.js
jackcmay: Dan tabs!

mvines: Lol I feel ya, but @danpaul000's last name is not "Tabs".  Just FYI

docs/src/cli/deploy-a-program.md
jackcmay: Would it make more sense if I went back and used "upgradeable" instead throughout?

jackcmay: I would argue the a different address is considered a deploy and a redeploy is only to the same address
deploy to X
deploy to Y
redeploy to X

CriesofCarrots: > I would argue the a different address is considered a deploy and a redeploy is only to the same address

That makes perfect sense from a code perspective, but I don't think that distinction is necessarily clear for a new solana user reading the word "redeploy".

> Would it make more sense if I went back and used "upgradeable" instead throughout?

That's an option, but I didn't find "redeploy" ambiguous or troublesome elsewhere. Really, if we can come up with something instead of `non-redeployable`, I'll be placated ;)

PR #15077, opened 2021-02-04
update hash - performance improvements - testing still

jeffwashington:
#### Problem
Calculating a hash is expensive.

#### Summary of Changes
Improve the performance of calculating hashes.

Fixes #


jeffwashington:
colo validator machine:
normal update_accounts_hash call in abs: ~6.3s
non-index update_accounts_hash call in abs: ~9.04s
non-index update_accounts_hash_v2 (this pr) call in abs: ~2.38s

PR #15080, opened 2021-02-04
Generate keypair file for c program deployment

jackcmay:
#### Problem

No default keypair generated for c program deployment

#### Summary of Changes

generate one

Fixes #


jackcmay:
@mvines What do you think of calling `solana-keygen` in the C makefile?

mvines:
> What do you think of calling `solana-keygen` in the C makefile? 

I like the idea

> Are the solana tools (solana-keygen specifically) not available in the stable-perf ci build?

Not in the PATH, I think we'd need to add a rule to build solana-keygen from source before using it in stable-perf

PR #15085, opened 2021-02-04
fix merge conflict

jeffwashington:
#### Problem
same function was added overlapping resulting in duplicates

#### Summary of Changes
remove 2nd added function

Fixes #


jeffwashington:
submitting because compile is broken

PR #15088, opened 2021-02-04
borrow storages

jeffwashington:
#### Problem
Clippy suggested this as part of larger changes and it is noisy, so I thought I'd pull it into its own pr.

#### Summary of Changes
type to &type

Fixes #


sakridge (approved):
lgtm

PR #15097, opened 2021-02-04
sentinel value for zero lamport accounts in hash scanning

jeffwashington:
#### Problem
it costs memory to keep track of original lamports and balance lamports during account scanning

#### Summary of Changes
use a sentinel value for balance lamports that indicates the raw lamports were zero (and the account should not be hashed)

Fixes #


carllin (approved):

sakridge (approved):
lgtm

PR #15104, opened 2021-02-04
change store account scan to not use dashmap

jeffwashington:
#### Problem
scanning accounts in stores for calculating hashes is slow.

#### Summary of Changes
collect all accounts (even dup entries), then sort, and keep only the final version of the account.
then, continue hashing as before.

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -3752,8 +3752,8 @@ impl AccountsDB {
                     end_index = len;
                 }
 
-                let first_slice = chunk_index == 0;
-                if !first_slice {
+                let is_first_slice = chunk_index == 0;
+                if !is_first_slice {
                     start_index -= 1;
                 }
 
@@ -3773,7 +3773,7 @@ impl AccountsDB {
     }
 
     fn de_dup_accounts_from_stores(
-        first_slice: bool,
+        is_first_slice: bool,
         slice: &[CalculateHashIntermediate],
     ) -> (Vec<Hash>, u128) {
         let len = slice.len();
@@ -3782,7 +3782,7 @@ impl AccountsDB {
         let mut sum: u128 = 0;
         if len > 0 {
             let mut i = 0;
-            let mut look_for_first_key = !first_slice;
+            let mut look_for_first_key = !is_first_slice;
             'outer: loop {
                 // at start of loop, item at 'i' is the first entry for a given pubkey - unless look_for_first
                 let now = &slice[i];
@@ -5268,7 +5268,7 @@ pub mod tests {
         type ExpectedType = (String, bool, u64, String);
         let expected:Vec<ExpectedType> = vec![
             // ("key/lamports key2/lamports ...",
-            // first_slice
+            // is_first_slice
             // result lamports
             // result hashes)
             // "a5" = key_a, 5 lamports
@@ -5316,11 +5316,11 @@ pub mod tests {
         for first_slice in 0..2 {
             for start in 0..COUNT {
                 for end in start + 1..COUNT {
-                    let first_slice = first_slice == 1;
+                    let is_first_slice = first_slice == 1;
                     let accounts = accounts.clone();
                     let slice = &accounts[start..end];
 
-                    let result = AccountsDB::de_dup_accounts_from_stores(first_slice, slice);
+                    let result = AccountsDB::de_dup_accounts_from_stores(is_first_slice, slice);
                     let (hashes2, lamports2) = AccountsDB::de_dup_accounts_in_parallel(slice, 1);
                     let (hashes3, lamports3) = AccountsDB::de_dup_accounts_in_parallel(slice, 2);
                     let (hashes4, _, lamports4) =
@@ -5359,12 +5359,12 @@ pub mod tests {
 
                     let packaged_result: ExpectedType = (
                         human_readable,
-                        first_slice,
+                        is_first_slice,
                         result.1 as u64,
                         hash_result_as_string,
                     );
 
-                    if first_slice {
+                    if is_first_slice {
                         // the parallel version always starts with 'first slice'
                         assert_eq!(
                             result.0, hashes,
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -5414,6 +5414,12 @@ pub mod tests {
         let (list, _) = AccountsDB::sort_hash_intermediate(list);
         assert_eq!(list, list_bkup);
 
+        let list = vec![val2.clone(), val.clone()]; // reverse args
+        let mut list_bkup = list.clone();
+        list_bkup.sort_by(AccountsDB::compare_two_hash_entries);
+        let (list, _) = AccountsDB::sort_hash_intermediate(list);
+        assert_eq!(list, list_bkup);
+
         // slot same, vers =
         let hash3 = Hash::new_unique();
         let val3 = CalculateHashIntermediate::new(1, hash3, 2, 1, key);
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -175,15 +175,17 @@ struct CalculateHashIntermediate {
     pub hash: Hash,
     pub lamports: u64,
     pub slot: Slot,
+    pub pubkey: Pubkey,
 }
 
 impl CalculateHashIntermediate {
-    pub fn new(version: u64, hash: Hash, lamports: u64, slot: Slot) -> Self {
+    pub fn new(version: u64, hash: Hash, lamports: u64, slot: Slot, pubkey: Pubkey) -> Self {
         Self {
             version,
             hash,
             lamports,
             slot,
+            pubkey,
         }
     }
 }
@@ -195,12 +197,6 @@ trait Versioned {
 impl Versioned for (u64, Hash) {
     fn version(&self) -> u64 {
         self.0
-    }
-}
-
-impl Versioned for CalculateHashIntermediate {
-    fn version(&self) -> u64 {
-        self.version
     }
 }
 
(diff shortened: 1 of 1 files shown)


jeffwashington:
measured avg of 2.561702439s vs mainnet-beta on colo machine.

jeffwashington:
note this is a reduced and simplified pr from a more complicated algorithm that bins by pubkey values and does more efficient copying, etc. The plan was to get this in and then build on it.

jeffwashington:
Also note that there are 2 other simple prs cherry-picked at the beginning of this that make the diff more complicated than it needs to be.

jeffwashington:
a perf log line:
with change:
calculate_accounts_hash_without_index accounts_scan=440286i eliminate_zeros=119986i hash=27602i sort=1190616i hash_total=15097762i flatten=454781i flatten_after_zeros=79885i unreduced_entries=15448300i

runtime/src/accounts_db.rs
```diff
@@ -3660,79 +3656,205 @@ impl AccountsDB {
...
+        let (data_sections_by_pubkey, time_scan) = accounts;
 
-        let mut zeros = Measure::start("eliminate zeros");
-        let hashes = Self::remove_zero_balance_accounts(account_maps);
-        zeros.stop();
-        let hash_total = hashes.len();
-        let (ret, (sort_time, hash_time)) =
-            Self::accumulate_account_hashes_and_capitalization(hashes, Slot::default(), false);
+        let (outer, flatten_time, raw_len) =
+            Self::flatten_hash_intermediate(data_sections_by_pubkey);
```
carllin, line 3830: Instead of flattening here, I'm curious to see how flattening within Rayon would perform, I've made a suggested change here: https://github.com/carllin/solana/commit/670103db2c8fb9b94440873d6b9404c08a326f0d#diff-1090394420d51617f3233275c2b65ed706b35b53b115fe65f82c682af8134a6fR3666

runtime/src/accounts_db.rs
```diff
@@ -3660,79 +3656,205 @@ impl AccountsDB {
...
-        snapshot_storages
-            .into_par_iter()
-            .flatten()
-            .map(|storage| {
-                let accounts = storage.accounts.accounts(0);
+        let items: Vec<_> = snapshot_storages.iter().flatten().collect();
+
+        // Without chunks, we end up with 1 output vec for each outer snapshot storage.
+        // This results in too many vectors to be efficient.
+        const MAX_ITEMS_PER_CHUNK: usize = 5_000;
```
carllin, line 3663: Should this be calculated based on `snapshot_storages.len()` and `num_cpus::get()`?

runtime/src/accounts_db.rs
```diff
@@ -3660,79 +3656,205 @@ impl AccountsDB {
...
-            .flatten()
             .collect();
-        hashes
+
+        let sum = *overall_sum.lock().unwrap();
+        (hashes, sum)
+    }
+
+    fn de_dup_accounts_from_stores(
+        first_slice: bool,
```
carllin, line 3776: nit: signal boolean by `first_slice`->`is_first_slice`

runtime/src/accounts_db.rs
```diff
@@ -3660,79 +3656,205 @@ impl AccountsDB {
...
+                }
+
+                let (result, sum) = Self::de_dup_accounts_from_stores(
+                    chunk_index == 0,
+                    &pubkey_division[start_index..end_index],
+                );
+                let mut overall = overall_sum.lock().unwrap();
+                *overall = Self::checked_cast_for_capitalization(sum + *overall as u128);
+
+                result
```
carllin, line 3767: Hmm this won't guarantee the most equitable distribution of work if the divisibility is an issue right? For instance, 19 work items, split among 10 workers here, would yield 9 works with 1 item, and 1 worker with 10 items here.  

How about something like this: https://github.com/carllin/solana/commit/9a8eb0482c2b951d34714193e17c7b8fed76eec8

runtime/src/accounts_db.rs
```diff
@@ -3660,79 +3656,205 @@ impl AccountsDB {
...
+        sorted_data_by_pubkey: Vec<CalculateHashIntermediate>,
+    ) -> (Vec<Vec<Hash>>, Measure, u64) {
+        // 1. eliminate zero lamport accounts
+        // 2. pick the highest slot or (slot = and highest version) of each pubkey
+        // 3. produce this output:
+        // vec: PUBKEY_BINS_FOR_CALCULATING_HASHES in pubkey order
+        //   vec: sorted sections from parallelism, in pubkey order
+        //     vec: individual hashes in pubkey order
+        let mut zeros = Measure::start("eliminate zeros");
+        const CHUNKS: usize = 10;
```
carllin, line 3727: Should this be based on number of cpus?

carllin:
@jeffwashington cool stuff, thanks for all the benchmarking!

One thing I noticed, probably should've caught this in the original PR, is that all the new Rayon parallel iterators for the new storage scans are no longer using the clean thread pool, which `calculate_accounts_hash()` uses: https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_db.rs#L3545.

In the past we've noticed that without an explicit thread pool which caps the number of thread resources, Rayon tends to consume all CPU in intensive scan operations, so I think we should probably introduce that again 😄

runtime/src/accounts_db.rs
jeffwashington: I will gladly try this out. I was trying to keep this pr straightforward, so all I did was separate out the flattens into their own area so they can be timed and dealt with in later PRs.

runtime/src/accounts_db.rs
jeffwashington: I'm not sure on the heuristics of getting a number here. I know no number is not good. I just picked something and it had reasonable performance so I haven't gone back to tune this yet. I assumed as I/we learned something more about the dynamics of the system, more intuitive understanding may appear. I have no idea how cpu or io bound these functions are.
I don't like it being a magic number, so at least its a constant, but I'm not sure what the right answer is yet.
I don't mean to sound defensive in these responses, fwiw.

runtime/src/accounts_db.rs
jeffwashington: This is also another constant number with not enough intuitive understanding yet. Empirically, 10 works better than 1. Once it is A number, my thinking is the algorithm can be tuned to 15M accounts, 80M, 100M, 1B, etc. My process is: Measure. See something expensive. Think about how to make it faster. Try something. Repeat. I agree that the tuning of any of this is not well done yet. I was making some assumptions like: machines have lots of cpus. There is lots of data. cpus and data likely will grow in the future. So, find places where we can split up work and see what effect that has and what side effects it introduces. And, then, the levers are there ready to pull as we hit the next hotspot. I'm open to feedback. 

But, also, this is the simpler changes I pulled out of the changes that were twice as fast as this. My thinking in this PR is these ideas seem sufficiently sound to build off and they have enough performance bang to be a good tradeoff for the moment while we keep getting better as time marches forward.

Again, I don't mean to sound defensive. Text is hard. I am trying to discuss/explore the tradeoffs and get us on the same page.

runtime/src/accounts_db.rs

sakridge:
> @jeffwashington cool stuff, thanks for all the benchmarking!
> 
> One thing I noticed, probably should've caught this in the original PR, is that all the new Rayon parallel iterators for the new storage scans are no longer using the clean thread pool, which `calculate_accounts_hash()` uses: https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_db.rs#L3545.
> 
> In the past we've noticed that without an explicit thread pool which caps the number of thread resources, Rayon tends to consume all CPU in intensive scan operations, so I think we should probably introduce that again 😄
> 
> @sakridge wonder if that might contribute to those stalls we're seeing on the Kin cluster, @CriesofCarrots probably remembers quite well when we were debugging that the replay stalls were due to Rayon eating all the resources 🍰

Yes, good catch.

runtime/src/accounts_db.rs
jeffwashington: My thinking was: some divisibility is better than none. Completely equitable is great. Always correct is necessary. You are right about len = 19. It isn't good. My assumption was data sizes that are already large (15M). But the last worker's share arrives at an equitable amount quickly enough. len=10k gives you 10.08% for the last worker. For 1M, it is 10.0004%. I can certainly tweak the algorithm for distribution to use float division or other techniques. I am not confident this algorithm will remain implemented as it is for even 6 months. Already there is the 80M and 1B talk ;-) I imagine this to be a sufficiently simple, and empirically faster implementation as we learn more and find the next hotspots and come up with even better implementations.

jeffwashington: I propose that we treat the optimization of the size per worker selection as a follow on discussion to this PR. Is that reasonable?

jeffwashington:
> > @jeffwashington cool stuff, thanks for all the benchmarking!
> > One thing I noticed, probably should've caught this in the original PR, is that all the new Rayon parallel iterators for the new storage scans are no longer using the clean thread pool, which `calculate_accounts_hash()` uses: https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_db.rs#L3545.
> > In the past we've noticed that without an explicit thread pool which caps the number of thread resources, Rayon tends to consume all CPU in intensive scan operations, so I think we should probably introduce that again 😄
> > @sakridge wonder if that might contribute to those stalls we're seeing on the Kin cluster, @CriesofCarrots probably remembers quite well when we were debugging that the replay stalls were due to Rayon eating all the resources 🍰
> 
> Yes, good catch.

[added thread pool](https://github.com/solana-labs/solana/pull/15104/commits/4b66360a46b21251072e64cd45dff60c0f86f57c). Hopefully I did it correctly.

runtime/src/accounts_db.rs
jeffwashington: another thing to note is that the next PR I am working on is binning the accounts by pubkey value. The result of this is another layer of vectors, which makes the flattening problem worse. For this PR, I was trying to keep standard, easy to understand flattening because the next change was likely to make the complexity higher but the performance better. And I was trying not to over optimize a way point along the route.

I am loving the engagement! It is such a relief that the basic index->non-index code is finally in so we can collaborate on making other parts much faster and better now!

jeffwashington: I cherry-picked @carllin 's code. Unfortunately, the results are much worse.
accounts_scan 1,082,644us
current pr measurement: 413,866us
I can investigate this more later.

core/src/accounts_hash_verifier.rs
```diff
@@ -38,6 +40,7 @@ impl AccountsHashVerifier {
         fault_injection_rate_slots: u64,
         snapshot_interval_slots: u64,
     ) -> Self {
+        let thread_pool = accounts_db::make_min_priority_thread_pool();
```
sakridge, line 43: Can we just create it with the flag?

sakridge:
```
let thread_pool = if package.test_hash {
    Some(accounts_db::make_min_priority_thread_pool())
} else {
    None
};
```

runtime/src/accounts_db.rs
carllin: no worries, thanks for trying it out!

runtime/src/accounts_db.rs
carllin: It's all good, the drift is at most `chunk_count` anyways, which is currently small. Was more for future-proofing in case I wanted such a utility in the future 😃

runtime/src/accounts_db.rs
```diff
@@ -3660,79 +3656,205 @@ impl AccountsDB {
...
+        is_first_slice: bool,
+        slice: &[CalculateHashIntermediate],
+    ) -> (Vec<Hash>, u128) {
+        let len = slice.len();
+        let mut result: Vec<Hash> = Vec::with_capacity(len);
+
+        let mut sum: u128 = 0;
+        if len > 0 {
+            let mut i = 0;
+            let mut look_for_first_key = !is_first_slice;
```
carllin, line 3785: Nit: How about an additional clarifying comment here that on slices other than the first one, this check is necessary because the last pubkey of the previous batch might be duplicated in the first `N` consecutive entries in this batch?

core/src/accounts_hash_verifier.rs
jeffwashington: this has been pulled into a separate pr. #15149

runtime/src/accounts_db.rs
```diff
@@ -3660,79 +3656,210 @@ impl AccountsDB {
...
+        chunk_count: usize,
+    ) -> (Vec<Vec<Hash>>, u64) {
+        let len = pubkey_division.len();
+        let max = if len > chunk_count { chunk_count } else { 1 };
+        let chunk_size = len / max;
+        let overall_sum = Mutex::new(0u64);
+        let hashes: Vec<Vec<Hash>> = (0..max)
+            .into_par_iter()
+            .map(|chunk_index| {
+                let mut start_index = chunk_index * chunk_size;
```
sakridge, line 3749: What is the difference between this and `par_chunks`?

runtime/src/accounts_db.rs
```diff
@@ -3660,79 +3656,210 @@ impl AccountsDB {
...
+            'outer: loop {
+                // at start of loop, item at 'i' is the first entry for a given pubkey - unless look_for_first
+                let now = &slice[i];
+                let last = now.pubkey;
+                if !look_for_first_key && now.lamports != ZERO_RAW_LAMPORTS_SENTINEL {
+                    // first entry for this key that starts in our slice
+                    result.push(now.hash);
+                    sum += now.lamports as u128;
+                }
+                for (k, now) in slice.iter().enumerate().take(len).skip(i + 1) {
```
sakridge, line 3800: `slice` is already sized to `len`. What does the `.take(len)` add?

jeffwashington: I think you're right. Looks like .take(len) is unnecessary.

runtime/src/accounts_db.rs
```diff
@@ -3660,79 +3656,210 @@ impl AccountsDB {
...
+        let result: Vec<_> = data_sections_by_pubkey.into_iter().flatten().collect();
+        let raw_len = result.len();
+        flatten_time.stop();
+        (result, flatten_time, raw_len)
+    }
+
+    fn compare_two_hash_entries(
+        a: &CalculateHashIntermediate,
+        b: &CalculateHashIntermediate,
+    ) -> std::cmp::Ordering {
```
sakridge, line 3697: Can we implement this as `PartialOrd` for `CalculateHashIntermediate` https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html

runtime/src/accounts_db.rs
jeffwashington: par_chunks regions don't overlap. I need all regions but the first region to also contain the last item in the previous region. This is due to the case of there being multiple entries for a single key that cross over a chunk boundary. The second chunk caller needs to ignore the continuation of a pubkey that began in the previous region.

jeffwashington: I added a comment to clarify this.

runtime/src/accounts_db.rs
jeffwashington: done

runtime/src/accounts_db.rs
jeffwashington: removed

jeffwashington:
No merge conflicts with master.

sakridge (approved):
lgtm

PR #15106, opened 2021-02-04
Add bank drop test and fix

sakridge:
#### Problem

Dropping a non-root bank may not clean up the keys it touches

#### Summary of Changes

Write a test and add a fix in clean_accounts to clean up the keys.

Fixes #

### DIFF
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -12224,4 +12224,41 @@ pub(crate) mod tests {
             genesis_config.epoch_schedule.get_first_slot_in_epoch(1),
         );
     }
+
+    #[test]
+    fn test_drop_bank_clears_storage() {
+        use std::str::FromStr;
+        solana_logger::setup();
+        let (genesis_config, mint_keypair) = create_genesis_config(100);
+        let bank0 = Arc::new(Bank::new(&genesis_config));
+
+        let pubkey1 = Pubkey::from_str("My11111111111111111111111111111111111111111").unwrap();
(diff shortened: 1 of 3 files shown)


sakridge:
@ryoqun do you want to take this over?

ryoqun:
> @ryoqun do you want to take this over?

@sakridge Yeah, I'd much like to take this on. But my backlog is rather long...

brooksprumo:
I'm looking into this one now.

runtime/src/bank.rs
```diff
@@ -12224,4 +12224,41 @@ pub(crate) mod tests {
...
+
+        bank0.transfer(2, &mint_keypair, &pubkey1).unwrap();
+        bank0.freeze();
+
+        let bank1 = Bank::new_from_parent(&bank0, &Pubkey::default(), 1);
+        bank1.transfer(3, &mint_keypair, &pubkey2).unwrap();
+        // todo: remove this and see that test passes
+        bank1.freeze();
+
+        let bank2 = Bank::new_from_parent(&bank0, &Pubkey::default(), 2);
```
brooksprumo, line 12246: @sakridge I'm still reading through all the code, but wanted to double check with you; can multiple banks have the same parent? I assume yes, since `bank1` and `bank2` are both children of `bank0`.

Is this what happens when there are forks too (multiple child banks off a single parent bank)?

brooksprumo:
I'm closing this PR. It is tracked in issue #16580, and was resolved with PR #16911.

PR #15136, opened 2021-02-05
program-test: Add warp tests for rent and stake rewards

joncinque:
#### Problem

Warping the bank forward in program-test worked for updating the sysvar clock, which is helpful for some testing usecases.  For stake pools, they will need to simulate stake rewards being distributed to various stake accounts.

#### Summary of Changes

Add the ability to simulate votes for staking rewards and a test.  I tried using the vote program directly, but without a history of hashes, I was unable to get any votes to succeed.

This also adds a rent collection test.

Fixes #

### DIFF
--- a/program-test/src/lib.rs
+++ b/program-test/src/lib.rs
@@ -817,6 +817,10 @@ impl ProgramTestContext {
         }
     }
 
+    pub fn genesis_config(&self) -> &GenesisConfig {
+        &self.genesis_config
+    }
+
     /// Manually increment vote credits for the current epoch in the specified vote account to simulate validator voting activity
     pub fn increment_vote_account_credits(
         &mut self,
--- a/program-test/src/lib.rs
+++ b/program-test/src/lib.rs
@@ -606,16 +606,17 @@ impl ProgramTest {
             });
         }
 
+        let rent = Rent::default();
         let bootstrap_validator_pubkey = Pubkey::new_unique();
-        let bootstrap_validator_stake_lamports = 10_000;
+        let bootstrap_validator_lamports = rent.minimum_balance(VoteState::size_of());
 
         let mut gci = create_genesis_config_with_leader(
             sol_to_lamports(1_000_000.0),
             &bootstrap_validator_pubkey,
-            bootstrap_validator_stake_lamports,
+            bootstrap_validator_lamports,
         );
         let genesis_config = &mut gci.genesis_config;
-        genesis_config.rent = Rent::default();
+        genesis_config.rent = rent;
         genesis_config.fee_rate_governor =
             solana_program::fee_calculator::FeeRateGovernor::default();
         debug!("Payer address: {}", gci.mint_keypair.pubkey());
@@ -765,7 +766,6 @@ pub struct ProgramTestContext {
     pub banks_client: BanksClient,
     pub last_blockhash: Hash,
     pub payer: Keypair,
-    pub voter: Keypair,
     pub genesis_config: GenesisConfig,
     bank_forks: Arc<RwLock<BankForks>>,
     block_commitment_cache: Arc<RwLock<BlockCommitmentCache>>,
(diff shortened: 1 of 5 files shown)


program-test/src/lib.rs
```diff
@@ -813,14 +808,35 @@ impl ProgramTestContext {
...
+    pub fn simulate_vote_activity(&mut self, voter: &Pubkey, number_of_slots: u64) {
+        let bank_forks = self.bank_forks.read().unwrap();
+        let bank = bank_forks.working_bank();
+        let working_slot = bank.slot();
+
+        // generate some vote activity for rewards
+        let mut voter_account = bank.get_account(voter).unwrap();
+        let mut vote_state = VoteState::from(&voter_account).unwrap();
+
+        // This can slow down testing for huge warp amounts, so limit to 2 dev epochs
```
mvines, line 830: "so limit to 2 dev epochs" -- where's this limit?  I don't see it in the code below

program-test/src/lib.rs
```diff
@@ -813,14 +808,35 @@ impl ProgramTestContext {
...
+            payer: genesis_config_info.mint_keypair,
+            genesis_config: genesis_config_info.genesis_config,
             bank_forks,
+            block_commitment_cache,
             _bank_task: bank_task,
         }
     }
 
+    /// Fake vote activity to accrue vote credits, and eventually stake rewards
+    pub fn simulate_vote_activity(&mut self, voter: &Pubkey, number_of_slots: u64) {
```
mvines, line 821:
```suggestion
    /// Manually increment vote credits for the current epoch in the specified vote account to simulate validator voting activity
    pub fn increment_vote_account_credits(&mut self, vote_account_address: &Pubkey, number_of_credits: u64) {
```

program-test/src/lib.rs
```diff
@@ -813,14 +808,35 @@ impl ProgramTestContext {
...
+        let working_slot = bank.slot();
+
+        // generate some vote activity for rewards
+        let mut voter_account = bank.get_account(voter).unwrap();
+        let mut vote_state = VoteState::from(&voter_account).unwrap();
+
+        // This can slow down testing for huge warp amounts, so limit to 2 dev epochs
+        for i in working_slot..=(working_slot + number_of_slots) {
+            let (epoch, _) = bank.get_epoch_and_slot_index(i);
+            vote_state.increment_credits(epoch);
```
mvines, line 833: Any reason to not simplify this to just apply credits for the current epoch?   Or alternatively just make the caller provide an `(epoch, credits)` pair as input?

joncinque: Nope, just my own lack of knowledge of how these things work!  For long periods, I thought it would be better to increment for each slot that passed, but someone can just pass `(epoch, slots_in_epoch)` if we use your suggestion.  `(epoch, credits)` makes more sense

program-test/src/lib.rs
joncinque: Oops, that was from some testing, I can get rid of the comment

program-test/src/lib.rs
joncinque: done in `20f5dfd`

program-test/src/lib.rs
joncinque: done in `20f5dfd`

program-test/tests/warp.rs
```diff
@@ -95,3 +106,145 @@ async fn custom_warp() {
...
+        .banks_client
+        .get_account(keypair.pubkey())
+        .await
+        .expect("account exists")
+        .unwrap();
+    assert_eq!(account.lamports, account_lamports);
+
+    // Warp forward and see that rent has been collected
+    // This test was a bit flaky with one warp, but two warps always works
+    context.warp_to_slot(DEFAULT_DEV_SLOTS_PER_EPOCH).unwrap();
```
mvines, line 148: Just a nit that resolving in a follow-up sometime seems fine: relying on `DEFAULT_DEV_SLOTS_PER_EPOCH` here is yucky.   Instead how about adding `fn genesis_config(&self) -> &GenesisConfig` method to `ProgramTestContext` so the test can do this: `context.genesis_config().epoch_schedule.slots_per_epoch`

mvines (approved):

PR #15139, opened 2021-02-05
Introduce minimalistic ttl eviction for Recycler/PinnedVec

ryoqun:
#### Problem

PinnedVec tends to leak indefinitely due to some past burst.

https://github.com/solana-labs/solana/issues/14366#issuecomment-761295421

> > @sakridge could this also be a legimate leak aside from the Vec HashMap capacity leak? I don't think we need 5GB / 3GB heap data at any given moment to run validator...
> 
> The recyclers just allocate to cover the transitional load and don't size down generally. Potentially we can size down with some heuristic if we can detect there is too much being held onto.

#### Summary of Changes

Amortized very-lightweight (1 vdso call for writes, 1 vdso + 2 `O(1)` VecDeque ops + 1 `T::drop()` for reads) eviction at each read from Recycler only when there has been enough capacity for the last hour.

### DIFF
--- a/perf/benches/recycler.rs
+++ b/perf/benches/recycler.rs
@@ -1,0 +1,23 @@
+#![feature(test)]
+
+extern crate test;
+
+use solana_perf::recycler::Recycler;
+use test::Bencher;
+
+#[bench]
+fn bench_recycler(bencher: &mut Bencher) {
+    solana_logger::setup();
+
+    let recycler = Recycler::default();
+
+    for _ in 0..1000 {
+        let mut p = solana_perf::packet::Packets::with_capacity(128);
+        p.packets.resize(128, solana_sdk::packet::Packet::default());
+        recycler.recycle_for_test(p.packets);
+    }
+
+    bencher.iter(move || {
+        recycler.recycle_for_test(recycler.allocate("me"));
+    });
+}
--- a/perf/src/recycler.rs
+++ b/perf/src/recycler.rs
@@ -1,3 +1,4 @@
+use log::*;
 use rand::{thread_rng, Rng};
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::{Arc, Mutex, Weak};
@@ -42,6 +43,9 @@ pub trait Reset {
     fn set_recycler(&mut self, recycler: Weak<RecyclerX<Self>>)
     where
         Self: std::marker::Sized;
+    fn unset_recycler(&mut self)
+    where
+        Self: std::marker::Sized;
 }
 
 lazy_static! {
@@ -55,6 +59,9 @@ pub fn enable_recycler_warming() {
 fn warm_recyclers() -> bool {
     WARM_RECYCLERS.load(Ordering::Relaxed)
 }
+
+pub const MAX_INVENTORY_COUNT_WITHOUT_EXPIRATION: usize = 10;
+pub const EXPIRATION_TTL_SECONDS: u64 = 3600;
 
 impl<T: Default + Reset + Sized> Recycler<T> {
     pub fn warmed(num: usize, size_hint: usize) -> Self {
@@ -76,14 +83,21 @@ impl<T: Default + Reset + Sized> Recycler<T> {
 
     pub fn allocate(&self, name: &'static str) -> T {
         let new = {
-            let mut gc = self
-                .recycler
-                .gc
-                .lock()
-                .expect("recycler lock in pb fn allocate");
-            if let Some((oldest_time, _old_item)) = gc.front() {
-                if oldest_time.elapsed().as_secs() > 3600 {
-                    gc.pop_front().unwrap();
+            let mut gc = self.recycler.gc.lock().unwrap();
+
+            // Don't expire when inventory is rather small. At least,
+            // we shouldn't expire (= drop) the last inventory; we can just
+            // return it from here.  Also add some buffer. Thus, we don't
+            // expire if less than 10.
+            if gc.len() > MAX_INVENTORY_COUNT_WITHOUT_EXPIRATION {
+                if let Some((oldest_time, _old_item)) = gc.front() {
+                    if oldest_time.elapsed().as_secs() >= EXPIRATION_TTL_SECONDS {
+                        let (_, mut expired) = gc.pop_front().unwrap();
+                        // unref-ing recycler here is crucial to prevent
+                        // the expired from being recycled again via Drop,
+                        // lading to dead lock!
+                        expired.unset_recycler();
+                    }
                 }
             }
             gc.pop_back().map(|(_added_time, item)| item)
@@ -108,6 +122,10 @@ impl<T: Default + Reset + Sized> Recycler<T> {
         let mut t = T::default();
         t.set_recycler(Arc::downgrade(&self.recycler));
         t
+    }
+
+    pub fn recycle_for_test(&self, x: T) {
+        self.recycler.recycle(x);
     }
 }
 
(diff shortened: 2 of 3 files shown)


perf/src/recycler.rs
```diff
@@ -1,7 +1,8 @@
 use rand::{thread_rng, Rng};
-use std::sync::atomic::AtomicBool;
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::{Arc, Mutex, Weak};
+use std::time::Instant;
+use std::{collections::VecDeque, sync::atomic::AtomicBool};
```
ryoqun, line 5: hehe, rustfmt became lazy. ;)

sakridge:
Nice, this seems pretty good, any perf change with the vec-deque and extra checks?

perf/src/recycler.rs
```diff
@@ -74,12 +75,19 @@ impl<T: Default + Reset + Sized> Recycler<T> {
...
-            .lock()
-            .expect("recycler lock in pb fn allocate")
-            .pop();
+        let new = {
+            let mut gc = self
+                .recycler
+                .gc
+                .lock()
+                .expect("recycler lock in pb fn allocate");
+            if let Some((oldest_time, _old_item)) = gc.front() {
```
sakridge, line 84: Thought about maybe checking the length of `gc` here. Maybe the recycler is just not used in a while and we are popping an extra one where we only have a handful of entries (1-5).

ryoqun: yeah, nice addition! Also, I noticed we shouldn't pop and drop the last one. We can just recycle it.. ;) I chose 10 for no particular strong reason. I just felt expiring with 5 items sounded too aggressive.... just that.

perf/benches/recycler.rs
```diff
@@ -0,0 +1,23 @@
...
+    let recycler = Recycler::default();
+
+    for _ in 0..1000 {
+        let mut p = solana_perf::packet::Packets::with_capacity(128);
+        p.packets.resize(128, solana_sdk::packet::Packet::default());
+        recycler.recycle_for_test(p.packets);
+    }
+
+    bencher.iter(move || {
+        recycler.recycle_for_test(recycler.allocate("me"));
```
ryoqun, line 21: > Nice, this seems pretty good, any perf change with the vec-deque and extra checks?

In summary, sadly this increased twice of fast path of recycle->allocate iteration around 70ns to 140ns. But, I think this shouldn't affect the overall cluster perf. (I'm running gcp system-performance-test to be sure).

In detail, simply changing to VecDeque increases around 10ns and 30ns * 2 by 2 Instance::now() calls by recycle and allocate.

```
base:
test bench_recycler ... bench:          73 ns/iter (+/- 22)

only changed Vec => VecDeque collection type:
test bench_recycler ... bench:          85 ns/iter (+/- 7)

this pr:
test bench_recycler ... bench:         144 ns/iter (+/- 19)
```

I think these numbers are reasonable. VecDec is basically Vec and it only some small additional offset calculation so the order should be 10ns-20ns.

As for Instance::now, I think this is decent value as well, considering it's basically vdso call, recalling my vague perf number for it. xD

Also, I also think `VecDeque` is faster than `BTreeMap` or `BinaryHeap` implementation. Although, I haven't tested.

This tested on my laptop and I know there is jitter. If you're curious, I can run more rigorous experiment in more clean room enviroment. :)

ryoqun:
As for the actual observation of leaks are fixed. I'm still running tested validator.. This takes a while.

perf/src/recycler.rs
```diff
@@ -74,12 +82,26 @@ impl<T: Default + Reset + Sized> Recycler<T> {
...
+            // we shouldn't expire (= drop) the last inventory; we can just
+            // return it from here.  Also add some buffer. Thus, we don't
+            // expire if less than 10.
+            if gc.len() > MAX_INVENTORY_COUNT_WITHOUT_EXPIRATION {
+                if let Some((oldest_time, _old_item)) = gc.front() {
+                    if oldest_time.elapsed().as_secs() >= EXPIRATION_TTL_SECONDS {
+                        let (_, mut expired) = gc.pop_front().unwrap();
+                        // unref-ing recycler here is crucial to prevent
+                        // the expired from being recycled again via Drop,
+                        // lading to dead lock!
```
ryoqun, line 98: my test validator was bitten by this. T_T

perf/src/recycler.rs
sakridge: yea, sounds good.

perf/benches/recycler.rs
sakridge: ok not bad, thanks.

ryoqun:
well, test validator survived running against mainnet-beta for a day without deadlocks. So, I think I fixed the bug.

Now testing running testnet via heaptrack to see leaks are really solved before last check.

ryoqun:
well it turned out, this fix doesn't fix the leak...... the gossip-receiver code doesn't recycle PinnedVecs at all actually. I'm trying to fix it as well.

ryoqun:
well it turned out, this fix doesn't fix the leak...... the gossip-receiver code doesn't recycle PinnedVecs at all actually. I'm trying to fix it as well.

ryoqun:
well it turned out, this fix doesn't fix the leak...... the gossip-receiver code doesn't recycle PinnedVecs at all actually. I'm trying to fix it as well.

ryoqun:
well it turned out, this fix doesn't fix the leak...... the gossip-receiver code doesn't recycle PinnedVecs at all actually. I'm trying to fix it as well.

carllin:
@ryoqun the reason this may not be recycling is because https://github.com/solana-labs/solana/pull/15139/files#diff-798271cb43dca47253c06d8ab6888d8bf4fb385fc8e74cc2cd70ac1b3b664ebfR93 checks the oldest element, but if I have `N` elements, then if:

1) I only need `1` element at a time
2) `N-1` elements are always sitting in the queue

As long as I have `N` total allocations before the max eviction time passes, then every packet will be refreshed in https://github.com/solana-labs/solana/pull/15139/files#diff-798271cb43dca47253c06d8ab6888d8bf4fb385fc8e74cc2cd70ac1b3b664ebfR136 and nothing will be recycled

ryoqun:
> @ryoqun the reason this may not be recycling is because https://github.com/solana-labs/solana/pull/15139/files#diff-798271cb43dca47253c06d8ab6888d8bf4fb385fc8e74cc2cd70ac1b3b664ebfR93 checks the oldest element, but if I have `N` elements, then if:
> 
>     1. I only need `1` element at a time
> 
>     2. `N-1` elements are always sitting in the queue
> 
> 
> As long as I have `N` total allocations before the max eviction time passes, then every packet will be refreshed by the recycle logic https://github.com/solana-labs/solana/pull/15139/files#diff-798271cb43dca47253c06d8ab6888d8bf4fb385fc8e74cc2cd70ac1b3b664ebfR136 and nothing will be recycled

@carllin Your understanding is mostly correct. Note that such `N` total allocations must be peak and at-all-once to void the intended eviction for unused inventry. That's because recyler are basically LIFO. So, recycler's inventory must be lent-over some other places at the same time, to reach to refresh the deepest VecDeque items.. Originally, I thought that it's enough to prevent mem leak due to some spikes. But I looks like my guess was off.

My current theory is that `solana-receiver` threads creates unbounded number of Packets out of Recycler and puts them in to the `mpsc::Channel` and `solana-gossip` couldn't keep with it?

ryoqun:
Also, this bug looks like a classic buffer bloat. ;)

perf/benches/recycler.rs
```diff
@@ -0,0 +1,23 @@
+#![feature(test)]
+
+extern crate test;
+
+use solana_perf::recycler::Recycler;
+use test::Bencher;
+
+#[bench]
+fn bench_recycler(bencher: &mut Bencher) {
```
ryoqun, line 9: this

perf/src/recycler.rs
```diff
@@ -74,12 +82,26 @@ impl<T: Default + Reset + Sized> Recycler<T> {
...
+        let new = {
+            let mut gc = self.recycler.gc.lock().unwrap();
+
+            // Don't expire when inventory is rather small. At least,
+            // we shouldn't expire (= drop) the last inventory; we can just
+            // return it from here.  Also add some buffer. Thus, we don't
+            // expire if less than 10.
+            if gc.len() > MAX_INVENTORY_COUNT_WITHOUT_EXPIRATION {
+                if let Some((oldest_time, _old_item)) = gc.front() {
+                    if oldest_time.elapsed().as_secs() >= EXPIRATION_TTL_SECONDS {
```
ryoqun, line 94: like this

ryoqun:
closing in favor of #15320. :)

PR #15149, opened 2021-02-05
use thread pool for non-index hash calculations

jeffwashington:
#### Problem
can't let hash calculations consume all threads

#### Summary of Changes
use a smaller thread pool

Fixes #

### DIFF
--- a/ledger-tool/src/main.rs
+++ b/ledger-tool/src/main.rs
@@ -1933,6 +1933,7 @@ fn main() {
                         Some(snapshot_version),
                         output_directory,
                         ArchiveFormat::TarZstd,
+                        &bank.get_thread_pool(),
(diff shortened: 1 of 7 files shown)


ledger-tool/src/main.rs
```diff
@@ -1933,6 +1933,7 @@ fn main() {
                         Some(snapshot_version),
                         output_directory,
                         ArchiveFormat::TarZstd,
+                        &bank.get_thread_pool(),
```
sakridge, line 1936: This could use threadpool with all threads.

jeffwashington: will address in followup pr

sakridge (approved):

PR #15151, opened 2021-02-05
Threadpool2

jeffwashington:
#### Problem
ledger tool can use all available threads for hashing

#### Summary of Changes
allow the default thread pool to be used.

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -3748,7 +3748,7 @@ impl AccountsDB {
             Self::calculate_accounts_hash_without_index(
                 &combined_maps,
                 simple_capitalization_enabled,
-                &self.thread_pool_clean,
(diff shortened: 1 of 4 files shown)


runtime/src/accounts_db.rs
```diff
@@ -3851,10 +3852,18 @@ impl AccountsDB {
     pub fn calculate_accounts_hash_without_index(
         storages: &[SnapshotStorage],
         simple_capitalization_enabled: bool,
+        thread_pool: Option<&ThreadPool>,
     ) -> (Hash, u64) {
-        let result = Self::scan_snapshot_stores(storages, simple_capitalization_enabled);
+        let function = || {
```
sakridge, line 3857: can we call it `scan_and_hash` or something more specific than `function`.

sakridge (approved):

Issue #14602, opened 2021-01-15
RPC: add caching to getLargestAccounts  


oJshua:
#### Problem
The `getLargestAccounts` RPC call takes ~14 seconds (based on haproxy logs on the server).  We should cache this in the RPC subsystem in the validator and probably only refresh it once every couple hours. 

#### Proposed Solution
Implement caching solution in the validator.

PR #15154, opened 2021-02-05
RPC: add caching to getLargestAccounts

oJshua:
#### Problem
The `getLargestAccounts` RPC call takes ~14 seconds (based on haproxy logs on the server).  We should cache this in the RPC subsystem in the validator and probably only refresh it once every couple hours. 

#### Summary of Changes
Implements a two hour cache on getLargestAccounts w/ its various config options.

Fixes https://github.com/solana-labs/solana/issues/14602

### DIFF
--- a/client/src/rpc_cache.rs
+++ b/client/src/rpc_cache.rs
@@ -1,0 +1,169 @@
+use crate::rpc_config::RpcLargestAccountsConfig;
+use crate::rpc_response::RpcAccountBalance;
+use std::collections::HashMap;
+use std::time::{Duration, SystemTime};
+
+#[derive(Debug, Clone)]
+pub struct LargestAccountsCache {
+    size: usize,
+    duration: u64,
+    cache: HashMap<RpcLargestAccountsConfig, LargestAccountsCacheValue>,
+}
+
+#[derive(Debug, Clone)]
+struct LargestAccountsCacheValue {
+    accounts: Vec<RpcAccountBalance>,
+    cached_time: SystemTime,
+}
+
+impl LargestAccountsCache {
+    pub fn new(size: usize, duration: u64) -> Self {
+        Self {
+            size,
+            duration,
+            cache: HashMap::new(),
+        }
+    }
+
+    pub fn get_largest_accounts(
+        &self,
+        config: &RpcLargestAccountsConfig,
+    ) -> Option<Vec<RpcAccountBalance>> {
+        self.cache.get(&config).map(|value| {
+            value.cached_time.elapsed().ok().map(|elapsed| {
+                if elapsed < Duration::from_secs(self.duration) {
+                    Some(value.accounts.clone())
+                } else {
+                    None
+                }
+            })?
+        })?
+    }
+
+    pub fn set_largest_accounts(
+        &mut self,
+        config: &RpcLargestAccountsConfig,
+        accounts: &[RpcAccountBalance],
+    ) {
+        if self.cache.len() >= self.size {
+            self.evict();
+        }
+
+        if self.cache.len() < self.size {
+            self.cache.insert(
+                config.clone(),
+                LargestAccountsCacheValue {
+                    accounts: accounts.to_owned(),
+                    cached_time: SystemTime::now(),
+                },
+            );
+        }
+    }
+
+    fn evict(&mut self) {
+        let duration = Duration::from_secs(self.duration);
+        self.cache.retain(|_key, value| {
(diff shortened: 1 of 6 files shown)


oJshua:
Note: this was originally intended for getProgramAccounts but I realized midway through I was supposed to be targeting getLargestAccounts. I kept the original branch in case we want to also implement caching on getProgramAccounts.

client/src/rpc_cache.rs
```diff
@@ -0,0 +1,169 @@
...
+    ) -> Option<Vec<RpcAccountBalance>> {
+        self.cache.get(&config).map(|value| {
+            value.cached_time.elapsed().ok().map(|elapsed| {
+                if elapsed < Duration::from_secs(self.duration) {
+                    Some(value.accounts.clone())
+                } else {
+                    None
+                }
+            })?
+        })?
```
CriesofCarrots, line 40:
```suggestion
        self.cache.get(&config).and_then(|value| {
            if let Ok(elapsed) = value.cached_time.elapsed() {
                if elapsed < Duration::from_secs(self.duration) {
                    return Some(value.accounts.clone());
                }
            }
            None
        })
```
I think this is slightly more Rusty and slightly easier to parse than the map-try syntax, but not a big deal.

client/src/rpc_cache.rs
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    pub fn set_largest_accounts(
+        &mut self,
+        config: &RpcLargestAccountsConfig,
+        accounts: &[RpcAccountBalance],
+    ) {
+        if self.cache.len() >= self.size {
+            self.evict();
```
CriesofCarrots, line 49: Not sure this (or the size field, for that matter) is necessary if we key by `Option<RpcLargestAccountsFilter>`, is it? There will only ever be max 3 HashMap records.

oJshua: Yep, we can do away with this now! Mostly a holdover from when I wrote this for program accounts.

client/src/rpc_cache.rs
```diff
@@ -0,0 +1,74 @@
...
+use std::time::{Duration, SystemTime};
+
+#[derive(Debug, Clone)]
+pub struct LargestAccountsCache {
+    duration: u64,
+    cache: HashMap<Option<RpcLargestAccountsFilter>, LargestAccountsCacheValue>,
+}
+
+#[derive(Debug, Clone)]
+struct LargestAccountsCacheValue {
```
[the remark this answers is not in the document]
CriesofCarrots: Might not be relevant, but if the explorer is set up to update largest accounts in real time, this could prevent the data going backward in time if it hits a different api endpoint with an older cache

client/src/rpc_cache.rs
```diff
@@ -0,0 +1,74 @@
+use crate::rpc_config::RpcLargestAccountsFilter;
+use crate::rpc_response::RpcAccountBalance;
+use std::collections::HashMap;
+use std::time::{Duration, SystemTime};
```
CriesofCarrots, line 4:
```suggestion
use std::{
    collections::HashMap,
    time::{Duration, SystemTime},
};
```

client/src/rpc_cache.rs
```diff
@@ -0,0 +1,74 @@
+use crate::rpc_config::RpcLargestAccountsFilter;
+use crate::rpc_response::RpcAccountBalance;
```
CriesofCarrots, line 2:
```suggestion
use crate::{
    rpc_config::RpcLargestAccountsFilter,
    rpc_response::RpcAccountBalance,
};
```

client/src/rpc_cache.rs
```diff
@@ -0,0 +1,74 @@
...
+        );
+    }
+}
+
+#[cfg(test)]
+pub mod test {
+    use crate::rpc_cache::LargestAccountsCache;
+    use crate::rpc_config::RpcLargestAccountsFilter;
+    use crate::rpc_response::RpcAccountBalance;
+    use std::time::Duration;
```
CriesofCarrots, line 60:
```suggestion
    use super::*;
```

client/src/rpc_cache.rs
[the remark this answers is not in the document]
oJshua: Good idea, just added.

CriesofCarrots (approved):
Code lgtm! Can you please add some documentation about this caching to https://github.com/solana-labs/solana/blob/master/docs/src/developing/clients/jsonrpc-api.md#getlargestaccounts ?

PR #15165, opened 2021-02-06
add metrics to hash calculation and consolidate

jeffwashington:
#### Problem
Performance of hash calculation is important to monitor as we get to 1B accounts.
Also, passing around metrics is cumbersome.

#### Summary of Changes
Add metrics for more parts of hash scanning and calculation and a total time.
Consolidate metrics.

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -169,6 +169,51 @@ type ShrinkCandidates = HashMap<Slot, HashMap<AppendVecId, Arc<AccountStorageEnt
 
 const ZERO_RAW_LAMPORTS_SENTINEL: u64 = std::u64::MAX;
 
+#[derive(Debug, Default)]
+pub struct HashStats {
+    pub scan_time_total_us: u64,
+    pub zeros_time_total_us: u64,
+    pub hash_time_total_us: u64,
+    pub sort_time_total_us: u64,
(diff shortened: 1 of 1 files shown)


runtime/src/accounts_db.rs
```diff
@@ -3925,41 +3945,43 @@ impl AccountsDB {
     fn scan_snapshot_stores(
         storage: &[SnapshotStorage],
         simple_capitalization_enabled: bool,
-    ) -> (Vec<Vec<CalculateHashIntermediate>>, Measure) {
+    ) -> (Vec<Vec<CalculateHashIntermediate>>, Measure, usize, Measure) {
```
sakridge, line 3948: Passing around `Measure` is getting awkward, we have a somewhat standard way of accumulating metrics into mutable stats struct, can we match that pattern so it looks like the rest of the code?

sakridge: CleanKeyTimings is one example:
https://github.com/solana-labs/solana/blob/ab0f4c69aaea10fa62b887d3212d0b09c026c0f9/runtime/src/accounts_db.rs#L1321-L1324
or
ShrinkStats:
https://github.com/solana-labs/solana/blob/ab0f4c69aaea10fa62b887d3212d0b09c026c0f9/runtime/src/accounts_db.rs#L874-L890

PR #15177, opened 2021-02-06
Add jit and caching args to ledger-tool

sakridge:
#### Problem

JIT and accounts cache arguments not plumbed in ledger-tool

#### Summary of Changes

Add them.

Fixes #

### DIFF
--- a/ledger-tool/src/main.rs
+++ b/ledger-tool/src/main.rs
@@ -742,6 +742,14 @@ fn main() {
         .long("no-snapshot")
         .takes_value(false)
         .help("Do not start from a local snapshot if present");
+    let bpf_jit_arg = Arg::with_name("bpf_jit")
+        .long("bpf-jit")
+        .takes_value(false)
+        .help("Process with JIT instead of interpreter.");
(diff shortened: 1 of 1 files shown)


ledger-tool/src/main.rs
```diff
@@ -742,6 +742,14 @@ fn main() {
         .long("no-snapshot")
         .takes_value(false)
         .help("Do not start from a local snapshot if present");
+    let bpf_jit_arg = Arg::with_name("bpf_jit")
+        .long("bpf-jit")
+        .takes_value(false)
+        .help("Process with JIT instead of interpreter.");
+    let accounts_db_caching_arg = Arg::with_name("accounts_db_caching_enabled")
```
mvines, line 749: Do you mind if we drop this and add `--no-accounts-db-caching` instead to match `solana-validator`'s comand-line interface?
https://github.com/solana-labs/solana/blob/47578895e1ac94aad5a1f18021e6525e19d464f1/validator/src/main.rs#L1419-L1423

sakridge: ok

Issue #15180, opened 2021-02-07
Capitalization panic in testnet simulation


sakridge:
#### Problem

Capitalization panic in testnet simulation:
https://discordapp.com/channels/428295358100013066/439194979856809985/807491877015388160

```
2021-02-06 05:26:32 | Eqr2dzMoqHZ3t8QpYyDs8JtUQMYpWmTzZ2ujSNgN3FEv | validator | solana-accounts-background | 
panicked at 'assertion failed: `(left == right)` left: `1000189065363148006`, right: `1000189065482728910`', runtime/src/bank.rs:4334:9

2021-02-06 05:16:32 | 9hm9DnKHS5zKYQDCnRiA58fKttVXb7BSPDyKnuCDzTMT | validator | solana-accounts-background | 
panicked at 'assertion failed: `(left == right)` left: `1000189066716389049`, right: `1000189067148716410`', runtime/src/bank.rs:4334:9
```

#### Proposed Solution

Debug and fix it.

sakridge:
cc @ryoqun

sakridge:
Nodes that experienced the panic:
`34.74.225.200     |   831 | Eqr2dzMoqHZ3t8QpYyDs8JtUQMYpWmTzZ2ujSNgN3FEv |    -    | 8001 | none | none | none | none | none | none | 19458`
`34.74.94.243      |   507 | 9hm9DnKHS5zKYQDCnRiA58fKttVXb7BSPDyKnuCDzTMT |  1.5.6  | 8001 | 8003 | 8004 | 8000 | 8002 | 8006 | 8007 | 19458`

sakridge:
From Eqr2dzMoqHZ3t8QpYyDs8JtUQMYpWmTzZ2ujSNgN3FEv
```
[2021-02-06T05:26:32.103469481Z INFO  solana_core::replay_stage] new fork:202368 parent:202367 root:202330
[2021-02-06T05:26:32.103551770Z INFO  solana_metrics::metrics] datapoint: bank-new_from_parent-heights slot_height=202368i block_height=190360i
[2021-02-06T05:26:32.103835935Z INFO  solana_metrics::metrics] datapoint: bank-timestamp get_timestamp_estimate_us=8i
[2021-02-06T05:26:32.103843608Z INFO  solana_metrics::metrics] datapoint: bank-timestamp-correction slot=202368i from_genesis=1612555439i corrected=1612588945i ancestor_timestamp=1612588945i
[...]

PR #15189, opened 2021-02-08
use index version of calculating hash

jeffwashington:
#### Problem
panic hit on simulation with capitalization mismatch after hash calculation.
#### Summary of Changes
revert to previous hash calculation algorithm.
Fixes #

### DIFF
--- a/runtime/src/accounts_background_service.rs
+++ b/runtime/src/accounts_background_service.rs
@@ -122,8 +122,11 @@ impl SnapshotRequestHandler {
 
                 let mut hash_time = Measure::start("hash_time");
                 let mut hash_for_testing = None;
-                snapshot_root_bank
-                    .update_accounts_hash_with_index_option(true, test_hash_calculation);
+                const DO_NOT_USE_INDEX: bool = false;
+                snapshot_root_bank.update_accounts_hash_with_index_option(
+                    DO_NOT_USE_INDEX,
+                    test_hash_calculation,
+                );
                 if test_hash_calculation {
                     hash_for_testing = Some(snapshot_root_bank.get_accounts_hash());
                 }
--- a/runtime/src/accounts_background_service.rs
+++ b/runtime/src/accounts_background_service.rs
@@ -123,10 +123,8 @@ impl SnapshotRequestHandler {
                 let mut hash_time = Measure::start("hash_time");
                 let mut hash_for_testing = None;
                 const USE_INDEX: bool = true;
-                snapshot_root_bank.update_accounts_hash_with_index_option(
-                    !USE_INDEX,
-                    test_hash_calculation,
-                );
+                snapshot_root_bank
+                    .update_accounts_hash_with_index_option(!USE_INDEX, test_hash_calculation);
                 if test_hash_calculation {
                     hash_for_testing = Some(snapshot_root_bank.get_accounts_hash());
                 }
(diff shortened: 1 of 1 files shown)


jeffwashington:
#15180

mvines:
(adding v1.5, as I think we want this patch there too)

runtime/src/accounts_background_service.rs
```diff
@@ -122,8 +122,11 @@ impl SnapshotRequestHandler {
 
                 let mut hash_time = Measure::start("hash_time");
                 let mut hash_for_testing = None;
-                snapshot_root_bank
-                    .update_accounts_hash_with_index_option(true, test_hash_calculation);
+                const DO_NOT_USE_INDEX: bool = false;
+                snapshot_root_bank.update_accounts_hash_with_index_option(
+                    DO_NOT_USE_INDEX,
+                    test_hash_calculation,
+                );
```
sakridge, line 129: unrelated, but we could write hash_for_testing as 
```
let hash_for_testing = if test_hash_calculations {
     Some(..)
} else {
     None
 };
 ```

**Review by sakridge — APPROVED** [review]

File: runtime/src/accounts_background_service.rs [hunk 2]
```diff
@@ -122,8 +122,11 @@ impl SnapshotRequestHandler {
 
                 let mut hash_time = Measure::start("hash_time");
                 let mut hash_for_testing = None;
-                snapshot_root_bank
-                    .update_accounts_hash_with_index_option(true, test_hash_calculation);
+                const DO_NOT_USE_INDEX: bool = false;
```
CriesofCarrots, line 125: Fwiw, I find the double-negative a little difficult to follow. If the churn isn't too awful, could we update this to `use_index` in `update_accounts_hash_with_index_option` and etc?

jeffwashington:
stable build failed once. I ran the test locally and it passed. I retry'd the build and it succeeded and is continuing.
https://buildkite.com/solana-labs/solana/builds/39884#_
test: test_stake_authorize_with_fee_payer

runtime/src/accounts_background_service.rs
jeffwashington: handled in #15191

runtime/src/accounts_background_service.rs
```diff
@@ -122,8 +122,11 @@ impl SnapshotRequestHandler {
 
                 let mut hash_time = Measure::start("hash_time");
                 let mut hash_for_testing = None;
-                snapshot_root_bank
-                    .update_accounts_hash_with_index_option(true, test_hash_calculation);
+                const USE_INDEX: bool = true;
+                snapshot_root_bank.update_accounts_hash_with_index_option(
+                    !USE_INDEX,
```
CriesofCarrots, line 127: I actually meant to continue the inversion upstream
https://github.com/solana-labs/solana/blob/88f22c360bce341a4aa10564f5f3ae76df8312c7/runtime/src/bank.rs#L4317, https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_db.rs#L3890
So all `do_not_use_index` > `use_index`

But this works for now if that's too much churn for this PR
But

jeffwashington: My preference would be to change this separately, once the dust settles, and it seems you are ok with that.

CriesofCarrots: Yep, works for me.

CriesofCarrots (approved):

PR #15191, opened 2021-02-08
pass expected capitalization to hash calculation to improve assert msg

jeffwashington:
Problem
With a debug flag to the validator (--accounts-db-test-hash-calculation), the validator can calculate hash and lamports 2 ways and compare. If the two ways conflict, we fail to report the comparison of lamports to the bank's expected capitalization.

Summary of Changes
Pass the bank's expected capitalization and compare it and print it on assertion.
Also, cleanup an if, and remove a double-negative from a parameter.

Fixes #

### DIFF
--- a/accounts-bench/src/main.rs
+++ b/accounts-bench/src/main.rs
@@ -113,6 +113,7 @@ fn main() {
                 solana_sdk::clock::Slot::default(),
                 &ancestors,
                 true,
+                None,
             );
             time_store.stop();
             if results != results_store {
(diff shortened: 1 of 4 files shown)


CriesofCarrots (approved):
lgtm

sakridge (approved):

jeffwashington:
@sakridge does this belong in 1.5?

sakridge:
> @sakridge does this belong in 1.5?

Seems pretty safe to put in 1.5. The extra information in the assert could be useful.

PR #15194, opened 2021-02-08
uses btree-map instead of hash-map for cluster-slots

behzadnouri:
#### Problem
retain traverses all values in the hashmap which is slow:
https://github.com/solana-labs/solana/blob/88f22c360/core/src/cluster_slots.rs#L45
btree-map instead allows more efficient prunning there.

In addition there is potential race condition here:
https://github.com/solana-labs/solana/blob/88f22c360/core/src/cluster_slots.rs#L68-L74
If another thread inserts a value at the same slot key between the read
and write lock, current thread will discard the inserted value.

#### Summary of Changes
* changed to btree-map for more efficient pruning.
* removed read-lock-then-write-lock code to avoid race condition.

### DIFF
--- a/core/src/cluster_slots.rs
+++ b/core/src/cluster_slots.rs
@@ -5,16 +5,15 @@ use crate::{
 use solana_runtime::{bank_forks::BankForks, epoch_stakes::NodeIdToVoteAccounts};
 use solana_sdk::{clock::Slot, pubkey::Pubkey};
 use std::{
-    collections::{HashMap, HashSet},
+    collections::{BTreeMap, HashMap, HashSet},
     sync::{Arc, RwLock},
 };
 
 pub type SlotPubkeys = HashMap<Pubkey, u64>;
-pub type ClusterSlotsMap = RwLock<HashMap<Slot, Arc<RwLock<SlotPubkeys>>>>;
 
 #[derive(Default)]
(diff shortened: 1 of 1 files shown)


behzadnouri:
Left chunk is this code:

`cluster_slots_service-timing::update_completed_slots_elapased`::
cluster_slots_service-timing update_completed_slots_elapased

core/src/cluster_slots.rs
```diff
@@ -42,7 +40,20 @@ impl ClusterSlots {
...
+            .keys()
+            .take_while(|slot| **slot <= root)
+            .copied()
+            .collect();
+        {
+            let mut cluster_slots = self.cluster_slots.write().unwrap();
+            for slot in old_slots {
+                cluster_slots.remove(&slot);
+            }
+        }
```
carllin, line 56: is a `split_off` followed by a `swap` more efficient here? i.e. something like: https://github.com/solana-labs/solana/blob/master/core/src/optimistic_confirmation_verifier.rs#L30-L34

behzadnouri: Doesn't seem different, but I switched to `split_off` anyways. thanks

sakridge (approved):
lgtm, nice improvement

carllin (approved):

PR #15202, opened 2021-02-09
Specify Custom VM Type

carllin:
#### Problem
Current gce deployment logic for bootstrap leader sometimes provision machines that have insufficient compute to pass the poh speed check, causing startup to fail

#### Summary of Changes
Specify a particular VM type

Thanks @t-nelson for the help 😃 

Fixes #


t-nelson (approved):
I'm pretty surprised we haven't been bit by this earlier.  Maybe E2 Skylake is a new thing...

Thanks for PRing the change!  LGTM!

PR #15204, opened 2021-02-09
Warp timestamp and extend max-allowable-drift for accommodate slow blocks

CriesofCarrots:
#### Problem
Blocks have slowed, making current `MAX_ALLOWABLE_DRIFT_PERCENTAGE` again too restrictive and causing timestamp to drift away from (behind) real-world time.

#### Summary of Changes
- Remove old feature gating and deprecated code (`test_bank_hash_consistency` needs updating, since it runs without active features)
- Allow independent adjustment of max_allowable_drift values for blocks slower than PoH vs blocks faster than PoH
- Extend max_allowable_drift percentage to 80% for slow blocks: 720ms at current PoH target
- On activation, warp timestamp to current estimate with no bounding
- Also fix a i64 to u64 casting issue (unlikely to occur in the wild, since it is in the too-fast path, but feature-gated for safety first)

### DIFF
--- a/sdk/src/feature_set.rs
+++ b/sdk/src/feature_set.rs
@@ -230,10 +230,6 @@ pub mod try_find_program_address_syscall_enabled {
     solana_sdk::declare_id!("EMsMNadQNhCYDyGpYH5Tx6dGHxiUqKHk782PU5XaWfmi");
 }
 
-pub mod warp_timestamp {
-    solana_sdk::declare_id!("Bfqm7fGk5MBptqa2WHXWFLH7uJvq8hkJcAQPipy2bAMk");
-}
-
 pub mod stake_program_v3 {
     solana_sdk::declare_id!("Ego6nTu7WsBcZBvVqJQKp6Yku2N3mrfG8oYCfaLZkAeK");
 }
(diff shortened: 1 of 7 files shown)


CriesofCarrots:
Not quite ready. One more chunk of deprecated code to remove.

CriesofCarrots:
This is ready for a look @mvines 
The actual behavior changes (bounding adjustment, time warp) are all in https://github.com/solana-labs/solana/pull/15204/commits/1300ba4ec483f3afbbc916c9a59f0aa1b0046f81

All the other commits are cleanup/refactoring, although if you have time, I'd appreciate a sanity check on them.

sdk/src/feature_set.rs
```diff
@@ -358,6 +347,7 @@ lazy_static! {
         (full_inflation::stakeconomy::vote::id(), "Community vote allowing Stakeconomy.com to enable full inflation"),
         (full_inflation::w3m::vote::id(), "Community vote allowing w3m to enable full inflation"),
         (full_inflation::w3m::enable::id(), "Full inflation enabled by w3m"),
+        (warp_timestamp_again::id(), "warp timestamp again, adjust bounding to 25% fast 80% slow #TODO"),
```
mvines, line 350:
```suggestion
        (warp_timestamp_again::id(), "warp timestamp again, adjust bounding to 25% fast 80% slow #15204"),
```

mvines (approved):

sdk/src/feature_set.rs
CriesofCarrots: ty!

mvines:
@CriesofCarrots - can this be merged?  I'm about to ship a v1.5, so we might as well get this in too

CriesofCarrots:
> @CriesofCarrots - can this be merged? I'm about to ship a v1.5, so we might as well get this in too

Yep! Sorry for the delay

PR #15210, opened 2021-02-09
Cleanup buffered packets

carllin:
1) `push_unprocessed()` calls in `process_packets()` check that the length of the local `unprocessed_packets` variable is under the max length, but then this entire thing is appended to the global buffered queue, regardless of the global buffered queue length:https://github.com/solana-labs/solana/blob/c1e93b3ef9d7c60b16e93f5a1a3144ca95488aec/core/src/banking_stage.rs#L442

2) Simplify `consume_buffered_packets()` as it's overly complicated and does not need to worry about the buffer's max limit if `process_packets()` manages that properly.

3) Through 1) and 2) remove redundant appends/extends that happen after calls to `process_packets()` and `consume_buffered_packets()` respectively

4) Add tests for `consume_buffered_packets()`

Fixes #


core/Cargo.toml
```diff
@@ -44,6 +44,7 @@ rand_chacha = "0.2.2"
 raptorq = "1.4.2"
 rayon = "1.5.0"
 regex = "1.3.9"
+retain_mut = "0.1.2"
```
ryoqun, line 47: :eyes: TIL

ryoqun:
@carllin FYI, beautified permalinks are actually easy to copy: 
image

core/src/banking_stage.rs
```diff
@@ -2108,4 +2098,177 @@ mod tests {
...
+                &mut buffered_packets,
+                None,
+                &gossip_vote_sender,
+                None::<Box<dyn Fn()>>,
+            );
+            assert_eq!(buffered_packets[0].1.len(), num_conflicting_transactions);
+
+            // When the poh recorder has a bank, should process all non conflicting buffered packets.
+            // Processes one packet per iteration of the loop
+            for i in 0..num_conflicting_transactions {
```
ryoqun, line 2178: nits: `for num_expected_unprocessed in (0..num_conflicting_transactions).iter().rev()` should do the trick to remove this rather complicated math?: `let num_expected_unprocessed = num_conflicting_transactions - i - 1;`

core/src/banking_stage.rs
```diff
@@ -2108,4 +2098,177 @@ mod tests {
...
+            Some((4, 4)),
+            bank.ticks_per_slot(),
+            &solana_sdk::pubkey::new_rand(),
+            &Arc::new(blockstore),
+            &Arc::new(LeaderScheduleCache::new_from_bank(&bank)),
+            &Arc::new(PohConfig::default()),
+        );
+        let poh_recorder = Arc::new(Mutex::new(poh_recorder));
+
+        // Set up conflicting transactions
```
ryoqun, line 2133: nits: `// Set up unparallelizable conflicting transactions`

core/src/banking_stage.rs
```diff
@@ -2108,4 +2098,177 @@ mod tests {
...
+            let mut buffered_packets: UnprocessedPackets = vec![(
+                all_packets,
+                (0..num_conflicting_transactions).into_iter().collect(),
+            )]
+            .into_iter()
+            .collect();
+
+            let (gossip_vote_sender, _gossip_vote_receiver) = unbounded();
+
+            // When the working bank in poh_recorder is None, no packets should be processed
```
ryoqun, line 2165: nits: as par this comment, add `assert!(!poh_recorder.lock().unwrap().has_bank());`?

core/src/banking_stage.rs
```diff
@@ -2108,4 +2098,177 @@ mod tests {
...
+                        None,
+                        &gossip_vote_sender,
+                        test_fn,
+                    );
+
+                    // Check everything is correct. All indexes after `interrupted_iteration`
+                    // should still be unprocessed
+                    assert_eq!(
+                        buffered_packets.len(),
+                        num_conflicting_transactions - interrupted_iteration - 1
```
ryoqun, line 2247: nits: `assert_eq!(buffered_packets.len(), packets[interrupted_iteration + 1..].iter().count())`?

core/src/banking_stage.rs
```diff
@@ -2108,4 +2098,177 @@ mod tests {
...
+                .clone()
+                .into_iter()
+                .map(|single_packets| (single_packets, vec![0]))
+                .collect();
+
+            let (continue_sender, continue_receiver) = unbounded();
+            let (finished_packet_sender, finished_packet_receiver) = unbounded();
+
+            let test_fn = Some(move || {
+                finished_packet_sender.send(1).unwrap();
```
ryoqun, line 2221: it looks like actually anything is ok to `send()`, how about using the unit type?: `()`

ryoqun: also, some naming is desired like `let control_flow_token = ();`.

core/src/banking_stage.rs
```diff
@@ -2108,4 +2098,177 @@ mod tests {
...
+                    {
+                        assert_eq!(
+                            remaining_unprocessed_packet.packets[0],
+                            original_packet.packets[0]
+                        );
+                    }
+                })
+                .unwrap();
+
+            for i in 0..=interrupted_iteration {
```
ryoqun, line 2261: nits: `for interrupt_now in [false, true] {`

core/src/banking_stage.rs
```diff
@@ -2108,4 +2098,177 @@ mod tests {
...
+                finished_packet_receiver.recv().unwrap();
+                if i == interrupted_iteration {
+                    // Set up a MaxHeightReached failure by clearing the working bank
+                    poh_recorder.lock().unwrap().reset(Hash::default(), 1, None);
```
ryoqun, line 2265: nits: If possible, I want to avoid comments; we don't want to maintain it. ;) and this looks somewhat common and useful thing to define at poh_recorder.rs. How about this eloquent-without-comment style?

```
impl PohRecorder {
    #[cfg(test)]
    fn schedule_dummy_max_height_reached_failure(&self) {
        self.lock().unwrap().reset(Hash::default(), 1, None);
    }
}

File: core/src/banking_stage.rs [hunk 9]
```diff
@@ -2108,4 +2098,177 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_consume_buffered_packets_interrupted() {
+        let ledger_path = get_tmp_ledger_path!();
+        {
+            let (transactions, bank, poh_recorder, _entry_receiver) =
+                setup_conflicting_transactions(&ledger_path);
+            let num_conflicting_transactions = transactions.len();
+            let packets = to_packets_chunked(&transactions, 1);
```
ryoqun, line 2206: (this is off-topic; but I'd like more verbose name like `packets_vec` than just `packets`. too hard to know the shape of given collection by name. Still `packets_vec` looks awful, showing our deficiency of proper abstraction; but it's better still, imho...)

core/src/banking_stage.rs
```diff
@@ -2108,4 +2098,177 @@ mod tests {
...
+                } else {
+                    assert_eq!(buffered_packets[0].1.len(), num_expected_unprocessed);
+                }
+            }
+        }
+        Blockstore::destroy(&ledger_path).unwrap();
+    }
+
+    #[test]
+    fn test_consume_buffered_packets_interrupted() {
```
ryoqun, line 2200: this and the above test are nicely-sized for readers. :) Well done. I learned banking_stage from these. :)

core/src/banking_stage.rs
ryoqun: or `continuation_token`?

core/benches/banking_stage.rs
```diff
@@ -81,9 +82,9 @@ fn bench_consume_buffered(bencher: &mut Bencher) {
                 &my_pubkey,
```
ryoqun, line 82: how about perf change? Maybe, negligible?

core/src/banking_stage.rs
```diff
@@ -180,78 +182,68 @@ impl BankingStage {
...
         poh_recorder: &Arc<Mutex<PohRecorder>>,
-        buffered_packets: &mut Vec<PacketsAndOffsets>,
-        batch_limit: usize,
+        buffered_packets: &mut UnprocessedPackets,
         transaction_status_sender: Option<TransactionStatusSender>,
         gossip_vote_sender: &ReplayVoteSender,
-    ) -> UnprocessedPackets {
-        let mut unprocessed_packets = vec![];
-        let mut rebuffered_packets = 0;
+        test_fn: Option<impl Fn()>,
```
ryoqun, line 188: it looks like I succeeded to join you to the `impl` static dispatch closure club. :)

core/src/banking_stage.rs
```diff
@@ -427,20 +413,11 @@ impl BankingStage {
...
-                        continue;
-                    }
-                    let num: usize = unprocessed_packets
-                        .iter()
-                        .map(|(_, unprocessed)| unprocessed.len())
-                        .sum();
-                    inc_new_counter_info!("banking_stage-buffered_packets", num);
-                    buffered_packets.append(&mut unprocessed_packets);
-                }
+                _ => (),
```
ryoqun, line 420: As this could swallow any newly-added `Err`s (maintenance burden), I think this should be `Ok(())`.

ryoqun: or more like this:

```rust
  Ok(()) | Err(RecvTimeoutError::Timeout) => (),
  Err(RecvTimeoutError::Disconnected) => break,
```

core/src/banking_stage.rs
```diff
@@ -427,20 +413,11 @@ impl BankingStage {
...
                 Err(RecvTimeoutError::Disconnected) => break,
-                Ok(mut unprocessed_packets) => {
-                    if unprocessed_packets.is_empty() {
-                        continue;
-                    }
-                    let num: usize = unprocessed_packets
-                        .iter()
-                        .map(|(_, unprocessed)| unprocessed.len())
-                        .sum();
-                    inc_new_counter_info!("banking_stage-buffered_packets", num);
```
ryoqun, line 441: maybe this metrics isn't important anymore?

core/src/banking_stage.rs
```diff
@@ -180,78 +182,68 @@ impl BankingStage {
...
-                    unprocessed_indexes.to_owned(),
-                    transaction_status_sender.clone(),
-                    gossip_vote_sender,
+                    &old_unprocessed_indexes,
+                    my_pubkey,
+                    *next_leader,
                 );
+                *old_unprocessed_indexes = new_unprocessed_indexes;
+                Self::packet_has_more_unprocessed_transactions(&old_unprocessed_indexes)
+            } else {
```
ryoqun, line 208: nits maybe, `} else if let Some(bank) = poh_recorder.lock().unwrap().bank() { ...` and `put rebuffered_packets_len += ...` into yet another  real `else { ... }`?

core/src/banking_stage.rs
```diff
@@ -180,78 +182,68 @@ impl BankingStage {
...
                     &msgs,
-                    unprocessed_indexes.to_owned(),
-                    transaction_status_sender.clone(),
-                    gossip_vote_sender,
+                    &old_unprocessed_indexes,
+                    my_pubkey,
+                    *next_leader,
                 );
+                *old_unprocessed_indexes = new_unprocessed_indexes;
+                Self::packet_has_more_unprocessed_transactions(&old_unprocessed_indexes)
```
ryoqun, line 207: (this snippet does `*old_unprocessed_indexes = new_unprocessed_indexe` even if `Self::packet_has_more_unprocessed_transactions()` could be `true`.)

core/src/banking_stage.rs
```diff
@@ -180,78 +182,68 @@ impl BankingStage {
...
+                }
+                let has_more_unprocessed_transactions =
+                    Self::packet_has_more_unprocessed_transactions(&new_unprocessed_indexes);
+                if has_more_unprocessed_transactions {
+                    *old_unprocessed_indexes = new_unprocessed_indexes;
                 }
+                if let Some(test_fn) = &test_fn {
+                    test_fn();
+                }
+                has_more_unprocessed_transactions
```
ryoqun, line 244: Thought about this subtle difference between this and there(https://github.com/solana-labs/solana/pull/15210/files#r573986343). And I just noticed this is basically same thing? How about defining a tiny helper like this?

```rust
let should_retain = update_retained_packets_with_still_unprocessed(...)
if let Some(test_fn) = ... {   ...   }
should_retain

...

fn update_retained_packets_with_still_unprocessed(old_unprocessed_indexes: &mut Vec<usize>, new_unprocessed_indexes: _) -> bool {
  let should_retain = Self::packet_has_more_unprocessed_transactions(&new_unprocessed_indexes);
  if should_retain {
    *old_unprocessed_indexes = new_unprocessed_indexes
  };
  should_retain;
}
```

ryoqun:
@carllin initial review pass is done! Looks like you really made this whole thing simpler. :) If you're asking lgtm specifically from me to merge, it'll take some more time; I'm complete newbie in BankingStage, still learning...

core/src/banking_stage.rs
```diff
@@ -180,78 +182,68 @@ impl BankingStage {
...
-                new_unprocessed_indexes,
-                &mut dropped_batches_count,
-                batch_limit,
-            );
-
-            if processed < verified_txs_len {
-                let next_leader = poh_recorder.lock().unwrap().next_slot_leader();
-                // Walk thru rest of the transactions and filter out the invalid (e.g. too old) ones
-                #[allow(clippy::while_let_on_iterator)]
-                while let Some((msgs, unprocessed_indexes)) = buffered_packets_iter.next() {
```
ryoqun, line 237: one less shallower looping gives my brain :candy:

core/src/banking_stage.rs
carllin: thanks for the review!

core/src/banking_stage.rs
carllin: hmmm `interrupted_iteration` isn't necessarily equal to `1`, so this is not always equivalent right?

core/src/banking_stage.rs
carllin: agreed, changed!

core/src/banking_stage.rs
carllin: yup that's better, fixed!

core/src/banking_stage.rs
carllin: yeah it's covered now by the `banking_stage-newly_buffered_packets_count` metric, but I think it's also not a bad idea to record the total buffered size in `banking_stage-total_buffered_packets_count`

carllin: Aggregated all the metrics under BankingStageStats

core/src/banking_stage.rs
carllin: yeah that's much better, thanks!

core/src/banking_stage.rs
carllin: heh yup, but then the `retain_mut` should remove the entire entry regardless right?

core/src/banking_stage.rs
carllin: Updated

core/src/banking_stage.rs
carllin: Edit: Unfortunately `else if let Some(bank) = poh_recorder.lock().unwrap().bank()` holds the `poh_recorder` lock for the entire block, which is needed for `process_received_packets`. Hence had to revert to the original where we have to assign the `poh_recorder.lock().unwrap().bank()` first to a variable

core/benches/banking_stage.rs
carllin: Heh comparing to the tip of master `f2f4003e706f7baa8534ac220c55eaa817fb1df3`, it actually seems about 10% faster, the bench went from average `92ns` to `82ns` over 10 trials. The numbers are small enough where this could be noise though

core/src/banking_stage.rs
ryoqun: odd? it's not `mut`:

https://github.com/solana-labs/solana/pull/15210/files#diff-bb8464c2f02c8863fb60e6e29f023f3e6da93a93fbb9199a70a44c9c1b41c7f3R2332
> `let interrupted_iteration = 1;`

Maybe, some debug leftover?

core/src/banking_stage.rs
```diff
@@ -2108,4 +2203,181 @@ mod tests {
...
+
+            for i in 0..=interrupted_iteration {
+                finished_packet_receiver.recv().unwrap();
+                if i == interrupted_iteration {
+                    poh_recorder
+                        .lock()
+                        .unwrap()
+                        .schedule_dummy_max_height_reached_failure();
+                }
+                continue_sender.send(1).unwrap();
```
ryoqun, line 2376: this `1` also can be replaced with `()`? hehe, my laziness here. ;) My other comment maybe didn't mention this.

core/src/banking_stage.rs
ryoqun: I see! Thanks for trying. :)

core/benches/banking_stage.rs
ryoqun: good. cleaner and faster. It can't go wrong. xD

core/src/banking_stage.rs
ryoqun: yeah, that's right. Also, the helper you added shines. :)

core/src/banking_stage.rs
ryoqun: :eyes:

core/src/banking_stage.rs
ryoqun: ... and please remember this: https://github.com/solana-labs/solana/pull/15210#discussion_r573865727 :)

core/src/banking_stage.rs
```diff
@@ -177,81 +253,84 @@ impl BankingStage {
...
-                        &mut unprocessed_packets,
-                        msgs,
-                        unprocessed_indexes,
-                        &mut dropped_batches_count,
-                        batch_limit,
-                    );
+                Self::update_buffered_packets_with_new_unprocessed(
+                    old_unprocessed_indexes,
+                    new_unprocessed_indexes,
+                )
```
ryoqun, line 298: nits: how about renaming `old_unprocessed_indexes` => `original_unprocessed_index`

this is a bit nuanced. Firstly, both old/new, original/new is common adjective pairs for temporal progression.

to me, old/new has equal weights with regards to each other; so, they should be treated equally. Usually, they're kept in scope to reference them at the same time.

on the other hand, `original`/`new` has some inbalance for their relations. this is like base/derived or master/slave or primary/secondary.

So, when I see `update_...(old_..., new_...);` I can't immediately know how this update_... does with the two arguments. It updates new with old? or updates old with new?

On the other hand, when I see `update_...(original_..., new_...)` I can immedaitely tell it updates original with something _new_. . :)

Also, you can pick `orig_` if you feel long for `original_`.

core/src/banking_stage.rs
```diff
@@ -177,81 +253,84 @@ impl BankingStage {
         Ok(())
     }
 
+    // Returns whether the given `Packets` has any more remaining unprocessed
+    // transactions
+    fn update_buffered_packets_with_new_unprocessed(
+        old_unprocessed_indexes: &mut Vec<usize>,
```
ryoqun, line 259: same as https://github.com/solana-labs/solana/pull/15210/files#r574501262

core/src/banking_stage.rs
```diff
@@ -1036,31 +1115,47 @@ impl BankingStage {
...
         dropped_batches_count: &mut usize,
+        newly_buffered_packets_count: &mut usize,
         batch_limit: usize,
     ) {
-        if !packet_indexes.is_empty() {
+        if Self::packet_has_more_unprocessed_transactions(&packet_indexes) {
             if unprocessed_packets.len() >= batch_limit {
-                unprocessed_packets.remove(0);
+                unprocessed_packets.pop_front();
                 *dropped_batches_count += 1;
```
ryoqun, line 1148: nits: how about swapping this? Let's coding thyme here: I see other places follows `some counter bookkeeping` then `actual collection mutation`. (Even the 3-lines below does this). So, let's align this outlier with that convention. :)

core/src/banking_stage.rs
```diff
@@ -177,81 +253,84 @@ impl BankingStage {
...
+                            gossip_vote_sender,
+                        );
+                    new_tx_count += processed;
+                    // Out of the buffered packets just retried, collect any still unprocessed
+                    // transactions in this batch for forwarding
+                    rebuffered_packets_len += new_unprocessed_indexes.len();
+                    if processed < verified_txs_len {
+                        reached_end_of_slot =
+                            Some((poh_recorder.lock().unwrap().next_slot_leader(), bank));
+                    }
```
ryoqun, line 318: how about moving this `if` to reside immediately after calling `process_received_packets()`? 

This code functions like proactive looping control code for the next iterations, so I rather don't want to put this in middle of bookkeeping of _this_ iteration if not required:

```rust
let (processed, verified_.., ...) = Self::process_...();
if processed < verified_... {
    // mark subsequent iterations as being skipped
    reached_... = ...
}

new_tx_count += processed;
// Out of the buffered packets just retried, collect any still unprocessed
// transactions in this batch for forwarding
rebuffered_packets_len += new_unprocessed_indexes.len()
let has_more_ ...

if let Some(test_fn) = ... {
    ...
}

has_more_....
```

(also note that my delicate preference is encoded in as a form of chunking by blank lines... lol)

Because of the current location of this code, I wondered while `return true/false` was needed inside this `if`

also, it seems that data dependency allows this. :)

core/src/banking_stage.rs
```diff
@@ -177,81 +253,84 @@ impl BankingStage {
...
+                            old_unprocessed_indexes,
+                            new_unprocessed_indexes,
+                        );
+                    if let Some(test_fn) = &test_fn {
+                        test_fn();
+                    }
+                    has_more_unprocessed_transactions
+                } else {
+                    rebuffered_packets_len += old_unprocessed_indexes.len();
+                    Self::packet_has_more_unprocessed_transactions(&old_unprocessed_indexes)
```
ryoqun, line 330: nits: this can be probed to always be `true` like the original has been doing? It, `old_unprocessed_indexes` in buffered_packets can't be empty if not processed.

core/src/banking_stage.rs
```diff
@@ -2108,4 +2203,181 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_consume_buffered_packets() {
+        let ledger_path = get_tmp_ledger_path!();
+        {
+            let (transactions, bank, poh_recorder, _entry_receiver) =
+                setup_conflicting_transactions(&ledger_path);
+            let num_conflicting_transactions = transactions.len();
+            let mut packets = to_packets_chunked(&transactions, num_conflicting_transactions);
```
ryoqun, line 2257: I miss `packets_vec` in the below test fn. :)

core/src/banking_stage.rs
```diff
@@ -2108,4 +2203,181 @@ mod tests {
...
+                &poh_recorder,
+                &mut buffered_packets,
+                None,
+                &gossip_vote_sender,
+                None::<Box<dyn Fn()>>,
+                None,
+            );
+            assert_eq!(buffered_packets[0].1.len(), num_conflicting_transactions);
+            // When the poh recorder has a bank, should process all non conflicting buffered packets.
+            // Processes one packet per iteration of the loop
```
ryoqun, line 2283: nits: `Processes exactly such a single transaction per iteration of the loop`. sadly, packet here is ambiguous between singular of `buffered packets` (= `Packets`), or really a singular of an element in `Packets`....

core/src/banking_stage.rs
```diff
@@ -2108,4 +2203,181 @@ mod tests {
...
+        let pubkey2 = solana_sdk::pubkey::new_rand();
+        let transactions = vec![
+            system_transaction::transfer(&mint_keypair, &pubkey0, 1, genesis_config.hash()),
+            system_transaction::transfer(&mint_keypair, &pubkey1, 1, genesis_config.hash()),
+            system_transaction::transfer(&mint_keypair, &pubkey2, 1, genesis_config.hash()),
+        ];
+        (transactions, bank, poh_recorder, entry_receiver)
+    }
+
+    #[test]
```
ryoqun, line 2250: how about writing very small unit test for `push_unprocessed` regarding `batch_limit`?

ryoqun: also, this would improve coverage...

image

ryoqun (dismissed):
LGTM with nits

Nice clean up, offering good opportunity to familiarize banking stage. :)

core/src/banking_stage.rs
carllin: hmmm forgot to mention tried this and got this clippy error: ``#[warn(clippy::unit_arg)]`

core/src/banking_stage.rs
carllin: oh I meant I prefer it remains just a placeholder value, i.e., we could make it `0, 2` in the future

core/src/banking_stage.rs
carllin: Changed, same comment as: https://github.com/solana-labs/solana/pull/15210#discussion_r574875850

core/src/banking_stage.rs
carllin: added an assert

core/src/banking_stage.rs
carllin: added references to `packets_vec` to be more explicit

core/src/banking_stage.rs
```diff
@@ -2108,4 +2208,250 @@ mod tests {
...
+                continue_sender.send(()).unwrap();
+            }
+
+            t_consume.join().unwrap();
+        }
+        Blockstore::destroy(&ledger_path).unwrap();
+    }
+
+    #[test]
+    fn test_push_unprocessed_batch_limit() {
```
ryoqun, line 2392: :+1:

core/src/banking_stage.rs
```diff
@@ -2108,4 +2208,250 @@ mod tests {
...
+                .collect();
+        // Set the limit to 2
+        let batch_limit = 2;
+        // Create some new unprocessed packets
+        let new_packets = single_element_packets;
+        let packet_indexes = vec![];
+
+        let mut dropped_batches_count = 0;
+        let mut newly_buffered_packets_count = 0;
+        // Because the unprocessed `packet_indexes` is empty, this packets are not aded to
```
ryoqun, line 2407: nits: `added`.

core/src/banking_stage.rs
```diff
@@ -2108,4 +2208,250 @@ mod tests {
...
+            packet_indexes,
+            &mut dropped_batches_count,
+            &mut newly_buffered_packets_count,
+            batch_limit,
+        );
+        assert_eq!(unprocessed_packets.len(), 1);
+        assert_eq!(dropped_batches_count, 0);
+        assert_eq!(newly_buffered_packets_count, 0);
+
+        // Because the unprocessed `packet_indexes` is non-empty, the packets are aded to
```
ryoqun, line 2421: nits: `added`

ryoqun (dismissed):
lgtm again with nits! Thanks for writing a test.

One day, I hope we could apply similar cleanup to `process_packets` too. :)

core/src/banking_stage.rs
```diff
@@ -177,81 +253,89 @@ impl BankingStage {
...
+                        test_fn();
+                    }
+                    has_more_unprocessed_transactions
+                } else {
+                    rebuffered_packets_len += original_unprocessed_indexes.len();
+                    // `original_unprocessed_indexes` must have remaining packets to process
+                    // if not yet processed.
+                    assert!(Self::packet_has_more_unprocessed_transactions(
+                        &original_unprocessed_indexes
+                    ));
```
ryoqun, line 334: hehe, it looks like you went ahead by adding protective `assert!`. :)

carllin: yeah, scary 😃

PR #15216, opened 2021-02-09
solana-test-validator now uses the BPF JIT by default, `--no-bpf-jit` to disable

mvines:


sakridge:
We want it as default?

mvines:
> We want it as default?

For dev purposes, I think so.  That'll get us more mileage on the JIT.    In 1.6/1.7 timeframe when we're fully comfortable with the JIT, we could do the same for `solana-validator`

sakridge:
> > We want it as default?
> 
> For dev purposes, I think so. That'll get us more mileage on the JIT. In 1.6/1.7 timeframe when we're fully comfortable with the JIT, we could do the same for `solana-validator`

Oh, my bad. I didn't notice was for test-validator only. No problems there I think.

mvines:
mergify 💔

PR #15219, opened 2021-02-09
Use spl-token-mint secondary index for relevant getProgramAccounts requests

CriesofCarrots:
#### Problem
Users want to query for all token accounts of a particular mint, eg:
```
curl https://api.mainnet-beta.solana.com -X POST -H "Content-Type: application/json" -d '
  {
    "jsonrpc": "2.0",
    "id": 1,
    "method": "getProgramAccounts",
    "params": [
      "TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA",
      {
        "encoding":"jsonParsed",
        "filters": [
          {
            "dataSize": 165
          },
          {
            "memcmp": {
              "offset": 0,
              "bytes": "<MINT_ADDRESS>"
            }
          }
        ]
      }
    ]
  }
'
```

However, this operation is currently very expensive, as it uses a full accounts scan.

#### Summary of Changes
When the `spl-token-mint` secondary index is available, use it for these requests (parallels existing fn for token-owner)

### DIFF
--- a/core/src/rpc.rs
+++ b/core/src/rpc.rs
@@ -334,6 +334,8 @@ impl JsonRpcRequestProcessor {
         let keyed_accounts = {
             if let Some(owner) = get_spl_token_owner_filter(program_id, &filters) {
                 self.get_filtered_spl_token_accounts_by_owner(&bank, &owner, filters)
+            } else if let Some(mint) = get_spl_token_mint_filter(program_id, &filters) {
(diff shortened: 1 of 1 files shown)


carllin (approved):

mvines:
Oh this one too, good to merge @CriesofCarrots ?

PR #15246, opened 2021-02-10
More failure codepath tracing wiring

ryoqun:
related to this #15237 , another some very low-risk niceties, this is needed when I was doing last periodic inflation verification.

### DIFF
--- a/programs/stake/src/stake_state.rs
+++ b/programs/stake/src/stake_state.rs
@@ -34,6 +34,21 @@ pub enum StakeState {
 }
 
 #[derive(Debug)]
+pub enum SkippedReason {
+    ZeroPoints,
+    ZeroPointValue,
+    ZeroReward,
+    ZeroCreditsAndReturnZero,
+    ZeroCreditsAndReturnCurrent,
+}
+
+impl From<SkippedReason> for InflationPointCalculationEvent {
+    fn from(reason: SkippedReason) -> Self {
+        InflationPointCalculationEvent::Skipped(reason)
+    }
+}
+
+#[derive(Debug)]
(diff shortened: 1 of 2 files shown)


ryoqun:
@CriesofCarrots could you sanity-check this at your free time? :)

programs/stake/src/stake_state.rs
```diff
@@ -592,8 +614,14 @@ impl Stake {
         // if there is no newer credits since observed, return no point
         if new_vote_state.credits() <= self.credits_observed {
             if fix_stake_deactivate {
+                if let Some(inflation_point_calc_tracer) = inflation_point_calc_tracer {
+                    inflation_point_calc_tracer(&SkippedReason::ZeroCreditsAndReturnCurrent.into());
+                }
                 return (0, self.credits_observed);
             } else {
+                if let Some(inflation_point_calc_tracer) = inflation_point_calc_tracer {
```
CriesofCarrots, line 622: We shouldn't ever be hitting this case, right?

CriesofCarrots (approved):
lgtm!

programs/stake/src/stake_state.rs
ryoqun: Right, we shouldn't ever be hitting. :) That's one of the things I actually want to reaffirm, if there should be any oddities in the distant future. :)

ryoqun: (Also, I think we're nearing to the phase where we can remove this dead altogether safely)

ryoqun:
now that another v1.5 is released; time to trickle down this.

Issue #8484, opened 2020-02-26
Searching leader schedule for leader slots is inefficient


carllin:
#### Problem
Follow up to: https://github.com/solana-labs/solana/pull/8468

In the current implementation here: https://github.com/solana-labs/solana/pull/8468/files#diff-bf2148039ffb77b38df2bbae2f74b418R146, 

Once the first leader slot `N` is found, the last slot is at most N + GRACE_TICKS_FACTOR * MAX_GRACE_SLOTS. This means:

1) We iterate through at most N + GRACE_TICKS_FACTOR * MAX_GRACE_SLOTS
2) We only do at most GRACE_TICKS_FACTOR * MAX_GRACE_SLOTS lookups in RocksDb because a validator only does a lookup if the slot belongs to them.

What still sucks: You still have to iterate through N entries in the schedule, granted these iterations are a lot faster because they do not incur a RocksDb lookup.

#### Proposed Solution
Build a `BTreeSet` with all the leader schedule slots a validator is scheduled for, which should make "next" queries fast with the `btreeset.range` function. Purge the set when roots are set.

carllin:
Resolved by #15249

PR #15249, opened 2021-02-10
adds an inverted index to leader schedule

behzadnouri:
#### Problem
next_leader_slot is doing a linear search for slots in which a pubkey is
the leader:
https://github.com/solana-labs/solana/blob/e59a24d9f/ledger/src/leader_schedule_cache.rs#L123-L157
This can be done more efficiently by adding an inverted index to leader
schedule.

#### Summary of Changes
* added a reverse index to leader-schedule

### DIFF
--- a/ledger/src/leader_schedule.rs
+++ b/ledger/src/leader_schedule.rs
@@ -1,6 +1,9 @@
+use itertools::Itertools;
 use rand::distributions::{Distribution, WeightedIndex};
 use rand_chacha::{rand_core::SeedableRng, ChaChaRng};
 use solana_sdk::pubkey::Pubkey;
+use std::collections::HashMap;
+use std::convert::identity;
 use std::ops::Index;
 use std::sync::Arc;
 
@@ -15,6 +18,8 @@ pub struct FixedSchedule {
 #[derive(Debug, Default, PartialEq)]
 pub struct LeaderSchedule {
     slot_leaders: Vec<Pubkey>,
+    // Inverted index from pubkeys to indices where they are the leader.
+    index: HashMap<Pubkey, Arc<Vec<usize>>>,
 }
 
 impl LeaderSchedule {
@@ -34,11 +39,22 @@ impl LeaderSchedule {
                 }
             })
             .collect();
-        Self { slot_leaders }
+        Self::new_from_schedule(slot_leaders)
     }
 
     pub fn new_from_schedule(slot_leaders: Vec<Pubkey>) -> Self {
-        Self { slot_leaders }
+        let index = slot_leaders
+            .iter()
+            .enumerate()
+            .map(|(i, pk)| (*pk, i))
+            .into_group_map()
+            .into_iter()
+            .map(|(k, v)| (k, Arc::new(v)))
+            .collect();
+        Self {
+            slot_leaders,
+            index,
+        }
     }
 
     pub fn get_slot_leaders(&self) -> &[Pubkey] {
@@ -47,6 +63,28 @@ impl LeaderSchedule {
 
     pub fn num_slots(&self) -> usize {
         self.slot_leaders.len()
+    }
+
+    /// Returns indices starting from 'offset' where pubkey is the leader.
+    pub(crate) fn get_schedule(
+        &self,
+        pubkey: &Pubkey,
+        offset: usize, // Starting index.
+    ) -> impl Iterator<Item = usize> {
+        let index = self.index.get(pubkey).cloned().unwrap_or_default();
+        let num_slots = self.slot_leaders.len();
+        let size = index.len();
+        #[allow(clippy::reversed_empty_ranges)]
+        let range = if index.is_empty() {
+            1..=0 // Intentionally empty range of type RangeInclusive.
+        } else {
+            let offset = index
+                .binary_search(&(offset % num_slots))
+                .unwrap_or_else(identity)
+                + offset / num_slots * size;
+            offset..=usize::MAX
+        };
+        range.map(move |k| index[k % size] + k / size * num_slots)
     }
 }
 
@@ -61,14 +99,14 @@ impl Index<u64> for LeaderSchedule {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use rand::Rng;
+    use std::iter::repeat_with;
 
     #[test]
     fn test_leader_schedule_index() {
         let pubkey0 = solana_sdk::pubkey::new_rand();
         let pubkey1 = solana_sdk::pubkey::new_rand();
-        let leader_schedule = LeaderSchedule {
-            slot_leaders: vec![pubkey0, pubkey1],
-        };
+        let leader_schedule = LeaderSchedule::new_from_schedule(vec![pubkey0, pubkey1]);
         assert_eq!(leader_schedule[0], pubkey0);
         assert_eq!(leader_schedule[1], pubkey1);
         assert_eq!(leader_schedule[2], pubkey0);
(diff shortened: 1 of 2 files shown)


behzadnouri:
right chunk is this code:

`replay-loop-timing-stats::reset_bank_elapsed`:
index_leader_schedule replay-loop-timing-stats reset_bank_elapsed

carllin:
ha! https://github.com/solana-labs/solana/issues/8484

ledger/src/leader_schedule.rs
```diff
@@ -34,11 +39,22 @@ impl LeaderSchedule {
...
+        Self::new_from_schedule(slot_leaders)
     }
 
     pub fn new_from_schedule(slot_leaders: Vec<Pubkey>) -> Self {
-        Self { slot_leaders }
+        let index = slot_leaders
+            .iter()
+            .enumerate()
+            .map(|(i, pk)| (*pk, i))
+            .into_group_map()
```
carllin, line 50: useful utitlity!

ledger/src/leader_schedule.rs
```diff
@@ -48,6 +64,28 @@ impl LeaderSchedule {
...
+    ) -> impl Iterator<Item = usize> {
+        let index = self.index.get(pubkey).cloned().unwrap_or_default();
+        let num_slots = self.slot_leaders.len();
+        let size = index.len();
+        #[allow(clippy::reversed_empty_ranges)]
+        let range = if index.is_empty() {
+            1..=0 // Intentionally empty range of type RangeInclusive.
+        } else {
+            let offset = index
+                .binary_search(&(offset % num_slots))
```
carllin, line 82: `offset % num_slots` is for future proofing right? I think today it should always be true `offset < num_slots` since `num_slots` is always an epoch number of slots.

ledger/src/leader_schedule.rs
```diff
@@ -48,6 +64,28 @@ impl LeaderSchedule {
...
+                .unwrap_or_else(identity)
+                + offset / num_slots * size;
```
carllin, line 84: The `offset % num_slots` and `offset / num_slots` is for future proofing right? I think today it should always be true offset < num_slots since num_slots is always an epoch number of slots.

ledger/src/leader_schedule.rs
```diff
@@ -48,6 +64,28 @@ impl LeaderSchedule {
     pub fn num_slots(&self) -> usize {
         self.slot_leaders.len()
     }
+
+    /// Returns indices starting from 'offset' where pubkey is the leader.
+    pub(crate) fn get_schedule(
```
carllin, line 69: nit, how about `get_schedule -> get_leader_schedule_indices`

ledger/src/leader_schedule.rs
```diff
@@ -48,6 +64,28 @@ impl LeaderSchedule {
     pub fn num_slots(&self) -> usize {
         self.slot_leaders.len()
     }
+
+    /// Returns indices starting from 'offset' where pubkey is the leader.
```
carllin, line 68: nit: how about defining `offset`, something like

"Given an input 'offset', an index into the leader schedule, the function returns an iterator of indices  `i > offset` at which the given pubkey is the leader`

ledger/src/leader_schedule.rs
```diff
@@ -48,6 +64,28 @@ impl LeaderSchedule {
...
+    pub(crate) fn get_schedule(
+        &self,
+        pubkey: &Pubkey,
+        offset: usize, // Starting index.
+    ) -> impl Iterator<Item = usize> {
+        let index = self.index.get(pubkey).cloned().unwrap_or_default();
+        let num_slots = self.slot_leaders.len();
+        let size = index.len();
```
carllin, line 76: nit: rename `size` to `leader_schedule_length`

ledger/src/leader_schedule.rs
```diff
@@ -48,6 +64,28 @@ impl LeaderSchedule {
...
         self.slot_leaders.len()
     }
+
+    /// Returns indices starting from 'offset' where pubkey is the leader.
+    pub(crate) fn get_schedule(
+        &self,
+        pubkey: &Pubkey,
+        offset: usize, // Starting index.
+    ) -> impl Iterator<Item = usize> {
+        let index = self.index.get(pubkey).cloned().unwrap_or_default();
```
carllin, line 74: nit: rename "index" to `pubkey_to_leader_indexes`

ledger/src/leader_schedule.rs
```diff
@@ -48,6 +64,28 @@ impl LeaderSchedule {
...
+        let num_slots = self.slot_leaders.len();
```
carllin, line 75: nits: "num_slots" -> "epoch_leader_schedule_len`

ledger/src/leader_schedule.rs
```diff
@@ -48,6 +64,28 @@ impl LeaderSchedule {
...
+        let range = if index.is_empty() {
+            1..=0 // Intentionally empty range of type RangeInclusive.
+        } else {
+            let offset = index
+                .binary_search(&(offset % num_slots))
+                .unwrap_or_else(identity)
+                + offset / num_slots * size;
+            offset..=usize::MAX
+        };
+        range.map(move |k| index[k % size] + k / size * num_slots)
```
carllin, line 87: nits: Maybe add a comment for the `k / size * num_slots` because I think it's primary use is also future proofing. Maybe something like:

If we want to know the `pubkey`'s `k`th leader slot, and `k` > `num_pubkey_owned_leader_indexes`, that means we've  already gone through `k/num_pubkey_owned_leader_indexes` iterations of the entire leader schedule. 
This means `pubkey_owned_leader_schedule_indexes[0]` is an offset from the `k/num_pubkey_owned_leader_indexes` *  total_leader_schedule_len` slots that have already passed since the last time a complete round of the leader schedule was finished.

ledger/src/leader_schedule.rs
behzadnouri: Expanded the comment.

ledger/src/leader_schedule.rs
behzadnouri: Renamed it to `get_indices`. The object itself is called `leader_schedule`, so i think `leader_schedule.get_leader_schedule_indices` has unnecessary repetition.

ledger/src/leader_schedule.rs
behzadnouri: Leaning on brevity here, as this is only a local variable in a short function, and I feel like longer names here cause more line breaks and make the code harder to read.

ledger/src/leader_schedule.rs
behzadnouri: I feel like `num_slots` is more consistent with existing code elsewhere in this struct:
https://github.com/solana-labs/solana/blob/ab0f4c69a/ledger/src/leader_schedule.rs#L48-L50

ledger/src/leader_schedule.rs
behzadnouri: again leaning on brevity here, but thanks for suggestion.

ledger/src/leader_schedule.rs
behzadnouri: This is actually for  _backward_ compatibility. Current code already cycles the leader schedule:
https://github.com/solana-labs/solana/blob/ab0f4c69a/ledger/src/leader_schedule.rs#L53-L59
so this has to replicate same logic.

ledger/src/leader_schedule.rs
behzadnouri: Added more comments, but again this is just replicating existing logic:
https://github.com/solana-labs/solana/blob/ab0f4c69a/ledger/src/leader_schedule.rs#L53-L59

ledger/src/leader_schedule_cache.rs
```diff
@@ -120,49 +119,40 @@ impl LeaderScheduleCache {
...
-            }
-            start_index = 0;
-        }
-        first_slot.map(|slot| (slot, last_slot))
+            });
+        let first_slot = schedule.next()?;
+        let max_slot = first_slot.saturating_add(max_slot_range);
+        let last_slot = schedule
+            .take_while(|slot| *slot < max_slot)
+            .zip(first_slot + 1..)
```
carllin, line 150: nit: Hehe, this was already an ambiguity from before this PR, but how about changing  the comment for this function to /// Return the (next slot, last slot) consecutive range of slots...

carllin (approved):

ledger/src/leader_schedule_cache.rs
behzadnouri: Updated the comment.

PR #15250, opened 2021-02-10
assert 2 items in stores don't have same slot and version with different values

jeffwashington:
#### Problem
This is an assumption we're making in the code. There are hash mismatches we're tracking. This check has no measurable cost.

#### Summary of Changes
assert the assumption.

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -4097,6 +4097,13 @@ impl AccountsDB {
                         i = k;
                         look_for_first_key = false;
                         continue 'outer;
+                    } else {
+                        let prev = &slice[k - 1];
+                        assert!(
+                            !(prev.slot == now.slot
+                                && prev.version == now.version
+                                && (prev.hash != now.hash || prev.lamports != now.lamports))
+                        );
                     }
                 }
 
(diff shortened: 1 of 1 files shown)


carllin (dismissed):

runtime/src/accounts_db.rs
```diff
@@ -3987,6 +3987,13 @@ impl AccountsDB {
                         i = k;
                         look_for_first_key = false;
                         continue 'outer;
+                    } else {
+                        let prev = &slice[k - 1];
+                        assert!(
+                            !(prev.slot == now.slot
+                                && prev.version == now.version
+                                && (prev.hash != now.hash || prev.lamports != now.lamports))
```
sakridge, line 3995: Would any debug message here be useful? cc @carllin

jeffwashington: Relevant information would seem to be: pubkey, slot, 2 hashes, and 2 lamports. I can add these. @sakridge

sakridge: Ok. I think it might be interesting to have if this slot has had a shrink or clean operation on it. That information doesn't exist yet though.

We can add that basic information like pubkey/slot/hash/lamports, but it may not really tell us much about why the write version is conflicting. Likely there's nothing about the specific account hash or lamports value itself that caused the write version confliction. It's about it's structure inside the stores files and potentially how it was moved around.

sakridge (approved):

PR #15267, opened 2021-02-11
create function div_ceil

jeffwashington:
#### Problem
We already use div_ceil and there will likely be a few more places soon.

#### Summary of Changes
implement div_ceil as a function

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -3391,6 +3391,14 @@ impl AccountsDB {
         })
     }
 
+    fn div_ceil(x: usize, y: usize) -> usize {
+        let mut result = x / y;
+        if x % y != 0 {
+            result += 1;
+        }
+        result
+    }
+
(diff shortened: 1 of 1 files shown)


runtime/Cargo.toml
```diff
@@ -26,6 +26,7 @@ libc = "0.2.81"
 libloading = "0.6.2"
 log = "0.4.11"
 memmap2 = "0.1.0"
+num = "0.3.1"
```
sakridge, line 29: Can it just be the `num-integer` crate?

sakridge:
I would just write it as a function actually locally in one of our crates. What is the advantage of the external crate? It just adds to build times, adds another vector for inserting malicious code. You added the tests here already anyway.

runtime/Cargo.toml
jeffwashington: stale

PR #15269, opened 2021-02-11
combine flatten and hash

jeffwashington:
#### Problem
Scanning accounts and accumulating hash and lamports is expensive. Flattening large vectors of data (from individual elements or from sub-vectors) is also expensive.

#### Summary of Changes
Allow hash calculation to operate on vectors of smaller size, resulting in lower time and memory use.
Later, we can do partial hashing of results to further reduce the data that has to be retained in memory during a scan.

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -176,7 +176,6 @@ pub struct HashStats {
     pub hash_time_total_us: u64,
     pub sort_time_total_us: u64,
     pub flatten_time_total_us: u64,
-    pub flatten_after_zeros_time_total_us: u64,
     pub pre_scan_flatten_time_total_us: u64,
     pub hash_total: usize,
     pub unreduced_entries: usize,
@@ -189,7 +188,6 @@ impl HashStats {
             + self.hash_time_total_us
             + self.sort_time_total_us
             + self.flatten_time_total_us
-            + self.flatten_after_zeros_time_total_us
             + self.pre_scan_flatten_time_total_us;
         datapoint_info!(
             "calculate_accounts_hash_without_index",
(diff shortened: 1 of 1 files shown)


jeffwashington:
This pr is related to #15275. After that pr is ready, we will be able to reduce some code paths and remove some redundancy. There are 2 separate ideas, so I broke it into 2 PRs.

runtime/src/accounts_db.rs
```diff
@@ -3813,14 +3953,31 @@ impl AccountsDB {
...
+    fn flatten_hashes_and_hash(
+        hashes: Vec<Vec<Hash>>,
+        fanout: usize,
+        test: bool,
+    ) -> (Hash, Measure, usize) {
         let mut flat2_time = Measure::start("flat2");
-        let hashes: Vec<Hash> = hashes.into_iter().flatten().collect();
+
+        let offsets = CumulativeOffsets1D::from_raw(&hashes);
+        let hash_bkup = if test {
```
sakridge, line 3964: I would have this path just in the test code itself. We should be able to call both paths and compare in a test.

runtime/src/accounts_db.rs
```diff
@@ -190,6 +190,64 @@ impl CalculateHashIntermediate {
...
+
+impl CumulativeOffsets1D {
+    pub fn from_raw<T>(raw: &[Vec<T>]) -> CumulativeOffsets1D {
+        let mut total_count: usize = 0;
+        let mut cumulative_offsets = Vec::new();
+        raw.iter().enumerate().for_each(|(i, v)| {
+            let len = v.len();
+            cumulative_offsets.push(CumulativeOffset1D::new(i, total_count, total_count + len));
+            total_count += len;
+        });
```
sakridge, line 226: seems like a natural collect
`let cumulative_offsets = raw.iter().enumerate().map(|(i, v)| ...).collect();`

runtime/src/accounts_db.rs
jeffwashington: it was already also being tested in tests

runtime/src/accounts_db.rs
jeffwashington: yes, and it would be good to remove zero length vectors here, so did that, too

sakridge (dismissed):

PR #15273, opened 2021-02-11
Bump Rust-BPF to v1.46

jackcmay:
#### Problem

The Solana Rust toolchain that supports BPF currently supports Rust v1.39

#### Summary of Changes

Bump the supported Rust version to v1.46

Fixes #


dmakarov (approved):

sdk/bpf/scripts/install.sh
```diff
@@ -178,7 +178,9 @@ if [[ ! -e rust-bpf-$machine-$version.md || ! -e rust-bpf-$machine ]]; then
 fi
 
 # Install Rust-BPF Sysroot sources
-version=v0.14
+rm -rf rust-bpf-sysroot*
+rm -rf ~/.cache/solana/v1.0
```
dmakarov, line 182: it seems this causes the sysroot to be cloned and rebuilt on the build of every program in `programs/rust/*`, which dramatically slows down testing.

jackcmay:
Yes, this is temporary and will not be in the final changes

dmakarov (dismissed):

sdk/bpf/scripts/install.sh
```diff
@@ -128,57 +129,36 @@ if [[ ! -e criterion-$version.md || ! -e criterion ]]; then
...
-  if [[ $exitcode -ne 0 ]]; then
-    exit 1
-  fi
-  touch llvm-native-$version.md
-fi
-
 # Install Rust-BPF
-version=v0.2.5
-if [[ ! -e rust-bpf-$machine-$version.md || ! -e rust-bpf-$machine ]]; then
+version=v1.0
```
dmakarov, line 133: why not use the same versioning as rust, i.e. v1.46?

sdk/bpf/scripts/install.sh
```diff
@@ -128,57 +129,36 @@ if [[ ! -e criterion-$version.md || ! -e criterion ]]; then
...
   if [[ $exitcode -ne 0 ]]; then
     exit 1
   fi
-  touch rust-bpf-$machine-$version.md
+  touch bpf-tools-$version.md
 fi
 
 # Install Rust-BPF Sysroot sources
-version=v0.14
+version=v1.0
```
dmakarov, line 161: btw, will this not conflict with v1.0 of rust-bpf-osx in .cache/ dir?

jackcmay: nope

sdk/bpf/scripts/install.sh
jackcmay: because we might make changes to the toolchain on our end that might necessitate v1.x,

dmakarov: 1.46.x then maybe?

jackcmay: We could, thought that also might conflict with a rust version

dmakarov (approved):

PR #15275, opened 2021-02-11
remove hash accumulation with lamports

jeffwashington:
#### Problem
Several parts of the merkle pipeline have been reworked previously. With the change in usage patterns, we can remove some code and improve both performance and simplicity as well as prepare for future changes that are coming to streamline this expensive operation further.

#### Summary of Changes
Remove some old code and rework.

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -3381,6 +3381,10 @@ impl AccountsDB {
         Self::compute_merkle_root_and_capitalization_loop(hashes, fanout, |t| (t.1, t.2))
     }
 
+    pub fn compute_merkle_root(hashes: Vec<(Pubkey, Hash)>, fanout: usize) -> Hash {
+        Self::compute_merkle_root_and_capitalization_loop(hashes, fanout, |t| (t.1, 0)).0
+    }
+
     // this function avoids an infinite recursion compiler error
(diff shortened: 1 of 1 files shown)


runtime/src/accounts_db.rs
```diff
@@ -4020,16 +4005,12 @@ impl AccountsDB {
...
-        let scan_result: ScanStorageResult<(Pubkey, Hash, u64), DashMapVersionHash> = self
+        let scan_result: ScanStorageResult<(Pubkey, Hash), DashMapVersionHash> = self
             .scan_account_storage(
                 slot,
                 |loaded_account: LoadedAccount| {
                     // Cache only has one version per key, don't need to worry about versioning
-                    Some((
-                        *loaded_account.pubkey(),
-                        *loaded_account.loaded_hash(),
-                        CACHE_VIRTUAL_WRITE_VERSION,
```
jeffwashington, line 4031: @carllin I think this was lamports. Can you please help me verify removing this field is the correct thing to do if we're not using lamports anyway?

carllin: looks great to me!. The lamports are always set to 0 anyways, and the call to `accumulate_account_hashes()` only returns the computed hash anyways. I've been meaning to clean this up at some point, thanks for taking it!

sakridge (approved):
lgtm

PR #15278, opened 2021-02-11
add another hash check

jeffwashington:
#### Problem
hash calculation from stores is not matching capitalization on the bank

#### Summary of Changes
with hash debug flag on, calculate hash and cap prior to process_stale_slot_with_budget and prior to force_flush_accounts_cache. This gives us 1 more data point as to when data may have changed.

Fixes #

### DIFF
--- a/runtime/src/accounts_background_service.rs
+++ b/runtime/src/accounts_background_service.rs
@@ -12,7 +12,7 @@ use crossbeam_channel::{Receiver, SendError, Sender};
 use log::*;
 use rand::{thread_rng, Rng};
 use solana_measure::measure::Measure;
-use solana_sdk::clock::Slot;
+use solana_sdk::{clock::Slot, hash::Hash};
 use std::{
     boxed::Box,
     fmt::{Debug, Formatter},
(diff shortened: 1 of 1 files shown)


jeffwashington:
Note this only runs when the validator test hash flag is passed.

carllin (approved):

runtime/src/accounts_background_service.rs
```diff
@@ -91,6 +91,13 @@ impl SnapshotRequestHandler {
                     status_cache_slot_deltas,
                 } = snapshot_request;
 
+                let previous_hash = if test_hash_calculation {
+                    // cannot calculate or verify here because cache has not been flushed and stores don't match
```
sakridge, line 95: I don't get the comment.. does it just apply to the non-index case?

jeffwashington: yes, it just applies to the non-index case. I can clarify the comment.

jeffwashington: updated the comment

sakridge (approved):

PR #15290, opened 2021-02-12
sdk: Add Borsh support for types and utilities

joncinque:
#### Problem

We want to support Borsh more widely for program development, but the core SDK types do not derive Borsh traits.

#### Summary of Changes

Derive those traits!  Also, this adds some utilities for working with Borsh in programs and BanksClient.  Also, add generic ProgramError / InstructionErrors for dev ease of use.

The IOError might be overkill, but it's what Borsh returns on serde, which allows us to write `Type::try_from_slice(&input)?` in a program processor without needing to constantly map the error.

https://github.com/solana-labs/solana-program-library/pull/1226 showcases how these are used

Fixes #


joncinque:
ABI update:
```
$ diff -u bank__BankSlotDelta_frozen_abi__test_abi_digest_GESbL4xC5ndicqqZeCW4fEHtNR71NMESVWM3i2Ty2MY9 bank__BankSlotDelta_frozen_abi__test_abi_digest_MUmkgPsCRrWL2HEsMEvpkWMis35kbBnaEZtrph5P6bk
--- bank__BankSlotDelta_frozen_abi__test_abi_digest_GESbL4xC5ndicqqZeCW4fEHtNR71NMESVWM3i2Ty2MY9        2021-02-12 17:13:02.708084045 +0100
+++ bank__BankSlotDelta_frozen_abi__test_abi_digest_MUmkgPsCRrWL2HEsMEvpkWMis35kbBnaEZtrph5P6bk 2021-02-12 17:22:31.761284740 +0100
@@ -140,7 +140,7 @@
                                                         field u8
                                                             primitive u8
                                                         field solana_program::instruction::InstructionError
-                                                            enum InstructionError (variants = 46)
+                                                            enum InstructionError (variants = 44)
                                                                 variant(0) GenericError (unit)
                                                                 variant(1) InvalidArgument (unit)
                                                                 variant(2) InvalidInstructionData (unit)
@@ -186,9 +186,6 @@
                                                                 variant(41) ProgramFailedToCompile (unit)
                                                                 variant(42) Immutable (unit)
                                                                 variant(43) IncorrectAuthority (unit)
-                                                                variant(44) IOError(alloc::string::String) (newtype)
-                                                                    primitive str
-                                                                variant(45) AccountNotRentExempt (unit)
                                                     variant(9) CallChainTooDeep (unit)
                                                     variant(10) MissingSignatureForFee (unit)
                                                     variant(11) InvalidAccountIndex (unit)
```

This is my first time with the ABI update process -- is it actually recommended to create a new struct as recommended in the docs https://docs.solana.com/implemented-proposals/abi-management#developers-workflow, or can this work for two new enum variants?

banks-client/src/lib.rs
```diff
@@ -218,6 +217,33 @@ impl BanksClient {
...
+            let account =
+                result?.ok_or_else(|| io::Error::new(io::ErrorKind::Other, "Account not found"))?;
+            T::unpack_from_slice(&account.data)
+                .map_err(|_| io::Error::new(io::ErrorKind::Other, "Failed to deserialize account"))
+        })
+    }
+
+    /// Return the unpacked account data at the given address
+    /// If the account is not found, an error is returned
+    pub fn get_account_data<T: BorshDeserialize>(
```
mvines, line 236: I think this function should be flagged as borsh specific, perhaps:
```suggestion
    pub fn get_account_data_with_borsh<T: BorshDeserialize>(
```

sdk/program/src/borsh_utils.rs
```diff
@@ -0,0 +1,55 @@
+//! Borsh utils
```
mvines, line 1: nit: let's just call this mod `borsh` instead of `borsh_utils`.  The utils suffix is superfluous and makes you have to remember if it's "utils" or "util"

sdk/program/src/instruction.rs
```diff
@@ -207,6 +214,19 @@ impl Instruction {
             accounts,
         }
     }
+
+    pub fn new_from_borsh<T: BorshSerialize>(
```
mvines, line 218:
```suggestion
    pub fn new_with_borsh<T: BorshSerialize>(
```

mvines (approved):
> This is my first time with the ABI update process -- is it actually recommended to create a new struct as recommended in the docs https://docs.solana.com/implemented-proposals/abi-management#developers-workflow, or can this work for two new enum variants?

Adding new variants to the end of an existing enum is allowed, so no issues here that I can see.

Thanks for doing this @joncinque!   A backport to v1.5 here would be nice too, so we don't have to wait for v1.6 to use it in the SPL repo

sdk/program/src/instruction.rs
```diff
@@ -207,6 +214,19 @@ impl Instruction {
             accounts,
         }
     }
+
+    pub fn new_with_borsh<T: BorshSerialize>(
```
jackcmay, line 218: Might be helpful to deprecate `Instruction::new` since it has the hidden bincode serialization and instead add `new_with_vec` and `new_with_bincode`?

joncinque: Finally the right terms! Is `new_with_vec` meant for just `Vec<u8>` input?

jackcmay: Yeah, `new_with_vec` takes a `Vec<u8>` on the assumption that it is in its final form.  Possibly already serialized by some other developer-defined mechanism.

PR #15299, opened 2021-02-12
fix: default preflightCommitment to Connection.commitment when sending tx

CriesofCarrots:
#### Problem
If a users instantiates a `Connection` object with a commitment, this commitment is used by default for all rpc queries and transaction confirmations. However, it is *not* used as the preflightCommitment for transaction sends, which defaults to `finalized` (max) on the rpc side. As such, if a client is using a commitment level less than `finalized`, they must specify preflightCommitment for all sends or preflight simulations may fail due to unfinalized state.

#### Summary of Changes
Default preflightCommitment to Connection.commitment if not specified
Test throws error without code change

### DIFF
--- a/web3.js/test/connection.test.js
+++ b/web3.js/test/connection.test.js
@@ -1361,6 +1361,25 @@ describe('Connection', () => {
         ).to.be.rejected;
       });
     });
+
+    it('consistent preflightCommitment', async () => {
+      const connection = new Connection(url, 'singleGossip');
+      const sender = new Account();
+      const recipient = new Account();
+      let signature = await connection.requestAirdrop(
+        sender.publicKey,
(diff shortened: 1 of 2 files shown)


web3.js/test/connection.test.js
```diff
@@ -1361,6 +1361,25 @@ describe('Connection', () => {
         ).to.be.rejected;
       });
     });
+
+    it.only('consistent preflightCommitment', async () => {
```
jstarry, line 1365: drop the only :)

CriesofCarrots: 🤦‍♀️ drat

jstarry (approved):

PR #15300, opened 2021-02-12
adds an upper bound on cluster-slots size

behzadnouri:
#### Problem
https://github.com/solana-labs/solana/issues/14366#issuecomment-769096305

#### Summary of Changes
If the cluster-slots map becomes too large, discard slots furthest away from the root.

### DIFF
--- a/core/src/cluster_slots.rs
+++ b/core/src/cluster_slots.rs
@@ -9,6 +9,10 @@ use std::{
     sync::{Arc, RwLock},
 };
 
+// Limit the size of cluster-slots map in case
+// of receiving bogus epoch slots values.
+const CLUSTER_SLOTS_TRIM_SIZE: usize = 524_288; // 512K
+
 pub type SlotPubkeys = HashMap<Pubkey, u64>;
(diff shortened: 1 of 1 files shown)


core/src/cluster_slots.rs
```diff
@@ -43,6 +47,12 @@ impl ClusterSlots {
         {
             let mut cluster_slots = self.cluster_slots.write().unwrap();
             *cluster_slots = cluster_slots.split_off(&(root + 1));
+            // Trimming is done at 2x size so that amortized it has a constant
+            // cost. The slots furthest away from the root are discarded.
+            if cluster_slots.len() > 2 * CLUSTER_SLOTS_TRIM_SIZE {
```
carllin, line 52: This is a good start!

At 8 bytes a pubkey, 1000 validators, this can still get rather large though. Would be interesting to figure out how this cross cluster pollution is happening 🤔

carllin (approved):

sakridge (approved):

PR #15303, opened 2021-02-12
divide account scan into bins for more efficient downstream processing

jeffwashington:
#### Problem
With 1B accounts, there is too much data to require contiguous vector allocations holding all data.

#### Summary of Changes
We can take advantage of the uniform distribution of pubkey values and divide the account scan into bins by pubkey values. This allows us to break the scan, sort, hashing, and other processing into smaller chunks and more quickly reduce sections of data from larger raw data to hashes and lamport sums. More changes are coming. This change introduces the concept of intermediate results binned by pubkey value.

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -281,6 +281,35 @@ impl CumulativeOffsets {
         }
     }
 
+    pub fn from_raw_2d<T>(raw: &[Vec<Vec<T>>]) -> CumulativeOffsets {
+        let mut total_count: usize = 0;
+        let cumulative_offsets: Vec<CumulativeOffset> = raw
+            .iter()
+            .enumerate()
+            .map(|(i, v)| {
+                v.iter()
+                    .enumerate()
+                    .filter_map(|(j, v)| {
+                        let len = v.len();
+                        if len > 0 {
+                            let result = CumulativeOffset::new(vec![i, j], total_count);
+                            total_count += len;
+                            Some(result)
+                        } else {
+                            None
+                        }
+                    })
+                    .collect::<Vec<_>>()
+            })
+            .flatten()
+            .collect();
+
+        Self {
+            cumulative_offsets,
+            total_count,
+        }
+    }
+
     // return the biggest slice possible that starts at 'start'
     pub fn get_slice<'a, T>(&self, raw: &'a [Vec<T>], start: usize) -> &'a [T] {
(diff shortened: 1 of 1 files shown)


runtime/src/accounts_db.rs
```diff
@@ -281,14 +281,62 @@ impl CumulativeOffsets1D {
...
+            .iter()
+            .enumerate()
+            .map(|(i, v)| {
+                v.iter()
+                    .enumerate()
+                    .filter_map(|(j, v)| {
+                        let len = v.len();
+                        if len > 0 {
+                            let result = CumulativeOffset::new(vec![i, j], total_count);
+                            total_count += len;
```
sakridge, line 296: With the 2d version, this might be cleaner with the iterative version because it ends up with less code and I believe the inner `collect` allocates which is just then copied into the outer `collect` vec, so it should be slightly more efficient without that copy.

```
let cumulative_offsets = Vec::with_capacity(raw.len() * raw[0].len());
for (i, v) in raw.iter().enumerate() {
   for (j, v) in v.iter().enumerate() {
       let len = v.len();
       if len > 0 {
           cumulative_offsets.push(CumulativeOffset::new(vec![i, j], total_count));
           total_count += len;
       }
   }
}
cumulative_offsets 
```

sakridge (approved):
lgtm with just the one suggestion. Seems like a pretty straightforward translation to 2d. What is the performance difference?

runtime/src/accounts_db.rs
jeffwashington: will address separately

jeffwashington: #15346

jeffwashington:
> lgtm with just the one suggestion. Seems like a pretty straightforward translation to 2d. What is the performance difference?

on colo machine, with 15M accounts, static data from scan captured last week.
flatten, sort, eliminate zeros, hash.
master: 1.3s
this pr: 0.8s.
It is notable that hash time got a little worse temporarily until we chunk larger amounts (than 16) of hashes together. That's in another pr. 
The real performance aim of this series of changes is to allow [store scan to final hash and cap] to occur with 1B accounts in minimal time with manageable dynamic memory usage. The next pr will be moving binning to scanning so that we only keep the data for a range of pubkeys at a time, process them to a small number of merkle hashes already a few levels deep and a lamport sum and then move to the next range of pubkeys.

PR #15320, opened 2021-02-14
Add limit and shrink policy for recycler

carllin:
#### Problem
Recycler can:

1) Grow without bound
2) Once it grows, it never shrinks

#### Summary of Changes
1) Set optional upper bound for recycler
2) Check recycler periodically, shrink if we are below the shrink threshold `N` times consecutively

Fixes #


perf/src/recycler.rs
```diff
@@ -73,41 +135,108 @@ impl<T: Default + Reset + Sized> Recycler<T> {
...
+            } else if let Some(limit) = object_pool.limit {
+                let should_allocate_new = object_pool.total_allocated_count < limit;
+                if should_allocate_new {
+                    object_pool.total_allocated_count += 1;
+                }
+                (None, false, should_allocate_new)
+            } else {
+                (None, false, true)
+            }
+        };
```
carllin, line 215: @ryoqun this elastic shrinking seems to be working well for my test case where I spam the node, I can see the total number of objects rising then falling over time after the spam is over. 

What do you think?

ryoqun: @carllin cool. this looks promising by superseding my attempt #15139. After all #15139 wasn't prepared for spamming, it's half way in that it only addresses spontaneous spikes, not spamming... ;)

perf/src/recycler.rs
```diff
@@ -73,41 +181,124 @@ impl<T: Default + Reset + Sized> Recycler<T> {
...
+            let mut t = T::default();
+            t.set_recycler(Arc::downgrade(&self.recycler));
+            Some(t)
+        } else {
+            None
+        }
+    }
+
+    fn get_shrink_target(shrink_ratio: f64, current_size: u32) -> u32 {
+        (shrink_ratio * current_size as f64).ceil() as u32
```
sakridge, line 293: We don't need float for this.

ryoqun:
(i'll review this today.)

perf/src/recycler.rs
```diff
@@ -137,26 +328,129 @@ impl<T: Default + Reset> RecyclerX<T> {
...
+        // actually recycles
+        let recycler = PacketsRecycler::new_with_limit("", limit);
+        let mut allocated_items = vec![];
+        for i in 0..limit * 2 {
+            let x = recycler.allocate();
+            if i >= limit {
+                assert!(x.is_none());
+            } else {
+                allocated_items.push(x.unwrap());
+            }
```
ryoqun, line 371: maybe invert the `then` clause and `else` clause? I usually want to put something happen first in the `then` clause.

ryoqun:
could you mix my perf test into this pr? https://github.com/solana-labs/solana/pull/15139/files#r578352709

ryoqun:
> Total allocated count

@carllin Oh, are you wondering excessive allocation? I know gossip doesn't recycle.

Here's my patch in the baking queue. Could you also incorporate this?

https://github.com/ryoqun/solana/tree/FixRecyclerGossip

ryoqun:
Also, I think we should test `cat /dev/zero > /dev/udp/localhost/<all_ports>`. :)

carllin:
heh well clearly there's a bug here because I underflowed in testing against mainnet: 
Screen Shot 2021-02-18 at 5 04 19 AM

perf/src/recycler.rs
```diff
@@ -11,36 +18,128 @@ struct RecyclerStats {
...
+}
+
+#[derive(Clone)]
+pub struct Recycler<T: Reset> {
     recycler: Arc<RecyclerX<T>>,
+    shrink_metric_name: &'static str,
+}
+
+impl<T: Default + Reset> Recycler<T> {
+    pub fn new(shrink_metric_name: &'static str) -> Self {
```
ryoqun, line 50: nits: `new_without_limit`? This looks like it's pretty dangerous to instantiate this if confronted to untrusted data flow. So indicate the name as so. :)

perf/src/recycler.rs
```diff
@@ -73,41 +181,132 @@ impl<T: Default + Reset + Sized> Recycler<T> {
...
+    pub fn allocate(&self) -> Option<T> {
+        let mut shrink_removed_objects = vec![];
+        let (mut allocated_object, did_reuse, should_allocate_new, mut shrink_stats) = {
+            let mut object_pool = self
+                .recycler
+                .gc
+                .lock()
+                .expect("recycler lock in pb fn allocate");
+
+            let now = timestamp();
```
ryoqun, line 193: I think it's generally more suited to use `Instant` just directly rather than `timestamp()` (which is backed by `SystemTime`). Because of monotonically non-decreasing nature of `Instant` over `SystemTime`.

perf/src/recycler.rs
```diff
@@ -73,41 +181,132 @@ impl<T: Default + Reset + Sized> Recycler<T> {
...
+            let mut shrink_stats = None;
+
+            if now.saturating_sub(object_pool.last_shrink_check_ts)
+                > object_pool.check_shrink_interval_ms as u64
```
ryoqun, line 197: By using Instant, you can write like this: https://github.com/solana-labs/solana/pull/15139/files#r578474676

perf/src/recycler.rs
```diff
@@ -73,41 +181,132 @@ impl<T: Default + Reset + Sized> Recycler<T> {
...
+                    let target_size =
+                        std::cmp::max(object_pool.minimum_object_count, shrink_threshold_count);
+                    let ideal_num_to_remove = object_pool.total_allocated_count - target_size;
+                    for _ in 0..ideal_num_to_remove {
+                        if let Some(mut expired_object) = object_pool.object_pool.pop() {
+                            expired_object.unset_recycler();
+                            // Drop these outside of the lock because the Drop() implmentation for
+                            // certain objects like PinnedVec's can be expensive
+                            shrink_removed_objects.push(expired_object);
+                            // May not be able to shrink exactly `ideal_num_to_remove` objects sinc
```
ryoqun, line 229: nits: `since`.

perf/src/recycler.rs
```diff
@@ -73,41 +181,132 @@ impl<T: Default + Reset + Sized> Recycler<T> {
...
+                        shrink_elapsed: recycler_shrink_elapsed.as_us(),
+                        // Filled in later
+                        drop_elapsed: 0,
+                    })
+                }
+            }
+
+            let reused_object = object_pool.object_pool.pop();
+            if reused_object.is_some() {
+                (reused_object, true, false, shrink_stats)
```
ryoqun, line 254: nits: I don't think this is really simpler writing, arguably. However I can't resist to propose this: 

```rust
if let Some(reused_object) = object_pool.object_pool.pop() {
  (Some(reused_object), true, false, shrink_stats)
} ...
```

perf/src/recycler.rs
```diff
@@ -73,41 +181,132 @@ impl<T: Default + Reset + Sized> Recycler<T> {
...
-            "allocating new: total {} {:?} id: {} reuse: {} max_gc: {}",
-            total,
-            name,
-            self.recycler.id,
-            self.recycler.stats.reuse.load(Ordering::Relaxed),
-            self.recycler.stats.max_gc.load(Ordering::Relaxed),
-        );
+    pub fn allocate(&self) -> Option<T> {
+        let mut shrink_removed_objects = vec![];
+        let (mut allocated_object, did_reuse, should_allocate_new, mut shrink_stats) = {
```
ryoqun, line 186: rather long chunk of code.. how about extracting to `fn`?

perf/src/recycler.rs
```diff
@@ -73,41 +181,132 @@ impl<T: Default + Reset + Sized> Recycler<T> {
...
+                            ("name", self.shrink_metric_name, String),
+                            ("count", object_pool.total_allocated_count as i64, i64),
+                            ("len", object_pool.len() as i64, i64),
+                        )
+                    }
+                    object_pool.total_allocated_count += 1;
+                }
+                (None, false, should_allocate_new, shrink_stats)
+            } else {
+                (None, false, true, shrink_stats)
```
ryoqun, line 270: these option, bool, bool are hurting my brain a bit...

how about this?:

```
enum AllocationStrategy {
  Reuse(T),
  Allocate,
  DontAllocate,
}
```

perf/src/recycler.rs
```diff
@@ -11,36 +18,128 @@ struct RecyclerStats {
...
+        Self {
+            recycler: Arc::new(RecyclerX::new(Some(limit))),
+            shrink_metric_name,
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct ObjectPool<T: Reset> {
+    object_pool: Vec<T>,
```
ryoqun, line 67: nits: `object_pool.object_pool` in the other place of this code looks a bit uncool. How about renaming this to `objects`?

btw, I don't like `RecyclerX` as well, unrelated to your pr...

perf/src/recycler.rs
carllin: fixed in separate `make_allocation_decision()` function~

perf/src/recycler.rs
carllin: Ah yup, was meaning to refactor this mess, thanks!

carllin:
> could you mix my perf test into this pr? https://github.com/solana-labs/solana/pull/15139/files#r578352709

@ryoqun, done! The results

Before this PR:
test bench_recycler ... bench:          65 ns/iter (+/- 6)

With this PR, not entering the shrink check: 
test bench_recycler ... bench:          76 ns/iter (+/- 6)

With this PR, when entering the shrink check here: https://github.com/solana-labs/solana/pull/15320/files#diff-798271cb43dca47253c06d8ab6888d8bf4fb385fc8e74cc2cd70ac1b3b664ebfR132. Seems like the creation of the `Instant` object is the expensive part of that operation.
test bench_recycler ... bench:         111 ns/iter (+/- 9)

carllin:
> > Total allocated count
> 
> @carllin Oh, are you wondering excessive allocation? I know gossip doesn't recycle.
> 
> Here's my patch in the baking queue. Could you also incorporate this?
> 
> https://github.com/ryoqun/solana/tree/FixRecyclerGossip

@ryoqun These look good, but would rather incorporate them into a follow-up PR as this one is getting rather large

perf/src/recycler.rs
```diff
@@ -11,36 +27,221 @@ struct RecyclerStats {
...
+        let is_consistent = self.total_allocated_count as usize >= self.len();
+        if !is_consistent {
+            warn!(
+                "Object pool inconsistent: {} {} {}",
+                self.total_allocated_count,
+                self.len(),
+                recycler_name
+            );
+        }
+        assert!(is_consistent);
```
ryoqun, line 131: nits: unify into a just `assert!` with the warn! message?

ryoqun (approved):
LGTM with nit

Thanks for proposing better fix than mine. :)

PR #15338, opened 2021-02-15
rename CumulativeOffset in preparation of 2d support

jeffwashington:
#### Problem
CumulativeOffset was unfortunately named too specific to 1d. It will soon be used for 1d or 2d.
This is in its own pr to make the next change less messy.

#### Summary of Changes
Rename.

Fixes #


sakridge (dismissed):
lgtm

PR #15344, opened 2021-02-15
compute merkle root on chunks of fanout^3

jeffwashington:
#### Problem
We need to hash 1B hash values in a merkle tree of 16.
Level 2 would have an array of 1B/16 = 62M hashes for us to hash next.
Level 3 would have 3.9M
Level 4 would have 244k

#### Summary of Changes
If we are hashing at least 4096 (16^3) hashes, then reduce 4096 hashes in a chunk to 1 hash by doing 3 levels of merkle tree reduction in 1 pass.
This means on the second round, we'd only have an array of 244k hash values left to hash.

For parallelism, we only enable this if the total we're hashing is at least 16^4. Otherwise, we would not be paralleizing merkle tree hashing for trees in the 4k-65k range.

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -3630,6 +3630,7 @@ impl AccountsDB {
     fn compute_merkle_root_from_slices<'a, F>(
         total_hashes: usize,
         fanout: usize,
+        max_levels_per_pass: Option<usize>,
         get_hashes: F,
     ) -> Hash
     where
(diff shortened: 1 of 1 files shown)


runtime/src/accounts_db.rs
```diff
@@ -3650,24 +3661,54 @@ impl AccountsDB {
...
+                        data_index += 1;
+                    }
+                } else {
+                    // hash 3 levels of fanout simultaneously.
+                    // The result of this loop is a single hash value from fanout^3 input hashes.
+                    let mut i = start_index;
+                    while i < end_index {
+                        let mut hasher_j = Hasher::default();
+                        for _j in 0..fanout {
+                            let mut hasher_k = Hasher::default();
```
sakridge, line 3692: ok, we have tests that cover all the edge-cases here?

jeffwashington: The case of having a partial last block is specifically tested with and without the 3 level optimization and compared.
I just made the test below more thorough to saturate the entire partial chunk solution space for a fanout of 3.
https://github.com/solana-labs/solana/blob/4ab1678b5b6c7859ec05ced8bc796588aec1a7a3/runtime/src/accounts_db.rs#L6544
In general the several merkle variations are tried with various calls and the results are compared with each other.
Note that the merkle code paths are about to be simplified in an upcoming PR as it will now be efficient for both uses - whether inputs are [Hash] or [(Pubkey, Hash)] where either of these types of input could be represented in 1-d to 3-d input vectors.

sakridge: OK. thanks

sakridge (approved):

PR #15347, opened 2021-02-16
cli: Speed up program deploys

jstarry:
#### Problem
The CLI fetches signature statuses with history instead of using the recent signature cache when confirming deploy transactions. Searching the history is a lot slower than checking the cache and often times out when checked a large batch of signatures.

#### Summary of Changes
- Search for deploy transactions in signature cache

Fixes #


jstarry:
This makes a huge difference in deploys, like 20x faster, no joke

CriesofCarrots (approved):
Yep, now that we've resolved a lot of the issues that caused transactions to pass out of the recent signature cache before this step, this looks great. Thanks!

PR #15350, opened 2021-02-16
add validator flag no-accounts-db-index-hashing

jeffwashington:
#### Problem
We want to calculate hashes with or without using the index.
#### Summary of Changes
add
--no-accounts-db-index-hashing
to validator to disable using the index. The hash will be calculated based only on the stores.

Fixes #


sakridge (approved):
lgtm

PR #15352, opened 2021-02-16
checks that prune-messages have the same inner/outer pubkey

behzadnouri:
#### Problem
The two pubkeys should be the same, and one is redundant.

#### Summary of Changes
In `Sanitize` implementation of `Protocol`, check that the two are the same.


sakridge (approved):
lgtm

PR #15353, opened 2021-02-16
Move hash code

jeffwashington:
#### Problem
hashing code doesn't depend on accounts_db

#### Summary of Changes
move hashing code out of accounts_db.rs into its own file, accounts_hash.rs

Fixes #


sakridge (approved):
lgtm

PR #15371, opened 2021-02-17
Update lockfile

leoluk:
Building v1.5.8 with `cargo build --release --locked` fails:

```
error: the lock file /build/solana/Cargo.lock needs to be updated but --locked was passed to prevent this
If you want to try to generate the lock file without accessing the network, use the --offline flag.
```

Ran ´cargo generate-lockfile` to update the lockfile.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,43 +12,43 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
-version = "0.12.1"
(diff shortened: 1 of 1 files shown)


mvines:
I think this actually bumps versions on some packages, such as fs_extras (the reason why the build failed)

leoluk:
¯\_(ツ)_/¯

PR #15377, opened 2021-02-17
multi-pass bin scanning

jeffwashington:
Problem
1B accounts requires more efficient use of memory in calculating hash for a slot.

Summary of Changes
calculate the hashes in passes, reducing memory footprint

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -3775,7 +3775,10 @@ impl AccountsDb {
             let num_scan_passes: usize = 2;
 
             let bins_per_pass = PUBKEY_BINS_FOR_CALCULATING_HASHES / num_scan_passes;
-            assert_eq!(bins_per_pass * num_scan_passes, PUBKEY_BINS_FOR_CALCULATING_HASHES); // evenly divisible
+            assert_eq!(
+                bins_per_pass * num_scan_passes,
+                PUBKEY_BINS_FOR_CALCULATING_HASHES
+            ); // evenly divisible
             let mut previous_pass = PreviousPass::default();
             let mut final_result = (Hash::default(), 0);
 
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -5282,7 +5282,6 @@ pub mod tests {
         let bins = 2;
         let result = AccountsDb::scan_snapshot_stores(
             &storages,
-            true,
             &mut stats,
             bins,
             &Range {
@@ -5298,7 +5297,6 @@ pub mod tests {
         // just the second bin of 2
         let result = AccountsDb::scan_snapshot_stores(
             &storages,
-            true,
             &mut stats,
             bins,
             &Range {
@@ -5317,7 +5315,6 @@ pub mod tests {
         for bin in 0..bins {
             let result = AccountsDb::scan_snapshot_stores(
                 &storages,
-                true,
                 &mut stats,
                 bins,
                 &Range {
@@ -5335,7 +5332,6 @@ pub mod tests {
         for bin in 0..bins {
             let result = AccountsDb::scan_snapshot_stores(
                 &storages,
-                true,
                 &mut stats,
                 bins,
                 &Range {
@@ -5367,7 +5363,6 @@ pub mod tests {
         let mut stats = HashStats::default();
         let result = AccountsDb::scan_snapshot_stores(
             &storages,
-            true,
             &mut stats,
             bins,
             &Range {
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -20,7 +20,7 @@
 
 use crate::{
     accounts_cache::{AccountsCache, CachedAccount, SlotCache},
-    accounts_hash::{AccountsHash, CalculateHashIntermediate, HashStats},
+    accounts_hash::{AccountsHash, CalculateHashIntermediate, HashStats, PreviousPass},
     accounts_index::{
         AccountIndex, AccountsIndex, AccountsIndexRootsStats, Ancestors, IndexKey, IsCached,
         SlotList, SlotSlice, ZeroLamport,
@@ -54,7 +54,7 @@ use std::{
     collections::{hash_map::Entry, BTreeMap, BTreeSet, HashMap, HashSet},
     convert::TryFrom,
     io::{Error as IoError, Result as IoResult},
-    ops::RangeBounds,
+    ops::{Range, RangeBounds},
     path::{Path, PathBuf},
     sync::atomic::{AtomicBool, AtomicU64, AtomicUsize, Ordering},
     sync::{Arc, Mutex, MutexGuard, RwLock},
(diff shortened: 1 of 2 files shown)


runtime/src/accounts_db.rs
```diff
@@ -3728,6 +3730,12 @@ impl AccountsDb {
             |loaded_account: LoadedAccount,
              accum: &mut Vec<Vec<CalculateHashIntermediate>>,
              slot: Slot| {
+                let pubkey = *loaded_account.pubkey();
+                let rng_index = pubkey.as_ref()[0] as usize * bins / max_plus_1;
```
carllin, line 3734: nit: how about renaming this to `rng_bin_index`, to signal we are mapping the pubkey to a `bin` index in the range `0...bins`

runtime/src/accounts_db.rs
```diff
@@ -3773,15 +3779,44 @@ impl AccountsDb {
...
-                &mut stats,
-                PUBKEY_BINS_FOR_CALCULATING_HASHES,
-            );
 
-            AccountsHash::rest_of_hash_calculation(result, &mut stats)
+            // # of passes should be a function of the total # of accounts that are active.
+            // higher passes = slower total time, lower dynamic memory usage
+            // lower passes = faster total time, higher dynamic memory usage
+            // passes=2 cuts dynamic memory usage in approximately half.
+            let passes: usize = 2;
```
carllin, line 3789: Maybe this can be passed in as a validator option so we can dynamically adjust it?

runtime/src/accounts_hash.rs
```diff
@@ -600,25 +629,111 @@ impl AccountsHash {
     pub fn rest_of_hash_calculation(
         data_sections_by_pubkey: Vec<Vec<Vec<CalculateHashIntermediate>>>,
         mut stats: &mut HashStats,
-    ) -> (Hash, u64) {
+        last_pass: bool,
```
carllin, line 632: nit: `is_last_pass` to signal boolean

runtime/src/accounts_db.rs
```diff
@@ -3773,15 +3779,44 @@ impl AccountsDb {
...
+                    simple_capitalization_enabled,
+                    &mut stats,
+                    PUBKEY_BINS_FOR_CALCULATING_HASHES,
+                    &bounds,
+                );
+
+                let (hash, lamports, for_next_pass) = AccountsHash::rest_of_hash_calculation(
+                    result,
+                    &mut stats,
+                    pass == passes - 1,
```
carllin, line 3813: nit: raname passes to num_passes` so that this reads `pass == num_passes - 1`

runtime/src/accounts_hash.rs
```diff
@@ -600,25 +629,111 @@ impl AccountsHash {
...
+            // Save hashes that don't evenly hash. They will be combined with hashes from the next pass.
+            let left_over_hashes = hash_total % target_fanout;
+
+            // move tail hashes that don't evenly hash into a 1d vector for next time
+            let mut i = hash_total - left_over_hashes;
+            while i < hash_total {
+                let data = cumulative.get_slice_2d(&hashes, i);
+                next_pass.remaining_unhashed.extend(data);
+                i += data.len();
+            }
```
carllin, line 672: Can we make this a `get_range()` function on `CumulativeOffsets`

runtime/src/accounts_db.rs
carllin: Edit: Never mind, I see that adjusting this `passes` value will result in a different hash, so in order for this be consistent it seems we'll need to make a new release to update this value anyways.

runtime/src/accounts_hash.rs
```diff
@@ -266,12 +274,13 @@ impl AccountsHash {
         fanout: usize,
         max_levels_per_pass: Option<usize>,
         get_hashes: F,
-    ) -> Hash
+        specific_level_count: Option<usize>,
```
carllin, line 277: nits: rename `get_hashes` to `get_hash_slice_containing_index`

runtime/src/accounts_hash.rs
```diff
@@ -600,25 +629,111 @@ impl AccountsHash {
...
 
-        let (hashes, total_lamports) =
+        let (mut hashes, mut total_lamports) =
             Self::de_dup_and_eliminate_zeros(sorted_data_by_pubkey, &mut stats);
 
-        let hash = Self::flatten_hashes_and_hash(hashes, MERKLE_FANOUT, &mut stats);
+        total_lamports += previous_state.lamports;
+
+        if !previous_state.remaining_unhashed.is_empty() {
+            // these items were not hashes last iteration because they didn't divide evenly
```
carllin, line 645: nit: hashes -> hashed

runtime/src/accounts_hash.rs
```diff
@@ -282,9 +291,18 @@ impl AccountsHash {
...
+        if three_level {
+            if let Some(specific_level_count_value) = specific_level_count {
+                three_level = specific_level_count_value >= THREE_LEVEL_OPTIMIZATION;
+            }
+        }
+        let (num_hashes_per_chunk, levels_hashed) = if three_level {
+            (target, THREE_LEVEL_OPTIMIZATION)
+        } else {
+            (fanout, 1)
+        };
```
carllin, line 305: Can we extract this into a separate function that returns whether to `use_three_levels`

carllin: Also could we add a few diagrams for this block of logic here: https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_hash.rs#L322-L344 clarifying the structure of the logic? For me specifically, I think it's a bit hard tracking what is being iterated on over the abstracted structure underlying the generic `get_hashes()` function:

For me, a diagram of
1) The structure of what is being indexed by `i`, "data_index" "end_index", "end",
2) What are the boundaries of the values being returned by `get_hashes` on every iteration
3) The structure of the "3 levels of fanout". This seems to imply a merkle three with three layers of children and parents but I don't think that's what's being done here? 

would be super helpful!

runtime/src/accounts_hash.rs
jeffwashington: I don't understand what you're asking. Can you please clarify your request?

runtime/src/accounts_db.rs
jeffwashington: The hash will be the same, regardless of the # of passes we use. This should be dynamically adjustable somehow. My thinking was this isn't urgent yet, but is valuable to have ready as we scale to more accounts. Perhaps we want it calculated somehow based on the current # of accounts.

jeffwashington: But, it could be passed in through the validator in a subsequent pr.

runtime/src/accounts_hash.rs
```diff
@@ -600,19 +692,106 @@ impl AccountsHash {
     pub fn rest_of_hash_calculation(
```
sakridge, line 692: comment above this has a typo: "items which fit in the containing bin"

sakridge (approved):
lgtm

jeffwashington:
updated metrics:
metrics on colo machine against mainnet.
Summary:
master: 2.30s
this pr with 2 passes: 2.76s (and half the mem usage)
(and it looks like I have some metric calculating bugs that need to be adjusted for this pr [hash time, unreduced entries])

master: datapoint: calculate_accounts_hash_without_index accounts_scan=705076i eliminate_zeros=113706i hash=75872i sort=481642i hash_total=18390432i flatten=924291i unreduced_entries=18842660i num_snapshot_storage=387263i pre_scan_flatten=1560i total=2302147i

2 pass scan: datapoint: calculate_accounts_hash_without_index accounts_scan=1053479i eliminate_zeros=135394i hash=76938i sort=577485i hash_total=18457173i flatten=911668i unreduced_entries=18867428i num_snapshot_storage=387616i pre_scan_flatten=5354i total=2760318i

jeffwashington:
downstream projects passed in this equivalent pr:
https://buildkite.com/solana-labs/solana/builds/42444

PR #15391, opened 2021-02-18
Cleanup old features

jackcmay:
#### Summary of Changes

Fixes #


jackcmay:
@mvines Cleaning up features older than 2 epochs, any objections?

sakridge:
We won’t be able to replay old ledger

mvines:
> We won’t be able to replay old ledger

We generally have this problem anyway.   When we archive ledger slices after a new epoch, we also record the solana version that was used so you can find the compatible release if needed

> Cleaning up features older than 2 epochs, any objections?

Nope

jackcmay:
> We won’t be able to replay old ledger

@sakridge You good?

PR #15407, opened 2021-02-18
adds metrics for inbound/outbound gossip packets counts

behzadnouri:
#### Problem
Need to monitor packets distribution. 

#### Summary of Changes
* Added metrics for inbound/outbound gossip packets counts

### DIFF
--- a/core/src/cluster_info.rs
+++ b/core/src/cluster_info.rs
@@ -257,6 +257,17 @@ struct GossipStats {
     handle_batch_pull_requests_time: Counter,
     handle_batch_pull_responses_time: Counter,
     handle_batch_push_messages_time: Counter,
+    packets_received_count: Counter,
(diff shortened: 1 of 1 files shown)


carllin:
Cool, hopefully this provides some good insight 😃 

Maybe a metric for pull responses tracking how much of the pull response is data we've already received might be useful as well?

carllin (approved):

behzadnouri:
> Maybe a metric for pull responses tracking how much of the pull response is data we've already received might be useful as well?

I think these metrics already track that:
https://github.com/solana-labs/solana/blob/4648439f5/core/src/crds_gossip_pull.rs#L164-L169
https://github.com/solana-labs/solana/blob/4648439f5/core/src/cluster_info.rs#L2780-L2799

PR #15410, opened 2021-02-18
Refactoring: Move KeyedAccounts to InvokeContext

Lichtso:
## Problem
`keyed_accounts` are currently passed as a parameter everywhere, but they need to be managed in one place so that their allocation scheme can be changed.

## Summary of Changes
Collects all parametric occurrences and the construction of `keyed_accounts` and puts them into `InvokeContext`.

### `next_keyed_account()` vs. `keyed_account_at_index()`
The replacement of `next_keyed_account` by `keyed_account_at_index` is necessary in order to drop references of `keyed_accounts` and retrieve them after function calls which involve passing `invoke_context`. The reference inside the iterator of `keyed_accounts` would lead to a collision of borrows otherwise. Ideas of alternative approaches are welcome!

#### Pro
Indices are better because they:
- Can be explicated later (e.g. as enums or consts).
- Make it possible to move parts of the code around more freely as it would not depend on the order of `.next()`.
- Enable leaving gaps in case a specific slot is not needed any longer.

#### Contra
Quoting @garious:
> In early versions of this code, we indexed the accounts as you're proposing now. We generally viewed that code as unreadable (so directly opposed to your "indices are better because..." arguments). As I recall, the indexing also caused more fights with the borrow checker and so we'd end up passing the whole accounts list into low-level functions when they only required references to one or two accounts.
> The first step to improving readability was to replace all the hardcoded indexes with constants. After that, we did a big refactor to replace all the indexing with the parsing style you see today (next_keyed_account()), at which point we were finally able to pass individual accounts to lower-level functions.

#### Neutral
Quoting @jackcmay:
> @garious I could go either way on the iter vs index debate. Both have access functions that provide sufficient protection. Also, our instruction definitions specify the accounts numerically, so it kinda lends itself to index-based accessing as well. We might also want to provide some kind of (maybe generated) accessor functions along with the instruction definitions to index a specific account of an instruction. Doing this by index also eliminates the need to convert to an iter to grab some accounts, and then back to a slice to pass off to other instructions. One place where it isn't as nice is in instructions where there is a group of accounts that will be iterated through but the length may not be known, like list of multisig acounts.

### The tale of MessageProcessor::process_cross_program_instruction
The main problem is getting the `KeyedAccount`s into the `InvokeContext` through the [stack push](https://github.com/solana-labs/solana/blob/bb30c47a936d6f2f427a2588aa2a766f9e9527d6/runtime/src/message_processor.rs#L893). This is hard as there are two references inside the [`KeyedAccount` struct](https://github.com/solana-labs/solana/blob/6d5c6c17c57d1e0a5afd2cc010f575b2056a7b88/sdk/src/keyed_account.rs#L17):
* `key: &'a Pubkey`
* `account: &'a RefCell<AccountSharedData>`

Both give the `KeyedAccount` struct the lifetime `'a`. Now, in oder to push the new `KeyedAccount`s into the `InvokeContext` we need to prove to the borrow checker that the new `KeyedAccount`s will live as long as `InvokeContext` does. We can't just recycle the existing `KeyedAccount`s, as they also have two other fields `is_signer` and `is_writable` which can be different at every invocation level.

#### Lifetime of KeyedAccounts
Because the `InvokeContext` is a [trait](https://github.com/solana-labs/solana/blob/ad9901d7c67aa29a8569e6caf879c74409a0345c/sdk/src/process_instruction.rs#L27) we can not alias the lifetime of its [implementation](https://github.com/solana-labs/solana/blob/bb30c47a936d6f2f427a2588aa2a766f9e9527d6/runtime/src/message_processor.rs#L266) with the [`push` method](https://github.com/solana-labs/solana/blob/bb30c47a936d6f2f427a2588aa2a766f9e9527d6/runtime/src/message_processor.rs#L315) directly. Instead the constructor of the `KeyedAccount`s has to be called from inside the specific implementations of the trait so that the two lifetimes correctly alias.

#### Lifetime of Account keys
The `key`s can not be borrowed from [`Vec<PreAccount>`](https://github.com/solana-labs/solana/blob/bb30c47a936d6f2f427a2588aa2a766f9e9527d6/runtime/src/message_processor.rs#L254) as the `Vec` is owned by `self` inside of `ThisInvokeContext`. Instead they must be borrowed from `message.account_keys` like it is done in [create_keyed_accounts](https://github.com/solana-labs/solana/blob/bb30c47a936d6f2f427a2588aa2a766f9e9527d6/runtime/src/message_processor.rs#L581).

#### Lifetime of Account RefCells
While the `key`s of the `KeyedAccount`s can be borrowed from the existing ones, the `Account`s are created on the Rust stack at these three locations:
* [BPF syscall](https://github.com/solana-labs/solana/blob/ad7f8e7f238ca98557bc57321590f870d46de287/programs/bpf_loader/src/syscalls.rs#L1674)
* [message_processor](https://github.com/solana-labs/solana/blob/92f4018b07524ef6a0c44492309bde04205a4b5a/runtime/src/message_processor.rs#L790)
* [program-test](https://github.com/solana-labs/solana/blob/dee655df3586946f419a7bf396ac6bd6f68391df/program-test/src/lib.rs#L288)

So we have to [`unsafe { transmute }` the lifetime](https://github.com/solana-labs/solana/blob/788ff6f486082460723f471e592dfc1d37b7cd16/runtime/src/message_processor.rs#L344) to get them into the `InvokeContext`. This is possible as we know that the stack frame will be [popped](https://github.com/solana-labs/solana/blob/788ff6f486082460723f471e592dfc1d37b7cd16/runtime/src/message_processor.rs#L978) again before the [original reference](https://github.com/solana-labs/solana/blob/788ff6f486082460723f471e592dfc1d37b7cd16/runtime/src/message_processor.rs#L935) goes out of scope.

Otherwise, if we instead correctly bind the existing `Account`s to the new `KeyedAccount`s [verify_and_update](https://github.com/solana-labs/solana/blob/92f4018b07524ef6a0c44492309bde04205a4b5a/runtime/src/message_processor.rs#L874) will sound alarm as the inner instruction supposedly changed the `Account`s of the outer one.

## Future Work (for subsequent PRs)
* Map `Account`s into the BPF VM using [`MemoryRegion`s](https://github.com/solana-labs/rbpf/blob/14f27ad0dde469d2d8897c3d7b93d307c704b77c/src/memory_region.rs#L12)
* Apply the `KeyedAccount`s `is_writable` flag to make some readonly
* Remove the `PreAccount`s from the `InvokeContext` as the `Account`s are readonly anyways
* Remove the `unsafe { transmute }` which was introduced in this PR

### DIFF
--- a/programs/bpf_loader/src/lib.rs
+++ b/programs/bpf_loader/src/lib.rs
@@ -33,7 +33,7 @@ use solana_sdk::{
     feature_set::{skip_ro_deserialization, upgradeable_close_instruction},
     ic_logger_msg, ic_msg,
     instruction::InstructionError,
-    keyed_account::{from_keyed_account, next_keyed_account, KeyedAccount},
+    keyed_account::{from_keyed_account, keyed_account_at_index, KeyedAccount},
     loader_instruction::LoaderInstruction,
     loader_upgradeable_instruction::UpgradeableLoaderInstruction,
     process_instruction::{stable_log, ComputeMeter, Executor, InvokeContext},
@@ -198,8 +198,7 @@ fn process_instruction_common(
 ) -> Result<(), InstructionError> {
     let logger = invoke_context.get_logger();
 
-    let account_iter = &mut keyed_accounts.iter();
-    let first_account = next_keyed_account(account_iter)?;
+    let first_account = keyed_account_at_index(keyed_accounts, 0)?;
     if first_account.executable()? {
         if first_account.unsigned_key() != program_id {
             ic_logger_msg!(logger, "Program id mismatch");
@@ -212,7 +211,7 @@ fn process_instruction_common(
                     programdata_address,
                 } = first_account.state()?
                 {
-                    let programdata = next_keyed_account(account_iter)?;
+                    let programdata = keyed_account_at_index(keyed_accounts, 1)?;
                     if programdata_address != *programdata.unsigned_key() {
                         ic_logger_msg!(
                             logger,
@@ -292,26 +291,25 @@ fn process_loader_upgradeable_instruction(
     use_jit: bool,
 ) -> Result<(), InstructionError> {
     let logger = invoke_context.get_logger();
-    let account_iter = &mut keyed_accounts.iter();
 
     match limited_deserialize(instruction_data)? {
         UpgradeableLoaderInstruction::InitializeBuffer => {
-            let buffer = next_keyed_account(account_iter)?;
+            let buffer = keyed_account_at_index(keyed_accounts, 0)?;
 
             if UpgradeableLoaderState::Uninitialized != buffer.state()? {
                 ic_logger_msg!(logger, "Buffer account already initialized");
                 return Err(InstructionError::AccountAlreadyInitialized);
             }
 
-            let authority = next_keyed_account(account_iter)?;
+            let authority = keyed_account_at_index(keyed_accounts, 1)?;
 
             buffer.set_state(&UpgradeableLoaderState::Buffer {
                 authority_address: Some(*authority.unsigned_key()),
             })?;
         }
         UpgradeableLoaderInstruction::Write { offset, bytes } => {
-            let buffer = next_keyed_account(account_iter)?;
-            let authority = next_keyed_account(account_iter)?;
+            let buffer = keyed_account_at_index(keyed_accounts, 0)?;
+            let authority = keyed_account_at_index(keyed_accounts, 1)?;
 
             if let UpgradeableLoaderState::Buffer { authority_address } = buffer.state()? {
                 if authority_address.is_none() {
@@ -338,14 +336,14 @@ fn process_loader_upgradeable_instruction(
             )?;
         }
         UpgradeableLoaderInstruction::DeployWithMaxDataLen { max_data_len } => {
-            let payer = next_keyed_account(account_iter)?;
-            let programdata = next_keyed_account(account_iter)?;
-            let program = next_keyed_account(account_iter)?;
-            let buffer = next_keyed_account(account_iter)?;
-            let rent = from_keyed_account::<Rent>(next_keyed_account(account_iter)?)?;
-            let clock = from_keyed_account::<Clock>(next_keyed_account(account_iter)?)?;
-            let system = next_keyed_account(account_iter)?;
-            let authority = next_keyed_account(account_iter)?;
+            let payer = keyed_account_at_index(keyed_accounts, 0)?;
+            let programdata = keyed_account_at_index(keyed_accounts, 1)?;
+            let program = keyed_account_at_index(keyed_accounts, 2)?;
+            let buffer = keyed_account_at_index(keyed_accounts, 3)?;
+            let rent = from_keyed_account::<Rent>(keyed_account_at_index(keyed_accounts, 4)?)?;
+            let clock = from_keyed_account::<Clock>(keyed_account_at_index(keyed_accounts, 5)?)?;
+            let system = keyed_account_at_index(keyed_accounts, 6)?;
+            let authority = keyed_account_at_index(keyed_accounts, 7)?;
             let upgrade_authority_address = Some(*authority.unsigned_key());
             let upgrade_authority_signer = authority.signer_key().is_none();
 
@@ -452,13 +450,13 @@ fn process_loader_upgradeable_instruction(
             ic_logger_msg!(logger, "Deployed program {:?}", program.unsigned_key());
         }
         UpgradeableLoaderInstruction::Upgrade => {
-            let programdata = next_keyed_account(account_iter)?;
-            let program = next_keyed_account(account_iter)?;
-            let buffer = next_keyed_account(account_iter)?;
-            let spill = next_keyed_account(account_iter)?;
-            let rent = from_keyed_account::<Rent>(next_keyed_account(account_iter)?)?;
-            let clock = from_keyed_account::<Clock>(next_keyed_account(account_iter)?)?;
-            let authority = next_keyed_account(account_iter)?;
+            let programdata = keyed_account_at_index(keyed_accounts, 0)?;
+            let program = keyed_account_at_index(keyed_accounts, 1)?;
+            let buffer = keyed_account_at_index(keyed_accounts, 2)?;
+            let spill = keyed_account_at_index(keyed_accounts, 3)?;
+            let rent = from_keyed_account::<Rent>(keyed_account_at_index(keyed_accounts, 4)?)?;
+            let clock = from_keyed_account::<Clock>(keyed_account_at_index(keyed_accounts, 5)?)?;
+            let authority = keyed_account_at_index(keyed_accounts, 6)?;
 
             // Verify Program account
 
@@ -581,9 +579,9 @@ fn process_loader_upgradeable_instruction(
             ic_logger_msg!(logger, "Upgraded program {:?}", program.unsigned_key());
         }
         UpgradeableLoaderInstruction::SetAuthority => {
-            let account = next_keyed_account(account_iter)?;
-            let present_authority = next_keyed_account(account_iter)?;
-            let new_authority = next_keyed_account(account_iter)
+            let account = keyed_account_at_index(keyed_accounts, 0)?;
+            let present_authority = keyed_account_at_index(keyed_accounts, 1)?;
+            let new_authority = keyed_account_at_index(keyed_accounts, 2)
                 .ok()
                 .map(|account| account.unsigned_key());
 
@@ -642,9 +640,9 @@ fn process_loader_upgradeable_instruction(
             if !invoke_context.is_feature_active(&upgradeable_close_instruction::id()) {
                 return Err(InstructionError::InvalidInstructionData);
             }
-            let close_account = next_keyed_account(account_iter)?;
-            let recipient_account = next_keyed_account(account_iter)?;
-            let authority = next_keyed_account(account_iter)?;
+            let close_account = keyed_account_at_index(keyed_accounts, 0)?;
+            let recipient_account = keyed_account_at_index(keyed_accounts, 1)?;
+            let authority = keyed_account_at_index(keyed_accounts, 2)?;
 
             if close_account.unsigned_key() == recipient_account.unsigned_key() {
                 ic_logger_msg!(logger, "Recipient is the same as the account being closed");
@@ -689,9 +687,7 @@ fn process_loader_instruction(
     invoke_context: &mut dyn InvokeContext,
     use_jit: bool,
 ) -> Result<(), InstructionError> {
-    let account_iter = &mut keyed_accounts.iter();
-
-    let program = next_keyed_account(account_iter)?;
+    let program = keyed_account_at_index(keyed_accounts, 0)?;
     if program.owner()? != *program_id {
         ic_msg!(
             invoke_context,
@@ -781,10 +777,8 @@ impl Executor for BpfExecutor {
         let logger = invoke_context.get_logger();
         let invoke_depth = invoke_context.invoke_depth();
 
-        let mut keyed_accounts_iter = keyed_accounts.iter();
-        let _ = next_keyed_account(&mut keyed_accounts_iter)?;
-        let parameter_accounts = keyed_accounts_iter.as_slice();
         let mut serialize_time = Measure::start("serialize");
+        let parameter_accounts = &keyed_accounts[1..];
         let mut parameter_bytes =
             serialize_parameters(loader_id, program_id, parameter_accounts, &instruction_data)?;
         serialize_time.stop();
--- a/programs/bpf_loader/src/lib.rs
+++ b/programs/bpf_loader/src/lib.rs
@@ -69,8 +69,7 @@ fn map_ebpf_error(
     InstructionError::InvalidAccountData
 }
 
-pub fn create_and_cache_executor(
-    program_id: Option<&Pubkey>,
+pub fn create_executor(
     program_account_index: usize,
     program_data_offset: usize,
     invoke_context: &mut dyn InvokeContext,
@@ -107,18 +106,9 @@ pub fn create_and_cache_executor(
             return Err(InstructionError::ProgramFailedToCompile);
         }
     }
-    let executor = Arc::new(BpfExecutor {
+    Ok(Arc::new(BpfExecutor {
         program: executable,
-    });
-    let key = if let Some(program_id) = program_id {
-        program_id
-    } else {
-        let keyed_accounts = invoke_context.get_keyed_accounts();
-        let program = keyed_account_at_index(keyed_accounts, program_account_index)?;
-        program.unsigned_key()
-    };
-    invoke_context.add_executor(key, executor.clone());
-    Ok(executor)
+    }))
 }
 
 fn write_program_data(
@@ -261,13 +251,11 @@ fn process_instruction_common(
 
         let executor = match invoke_context.get_executor(program_id) {
             Some(executor) => executor,
-            None => create_and_cache_executor(
-                Some(program_id),
-                0,
-                program_data_offset,
-                invoke_context,
-                use_jit,
-            )?,
+            None => {
+                let executor = create_executor(0, program_data_offset, invoke_context, use_jit)?;
+                invoke_context.add_executor(program_id, executor.clone());
+                executor
+            }
         };
         executor.execute(
             loader_id,
@@ -443,13 +431,8 @@ fn process_loader_upgradeable_instruction(
             )?;
 
             // Load and verify the program bits
-            let _ = create_and_cache_executor(
-                Some(program_id),
-                3,
-                buffer_data_offset,
-                invoke_context,
-                use_jit,
-            )?;
+            let executor = create_executor(3, buffer_data_offset, invoke_context, use_jit)?;
+            invoke_context.add_executor(program_id, executor);
 
             // Update the ProgramData account and record the program bits
             programdata.set_state(&UpgradeableLoaderState::ProgramData {
@@ -570,8 +553,12 @@ fn process_loader_upgradeable_instruction(
 
             // Load and verify the program bits
 
-            let _ =
-                create_and_cache_executor(None, 2, buffer_data_offset, invoke_context, use_jit)?;
+            let executor = create_executor(2, buffer_data_offset, invoke_context, use_jit)?;
+            {
+                let keyed_accounts = invoke_context.get_keyed_accounts();
+                let program = keyed_account_at_index(keyed_accounts, 1)?;
+                invoke_context.add_executor(program.unsigned_key(), executor);
+            }
 
             // Update the ProgramData account, record the upgraded data, and zero
             // the rest
@@ -734,7 +721,12 @@ fn process_loader_instruction(
                 return Err(InstructionError::MissingRequiredSignature);
             }
 
-            let _ = create_and_cache_executor(None, 0, 0, invoke_context, use_jit)?;
+            let executor = create_executor(0, 0, invoke_context, use_jit)?;
+            {
+                let keyed_accounts = invoke_context.get_keyed_accounts();
+                let program = keyed_account_at_index(keyed_accounts, 0)?;
+                invoke_context.add_executor(program.unsigned_key(), executor);
+            }
             program.try_account_ref_mut()?.executable = true;
             ic_msg!(
                 invoke_context,
--- a/programs/bpf_loader/src/lib.rs
+++ b/programs/bpf_loader/src/lib.rs
@@ -346,7 +346,8 @@ fn process_loader_upgradeable_instruction(
             let buffer = keyed_account_at_index(keyed_accounts, 3)?;
             let rent = from_keyed_account::<Rent>(keyed_account_at_index(keyed_accounts, 4)?)?;
             let clock = from_keyed_account::<Clock>(keyed_account_at_index(keyed_accounts, 5)?)?;
-            let system = keyed_account_at_index(keyed_accounts, 6)?;
+            // TODO [KeyedAccounts to InvokeContext refactoring]
+            // let _system = keyed_account_at_index(keyed_accounts, 6)?;
             let authority = keyed_account_at_index(keyed_accounts, 7)?;
             let upgrade_authority_address = Some(*authority.unsigned_key());
             let upgrade_authority_signer = authority.signer_key().is_none();
@@ -420,7 +421,7 @@ fn process_loader_upgradeable_instruction(
                     programdata_len as u64,
                     program_id,
                 ),
-                &[payer, programdata, system],
+                &[0, 1, 6],
                 &[&[program.unsigned_key().as_ref(), &[bump_seed]]],
             )?;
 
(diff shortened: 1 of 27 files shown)


mvines:
(I'm mailed Travis CI support about the missing status here, no need to block this PR on them)

runtime/src/message_processor.rs
```diff
@@ -832,7 +843,13 @@ impl MessageProcessor {
                 Self::create_keyed_accounts(message, instruction, executable_accounts, accounts);
```
Lichtso, line 843: This mess right here will be reimplemented differently in a [subsequent PR](https://github.com/solana-labs/solana/pull/15720). So polishing it would be somewhat pointless.

runtime/src/message_processor.rs
```diff
@@ -307,7 +310,11 @@ impl<'a> ThisInvokeContext<'a> {
     }
 }
 impl<'a> InvokeContext for ThisInvokeContext<'a> {
```
Lichtso, line 312: See my remarks in `MessageProcessor::process_cross_program_instruction`

runtime/src/native_loader.rs
```diff
@@ -134,23 +134,29 @@ impl NativeLoader {
...
+            let program = keyed_account_at_index(keyed_accounts, 0)?;
+            if native_loader::id() != *program_id {
+                error!("Program id mismatch");
+                return Err(InstructionError::IncorrectProgramId);
+            }
+            if program.owner()? != *program_id {
+                error!("Executable account now owned by loader");
+                return Err(InstructionError::IncorrectProgramId);
+            }
+            // TODO: Remove these two copies (* deref is also a copy)
```
Lichtso, line 151: I really dislike introducing more copy / clone calls, so I marked this.
But, the comment in the code can probably be removed.

sdk/src/process_instruction.rs
```diff
@@ -15,20 +15,26 @@ use std::{cell::RefCell, fmt::Debug, rc::Rc, sync::Arc};
...
     instruction_data: &[u8],
     invoke_context: &dyn InvokeContext,
 ) -> Result<(), InstructionError>;
 
 pub type ProcessInstructionWithContext =
-    fn(&Pubkey, &[KeyedAccount], &[u8], &mut dyn InvokeContext) -> Result<(), InstructionError>;
+    fn(&Pubkey, &[u8], &mut dyn InvokeContext) -> Result<(), InstructionError>;
 
 /// Invocation context passed to loaders
 pub trait InvokeContext {
```
Lichtso, line 26: See my remarks in `MessageProcessor::process_cross_program_instruction`

Lichtso (commented):
There are still some TODOs left, but they can be addressed by future PRs, because the main objective of this one was achieved.

sdk/src/process_instruction.rs
```diff
@@ -41,6 +47,10 @@ pub trait InvokeContext {
     ) -> Result<(), InstructionError>;
     /// Get the program ID of the currently executing program
     fn get_caller(&self) -> Result<&Pubkey, InstructionError>;
+    /// Removes the first keyed account
+    fn pop_first_keyed_account(&mut self);
```
garious, line 51: "pop" is usually used for the constant-time operation to pop an item off the top of a stack.  Rust libs usually use "next" to remove the first item from an iterator. Should we call this "next_keyed_account"?

programs/vest/src/vest_processor.rs
```diff
@@ -97,25 +97,25 @@ pub fn process_instruction(
         VestInstruction::InitializeAccount { .. } => {}
         VestInstruction::SetTerminator(new_pubkey) => {
             verify_signed_account(
-                next_keyed_account(keyed_accounts_iter)?,
+                keyed_account_at_index(keyed_accounts, 1)?,
```
garious, line 100: This looks fragile compared to the previous code. Why not remove one account at a time like it was?

garious: Just read the PR description.  I'll like to spend some time understanding the "borrow collision" more.

runtime/src/message_processor.rs
jackcmay: Can you give more details on what you think the mess so it can be addressed in future prs?

sdk/src/process_instruction.rs
Lichtso: This actually pops the front (of a queue if you will), and it is constant time as it operates on a ref of a slice.

programs/vest/src/vest_processor.rs
Lichtso: Iterators won't do as they keep an internal reference which would collide with the one in `invoke_context`.
See my comments on that in the header / description of the PR.

sdk/src/process_instruction.rs
Lichtso: But, you are right in that it is similar to the `next` method of iterators.
However, it only advances the internal pointers and does not return anything.

garious:
Some context for my earlier comments: in early versions of this code, we indexed the accounts as you're proposing now.  We generally viewed that code as unreadable (so directly opposed to your "indices are better because..." arguments). As I recall, the indexing also caused more fights with the borrow checker and so we'd end up passing the whole accounts list into low-level functions when they only required references to one or two accounts.

The first step to improving readability was to replace all the hardcoded indexes with constants.  After that, we did a big refactor to replace all the indexing with the parsing style you see today (`next_keyed_account()`), at which point we were finally able to pass individual accounts to lower-level functions.

Lichtso:
@garious Interesting, so you are saying that it would basically be a revert?

Anyway, I don't think there is / I don't see a way to not have indexed access with the way the borrow checker works, when the `keyed_accounts` are in `invoke_context`. Note, that this does not prevent a `&KeyedAccount` from being passed into a subroutine as long as the `invoke_context` is not also passed along. In that case an index is needed there as well.

garious:
In some ways, it'd be a little worse than a revert. To get closer to a full revert, the PR would also include:
1. Implementing the `Range` op
2. Replacing the hardcoded indexes with nice readable constants

And even that would be harder to grok than the old code because you'd have an abstraction layer instead of just working with a slice of keyed accounts.

sdk/src/process_instruction.rs
garious: `remove_first`?

runtime/src/message_processor.rs
Lichtso: There is some `unsafe` code to erase life times as we can not tell rust that the `keyed_accounts` created here will exist only in this VM stack frame. But if we only recycle existing `keyed_accounts` this could be avoided, I think.

runtime/src/native_loader.rs
```diff
@@ -134,23 +134,29 @@ impl NativeLoader {
...
+            // Both could be avoided as we know that the first KeyedAccount
+            // still exists even after invoke_context.remove_first_keyed_account() is called
+            (
+                *program.unsigned_key(),
+                &program.try_account_ref()?.data.clone(),
```
jeffwashington, line 156: note that if/when data becomes an Arc, this won't be the same expensive - just different. Presumably this data is already small.

runtime/src/message_processor.rs
```diff
@@ -263,6 +263,7 @@ pub struct ThisInvokeContext<'a> {
     rent: Rent,
     pre_accounts: Vec<PreAccount>,
     account_deps: &'a [(Pubkey, RefCell<Account>)],
+    keyed_accounts: &'a [KeyedAccount<'a>],
```
jackcmay, line 266: Can we instead keep the keyed_accounts stored with the program id for each call stack level:
```rust
call_stack: Vec<(Pubkey, &'a [KeyedAccount<'a>)>,
```
And `push` and `pop` just push and pop entries and `get_keyed_accounts` returns the keyed accounts from the top?

runtime/src/message_processor.rs
```diff
@@ -316,10 +323,13 @@ impl<'a> InvokeContext for ThisInvokeContext<'a> {
             return Err(InstructionError::ReentrancyNotAllowed);
         }
         self.program_ids.push(*key);
-        Ok(())
+        let mut keyed_accounts = unsafe { std::mem::transmute(keyed_accounts) };
```
jackcmay, line 326: I'd really like to see all the unsafe code gone, we've spent a lot of care to avoid unsafe code in the runtime (besides the program's ffi boundaries).  There has to be a better way.  Maybe `Rc<RefCell>` or some other Rust mechanism that the borrow checker plays nice with.

programs/bpf_loader/src/syscalls.rs
```diff
@@ -1575,8 +1562,8 @@ fn call<'a>(
             signers_seeds_len,
             memory_mapping,
         )?;
-        let keyed_account_refs = syscall
-            .get_callers_keyed_accounts()
+        let keyed_account_refs = invoke_context
```
jackcmay, line 1565: I'm not following how we can iter here but not use `next_keyed_account` in native program's `process_instruction`, can you detail what the conflict is?

jackcmay (changes requested):
.

programs/bpf_loader/src/syscalls.rs
Lichtso: Because there are no calls which pass `invoke_context` deeper.
That would implicitly incur a reference to `keyed_accounts` which would collide with the iterator.

runtime/src/message_processor.rs
Lichtso: I tried that but it does not rid us of all the unsafe statements here,
because there still needs to be a lifetime upcast from the lifetime of the freshly created `keyed_accounts`,
which are only valid in this stack frame, to the general lifetime of the `invoke_context`.

However, if we would not create "new" `keyed_accounts` here and only reorder existing ones, then the problem should disappear entirely.

runtime/src/message_processor.rs
```diff
@@ -302,22 +303,28 @@ impl<'a> ThisInvokeContext<'a> {
...
+    fn push(
+        &mut self,
+        keyed_accounts: &[KeyedAccount],
+        key: &Pubkey,
+    ) -> Result<&[KeyedAccount], InstructionError> {
+        if self.invoke_stack.len() > self.bpf_compute_budget.max_invoke_depth {
             return Err(InstructionError::CallDepth);
         }
-        if self.program_ids.contains(key) && self.program_ids.last() != Some(key) {
+        let frame_index = self.invoke_stack.iter().position(|frame| frame.0 == *key);
```
jackcmay, line 314: Was there a corner case here that was missing before?

runtime/src/message_processor.rs
```diff
@@ -302,22 +303,28 @@ impl<'a> ThisInvokeContext<'a> {
     }
 }
 impl<'a> InvokeContext for ThisInvokeContext<'a> {
-    fn push(&mut self, key: &Pubkey) -> Result<(), InstructionError> {
-        if self.program_ids.len() > self.bpf_compute_budget.max_invoke_depth {
+    fn push(
+        &mut self,
+        keyed_accounts: &[KeyedAccount],
+        key: &Pubkey,
+    ) -> Result<&[KeyedAccount], InstructionError> {
```
jackcmay, line 310: Instead of returning the caller's keyed_accounts here it would be more flexible to add a `get_callers_keyed_accounts` accessor

runtime/src/message_processor.rs
```diff
@@ -302,22 +303,28 @@ impl<'a> ThisInvokeContext<'a> {
...
         }
-        if self.program_ids.contains(key) && self.program_ids.last() != Some(key) {
+        let frame_index = self.invoke_stack.iter().position(|frame| frame.0 == *key);
+        if frame_index != None && frame_index != Some(self.invoke_stack.len() - 1) {
             // Reentrancy not allowed unless caller is calling itself
             return Err(InstructionError::ReentrancyNotAllowed);
         }
-        self.program_ids.push(*key);
-        Ok(())
+        let keyed_accounts = unsafe { std::mem::transmute(keyed_accounts) };
```
jackcmay, line 319: Gotta get rid of this, maybe we need to rethink keyed_account, those direct references have always been a pain.  Maybe `Rc` would help here?

Lichtso: I think we can just put the `KeyedAccount`s in the `InvokeContext` directly, without a reference. As it is now the only place where it resides.

runtime/src/message_processor.rs
```diff
@@ -341,10 +348,19 @@ impl<'a> InvokeContext for ThisInvokeContext<'a> {
...
-        self.program_ids
+        self.invoke_stack
             .last()
+            .map(|frame| &frame.0)
             .ok_or(InstructionError::GenericError)
     }
+    fn remove_first_keyed_account(&mut self) {
+        let index = self.invoke_stack.len() - 1;
+        let last_frame = &mut self.invoke_stack[index];
+        last_frame.1 = &last_frame.1[1..];
```
jackcmay, line 359: Tuple references are tough on the eyes, maybe make stack a vec of structs instead?

runtime/src/message_processor.rs
```diff
@@ -341,10 +348,19 @@ impl<'a> InvokeContext for ThisInvokeContext<'a> {
...
     }
     fn get_caller(&self) -> Result<&Pubkey, InstructionError> {
-        self.program_ids
+        self.invoke_stack
             .last()
+            .map(|frame| &frame.0)
             .ok_or(InstructionError::GenericError)
     }
+    fn remove_first_keyed_account(&mut self) {
+        let index = self.invoke_stack.len() - 1;
```
jackcmay, line 357: `.last()` would be safer, if for whatever reason `.len()` was zero this would underflow

Lichtso: Yes, will have to do that anyway to get past `clippy::integer-arithmetic`

runtime/src/message_processor.rs
Lichtso: No, but it is a tuple / struct now so `contains` and `Some(key) `do not work anymore.

runtime/src/message_processor.rs
Lichtso: Sure

runtime/src/message_processor.rs
```diff
@@ -677,36 +681,43 @@ impl MessageProcessor {
     pub fn native_invoke(
         invoke_context: &mut dyn InvokeContext,
         instruction: Instruction,
-        keyed_accounts: &[&KeyedAccount],
-        signers_seeds: &[&[&[u8]]],
+        keyed_account_indices: &[usize],
```
jackcmay, line 684: If we passed `&[Rc<KeyedAccounts>]` would that allow us to avoid using indicies?

Lichtso: The problem here is not that the lifetime or ownership is unclear, but that there could be multiple references of which at least one is mutable (the one in `&mut dyn InvokeContext`). So I think a runtime borrow check like `Cell` would be more appropriate here.

Lichtso: But it is not necessary, that is why I used indices.

runtime/src/message_processor.rs
```diff
@@ -1037,27 +1054,23 @@ impl MessageProcessor {
...
         let program_id = instruction.program_id(&message.account_keys);
+        let keyed_accounts =
+            Self::create_keyed_accounts(message, instruction, executable_accounts, accounts);
         let mut invoke_context = ThisInvokeContext::new(
             program_id,
             rent_collector.rent,
             pre_accounts,
             executable_accounts,
             account_deps,
+            keyed_accounts.as_slice(),
```
jackcmay, line 1065: Could `as_slice()` be causing some of the lifetime challenges?
https://stackoverflow.com/questions/24542064/why-does-the-variable-not-live-long-enough

Lichtso: It certainly does, and as I wrote in one of the other comments, in the end I want to move the actual object in here, not just the reference.

runtime/src/message_processor.rs
jackcmay: Lets try to get rid of the indicies

sdk/src/process_instruction.rs
```diff
@@ -15,19 +15,59 @@ use std::{cell::RefCell, fmt::Debug, rc::Rc, sync::Arc};
...
+    pub keyed_accounts: Vec<KeyedAccount<'a>>,
+    pub keyed_accounts_range: std::ops::Range<usize>,
+}
+
+impl<'a> InvokeContextStackFrame<'a> {
+    pub fn new(
+        key: Pubkey,
+        keyed_accounts: &[(&'a Pubkey, bool, bool, &'a RefCell<AccountSharedData>)],
+    ) -> Self {
+        let keyed_accounts: Vec<_> = keyed_accounts
```
jackcmay, line 36: Not cloning here because we want to create new lifetimes?

sdk/src/process_instruction.rs
```diff
@@ -314,7 +373,25 @@ impl InvokeContext for MockInvokeContext {
...
+            .last()
+            .map(|frame| &frame.key)
+            .ok_or(InstructionError::GenericError)
+    }
+    fn remove_first_keyed_account(&mut self) -> Result<(), InstructionError> {
+        let stack_frame = &mut self
+            .invoke_stack
+            .last_mut()
+            .ok_or(InstructionError::GenericError)?;
+        stack_frame.keyed_accounts_range.start =
```
jackcmay, line 386: Doing this to avoid re-arranging the keyed accounts list?

runtime/src/message_processor.rs
```diff
@@ -325,24 +383,41 @@ impl<'a> InvokeContext for ThisInvokeContext<'a> {
...
+            &mut self.timings,
+            self.feature_set.is_active(&demote_sysvar_write_locks::id()),
+        )
     }
     fn get_caller(&self) -> Result<&Pubkey, InstructionError> {
-        self.program_ids
+        self.invoke_stack
+            .last()
+            .map(|frame| &frame.key)
+            .ok_or(InstructionError::GenericError)
```
jackcmay, line 406: Avoid the use of `GenericError`, add a new one if needed

programs/bpf_loader/src/lib.rs
```diff
@@ -61,70 +61,74 @@ pub enum BpfError {
...
+        if let Err(err) = executable.jit_compile() {
             ic_msg!(invoke_context, "Failed to compile program {:?}", err);
             return Err(InstructionError::ProgramFailedToCompile);
         }
     }
-    let executor = Arc::new(BpfExecutor { program });
-    invoke_context.add_executor(key, executor.clone());
-    Ok(executor)
+    Ok(Arc::new(BpfExecutor {
+        program: executable,
```
jackcmay, line 107: nit, change the BpfExecutor member name to `executable` too?

Also, would be helpful to the reviewer to move the naming changes out and mainline them in a different pr

sdk/src/process_instruction.rs
jackcmay: Never really liked that we have to create these with two different new functions.  I can see for `Accountmeta` making the distinction for developers but for `KeyedAccounts` it just looks messy.

I would recommend
- `.clone()` here unless I missing something
- replace new and new_readonly with just a new function that takes readonly as an argument

sdk/src/keyed_account.rs
```diff
@@ -186,11 +186,14 @@ where
         .collect::<A>()
 }
 
-/// Return the next KeyedAccount or a NotEnoughAccountKeys error
-pub fn next_keyed_account<'a, 'b, I: Iterator<Item = &'a KeyedAccount<'b>>>(
```
jackcmay, line 190: @garious I could go either way on the iter vs index debate.  Both have access functions that provide sufficient protection.  Also, our instruction definitions spcify the accounts numerically, so it kinda lends itself to index-based accessing as well.  We might also want to provide some kind of (maybe generated) accessor functions along with the instruction definitions to index a specific account of an instruction.  Doing this by index also eliminates the need to convert to an iter to grab some accounts, and then badk to a slice to pass off to other instructions.  One place where it isn't as nice is in instructions where there is a group of accounts that will be iterated through but the length may not be known, like list of multisig acounts.

runtime/src/message_processor.rs
```diff
@@ -297,26 +307,74 @@ impl<'a> ThisInvokeContext<'a> {
...
+                            (
+                                &self.message.account_keys[index],
+                                *is_signer,
+                                *is_writable,
+                                // TODO
+                                // Currently we are constructing new accounts on the stack
+                                // before calling MessageProcessor::process_cross_program_instruction
+                                // Ideally we would recycle the existing accounts here like this:
+                                // &self.accounts[index] as &RefCell<AccountSharedData>,
+                                unsafe { std::mem::transmute(*account) },
```
jackcmay, line 360: Lets get rid of this last unsafe.  If there are other changes that have to go in first, then lets do those.

sdk/src/keyed_account.rs
garious: I can stop blocking progress here.  My hope was that there was just some lifetime issue that needed to be solved and then the original code could be left alone.  Sounds like the situation is more nuanced.  Feel free to move forward in whatever way you feel makes the most sense.

runtime/src/message_processor.rs
Lichtso: Well, as discussed in the header of this PR, this one is really tricky. Unless I am missing something, there is no way for the `PreAccount`s `verify` to distinguish between "unlawful" changes by one of the programs and the seeming "changes" caused by correctly aliasing the accounts. The one solution I can see is to remove the need for `PreAccount`s, but that depends on this PR being done and the planned `KeyedAccounts` mapping too.

sdk/src/process_instruction.rs
Lichtso: It avoids possibly reallocating the `Vec` if we were to call `pop` or `remove` instead. And for the `KeyedAccount`s to be mapped directly it is beneficial to have them stay at one constant pointer over their lifetime.

sdk/src/process_instruction.rs
Lichtso: I originally wanted to merge the two `KeyedAccount` constructors but was worried about breaking changes in the SDK / API. But, if you also think that it would be better that way, then I will do it so.

jackcmay:
@Lichtso would it help the lifetime issue if we reworked `KeyedAccount` to remove the pubkey ref and put `Account` behind an `Rc`, thus removing the explicitly `'a` lifetime?

sdk/src/process_instruction.rs
jackcmay: Yeah, that's what I figured, be nice to clean up this later

sdk/src/process_instruction.rs
jackcmay: We could only break backward compatibility at a major version change (https://docs.solana.com/developing/backwards-compatibility).  We could create a new function that takes writable as a parameter

runtime/src/message_processor.rs
jackcmay: @Lichtso would it help the lifetime issue if we reworked KeyedAccount to remove the pubkey ref and put Account behind an Rc, thus removing the explicitly 'a lifetime?

Lichtso:
@jackcmay No, the lifetime issues of the `KeyedAccount` struct itself and the one of the `key` are both solved.
The problem stems from the `Account`s they point to, not the reference.
We create new (duplicate) `Account` structs on the stack for every invocation frame, which currently do not use / alias with the existing `Account` structs from the earlier invocation frames. And that is the core problem. If we could alias all references to one `Account` struct per `key`, we wouldn't need the `unsafe transmute`.

PR #15415, opened 2021-02-18
dump details of hash mismatches

jeffwashington:
#### Problem
There are intermittent inconsistencies when calculating hash and lamports from stores vs. using index. Turning on full checking is too expensive.

#### Summary of Changes
Turn on checking once we detect a cap mismatch. And, dump account differences between the two methods prior to panicing as before.

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -3555,14 +3555,21 @@ impl AccountsDB {
         ancestors: &Ancestors,
         simple_capitalization_enabled: bool,
     ) -> (Hash, u64) {
-        self.update_accounts_hash_with_index_option(
+        let result = self.update_accounts_hash_with_index_option(
             true,
             true,
             slot,
             ancestors,
             simple_capitalization_enabled,
             None,
-        )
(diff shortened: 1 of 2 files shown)


runtime/src/accounts_db.rs
```diff
@@ -3651,14 +3663,219 @@ impl AccountsDB {
...
+        ancestors: &Ancestors,
+        simple_capitalization_enabled: bool,
+    ) -> bool {
+        // this code path will only execute when debug_verify is true
+        // this code is cobbled together from the normal execution path
+        let storages = self.get_snapshot_storages(slot);
+        let scan_and_hash = || {
+            let mut at_least_one_mismatch = false;
+            let mut stats = HashStats::default();
+            // clipped and modified from calculate_accounts_hash_without_index
```
sakridge, line 3700: It's a lot of copied code, seems like a maintenance burden for just a debug function. Can we go the way have having two paths which compute the intermediates in both ways which is shared with the normal code and then doing like a post-processing on both of them to determine the diff?

Issue #14398, opened 2021-01-03
explorer: Detect and display if the tx has been optimistically confirmed.


aeyakovenko:
Users may want to know that the transaction has been optimistically confirmed before it’s rooted.

PR #15422, opened 2021-02-19
explorer: add confirmation status to transaction details page

oJshua:
#### Problem
It would be helpful to know if a transaction is optimistically confirmed before it's rooted.

#### Summary of Changes
Add confirmation status to the transaction details page.

Fixes https://github.com/solana-labs/solana/issues/14398


jstarry:
Sweet!

PR #15425, opened 2021-02-19
Load memo v2 into genesis for test validator

jstarry:
#### Problem
Memo 2 is not available in the solana-test-validator

#### Summary of Changes
- Load memo 2 into test validator genesis
- Update program test to use memo v2
- Add label in explorer for memo 2

Fixes #


fetch-spl.sh
```diff
@@ -40,6 +40,7 @@ fetch_program() {
 
 fetch_program token 3.1.0 TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA BPFLoader2111111111111111111111111111111111
 fetch_program memo  1.0.0 Memo1UhkJRfHyvLMcVucJwxXeuD728EqVDDwQDxFMNo BPFLoader1111111111111111111111111111111111
```
mvines, line 42:
```suggestion
```

jstarry: backwards incompatible!

jstarry: I don't mind tossing it, just didn't think it would do any harm.

mvines: actually yeah that was a bad idea.  But `program-test/src/programs.rs` should then include both memo programs too right?

jstarry: Yeah, that was my laziness. I'll add v1 / v2 modules

program-test/src/programs.rs
```diff
@@ -4,15 +4,15 @@ mod spl_token {
...
+    solana_sdk::declare_id!("MemoSq4gqABAXKb96qnH8TysNcWxMyWCqXgDLGmfcHr");
 }
 mod spl_associated_token_account {
     solana_sdk::declare_id!("ATokenGPvbdGVxr1b2hvZbsiqW5xWH25efTNsLJA8knL");
 }
 
 static SPL_PROGRAMS: &[(Pubkey, &[u8])] = &[
     (spl_token::ID, include_bytes!("programs/spl_token-3.1.0.so")),
-    (spl_memo::ID, include_bytes!("programs/spl_memo-1.0.0.so")),
+    (spl_memo::ID, include_bytes!("programs/spl_memo-3.0.0.so")),
```
mvines, line 15: The new SO needs to be added to the repo as well in here: https://github.com/solana-labs/solana/tree/master/program-test/src/programs

jstarry: Ah thanks, for some reason I thought this picked up from fetch-spl

mvines: Would be nice :)
The problem is when the program-test crate is published to crates.io, ` fetch-spl.sh` is no longer a thing

fetch-spl.sh
jstarry: To be clear, this memo is v2 despite the 3.0.0 semver right?

mvines: It's the second memo program we've published to mainnet, yes.  Using spl_memo_1_0 and spl_memo_3_0 is probably the least confusing from a Cargo.toml/program.rs perspective

jstarry: sounds good, thanks

mvines (approved):
lgtm beyond program-test/src/programs/spl_memo-3.0.0.so being empty

jstarry:
> lgtm beyond program-test/src/programs/spl_memo-3.0.0.so being empty

🤕

jstarry:
The file appears to be non-empty based on this commit diff https://github.com/solana-labs/solana/pull/15425/commits/b314b13566e30d04c7d177d44a279e14a90b7633

jstarry:
ok now that diff shows +0... it was *just* showing +73KB

jstarry:
Screen Shot 2021-02-19 at 2 26 22 PM
bizarre

jstarry:
Merging because when I view the file, it shows 73 KB.

PR #15428, opened 2021-02-19
Upgrade to Rust v1.50

mvines:


mvines:
Oof, https://rust-lang.github.io/rust-clippy/master/index.html#upper_case_acronyms is going to cause some code churn here.

PR #15437, opened 2021-02-19
prevent data copies to verify instructions don't modify data

jeffwashington:
#### Problem
When executing instructions, we make a copy of data solely for the purpose of verifying that the data wasn't changed when it wasn't supposed to change. This is very expensive.

#### Summary of Changes
After executing an instruction, prior to deserializing, compare the data we would deserialize vs the data we passed into the function. If the size or content changed, then note that on the PreAccount.

Fixes #


jeffwashington:
going a different way

PR #15439, opened 2021-02-19
Rename IOError to BorshIoError

mvines:
Clippy bucks hard against `IOError`, it wants `IoError`.  But wait, this isn't really `IoError`.  It's a `BorshIoError`.  
It's ok to rename this SDK enum entry since it was just added days ago.

Prerequisite for https://github.com/solana-labs/solana/pull/15428


joncinque:
Bah sorry about that -- how could I have caught this?

mvines:
> Bah sorry about that -- how could I have caught this?

* The new clippy in 1.50 doesn't like IOError and we have other instance of "IOError" about the tree, so hard to have caught that one.
* It seems you edited the generated file, `storage-proto/proto/solana.storage.transaction_by_addr.rs`, instead of modifying the source proto file and running the generator.    There's like zero docs about this process, and I reviewed your PR and missed it, so hard to have caught that one too.

So no need to lose any sleep over this :)

PR #15447, opened 2021-02-19
Make UiTokenAmount::ui_amount a String

CriesofCarrots:
#### Problem
UiTokenAmount::ui_amount is an f64, which has precision problems, and also does a poor job handling Mints with very large decimals.

#### Summary of Changes
Convert UiTokenAmount::ui_amount to a String

### DIFF
--- a/account-decoder/src/lib.rs
+++ b/account-decoder/src/lib.rs
@@ -24,6 +24,7 @@ use {
 };
 
 pub type StringAmount = String;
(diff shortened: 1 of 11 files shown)


CriesofCarrots:
This will break CI downstream-projects until spl-token-cli is patched. I thought I'd ignore that CI task for now?

mvines (approved):

PR #15449, opened 2021-02-20
Remove read only locks when they hit ref count zero, cleanup accounts locking

carllin:
#### Problem
1. Read only locks are never removed when they hit zero ref count
2. Accounts locking has separate locks for read only and write, even though in all cases the write accounts lock is held while reading/modifying read-only account locks. This makes it hard for readers to reason about what's happening/synchronization model between the two.

#### Summary of Changes
1. Remove zero ref count read only locks
2. Move readonly and write locks under same lock for clarity and to remove redundant locking.

Fixes #

### DIFF
--- a/runtime/src/accounts.rs
+++ b/runtime/src/accounts.rs
@@ -42,7 +42,7 @@ pub(crate) struct ReadonlyLock {
     lock_count: Mutex<u64>,
 }
 
-#[derive(Debug, Default)]
+#[derive(Debug, Default, AbiExample)]
 pub struct AccountLocks {
     write_locks: HashSet<Pubkey>,
     readonly_locks: HashMap<Pubkey, u64>,
--- a/runtime/src/accounts.rs
+++ b/runtime/src/accounts.rs
@@ -31,15 +31,58 @@ use solana_sdk::{
     transaction::{Transaction, TransactionError},
 };
 use std::{
-    collections::{HashMap, HashSet},
+    collections::{hash_map, HashMap, HashSet},
     ops::RangeBounds,
     path::PathBuf,
-    sync::{Arc, Mutex, RwLock},
+    sync::{Arc, Mutex},
 };
 
 #[derive(Default, Debug, AbiExample)]
 pub(crate) struct ReadonlyLock {
     lock_count: Mutex<u64>,
+}
+
+#[derive(Debug, Default)]
+pub struct AccountLocks {
+    write_locks: HashSet<Pubkey>,
+    readonly_locks: HashMap<Pubkey, u64>,
+}
+
+impl AccountLocks {
+    fn is_locked_readonly(&self, key: &Pubkey) -> bool {
+        self.readonly_locks
+            .get(key)
+            .map_or(false, |count| *count > 0)
+    }
+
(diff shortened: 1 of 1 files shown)


runtime/src/accounts.rs
```diff
@@ -55,10 +98,7 @@ pub struct Accounts {
     pub accounts_db: Arc<AccountsDB>,
 
     /// set of writable accounts which are currently in the pipeline
-    pub(crate) account_locks: Mutex<HashSet<Pubkey>>,
-
-    /// Set of read-only accounts which are currently in the pipeline, caching number of locks.
-    pub(crate) readonly_locks: Arc<RwLock<Option<HashMap<Pubkey, ReadonlyLock>>>>,
+    pub(crate) account_locks: Mutex<AccountLocks>,
```
sakridge, line 101: update comment 'set of read-only and writable accounts'

sakridge (dismissed):
looks good

runtime/src/accounts.rs
```diff
@@ -694,103 +731,53 @@ impl Accounts {
...
 
         for k in writable_keys {
-            locks.insert(*k);
+            account_locks.write_locks.insert(*k);
         }
 
         let readonly_writes: Vec<&&Pubkey> = readonly_keys
             .iter()
-            .filter(|k| !self.lock_readonly(k))
+            .filter(|k| !account_locks.lock_readonly(k))
```
sakridge, line 759: might as well do this all in one pass, no?

runtime/src/accounts.rs
```diff
@@ -694,103 +731,53 @@ impl Accounts {
...
         &self,
-        locks: &mut HashSet<Pubkey>,
+        account_locks: &mut AccountLocks,
         writable_keys: Vec<&Pubkey>,
         readonly_keys: Vec<&Pubkey>,
     ) -> Result<()> {
         for k in writable_keys.iter() {
-            if locks.contains(k) || self.is_locked_readonly(k) {
+            if account_locks.is_locked_write(k) || account_locks.is_locked_readonly(k) {
                 debug!("CD Account in use: {:?}", k);
```
CriesofCarrots, line 742: Nit: update debug, since we don't call anything credit-debit anymore
```suggestion
                debug!("Writable account in use: {:?}", k);
```

runtime/src/accounts.rs
```diff
@@ -694,103 +731,53 @@ impl Accounts {
...
                 return Err(TransactionError::AccountInUse);
             }
         }
         for k in readonly_keys.iter() {
-            if locks.contains(k) {
+            if account_locks.is_locked_write(k) {
                 debug!("CO Account in use: {:?}", k);
```
CriesofCarrots, line 748: Ditto for credit-only
```suggestion
                debug!("Read-only account in use: {:?}", k);
```

runtime/src/accounts.rs
CriesofCarrots: ➕

CriesofCarrots (approved):
Thanks!

CriesofCarrots (approved):

Issue #15374, opened 2021-02-17
Unexpected ProgramAccountNotFound


sakridge:
#### Problem

Validator on discord (nyet | blockdaemon) experienced a ProgramAccountNotFound, but the program account should exist:
```
2021-02-17T12:23:15.498710995Z WARN  solana_ledger::blockstore_processor] Unexpected validator error: ProgramAccountNotFound, transaction: Transaction { signatures: [MEbLYnemteXmmnCR
aYj6f8C89J2goMf5NoYawisgpLCUE6xKaBmZuRSEA5XNEJT2oUbZeeARmiCSWg5HLHFB1NQ], message: Message { header: MessageHeader { num_required_signatures: 1, num_readonly_signed_accounts: 0, num_r
eadonly_unsigned_accounts: 2 }, account_keys: [8Q9Kdp1b9GQh5qmc8fYesGkfAZMMw8uF3ButvZ1p78fY, 9Jyad5hREdVADqobAstaxBJCnivEUNRXJweo6RuXkmB, 8G7KdckAw8XCDDEN8G2gZ3A2EFZ9Nvo4TDq9HuwbUg3W,
 Bdx9Hx7wh7TBiQ5wFdakaLzX49uWaYuVKysPibYqPfjb, EUqojwWA2rd19FZrzeBncJsm38Jm1hEhE3zsmX3bRc2o, 11111111111111111111111111111111], recent_blockhash: 3BXoLEvpu8ZYNqkRKmmSsBKcCdRxkXviauoFk
mJk7VRG, instructions: [CompiledInstruction { program_id_index: 4, accounts: [1, 2, 3, 2, 2], data: [0, 3, 0, 0, 0, 1, 0] }, CompiledInstruction { program_id_index: 5, accounts: [0, 0
], data: [2, 0, 0, 0, 99, 5, 0, 0, 0, 0, 0, 0] }] } }
[2021-02-17T12:23:15.499329365Z ERROR solana_metrics::metrics] datapoint: validator_process_entry_error error="error: ProgramAccountNotFound, transaction: Transaction { signatures: [M
EbLYnemteXmmnCRaYj6f8C89J2goMf5NoYawisgpLCUE6xKaBmZuRSEA5XNEJT2oUbZeeARmiCSWg5HLHFB1NQ], message: Message { header: MessageHeader { num_required_signatures: 1, num_readonly_signed_acc
[...]

PR #15454, opened 2021-02-20
Account for possibility of cache flush in load()

carllin:
#### Problem
https://github.com/solana-labs/solana/issues/15374#issuecomment-782594750
#### Summary of Changes
TODO: Testing
Fixes #


ryoqun:
@carllin (sorry for delayed review for this. But I'm just too tired now. I'll review this with joy after I refresh my brain....)

runtime/src/accounts_db.rs
```diff
@@ -2269,7 +2404,11 @@ impl AccountsDb {
         offset: usize,
```
ryoqun, line 2404: this function's comment is a bit outdated. :)

runtime/src/accounts_db.rs
```diff
@@ -184,11 +187,12 @@ impl Versioned for (u64, AccountInfo) {
...
+    // None value in Cached variant means the cache was flushed
+    Cached(Option<(Pubkey, Cow<'a, CachedAccount>)>),
 }
 
 impl<'a> LoadedAccountAccessor<'a> {
-    fn get_loaded_account(&self) -> Option<LoadedAccount> {
+    fn get_loaded_account(&mut self) -> Option<LoadedAccount> {
         match self {
             LoadedAccountAccessor::Stored(storage_entry) => {
                 // May not be present if slot was cleaned up in between
```
ryoqun, line 198: nits: how about documenting our learnings?

```rust
// This not-deterministic behavior can't happen via consensus-sensitive tx execution path (like `load()`) but can do via rpc.
```

runtime/src/accounts_db.rs
```diff
@@ -2253,7 +2388,7 @@ impl AccountsDb {
     }
 
     pub fn load_slow(&self, ancestors: &Ancestors, pubkey: &Pubkey) -> Option<(Account, Slot)> {
-        self.load(ancestors, pubkey)
+        self.load(ancestors, pubkey, false)
```
ryoqun, line 2391: Hmm, this is called from `get_sysvar_account()` which could be called via the code path of bank-hash calculation like `update_sysvar_account`. So, I think we need this flag switch one or two higher level... ;)

runtime/src/accounts.rs
```diff
@@ -211,7 +211,7 @@ impl Accounts {
                     } else {
                         let (account, rent) = self
                             .accounts_db
-                            .load(ancestors, key)
+                            .load(ancestors, key, true)
```
ryoqun, line 214: nits: it'll be hard to know the meaning of `true` here in the future. how about defining small wrappers like `load_on_chain`/`load_off_chain`? (well, i couldn't come up with good name...)

runtime/src/accounts_db.rs
```diff
@@ -2200,33 +2212,156 @@ impl AccountsDb {
...
-            (slot, store_id, offset)
-            // `lock` released here
+        // If there are no entries in the index, return
+        let (slot, store_id, offset) = self.accounts_index_get(ancestors, pubkey, max_root)?;
+
+        #[cfg(test)]
+        {
+            // Give some time for cache flushing to occur here for unit tests
+            sleep(Duration::from_millis(self.load_delay));
+        }
```
ryoqun, line 2255: I  rather like this instead of hogging cpus for awhile. xD

runtime/src/accounts_db.rs
```diff
@@ -2200,33 +2212,156 @@ impl AccountsDb {
...
+                    // RPC may have fetched this storage entry after it was already `reset` by clean
+                    load_result
+                }
+            }
         };
 
-        //TODO: thread this as a ref
-        self.get_account_accessor_from_cache_or_storage(slot, pubkey, store_id, offset)
-            .get_loaded_account()
-            .map(|loaded_account| (loaded_account.account(), slot))
```
ryoqun, line 2229: I really miss this older yet simpler world... lol

runtime/src/accounts_db.rs
```diff
@@ -2200,33 +2212,156 @@ impl AccountsDb {
...
+        loaded_account.map(|loaded_account| (loaded_account.account(), slot))
     }
 
     pub fn load_account_hash(&self, ancestors: &Ancestors, pubkey: &Pubkey) -> Hash {
```
ryoqun, line 2367: :red_circle: I thought this bad code duplication might cause bugs. but this is only called from tests.... how about marking this  as `config(test)` or fixing the implementation?

runtime/src/accounts_db.rs
ryoqun: dumb question: can we hold the index key only if `is_transaction_load == true and store_id == CACHE_VIRTUAL_STORAGE_ID` while avoiding the possibility of heavy lock contention and dead locks?

runtime/src/accounts_db.rs
```diff
@@ -2200,33 +2212,156 @@ impl AccountsDb {
...
+                    // removes the slots' storage entries but doesn't purge from the accounts index
+                    // (account index cleanup is left to clean for stored slots).
+                    else {
+                        // Increment failsafe
+                        num_acceptable_failed_iterations += 1;
+                    }
+                }
+                LoadedAccountAccessor::Cached(None) => {
+                    // Cache was flushed in between checking the index and retrieving from the cache,
+                    // so retry. Don't increment `num_failed_iterations` here because it's impossible
```
ryoqun, line 2303: `num_acceptable_failed_iterations` looks to be incremented; maybe stale comment? (also, variable name is out of sync?)

runtime/src/accounts_db.rs
```diff
@@ -2200,33 +2212,156 @@ impl AccountsDb {
...
+                    // for this to fail for transaction loads more than once. This is because for a slot
+                    // `X` that's being replayed, there is only one latest ancestor containing the latest
+                    // update for the account, and this ancestor can only be flushed once.
+                    num_acceptable_failed_iterations += 1;
+                    if is_transaction_load {
+                        assert!(num_acceptable_failed_iterations <= 1);
+                    }
+                }
+                _ => {
```
ryoqun, line 2312: nits: I think `LoadedAccountAccessor::Cached(Some(_) | LoadedAccountAccessor::Stored(Some(_))` won't hurt us for now so much, trading for future maintainability aided by rust's exhaustiveness check. :)

runtime/src/accounts_db.rs
```diff
@@ -2200,33 +2212,156 @@ impl AccountsDb {
...
+                new_store_id,
+                new_offset,
+            );
+        }
+
+        let loaded_account: Option<LoadedAccount> = match &mut account_accessor {
+            LoadedAccountAccessor::Cached(None) | LoadedAccountAccessor::Stored(None) => {
+                panic!("Should have already been taken care of above, see comment above");
+            }
+            LoadedAccountAccessor::Cached(_) => {
```
ryoqun, line 2346: nits: `_` here is always `Some(_)` in fact?

runtime/src/accounts_db.rs
ryoqun: ... and that avoids rather complex new code?

runtime/src/accounts_db.rs
carllin: @ryoqun ha! I think there is actually an additional edge case to account for, fixed here: https://github.com/solana-labs/solana/pull/15454/commits/e7ae7620f23cf849b16efa621f5498f0ef7813c1

runtime/src/accounts_db.rs
carllin: hmmm I would like to avoid the clone of the entire account, especially large `> 1MB` ones, while holding the lock for now, maybe after the changes to not clone the entire account this will be a more performant option?

runtime/src/accounts_db.rs
carllin: ah yeah updated!

runtime/src/accounts_db.rs
carllin: yup that's correct, updated!

runtime/src/accounts_db.rs
carllin: Reminder: I should write additional test case for this...

runtime/src/accounts_db.rs
carllin: Unfortunately it was in `tests/accounts_db` which compiles this `runtime` without test flags, so instead I refactored and fixed the implementation: https://github.com/solana-labs/solana/pull/15454/commits/cdc6f15ec951d8bd5d1e796705fb8a4516a2e1ea

runtime/src/accounts_db.rs
carllin: This is a good catch, updated here: https://github.com/solana-labs/solana/pull/15454/commits/9d8a76a5306a6fe53e1f2cdbc2fb3152a6414510

runtime/src/accounts.rs
carllin: I renamed the `is_transaction_load` to `is_root_fixed` and renamed these functions to `load_slow/load_slow_with_fixed_root` communicating whether the root can move while the load is happening, which should be the most important bit

ryoqun:
Hooray, this pr is getting better. I'll do the 2nd pass.

runtime/src/bank.rs
```diff
@@ -3985,13 +3985,17 @@ impl Bank {
...
     // Generally, all of sysvar update granularity should be slot boundaries.
-    fn get_sysvar_account(&self, pubkey: &Pubkey) -> Option<Account> {
+    fn get_sysvar_account(&self, pubkey: &Pubkey, is_root_fixed: bool) -> Option<Account> {
         let mut ancestors = self.ancestors.clone();
         ancestors.remove(&self.slot());
-        self.rc
-            .accounts
-            .load_slow(&ancestors, pubkey)
-            .map(|(acc, _slot)| acc)
+        let res = if is_root_fixed {
```
ryoqun, line 3991: Maybe, my prev round of comment was confusing, but I don't think this condition is needed because this fn's sole caller is just always giving us `true`.

runtime/src/accounts_db.rs
ryoqun: I see. potentially large Account copies are fair point. :)

runtime/src/accounts_background_service.rs
```diff
@@ -441,9 +441,19 @@ mod test {
         bank0.store_account(&account_key, &Account::new(264, 0, &Pubkey::default()));
         assert!(bank0.get_account(&account_key).is_some());
         pruned_banks_sender.send(0).unwrap();
+
+        assert!(!bank0
+            .rc
+            .accounts
+            .scan_slot(0, |stored_account| Some(stored_account.stored_size()))
```
ryoqun, line 448: nits: maybe replace `Some(stored_account.stored_size())` with `unreachable!()` to better convey this closure doesn't matter?

runtime/src/accounts_background_service.rs
```diff
@@ -441,9 +441,19 @@ mod test {
...
+            .is_empty());
+
         AccountsBackgroundService::remove_dead_slots(&bank0, &request_handler, &mut 0, &mut 0);
 
-        // Slot should be removed
-        assert!(bank0.get_account(&account_key).is_none());
+        assert!(bank0
+            .rc
+            .accounts
+            .scan_slot(0, |stored_account| Some(stored_account.stored_size()))
```
ryoqun, line 456: same as above https://github.com/solana-labs/solana/pull/15454/files#r590761091

runtime/src/bank.rs
ryoqun: So, I'm saying this time that you can directly call `self.rc.accounts.load_slow_with_fixed_root(...)` from here.

runtime/src/accounts_db.rs
```diff
@@ -2209,67 +2251,186 @@ impl AccountsDb {
...
+
+            slot = new_slot;
+            store_id = new_store_id;
+            offset = new_offset;
+
+            // Accounts cache was just flushed, look in the index to find
+            // the account storage again. This works because in accounts cache flush,
+            // an account is written to storage *before* it is removed from the cache
+            account_accessor =
+                self.get_account_accessor_from_cache_or_storage(slot, pubkey, store_id, offset);
```
ryoqun, line 2398: this

runtime/src/accounts_db.rs
```diff
@@ -2209,67 +2251,186 @@ impl AccountsDb {
...
-            .map(|loaded_account| *loaded_account.loaded_hash())
-            .unwrap()
+        #[cfg(test)]
+        {
+            // Give some time for cache flushing to occur here for unit tests
+            sleep(Duration::from_millis(self.load_delay));
+        }
+
+        let mut account_accessor =
+            self.get_account_accessor_from_cache_or_storage(slot, pubkey, store_id, offset);
```
ryoqun, line 2302: Maybe, we can simplify this crutial looping code a bit by unifying `get_account_accessor_from_cache_or_storage` below (https://github.com/solana-labs/solana/pull/15454/files#r590768524), by moving this call to the top inside the `loop`?

runtime/src/accounts_db.rs
```diff
@@ -2209,67 +2251,186 @@ impl AccountsDb {
...
     }
 
-    pub fn load_slow(&self, ancestors: &Ancestors, pubkey: &Pubkey) -> Option<(Account, Slot)> {
-        self.load(ancestors, pubkey)
+    fn do_load(
+        &self,
+        ancestors: &Ancestors,
+        pubkey: &Pubkey,
+        max_root: Option<Slot>,
+        // true if the root will not move during this load
```
ryoqun, line 2409: maybe better of defining this `move`? like `(i.e. not flushed from AccountsCache nor cleaned and shrinked)` (Yeah, I know this is a bit verbose... But I want some documented definition of this flag in accessible way for future me).

runtime/src/accounts_db.rs
```diff
@@ -2209,67 +2251,186 @@ impl AccountsDb {
...
+                        // latest ancestor containing the latest update for the account, and this
+                        // ancestor can only be flushed once.
+                        // 2) The root cannot move while replaying, so the index cannot continually
+                        // find more up to date entries than the current `slot`
+                        assert!(num_acceptable_failed_iterations <= 1);
+                    }
+                }
+                LoadedAccountAccessor::Cached(Some(_)) | LoadedAccountAccessor::Stored(Some(_)) => {
+                    // Everything else means there was no race, so break out and continue
+                    break;
```
ryoqun, line 2373: To realize this: https://github.com/solana-labs/solana/pull/15454/files#r590769646, maybe this `break` can simply replaced with `return Some((account_accessor, slot))`. which I think more readable in this case.

runtime/src/accounts_db.rs
```diff
@@ -2209,67 +2251,186 @@ impl AccountsDb {
...
+            offset = new_offset;
+
+            // Accounts cache was just flushed, look in the index to find
+            // the account storage again. This works because in accounts cache flush,
+            // an account is written to storage *before* it is removed from the cache
+            account_accessor =
+                self.get_account_accessor_from_cache_or_storage(slot, pubkey, store_id, offset);
+        }
+
+        Some((account_accessor, slot))
```
ryoqun, line 2401: ... (continueing from https://github.com/solana-labs/solana/pull/15454/files#r590795305), then, don't forget to put `unreachable!("foo bar...")` for human sanity. :)

runtime/src/accounts_db.rs
ryoqun: that's because I don't think we should generally mix `break` and `return` inside a loop.. xD

ryoqun: let's stick to `return` for this case. :)

runtime/src/accounts_db.rs
```diff
@@ -2209,67 +2251,186 @@ impl AccountsDb {
...
+
+                        // RPC get_account() may have fetched an old root from the index that was
+                        // either:
+                        // 1) Cleaned up by clean_accounts(), so the accounts index has been updated
+                        // and the storage entries have been removed.
+                        // 2) Dropped by purge_slots() because the slot was on a minor fork, which
+                        // removes the slots' storage entries but doesn't purge from the accounts index
+                        // (account index cleanup is left to clean for stored slots).
+                        num_acceptable_failed_iterations += 1;
+                    }
```
ryoqun, line 2354: if `is_root_fixed` is `true`, and `LoadedAccountAccessor::Stored(None)`, this could possibly lead into infinite loop. Anyway, this condition should never occur? how about like this?

```
    ...
} else {
    panic!("this shouldn't happen because of foo bar...");
}

```

runtime/src/accounts_db.rs
```diff
@@ -2209,67 +2251,186 @@ impl AccountsDb {
...
+                        // find more up to date entries than the current `slot`
+                        assert!(num_acceptable_failed_iterations <= 1);
+                    }
+                }
+                LoadedAccountAccessor::Cached(Some(_)) | LoadedAccountAccessor::Stored(Some(_)) => {
+                    // Everything else means there was no race, so break out and continue
+                    break;
+                }
+            }
+            let (new_slot, new_store_id, new_offset) =
```
ryoqun, line 2376: nits: how about adding this one line comment? `// start retrying from the index read`

runtime/src/accounts_db.rs
```diff
@@ -2209,67 +2251,186 @@ impl AccountsDb {
...
+                        // 2) The root cannot move while replaying, so the index cannot continually
+                        // find more up to date entries than the current `slot`
+                        assert!(num_acceptable_failed_iterations <= 1);
+                    }
+                }
+                LoadedAccountAccessor::Cached(Some(_)) | LoadedAccountAccessor::Stored(Some(_)) => {
+                    // Everything else means there was no race, so break out and continue
+                    break;
+                }
+            }
```
ryoqun, line 2375: nits: how about adding this one line comment? `// start retrying from the index read`

runtime/src/accounts_db.rs
```diff
@@ -2209,67 +2251,186 @@ impl AccountsDb {
...
+            let (new_slot, new_store_id, new_offset) =
+                self.accounts_index_get(ancestors, pubkey, max_root)?;
+            if new_slot == slot && new_store_id == store_id {
+                // If the entry was missing from the cache, that means it must have been flushed,
+                // and the accounts index is always updated before cache flush, so this
+                // should be impossible
+                assert!(new_store_id != CACHE_VIRTUAL_STORAGE_ID);
+
```
ryoqun, line 2383: we can also add `assert!(new_offset == offset)` here? Otherwise, it's a bit hard to determine we should add the condition to the enclosing `if` or not... In this case, I think assert is more rigid.

ryoqun:
Okay, I went through formal review of the meaty logic. Pretty well thought code. I think the next review round will be the last one after some more units tests as you said. :)

runtime/src/bank.rs
carllin: hmmm I was slightly worried somebody might call `get_sysvar_account` in the future in a nonsafe way,  if there wasn't some indication that it may be potentially dangerous. Maybe I should just rename this function to `get_sysvar_account_with_fixed_root`?

runtime/src/accounts_background_service.rs
carllin: yeah the contents of the closure don't matter, but the closure will still be executed. Replaced with `()`

runtime/src/accounts_db.rs
carllin: Heh looks like clippy doesn't like that since it knows the loop doesn't break, so the `unreachable!("foo bar...")` is itself unreachable

runtime/src/accounts_db.rs
carllin: I think it can happen. Even during replay, you could have continuous shrink + clean of an existing storage entry, see comments here: https://github.com/solana-labs/solana/commit/e7ae7620f23cf849b16efa621f5498f0ef7813c1#diff-1090394420d51617f3233275c2b65ed706b35b53b115fe65f82c682af8134a6fR2289-R2296

carllin: This is part of the reason I went with the `is_root_fixed` name, b/c if the root is fixed, there should be a bounded number of pending cleans/shrinks

runtime/src/accounts_db.rs
carllin: > Also, this is really only `root`? I think random AccountsCache evinction and full-of-cache condition could flush the un rooted ones too? so, this could more be named like `is_slot_fixed`?

This is true, but based on the *usage* of that flag, I couldn't come up with a better name than `is_root_fixed`, see comment here: https://github.com/solana-labs/solana/pull/15454/files#r590921257

sakridge:
@ryoqun can you take over this PR from carl?

ryoqun:
> @ryoqun can you take over this PR from carl?

@sakridge Yeah, I can. :)

ryoqun:
Sorry for delay; other various interrupts.., I'm going to start to work on this within 24 hours.

ryoqun:
work time!

runtime/src/accounts.rs
```diff
@@ -451,19 +451,42 @@ impl Accounts {
...
     /// Slow because lock is held for 1 operation instead of many
     pub fn load_slow(
         &self,
         ancestors: &Ancestors,
         pubkey: &Pubkey,
+        is_root_fixed: bool,
     ) -> Option<(AccountSharedData, Slot)> {
-        let (account, slot) = self.accounts_db.load_slow(ancestors, pubkey)?;
+        let (account, slot) = self.accounts_db.load(ancestors, pubkey, is_root_fixed)?;
+        Self::filter_zero_lamport_account(account, slot)
```
ryoqun, line 473: I don't like this particularly quirk logic insertion here at wrong layer, imo. But this is different story from this pr. and let's just keep existing behavior.

runtime/src/accounts_db.rs
```diff
@@ -279,7 +312,7 @@ impl<'a> LoadedAccount<'a> {
         }
     }
 
-    pub fn account(self) -> AccountSharedData {
+    pub fn take_account(self) -> AccountSharedData {
```
ryoqun, line 315: As this takes ownership, I think using `take_*` is more natural in rust. Otherwise, this just looks like a simple no side-effect getter.

runtime/src/accounts_db.rs
```diff
@@ -9537,4 +9909,273 @@ pub mod tests {
...
+                            .unwrap();
+
+                        sleep(Duration::from_millis(RACY_SLEEP_MS));
+
+                        let mut account_accessor =
+                            db.get_account_accessor(slot, &pubkey, store_id, offset);
+                        match account_accessor.get_loaded_account() {
+                            None => continue, // for RPC case, observing None is valid sometimes
+                            Some(account_accessor) => (account_accessor.take_account(), slot),
+                        }
```
ryoqun, line 9959: here

runtime/src/accounts_db.rs
```diff
@@ -185,27 +188,57 @@ impl Versioned for (u64, AccountInfo) {
...
-            LoadedAccountAccessor::Cached((cache, slot, pubkey)) => {
-                // May not be present if slot was cleaned up in between
-                cache.load(*slot, pubkey).map(|cached_account| {
-                    LoadedAccount::Cached((**pubkey, Cow::Owned(cached_account)))
-                })
+            LoadedAccountAccessor::Cached(Some(_cached_account)) => {
+                // Cached(Some(x)) variant always produces `Some` for get_loaded_account() since
+                // it just returns the inner `x` without additional fetches
+                self.get_loaded_account()
+                    .expect("Cache flushed should be handled before trying to fetch account")
```
ryoqun, line 210: Previously, this `.expect()` existed in `get_loaded_account`. But, for the rpc case, this `panic!`s. So, I moved here. Also, I enhanced a test https://github.com/solana-labs/solana/pull/15454/files?diff=split&w=1#r606629658

runtime/src/accounts_db.rs
```diff
@@ -2133,16 +2163,21 @@ impl AccountsDb {
...
             |pubkey, (account_info, slot)| {
+                // unlike other scan fns, this is called from Bank::collect_rent_eagerly(),
+                // which is on-consensus processing in the banking/replaying stage.
+                // This requires infallible and consistent account loading.
+                // So, we unwrap Option<LoadedAccount> from get_loaded_account() here.
+                // This is safe because this closure is invoked with the account_info,
+                // while we lock the index entry at AccountsIndex::do_scan_accounts() ultimately,
+                // meaning no other subsystems can invalidate the account_info before making their
+                // changes to the index entry.
+                // For details, see the comment in retry_to_get_account_accessor()
```
ryoqun, line 2174: this is another additional subtlety I found while finishing up this pr... pretty hard to spot...

runtime/src/accounts_db.rs
```diff
@@ -2259,28 +2289,279 @@ impl AccountsDb {
...
+        &'a self,
+        mut slot: Slot,
+        mut store_id: usize,
+        mut offset: usize,
+        ancestors: &'a Ancestors,
+        pubkey: &'a Pubkey,
+        max_root: Option<Slot>,
+        is_root_fixed: bool,
+    ) -> Option<(LoadedAccountAccessor<'a>, Slot)> {
+        // Happy drawing time! :)
```
ryoqun, line 2343: this jumbo comment is my internalizing process of this fn. lol

runtime/src/accounts_db.rs
```diff
@@ -2259,28 +2289,279 @@ impl AccountsDb {
...
+                        // 1) Cleaned up by clean_accounts(), so the accounts index has been updated
+                        // and the storage entries have been removed.
+                        // 2) Dropped by purge_slots() because the slot was on a minor fork, which
+                        // removes the slots' storage entries but doesn't purge from the accounts index
+                        // (account index cleanup is left to clean for stored slots).
+                        num_acceptable_failed_iterations += 1;
+                    }
+                }
+            }
+            if num_acceptable_failed_iterations >= 10 {
```
ryoqun, line 2513: I moved this check from the top of this `loop` to this place because otherwise this `datapoint_warn!` would print after updating the `(slot, store_id, offset)` tuple... much less helpful.

runtime/src/accounts_db.rs
```diff
@@ -2259,28 +2289,279 @@ impl AccountsDb {
...
+                // The latest version of the account existed in the index, but could not be
+                // fetched from storage. This means a race occurred between this function and clean
+                // accounts/purge_slots
+                let message = format!(
+                    "do_load() failed to get key: {} from storage, latest attempt was for \
+                     slot: {}, storage_entry: {} is_root_fixed: {}",
+                    pubkey, slot, store_id, is_root_fixed,
```
ryoqun, line 2520: `is_root_fixed` can be prooved to be `false` always. but no harm to print it anyway.

runtime/src/accounts_db.rs
```diff
@@ -9537,4 +9909,273 @@ pub mod tests {
...
+        do_test_load_account_and_shrink_race(true);
+    }
+
+    #[test]
+    fn test_load_account_and_shrink_race_without_retry() {
+        do_test_load_account_and_shrink_race(false);
+    }
+
+    #[test]
+    fn test_load_account_and_purge_race() {
```
ryoqun, line 10125: here

runtime/src/accounts_db.rs
```diff
@@ -2259,28 +2289,279 @@ impl AccountsDb {
...
+
+                // If this is not a cache entry, then this was a minor fork slot
+                // that had its storage entries cleaned up by purge_slots() but hasn't been
+                // cleaned yet. That means this must be rpc access and not replay/banking.
+                // But we can't guarantee `is_root_fixed' must be false, because
+                // the dangling bank at the tip of minor fork might not be started to freeze
+                // (!freeze_started =~> is_root_fixed)
+
+                // Everything being said, let's return None as it's an error condition after all.
+                return None;
```
ryoqun, line 2548: I adjust these assertion ordering and commentary and added a test for this: https://github.com/solana-labs/solana/pull/15454/files?diff=split&w=1#r606630524

runtime/src/accounts_db.rs
```diff
@@ -2259,28 +2289,279 @@ impl AccountsDb {
...
+
+        // Failsafe for potential race conditions with other subsystems
+        let mut num_acceptable_failed_iterations = 0;
+        loop {
+            let account_accessor = self.get_account_accessor(slot, pubkey, store_id, offset);
+            match account_accessor {
+                LoadedAccountAccessor::Cached(Some(_)) | LoadedAccountAccessor::Stored(Some(_)) => {
+                    // Great! There was no race, just return :) This is the most usual situation
+                    return Some((account_accessor, slot));
+                }
```
ryoqun, line 2456: Generally, I like to put normal cases at the front of sibling branches so that our brain's stack can be retained low.

runtime/src/accounts_db.rs
```diff
@@ -2259,28 +2289,279 @@ impl AccountsDb {
...
-            let (
-                slot,
-                AccountInfo {
-                    store_id, offset, ..
-                },
-            ) = slot_list[index];
-            (slot, store_id, offset)
-            // `lock` released here
-        };
+        let (slot, store_id, offset) = self.read_index_for_accessor(ancestors, pubkey, max_root)?;
```
ryoqun, line 2564: I moved this initial index read out of the retry function, thanks to the read cache.

runtime/src/accounts_db.rs
```diff
@@ -2289,91 +2570,77 @@ impl AccountsDb {
...
+            store_id,
+            offset,
+            ancestors,
+            pubkey,
+            max_root,
+            is_root_fixed,
+        )?;
+        let loaded_account = account_accessor.check_and_get_loaded_account();
+        let is_cached = loaded_account.is_cached();
+        let account = loaded_account.take_account();
```
ryoqun, line 2584: So, this is the heart of this pr. And I renamed bunch to make this chunk of code to read best.

runtime/src/accounts_db.rs
```diff
@@ -3660,7 +3929,11 @@ impl AccountsDb {
                                 {
                                     let (slot, account_info) = &lock.slot_list()[index];
                                     if account_info.lamports != 0 {
-                                        self.get_account_accessor_from_cache_or_storage(
+                                        // Because we're keeing the `lock' here, there is no need
+                                        // to use retry_to_get_account_accessor()
+                                        // In other words, flusher/shrinker/cleaner is blocked to
+                                        // cause any Accessor(None) situtation.
```
ryoqun, line 3935: I hope my reasoning is right here?

runtime/src/accounts_db.rs
```diff
@@ -3416,8 +3683,10 @@ impl AccountsDb {
                 );
             }
 
-            // Remove this slot from the cache, which will to AccountsDb readers should look like an
-            // atomic switch from the cache to storage
+            // Remove this slot from the cache, which will to AccountsDb's new readers should look like an
+            // atomic switch from the cache to storage.
+            // There is some racy condition for existing readers who just has read exactly while
+            // flushing. That case is handled by retry_to_get_account_accessor()
```
ryoqun, line 3689: let's cross reference. :)

runtime/src/bank.rs
```diff
@@ -4003,7 +4003,21 @@ impl Bank {
...
+
+    fn load_slow(
+        &self,
+        ancestors: &Ancestors,
+        pubkey: &Pubkey,
+    ) -> Option<(AccountSharedData, Slot)> {
+        // get_account (= this fn caller) can be called for various on-consensus
+        // code like inflation, feature activation and native program loading
+        // before freezing and it's generally impossible to make static/lexical guarantee of RPC or
+        // on-consensus.. so just pass !freeze_started for is_root_fixed as a close approximation
```
ryoqun, line 4017: this is another potential bug fix.

ryoqun:
@carllin could you review this? I've gone through the whole relevant code adding comments and a few fixes. So, it turned out the new changes are large. For reference, here's the pre-rebase branch (most of changes are in separated commits there): https://github.com/ryoqun/solana/pull/new/FixAccountsDbBeforeRebase

carllin:
> @carllin could you review this? I've gone through the whole relevant code adding comments and a few fixes. So, it turned out the new changes are large. For reference, here's the pre-rebase branch (most of changes are in separated commits there): https://github.com/ryoqun/solana/pull/new/FixAccountsDbBeforeRebase

Yup! Looks like a lot of good stuff here, will take some time to carefully parse through it, thanks!

runtime/src/accounts_db.rs
```diff
@@ -3660,7 +3929,11 @@ impl AccountsDb {
                                 {
                                     let (slot, account_info) = &lock.slot_list()[index];
                                     if account_info.lamports != 0 {
-                                        self.get_account_accessor_from_cache_or_storage(
+                                        // Because we're keeing the `lock' here, there is no need
```
carllin, line 3932: nit: keeping

runtime/src/accounts_db.rs
carllin: 1) It should be safe with cache flush because cache flush `AccountsDb::flush_slot()`doesn't remove the slot entry from the cache until the accounts index gets updated first, which can't happen while we hold the lock here.

2) Any reclaims in `clean_accounts()` needs to go though the accounts index lock: https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_db.rs#L1493-L1494 first, so the slot can't be dead yet, and the slot storages should not be cleaned up by `process_dead_slots()` yet

3) In shrink, this logic which updates the accounts index: https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_db.rs#L1908-L1914 needs to run first before storage entries are removed: https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_db.rs#L1926-L1930, so this should also be ok

Moreover for 2) and 3) loads in `calculate_accounts_hash()` should not  *currently* race with clean/shrink because the full hash is synchronous with clean/shrink in AccountsBackgroundService, but this could change in the future, so it's good you called it out here 😄

runtime/src/accounts_db.rs
carllin: oh hmm, for the RPC panic:

1) is it happening from a scan like `scan_accounts()`/`unchecked_scan_accounts()`. Hopefully these scans shouldn't encounter this panic because they hold the index lock: https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_index.rs#L507-L515. If this is panicking worried this means that RPC is getting inconsistent state.

2)And double checking here. For point lookups/fetch of a specific account through `do_load()`, even if we hit the max retryable iterations, then this should just return: https://github.com/solana-labs/solana/blob/3e0377b45867a2579710632fe00099fb13c3322d/runtime/src/accounts_db.rs#L2581 before calling `check_and_get_loaded_account()`

runtime/src/accounts_db.rs
```diff
@@ -2259,28 +2289,279 @@ impl AccountsDb {
...
+        // P3 purge_slots_from_cache_and_store()/ | index
+        //        purge_slot_cache_pubkeys()      | (removes existing store_id, offset for caches)
+        //      OR                                |
+        //    clean_accounts()/                   |
+        //        clean_old_rooted_accounts()     | (removes existing store_id, offset for stores)
+        //                                        V
+        //
+        // Remarks for purger: So, for any reading operations, it's a race conditon
+        // where P2 happens between R1 and R2. In that case, retrying from R1 is safu.
+        // In that case, we may bail at index read retry when P3 hasn't been run
```
ryoqun, line 2440: here

runtime/src/accounts_db.rs
ryoqun: @carllin 

> for the RPC panic:

good question. From this discussion, I think we can revert to your original change and even simplify more.

> 1. is it happening from a scan like `scan_accounts()`/`unchecked_scan_accounts()`. Hopefully these scans shouldn't encounter this panic because they hold the index lock: 

This is correct. The only concern I had was `purging` ( https://github.com/solana-labs/solana/pull/15454/files?diff=split&w=1#r606978231 ), which does irregular update order of `cache/store remove => index` update unlike other race-condition triggers. In conclusion, I think we can promote the purging race-condition detection a hard error (https://github.com/solana-labs/solana/pull/15454/files?diff=split&w=1#r606630573).

let me explain in pieces. :)

My original motivation of this `.expect()` move was that I thought we need to consider the purging case because of the existence of the purging race-condition detection: If `do_load(with_fixed_root = false)` from rpc must be handled for the purging, we likewise also must handle the purging case for the `scan_accounts()` and its families.

if that's true (I think it's not now), holding the index lock isn't enough because the purging code firstly removes caches/stores from the store:  https://github.com/solana-labs/solana/pull/15454/files?diff=split&w=1#r606978231

But then, I now think there should be no possibilities of purging race to begin with: All of code path for both `do_load` and `scan` actually ensures Bank's liveliness because they're called via as a method of `Bank` from external subsystems (rpc, replay&banking). So, combined with `Bank::parent` fields, all of ancestors should be alive then none of them should be purged/pruned. 


image

image



With these things are done, am I right? Or, did you add the original purging detection code in `do_load` ( https://github.com/solana-labs/solana/commit/2885d989a4d6a93fcce0adca7f6210709550184e#r49109545 ) with some concrete concerns in mind at the time? 


> 2) And double checking here. For point lookups/fetch of a specific account through `do_load()`, even if we hit the max retryable iterations, then this should just return:

Yeah, this should be safe. So, in this regard too, we can restore the `.expect` in `get_loaded_account()`.

runtime/src/accounts_db.rs
```diff
@@ -9537,4 +9909,273 @@ pub mod tests {
...
+            let exit = exit.clone();
+
+            std::thread::Builder::new()
+                .name("account-purge".to_string())
+                .spawn(move || loop {
+                    if exit.load(Ordering::Relaxed) {
+                        return;
+                    }
+                    // Simulate purge_slots()
+                    db.purge_slot(slot);
```
ryoqun, line 10156: here

runtime/src/accounts_db.rs
ryoqun: Also, It means the purge race-condition test is meaningless because it's doing which can't happen in the wild. xD

runtime/src/accounts_db.rs
ryoqun: cool! thanks for confirming. :)

runtime/src/accounts_db.rs
carllin: All these nice links aren't working for me 😢 

> But then, I now think there should be no possibilities of purging race to begin with: All of code path for both `do_load` and `scan` actually ensures Bank's liveliness because they're called via as a method of `Bank` from external subsystems (rpc, replay&banking). So, combined with `Bank::parent` fields, all of ancestors should be alive then none of them should be purged/pruned.

Haha yup I'm pretty sure I went through this exact line of reasoning when fixing up the scan the scan: https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_index.rs#L415-L419

There's also a test for this here: https://github.com/solana-labs/solana/blob/master/runtime/src/bank.rs#L11533 which constantly scans unrooted slots that are being removed from BankForks by `squash()`

carllin: > good question. From this discussion, I think we can revert to your original change and even simplify more.
> 
> In conclusion, I think we can promote the purging race-condition detection a hard error (https://github.com/solana-labs/solana/pull/15454/files?diff=split&w=1#r606630573).

Could you point me to the commit with this change, I don't think I see it yet.

runtime/src/accounts_db.rs
```diff
@@ -2259,28 +2289,279 @@ impl AccountsDb {
...
+        //          |                           |
+        //          V                           |
+        // F3 store_accounts_frozen()/          | index
+        //        update_index()                | (replaces existing store_id, offset in caches)
+        //          |                           |
+        //          V                           |
+        // F4 accounts_cache.remove_slot()      | map of caches (removes old entry)
+        //                                      V
+        //
+        // Remarks for flusher: So, for any reading operations, it's a race conditon where F4 happens
```
carllin, line 2383: nit: conditon -> condition

runtime/src/accounts_db.rs
```diff
@@ -2259,28 +2289,279 @@ impl AccountsDb {
...
+
+                // If the entry was missing from the cache, that means it must have been flushed,
+                // and the accounts index is always updated before cache flush, so store_id must
+                // not indicate being cached.
+                assert!(new_store_id != CACHE_VIRTUAL_STORAGE_ID);
+
+                // If this is not a cache entry, then this was a minor fork slot
+                // that had its storage entries cleaned up by purge_slots() but hasn't been
+                // cleaned yet. That means this must be rpc access and not replay/banking.
+                // But we can't guarantee `is_root_fixed' must be false, because
```
carllin, line 2543: Hmmm I don't know if it's currently possible for RPC to query unfrozen banks, I think all the commitment levels are at least frozen: https://github.com/solana-labs/solana/blob/master/sdk/src/commitment_config.rs#L139-L196, which maps to  https://github.com/carllin/solana/blob/master/runtime/src/commitment.rs#L112-L120

runtime/src/bank.rs
carllin: hmmm I'm worried about hiding this behind `bank.freeze()`. There's no guarantee in the future that we won't need to load something from a frozen bank where the `load` is not allowed to fail (some services just keep a bank around as an accessor to the AccountsDb). 

I'm thinking it should be up to the caller, like RPC, to decide if a failure is acceptable.

runtime/src/accounts_db.rs
carllin: Yup, thanks for taking care of that conflict!

runtime/src/accounts_db.rs
carllin: ha yup, this one is tricky. Lucky there's no bug in existing code for this case thought!

runtime/src/accounts_db.rs
```diff
@@ -2259,28 +2289,279 @@ impl AccountsDb {
...
+                // the dangling bank at the tip of minor fork might not be started to freeze
+                // (!freeze_started =~> is_root_fixed)
```
carllin, line 2545: Relevant to this comment: https://github.com/solana-labs/solana/pull/15454/files#r607473929 as well if we decide to change it.

runtime/src/accounts_db.rs
ryoqun: > These nice links aren't working for me cry

(github review functionality sucks. Usually, you can view what I'm referring to by opening a new browser window from the link if just clicking didn't work...)

Well, I waited for your go to promote this clause to `panic!()`, here's the freshly written commit: e02974b27a7afcafdfa34b54afcd69c072dc17c5

ryoqun: ^^ @carllin (I'm back working on this pr!)

runtime/src/accounts_db.rs
ryoqun: done!

runtime/src/accounts_db.rs
ryoqun: And finally I reverted to your original code: ecdf6ae1bc981a1884135482394ee818097f8ae0

runtime/src/bank.rs
ryoqun: So, it took some back-and-forth by myself. Here's my revamped solution with another beefy commit: e4590ebae67ba4884c69c84c50e84c3b9344c3be

Let's reframe the situation a bit.

Namely, we can elide the externally observable race conditions altogether, _if we re-interpret the is_root_fixed flag to merely be an optimization hint and make the RPC loading route to newly-added slow path inside the index lock._

First of all, plumbing is_root_fixed up to RPC (as suggested by you: https://github.com/solana-labs/solana/pull/15454#discussion_r607473929) is very tedious(77010dbbe2f2f2cd73b228bf5a0e86f90632eac4). Also, the outcome is that rpc remain to be exposed to the race condition (which is very undesirable, come to think of it), even including very fundamental methods like `getAccount` with the `confirmed/finalized` commitment while holding the bank correctly. That's bad. we're just externalizing our bad race bug just to others shoulders. I think rpc should be treated as a first-citizen. Also, I don't want to handle supporting tickets for these insane flaky rpc behavior and recommend partners to retry even getAccount as it could _be silently faulty_ (returning empty account at the definitely existing address), which sounds pretty lame. ;) Maybe unrelated to this, As a illustration, I want to avoid these hard-to-debug rpc bug reports: https://github.com/solana-labs/solana/issues/15681, people would eventually encounter in the future if we had this data race rather permanent thing.

Finally, about some supportive arguments. Firstly, I think it's acceptable from RPC consumer's perspective for rpc to be slightly slow and to affect on-consensus processing a bit, rather than the hard-to-debug data-race bugs. (so this is revised version of my previous failed suggestion: https://github.com/solana-labs/solana/pull/15454#discussion_r590759082).

Also, I believe newer implementation for rpc load doesn't cause problems:
- I made relevant `match` arms be `unreachable!()`, removing the need of `num_acceptable_failed_iterations` and making `retry_to_get_account_accessor` be used only for `LoadSafety::FixedMaxRoot` (this makes things simpler)
- With the assumption no purge race (e02974b27a7afcafdfa34b54afcd69c072dc17c5 ecdf6ae1bc981a1884135482394ee818097f8ae0) with RPC, we can prove that the new rpc load doesn't fail anymore, as long as we hold the corresponding lock in read_index_for_accessor_or_load_slow.
- replaying/banking works as fast as before if there is no rpc access. :)

@carllin How do you like this?

carllin: hmmmm, I'm still worried about the perf implications of loading these large accounts under heavier RPC load. I'm especially hesitant to incur these costs, since these failures should be very rare (race means you would have to consistently encounter a new root between the accounts index fetch and the accounts load, especially since roots should not be being set more than once every few hundred ms). 

One thing we could do is get rid of the failure case altogether, and just keep retrying with some very absurdly high limit. I would be very surprised if this ever ended up in a loop that lasted more than a few iterations. We can keep the logging around to observe if this ever does happen.

ryoqun: I see! thanks for quick suggestion. I'll adjust this pr yet again. :)

ryoqun: @carllin hi, here's my another patch, believing it's done accordingly as suggested by you: https://github.com/solana-labs/solana/pull/15454/commits/b8ced7f7334a4e2f2558affd2b461c456f445a6b

Could you review this? thanks for in-depth review!

ryoqun: This patch restores the previous behavior and I checked all of new tests fails with this:

```patch
$ git diff carllin/FixAccountsDb
index 4fa24ede44..3799f73301 100644
@@ -2499,6 +2499,8 @@ impl AccountsDb {
         let mut num_acceptable_failed_iterations = 0;
         loop {
             let account_accessor = self.get_account_accessor(slot, pubkey, store_id, offset);
+            return Some((account_accessor, slot));
+
             match account_accessor {
                 LoadedAccountAccessor::Cached(Some(_)) | LoadedAccountAccessor::Stored(Some(_)) => {
                     // Great! There was no race, just return :) This is the most usual situation
```

```
failures:
    accounts_db::tests::test_load_account_and_cache_flush_race_with_retry
    accounts_db::tests::test_load_account_and_cache_flush_race_without_retry
    accounts_db::tests::test_load_account_and_purge_race_with_retry
    accounts_db::tests::test_load_account_and_purge_race_without_retry
    accounts_db::tests::test_load_account_and_shrink_race_with_retry
    accounts_db::tests::test_load_account_and_shrink_race_without_retry

test result: FAILED. 0 passed; 6 failed; 0 ignored; 0 measured; 489 filtered out; finished in 30.10s

error: test failed, to rerun pass '-p solana-runtime --lib'

```

ryoqun: Also, I'll run a test validator against mainnet-beta for a day or two before merging this to make sure everything works.

runtime/src/bank.rs
```diff
@@ -4003,7 +4003,25 @@ impl Bank {
...
+        // get_account (= this fn caller) can be called for various on-consensus
+        // code like inflation, feature activation and native program loading
+        // before freezing and it's generally impossible to make static/lexical guarantee of RPC or
+        // on-consensus.. so just pass !freeze_started for LoadHint::FixedMaxRoot as a close
+        // approximation of replaying/banking.
+        let load_hint = if !self.freeze_started() {
+            LoadHint::FixedMaxRoot
+        } else {
+            LoadHint::Unspecified
+        };
```
carllin, line 4023: I fell like this this might not be the most intuitive mapping to what each `LoadHint` does. The purpose of the load hint is just this assert here: https://github.com/solana-labs/solana/blob/b8ced7f7334a4e2f2558affd2b461c456f445a6b/runtime/src/accounts_db.rs#L2522 right?

For instance, if a different thread than replay keeps around a reference to an unfrozen bank and calls load, they may see more than one failure if the root keeps moving, triggering that assert.

I think the safest thing to do here is to default to the same behavior `LoadHint::Unspecified`, but in specific paths like inflation, feature activation and native program loading, plumb the `LoadHint::FixedMaxRoot` where we know it's safe.

runtime/src/bank.rs
carllin: @ryoqun looks good! I had one last remaining concern here: https://github.com/solana-labs/solana/pull/15454#discussion_r612045902.

runtime/src/bank.rs
ryoqun: @carllin thanks for raising that concern. Yeah, let's be on the safer side. I've went through all uses of `get_account`: 0465e355ee196c2c1fe330eff6dc87b85389070a. Maybe this is what you wanted?

fyi, this is much simpler version of https://github.com/solana-labs/solana/commit/77010dbbe2f2f2cd73b228bf5a0e86f90632eac4. ;)

runtime/src/bank.rs
```diff
@@ -4170,12 +4209,9 @@ impl Bank {
         self.rc.accounts.load_by_program_slot(self.slot(), None)
     }
 
-    pub fn get_account_modified_since_parent(
-        &self,
-        pubkey: &Pubkey,
-    ) -> Option<(AccountSharedData, Slot)> {
+    fn get_account_for_incinerator(&self, pubkey: &Pubkey) -> Option<(AccountSharedData, Slot)> {
```
ryoqun, line 4212: - Removed `pub`
- And renamed as it's only used by `incinerator` to prevent this helper fn from accidentally being used outside on-chain.

carllin: This seems like a potentially useful utility function, maybe the more generic name `get_account_modified_since_parent(is_fixed_root: bool)` makes sense?

runtime/src/bank.rs
carllin: Awesome, looks great 😄

carllin:
@ryoqun seems like I can't approve my own PR 😃 , but lgtm!

ryoqun:
> @ryoqun seems like I can't approve my own PR smiley , but lgtm!

woo! here I've finally got the badge! this pr is now in the testing phase. Yeah, I noticed you can't self-approve mine other time.

ryoqun:
> > @ryoqun seems like I can't approve my own PR smiley , but lgtm!
> 
> woo! here I've finally got the badge! this pr is now in the testing phase. Yeah, I noticed you can't self-approve mine other time.

now, testing phase is complete, at least it ran without problems for 24 hours. (although the test validator is interrupted by different reason). I'm merging this soon.

PR #15459, opened 2021-02-21
indexes epoch slots in crds table

behzadnouri:
#### Problem
ClusterInfo::get_epoch_slots_since scans the entire crds table to obtain
epoch-slots inserted since a timestamp:
https://github.com/solana-labs/solana/blob/013daa8f4/core/src/cluster_info.rs#L1245-L1262
The alternative is to index epoch-slots in crds table ordered by their
insert timestamp.

#### Summary of Changes
* Added an index for epoch-slots crds values ordered by insert-timestamp.
* Partially rewrote ClusterSlots update for fewer unlock/lock instances.

### DIFF
--- a/core/src/cluster_slots.rs
+++ b/core/src/cluster_slots.rs
@@ -2,11 +2,15 @@ use crate::{
     cluster_info::ClusterInfo, contact_info::ContactInfo, epoch_slots::EpochSlots,
     serve_repair::RepairType,
 };
+use itertools::Itertools;
 use solana_runtime::{bank_forks::BankForks, epoch_stakes::NodeIdToVoteAccounts};
 use solana_sdk::{clock::Slot, pubkey::Pubkey};
 use std::{
     collections::{BTreeMap, HashMap, HashSet},
-    sync::{Arc, RwLock},
+    sync::{
+        atomic::{AtomicU64, Ordering},
+        Arc, RwLock,
+    },
 };
 
 // Limit the size of cluster-slots map in case
@@ -18,10 +22,9 @@ pub type SlotPubkeys = HashMap<Pubkey, u64>;
 #[derive(Default)]
 pub struct ClusterSlots {
     cluster_slots: RwLock<BTreeMap<Slot, Arc<RwLock<SlotPubkeys>>>>,
-    since: RwLock<Option<u64>>,
+    since: AtomicU64,
     validator_stakes: RwLock<Arc<NodeIdToVoteAccounts>>,
     epoch: RwLock<Option<u64>>,
-    self_id: RwLock<Pubkey>,
 }
 
 impl ClusterSlots {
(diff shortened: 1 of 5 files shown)


behzadnouri:

`cluster_slots_service-timing::update_completed_slots_elapsed`:
btree_index cluster_slots_service-timing update_completed_slots_elapsed_1

core/src/cluster_slots.rs
```diff
@@ -18,34 +22,56 @@ pub type SlotPubkeys = HashMap<Pubkey, u64>;
...
+                epoch_slots
+                    .to_slots(root)
+                    .into_iter()
+                    .filter(|slot| *slot > root)
+                    .zip(std::iter::repeat((epoch_slots.from, stake)))
+            })
+            .into_group_map();
         {
             let mut cluster_slots = self.cluster_slots.write().unwrap();
+            for (slot, node_stakes) in slot_node_stakes {
```
carllin, line 67: Hmm, so I'm worried that if there are a lot of updates here, holding the write lock may block repair which is also consuming this `self.cluster_slots`. 

Maybe we can try:
1)  Grabbing the read lock for the `cluster_slots`, cloning out the existing `Arc<RwLock<SlotPubkeys>>` and modifying those.
2) Only grab the write lock on `cluster_slots` to create + clone the missing set of `slot` keys, drop the lock. Modify the cloned `Arc<RwLock<SlotPubkeys>>`

behzadnouri: makes sense. updated the code.

core/src/cluster_slots.rs
```diff
@@ -62,16 +62,21 @@ impl ClusterSlots {
                     .zip(std::iter::repeat((epoch_slots.from, stake)))
             })
             .into_group_map();
+        let slot_nodes_stakes: Vec<_> = {
+            let mut cluster_slots = self.cluster_slots.write().unwrap();
```
carllin, line 66: hmm here can we grab the read lock here, clone out the ones that are needed, and then only grab the write lock to insert the missing slots? May avoid grabbing the write lock/blocking in a lot of cases.

behzadnouri: The older code was doing something like that:
https://github.com/solana-labs/solana/blob/88f22c360/core/src/cluster_slots.rs#L68-L74
and it had a race condition (fixed in https://github.com/solana-labs/solana/pull/15194)

I think that will make the code more convoluted here, and potentially may actually slow it down if it ends up obtaining the write lock anyways. I would suggest to keep this simple for now until we have a benchmark that demonstrates improvement.

carllin: > The older code was doing something like that:
> https://github.com/solana-labs/solana/blob/88f22c360/core/src/cluster_slots.rs#L68-L74
> and it had a race condition (fixed in #15194)
> 

I see that check needs to check for existence after the write lock is grabbed as well

Then let's add a metric to `RepairStats` in this function call: https://github.com/solana-labs/solana/blob/master/core/src/repair_service.rs#L265, time to grab the read lock in `cluster_slots.compute_weights`. Should give us a benchmark.

behzadnouri: This is `serve_repair-repair-timing::send-repairs-elapsed`:
https://github.com/solana-labs/solana/blob/d5f235d99/core/src/repair_service.rs#L263-L278
which encompasses the call to `cluster_slots.compute_weights`, and shows that effectively there was no change before and after this change.

This benchmarking was done with the earlier version of this pull request which was extending inner hash-maps inside the write lock (so possibly was even worse).

btree_index serve_repair-repair-timing send-repairs-elapsed_1

carllin: thanks!

carllin (approved):

behzadnouri:
**mainnet** improvement with this patch included:

`cluster_slots_service-timing::update_completed_slots_elpased`:
cluster_slots_service-timing update_completed_slots_elpased

PR #15475, opened 2021-02-23
Add max retransmit and shred insert slot

sakridge:
#### Problem

No easy way to tell over RPC how far behind a node might be in processing.

#### Summary of Changes

Add two points of max slots in retransmit and also after shred insert. A node could use this to tell how far behind the latest shreds in the network it is.

Fixes #


mvines:
(Don't forget to add these new methods to `developing/clients/jsonrpc-api.md` as well please, it's a totally manual process)

CriesofCarrots (approved):
r+ jsonrpc docs

PR #15482, opened 2021-02-23
fix: createAccountWithSeed account handling

### LINKED ISSUE #15477: web3: createAccountWithSeed mishandles accounts when basePubkey != fromPubkey
#### Problem
The TransactionInstruction created by `createAccountWithSeed()` never adds the optional 3rd account (the base pubkey) as signer, if it does not equal the from pubkey.
solana-sdk reference: https://github.com/solana-labs/solana/blob/123de5de54140dbe88638201e345b3324ac7a64f/sdk/program/src/system_instruction.rs#L92-L97

#### Proposed Solution
Fix it

CriesofCarrots:
#### Problem
The system-program `createAccountWithSeed` method passes 2 account addresses, regardless of whether the from and base pubkeys are equivalent. If they are not, transactions always fail:
`Error: failed to send transaction: Transaction simulation failed: Error processing Instruction 0: missing required signature for instruction`

#### Summary of Changes
Add optional third account when from and base pubkeys are different

Fixes #15477

### DIFF
--- a/web3.js/src/system-program.js
+++ b/web3.js/src/system-program.js
@@ -745,12 +745,16 @@ export class SystemProgram {
       space: params.space,
       programId: params.programId.toBuffer(),
     });
+    let keys = [
+      {pubkey: params.fromPubkey, isSigner: true, isWritable: true},
+      {pubkey: params.newAccountPubkey, isSigner: false, isWritable: true},
+    ];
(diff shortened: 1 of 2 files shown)


CriesofCarrots:
CI failure due to https://github.com/solana-labs/solana/pull/15447
Fix coming shortly

PR #15484, opened 2021-02-23
web3: Rewrite in TypeScript and update superstruct

jstarry:


web3.js/tsconfig.json
```diff
@@ -0,0 +1,24 @@
+{
```
bartosz-lipinski, line 1: it might be worth removing babel/rollup, and have two typescript files similar to https://github.com/solana-labs/token-list

bartosz-lipinski (commented):
Looks awesome @jstarry this will make development in web3 so much easier

web3.js/tsconfig.json
jstarry: Yeah that would be quite nice, unfortunately we need node polyfills for the browser and afaik typescript's bundler doesn't support that

jstarry:
Closed in favour of https://github.com/solana-labs/solana/pull/15858

PR #15488, opened 2021-02-23
Plumb slot update pubsub notifications

carllin:
#### Summary of Changes

Fixes #

### DIFF
--- a/client/src/rpc_response.rs
+++ b/client/src/rpc_response.rs
@@ -101,6 +101,16 @@ pub struct SlotInfo {
     pub root: Slot,
 }
 
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(tag = "type")]
+pub enum SlotUpdate {
+    OptimisticConfirmation { slot: Slot, timestamp: u64 },
+    FirstShredReceived { slot: Slot, timestamp: u64 },
+    BlockEnd { slot: Slot, timestamp: u64 },
+    BlockComplete { slot: Slot, timestamp: u64 },
+    Frozen { slot: Slot, timestamp: u64 },
+}
+
 #[derive(Serialize, Deserialize, Clone, Debug)]
 #[serde(rename_all = "camelCase", untagged)]
 pub enum RpcSignatureResult {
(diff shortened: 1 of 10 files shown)


jstarry (dismissed):
looks good once ci is fixed

client/src/rpc_response.rs
```diff
@@ -101,6 +101,16 @@ pub struct SlotInfo {
     pub root: Slot,
 }
 
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(rename_all = "camelCase", tag = "type")]
+pub enum SlotUpdate {
+    OptimisticConfirmation { slot: Slot, timestamp: u64 },
```
sakridge, line 107: How about errored blocks? Missing I guess we know if there is just no update for a given slot.

carllin: hmmm yeah if we sent all the roots, then you could tell which ones are missing due to wrong fork/errors. If we want to distinguish between wrong fork/replayed error we could add an extra errored block notification

core/src/rpc_subscriptions.rs
```diff
@@ -387,6 +393,8 @@ struct Subscriptions {
     gossip_program_subscriptions: Arc<RpcProgramSubscriptions>,
     gossip_signature_subscriptions: Arc<RpcSignatureSubscriptions>,
     slot_subscriptions: Arc<RpcSlotSubscriptions>,
+    // TODO: Figure out when to remove a subscriber (some number of slots)
```
jstarry, line 396: this can be removed now :)

jstarry (approved):

core/src/retransmit_stage.rs
```diff
@@ -341,6 +382,16 @@ fn retransmit(
...
             max_slot = max_slot.max(shred_slot);
+
+            if let Some(rpc_subscriptions) = rpc_subscriptions {
+                notify_first_shred_received(
+                    shred_slot,
+                    rpc_subscriptions,
+                    sent_received_slot_notification,
+                    &root_bank,
+                );
+            }
```
behzadnouri, line 393: @carllin Would it be ok if we move this rpc notification to window-service? 
Or is there some reason this notification is sent in retransmit-stage?

PR #15492, opened 2021-02-23
Count if optimistically confirmed slot is already rooted

CriesofCarrots:
#### Problem
The optimistically-confirmed-bank-tracker assumes that an slot for which it's receiving a `BankNotification::OptimisticallyConfirmed` is not yet rooted, but we aren't certain that's true.

#### Summary of Changes
Add a counter to track how frequently slots are rooted (and removed from BankForks) before `BankNotification::OptimisticallyConfirmed` is received.

### DIFF
--- a/core/src/optimistically_confirmed_bank_tracker.rs
+++ b/core/src/optimistically_confirmed_bank_tracker.rs
@@ -127,6 +127,8 @@ impl OptimisticallyConfirmedBankTracker {
                     drop(w_optimistically_confirmed_bank);
                 } else if slot > bank_forks.read().unwrap().root_bank().slot() {
                     pending_optimistically_confirmed_banks.insert(slot);
+                } else {
+                    inc_new_counter_info!("dropped-already-rooted-optimistic-bank-notification", 1);
                 }
             }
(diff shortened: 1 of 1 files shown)


core/src/optimistically_confirmed_bank_tracker.rs
```diff
@@ -127,6 +127,8 @@ impl OptimisticallyConfirmedBankTracker {
                     drop(w_optimistically_confirmed_bank);
                 } else if slot > bank_forks.read().unwrap().root_bank().slot() {
                     pending_optimistically_confirmed_banks.insert(slot);
+                } else {
+                    inc_new_counter_info!("optimistically-conf-bank-already-rooted", 1);
```
carllin, line 131: How about something like `dropped-already-rooted-optimistic-bank-notification`

carllin (approved):

core/src/optimistically_confirmed_bank_tracker.rs
CriesofCarrots: Thx!

carllin (approved):

PR #15495, opened 2021-02-23
check program owners

sakridge:
#### Problem

It's easier to reason about program state if only the program that owns it can operate on it.

#### Summary of Changes

Check the program owners in a few of the builtin programs which don't check it.

Fixes #

### DIFF
--- a/storage-proto/src/convert.rs
+++ b/storage-proto/src/convert.rs
@@ -483,6 +483,9 @@ impl TryFrom<tx_by_addr::TransactionError> for TransactionError {
                     41 => InstructionError::ProgramFailedToCompile,
                     42 => InstructionError::Immutable,
                     43 => InstructionError::IncorrectAuthority,
+                    44 => InstructionError::BorshIoError(String::new()),
+                    45 => InstructionError::AccountNotRentExempt,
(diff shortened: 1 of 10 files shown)


storage-proto/src/convert.rs
```diff
@@ -717,6 +717,9 @@ impl From<TransactionError> for tx_by_addr::TransactionError {
                             InstructionError::AccountNotRentExempt => {
                                 tx_by_addr::InstructionErrorType::AccountNotRentExempt
                             }
+                            InstructionError::InvalidAccountOwner => {
```
CriesofCarrots, line 720: Also needs to be included in https://github.com/solana-labs/solana/blob/cf4e31964b1ed6e94f9abf4303d6816c3e498e39/storage-proto/src/convert.rs#L487-L497

BorshIoError and AccountNotRentExempt are also missing there. Do you mind including those as well?

CriesofCarrots (commented):
r+ cargo fmt and storage-proto/src/convert update

storage-proto/src/convert.rs
sakridge: added, thanks

t-nelson (approved):

PR #15496, opened 2021-02-23
feat: add solana-token-registry to explorer

oJshua:
#### Problem
We've consolidated token registries into one package (solana/token-registry) and need to use this in Explorer.

#### Summary of Changes
Swaps hardcoded token registry for combined package. 

(Pending 0.1.2 release of token-registry)


oJshua:
Still looking into why this is failing CI.

oJshua:
Screenshot from 2021-02-23 19-33-01

@bartosz-lipinski looks like the published npm module is missing some dist files. I can get added to npm and republish.

PR #15497, opened 2021-02-23
Drop write lock on sysvars

sakridge:
#### Problem

Programs can't write to sysvars and taking a write lock on them can make batch sizes smaller than they should be.

#### Summary of Changes

Lower write lock to read lock for sysvars.

Fixes #


sdk/program/src/message.rs
```diff
@@ -335,7 +335,7 @@ impl Message {
         let mut writable_keys = vec![];
         let mut readonly_keys = vec![];
         for (i, key) in self.account_keys.iter().enumerate() {
-            if self.is_writable(i) {
+            if self.is_writable(i) && !crate::sysvar::is_sysvar_id(key) {
```
behzadnouri, line 338: Wouldn't it make sense to reject all together the ones which write-lock sysvar earlier?
e.g. in sanitize or like the ones with duplicate keys here:
https://github.com/solana-labs/solana/blob/34bebb7d0/runtime/src/accounts.rs#L794-L796

sakridge: That's an option, it would be a change of behavior which would break some applications on mainnet.

t-nelson:
Do we still just record TX success/failure as a bool?  Should be fine if so.  Otherwise I'm wondering if this would change the error type in cases where a program attempt to write to a sysvar

sakridge:
> Do we still just record TX success/failure as a bool? Should be fine if so. Otherwise I'm wondering if this would change the error type in cases where a program attempt to write to a sysvar

bool? there is an error code. `TransactionError`.

behzadnouri:
I moved the demote logic to `is_writable`.
Let me know if I should revert.

sakridge:
> I moved the demote logic to `is_writable`.
> Let me know if I should revert.

Ok. Can you plumb the feature gate and also write a test to verify the error behavior is the same before and after this change for the case of referencing and writing to a sysvar?

behzadnouri:
> Ok. Can you plumb the feature gate and also write a test to verify the error behavior is the same before and after this change for the case of referencing and writing to a sysvar? cc @behzadnouri

@sakridge which bank should we use to look up the feature-set?

behzadnouri:
@sakridge  I added a feature gate for the write-lock demotion.

* Moved the logic from `Message::is_writable` back to `Message::get_account_keys_by_lock_type`. Because the other one seems to impact `serialize_instructions`.
* Not sure which bank  to use for looking up `feature-set`. Currently this is using the same bank which calls `unlock_accounts` and `prepare_batch`. In `transaction_status_service.rs` it is the bank received through the queue.

core/src/transaction_status_service.rs
```diff
@@ -99,8 +99,9 @@ impl TransactionStatusService {
                     })
                     .expect("FeeCalculator must exist");
                 let fee = fee_calculator.calculate_fee(transaction.message());
-                let (writable_keys, readonly_keys) =
-                    transaction.message.get_account_keys_by_lock_type();
+                let (writable_keys, readonly_keys) = transaction
+                    .message
+                    .get_account_keys_by_lock_type(bank.demote_sysvar_write_locks());
```
sakridge, line 104: @CriesofCarrots @mvines I think this will affect how it shows in the explorer. Should we show sysvar as demoted to read locked even if it is write in the transaction? or as the original raw write lock?

behzadnouri:
> I think we might want to demote for the builtin keys as well:
> 
> ```
> solana_sdk::native_loader::check_id(owner)
> solana_budget_program::check_id()
> solana_vote_program::check_id()
> solana_stake_program::check_id()
> ..
> anything under `programs/`
> ```

Added the builtin keys. I had to copy over keys because direct references create cyclical build dependency.

core/src/transaction_status_service.rs
CriesofCarrots: My preference would be to show sysvar accounts as they were processed, so demoted to read-locked.

sdk/program/src/message.rs
```diff
@@ -8,10 +8,34 @@ use crate::{
...
+        "ExchangeFaucet11111111111111111111111111111",
+        "FaiLure111111111111111111111111111111111111",
+        "NativeLoader1111111111111111111111111111111",
+        "Noop111111111111111111111111111111111111111",
+        "Stake11111111111111111111111111111111111111",
+        "StakeConfig11111111111111111111111111111111",
+        "Vest111111111111111111111111111111111111111",
+        "Vote111111111111111111111111111111111111111",
+        "ownab1e111111111111111111111111111111111111",
+    ]
```
CriesofCarrots, line 32: Should the SystemProgram be included here?

behzadnouri: added

sdk/program/src/message.rs
```diff
@@ -8,10 +8,37 @@ use crate::{
...
+    short_vec, system_instruction, system_program, sysvar,
 };
 use itertools::Itertools;
-use std::convert::TryFrom;
+use lazy_static::lazy_static;
+use std::convert::{TryFrom, TryInto};
+
+lazy_static! {
+    // Copied keys over since direct references create cyclical dependency.
+    static ref BUILTIN_PROGRAMS_KEYS: [Pubkey; 13] = vec![
```
mvines, line 19: `Feature111111111111111111111111111111111111` can be added in here as well

sdk/program/src/message.rs
```diff
@@ -8,10 +8,34 @@ use crate::{
...
+    short_vec, system_instruction, sysvar,
 };
 use itertools::Itertools;
-use std::convert::TryFrom;
+use lazy_static::lazy_static;
+use std::convert::{TryFrom, TryInto};
+
+lazy_static! {
+    // Copied keys over since direct references create cyclical dependency.
+    static ref BUILTIN_PROGRAMS_KEYS: [Pubkey; 12] = vec![
```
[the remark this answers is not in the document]
behzadnouri: Wouldn't that give us less test coverage on TDS?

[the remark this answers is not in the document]
mvines: I don't think so.  The stale/legacy built-ins are only active in a local cluster, not on devnet/testnet/mainnet.  I'd like to 👢 them all eventually

sdk/program/src/message.rs
```diff
@@ -8,10 +8,37 @@ use crate::{
...
     pubkey::Pubkey,
-    short_vec, system_instruction,
+    short_vec, system_instruction, system_program, sysvar,
 };
 use itertools::Itertools;
-use std::convert::TryFrom;
+use lazy_static::lazy_static;
+use std::convert::{TryFrom, TryInto};
+
+lazy_static! {
```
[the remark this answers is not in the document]
behzadnouri: added a manually written abi-freeze test.
not sure how to do it with existing abi-freeze macros.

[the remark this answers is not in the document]
mvines: That looks perfect, I was thinking briefly here about how to do this with the existing macros and I couldn't think of a nice way either

sdk/program/src/message.rs
```diff
@@ -8,10 +8,38 @@ use crate::{
...
+        "ExchangeFaucet11111111111111111111111111111",
+        "FaiLure111111111111111111111111111111111111",
+        "Feature111111111111111111111111111111111111",
+        "NativeLoader1111111111111111111111111111111",
+        "Noop111111111111111111111111111111111111111",
+        "Stake11111111111111111111111111111111111111",
+        "StakeConfig11111111111111111111111111111111",
+        "Vest111111111111111111111111111111111111111",
+        "Vote111111111111111111111111111111111111111",
+        "ownab1e111111111111111111111111111111111111",
```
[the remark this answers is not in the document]
mvines: Yep we'd need the deprecated loader (`BPFLoader1111111111111111111111111111111111`) as well for backwards compat

sdk/program/src/message.rs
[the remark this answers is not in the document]
ryoqun: hehe, you called me? let me try to use the frozen_abi macro magic. xD

sdk/program/src/message.rs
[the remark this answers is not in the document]
behzadnouri: added bpf loader keys

sdk/program/src/message.rs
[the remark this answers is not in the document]
behzadnouri: thanks @ryoqun
I am leaning on keeping it simple for now, avoiding maintenance cost if this is going to be the only use case.
I appreciate your help though.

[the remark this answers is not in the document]
ryoqun: @behzadnouri no problem. I'll just create a follow up pr for the abi thing with other use cases in the our codebase. I believe surely we have some decent usecases for this. I don't want to block your pr. :)

ryoqun:
@behzadnouri (CC: @sakridge ), I think this pr starts to prohibit this currently-allowed-but-bizarre tx: sending lamports to sysvars.

I think it's ok to prohibit that in the future for the sake of performance. Could you add such a test?

behzadnouri:
> I think this pr starts to prohibit this currently-allowed-but-bizarre tx: sending lamports to sysvars.

@ryoqun do you see any more serious consequences from that or similar transactions? e.g. breaking already deployed programs, or conflicting block producers vs validators?

sakridge:
> @behzadnouri (CC: @sakridge ), I think this pr starts to prohibit this currently-allowed-but-bizarre tx: sending lamports to sysvars.
> 
> I think it's ok to prohibit that in the future for the sake of performance. Could you add such a test?
> 
> (Ah, also please check that capitalization isn't messed up by that, too.)

Yea. I dont' think there are any use-cases of transferring lamports to sysvar accounts and I haven't seen any on-chain. Although, I haven't done any kind of scan for it.

It would be a good idea to have a test to see what happens in this pathway before and after this change in terms of the capitalization and error behavior though. Both of modifying sysvar data and lamports values.

With the feature cutoff I would think there shouldn't be any difference of behavior with block producer and validator since they would be activated at the same time.

behzadnouri:
> Yea. I dont' think there are any use-cases of transferring lamports to sysvar accounts and I haven't seen any on-chain. Although, I haven't done any kind of scan for it.

How can I do this scan?

Theoretically at least, I _think_ it is possible to add lamports to sysvars:
https://github.com/solana-labs/solana/blob/1061d021c/runtime/src/message_processor.rs#L121-L137
note that `is_writable` comes from `Message::is_writable` which this pr is not changing (because it will impact `serialize_instructions`).

and it seems like they are used in capitalization:
https://github.com/solana-labs/solana/blob/1061d021c/runtime/src/accounts_db.rs#L3500-L3507

sakridge:
> > Yea. I dont' think there are any use-cases of transferring lamports to sysvar accounts and I haven't seen any on-chain. Although, I haven't done any kind of scan for it.
> 
> How can I do this scan?

You could run `ledger-tool print` on some recent mainnet ledger and check that post-lamports is not changed for sysvar keys. Or walk the slots with the rpc  `getConfirmedBlock` which has pre- and post- lamport balances. Or the `blockstore.get_confirmed_block` api.

> Theoretically at least, I _think_ it is possible to add lamports to sysvars:
> https://github.com/solana-labs/solana/blob/1061d021c/runtime/src/message_processor.rs#L121-L137
> note that `is_writable` comes from `Message::is_writable` which this pr is not changing (because it will impact `serialize_instructions`).
> 
> and it seems like they are used in capitalization:
> https://github.com/solana-labs/solana/blob/1061d021c/runtime/src/accounts_db.rs#L3500-L3507

That's fine, they just won't be able to be affected after this change.

ryoqun:
@behzadnouri what @sakridge is saying is right. (Thanks for follow-up) Please ping me if you have more questions regarding the special handling of sysvars in bank. I'm mostly in charge of it.

> That's fine, they just won't be able to be affected after this change.

That's true. In fact, I just removed the code: https://github.com/solana-labs/solana/pull/15763 

> @ryoqun do you see any more serious consequences from that or similar transactions? 

What I worried originally about is that a tx transfers lamports and the updated sysvar account is silently discarded. You can check that just running `run.sh` and `solana -ul pay SysvarC1ock11111111111111111111111111111111 1` and see what happens after awhile (so that the next periodic background accounts hashing/capitalization check starts)

sdk/program/src/message.rs
```diff
@@ -345,9 +345,14 @@ impl Message {
         self.program_position(i).is_some()
     }
 
-    pub fn is_writable(&self, i: usize) -> bool {
-        i < (self.header.num_required_signatures - self.header.num_readonly_signed_accounts)
-            as usize
+    pub fn is_writable(&self, i: usize, demote_sysvar_write_locks: bool) -> bool {
```
behzadnouri, line 348: From discord chat, I was assuming you do not want to change `is_writable`. Has that changed?

ryoqun:
> You can check that just running `run.sh` and `solana -ul pay SysvarC1ock11111111111111111111111111111111 1` and see what happens after awhile (so that the next periodic background accounts hashing/capitalization check starts)

@sakridge  @behzadnouri  Well, I found this test procedure doesn't work generally in some sense: you can't observe lamports sent to sysvars. Due to historical reason, sysvar handling a bit odd... I think it just burns sent lamports, remaining to be `1 lamports`. I'll fix this odd behavior anyway.

sdk/program/src/message.rs
sakridge: Yea. I'm thinking we want to change that now.

sakridge: Because if you could change the lamports without the write lock, then multiple threads doing that might produce inconsistent results.

behzadnouri: yes, that makes sense

behzadnouri:
@sakridge I updated this pull request.

* `test_program_bpf_instruction_introspection` was failing and had to be modified in this line:
https://github.com/solana-labs/solana/blob/664ed7652/programs/bpf/tests/programs.rs#L1326
because with the new logic, this branch:
https://github.com/solana-labs/solana/blob/664ed7652/runtime/src/accounts.rs#L206-L208
is no longer taken.

* A number of places which have to do with serialization/formatting I have hard coded `demote_sysvar_write_locks` to `false` not to change the serialization. In some cases having a hard coded value is inevitable. For example, the `Display` or `Debug` trait implementations cannot take an extra argument. These places are marked with `TODO` in the commit to confirm.

* My suggestion is to also  hard code `demote_sysvar_write_locks` to `false` in `Message::serialize_instructions`. So that this demotion logic is only a runtime thing for execution and does not change the serializations. Also marked with `TODO` in the commit.

sakridge:
> @sakridge I updated this pull request.
> 
> * `test_program_bpf_instruction_introspection` was failing and had to be modified in this line:
>   https://github.com/solana-labs/solana/blob/664ed7652/programs/bpf/tests/programs.rs#L1326
>   because with the new logic, this branch:
>   https://github.com/solana-labs/solana/blob/664ed7652/runtime/src/accounts.rs#L206-L208
>   is no longer taken.
> * A number of places which have to do with serialization/formatting I have hard coded `demote_sysvar_write_locks` to `false` not to change the serialization. In some cases having a hard coded value is inevitable. For example, the `Display` or `Debug` trait implementations cannot take an extra argument. These places are marked with `TODO` in the commit to confirm.
> * My suggestion is to also  hard code `demote_sysvar_write_locks` to `false` in `Message::serialize_instructions`. So that this demotion logic is only a runtime thing for execution and does not change the serializations. Also marked with `TODO` in the commit.

Ok. I think that all sounds good. Thanks!

behzadnouri:
With this last commit, `demote_sysvar_write_locks` is now hard coded to `false` for serializations/encoding methods so not to change the serialized messages (so the write lock demotion is only a runtime thing for tx execution).

sakridge:
> With this last commit, `demote_sysvar_write_locks` is now hard coded to `false` for serializations/encoding methods so not to change the serialized messages (so the write lock demotion is only a runtime thing for tx execution).

From this, I think the preference was to show it demoted and as-processed in the UI: https://github.com/solana-labs/solana/pull/15497/files#r590501063

So I think we should hard-code the display cases to `demote=true`

For the serialization, I would go with just passing `demote_sysvar_write_locks` as-is. You're worried about breaking apps on-chain?

sakridge:
Here's all the usage of the Sysvar1nstructions key.

https://explorer.solana.com/address/Sysvar1nstructions1111111111111111111111111

Looks like wormhole transactions and doesn't seem to pass any other sysvar or builtin keys as writable.

behzadnouri:
> From this, I think the preference was to show it demoted and as-processed in the UI: https://github.com/solana-labs/solana/pull/15497/files#r590501063
> 
> So I think we should hard-code the display cases to `demote_sysvar_write_locks=true`
> 
> For the serialization, I would go with just passing `demote_sysvar_write_locks` as-is. You're worried about breaking apps on-chain?

I reverted last commit adding back the argument to `serialize_instructions`.

Also, changed the hard coded values to `demote_sysvar_write_locks=true`. There are 2 places:
https://github.com/solana-labs/solana/blob/66c42f62d/transaction-status/src/parse_accounts.rs#L16
https://github.com/solana-labs/solana/blob/66c42f62d/cli-output/src/display.rs#L136

Not sure if this causes breakage though. Hard coding to `true` feels wrong given that the feature is not enabled yet, and there will not be any feature gating.

sakridge:
> > From this, I think the preference was to show it demoted and as-processed in the UI: https://github.com/solana-labs/solana/pull/15497/files#r590501063
> > So I think we should hard-code the display cases to `demote_sysvar_write_locks=true`
> > For the serialization, I would go with just passing `demote_sysvar_write_locks` as-is. You're worried about breaking apps on-chain?
> 
> I reverted last commit adding back the argument to `serialize_instructions`.
> 
> Also, changed the hard coded values to `demote_sysvar_write_locks=true`. There are 2 places:
> https://github.com/solana-labs/solana/blob/66c42f62d/transaction-status/src/parse_accounts.rs#L16
> https://github.com/solana-labs/solana/blob/66c42f62d/cli-output/src/display.rs#L136
> 
> Not sure if this causes breakage though. Hard coding to `true` feels wrong given that the feature is not enabled yet, and there will not be any feature gating.

Well with no gate, it's going to be 'wrong' either way in some window of time. Seems better to be wrong toward the long term status of these given the feature will be enabled eventually. The keys are effectively read-only anyway today because nobody can sign with the Sysvar111 key.

If you hard code to demote=false what do you do when you enable the feature? push another change to update to hard-code true?

behzadnouri:
> If you hard code to demote=false what do you do when you enable the feature? push another change to update to hard-code true?

I was thinking of not changing it (i.e. leave it false).

sakridge (commented):
ok, lgtm. reviewers?

cli-output/src/display.rs
```diff
@@ -133,7 +133,7 @@ fn format_account_mode(message: &Message, index: usize) -> String {
         } else {
             "-"
         },
-        if message.is_writable(index) {
+        if message.is_writable(index, /*demote_sysvar_write_locks=*/ true) {
```
behzadnouri, line 136: Note: One of the 2 hard-coded values. This one is invoked from `Display` implementation (and maybe other places). So having a hard coded value somewhere is inevitable.

transaction-status/src/parse_accounts.rs
```diff
@@ -13,7 +13,7 @@ pub fn parse_accounts(message: &Message) -> Vec<ParsedAccount> {
     for (i, account_key) in message.account_keys.iter().enumerate() {
         accounts.push(ParsedAccount {
             pubkey: account_key.to_string(),
-            writable: message.is_writable(i),
+            writable: message.is_writable(i, /*demote_sysvar_write_locks=*/ true),
```
behzadnouri, line 16: Note: The other of the 2 hardcoded places. This one is used from `ledger-tool/src/bigtable.rs`, `core/src/rpc.rs`, `cli/src/cli.rs` and possibly elsewhere.

behzadnouri (commented):
Noted places where the `demote_sysvar_write_locks` is hard coded for the reviewers to also take a look:
https://github.com/solana-labs/solana/blob/66c42f62d/transaction-status/src/parse_accounts.rs#L16
https://github.com/solana-labs/solana/blob/66c42f62d/cli-output/src/display.rs#L136

CriesofCarrots (approved):
Lgtm. I'm comfortable with the 2 hardcoded `demote_sysvar_write_locks`. It seems like the upshot is that when parsed, some old transactions may appear different than how they were actually processed, but the unparsed message is available for anyone who needs it.

sdk/program/src/message.rs
```diff
@@ -326,23 +347,31 @@ impl Message {
...
             as usize
             || (i >= self.header.num_required_signatures as usize
                 && i < self.account_keys.len()
-                    - self.header.num_readonly_unsigned_accounts as usize)
+                    - self.header.num_readonly_unsigned_accounts as usize))
+            && !{
+                let key = self.account_keys[i];
+                demote_sysvar_write_locks
+                    && (sysvar::is_sysvar_id(&key) || BUILTIN_PROGRAMS_KEYS.contains(&key))
+            }
```
ryoqun, line 360: @oJshua txes with writable flag to Sysvars will be demoted as if not-writable. Maybe, explorer isn't affected because it's not parsing txes, just displaying whatever returneed from rpcs?

ryoqun:
@behzadnouri iirc, we opted to demote (not reject) the affected txes for compatibility, right? Could you add this information into the pr description for future reference?

sdk/program/src/message.rs
```diff
@@ -326,23 +347,31 @@ impl Message {
...
-        i < (self.header.num_required_signatures - self.header.num_readonly_signed_accounts)
+    pub fn is_writable(&self, i: usize, demote_sysvar_write_locks: bool) -> bool {
+        (i < (self.header.num_required_signatures - self.header.num_readonly_signed_accounts)
             as usize
             || (i >= self.header.num_required_signatures as usize
                 && i < self.account_keys.len()
-                    - self.header.num_readonly_unsigned_accounts as usize)
+                    - self.header.num_readonly_unsigned_accounts as usize))
+            && !{
+                let key = self.account_keys[i];
```
ryoqun, line 357: I think this indexing is free of out-of-bound access, because `i` should be within `self.account_keys.len()` always.

transaction-status/src/parse_accounts.rs
ryoqun: @jstarry maybe, this means this recently-added useful stats on explorer (https://explorer.solana.com/block/71320343/accounts) starts to reflect the demoted definition as soon as deployed without feature gate? I think it's ok, just checking :)

sdk/benches/serialize_instructions.rs
```diff
@@ -27,7 +27,9 @@ fn bench_manual_instruction_serialize(b: &mut Bencher) {
     let instructions = make_instructions();
     let message = Message::new(&instructions, None);
     b.iter(|| {
-        test::black_box(message.serialize_instructions());
+        test::black_box(message.serialize_instructions(
```
ryoqun, line 30: maybe negligible perf impact? :)

ryoqun (dismissed):
lgtm; carefully-done patch to change such a fundamental definition of the writable flag. :)

behzadnouri:
> @behzadnouri iirc, we opted to demote (not reject) the affected txes for compatibility, right? Could you add this information into the pr description for future reference?

yeah, it might break some applications on mainnet: https://github.com/solana-labs/solana/pull/15497#discussion_r587694534

PR #15501, opened 2021-02-23
gracefully handle vote account without authorized voter

jackcmay:
#### Problem

Epoch stakes doesn't like a vote account without an authorized voter

#### Summary of Changes

Handle it gracefully

Fixes #


CriesofCarrots (approved):
lgtm

PR #15503, opened 2021-02-24
Check vote account initialization

CriesofCarrots:
#### Problem
Epoch stakes only needs to know about properly sized, initialized Vote accounts

#### Summary of Changes
- Check vote account size on initialization
- Check that vote account is initialized -- ie. non-empty and the correct size -- before it is included in cache. The size check should be redundant, given the vote-processor change.

### DIFF
--- a/programs/vote/src/vote_state/mod.rs
+++ b/programs/vote/src/vote_state/mod.rs
@@ -684,7 +684,11 @@ pub fn initialize_account<S: std::hash::BuildHasher>(
     vote_init: &VoteInit,
     signers: &HashSet<Pubkey, S>,
     clock: &Clock,
+    check_data_size: bool,
 ) -> Result<(), InstructionError> {
+    if check_data_size && vote_account.data_len()? != VoteState::size_of() {
+        return Err(InstructionError::InvalidAccountData);
+    }
     let versioned = State::<VoteStateVersions>::state(vote_account)?;
 
     if !versioned.is_uninitialized() {
@@ -812,6 +816,7 @@ mod tests {
             },
             &signers,
             &Clock::default(),
+            true,
         );
         assert_eq!(res, Err(InstructionError::MissingRequiredSignature));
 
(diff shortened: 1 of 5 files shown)


programs/vote/src/vote_state/mod.rs
```diff
@@ -569,6 +572,12 @@ impl VoteState {
         self.last_timestamp = BlockTimestamp { slot, timestamp };
         Ok(())
     }
+
+    pub fn is_uninitialized_no_deser(data: &[u8]) -> bool {
+        const VERSION_OFFSET: usize = 4;
+        &data[VERSION_OFFSET..VERSION_OFFSET + DEFAULT_PRIOR_VOTERS_OFFSET]
```
jackcmay, line 578: Is there a chance that the data size could be less than the indices here?

runtime/src/stakes.rs
```diff
@@ -116,16 +117,19 @@ impl Stakes {
...
             // delegated vote pubkey)
             let old = self.vote_accounts.remove(pubkey);
-            // when account is removed (lamports == 0), don't readd so that given
-            // `pubkey` can be used for any owner in the future, while not
-            // affecting Stakes.
-            if account.lamports != 0 {
+            // when account is removed (lamports == 0 or data uninitialized), don't read so that
+            // given `pubkey` can be used for any owner in the future, while not affecting Stakes.
+            if account.lamports != 0
+                && !(check_vote_init && (account.data.len() != VoteState::size_of())
```
jackcmay, line 130: Maybe move this size check into `is_unitialized_no_deser`?

programs/vote/src/vote_state/mod.rs
CriesofCarrots: Not with the size check coming first ;)

runtime/src/stakes.rs
CriesofCarrots: Good idea; makes the indexing more clear

CriesofCarrots: Adding a unit test for `is_unitialized_no_deser` also

CriesofCarrots: Updated

jackcmay (approved):

PR #15506, opened 2021-02-24
Fix received notifications for gossip signature subscriptions

jstarry:
#### Problem
Gossip signature subscriptions do not receive received notifications because the gossip signature subscriptions map is not checked.


carllin (approved):
Thanks for taking the fix!

ryoqun:
@carllin @jstarry Hi, I'm now wearing rel-notes hat. :) How does relates to one of our ~500 open issues?

jstarry:
Hi! This fixes the behaviour of an undocumented RPC pubsub api option. I think it can be excluded from the release notes

PR #15518, opened 2021-02-24
More debug on shrink and flush

sakridge:
#### Problem

Not clear how slots are getting flushed and shrinking.

#### Summary of Changes

Add prints on shrink and flush.

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -859,6 +859,15 @@ impl PurgeStats {
     }
 }
 
+#[derive(Debug)]
+struct FlushStats {
+    slot: Slot,
+    num_flushed: usize,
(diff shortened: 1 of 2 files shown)


runtime/src/accounts_db.rs
```diff
@@ -3254,7 +3257,12 @@ impl AccountsDb {
             true
         } else {
             false
-        }
+        };
+        info!(
```
carllin, line 3261: I think it would also be good to log:

1)  the `requested_flush_root` and `max_clean_root` in  `flush_rooted_accounts_cache`

2) Both `max_clean_root` in `clean_accounts()` here: https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_db.rs#L1376-L1377

Issue #14509, opened 2021-01-09
Cli: audit OutputFormat handling


CriesofCarrots:
#### Problem
Not all cli commands have been updated to produce properly formatted output.
eg. `solana leader-schedule --output json` doesn't output json, and that's a case where json could be really useful.

#### Proposed Solution
Audit `solana-cli` subcommands and implement OutputFormat handling where needed (esp for requests that return a lot of data or would likely be consumed programmatically), incl:
- [x] solana leader-schedule
- [x] solana inflation
- [x] solana block
- [x] solana confirm / decode-transaction
- [x] solana transaction-history

CriesofCarrots:
Reported on discord:
the recent versions 1.4.22 and 1.5.2 write error messages into the json output of solana block-production like Note: Requested start slot was... , this causes errors with json parsing.

PR #15528, opened 2021-02-25
Implement OutputFormat for confirm in Cli and ledger-tool bigtable

CriesofCarrots:
#### Problem
Not all cli commands have been updated to produce properly formatted output.
Also, consumers of solana-ledger-tool bigtable confirm should be consistent wrt json output

#### Summary of Changes
- Implement OutputFormat for solana confirm and solana decode-transaction
- Use in solana-ledger-tool bigtable confirm

`transaction-history` still needs handling for both cli and ledger-tool bigtable, but will save for a separate PR

Toward #14509

### DIFF
--- a/cli-output/src/cli_output.rs
+++ b/cli-output/src/cli_output.rs
@@ -29,7 +29,9 @@ use {
         transaction::Transaction,
     },
     solana_stake_program::stake_state::{Authorized, Lockup},
-    solana_transaction_status::EncodedConfirmedBlock,
+    solana_transaction_status::{
+        EncodedConfirmedBlock, EncodedTransaction, UiTransactionStatusMeta,
+    },
     solana_vote_program::{
         authorized_voters::AuthorizedVoters,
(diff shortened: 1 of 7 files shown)


CriesofCarrots:
@t-nelson this isn't as clean as I'd like. Let me know if you have any ideas, particularly around CliTransaction::decoded_transaction. The tricky bit is that I want the json to return a Json-encoded transaction to match Rpc, but writeln_transaction takes a decoded `Transaction` (which can't be derived from a Json-encoded one).

t-nelson (approved):
> @t-nelson this isn't as clean on the calling side as I'd like. Let me know if you have any ideas, particularly around CliTransaction::decoded_transaction. The tricky bit is that I want the json to return a Json-encoded transaction to match Rpc, but writeln_transaction takes a decoded Transaction (which can't be derived from a Json-encoded one).

I'm not coming up with anything less intrusive off hand.  A custom Serialize impl for `CliTransaction` seems like the shortest path, but that path isn't all that short. Getting back a `Transaction` from an `EncodedTransaction::Json()` in a backward compatible way looks like it'd also be a fair amount of churn, though ultimately useful.

Let's just go with this for now.  LGTM! :wink:

PR #15532, opened 2021-02-25
Fix root scan in ledger tool

carllin:
#### Problem
Ledger tool root search starts at slot 0 which will panic for any ledgers that don't contain slot 0

#### Summary of Changes
Pass a starting slot
Fixes #


CriesofCarrots:
Oh thank you! I hacked together this same change last week; should have cleaned up!

PR #15534, opened 2021-02-25
Fix finalize_dead_slot_removal() of cached slots decrementing refcount

carllin:
#### Problem
finalize_dead_slot_removal() decrements the ref count of deleted account for cached slots, even though the ref count was never incremented for accounts in the cache anyways.

#### Summary of Changes
Don't decrement the ref count when purging dead cached slots

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -3946,17 +3946,19 @@ impl AccountsDb {
         &'a self,
         dead_slots_iter: impl Iterator<Item = &'a Slot> + Clone,
         purged_slot_pubkeys: HashSet<(Slot, Pubkey)>,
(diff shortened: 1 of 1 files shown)


runtime/src/accounts_db.rs
```diff
@@ -3946,17 +3946,19 @@ impl AccountsDb {
         &'a self,
         dead_slots_iter: impl Iterator<Item = &'a Slot> + Clone,
         purged_slot_pubkeys: HashSet<(Slot, Pubkey)>,
-        mut purged_account_slots: Option<&mut AccountSlots>,
+        // If the slot being are not cached, return
+        is_cached: Option<&mut AccountSlots>,
```
ryoqun, line 3950: nits: keep original var name? This paused me a whil because the type isn't bool.

PR #15535, opened 2021-02-25
Revert "v1.5: Make UiTokenAmount::ui_amount a String (#15447)"

ryoqun:
Reverts solana-labs/solana#15472

This is causing issues across the board.. Maybe, we can't change rpc field's type, expecting at-once micrale over-night transition.. `\(o_o)/`:

- explorer
- `solana`
- `spl-token`
- web3.js (I haven't checked; I'm pretty sure this is affected as well).

x

- mainnet-beta
- testnet
- devnet

### DIFF
--- a/scripts/build-downstream-projects.sh
+++ b/scripts/build-downstream-projects.sh
@@ -105,5 +105,5 @@ EOF
 
 
 _ example_helloworld
-# _ spl
+_ spl
 _ serum_dex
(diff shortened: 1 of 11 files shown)


ryoqun:
ref: https://github.com/solana-labs/solana-program-library/issues/1341

scripts/build-downstream-projects.sh
```diff
@@ -105,5 +105,5 @@ EOF
 
 
 _ example_helloworld
-# _ spl
+_ spl
```
ryoqun, line 108: :eyes: no wonder pure me didn't notice the expected spl cli breakages even if I waited for my last. here was dragons. :)

PR #15539, opened 2021-02-25
limits number of unique pubkeys in the crds table

behzadnouri:
#### Summary of Changes

Fixes #

### DIFF
--- a/core/src/cluster_info.rs
+++ b/core/src/cluster_info.rs
@@ -114,6 +114,8 @@ pub const DEFAULT_CONTACT_DEBUG_INTERVAL_MILLIS: u64 = 10_000;
 pub const DEFAULT_CONTACT_SAVE_INTERVAL_MILLIS: u64 = 60_000;
 /// Minimum serialized size of a Protocol::PullResponse packet.
 const PULL_RESPONSE_MIN_SERIALIZED_SIZE: usize = 161;
+// Limit number of unique pubkeys in the crds table.
+const CRDS_UNIQUE_PUBKEY_CAPACITY: usize = 1024;
 
 #[derive(Debug, PartialEq, Eq)]
 pub enum ClusterInfoError {
@@ -601,16 +603,6 @@ impl ClusterInfo {
 
     pub fn set_contact_debug_interval(&mut self, new: u64) {
         self.contact_debug_interval = new;
-    }
-
-    pub fn update_contact_info<F>(&self, modify: F)
-    where
-        F: FnOnce(&mut ContactInfo),
-    {
-        let my_id = self.id();
-        modify(&mut self.my_contact_info.write().unwrap());
-        assert_eq!(self.my_contact_info.read().unwrap().id, my_id);
-        self.insert_self()
     }
 
     fn push_self(
@@ -1726,6 +1718,7 @@ impl ClusterInfo {
         stakes: &HashMap<Pubkey, u64>,
         generate_pull_requests: bool,
     ) -> Vec<(SocketAddr, Protocol)> {
+        self.trim_crds_table(CRDS_UNIQUE_PUBKEY_CAPACITY, &stakes);
         let mut pulls: Vec<_> = if generate_pull_requests {
             self.new_pull_requests(&thread_pool, gossip_validators, stakes)
         } else {
@@ -1837,6 +1830,31 @@ impl ClusterInfo {
             .time_gossip_write_lock("purge", &self.stats.purge)
             .purge(thread_pool, timestamp(), &timeouts);
         inc_new_counter_info!("cluster_info-purge-count", num_purged);
+    }
+
+    // Trims the CRDS table by dropping all values associated with the pubkeys
+    // with the lowest stake, so that the number of unique pubkeys are bounded.
+    fn trim_crds_table(&self, cap: usize, stakes: &HashMap<Pubkey, u64>) {
+        if !self.gossip.read().unwrap().crds.should_trim(cap) {
+            return;
+        }
+        let keep: Vec<_> = self
+            .entrypoints
+            .read()
+            .unwrap()
+            .iter()
+            .map(|k| k.id)
+            .chain(std::iter::once(self.id))
+            .collect();
+        let mut gossip = self.gossip.write().unwrap();
+        match gossip.crds.trim(cap, &keep, stakes) {
+            Err(err) => error!("crds table trim failed: {:?}", err),
+            Ok(purged_values) => gossip.pull.purged_values.extend(
+                purged_values
+                    .into_iter()
+                    .map(|v| (v.value_hash, v.local_timestamp)),
+            ),
+        }
     }
 
     /// randomly pick a node and ask them for updates asynchronously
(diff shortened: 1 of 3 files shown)


leoluk:
What about unstaked nodes? Those are quite important to the network, too

mvines:
> What about unstaked RPC nodes? Those are quite important to the network, too

Yeah good point.  I've been wondering if it makes sense to strongly recommend that RPC nodes have at least a couple SOL staked to them.   Not enough to get in the leader schedule but enough so that they actually get prioritized over the rest of the 0 nodes

core/src/cluster_info.rs
```diff
@@ -114,6 +114,8 @@ pub const DEFAULT_CONTACT_DEBUG_INTERVAL_MILLIS: u64 = 10_000;
 pub const DEFAULT_CONTACT_SAVE_INTERVAL_MILLIS: u64 = 60_000;
 /// Minimum serialized size of a Protocol::PullResponse packet.
 const PULL_RESPONSE_MIN_SERIALIZED_SIZE: usize = 167;
+// Limit number of unique pubkeys in the crds table.
+const CRDS_UNIQUE_PUBKEY_CAPACITY: usize = 1024;
```
sakridge, line 118: 1k seems really low. I would expect like 10-20k is probably ok?

behzadnouri:
> > What about unstaked RPC nodes? Those are quite important to the network, too
> 
> Yeah good point. I've been wondering if it makes sense to strongly recommend that RPC nodes have at least a couple SOL staked to them. Not enough to get in the leader schedule but enough so that they actually get prioritized over the rest of the 0 nodes

We may do some whitelisting to. The commit includes a `keep: &[Pubkey]` for pubkeys to never drop. Currently this is populated with the entrypoints and self-pubkey. But potentially can be expanded.

leoluk:
> > What about unstaked RPC nodes? Those are quite important to the network, too
> 
> Yeah good point. I've been wondering if it makes sense to strongly recommend that RPC nodes have at least a couple SOL staked to them. Not enough to get in the leader schedule but enough so that they actually get prioritized over the rest of the 0 nodes

We're running a whole bunch of private RPC-only nodes these days and having to stake them (and manage the keys) would be a big headache... not that I have a better idea on solving this

sakridge:
> > > What about unstaked RPC nodes? Those are quite important to the network, too
> > 
> > 
> > Yeah good point. I've been wondering if it makes sense to strongly recommend that RPC nodes have at least a couple SOL staked to them. Not enough to get in the leader schedule but enough so that they actually get prioritized over the rest of the 0 nodes
> 
> We're running a whole bunch of private RPC-only nodes these days and having to stake them (and manage the keys) would be a big headache... not that I have a better idea on solving this

Is it really a problem? the infos are not that expensive, I think we can make the limit high enough that unless we are in constant DOS from nodes spamming ContactInfo it will be ok.

CherryWorm:
The limit can't be too high as the DOS primarily depends on the contact infos being in the table long enough to get pushed out via gossip (which is not long). They get evicted after 15 seconds anyways. II'm not sure what the number has to be, but 20k already seems very high, that should probably be tested first.

leoluk:
> Is it really a problem? the infos are not that expensive, I think we can make the limit high enough that unless we are in constant DOS from nodes spamming ContactInfo it will be ok.

But not doing it would leave the network latently vulnerable to such DoS attacks, so I would assume that anyone running an important RPC node would want to stake it?

behzadnouri:
I do not think `ContactInfo`s are special. Any type can cause problems. The key is that the upper bound on the number of values is proportional to the number of unique pubkeys.

So, the limit is on the number of unique pubkeys. With each pubkey there can be 830 values associated. So with 1000 limit on unique pubkeys, potentially the table can be as large as 830k records.

sakridge:
> I do not think `ContactInfo`s are special. Any type can cause problems. The key is that the upper bound on the number of values is proportional to the number of unique pubkeys.
> https://github.com/solana-labs/solana/blob/ebd43938a/core/src/crds_value.rs#L435-L446
> 
> So, the limit is on the number of unique pubkeys. With each pubkey there can be 830 values associated. So with 1000 limit on unique pubkeys, potentially the table can be as large as 830k records.

Well for unstaked nodes, we don't need to propagate votes and the other labels.

behzadnouri:
> Well for unstaked nodes, we don't need to propagate votes and the other labels.

That might be a good a idea to implement. But currently the gossip propagation code does not look at the stakes. So this will be a new logic in the code.

On the other hand it only takes 1 sol to work around that. So might not be enough.

mvines:
> > Well for unstaked nodes, we don't need to propagate votes and the other labels.
> 
> That might be a good a idea to implement. But currently the gossip propagation code does not look at the stakes. So this will be a new logic in the code.

Oh yeah that'd be really nice.  There's now probably hundreds of RPC nodes on mainnet-beta

sakridge:
> > Well for unstaked nodes, we don't need to propagate votes and the other labels.
> 
> That might be a good a idea to implement. But currently the gossip propagation code does not look at the stakes. So this will be a new logic in the code.
> 
> On the other hand it only takes 1 sol to work around that. So might not be enough.

True, but it limits the attack surface, you cannot stake/unstake quickly. I think for the stalling issue actually we might need to do something to limit the outgoing unstaked entries even further and then as came up in the thread have the whitelist for like a staked node + unstaked RPC setups. Since each node doesn't really know what the rest of the network is sending, then the limits to stop a distributed reflection type attack has to be pretty low.

behzadnouri:
> > > Well for unstaked nodes, we don't need to propagate votes and the other labels.
> > 
> > 
> > That might be a good a idea to implement. But currently the gossip propagation code does not look at the stakes. So this will be a new logic in the code.
> 
> Oh yeah that'd be really nice. There's now probably hundreds of RPC nodes on mainnet-beta

sent out https://github.com/solana-labs/solana/pull/15561

core/src/cluster_info.rs
behzadnouri: Increased it to 4k, which is ~6x current mainnet cluster (~4x current tds).
This is the number of unique pubkeys and the size of table itself is 830x of this value. So 10k might be a bit stretch performance wise. We can update this number as the code is optimized further to handle larger cluster.

aeyakovenko:
@behzadnouri once we have this, we can start optimizing for stake weighted propagation in gossip as well.

mvines:
> we might also want to take into consideration the vote frequency, basically votes burn sol, which pays for message access on the network. That might be a stronger spam filter then just stake weighted filters

Oh I like this.   We've been talking about giving RPC backend nodes a pinch of stake for network access, but that's annoying to manage.  Easier to just get all the RPC backends to vote with no delegated stake

sakridge:
> > we might also want to take into consideration the vote frequency, basically votes burn sol, which pays for message access on the network. That might be a stronger spam filter then just stake weighted filters
> 
> Oh I like this. We've been talking about giving RPC backend nodes a pinch of stake for network access, but that's annoying to manage. Easier to just get all the RPC backends to vote with no delegated stake

sounds like a nightmare for block sizes...

mvines:
> sounds like a nightmare for block sizes...

But if we can't deal with a couple thousand votes per second how's 50k TPS going to ever work?

sakridge (approved):
lgtm

PR #15540, opened 2021-02-25
feat: support blocktimes on confirmed blocks

oJshua:
#### Problem
ConfirmedBlock types do not have blockTime, which would be helpful now that these are available. 

#### Summary of Changes
Add blockTime to ConfirmedBlock.

### DIFF
--- a/web3.js/module.d.ts
+++ b/web3.js/module.d.ts
@@ -200,6 +200,7 @@ declare module '@solana/web3.js' {
       postBalance: number | null;
       rewardType: string | null;
     }>;
+    blockTime: number | null;
   };
 
   export type PerfSample = {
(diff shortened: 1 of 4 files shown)


oJshua:
@CriesofCarrots I'm trying to decide if I should include the undefined case for blockTime like we're doing elsewhere (ConfirmedTransaction). These will be going away in the near future.

CriesofCarrots:
> @CriesofCarrots I'm trying to decide if I should include the undefined case for blockTime like we're doing elsewhere (ConfirmedTransaction). These will be going away in the near future.

Since the latest v1.4 also supports ConfirmedBlock::block_time, I am okay skipping the undefined case

CriesofCarrots (dismissed):

PR #15543, opened 2021-02-25
validator: Add wait-for-restart-window subcommand

mvines:
The `wait-for-restart-window` subcommand will connect locally to your validator over RPC, and when the node is
1. healthy,
2. in a time window when it won't be leader (default is 10 minutes, can be specified on the command-line),
3. produced a new snapshot, and 
4. there's no more than 5% cluster delinquency 
 
the command will exit with 0 and you can restart your validator using whatever method is preferred.

----
```
$ solana-validator wait-for-restart-window --help
solana-validator-wait-for-restart-window 
Monitor the validator for a good time to restart

USAGE:
    solana-validator --ledger <DIR> wait-for-restart-window [min_idle_time_in_minutes]

FLAGS:
    -h, --help    Prints help information

ARGS:
    <min_idle_time_in_minutes>    Minimum time that the validator should not be leader [default: 10]

Note: If this command exits with a non-zero status then this is not a good time for a restart

### DIFF
--- a/validator/src/dashboard.rs
+++ b/validator/src/dashboard.rs
@@ -1,7 +1,10 @@
 use {
-    crate::{get_validator_rpc_addr, get_validator_start_time},
+    crate::{
+        get_validator_rpc_addr, get_validator_start_time, new_spinner_progress_bar,
+        println_name_value,
+    },
     console::style,
-    indicatif::{ProgressBar, ProgressDrawTarget, ProgressStyle},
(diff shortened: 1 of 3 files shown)


mvines:
Oh!  I can add a delinquent stake check in here too, no restart if more than 5% offline

mvines:
Added delinquency check, hard coded to 5% for now

PR #15566, opened 2021-02-26
coalesces vote packets into one Packets

behzadnouri:
#### Problem
banking-stage is iterating over individual vote packets:
https://github.com/solana-labs/solana/blob/d47f1fae4/core/src/banking_stage.rs#L1045-L1048
which breaks the windowing logic in vote-listener:
https://github.com/solana-labs/solana/blob/d47f1fae4/core/src/cluster_info_vote_listener.rs#L395-L409

#### Summary of Changes
coalesce vote packets into one Packets


sakridge (approved):
lgtm

PR #15567, opened 2021-02-26
ledger-tool: Trivialize creating a snapshot at the current root

mvines:
It's currently too hard to create a new snapshot at the node's current root.   Add `ROOT` keyword to make this trivial:
```
$ solana-ledger-tool -l ledger/ create-snapshot ROOT
```


mvines:
Mergify?

PR #15578, opened 2021-02-27
Fix blockstore processor error severity

sakridge:
#### Problem

Blockstore processor error reporting logic in replay stage is inverted. Reporting InvalidTickCount as bad and others as not, when the reverse should be the case.

#### Summary of Changes

Fix the logic and split InvalidTickCount into two cases, too many ticks which is unexpected and too few ticks which is relatively expected.

Fixes #


carllin (approved):

carllin:
Good catch!

PR #15582, opened 2021-02-28
removes VerifiedVotePackets and stops writing to Bank::last_vote_sync

behzadnouri:
#### Problem
update_version and VerifiedVotePackets are both local variables to
bank_send_loop, implementing an ordinal counter, not meaningful outside
the function:
https://github.com/solana-labs/solana/blob/21e08b5b2/core/src/cluster_info_vote_listener.rs#L372-L374
So writing and reading last_version to Bank::last_vote_sync is
redundant.

If more than one thread runs bank_send_loop, then they should use crds
values timestamp instead of a local counter each (which will be a race
condition in that case).

#### Summary of Changes
removed VerifiedVotePackets and stopped writing to Bank::last_vote_sync.


sakridge (approved):
Looks equivalent to me, I'll let @carllin comment on the history of why that was shared if there was any reason to have it on the bank.

carllin:
So I believe the purpose was to ensure votes were retried even if we were leaders on different forks. So if the slots looks like

    /----1---2 (leader)
0
   \---3 (leader) ---4 (leader) ---5 (leader)

Then if we ingested votes for slot 2, then those should also be included in slot 3. This is because the vote sync was only inherited from parent banks, so an update in slot 2 on a different would not be seen on slot 3.

PR #15584, opened 2021-03-01
fix: secp256k1 instruction should accept 64 byte public key

jstarry:
#### Problem
web3 secp256k1 instruction api expects a 65 byte secp256k1 public key which is an implementation detail of how bitcoin keys are used (leading byte indicates compressed / uncompressed). Since our implementation is for verifying eth address signatures, we should use 64 byte public keys directly

#### Summary of Changes
- Use 64 byte public keys

Fixes https://github.com/solana-labs/solana-web3.js/issues/1012


oJshua (approved):

oJshua (approved):
Makes sense going to 64!

PR #15607, opened 2021-03-01
Skip leader slots until a vote lands

sakridge:
#### Problem

Nodes starting up after clearing their ledger can double sign for a slot when they start their leader slot again and are not aware they already produced a block for that slot.

#### Summary of Changes

Check to see if we have landed a vote since starting the node and wait to produce leader slots until then.

Fixes #

### DIFF
--- a/core/src/replay_stage.rs
+++ b/core/src/replay_stage.rs
@@ -41,6 +41,7 @@ use solana_sdk::{
     genesis_config::ClusterType,
     hash::Hash,
     pubkey::Pubkey,
+    signature::Signature,
     signature::{Keypair, Signer},
     timing::timestamp,
     transaction::Transaction,
@@ -63,6 +64,7 @@ pub const SUPERMINORITY_THRESHOLD: f64 = 1f64 / 3f64;
 pub const MAX_UNCONFIRMED_SLOTS: usize = 5;
 pub const DUPLICATE_LIVENESS_THRESHOLD: f64 = 0.1;
 pub const DUPLICATE_THRESHOLD: f64 = 1.0 - SWITCH_FORK_THRESHOLD - DUPLICATE_LIVENESS_THRESHOLD;
+const MAX_VOTE_SIGNATURES: usize = 200;
 
 #[derive(PartialEq, Debug)]
 pub(crate) enum HeaviestForkFailures {
@@ -111,6 +113,7 @@ pub struct ReplayStageConfig {
     pub rewards_recorder_sender: Option<RewardsRecorderSender>,
     pub cache_block_time_sender: Option<CacheBlockTimeSender>,
     pub bank_notification_sender: Option<BankNotificationSender>,
+    pub wait_for_vote_to_start_leader: bool,
 }
 
 #[derive(Default)]
@@ -282,6 +285,7 @@ impl ReplayStage {
             rewards_recorder_sender,
             cache_block_time_sender,
             bank_notification_sender,
+            wait_for_vote_to_start_leader,
         } = config;
 
         trace!("replay stage");
@@ -312,6 +316,8 @@ impl ReplayStage {
                 let mut skipped_slots_info = SkippedSlotsInfo::default();
                 let mut replay_timing = ReplayTiming::default();
                 let mut gossip_duplicate_confirmed_slots: GossipDuplicateConfirmedSlots = BTreeMap::new();
+                let mut voted_signatures = Vec::new();
+                let mut has_new_vote_been_rooted = !wait_for_vote_to_start_leader;
                 loop {
                     let allocated = thread_mem_usage::Allocatedp::default();
 
@@ -523,6 +529,8 @@ impl ReplayStage {
                             &cache_block_time_sender,
                             &bank_notification_sender,
                             &mut gossip_duplicate_confirmed_slots,
+                            &mut voted_signatures,
+                            &mut has_new_vote_been_rooted,
                         );
                     };
                     voting_time.stop();
@@ -614,6 +622,7 @@ impl ReplayStage {
                             &progress,
                             &retransmit_slots_sender,
                             &mut skipped_slots_info,
+                            has_new_vote_been_rooted,
                         );
 
                         let poh_bank = poh_recorder.lock().unwrap().bank();
@@ -1020,7 +1029,12 @@ impl ReplayStage {
         progress_map: &ProgressMap,
         retransmit_slots_sender: &RetransmitSlotsSender,
         skipped_slots_info: &mut SkippedSlotsInfo,
+        has_new_vote_been_rooted: bool,
     ) {
+        if !has_new_vote_been_rooted {
+            info!("Haven't landed a vote, so skipping my leader slot");
+            return;
+        }
         // all the individual calls to poh_recorder.lock() are designed to
         // increase granularity, decrease contention
 
@@ -1238,6 +1252,8 @@ impl ReplayStage {
         cache_block_time_sender: &Option<CacheBlockTimeSender>,
         bank_notification_sender: &Option<BankNotificationSender>,
         gossip_duplicate_confirmed_slots: &mut GossipDuplicateConfirmedSlots,
+        vote_signatures: &mut Vec<Signature>,
+        has_new_vote_been_rooted: &mut bool,
     ) {
         if bank.is_empty() {
             inc_new_counter_info!("replay_stage-voted_empty_bank", 1);
@@ -1290,6 +1306,8 @@ impl ReplayStage {
                 highest_confirmed_root,
                 heaviest_subtree_fork_choice,
                 gossip_duplicate_confirmed_slots,
+                has_new_vote_been_rooted,
+                vote_signatures,
             );
             subscriptions.notify_roots(rooted_slots);
             if let Some(sender) = bank_notification_sender {
@@ -1319,6 +1337,8 @@ impl ReplayStage {
             last_vote,
             &tower_slots,
             switch_fork_decision,
+            vote_signatures,
+            *has_new_vote_been_rooted,
         );
     }
 
@@ -1330,6 +1350,8 @@ impl ReplayStage {
         vote: Vote,
         tower: &[Slot],
         switch_fork_decision: &SwitchForkDecision,
+        vote_signatures: &mut Vec<Signature>,
+        has_new_vote_been_rooted: bool,
     ) {
         if authorized_voter_keypairs.is_empty() {
             return;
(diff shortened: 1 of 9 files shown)


sakridge:
@carllin does it look reasonable at all?

carllin:
will this break if the entire network is restarting at the same time? Maybe it should just be disabled on hard forks?

core/src/replay_stage.rs
```diff
@@ -1238,6 +1252,9 @@ impl ReplayStage {
 
         let mut vote_tx = Transaction::new_with_payer(&[vote_ix], Some(&node_keypair.pubkey()));
 
+        if *first_voted_slot != Slot::MAX {
```
carllin, line 1255: should this be `==`

sakridge: yes

carllin:
hmmm so if somebody is blowing away their ledger and restarting, I'm not sure this would help, i.e.

0 (root)----1---2---3 (leader slot)

They then blow away the ledger, vote on slot 1, set `has_voted` to true on slot 2, and recreate another version of the leader slot on slot 3.

sakridge:
> hmmm so if somebody is blowing away their ledger and restarting, I'm not sure this would help, i.e.
> 
> `0 (root)----1---2---3 (leader slot)`
> 
> They then blow away the ledger, vote on slot 1, set `has_voted` to true on slot 2, and recreate another version of the leader slot on slot 3.

Slot 2 will be frozen already if they created slot 3 on top of it. So they won't be able to land their vote for slot 1 in slot 2. The `has_voted` tower state maybe is the problem. It actually does need to check the specific vote signatures to see if they landed when it came up the 2nd time.

carllin:
> > hmmm so if somebody is blowing away their ledger and restarting, I'm not sure this would help, i.e.
> > `0 (root)----1---2---3 (leader slot)`
> > They then blow away the ledger, vote on slot 1, set `has_voted` to true on slot 2, and recreate another version of the leader slot on slot 3.
> 
> Slot 2 will be frozen already if they created slot 3 on top of it. So they won't be able to land their vote for slot 1 in slot 2. The `has_voted` tower state maybe is the problem. It actually does need to check the specific vote signatures to see if they landed when it came up the 2nd time.

yeah it seems at the very least you'll have to check that the vote signature landed in a rooted fork to ensure you're near the tip. 

Also of note, I think there's still an edge case if you were a bit ahead of the network on a different fork, i.e. 

```
             1-2----20 (your leader slot) (minor fork)
         /
(root)
         \
            3---4-- 5 (major fork) --RESTART-- 6---7
```

If you restart on the major fork and vote on slot 6, which lands in slot 7, you may recreate your vote slot again for slot 20 on the major fork.

mvines:
I was thinking we'd suppress this behavior when `--wait-for-supermajority` is given for the restart case

sakridge:
I need to handle the case where the node is the sole bootstrap leader or maybe 1 of many  or maybe any case where the node might not be able to land a vote and needs to produce slots anyway. I was trying to think if there is an elegant way to figure that out.

t-nelson (approved):
lgtm with nits!

core/src/replay_stage.rs
```diff
@@ -885,7 +893,12 @@ impl ReplayStage {
         progress_map: &ProgressMap,
         retransmit_slots_sender: &RetransmitSlotsSender,
         skipped_slots_info: &mut SkippedSlotsInfo,
+        has_voted: bool,
```
carllin, line 896: nit: rename `has_voted` here and in other use cases to `has_new_vote_been_rooted` to be more explicit about what it represents

validator/src/main.rs
```diff
@@ -1360,6 +1360,13 @@ pub fn main() {
                 .help("After processing the ledger and the next slot is SLOT, wait until a \
                        supermajority of stake is visible on gossip before starting PoH"),
         )
+        .arg(
+            Arg::with_name("no_wait_for_vote_to_start_leader")
```
carllin, line 1364: might be good to add a warning here that if > 33% of the cluster goes down and restarts, everyone will need to set this flag, otherwise even on restart nobody will make progress.

core/src/validator.rs
```diff
@@ -1293,23 +1299,23 @@ fn wait_for_supermajority(
     cluster_info: &ClusterInfo,
     rpc_override_health_check: Arc<AtomicBool>,
     start_progress: &Arc<RwLock<ValidatorStartProgress>>,
-) -> bool {
+) -> (bool, bool) {
```
carllin, line 1302: hmmm could we add to the function comment on what the second boolean represents?

core/src/validator.rs
```diff
@@ -627,16 +629,19 @@ impl Validator {
             check_poh_speed(&genesis_config, None);
         }
 
-        if wait_for_supermajority(
+        let (failed, did_wait) = wait_for_supermajority(
```
carllin, line 632: nit: `did_wait` -> `is_wait_unnecessary`, since in some cases like on hard forks, a wait didn't occur but is necessary (or even better `is_wait_necessary`, but then we have to flip the booleans around below)

core/src/validator.rs
carllin: Seems like this might be better modeled as a `Result<bool>` since we exit the validator whenever the first bool, which represents whether an error occurred, is `true`

core/src/validator.rs
```diff
@@ -1293,23 +1299,23 @@ fn wait_for_supermajority(
     cluster_info: &ClusterInfo,
     rpc_override_health_check: Arc<AtomicBool>,
     start_progress: &Arc<RwLock<ValidatorStartProgress>>,
-) -> bool {
+) -> (bool, bool) {
     if let Some(wait_for_supermajority) = config.wait_for_supermajority {
         match wait_for_supermajority.cmp(&bank.slot()) {
-            std::cmp::Ordering::Less => return false,
+            std::cmp::Ordering::Less => return (false, false),
```
carllin, line 1305: is the second bool equal to `false` ever an acceptable case? I.e. if everyone  restarts together and is waiting for their vote to land in a root, nobody will vote, and so everyone will stall right?

sakridge: `did_wait` can be false, yes. Because nodes could still have --wait-for-supermajority even when the network has passed the wait slot given and they are trying to join the network. They wouldn't wait at all because the slot they loaded from is already past the wait-for slot and the shred version matches. In this case they should wait to vote because they have joined the network that is already started producing blocks.

validator/src/main.rs
sakridge: If 33% goes down, we have to do a manual restart with `--wait-for-supermajority` flag, right? In that we detect that we had to wait for supermajority and we skip this check for the vote. So, it shouldn't be necessary to use it in that case. The only case we need it is if you are starting a bootstrap leader(s) without `--wait-for-supermajority`

core/src/replay_stage.rs
```diff
@@ -293,6 +296,8 @@ impl ReplayStage {
                 let mut partition_exists = false;
                 let mut skipped_slots_info = SkippedSlotsInfo::default();
                 let mut replay_timing = ReplayTiming::default();
+                let mut voted_signatures = Vec::new();
```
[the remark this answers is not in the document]
mvines: Why 201?

core/src/replay_stage.rs
```diff
@@ -1263,6 +1284,14 @@ impl ReplayStage {
 
         let mut vote_tx = Transaction::new_with_payer(&[vote_ix], Some(&node_keypair.pubkey()));
 
+        if !has_new_vote_been_rooted {
+            vote_signatures.push(vote_tx.signatures[0]);
+            if vote_signatures.len() > 200 {
```
mvines, line 1289: Use a constant for `200`?

core/src/replay_stage.rs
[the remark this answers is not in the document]
sakridge: I chose a 200 signature limit for how large this can be. I'm not sure we need to size it initially, most cases should not use the full 200 and the path isn't that performance sensitive.

mvines (commented):
I bet we'll need a `no_wait_for_vote_to_start_leader: true` in this config block as well for `TestValidator`:
https://github.com/solana-labs/solana/blob/d76ad33597f5661c13a9918f617e1271004d1d35/core/src/test_validator.rs#L402-L426

sakridge:
> I bet we'll need a `no_wait_for_vote_to_start_leader: true` in this config block as well for `TestValidator`:
> https://github.com/solana-labs/solana/blob/d76ad33597f5661c13a9918f617e1271004d1d35/core/src/test_validator.rs#L402-L426

added, thanks

behzadnouri:
Should have `multinode-demo/validator.sh` also been updated?
Starting a gce cluster shows that nodes get stuck in 
```
INFO  solana_core::replay_stage] Haven't landed a vote, so skipping my leader slot
```

sakridge:
> Should have `multinode-demo/validator.sh` also been updated?
> Starting a gce cluster shows that nodes get stuck in
> 
> ```
> INFO  solana_core::replay_stage] Haven't landed a vote, so skipping my leader slot
> ```

let me look

sakridge:
> Should have `multinode-demo/validator.sh` also been updated?
> Starting a gce cluster shows that nodes get stuck in
> 
> ```
> INFO  solana_core::replay_stage] Haven't landed a vote, so skipping my leader slot
> ```

I don't think that is the cause of it. Those validators are not staked at the beginning, so they can wait to land a rooted vote. I had that print in a bad spot but: https://github.com/solana-labs/solana/pull/16156 should fix it.

behzadnouri:
I see `buildkite/solana/stable` failing but it passes when I revert this commit.
Also on gce it seems to have impacted replay metrics.
Left is master, right is this commit reverted.

`replay-slot-stats::total_shreds`:
replay-slot-stats total_shreds

`replay_stage-replay-transactions`:
replay_stage-replay-transactions

sakridge:
> I see `buildkite/solana/stable` failing but it passes when I revert this commit.
> Also on gce it seems to have impacted replay metrics.
> Left is master, right is this commit reverted.
> 
> `replay-slot-stats::total_shreds`:
> replay-slot-stats total_shreds
> 
> `replay_stage-replay-transactions`:
> replay_stage-replay-transactions

can you try with https://github.com/solana-labs/solana/pull/16156

behzadnouri:
https://github.com/solana-labs/solana/pull/16156 is also failing `buildkite/solana/stable`: 
https://buildkite.com/solana-labs/solana/builds/42989

sakridge:
Stabled passed here with it:
https://buildkite.com/solana-labs/solana/builds/42978
https://buildkite.com/solana-labs/solana/builds/42993
https://buildkite.com/solana-labs/solana/builds/42970
https://buildkite.com/solana-labs/solana/builds/42962
https://buildkite.com/solana-labs/solana/builds/42962

It looks like a flaky condition in the test to me, it shouldn't really affect that test. The test doesn't actually trigger any of the modified paths of this PR.

PR #15610, opened 2021-03-01
Fix typos

garious:
#### Problem

Typos and some dead code

#### Summary of Changes

Fix typos, clean up some comments, and remove dead code

### DIFF
--- a/runtime/src/append_vec.rs
+++ b/runtime/src/append_vec.rs
@@ -1,3 +1,6 @@
+//! Persistent storage for accounts. For more information, see:
+//! https://docs.solana.com/implemented-proposals/persistent-account-storage
+
 use log::*;
 use memmap2::MmapMut;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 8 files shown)


runtime/src/append_vec.rs
```diff
@@ -165,10 +168,13 @@ impl AppendVec {
             })
             .unwrap();
 
+        // Performance optimization: write a zero to the end of the file to
+        // force the disk space to be allocated upfront.
```
garious, line 172: @aeyakovenko or @sakridge, is this accurate?

runtime/src/append_vec.rs
```diff
@@ -165,10 +168,14 @@ impl AppendVec {
             })
             .unwrap();
 
+        // Theoretical performance optimization: write a zero to the end of
+        // the file so that we won't have to resize it later, which may be
+        // expensive.
```
garious, line 173: @sakridge, better?

sakridge: looks good

PR #15613, opened 2021-03-02
Enable BPF program instruction traces

jackcmay:
#### Problem

There is no way to enable BPF instruction tracing in program-test, validator

#### Summary of Changes

Emit the BPF instruction trace for `RUST_LOG=solana_bpf_loader=trace`

Fixes #


Lichtso (approved):
Looks good

PR #15629, opened 2021-03-02
sends only the latest vote of each validator to the banking stage

behzadnouri:
#### Problem
Votes do not seem to coalesce in entries

#### Summary of Changes
Sends only the latest vote of each validator to the banking stage

### DIFF
--- a/core/src/cluster_info_vote_listener.rs
+++ b/core/src/cluster_info_vote_listener.rs
@@ -44,8 +44,10 @@ use std::{
 };
 
 // Map from a vote account to the authorized voter for an epoch
-pub type VerifiedLabelVotePacketsSender = CrossbeamSender<Vec<(CrdsValueLabel, Packets)>>;
-pub type VerifiedLabelVotePacketsReceiver = CrossbeamReceiver<Vec<(CrdsValueLabel, Packets)>>;
+pub type VerifiedLabelVotePacketsSender =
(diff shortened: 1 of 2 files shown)


core/src/cluster_info_vote_listener.rs
```diff
@@ -353,8 +359,10 @@ impl ClusterInfoVoteListener {
             msgs,
         )
         .filter_map(|(label, vote, verify_result, packet)| {
+            let slot = vote_transaction::parse_vote_transaction(&vote)
```
sakridge, line 362: So my point was, yea we just filter them later and also I think we filter on the other consumer thread (where the `Transaction` objects go to the optimistic confirmation code) with this same function:
https://github.com/solana-labs/solana/blob/2ff6942ed3e6be0b80db3387f74a4f1b32defb42/core/src/cluster_info_vote_listener.rs#L687-L691
So we could just filter them here and skip the filtering later.

behzadnouri: done.
Just to be clear the new commit is filtering out votes with empty `Vote.slots`:
https://github.com/solana-labs/solana/blob/2ff6942ed/programs/vote/src/vote_state/mod.rs#L39-L43
from both outputs, even if the `parse_vote_transaction` succeeds.

carllin (commented):
Change looks good, thanks!

PR #15636, opened 2021-03-02
remove new alloc on deserialization

jeffwashington:
#### Problem
For every account in a transaction, we reallocate a new pubkey to replace the existing pubkey on deserialization. This seems like unnecessary exercising of the memory manager.

#### Summary of Changes
Overwrite the existing pubkey in place instead of allocating a new one. This should execute less code for the same result.

Fixes #

### DIFF
--- a/programs/bpf_loader/src/serialization.rs
+++ b/programs/bpf_loader/src/serialization.rs
@@ -251,7 +251,10 @@ pub fn deserialize_parameters_aligned(
                 + size_of::<u8>() // executable
                 + 4 // padding to 128-bit aligned
                 + size_of::<Pubkey>(); // key
-            account.owner = Pubkey::new(&buffer[start..start + size_of::<Pubkey>()]);
+            account
+                .owner
+                .as_mut()
+                .copy_from_slice(&buffer[start..start + size_of::<Pubkey>()]);
             start += size_of::<Pubkey>(); // owner
             account.lamports = LittleEndian::read_u64(&buffer[start..]);
             start += size_of::<u64>(); // lamports
--- a/sdk/program/src/pubkey.rs
+++ b/sdk/program/src/pubkey.rs
@@ -325,6 +325,12 @@ impl AsRef<[u8]> for Pubkey {
     }
 }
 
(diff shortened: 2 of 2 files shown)


jeffwashington:
@jackcmay do you like [DeRefMut](https://github.com/solana-labs/solana/pull/15636/commits/f6ae04115509202cbb7e97f21168d2ee52f7a606) or [AsMut](https://github.com/solana-labs/solana/pull/15636/commits/0a531918cc4a1a0696e4c538415b49197579b7a4) better?

jackcmay:
AsRef would work here and we already have that defined.  I was thinking that we might want to use DeRef for Pubkey to &Pubkey comparisons (I think we could).  Did you have a reason for using DeRef over AsRef?

jeffwashington:
> AsRef would work here and we already have that defined. I was thinking that we might want to use DeRef for Pubkey to &Pubkey comparisons (I think we could). Did you have a reason for using DeRef over AsRef?
Unless I'm missing something, I need a mutable reference to copy over the pubkey's contents and AsRef only works for immutable references. AsMut appears to be the related trait that gets me a mutable reference.
DeRef is nice because it doesn't require `.as_mut()`on the caller. I didn't find another AsMut implementation in our code base. I found other DeRefMut implementations. There is more Rust for me to learn, I'm sure.

jackcmay:
> > AsRef would work here and we already have that defined. I was thinking that we might want to use DeRef for Pubkey to &Pubkey comparisons (I think we could). Did you have a reason for using DeRef over AsRef?
> > Unless I'm missing something, I need a mutable reference to copy over the pubkey's contents and AsRef only works for immutable references. AsMut appears to be the related trait that gets me a mutable reference.
> > DeRef is nice because it doesn't require `.as_mut()`on the caller. I didn't find another AsMut implementation in our code base. I found other DeRefMut implementations. There is more Rust for me to learn, I'm sure.

Yeah, I was using AsRef/DeRef generically here, you would use the mut versions.  I'm not opposed to using `.as_mut()` in this case it makes it a little more explicit what's happening in this case.

PR #15637, opened 2021-03-02
use copy_from_slice for [u8]

jeffwashington:
#### Problem
performance improvements

#### Summary of Changes
use copy_from_slice instead of clone_from_slice
[Rust docs say "If T implements Copy, it can be more performant to use copy_from_slice."](https://doc.rust-lang.org/std/primitive.slice.html#method.clone_from_slice)
T is a u8 in this case.
copy seems to be the better term and usage for what is happening here anyway.

Fixes #


jeffwashington:
this change went in as part of a different refactoring

PR #15638, opened 2021-03-02
add execute detail timings

jeffwashington:
#### Problem
performance
#### Summary of Changes
add metrics for serialize, create vm, execute, and deserialize

Fixes #

### DIFF
--- a/core/src/banking_stage.rs
+++ b/core/src/banking_stage.rs
@@ -701,6 +701,11 @@ impl BankingStage {
             txs.len(),
         );
 
+        debug!(
+            "process_and_record_transactions_locked: {:?}",
+            execute_timings
+        );
+
         (Ok(num_to_commit), retryable_txs)
     }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4134,6 +4134,7 @@ dependencies = [
  "rand 0.7.3",
  "rand_core 0.6.2",
  "rustversion",
+ "solana-measure",
  "solana-runtime",
  "solana-sdk",
  "solana_rbpf",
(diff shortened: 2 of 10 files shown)


jeffwashington:
example output:
`confirm_full_slot ExecuteTimings load_us 114784 execute_us 342426 store_us 92468 details ExecuteDetailsTimings serialize_us 47660 create_vm_us 8856 execute_us 70725 deserialize_us 29047 changed_account_count 0 total_account_count 5348 total_data_size 347398402 data_size_changed 0
ledger processed in 1 minute and 4 seconds. 3955 MB allocated. root slot is 67266403 6 forks at 67266419 67266448 67266456 67266459 67266471 67266487 with 81 frozen banks`

core/src/banking_stage.rs
```diff
@@ -701,6 +701,11 @@ impl BankingStage {
             txs.len(),
         );
 
+        debug!(
```
jeffwashington, line 704: I wasn't sure where to report all of these. Stephen has preferred debug! in the past, so I assumed these would be available if someone wants to look for them.

jeffwashington: I think [this](https://github.com/solana-labs/solana/blob/830be855dc3c38fe005ba07cf19d71d31eb0f24d/core/src/progress_map.rs#L64) may be a place to add some or all of the new metrics.

sakridge: Yes, in here will be too noisy to turn on by default. Generally we just want to print things at a rate of 1-2 seconds.

sakridge (approved):

jackcmay (approved):

Lichtso (approved):

PR #15639, opened 2021-03-02
Put Account's data in an Arc

jeffwashington:
#### Summary of Changes

Fixes #

### DIFF
--- a/sdk/src/account.rs
+++ b/sdk/src/account.rs
@@ -1,6 +1,6 @@
 use crate::{clock::Epoch, pubkey::Pubkey};
 use solana_program::{account_info::AccountInfo, sysvar::Sysvar};
-use std::{cell::RefCell, cmp, fmt, rc::Rc};
+use std::{cell::RefCell, cmp, fmt, rc::Rc, sync::Arc};
 
 /// An Account with data that is stored on chain
 #[repr(C)]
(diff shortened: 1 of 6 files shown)


sdk/src/account.rs
```diff
@@ -21,6 +21,36 @@ pub struct Account {
     pub rent_epoch: Epoch,
 }
 
+#[derive(Clone, Default, Debug, PartialEq, Eq, AbiExample)]
+pub struct AccountNoData {
+    /// lamports in the account
+    pub lamports: u64,
+    /// data held in this account
+    pub data: Arc<Vec<u8>>,
```
jeffwashington, line 29: @sakridge , @carllin
I've been working on replay timing. We copy data around a lot. This is an attempt to greatly reduce copying data. I've prototyped this all out in a separate branch. Overall, I measure we're saving 30% of the time for a ledger verify of a recent mainnet-beta snapshot with a handful of changes that will each be discussed in their own pull request. This is one of the larger ones. It is intended to be consistent with comments in the code that we should return a ref or a cow from the cache. It looks like it needs to be an Arc vs Rc because we move accounts between threads frequently and Rc is no good. It looks like it needs to be an Arc vs Cow because of lifetime issues. I worked a change almost all the way through that attempted to add the right lifetimes to everything from the cache out and eventually ran into a few problems that looked fundamental. Perhaps this lifetime approach can work at some point.

Note this change doesn't 'work' all the way yet. In my prototype implementation, I made a lot of functions generic on the AnAccount trait. That allows Account and AccountNoData to coexist. It looks like they need to coexist because Account has to remain as is because it is an ABI? This is just a guess. If so, the internals of all our code would likely switch to generating and using AccountNoData instances, and we'd have to convert when we go to some api that requires an Account. This isn't different than cloning an Account today. This change is meant to give us a specific thing to discuss.

Some alternatives include the whole account being behind an Arc or the account being in an Arc AND the data being in an Arc. Updating the data and copying it gets quite expensive. Arc seems to give us the copy on write behavior I was imagining we'd want. Arc also frees us from the lifetime issues of Cow.

jeffwashington: @ryoqun Carl says you've worked on something similar.

ryoqun: Don't worry about my prior work around here. I was just experimenting like you. And nothing concrete.. I'm constantly interrupted by other things... xD
I'll look in depth this later. 
@Lichtso This pr might be interesting because you're also working on similar area. ie: https://github.com/solana-labs/solana/pull/15410

jeffwashington: @Lichtso and I have communicated some on this topic. I am happy to collaborate more. The lifetime and copies of accounts with data from account load and caching to locking, serialization, vms, program execution, deserialization, verification, rollbacks, temporary storage, permanent storage, etc. are all in a circle of life. Sounds fun!

jeffwashington:
replaced by #15691

PR #15640, opened 2021-03-02
Document and clean up AppendVec

garious:
#### Problem

Difficult to grok AppendVec, which uses `unsafe` and may be more vulnerable to bugs

#### Summary of Changes

* Document its methods
* Remove redundant state variable so that it's clear the Mutex is only there for thread synchronization
* Remove `get_mut()` from design doc


sakridge (approved):

mvines:
Rebase on the tip of master to pick up the local-cluster fix

PR #15642, opened 2021-03-02
adds more metrics for tx counts and batch sizes

behzadnouri:
#### Problem
Need more metrics to investigate where transactions are split

#### Summary of Changes
Added more metrics

### DIFF
--- a/core/src/poh_recorder.rs
+++ b/core/src/poh_recorder.rs
@@ -415,8 +415,6 @@ impl PohRecorder {
                     timing::duration_as_us(&now.elapsed()) as usize
                 );
                 if let Some(poh_entry) = res {
-                    inc_new_counter_info!("poh_recorder-num_entries", 1);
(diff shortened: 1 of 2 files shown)


core/src/poh_recorder.rs
```diff
@@ -415,6 +415,8 @@ impl PohRecorder {
                     timing::duration_as_us(&now.elapsed()) as usize
                 );
                 if let Some(poh_entry) = res {
+                    inc_new_counter_info!("poh_recorder-num_entries", 1);
+                    inc_new_counter_info!("poh_recorder-num_transactions", transactions.len());
```
sakridge, line 419: We have other counters for entries and transactions, I don't think it will be significantly different than here.

behzadnouri: removed

sakridge (approved):

carllin (approved):

PR #15647, opened 2021-03-03
Remove ValidatorConfig derive Clone, and fix local-cluster tests

CriesofCarrots:
#### Problem
The local-cluster CI job is timing out on `test_no_optimistic_confirmation_violation_with_tower()`.
This is because https://github.com/solana-labs/solana/pull/15606 adds ValidatorExit to the ValidatorConfig struct, and the `vec![ValdiatorConfig; NUM` pattern used all over local-cluster implicitly clones the same ValidatorExit among all the validators.

#### Summary of Changes
Remove Clone footgun from ValidatorConfig
Add helper functions for local-cluster tests and benches that need a bunch of identical ValidatorConfigs

### DIFF
--- a/accounts-cluster-bench/src/main.rs
+++ b/accounts-cluster-bench/src/main.rs
@@ -561,7 +561,10 @@ fn main() {
 pub mod test {
     use super::*;
     use solana_core::validator::ValidatorConfig;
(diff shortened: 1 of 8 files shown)


mvines (approved):
tysm

PR #15686, opened 2021-03-03
Clarify Instruction::new

### LINKED ISSUE #5970: SDK Instruction constructor implicitly uses bincode serializer
#### Problem
There is an implicit bincode serialization in the Instruction constructor which is not desirable for creating BPF program instructions because they can't use bincode.
```rust
use bincode::serialize;

impl Instruction {
    pub fn new<T: Serialize>(program_id: Pubkey, data: &T, accounts: Vec<AccountMeta>) -> Self {
        let data = serialize(data).unwrap();
        Self {
            program_id,
            data,
            accounts,
        }
    }
}
```

#### Proposed Solution
Probably just rename `Instruction::new` to indicate that only native program instructions should be created with this method.

jackcmay:
#### Problem

`Instruction::new` has a hidden dependency on bincode.  Now that we have `new_with_borsh` it makes sense to continue that pattern and make the data type/serialization explicit.

#### Summary of Changes

Added `new_with_bincode` and `new_with_bytes` (open to name suggestions on the latter :-) )

Fixes #5970

### DIFF
--- a/sdk/program/src/instruction.rs
+++ b/sdk/program/src/instruction.rs
@@ -212,6 +212,14 @@ pub struct Instruction {
 
 impl Instruction {
     pub fn new<T: Serialize>(program_id: Pubkey, data: &T, accounts: Vec<AccountMeta>) -> Self {
+        Self::new_with_bincode(program_id, data, accounts)
+    }
+
+    pub fn new_with_bincode<T: Serialize>(
+        program_id: Pubkey,
+        data: &T,
+        accounts: Vec<AccountMeta>,
+    ) -> Self {
         let data = serialize(data).unwrap();
         Self {
             program_id,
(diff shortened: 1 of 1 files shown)


jackcmay:
@mvines, if these changes look good think we should also officially deprecate `Instruction::new` and change all uses in the repo to `new_with_bincode`?

```
    #[deprecated(
        since = "1.6.0",
        note = "Please select another `Instruction constructor instead"
    )]
 ```

File: sdk/program/src/instruction.rs [hunk 1]
```diff
@@ -232,6 +240,19 @@ impl Instruction {
             accounts,
         }
     }
+
+    pub fn new_with_bytes<T: Serialize>(
+        program_id: Pubkey,
+        data: &[u8],
+        accounts: Vec<AccountMeta>,
+    ) -> Self {
+        let data = serialize(data).unwrap();
```
joncinque, line 249: Copy / pasta of `T: Serialize` and `let data = serialize(data)`?

joncinque (commented):
Thanks for taking the lead on this!

joncinque (dismissed):
Looks good!

PR #15689, opened 2021-03-03
invert default behavior of index hashing

jeffwashington:
#### Problem
we want to enable non-index hashing by default
#### Summary of Changes
obsolete previous flag, add new flag to enable forcing use of index scan.
Fixes #


sakridge (approved):
lgtm

PR #15691, opened 2021-03-03
Account -> AccountSharedData

jeffwashington:
#### Problem
We don't want to copy account and account data around unnecessarily.
Account structure is fixed for abi. It needs to remain.

#### Summary of Changes
Step 1 in creating a new AccountSharedData struct which holds data in an Arc.
This change intends to be as much as possible just a rename since the rename touches many files.
Subsequent changes will change the implementation of AccountSharedData and handle its fallout.
I accomplished this by creating a parallel AccountSharedData which is a duplicate of Account. This let me produce something that should be identical in performance and allow us to review the mass rename changes separately from what should be interesting changes.
#15714
Fixes #

### DIFF
--- a/account-decoder/src/lib.rs
+++ b/account-decoder/src/lib.rs
@@ -16,7 +16,10 @@ pub mod validator_info;
 
 use {
     crate::parse_account_data::{parse_account_data, AccountAdditionalData, ParsedAccount},
-    solana_sdk::{account::Account, clock::Epoch, fee_calculator::FeeCalculator, pubkey::Pubkey},
+    solana_sdk::{
+        account::ReadableAccount, account::WritableAccount, clock::Epoch,
+        fee_calculator::FeeCalculator, pubkey::Pubkey,
+    },
     std::{
         io::{Read, Write},
         str::FromStr,
@@ -57,58 +60,61 @@ pub enum UiAccountEncoding {
 }
 
 impl UiAccount {
-    pub fn encode(
+    pub fn encode<T: ReadableAccount>(
         pubkey: &Pubkey,
-        account: Account,
+        account: T,
         encoding: UiAccountEncoding,
         additional_data: Option<AccountAdditionalData>,
         data_slice_config: Option<UiDataSliceConfig>,
     ) -> Self {
         let data = match encoding {
             UiAccountEncoding::Binary => UiAccountData::LegacyBinary(
-                bs58::encode(slice_data(&account.data, data_slice_config)).into_string(),
+                bs58::encode(slice_data(&account.data(), data_slice_config)).into_string(),
             ),
             UiAccountEncoding::Base58 => UiAccountData::Binary(
-                bs58::encode(slice_data(&account.data, data_slice_config)).into_string(),
+                bs58::encode(slice_data(&account.data(), data_slice_config)).into_string(),
                 encoding,
             ),
             UiAccountEncoding::Base64 => UiAccountData::Binary(
-                base64::encode(slice_data(&account.data, data_slice_config)),
+                base64::encode(slice_data(&account.data(), data_slice_config)),
                 encoding,
             ),
             UiAccountEncoding::Base64Zstd => {
                 let mut encoder = zstd::stream::write::Encoder::new(Vec::new(), 0).unwrap();
                 match encoder
-                    .write_all(slice_data(&account.data, data_slice_config))
+                    .write_all(slice_data(&account.data(), data_slice_config))
                     .and_then(|()| encoder.finish())
                 {
                     Ok(zstd_data) => UiAccountData::Binary(base64::encode(zstd_data), encoding),
                     Err(_) => UiAccountData::Binary(
-                        base64::encode(slice_data(&account.data, data_slice_config)),
+                        base64::encode(slice_data(&account.data(), data_slice_config)),
                         UiAccountEncoding::Base64,
                     ),
                 }
             }
             UiAccountEncoding::JsonParsed => {
                 if let Ok(parsed_data) =
-                    parse_account_data(pubkey, &account.owner, &account.data, additional_data)
+                    parse_account_data(pubkey, &account.owner(), &account.data(), additional_data)
                 {
                     UiAccountData::Json(parsed_data)
                 } else {
-                    UiAccountData::Binary(base64::encode(&account.data), UiAccountEncoding::Base64)
+                    UiAccountData::Binary(
+                        base64::encode(&account.data()),
+                        UiAccountEncoding::Base64,
+                    )
                 }
             }
         };
         UiAccount {
-            lamports: account.lamports,
+            lamports: account.lamports(),
             data,
-            owner: account.owner.to_string(),
-            executable: account.executable,
-            rent_epoch: account.rent_epoch,
+            owner: account.owner().to_string(),
+            executable: account.executable(),
+            rent_epoch: account.rent_epoch(),
         }
     }
 
-    pub fn decode(&self) -> Option<Account> {
+    pub fn decode<T: WritableAccount>(&self) -> Option<T> {
         let data = match &self.data {
             UiAccountData::Json(_) => None,
             UiAccountData::LegacyBinary(blob) => bs58::decode(blob).into_vec().ok(),
@@ -128,13 +134,13 @@ impl UiAccount {
                 UiAccountEncoding::Binary | UiAccountEncoding::JsonParsed => None,
             },
         }?;
-        Some(Account {
-            lamports: self.lamports,
+        Some(T::create(
+            self.lamports,
             data,
-            owner: Pubkey::from_str(&self.owner).ok()?,
-            executable: self.executable,
-            rent_epoch: self.rent_epoch,
-        })
+            Pubkey::from_str(&self.owner).ok()?,
+            self.executable,
+            self.rent_epoch,
+        ))
     }
 }
 
@@ -184,6 +190,7 @@ fn slice_data(data: &[u8], data_slice_config: Option<UiDataSliceConfig>) -> &[u8
 #[cfg(test)]
 mod test {
     use super::*;
+    use solana_sdk::account::{Account, AccountSharedData};
 
     #[test]
     fn test_slice_data() {
@@ -217,9 +224,9 @@ mod test {
     fn test_base64_zstd() {
         let encoded_account = UiAccount::encode(
             &Pubkey::default(),
-            Account {
+            AccountSharedData {
                 data: vec![0; 1024],
-                ..Account::default()
+                ..AccountSharedData::default()
             },
             UiAccountEncoding::Base64Zstd,
             None,
@@ -230,7 +237,9 @@ mod test {
             UiAccountData::Binary(_, UiAccountEncoding::Base64Zstd)
         ));
 
-        let decoded_account = encoded_account.decode().unwrap();
+        let decoded_account = encoded_account.decode::<Account>().unwrap();
+        assert_eq!(decoded_account.data, vec![0; 1024]);
+        let decoded_account = encoded_account.decode::<AccountSharedData>().unwrap();
         assert_eq!(decoded_account.data, vec![0; 1024]);
     }
 }
--- a/cli/src/stake.rs
+++ b/cli/src/stake.rs
@@ -1747,9 +1747,10 @@ pub fn process_show_stake_history(
     use_lamports_unit: bool,
 ) -> ProcessResult {
     let stake_history_account = rpc_client.get_account(&stake_history::id())?;
-    let stake_history = from_account::<StakeHistory>(&stake_history_account).ok_or_else(|| {
-        CliError::RpcRequestError("Failed to deserialize stake history".to_string())
-    })?;
+    let stake_history =
+        from_account::<StakeHistory, _>(&stake_history_account).ok_or_else(|| {
+            CliError::RpcRequestError("Failed to deserialize stake history".to_string())
+        })?;
 
     let mut entries: Vec<CliStakeHistoryEntry> = vec![];
     for entry in stake_history.deref() {
--- a/runtime/src/vote_account.rs
+++ b/runtime/src/vote_account.rs
@@ -1,6 +1,8 @@
 use serde::de::{Deserialize, Deserializer};
 use serde::ser::{Serialize, Serializer};
-use solana_sdk::{account::Account, instruction::InstructionError, pubkey::Pubkey};
+use solana_sdk::{
+    account::Account, account::AccountSharedData, instruction::InstructionError, pubkey::Pubkey,
+};
 use solana_vote_program::vote_state::VoteState;
 use std::{
     borrow::Borrow,
@@ -172,9 +174,24 @@ impl<'de> Deserialize<'de> for ArcVoteAccount {
     }
 }
 
+impl From<AccountSharedData> for ArcVoteAccount {
+    fn from(account: AccountSharedData) -> Self {
+        Self(Arc::new(VoteAccount::from(account)))
+    }
+}
 impl From<Account> for ArcVoteAccount {
     fn from(account: Account) -> Self {
         Self(Arc::new(VoteAccount::from(account)))
+    }
+}
+
+impl From<AccountSharedData> for VoteAccount {
+    fn from(account: AccountSharedData) -> Self {
+        Self {
+            account: Account::from(account),
+            vote_state: RwLock::new(INVALID_VOTE_STATE),
+            vote_state_once: Once::new(),
+        }
     }
 }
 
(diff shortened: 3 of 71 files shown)


jeffwashington:
I ran ledger-tool verify on a colo machine against a recent mainnet snapshot. This change should have no or minimal effect.
master:
`ledger processed in 1 minute and 31 seconds. 8053 MB allocated. root slot is 67566632 2 forks at 67566647 67566674 with 42 frozen banks`
this pr:
`ledger processed in 1 minute and 31 seconds. 7982 MB allocated. root slot is 67566632 2 forks at 67566647 67566674 with 42 frozen banks`

account-decoder/src/lib.rs
```diff
@@ -58,6 +61,22 @@ pub enum UiAccountEncoding {
 
 impl UiAccount {
     pub fn encode(
+        pubkey: &Pubkey,
+        account: AccountNoData,
```
carllin, line 65: Is it possible to take a reference to a `AnAccount` here to generalize over both `AccountNoData` and `Account`?

sdk/src/account.rs
```diff
@@ -21,6 +21,159 @@ pub struct Account {
...
 
+//pub type AccountNoData = Account;
+
+#[derive(Serialize, Deserialize, PartialEq, Eq, Clone, Default, AbiExample)]
+pub struct AccountNoData {
+    /// lamports in the account
+    pub lamports: u64,
+    /// data held in this account
+    #[serde(with = "serde_bytes")]
+    pub data: Vec<u8>, // will be: Arc<Vec<u8>>,
```
carllin, line 32: Just another thought for exploration, what what the tradeoffs be if instead of a separate trait/struct for AccountNoData, we generalized the `data` field of `Account` to be a

`
enum AccountData {
     Owned(Vec<u8>),
     Shared(Arc<Vec<u8>>)
}
`
Would that reduce a lot of the needs to implement separate traits for both `AccountNoData` and `Account`, or is it just a lot of headache from having to replumb everything in the system?

jeffwashington:
Some archived discussion here:
#15639

sdk/src/account.rs
jeffwashington: @carllin, I am under the impression that we cannot change the contents of struct Account. Certainly the ABI tests complain if I change it. I already ran into a few other structs that had to remain using Account. Perhaps we just can't change the serialization. I am unclear of the actual constraints. One thing I thought I understood from the discussion we had with @sakridge was that the clearest way forward was a new struct, currently proposed to be called AccountNoData. I am happy for feedback on another way forward.

jeffwashington: As part next of this topic on enum, I like the idea. I just talked with @Lichtso about serialization/deserialization of accounts for bpf execution, whether accounts are read-only or writeable. I think the new AccountNoData may at some point BE an enum similar to or maybe an extension of [LoadedAccount](https://github.com/solana-labs/solana/blob/ee621878b058ea1068f007e3261b2b0acae386be/runtime/src/accounts_db.rs#L185). Perhaps the representation in memory of an account after being written by a bpf program could be indexes into a slice that was mapped into the vm's address space. account.lamports() would read the enum and go to the slice to read out the memory or the data. Once we wrote to a store, we would free the slice. When we next loaded the account, it woudl begin life as an arc to an account struct that had an arc to the data. But the memory representation of the actual account data would be abstracted behind impl functions on the AccountNoData struct. Perhaps similar to the [AnAccount](https://github.com/solana-labs/solana/blob/dd5554112d0edfd2d83b22789718893afc4426b8/sdk/src/account.rs#L104) trait I've been working out and trying to avoid ;-).

account-decoder/src/lib.rs
jeffwashington: These are the kind of replication that I had mentioned to you in slack I wasn't happy with yet. I figured I'd revisit things like this now that the whole ci is happy. I would love to combine all these _legacy or _no_data functions as much as it makes sense. I'm open to ideas or rust ways to accomplish some of this. I haven't become an expert in rust alternatives to specialization yet. I'm also not an expert in serialization/deserialization yet, and I'm trying to keep things simple. So, there are a few tradeoffs and a few areas where I'm unsure how to tradeoff cost/benefit. In short, yes, it is worth trying to reduce all of these to reduce test burden, reduce duplicated code, etc. But, we also have to consider execution efficiency, of course. Please do provide feedback and ideas. Thanks!

jeffwashington: yes, and I did that and realized that nobody called encode_legacy anyway. testing that now on the ci. But, I saved the AnAccount version anyway ;-)

sdk/src/account.rs
jeffwashington: I talked with Carl. We think Account can't change.

client/src/rpc_client.rs
```diff
@@ -728,6 +729,14 @@ impl RpcClient {
             .ok_or_else(|| RpcError::ForUser(format!("AccountNotFound: pubkey={}", pubkey)).into())
     }
 
+    /// Note that `get_account` returns `Err(..)` if the account does not exist whereas
+    /// `get_account_with_commitment` returns `Ok(None)` if the account does not exist.
+    pub fn get_account_no_data(&self, pubkey: &Pubkey) -> ClientResult<AccountNoData> {
```
jeffwashington, line 734: @CriesofCarrots @jackcmay : @carllin suggested I call this type of rpc change out to you two in case we needed to do something different here.

sdk/src/account.rs
```diff
@@ -21,24 +21,184 @@ pub struct Account {
     pub rent_epoch: Epoch,
 }
 
+/// An Account with data that is stored on chain
+/// This will become a new in-memory representation of the 'Account' struct data.
+/// The existing 'Account' structure cannot easily change due to downstream projects.
+/// This struct will shortly rely on something like the AnAccount trait for access to the fields.
+#[derive(Serialize, Deserialize, PartialEq, Eq, Clone, Default, AbiExample)]
+pub struct AccountNoData {
```
sakridge, line 29: I think `AccountRefData` or `AccountSharedData` might be better. `AccountNoData` sounds to me like there's no data at all, but I think it just means it has a reference to the data instead of owning it outright. Does it sound right?

jeffwashington:
I like shared of the choices ;)

client/src/rpc_client.rs
CriesofCarrots: Thanks for the heads up! Is there anything in particular in the client or rpc tooling on which you have questions?

jeffwashington: @CriesofCarrots: @carllin and I expressed some lack of knowledge in the rpc area. @sakridge raised rpc as a concern early on with this refactoring and rename. So, we need an expert to make sure we can add methods here and things will work ok and to see if we missed something.

client/src/rpc_client.rs
```diff
@@ -843,25 +880,36 @@ impl RpcClient {
         )
     }
 
+    fn get_config_for_program_accounts(&self) -> RpcProgramAccountsConfig {
```
CriesofCarrots, line 883:
```suggestion
    fn make_program_accounts_config(&self) -> RpcProgramAccountsConfig {
```
`get` typically implies a call or request; wdyt about this alternative?

sdk/src/account.rs
```diff
@@ -21,112 +21,455 @@ pub struct Account {
...
+            lamports: other.lamports,
+            data: other.data,
+            owner: other.owner,
+            executable: other.executable,
+            rent_epoch: other.rent_epoch,
+        }
+    }
+}
+
+pub trait AnAccountWritable: AnAccount {
```
CriesofCarrots, line 75: Fwiw, I find the `An` prefix a little hard to parse out of context. This trait could be `WritableAccount`. Trait below... maybe `GenericAccount`? or `ReadableAccount`?

CriesofCarrots (commented):
Is the intent to deprecate Account altogether? I am wondering if the rpc_client changes are necessary, or if downstream clients should still be operating using Account (with AccountSharedData just an internal node concept). The account will always have to be copied to be returned from RPC anyway, so could use an `AccountSharedData::into<Account>()` method.

client/src/rpc_client.rs
jeffwashington: just trying to share code an eliminate duplication

sdk/src/account.rs
jeffwashington: I like WriteableAccount and ReadableAccount. Those work for me.

client/src/rpc_client.rs
CriesofCarrots: Yep, just nitpicking the method name :)

jeffwashington:
> Is the intent to deprecate Account altogether? I am wondering if the rpc_client changes are necessary, or if downstream clients should still be operating using Account (with AccountSharedData just an internal node concept). The account will always have to be copied to be returned from RPC anyway, so could use an `AccountSharedData::into<Account>()` method.

The intuition of @carllin , @sakridge , @jackcmay appears to be that we will need to keep the existing Account struct around for things like 'downstream projects' that will have to be modified if we make the fields of 'Account' private and abstract away how the conceptual account is actually stored in memory.  For example, if lamports becomes private, and we require traits to access ie. account.lamports**()** or account.set_lamports(new_lamports), then existing code in downstream projects or 3rd parties could break.

I tried to walk the right line for things like rpc where we could at the beginning get an account in either format (Account or AccountSharedData). The existing function call (.get_account()) still needed to return Account. This would require us to litter in various places AccountSharedData::from(_.get_account()). This at a minimum could result in extra struct copying and such on every account load. Perhaps this is inlined and optimized away. I went with the approach that in the few origins of Account[SharedData], the caller could be explicit about what storage they wanted initially, which should result in higher performance code. It didn't seem there were too many places where we needed to replicate the 'get' functions. I am happy to consider other options.

sdk/src/account.rs
```diff
@@ -21,112 +21,455 @@ pub struct Account {
...
+/// This will become a new in-memory representation of the 'Account' struct data.
+/// The existing 'Account' structure cannot easily change due to downstream projects.
+/// This struct will shortly rely on something like the AnAccount trait for access to the fields.
+#[derive(Serialize, Deserialize, PartialEq, Eq, Clone, Default, AbiExample)]
+pub struct AccountSharedData {
+    /// lamports in the account
+    pub lamports: u64,
+    /// data held in this account
+    #[serde(with = "serde_bytes")]
+    pub data: Vec<u8>, // will be: Arc<Vec<u8>>,
```
jackcmay, line 34: If this data is shared betwen threads (central cache?) do we also need to wrap with a mutex or are you thinking copy on write?

sdk/src/account.rs
```diff
@@ -21,112 +21,455 @@ pub struct Account {
...
+    /// the program that owns this account. If executable, the program that loads this account.
+    pub owner: Pubkey,
+    /// this account's data contains a loaded program (and is now read-only)
+    pub executable: bool,
+    /// the epoch at which this account will next owe rent
+    pub rent_epoch: Epoch,
+}
+
+pub fn accounts_equal<T: AnAccount, U: AnAccount>(me: &T, other: &U) -> bool {
```
jackcmay, line 43: We haven't gone back andd one this retroactively yet, but in general, all public functions in the sdk should be documented

sdk/src/account.rs
jeffwashington: > If this data is shared betwen threads (central cache?) do we also need to wrap with a mutex or are you thinking copy on write?

copy on write. The semantics today are multiple identical copies of the data, but any copy can be modified. This PR is moving towards multiple references to shared data (for performance) where if any reference wants to modify its data, it gets a new copy of the data and the other refers retain references to the unchanged data.

sdk/src/account.rs
jeffwashington: > We haven't gone back andd one this retroactively yet, but in general, all public functions in the sdk should be documented

Where should they be documented? In code or somewhere else also?

jeffwashington:
I just pushed commits to respond to pr feedback and rebase on master.

sdk/src/account.rs
jackcmay: documenting it in code like this:
https://github.com/solana-labs/solana/blob/a7d5645f72265e243c60084651f30d5dc0b3a5cd/sdk/program/src/pubkey.rs#L158
Will translate into Rust docs like this:
https://docs.rs/solana-program/1.5.13/solana_program/pubkey/struct.Pubkey.html#method.create_program_address

cli/src/program.rs
```diff
@@ -1137,7 +1137,7 @@ fn do_process_program_write_and_deploy(
     let (initial_message, write_messages, balance_needed) =
         if let Some(buffer_authority_signer) = buffer_authority_signer {
             let (initial_instructions, balance_needed) = if let Some(account) = rpc_client
-                .get_account_with_commitment(buffer_pubkey, config.commitment)?
+                .get_account_shared_data_with_commitment(buffer_pubkey, config.commitment)?
```
jackcmay, line 1140: With the new name change should this be `get_writeable/readable_account_with_commitment`?

cli/src/stake.rs
```diff
@@ -1746,10 +1746,11 @@ pub fn process_show_stake_history(
     config: &CliConfig,
     use_lamports_unit: bool,
 ) -> ProcessResult {
-    let stake_history_account = rpc_client.get_account(&stake_history::id())?;
-    let stake_history = from_account::<StakeHistory>(&stake_history_account).ok_or_else(|| {
-        CliError::RpcRequestError("Failed to deserialize stake history".to_string())
-    })?;
+    let stake_history_account = rpc_client.get_account_shared_data(&stake_history::id())?;
```
jackcmay, line 1749: Are we going to expose the new structures to rpc_client's or keep them on `Account`?

sdk/src/account.rs
jeffwashington: I documented this function. I didn't see anything else new and pub except traits in this file.

cli/src/program.rs
jeffwashington: get_account_shared_data* returns AccountSharedData vs get_account* returns Account. It isn't returning a trait or boxed ReadableAccount. The only difference is the type of the thing it returns. And, in trying to avoid additional copies or overhead each time we get an account, I was imagining we'd need to duplicate these code paths.
There is some discussion [here](https://github.com/solana-labs/solana/pull/15691#issuecomment-793034139).

cli/src/stake.rs
jeffwashington: Perhaps I need to take another path at detangling things. It seemed there was cross use where some number of callers in our mono repo were calling get_account. It seems that anything like the cli which perhaps could make copies could benefit from the arc ideas or the differing storage formats internally? I guess our cli is considered an rpc client? And would you rather keep those as Account?

jeffwashington: Basically, anything that ends up calling any shared code should ideally be using AccountSharedData, or else we could end up with conversions, which could become hidden copies at every intersection of the types.

jackcmay: +1 on reducing copies, @CriesofCarrots what are your thoughts on rpc_client switching to `AccountSharedData`?

jeffwashington:
bench-tps #s
master: Average TPS: 52101.16
shareddata: Average TPS: 52483.273

colo machine, single node testnet

cli/src/stake.rs
CriesofCarrots: Our cli uses rpc_client under the hood, yes.
I'm not sure I see a strong need to use `AccountSharedData` in rpc_client. The account has to be copied to be transmitted via rpc regardless, with allocation on the receiving end. Essentially the main benefit the duplicated methods provide is performing a conversion to `AccountSharedData`, and I wonder if that, or other Arc wrapping, should be explicit on the part of the caller? (Fwiw, I haven't thought of a viable use-case for WritableAccounts in clients.)

That said, the only real harm in supporting the additional methods is the added complexity of choosing when to use one or the other. So if we go that route, I'd like to see thorough documentation on how a client should make that choice.

jackcmay: Agree, no reason to extend this type through the rpc boundary at this moment.  Jeff, how much work is it to pull this back a bit and keep AccountSharedData on the "inside"?

jeffwashington:
I reworked a lot, squashed it all, rolled back some changes. Squashed it.

runtime/src/vote_account.rs
```diff
@@ -172,12 +174,27 @@ impl<'de> Deserialize<'de> for ArcVoteAccount {
...
+        Self(Arc::new(VoteAccount::from(account)))
+    }
+}
 impl From<Account> for ArcVoteAccount {
     fn from(account: Account) -> Self {
         Self(Arc::new(VoteAccount::from(account)))
     }
 }
 
+impl From<AccountSharedData> for VoteAccount {
```
carllin, line 188: @behzadnouri does this look ok now?

CriesofCarrots (approved):

runtime/src/vote_account.rs
behzadnouri: the issue was changing `Serialize/Deserialize` impl which is now reverted.
changes to `new_rand_vote_account` also seem redundant but i guess it does not make a difference.

PR #15702, opened 2021-03-04
explorer: Add tabs for block program and account stats

jstarry:
#### Problem
The current block page isn't great for understanding what actually happened in the block. There's little transparency on which transactions were actually processed.

#### Summary of Changes
- Split block cards into separate tabs
- Add tab for block programs which breaks down the distribution of programs called
- Add tab for block accounts which breaks down the distribution of loaded accounts

Fixes #

### DIFF
--- a/explorer/tsconfig.json
+++ b/explorer/tsconfig.json
@@ -4,6 +4,7 @@
     "lib": ["dom", "dom.iterable", "esnext"],
     "allowJs": true,
     "skipLibCheck": true,
+    "downlevelIteration": true,
     "esModuleInterop": true,
     "allowSyntheticDefaultImports": true,
     "strict": true,
--- a/explorer/src/App.tsx
+++ b/explorer/src/App.tsx
@@ -46,8 +46,10 @@ function App() {
           />
           <Route
             exact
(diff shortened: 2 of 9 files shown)


jstarry:
Program Stats:
http://localhost:3000/block/67419221/programs
Screen Shot 2021-03-04 at 7 14 35 PM

Account Stats:
http://localhost:3000/block/67419221/accounts
Screen Shot 2021-03-04 at 7 14 39 PM

explorer/tsconfig.json
```diff
@@ -4,6 +4,7 @@
     "lib": ["dom", "dom.iterable", "esnext"],
     "allowJs": true,
     "skipLibCheck": true,
+    "downlevelIteration": true,
```
jstarry, line 7: Useful for iterating through `Map`

oJshua: Ah, I ran into this last week. Ended up going with `for(let entry of Array.from(map.entries())) {`. A welcome change.

oJshua (commented):
Looks great! Will be nice to have the extra insight.

PR #15714, opened 2021-03-04
Data as arc

jeffwashington:
Problem
We don't want to copy account and account data around unnecessarily.
Account structure is fixed for abi. It needs to remain.

Summary of Changes
Step 1 is creating a new AccountSharedData struct which we can manipulate. That is PR #15691 
Step 2 is this PR. This is to make AccountSharedData.data hold data: Arc<Vec<u8>>. The major changes have comments next to them in the source below.

### DIFF
--- a/runtime/src/accounts_cache.rs
+++ b/runtime/src/accounts_cache.rs
@@ -1,5 +1,5 @@
 use dashmap::DashMap;
-use solana_sdk::{account::Account, clock::Slot, hash::Hash, pubkey::Pubkey};
+use solana_sdk::{account::AccountSharedData, clock::Slot, hash::Hash, pubkey::Pubkey};
 use std::{
     collections::BTreeSet,
     ops::Deref,
@@ -42,7 +42,7 @@ impl SlotCacheInner {
         );
     }
 
-    pub fn insert(&self, pubkey: &Pubkey, account: Account, hash: Hash) {
+    pub fn insert(&self, pubkey: &Pubkey, account: AccountSharedData, hash: Hash) {
         if self.cache.contains_key(pubkey) {
             self.same_account_writes.fetch_add(1, Ordering::Relaxed);
             self.same_account_writes_size
@@ -86,7 +86,7 @@ impl Deref for SlotCacheInner {
 
 #[derive(Debug, Clone)]
 pub struct CachedAccount {
-    pub account: Account,
+    pub account: AccountSharedData,
     pub hash: Hash,
 }
 
@@ -123,7 +123,7 @@ impl AccountsCache {
         );
     }
 
-    pub fn store(&self, slot: Slot, pubkey: &Pubkey, account: Account, hash: Hash) {
+    pub fn store(&self, slot: Slot, pubkey: &Pubkey, account: AccountSharedData, hash: Hash) {
         let slot_cache = self.slot_cache(slot).unwrap_or_else(||
             // DashMap entry.or_insert() returns a RefMut, essentially a write lock,
             // which is dropped after this block ends, minimizing time held by the lock.
@@ -235,7 +235,7 @@ pub mod tests {
         cache.store(
             inserted_slot,
             &Pubkey::new_unique(),
-            Account::new(1, 0, &Pubkey::default()),
+            AccountSharedData::new(1, 0, &Pubkey::default()),
             Hash::default(),
         );
         // If the cache is told the size limit is 0, it should return the one slot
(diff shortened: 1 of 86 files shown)


jeffwashington:
Processing a ledger with ledger-tool on a recent mainnet snapshot with this change produced this difference vs master:
`ledger processed in 1 minute and 21 seconds. 7124 MB allocated. root slot is 67566632 2 forks at 67566647 67566674 with 42 frozen banks`
`ledger processed in 1 minute and 31 seconds. 8053 MB allocated. root slot is 67566632 2 forks at 67566647 67566674 with 42 frozen banks`

jeffwashington:
note this is based off #15691
That pr is intended to have as much as possible the bulk, boring code changes due to replacing a struct (while preserving the previous one)

sdk/src/account.rs
```diff
@@ -21,6 +22,190 @@ pub struct Account {
...
 }
 
+//pub type AccountNoData = Account;
+
+#[derive(PartialEq, Eq, Clone, Default, AbiExample)]
+pub struct AccountNoData {
+    /// lamports in the account
+    pub lamports: u64,
+    /// data held in this account
+    pub data: Arc<Vec<u8>>,
```
jeffwashington, line 32: This is the root change in this pr. data becomes an Arc.

jeffwashington: @Lichtso fyi. We can discuss this Freitag.

runtime/src/accounts_cache.rs
```diff
@@ -86,7 +86,7 @@ impl Deref for SlotCacheInner {
 
 #[derive(Debug, Clone)]
 pub struct CachedAccount {
-    pub account: Account,
+    pub account: AccountNoData,
```
jeffwashington, line 89: The root place this impacts many situations is here. The cache now holds AccountNoData, which shares its data behind an Arc. So, as an item is loaded from the cache, the data is never copied until it needs to be modified by someone. Cloning the account clones the non-data fields and addrefs the Arc to the data.

A subsequent change will attempt to move this AccountNoData struct to AccountNoDataInner and make AccountNoData have an Arc to an AccountNoDataInner. This would reduce copying of the rest of the Account contents (lamports, executable, owner, rent_epoch)

sdk/src/account.rs
jackcmay: Is there a reason the whole `Account` doesn't move under `Arc`?

jeffwashington: We imagine the eventual state may be AccountNoData = Arc<AccountInner>
where AccountInner has an Arc<Data>

The thinking is the data is large and we want to avoid copying it. The account itself is moderate, but fixed size and we copy it a lot. So, if a program changes lamports, the account data changes, but we don't want to copy the data. I was trying to move things in easier to understand stages.
Moving the whole account under and Arc was a next thing to do. I have not built it out and made it work yet to determine the performance differences or the difficulty in getting it to pass the CI.

jeffwashington: But another next state is where AccountNoData is an enum which either owns the data or it references data from a slice where the slice was the serialized modified accounts sent to a program. I talked with Alex and Carl about this today.

jeffwashington: Feel free to provide feedback and discuss more. You (Jack) were on my list to discuss with but I haven't been able to connect yet.

jeffwashington: I just tried this:
`
pub type AccountNoData = Arc<AccountNoDataInner>;
    
#[derive(PartialEq, Eq, Clone, Default, AbiExample)]
pub struct AccountNoDataInner {
    /// lamports in the account
    pub lamports: u64,
    /// data held in this account
    pub data: Arc<Vec<u8>>,
    /// the program that owns this account. If executable, the program that loads this account.
    pub owner: Pubkey,
    /// this account's data contains a loaded program (and is now read-only)
    pub executable: bool,
    /// the epoch at which this account will next owe rent
    pub rent_epoch: Epoch,
}

`
It didn't seem to quite work easily. I have more Rust to learn still. I was imagining we'd want to move the members of AccountNoData to private and use the AnAccount trait to access them. That allows us later to abstract away that the data is stored in a slice from a program's execution. Moving fields to private was another follow on pr I was imagining shortly after these 2 went in.

jeffwashington: Attempt 2 at Arc would be making AccountNoData a struct with a single private member that was an Arc. Once again, I have not measured this. I don't know how to quantify how much the additional arc affects performance. I know the data Arc helps.

jackcmay: The account is a unit that reflects a whole state, are there cases where we want to specifically copy parts (say lamports) but share others (data)?  If, not it seems more clear to pass a shared copy around and only do a deep copy if necessary.  I'm available Monday if you'd like to talk it out

PR #15731, opened 2021-03-05
Wallclock BankingStage Throttle

carllin:
#### Problem
PoH can drift behind the target wallclock duration, allowing BankingStage to process too many transactions in a slot.

#### Summary of Changes
Add wallclock throttle to BankingStage

Fixes #

### DIFF
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -89,8 +89,7 @@ use std::{
         atomic::{AtomicBool, AtomicU64, Ordering::Relaxed},
         LockResult, RwLockWriteGuard, {Arc, RwLock, RwLockReadGuard},
     },
-    time::{Duration, Instant},
-    sync::Mutex,
+    time::Duration,
 };
 
 pub const SECONDS_PER_YEAR: f64 = 365.25 * 24.0 * 60.0 * 60.0;
@@ -881,8 +880,6 @@ pub struct Bank {
     pub drop_callback: RwLock<OptionalDropCallback>,
 
     pub freeze_started: AtomicBool,
-
-    pub process_transactions_start: Mutex<Option<Instant>>,
 }
 
 impl Default for BlockhashQueue {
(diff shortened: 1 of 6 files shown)


runtime/src/bank.rs
```diff
@@ -4652,6 +4657,19 @@ impl Bank {
...
 
+    // Get the current processing bank if the processing time hasn't expired
+    pub fn is_bank_still_processing_txs(
+        bank_creation_time: &Instant,
+        max_tx_ingestion_time: Option<u64>,
+    ) -> bool {
+        // Do this check outside of the poh lock, hence not a method on PohRecorder
+        max_tx_ingestion_time
+            .map(|max_tx_ingestion_time| {
+                bank_creation_time.elapsed().as_millis() <= max_tx_ingestion_time as u128
```
sakridge, line 4668: why not down-cast millis to u64?

runtime/src/bank.rs
```diff
@@ -4652,6 +4657,19 @@ impl Bank {
                 .is_active(&feature_set::simple_capitalization::id())
     }
 
+    // Get the current processing bank if the processing time hasn't expired
+    pub fn is_bank_still_processing_txs(
```
sakridge, line 4661: The name and comment is kind of confusing to me. Seems like it should be `should_bank_still_be_processing_txs` and the comment like "See if the wallclock time from bank creation to now has exceeded the slot time".

runtime/src/bank.rs
```diff
@@ -1199,6 +1203,7 @@ impl Bank {
             feature_set: new(),
             drop_callback: RwLock::new(OptionalDropCallback(None)),
             freeze_started: AtomicBool::new(fields.hash != Hash::default()),
+            max_tx_ingestion_time: Some(SLOT_MS),
```
sakridge, line 1206: This should come from genesis config, not the constants. Does it need to be on the bank? Why not just pass it as a parameter to banking_stage?

sakridge (approved):

PR #15732, opened 2021-03-05
Add tracer key for tracing transaction path through the network

carllin:
#### Problem
Hard to trace what's going on in forwarding/transaction processing

#### Summary of Changes
Add a special signing key that will be detected in sigverify, will set a signature in the packet metadata, and later used for metric reporting in BankingStage (still TODO in follow-up PR)

Fixes #

### DIFF
--- a/core/src/sigverify.rs
+++ b/core/src/sigverify.rs
@@ -31,33 +31,7 @@ impl Default for TransactionSigVerifier {
 
 impl SigVerifier for TransactionSigVerifier {
     fn verify_batch(&self, mut batch: Vec<Packets>) -> Vec<Packets> {
-        let r = sigverify::ed25519_verify(&mut batch, &self.recycler, &self.recycler_out);
-        mark_disabled(&mut batch, &r);
+        sigverify::ed25519_verify(&mut batch, &self.recycler, &self.recycler_out);
         batch
-    }
-}
-
-pub fn mark_disabled(batches: &mut Vec<Packets>, r: &[Vec<u8>]) {
-    batches.iter_mut().zip(r).for_each(|(b, v)| {
-        b.packets.iter_mut().zip(v).for_each(|(p, f)| {
-            p.meta.discard = *f == 0;
-        })
-    });
-}
-
-#[cfg(test)]
-mod tests {
-    use super::*;
-    use solana_perf::packet::Packet;
-
-    #[test]
-    fn test_mark_disabled() {
-        let mut batch = Packets::default();
-        batch.packets.push(Packet::default());
-        let mut batches: Vec<Packets> = vec![batch];
-        mark_disabled(&mut batches, &[vec![0]]);
-        assert_eq!(batches[0].packets[0].meta.discard, true);
-        mark_disabled(&mut batches, &[vec![1]]);
-        assert_eq!(batches[0].packets[0].meta.discard, false);
     }
 }
--- a/core/src/sigverify.rs
+++ b/core/src/sigverify.rs
@@ -31,7 +31,7 @@ impl Default for TransactionSigVerifier {
 
 impl SigVerifier for TransactionSigVerifier {
     fn verify_batch(&self, mut batch: Vec<Packets>) -> Vec<Packets> {
-        let r = sigverify::ed25519_verify(&batch, &self.recycler, &self.recycler_out);
+        let r = sigverify::ed25519_verify(&mut batch, &self.recycler, &self.recycler_out);
         mark_disabled(&mut batch, &r);
         batch
     }
@@ -39,10 +39,9 @@ impl SigVerifier for TransactionSigVerifier {
 
 pub fn mark_disabled(batches: &mut Vec<Packets>, r: &[Vec<u8>]) {
     batches.iter_mut().zip(r).for_each(|(b, v)| {
-        b.packets
-            .iter_mut()
-            .zip(v)
-            .for_each(|(p, f)| p.meta.discard = *f == 0)
+        b.packets.iter_mut().zip(v).for_each(|(p, f)| {
+            p.meta.discard = *f == 0;
+        })
     });
 }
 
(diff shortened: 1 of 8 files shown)


perf/src/sigverify.rs
```diff
@@ -99,20 +103,32 @@ fn verify_packet(packet: &Packet) -> u8 {
...
+        {
+            packet.meta.tracer_tx_signature = Some(signature);
         }
+
         pubkey_start += size_of::<Pubkey>();
         sig_start += size_of::<Signature>();
     }
-    1
+
+    packet.meta.discard = false;
```
jstarry, line 131: This could potentially cause a discarded packet to be undiscarded. I think it would be better to check above whether the packet is discarded already and only verify if it's not discarded yet just to be safe

perf/src/sigverify.rs
```diff
@@ -320,13 +334,14 @@ pub fn get_checked_scalar(scalar: &[u8; 32]) -> Result<[u8; 32], PacketError> {
 }
 
 pub fn ed25519_verify(
-    batches: &[Packets],
+    batches: &mut [Packets],
```
jstarry, line 337: You will need to mark gpu-verified packets that fail verification as discarded too

core/src/sigverify.rs
```diff
@@ -31,18 +31,17 @@ impl Default for TransactionSigVerifier {
 
 impl SigVerifier for TransactionSigVerifier {
     fn verify_batch(&self, mut batch: Vec<Packets>) -> Vec<Packets> {
-        let r = sigverify::ed25519_verify(&batch, &self.recycler, &self.recycler_out);
+        let r = sigverify::ed25519_verify(&mut batch, &self.recycler, &self.recycler_out);
         mark_disabled(&mut batch, &r);
```
jstarry, line 35: This is redundant

perf/src/sigverify.rs
```diff
@@ -99,20 +103,32 @@ fn verify_packet(packet: &Packet) -> u8 {
...
             &packet.data[pubkey_start..pubkey_end],
             &packet.data[msg_start..msg_end],
         ) {
-            return 0;
+            // Check for tracer pubkey here
+            packet.meta.discard = true;
+            return;
+        }
+
+        if &packet.data[pubkey_start..pubkey_end] == TRACER_KEY.as_ref()
```
jstarry, line 120: I think that lazy static references can be slow, not sure this is good to do for every packet. I think a const or non-lazy static would be better

sdk/src/packet.rs
```diff
@@ -24,6 +24,7 @@ pub struct Meta {
     pub v6: bool,
     pub seed: [u8; 32],
     pub slot: Slot,
+    pub tracer_tx_signature: Option<Signature>,
```
jstarry, line 27: Is this a non-trivial memory increase? Can we instead track a bool for if it's a tracer tx?

jstarry (commented):
Nice, looking forward to this debugging power :) Added some comments about perf / memory

perf/src/sigverify.rs
carllin: So all cases where `packet.meta.discard = true` should early return (parallels the `return 0` from the previous code). Also updated to set discard to false at the beginning of the function instead, as a failsafe

perf/src/sigverify.rs
carllin: those will still get marked as discarded by the old path in `mark_disabled` 

```
let r = verify_shreds_gpu(&batches, &leader_slots, &self.recycler_cache);
sigverify::mark_disabled(&mut batches, &r);
```

core/src/sigverify.rs
carllin: It's still necessary for the gpu case

sdk/src/packet.rs
carllin: yeah you're right. I was trying to avoid deserializing the signature again in BankingStage when reporting the metrics since Sigverify already owns the signature.

carllin: The best thing I could think of is send a `Option<Box<Signature>>`  along with every Packet to BankingStage so that the Packet doesn't accrue extra memory, but the channel would have an extra `size_of` pointer for every packet.

Maybe I should just eat the cost of the extra deserialization in BankingStage...

carllin: Another options is a separate channel just for tracer transactions that takes (Packet,Signature) tuples

jstarry: Extra deserialization in banking stage seems fine if you only do it for packets that have a `tracer` metadata flag

perf/src/sigverify.rs
carllin: Changed to initialize from `const` pubkey bytes

core/src/sigverify.rs
jstarry: Maybe move this inside `sigverify::ed25519_verify` then? Seems wasteful to do the disabled check twice

perf/src/sigverify.rs
jstarry: What's the point of marking as disabled within `verify_shreds` if we are still marking as discarded outside?

perf/src/sigverify.rs
```diff
@@ -773,13 +816,22 @@ mod tests {
             }
 
             // verify packets
-            let ans = sigverify::ed25519_verify(&batches, &recycler, &recycler_out);
-
-            let cpu_ref = ed25519_verify_cpu(&batches);
+            let ans = sigverify::ed25519_verify(&mut batches, &recycler, &recycler_out);
+            ed25519_verify_cpu(&mut batches);
```
jstarry, line 820: I think this test intended to isolate the behavior of `ed25519_verify_cpu` against `sigverify::ed25519_verify` but now that `batches` is passed by mutable reference, this is no longer the case. Feels like batches should be cloned into `sigverify::ed25519_verify` to prevent that.

perf/src/sigverify.rs
jstarry: > Also updated to set discard to false at the beginning of the function instead, as a failsafe

So is it safe to set discard to false at the beginning because we know that `FetchStage` never marks packets as discarded? I think it would be a little safer to just return if discard is true at the top

perf/src/sigverify.rs
carllin: hehe yeah, updated to do that!

jstarry (approved):

PR #15734, opened 2021-03-05
Add more slot update notifications

jstarry:
#### Problem
Want more visibility into block processing and forking

#### Summary of Changes
- Added transaction stats to the `Frozen` notification
- Added `Completed`, `CreatedBank`, and `Dead` notifications

Fixes #


core/src/cluster_slots_service.rs
```diff
@@ -120,13 +130,23 @@ impl ClusterSlotsService {
...
         cluster_info: &ClusterInfo,
+        rpc_subscriptions: &Option<Arc<RpcSubscriptions>>,
     ) {
         let mut slots: Vec<Slot> = vec![];
         while let Ok(mut more) = completed_slots_receiver.try_recv() {
             slots.append(&mut more);
         }
         #[allow(clippy::stable_sort_primitive)]
         slots.sort();
+        if let Some(rpc_subscriptions) = rpc_subscriptions {
```
carllin, line 141: This is run in a loop() with a 200ms sleep, so the responses may be a little delayed. Any reason not to plumb it directly into the blockstore insert?

core/src/replay_stage.rs
```diff
@@ -997,7 +999,17 @@ impl ReplayStage {
         transaction_status_sender: Option<TransactionStatusSender>,
         replay_vote_sender: &ReplayVoteSender,
         verify_recyclers: &VerifyRecyclers,
+        subscriptions: Option<Arc<RpcSubscriptions>>,
     ) -> result::Result<usize, BlockstoreProcessorError> {
+        if bank_progress.replay_progress.num_entries == 0 {
```
carllin, line 1004: hmm this may not be exactly what you want, since it's possible here that `num_entries ==0` both before and after this function call.  

If you want the time when the bank is actually created in `generate_new_bank_forks()`, then at the line `let child_bank = Self::new_bank_from_parent_with_notify(` for validators, and in `maybe_start_leader`, `Self::new_bank_from_parent_with_notify(` for leaders will be the most straightforward places to put the notification

core/src/retransmit_stage.rs
```diff
@@ -235,49 +235,75 @@ fn check_if_already_received(
...
-        } else {
-            None
+) -> bool {
+    if shred_slot <= root_bank.slot() {
+        return false;
+    }
+
+    let mut first_shreds_received_locked = first_shreds_received.lock().unwrap();
+    if !first_shreds_received_locked.contains_key(&shred_slot) {
+        info!("First time receiving a shred from slot: {}", shred_slot);
```
carllin, line 267: Could we actually make this a datapoint, I think it would be useful for debugging purposes 😃

core/src/cluster_slots_service.rs
jstarry: No particular reason, I'll move into blockstore

core/src/replay_stage.rs
carllin: Also of note, the only time this will be dramatically different than the first data shred received should be if replay is behind, i.e. still replaying the previous slot/different fork when the next slot's data shred has already arrived.

jstarry: Ah good point. My intent was to get a timestamp for when a node actually started replaying the first entry of a block which I think I could do by moving this into `confirm_slot` and checking if entries isn't empty. Does that feel useful?

Regardless, sending a notification for when the bank is created seems more useful since it indicates when a node decided to build off another fork

jstarry: > Also of note, the only time this will be dramatically different than the first data shred received should be if replay is behind, i.e. still replaying the previous slot/different fork when the next slot's data shred has already arrived.

Well, my hope was to capture any issues with turbine here where replay got delayed due to slow shred propagation. Sounds like the only way to do that is to capture whether or not the node is behind with a "block start" notification.

jstarry: Replaced this with a `CreatedBank` notification

core/src/cluster_slots_service.rs
carllin: hmmm looking at it now the `push_epoch_slots()` may actually be slow sometimes due to some 

1) Contention on the cluster_info lock in `self.time_gossip_read_lock` (this has been known to spike)
2) Serializing/deserializing epoch slots

It's kind of ugly, but we could plumb another thread in between here that reads from completed_slots_receiver and does notification...

core/src/rpc_subscriptions.rs
```diff
@@ -952,6 +952,11 @@ impl RpcSubscriptions {
 
     pub fn notify_slot(&self, slot: Slot, parent: Slot, root: Slot) {
         self.enqueue_notification(NotificationEntry::Slot(SlotInfo { slot, parent, root }));
+        self.enqueue_notification(NotificationEntry::SlotUpdate(SlotUpdate::CreatedBank {
```
carllin, line 955: nice, repurposing existing things 😃

core/src/cluster_slots_service.rs
jstarry: Done!

core/src/rpc_completed_slots_service.rs
```diff
@@ -0,0 +1,54 @@
...
+                {
+                    Ok(slots) => slots,
+                    Err(RecvTimeoutError::Timeout) => continue,
+                    Err(RecvTimeoutError::Disconnected) => {
+                        warn!("RPC completed slots service - sender disconnected");
+                        break;
+                    }
+                };
+
+                while let Ok(mut more) = completed_slots_receiver.try_recv() {
```
carllin: is this equivalent to `for more in completed_slots_receiver.iter()`

carllin (dismissed):

CriesofCarrots:
@jstarry , sorry for necroposting, but I'm trying to understand `Bank::transaction_entries_count`. Is it actually the count of (executed) transaction batches?
It doesn't seem like it has anything to do with `Entry`s. Or am I missing something?

jstarry:
No worries! @CriesofCarrots it looks like this code assumes that transaction entries and transaction batches are the same thing. I think that's true for the leader, but during replay validators could process a few transaction entries together as a batch. Is that what you're getting at?

CriesofCarrots:
> this code assumes that transaction entries and transaction batches are the same thing. I think that's true for the leader, during replay validators could process a few transaction entries together as a batch. Is that what you're getting at?

Oh right. I realized that the each of a leader's batches represent the txs in one Entry, but didn't really think of that as the reason for this field name.
Yeah, that's exactly what I'm getting at; this value would not necessarily be consistent between a leader and other validators. Okay, thank you for confirming!

PR #15739, opened 2021-03-05
fix: add TokenAmount.uiAmountString

CriesofCarrots:
#### Problem
New TokenAmount.uiAmountString field is not exposed

#### Summary of Changes
Add it to web3.js struct/types
Allow TokenAmount.uiAmount to be null


oJshua (approved):
Yes, this will work!

PR #15742, opened 2021-03-05
Explorer: support uiAmountString api updates

oJshua:
#### Problem
We have deprecated uiAmount in turn for uiAmountString to resolve some precision issues.

#### Summary of Changes
Use uiAmountString in place of uiAmount.

* Waiting on web3 build and release


CriesofCarrots (approved):
ty!

PR #15753, opened 2021-03-06
SoM: Avoid destaking all nodes when there are no quality block producers

mvines:
The `--max-poor-block-producer-percentage` safely belt doesn't work when ALL nodes are classified as poor.   

Thankfully due to an RPC error, the https://buildkite.com/solana-labs/stake-o-matic/builds/456#86b06df8-014d-4330-9623-9817cfe8da56 job from yesterday failed instead of accidentally removing stake from all nodes

### DIFF
--- a/stake-o-matic/src/main.rs
+++ b/stake-o-matic/src/main.rs
@@ -617,8 +617,8 @@ where
 
 type BoxResult<T> = Result<T, Box<dyn error::Error>>;
 
-///                    quality          poor             cluster_skip_rate
-type ClassifyResult = (HashSet<Pubkey>, HashSet<Pubkey>, usize);
+///                    quality          poor             cluster_skip_rate, too_many_poor_block_producers
+type ClassifyResult = (HashSet<Pubkey>, HashSet<Pubkey>, usize, bool);
 
 fn classify_producers(
     first_slot: Slot,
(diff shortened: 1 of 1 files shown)


stake-o-matic/src/main.rs
```diff
@@ -1134,8 +1134,11 @@ fn main() -> Result<(), Box<dyn error::Error>> {
     let (quality_block_producers, poor_block_producers, cluster_average_skip_rate) =
         classify_block_producers(&rpc_client, &config, last_epoch)?;
 
-    let too_many_poor_block_producers = poor_block_producers.len()
-        > quality_block_producers.len() * config.max_poor_block_producer_percentage / 100;
+    let too_many_poor_block_producers = if quality_block_producers.is_empty() {
+        true
+    } else {
+        poor_block_producers.len() > quality_block_producers.len() * config.max_poor_block_producer_percentage / 100
```
mvines, line 1140: When quality_block_producers.is_empty(), we previously had:
`poor_block_producers.len() > 0 * config.max_poor_block_producer_percentage / 100`

So no quality block producers means that `too_many_poor_block_producers` is false and we destake the world!

stake-o-matic/src/main.rs
```diff
@@ -1134,8 +1134,12 @@ fn main() -> Result<(), Box<dyn error::Error>> {
     let (quality_block_producers, poor_block_producers, cluster_average_skip_rate) =
         classify_block_producers(&rpc_client, &config, last_epoch)?;
 
-    let too_many_poor_block_producers = poor_block_producers.len()
-        > quality_block_producers.len() * config.max_poor_block_producer_percentage / 100;
+    let too_many_poor_block_producers = if quality_block_producers.is_empty() {
+        true
+    } else {
+        poor_block_producers.len()
+            > quality_block_producers.len() * config.max_poor_block_producer_percentage / 100
```
sakridge, line 1141: extract to function & unit test?

sakridge (approved):
logic looks right

sakridge:
Would it have been just bonus stake? or all stake?

stake-o-matic/src/main.rs
mvines: done

mvines:
> Would it have been just bonus stake? or all stake?

just bonus I think, ~◎85m 😬

PR #15754, opened 2021-03-07
Remove old feature: cumulative_rent_related_fixes

ryoqun:
#### Problem

There is important planned rent change (will backport to v1.5) with another feature gate; and there are too many feature flags.

#### Summary of Changes

Clean up before the next one!

Because of the blocked change, this clean will also be v1.5-back-ported.

Fixes #

### DIFF
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -3167,7 +3167,6 @@ impl Bank {
             &self.rent_collector,
             &self.last_blockhash_with_fee_calculator(),
             self.fix_recent_blockhashes_sysvar_delay(),
-            self.cumulative_rent_related_fixes_enabled(),
         );
         self.collect_rent(executed, loaded_accounts);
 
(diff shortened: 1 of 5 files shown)


ryoqun:
(Also note that I also need to remove the `simple_capitalization` as the next and last preparatory cleanup pr)

ryoqun:
@t-nelson could you sanity-check this? I've tested this against testnet and mainnet-beta using the live-cluster tests: https://github.com/solana-labs/solana/pull/12175/files#diff-ed9d8e9c9bc044a65944d919be16361b1969bca5122e281ff644a0bb4cbfa198L58

https://buildkite.com/solana-labs/solana/builds/41661#231c7504-d8b7-443a-b43e-e1e2ffac885f

t-nelson (approved):
lgtm!

runtime/src/bank.rs
```diff
@@ -10227,19 +10185,19 @@ pub(crate) mod tests {
             if bank.slot == 32 {
                 assert_eq!(
                     bank.hash().to_string(),
-                    "4syPxVrVFUpksTre5BB5w7qd3BxSU4WzUT6R2fjFgMJ2"
+                    "7AkMgAb2v4tuoiSf3NnVgaBxSvp7XidbrSwsPEn4ENTp"
```
ryoqun, line 10188: For later reference, this is safe because I'm aware the specific behaviors of rent collection is changed by making some gated behavior now default.

PR #15758, opened 2021-03-08
Add more Poh metrics

sakridge:
#### Problem

* batch size 64 was a tps regression
* unit test doesn't accurately reflect bench-tps or real system load.

#### Summary of Changes

* Revert to batch-size=1
* Add mode to par_iter for record to show lock contention and induce CPU load.

Fixes #


genesis/src/main.rs
```diff
@@ -462,7 +462,7 @@ fn main() -> Result<(), Box<dyn error::Error>> {
             ClusterType::Development => {
                 let hashes_per_tick =
                     compute_hashes_per_tick(poh_config.target_tick_duration, 1_000_000);
-                poh_config.hashes_per_tick = Some(hashes_per_tick);
+                poh_config.hashes_per_tick = Some(hashes_per_tick / 2);
```
jeffwashington, line 465: @sakridge what is this?

sakridge: That was experimenting with lowering the target rate to 1/2 the peak PoH rate to test in bench-tps. We can revert it for now.

sakridge:
added with https://github.com/solana-labs/solana/pull/15931

PR #15763, opened 2021-03-08
Remove old feature: simple_capitalization

ryoqun:
#### Problem

There is important planned rent change (will backport to v1.5) with another feature gate; and there are too many feature flags.

#### Summary of Changes

Clean up before the next one!

Because of the blocked change, this clean will also be v1.5-back-ported.

this is 2 of 2 of this old feature cleaning: https://github.com/solana-labs/solana/pull/15754

Also, this passed live-cluter test.

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -3489,25 +3489,10 @@ impl AccountsDb {
 
     pub fn account_balance_for_capitalization(
         lamports: u64,
-        owner: &Pubkey,
-        executable: bool,
-        simple_capitalization_enabled: bool,
+        _owner: &Pubkey,
+        _executable: bool,
     ) -> u64 {
-        if simple_capitalization_enabled {
-            return lamports;
-        }
-
-        let is_specially_retained = (solana_sdk::native_loader::check_id(owner) && executable)
-            || solana_sdk::sysvar::check_id(owner);
-
-        if is_specially_retained {
-            // specially retained accounts always have an initial 1 lamport
-            // balance, but could be modified by transfers which increase
-            // the balance but don't affect the capitalization.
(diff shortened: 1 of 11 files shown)


runtime/src/accounts_db.rs
```diff
@@ -3487,35 +3487,20 @@ impl AccountsDb {
         AccountsHash::checked_cast_for_capitalization(balances.map(|b| b as u128).sum::<u128>())
     }
 
+    // remove this by inlining and remove extra unused params upto all callchain
```
ryoqun, line 3490: @jeffwashington FYI, this is possible further optimization opportunity for accounts hash calculation. I'm leaving this alone for now because I want to minimize this pr (this needs to be back-ported to v1.5)

runtime/src/bank.rs
```diff
@@ -7001,7 +6923,11 @@ pub(crate) mod tests {
         // not being eagerly-collected for exact rewards calculation
         bank.restore_old_behavior_for_fragile_tests();
 
-        assert_eq!(bank.capitalization(), 42 * 1_000_000_000);
+        let sysvar_and_native_proram_delta = 10;
```
ryoqun, line 6926: there are bunch of similar hard coded magic numbers in this pr; fyi, I'll soon replace them once I create the next and final pr to change the rent calculation for sysvars.

ryoqun: ... so, please bear with this for now. ;)

ryoqun:
@t-nelson similar to #15754, could you sanity-check this? Thanks!

t-nelson (approved):
LGTM! (so long as those magic numbers go away in PR++ :wink:)

PR #15777, opened 2021-03-09
Use trait for Account data

jeffwashington:
#### Problem
We need to abstract how accounts store data.
#### Summary of Changes

Fixes #


sdk/src/account.rs
```diff
@@ -436,6 +450,53 @@ impl AccountSharedData {
     pub fn serialize_data<T: serde::Serialize>(&mut self, state: &T) -> Result<(), bincode::Error> {
         shared_serialize_data(self, state)
     }
+
```
jeffwashington, line 453: @carllin here was a first pass at changing constructors

PR #15780, opened 2021-03-09
Account Data Push from a test

jeffwashington:
#### Problem
A test does a push on account.data. data will become abstracted and copy on write. Push is a complicated operation to support efficiently and we don't want to make it first class.

#### Summary of Changes
Add helper method to isolate data push for the test so we can change how account data is represented.

Fixes #


carllin:
Hmm instead of making a new function for each of these operations on `data`, would it be easier to make the`ReadableAccounts/WritableAccounts` derive `Deref` and `DerefMut`  to get `&Vec<>` and `&mut Vec<>` references and then directly call the vector methods on those?

PR #15784, opened 2021-03-10
Default to the BPF JIT. Use the `solana-validator --no-bpf-jit` flag to disable

mvines:
1.6 seems like a good release train to enable the BPF JIT by default.  We've had mainnet-beta nodes running with the JIT enabled for months now with no real issues.  It's time!


jackcmay (approved):

Lichtso (commented):
Looks good

PR #15785, opened 2021-03-10
Improve load_largest_accounts more

### LINKED ISSUE #15711: load_largest_accounts() consumes a lot of memory
#### Problem

Loading the largest accounts consumes 40 bytes per account.

#### Proposed Solution

Use a fixed size max heap, such that it only consumes `40 * num` bytes.

CriesofCarrots:
#### Problem
Very fine post-merge suggestions on #15775 

#### Summary of Changes
- Add bench for `load_largest_accounts`; 3 runs of 10_000 accounts with random lamport balances:
Before: 6.546ms
After: 6.058ms
- Use `min-heap` BinaryHeap
- Add unit test in accounts.rs

Fixes #15711 more

### DIFF
--- a/runtime/src/accounts.rs
+++ b/runtime/src/accounts.rs
@@ -542,11 +542,15 @@ impl Accounts {
             .scan_accounts(ancestors, |collector: &mut Vec<(Pubkey, u64)>, option| {
                 if let Some(data) = option
                     .filter(|(pubkey, account, _)| {
-                        let should_include_pubkey = match filter {
-                            AccountAddressFilter::Exclude => !filter_by_address.contains(&pubkey),
-                            AccountAddressFilter::Include => filter_by_address.contains(&pubkey),
-                        };
-                        should_include_pubkey && account.lamports != 0
+                        account.lamports != 0
+                            && match filter {
+                                AccountAddressFilter::Exclude => {
+                                    !filter_by_address.contains(&pubkey)
+                                }
+                                AccountAddressFilter::Include => {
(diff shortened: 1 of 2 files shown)


garious (approved):
Awesome, that looks great! Thank you!

runtime/src/accounts.rs
```diff
@@ -538,36 +539,45 @@ impl Accounts {
...
+                                AccountAddressFilter::Exclude => {
+                                    !filter_by_address.contains(&pubkey)
+                                }
+                                AccountAddressFilter::Include => {
+                                    filter_by_address.contains(&pubkey)
+                                }
+                            }
                     })
-                    .map(|(pubkey, account, _slot)| (*pubkey, account.lamports))
+                    .map(|(pubkey, account, _slot)| (account.lamports, *pubkey))
```
garious, line 560: Nit: put `&Pubkey` in the min-heap instead of `Pubkey`. Only copy at the very end where you do that `map(|Reverse(...)|)`.

CriesofCarrots: Hmm, the borrow-checker's not too keen on that one, as the `&Pubkey` comes from `option`, which can't escape the closure:
```
`option` is a reference that is only valid in the closure body
```
Do you know any Rust tricks that might work here?

garious: I see.  Best we can do is only copy when adding to the collector.  I'll push a patch shortly.

CriesofCarrots: ty!

PR #15790, opened 2021-03-10
AccountSharedData construction changes

jeffwashington:
#### Problem
We construct accounts in many combinations of the 5 fields.
We want to abstract away the data field and hide it. Doing so requires allowing construction somehow.
#### Summary of Changes
Reuse the existing Account struct and AccountSharedData::from() to allow rust-like construction of AccountSharedData without adding new structs or methods.
Fixes #

### DIFF
--- a/account-decoder/src/lib.rs
+++ b/account-decoder/src/lib.rs
@@ -224,10 +224,10 @@ mod test {
     fn test_base64_zstd() {
         let encoded_account = UiAccount::encode(
             &Pubkey::default(),
-            AccountSharedData {
+            AccountSharedData::from(Account {
                 data: vec![0; 1024],
-                ..AccountSharedData::default()
-            },
+                ..Account::default()
+            }),
             UiAccountEncoding::Base64Zstd,
             None,
             None,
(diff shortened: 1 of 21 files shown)


garious:
Using `..AccountSharedData::default()` looks cleaner to me:

```rust
        let account = AccountSharedData {
            lamports: 1,
            ..AccountSharedData::default()
        };
```

If you add another field, no callers needs to change, whereas with your proposal, they all would need to add a `None` parameter. Also, it's not clear what `None` means, unless you're using an IDE that adds the `lamports:` label for you.  Also, an optional parameter would need to be double-wrapped (i.e. Some(Some(42))).

A common pattern elsewhere in the codebase is to add a "Config" struct, which allows you to make all of the original struct members private.  For example, in this case, you could have a `AccountSharedData` with private field `lamports` and a `AccountSharedDataConfig` with public field `lamports`, and static method `AccountSharedData::new_with_config(AccountSharedDataConfig)`.  Usage:

```rust
    let account = AccountSharedData::new_with_config(
        AccountSharedDataConfig {
            lamports: 1,
            ..AccountSharedDataConfig::default()
        }
    );
```

jeffwashington:
@garious and I arrived at:
let account = AccountSharedData::from(
        Account {
            lamports: 1,
            ..Account::default()
        }
    );

garious (approved):

PR #15793, opened 2021-03-10
limits CrdsGossipPull::pull_request_time size

behzadnouri:
#### Problem
There is no pruning logic on CrdsGossipPull::pull_request_time
https://github.com/solana-labs/solana/blob/79ac1997d/core/src/crds_gossip_pull.rs#L172-L174
potentially allowing this to take too much memory.

Additionally, CrdsGossipPush::last_pushed_to is pruning recent push
timestamps:
https://github.com/solana-labs/solana/blob/79ac1997d/core/src/crds_gossip_push.rs#L275-L279
instead of the older ones.

#### Summary of Changes
* Changed the `CrdsGossipPull::pull_request_time` type from `HashMap` to `LruCache`.
* Similar logic for `CrdsGossipPush::last_pushed_to`.
* Limit the impact of last push/pull time in push/pull options.

### DIFF
--- a/core/src/cluster_info.rs
+++ b/core/src/cluster_info.rs
@@ -4763,18 +4763,32 @@ mod tests {
             rand_ci.shred_version = shred_version;
             rand_ci.wallclock = timestamp();
             CrdsValue::new_signed(CrdsData::ContactInfo(rand_ci), &keypair)
-        }).take(NO_ENTRIES).collect();
+        })
+        .take(NO_ENTRIES)
+        .collect();
         let timeouts = cluster_info.gossip.read().unwrap().make_timeouts_test();
-        assert_eq!((0, 0, NO_ENTRIES), cluster_info.handle_pull_response(
-            &entrypoint_pubkey,
-            data,
-            &timeouts
-        ));
+        assert_eq!(
+            (0, 0, NO_ENTRIES),
+            cluster_info.handle_pull_response(&entrypoint_pubkey, data, &timeouts)
+        );
 
         let now = timestamp();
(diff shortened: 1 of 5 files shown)


behzadnouri:
> I think it looks good. I think we might need a test for it though.

@sakridge added more test coverage

behzadnouri:
@sakridge also merged in @utsl42 commit adding more test coverage.

core/src/cluster_info.rs
```diff
@@ -115,7 +115,7 @@ pub const DEFAULT_CONTACT_SAVE_INTERVAL_MILLIS: u64 = 60_000;
 /// Minimum serialized size of a Protocol::PullResponse packet.
 const PULL_RESPONSE_MIN_SERIALIZED_SIZE: usize = 161;
 // Limit number of unique pubkeys in the crds table.
-const CRDS_UNIQUE_PUBKEY_CAPACITY: usize = 4096;
+pub(crate) const CRDS_UNIQUE_PUBKEY_CAPACITY: usize = 4096;
```
sakridge, line 118: Could we handle 8k?

behzadnouri: This was mentioned in: https://github.com/solana-labs/solana/pull/15539#discussion_r590519612
It will have 830x impact on size of the crds table. So I am worried that would be too much. For now this is 4x tds cluster and 6x mainnet.

sakridge: Ok.

sakridge (approved):
lgtm

PR #15796, opened 2021-03-10
Unpack snapshot AppendVecs directly into account paths

mvines:
Unpacking snapshot archives are quite inefficient when the account paths are located on a different drive than the ledger itself.   In this case the append vecs are first unpacked onto the ledger drive, and then copied over to the account drive(s).   Instead unpack the append vecs directory into the account drive(s).

Quick test on `bvb` showed a ~35 second reduction in mainnet-beta validator startup time.


mvines:
@ryoqun or @carllin, wdyt?

runtime/src/hardened_unpack.rs
```diff
@@ -139,7 +146,7 @@ where
 
         // unpack_in does its own sanitization
         // ref: https://docs.rs/tar/*/tar/struct.Entry.html#method.unpack_in
-        check_unpack_result(entry.unpack_in(&unpack_dir)?, path_str)?;
+        check_unpack_result(entry.unpack_in(unpack_dir)?, path_str)?;
```
carllin, line 149: looks safe to me 😃

PR #15800, opened 2021-03-11
AccountsSharedData: data copy on write

jeffwashington:
#### Problem
This pr gives us copy on write for AccountSharedData's 'data' vec. The result is improved performance at least on replay and transaction execution.

This is the version where AccountSharedData has:
`Arc<Vec<u8>>`
as compared to:
AccountData has:
`Arc<Vec<u8>>`[PR](https://github.com/solana-labs/solana/pull/15850)
From various discussions, it seems the preferred approach is this one. Arc<AccountData>
#### Summary of Changes
AccountSharedData uses Arc<Vec<u8>> instead of Vec<u8>.
Fixes #


carllin:
@jeffwashington what if it looked something like this? https://github.com/carllin/solana/commit/e540cb5a3c73ae6e3f9455110f037819579295b6

sdk/src/account.rs
```diff
@@ -505,12 +517,12 @@ impl solana_program::account_info::Account for Account {
         )
     }
 }
-
+/*
 /// Create `AccountInfo`s
```
jeffwashington, line 521: this comment gets me a link

sdk/src/account.rs
```diff
@@ -40,6 +41,142 @@ pub struct AccountSharedData {
     pub rent_epoch: Epoch,
 }
 
+impl<'de> Deserialize<'de> for AccountSharedData {
```
jeffwashington, line 44: This is not desirable.

jeffwashington: I found a simpler way to implement serialize/deserialize with minimal performance impact and not a lot of new code.

carllin (dismissed):

jeffwashington:
> @jeffwashington what if it looked something like this? [carllin@e540cb5](https://github.com/carllin/solana/commit/e540cb5a3c73ae6e3f9455110f037819579295b6)

Carl and I discussed. This api change can be made later if we need to. For now, the 2 users are test code only and the solution is localized to the test code and not first class in the api.

sdk/src/account.rs
sakridge: Does it need to be serialize/deserialize? Maybe it makes sense for it to be in another module and then that module excluded from contract code.

It can be in this list as `account_shared_data` module for instance:
```
default = [
  "full" # functionality that is not compatible or needed for on-chain programs
]
full = [
    "assert_matches",
    "byteorder",
    ```

Comment (behzadnouri) [reply sdk/src/account.rs:44]: This might need some debugging what is going on. We have examples of `Serialize/Deserialize` with inner `Arc` fields which work fine:
https://github.com/solana-labs/solana/blob/ff2242d0e/runtime/src/epoch_stakes.rs#L15-L21

Comment (sakridge) [reply sdk/src/account.rs:44]: I think it's compiling in the BPF toolchain environment. Because Account is used for both, so everything in this module would get compiled for both. That environment has more limitations than the normal native dev environment.

Comment (sakridge) [reply sdk/src/account.rs:44]: If we don't need something in the BPF environment, we should probably move it out if we can.

Comment (jeffwashington) [reply sdk/src/account.rs:44]: @behzadnouri 
[here's a failing build:](https://buildkite.com/solana-labs/solana/builds/42372#cf928dd8-25a9-4a5c-9500-696e3e4a7d45)

Comment (jeffwashington) [reply sdk/src/account.rs:44]: I will try a different approach where we separate it out. it looks like it doesn't need to serialize.

Comment (jeffwashington) [reply sdk/src/account.rs]: I have removed serialization completely. Moving out AccountSharedData is a different issue. Yay for the pr process!

File: sdk/src/account_data.rs [hunk 3]
```diff
@@ -0,0 +1,103 @@
...
+use std::{
+    ops::{Deref, DerefMut, Index, IndexMut},
+    slice::SliceIndex,
+};
+
+#[derive(Debug, Default)]
+pub struct AccountData {
+    data: Vec<u8>,
+}
+
```
behzadnouri, line 13: Per our previous discussions, I will integrate https://github.com/solana-labs/solana/pull/15756 
once you are done with your changes. It will be easier if you just keep `data` in `AccountSharedData` as `Arc<Vec<u8>>` as in your initial plans and what we discussed.

Can you please remove this `account_data.rs` from this pull request?

sdk/src/account.rs
jeffwashington: Could not make AccountSharedData unavailable in bpf due to KeyedAccounts.

sakridge (approved):
code looks good. interested in bench results

jeffwashington:
bench tps on colo:
this pr:
    Highest TPS: 55541.73 sampling period 1s max transactions: 2862723 clients: 1 drop rate: 0.89
[2021-03-23T20:00:36.924698248Z INFO  solana_bench_tps::bench] 	Average TPS: 31444.836

master:
    Highest TPS: 54838.24 sampling period 1s max transactions: 2921572 clients: 1 drop rate: 0.89
[2021-03-23T20:07:49.986106035Z INFO  solana_bench_tps::bench] 	Average TPS: 32079.297

PR #15812, opened 2021-03-12
Add trait for saturating arithmetic

CriesofCarrots:
#### Problem
The `Duration` type has saturating arithmetic coming, but they are not yet stable. We should include Durations in our to endeavor to prevent overflowing, but it is cumbersome to use `.checked_MATH().unwrap_or()` everywhere it's needed, especially as `Duration` consts MAX/ZERO are similarly not yet stable.

#### Summary of Changes
Add a helper trait with saturating-math methods that will be easy to replace with the standard methods when stable.

Addresses https://github.com/solana-labs/solana/pull/15736#discussion_r590890444
Toward #15368

### DIFF
--- a/sdk/src/arithmetic.rs
+++ b/sdk/src/arithmetic.rs
@@ -1,0 +1,45 @@
+use std::time::Duration;
+
+/// A helper trait for primitive types that do not yet implement saturating arithmetic methods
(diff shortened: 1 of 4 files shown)


t-nelson (approved):
Awesome!  These will come in handy.

PR #15820, opened 2021-03-12
explorer: Filter program search results by cluster

jstarry:
#### Problem
Program search isn't smart enough to filter out programs that don't exist on non-mainnet clusters.

#### Summary of Changes
- Annotate each program with which clusters it's deployed on
- Add labels for spl-feature-proposal, raydium and bonfida
- Add label for the instructions sysvar

Fixes #


jstarry:
@bartosz-lipinski I think we'll want a program registry as well fyi

oJshua (approved):
Much needed!

Issue #15817, opened 2021-03-12
rpc: Add options for limiting getConfirmedBlock response data


jstarry:
#### Problem
The `getConfirmedBlock` RPC response can be quite massive when considering transaction details and metadata as well as a ton of reward information for blocks on the epoch boundary. Not all consumers will need this information but currently everyone has to pay the cost of bandwidth and response time.

#### Proposed Solution
1. Add an option to not return reward info
2. Add an option to return either full tx details, only signatures, or nothing at all.

mvines:
The "nothing at all" case for transactions in a block would speed up the rewards output in `solana vote-account` and `solana stake-account` -- these two commands are getting slower and slower for each new epoch post-inflation.   This is probably worth spinning off into another issue, once this issue is fixed, assuming it doesn't happen at the same time

PR #15827, opened 2021-03-12
Rpc: support extended config for getConfirmedBlock

CriesofCarrots:
#### Problem
#15817 requests customizing `getConfirmedBlock` responses, but the current request structure is difficult to extended with additional optional parameters.

#### Summary of Changes
- Add a wrapper enum that supports the previous encoding input (string) as well as an alternate config object, which can be extended with additional optional parameters

### DIFF
--- a/client/src/rpc_config.rs
+++ b/client/src/rpc_config.rs
@@ -111,24 +111,48 @@ pub struct RpcGetConfirmedSignaturesForAddress2Config {
 
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 #[serde(untagged)]
-pub enum RpcConfirmedBlockConfigWrapper {
+pub enum RpcEncodingConfigWrapper<T> {
     Deprecated(Option<UiTransactionEncoding>),
-    Current(Option<RpcConfirmedBlockConfig>),
+    Current(Option<T>),
 }
 
-impl RpcConfirmedBlockConfigWrapper {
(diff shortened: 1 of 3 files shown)


mvines (approved):
Backport to v1.5 so we can start flushing out the old version sooner?

For a smooth client-side rollout, I feel like we'll need node version detection logic in web3.js like we have in solana-client with `RpcClient::get_node_version()` and friends.

CriesofCarrots:
> For a smooth client-side rollout, I feel like we'll need node version detection logic in web3.js like we have in solana-client with `RpcClient::get_node_version()` and friends.

Agreed, I'll file an issue.

Incidentally, with these changes, `getConfirmedTransaction` is the only remaining method that uses the positional string encoding-parameter. While we don't have an immediate need for more control over those responses as we do for confirmed-block, I'm inclined to give it this same treatment now for more consistency across methods. Any objections @mvines?

CriesofCarrots:
> For a smooth client-side rollout, I feel like we'll need node version detection logic in web3.js like we have in solana-client with `RpcClient::get_node_version()` and friends.

Agreed, I'll file an issue.

Incidentally, with these changes, `getConfirmedTransaction` is the only remaining method that uses the positional string encoding-parameter. While we don't have an immediate need for more control over those responses as we do for confirmed-block, I'm inclined to give it this same treatment now for more consistency across methods. Any objections @mvines?

PR #15829, opened 2021-03-12
Change poh to not use mutex for mixin

jeffwashington:
#### Summary of Changes

Fixes #


jeffwashington:
metrics: 25M loop that adds 1 to a u32 and does some sync thing:
mutex locks: 222ms
try_recv: 74ms
try_recv with 25M ints pushed to it: 386ms
try_recv with 1 int pushed to it: 143ms

PR #15840, opened 2021-03-13
Separate snapshot location

diman-io:
#### Problem

The size of unarchived snapshot is about 7.5 Gb
1 snapshot per 10 min = 7.5 * 6 * 24 = 1080 Gb writing to disk per day

But they rotate.
And in one time it use 7.5 Gb * 3 (2 - completed, 1 - in progress) + 300 Mb (`snapshots` folder).
Total ~ 22 Gb

Idea is to use ramdisk for them.

#### Summary of Changes

I have added option `--snapshots` (similar `--accounts`).
I have done changes on v1.5.14 (commit 7f1368e7925d388ec0c6978dfdf282106b0f44ba).
You could use https://github.com/solana-labs/solana/compare/v1.5.14...diman-io:separate-snapshot-location?expand=1 for see clear diff. I'm able to move changes to 1.7.x

It works at my no-voting node now in mainnet 😎

Fixes #


mvines:
Nice I've been wanting this for a while, can you rebase on the tip of the master branch so we can get this PR reviewed and landed

mvines:
> The size of unarchived snapshot is about 7.5 Gb

I think you want to be using the default zstd-compressed snapshots these days.  The extra CPU needed to compress the tarball is typically a better trade-off than the IOPS cost of writing the uncompressed snapshot.  But even at 1.7 Gb for a compressed snapshot, being able to relocate the snapshot elsewhere is nice.

If you store the snapshot in a ramdisk, you'll probably want some other process that:
1. Saves the snapshot to persistent storage before a system reboot
2. Less frequently saves the snapshot to persistent storage in case the system crashes or power is lost
so that in those cases it won't be necessary to download a snapshot from some other mainnet-beta node

diman-io:
> Nice I've been wanting this for a while, can you rebase on the tip of the master branch so we can get this PR reviewed and landed

Done.

mvines:
@diman-io - hey I rebased this PR into your github fork for you.  You had a merge commit in there

validator/src/main.rs
```diff
@@ -1203,6 +1204,13 @@ pub fn main() {
                 .multiple(true)
                 .help("Path to accounts shrink path which can hold a compacted account set."),
         )
+        .arg(
+            Arg::with_name("snapshots_path")
+                .long("snapshots")
+                .value_name("PATHS")
```
mvines, line 1210:
```suggestion
                .value_name("DIR")
```

runtime/src/snapshot_utils.rs
```diff
@@ -206,11 +207,12 @@ fn get_archive_ext(archive_format: ArchiveFormat) -> &'static str {
...
         for entry in entries.filter_map(|entry| entry.ok()) {
-            if entry
+            let filename = entry
                 .file_name()
                 .into_string()
-                .unwrap_or_else(|_| String::new())
-                .starts_with(TMP_SNAPSHOT_PREFIX)
+                .unwrap_or_else(|_| String::new());
+            if filename.starts_with(TMP_SNAPSHOT_PREFIX)
+                | filename.ends_with(TMP_SNAPSHOT_EXTENSION)
```
mvines, line 215: I'm a little uncomfortable with deleting *.tmp from wherever `--snapshot-dir` is aimed at, can you explain the issue that this line is aiming to resolve?

validator/src/main.rs
```diff
@@ -2058,7 +2066,12 @@ pub fn main() {
 
     let snapshot_interval_slots = value_t_or_exit!(matches, "snapshot_interval_slots", u64);
     let maximum_local_snapshot_age = value_t_or_exit!(matches, "maximum_local_snapshot_age", u64);
-    let snapshot_path = ledger_path.join("snapshot");
+    let snapshot_output_dir = if matches.is_present("snapshots_path") {
+        PathBuf::from(matches.value_of("snapshots_path").unwrap())
+    } else {
+        ledger_path.clone()
+    };
+    let snapshot_path = snapshot_output_dir.join("snapshot");
```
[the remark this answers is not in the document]
diman-io: I don't understand ((
`snapshot_path` contains path to `snapshot` folder in `ledger` folder or in new `snapshots_path`-folder

In ledger:
image

In separate folder:
image

runtime/src/snapshot_utils.rs
diman-io: tmp is extension of snapshot while it downloaded. If process was halted, than tmp file stays here forever. I saw 10 files one time)

When we use `--snapshot-dir` could are there other files with tmp extension in this separate folder?
But I don't sure that's true if we don't use `--snapshot-dir`.

You are right. Even if there are no these files now, but it could appear in future.

What about this?
```
const SNAPSHOT_PREFIX: &str = "snapshot-";

if filename.starts_with(TMP_SNAPSHOT_PREFIX)
    | filename.starts_with(SNAPSHOT_PREFIX) & filename.ends_with(TMP_SNAPSHOT_EXTENSION)
```

mvines: What do you think about https://github.com/solana-labs/solana/pull/15939 as an alternative to changing `remove_tmp_snapshot_archives()`

validator/src/main.rs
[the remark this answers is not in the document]
mvines: Ah, yes.  You're right.  I forgot that's how it works!  Nevermind this comment :)

validator/src/main.rs
```diff
@@ -1203,6 +1204,13 @@ pub fn main() {
                 .multiple(true)
                 .help("Path to accounts shrink path which can hold a compacted account set."),
         )
+        .arg(
+            Arg::with_name("snapshot_dir")
+                .long("snapshot-dir")
```
mvines, line 1209: > I have added option --snapshots (similar --accounts).

Looking at this now and re-reading what you wrote in the PR description, I completely agree that `--snapshots` is a nicer name for this parameter.   Sorry for the wrong turn
```suggestion
            Arg::with_name("snapshots")
                .long("snapshots")
```

runtime/src/snapshot_utils.rs
diman-io: I like it. More universal.

diman-io: I have reverted this file to origin

PR #15843, opened 2021-03-14
explorer: Fix midnight timestamps

jstarry:
#### Problem
Timestamps at midnight display as "24:00:00" instead of "00:00:00"

Before

Screen Shot 2021-03-14 at 10 57 03 AM

After

Screen Shot 2021-03-14 at 10 57 10 AM

### DIFF
--- a/explorer/src/utils/date.ts
+++ b/explorer/src/utils/date.ts
@@ -12,7 +12,7 @@ export function displayTimestamp(
     hour: "numeric",
     minute: "numeric",
     second: "numeric",
-    hour12: false,
+    hourCycle: "h23",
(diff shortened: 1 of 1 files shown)


CriesofCarrots:
Just checking: Is the date right? I would expect 24:00 Dec 22 to be equivalent to 00:00 Dec 23 (but perhaps the date is currently wrong for 24:00)

jstarry:
> Just checking: Is the date right? I would expect 24:00 Dec 22 to be equivalent to 00:00 Dec 23 (but perhaps the date is currently wrong for 24:00)

@CriesofCarrots date is currently wrong :/

PR #10095, opened 2020-05-18
Fix erasure

carllin:
#### Problem
Erasure FEC coding blocks are overlapping
#### Summary of Changes

Fixes #

PR #15848, opened 2021-03-14
renames is_last_in_fec_set back to is_last_data

behzadnouri:
#### Problem
https://github.com/solana-labs/solana/pull/10095 renamed `is_last_data` to `is_last_in_fec_set`.
However, the code shows that this is actually meant to indicate where the serialized data is complete:
https://github.com/solana-labs/solana/blob/420174d3d/ledger/src/shred.rs#L599-L600
https://github.com/solana-labs/solana/blob/420174d3d/ledger/src/shred.rs#L229-L231

There are multiple FEC sets for each `&[Entry]` serialized and this flag does not represent shreds last in FEC sets (only the very last one by overlap). So the name is wrong and confusing

#### Summary of Changes
Rolled back the name change.

### DIFF
--- a/ledger/src/shred.rs
+++ b/ledger/src/shred.rs
@@ -201,7 +201,7 @@ impl Shred {
         index: u32,
         parent_offset: u16,
         data: Option<&[u8]>,
-        is_last_in_fec_set: bool,
+        is_last_data: bool,
         is_last_in_slot: bool,
         reference_tick: u8,
         version: u16,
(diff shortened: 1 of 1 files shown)


carllin (approved):
Ah good catch!, yeah there can be multiple FEC sets that aren't data complete (individually deserializable). Maybe it would be better to make each FEC set cut off at a transaction boundary so they are individually deserializable/replayable.

PR #15849, opened 2021-03-14
buffers data shreds to make larger erasure coded sets

behzadnouri:
#### Problem
Broadcast stage batches up to 8 entries:
https://github.com/solana-labs/solana/blob/79280b304/core/src/broadcast_stage/broadcast_utils.rs#L26-L29
which will be serialized into some number of shreds and chunked into FEC
sets of at most 32 shreds each:
https://github.com/solana-labs/solana/blob/79280b304/ledger/src/shred.rs#L576-L597
So depending on the size of entries, FEC sets can be small, which may
aggravate loss rate.
For example 16 FEC sets of 2:2 data/code shreds each have higher loss
rate than one 32:32 set.

#### Summary of Changes
This commit broadcasts data shreds immediately, but also buffers them
until it has a batch of 32 data shreds, at which point 32 coding shreds
are generated and broadcasted.

### DIFF
--- a/core/src/broadcast_stage/standard_broadcast_run.rs
+++ b/core/src/broadcast_stage/standard_broadcast_run.rs
@@ -7,12 +7,13 @@ use super::{
 use crate::broadcast_stage::broadcast_utils::UnfinishedSlotInfo;
 use solana_ledger::{
     entry::Entry,
-    shred::{ProcessShredsStats, Shred, Shredder, RECOMMENDED_FEC_RATE, SHRED_TICK_REFERENCE_MASK},
+    shred::{
+        ProcessShredsStats, Shred, Shredder, MAX_DATA_SHREDS_PER_FEC_BLOCK, RECOMMENDED_FEC_RATE,
+        SHRED_TICK_REFERENCE_MASK,
+    },
 };
 use solana_sdk::{pubkey::Pubkey, signature::Keypair, timing::duration_as_us};
-use std::collections::HashMap;
-use std::sync::RwLock;
-use std::time::Duration;
+use std::{collections::HashMap, ops::Deref, sync::RwLock, time::Duration};
 
 #[derive(Clone)]
 pub struct StandardBroadcastRun {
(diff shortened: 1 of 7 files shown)


aeyakovenko:
does this accumulate data shreds until we see 32?

behzadnouri:
> does this accumulate data shreds until we see 32?

@aeyakovenko yes, `data_shreds_buffer` is buffering data-shreds, and `make_coding_shreds` is consuming from the buffer a multiple of `MAX_DATA_SHREDS_PER_FEC_BLOCK` which is 32.

aeyakovenko:
once this works, we should be able to add a new confirmation status
https://docs.solana.com/developing/clients/jsonrpc-api#configuring-state-commitment

* executed - Each of the turbine batches have over 51% of the stake successfully retransmitting the shreds + this block is attached to an optimistic or executed block + this block was replayed successfully without an error (tag @carllin)

core/src/broadcast_stage/standard_broadcast_run.rs
```diff
@@ -40,94 +41,114 @@ impl StandardBroadcastRun {
...
+            process_stats,
+        );
+        let mut data_shreds_buffer = match &mut self.unfinished_slot {
+            Some(state) => {
+                assert_eq!(state.slot, slot);
+                std::mem::take(&mut state.data_shreds_buffer)
+            }
+            None => Vec::default(),
+        };
+        data_shreds_buffer.extend(data_shreds.clone());
```
carllin, line 144: nice! could we add a simple test `entries_to_data_shreds() -> make_coding_shreds()` with a few iterative batches and then check at the end that one big batch was made?

core/src/broadcast_stage/standard_broadcast_run.rs
```diff
@@ -40,94 +41,114 @@ impl StandardBroadcastRun {
...
-            shredder.entries_to_data_shreds(entries, is_slot_end, next_shred_index, process_stats);
-
+        let (slot, parent_slot) = self.current_slot_and_parent.unwrap();
+        let (next_shred_index, fec_set_offset) = match &self.unfinished_slot {
+            Some(state) => (state.next_shred_index, state.fec_set_offset),
+            None => match blockstore.meta(slot).unwrap() {
+                Some(slot_meta) => {
+                    let shreds_consumed = slot_meta.consumed as u32;
+                    (shreds_consumed, shreds_consumed)
+                }
```
carllin, line 117: I think we may actually be able to remove this outdated case, since if the blockstore has metadata for that slot, it should not recreate the slot due to this check `https://github.com/solana-labs/solana/blob/master/ledger/src/leader_schedule_cache.rs#L141-L144`

core/src/broadcast_stage/standard_broadcast_run.rs
```diff
@@ -418,6 +469,8 @@ impl BroadcastRun for StandardBroadcastRun {
         blockstore_sender: &Sender<(Arc<Vec<Shred>>, Option<BroadcastShredBatchInfo>)>,
     ) -> Result<()> {
         let receive_results = broadcast_utils::recv_slot_entries(receiver)?;
+        // TODO: Confirm that last chunk of coding shreds
```
carllin, line 472: A good way to confirm this might be by monitoring the `broadcast-transmit-shreds-stats` metric, `end_to_end_elapsed` field. It doesn't look like it should be too big of a deal because even with these changes, it looks like at most we will have at most an additional `MAX_DATA_SHREDS_PER_FEC_BLOCK` at the last iteration of `process_receive_results` for this slot compared to the code today

core/src/broadcast_stage/standard_broadcast_run.rs
```diff
@@ -255,11 +271,15 @@ impl StandardBroadcastRun {
         // Send data shreds
         let data_shreds = Arc::new(data_shreds);
         socket_sender.send(((stakes.clone(), data_shreds.clone()), batch_info.clone()))?;
-        blockstore_sender.send((data_shreds.clone(), batch_info.clone()))?;
+        blockstore_sender.send((data_shreds, batch_info.clone()))?;
 
         // Create and send coding shreds
-        let coding_shreds = shredder
-            .data_shreds_to_coding_shreds(&data_shreds[0..last_data_shred], &mut process_stats);
+        let coding_shreds = make_coding_shreds(
```
carllin, line 277: Nice, this is a clever way to avoid changing the `num_batches` and `num_expected_batches` logic above because it always returns a vector, even if that vector is empty.

carllin:
@behzadnouri whoo, finally knocking this one out! Really clean work 🧠 , just a few nits 😃

core/src/broadcast_stage/standard_broadcast_run.rs
behzadnouri: You are probably, but I think we better have a separate commit for that so that this one does not change the logic here, just in case there are unknown consequences.

core/src/broadcast_stage/standard_broadcast_run.rs
behzadnouri: Added more test coverage checking the batching behavior.

carllin (approved):

PR #15851, opened 2021-03-15
refactor test truncate data

jeffwashington:
#### Problem
In refactoring accounts, there are a few tests which truncate the data vector. This is complicated if data is abstracted.
#### Summary of Changes
Refactor the test code.
Fixes #

### DIFF
--- a/programs/bpf_loader/src/lib.rs
+++ b/programs/bpf_loader/src/lib.rs
@@ -1521,9 +1521,8 @@ mod tests {
     }
 
     fn truncate_data(account: &mut AccountSharedData, len: usize) {
-        let mut data = account.data.to_vec();
-        data.truncate(len);
-        account.set_data(data);
+        // when account data becomes copy on write, this operation will be more complicated
+        account.data.truncate(len);
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


programs/bpf_loader/src/lib.rs
```diff
@@ -1520,6 +1520,11 @@ mod tests {
         );
     }
 
+    fn truncate_data(account: &mut AccountSharedData, len: usize) {
```
carllin, line 1523: won't `truncate` be a method on either `AccountSharedData` as a required trait method on `WritableAccount` that `AccountSharedData` implements?

jeffwashington:
it is possible this could be a method on AccountSharedData, AccountData, or WritableAccount as you have suggested. It is only used by tests right now, so it doesn't seem 'necessary' to add it to account. It is troublesome in a few ways. I planned to engage with you once these other trivial changes were all in so that we could choose the right design for AccountSharedData & AccountData, where the Arc goes, how to integrate with Behzad's changes, how to put an arc over the entire account, what belongs on a trait, etc. My plan was to punt on truncate and push since they were only used by testing code, so we don't have to work those into the api yet in a 1st class way. I'm open to other suggestions. i have things to talk to you about concerning the suggestion of putting it on WriteableAccount. We can delay this pr if you want. I recommend we put it in and visit it later if it becomes important, but I'm open to waiting to put this in with the bigger change. This seems of low importance to me.

carllin (approved):

PR #15854, opened 2021-03-15
set_data_from_slice

jeffwashington:
#### Problem
This is an expensive piece of code working with creating copies of account data.
When account.data becomes copy on write, then this algorithm will be reworked for higher efficiency.

#### Summary of Changes
Isolate the algorithm to the account so it can be optimized.

Fixes #

### DIFF
--- a/sdk/src/account.rs
+++ b/sdk/src/account.rs
@@ -395,7 +395,7 @@ impl Account {
 
 impl AccountSharedData {
     /// make account's data equal to 'data'. This may require resizing and copying data.
-    pub fn data_ensure_equal_to_slice(&mut self, data: &[u8]) {
+    pub fn set_data_from_slice(&mut self, data: &[u8]) {
         let len = self.data.len();
         let len_different = len != data.len();
         if len_different {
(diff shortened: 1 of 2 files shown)


sdk/src/account.rs
```diff
@@ -394,6 +394,20 @@ impl Account {
...
+        let len_different = len != data.len();
+        if len_different {
+            // if the resize causes a reallocation and copy, it would be better to create a new copy of the final data
+            //  rather than resize (+ copy current) and then copy over below.
+            // however, the implementation of account's data is soon to be copy on write, so the tradeoffs will soon be different.
+            self.data.resize(data.len(), 0);
+        }
+        // we could compare here to determine whether we need to modify the original data or not. In the current implementation, that would
+        //  not make a positive difference.
+        self.data.clone_from_slice(data);
```
carllin, line 409: `copy_from_slice` may be more efficient here

sdk/src/account.rs
```diff
@@ -394,6 +394,20 @@ impl Account {
 }
 
 impl AccountSharedData {
+    /// make account's data equal to 'data'. This may require resizing and copying data.
+    pub fn data_ensure_equal_to_slice(&mut self, data: &[u8]) {
```
carllin, line 398: how about renaming this to `set_data_from_slice()`

sdk/src/account.rs
jeffwashington: That's funny. I had a separate change where I changed this to copy_from_slice as the docs indicate it could be more efficient. Benchmarking showed no difference. I left it like this so it matched exactly the code it was replacing, but my inclination was to change it to copy. You have pushed me over the edge.

carllin (approved):

PR #15855, opened 2021-03-15
explorer: minor cleanup and fixes

jstarry:
#### Problem
- Vote details row is not keyed
- Confirmation status can be undefined
- Rewards may be undefined

#### Summary of Changes

Fixes #


bartosz-lipinski:
@jstarry, Looks, good but can you fix type errors null vs undefined.

PR #15856, opened 2021-03-15
web3: clean up nullable typing and validate confirmation status string

jstarry:
#### Summary of Changes
- `confirmationStatus` may not be set (`undefined`) but will never be `null`. 
- validate confirmation status strings

Fixes #

### DIFF
--- a/web3.js/module.d.ts
+++ b/web3.js/module.d.ts
@@ -106,7 +106,7 @@ declare module '@solana/web3.js' {
     slot: number;
     err: TransactionError | null;
     confirmations: number | null;
-    confirmationStatus: TransactionConfirmationStatus | null;
+    confirmationStatus?: TransactionConfirmationStatus;
   };
 
   export type ConfirmedSignatureInfo = {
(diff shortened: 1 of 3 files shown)


jstarry:
The corresponding types in `core/src/rpc.rs` are all `Option` wrapped but if I understand correctly, it's to handle old rpc servers which don't respond with new attributes. But any new rpc server will never return null for one of these values. @CriesofCarrots can you confirm if this is true?

CriesofCarrots:
> The corresponding types in `core/src/rpc.rs` are all `Option` wrapped but if I understand correctly, it's to handle old rpc servers which don't respond with new attributes. But any new rpc server will never return null for one of these values. @CriesofCarrots can you confirm if this is true?

Correct, `confirmation_status` and `feature_set` should never be None from a new rpc server.
`ui_amount_string` is a String, not an Option; as you've discovered, it can be erroneously empty atm, but should never be after https://github.com/solana-labs/solana/pull/15873 lands.

jstarry:
Merging as part of https://github.com/solana-labs/solana/pull/15858

PR #15857, opened 2021-03-15
web3: require programId and keys for TransactionInstruction

jstarry:
#### Problem
The `TransactionInstruction` class claims to always have a defined `programId` but can be constructed without one. This is unnecessary because there's no reason a `TransactionInstruction` should be constructed without a program id or key list.

#### Summary of Changes
- Require `TransactionInstruction` to have defined programId

Fixes #


jstarry:
@mvines  This is a breaking change that shouldn't break anyone, but if it does, it's probably fixing a bug

mvines (approved):

PR #15858, opened 2021-03-15
web3: migrate to typescript

jstarry:
#### Problem
Maintainability of web3.js is a pain due to manual management of typing definitions

#### Summary of Changes
- Migrate web3 source to TypeScript
- Generate TS and Flow type definition files during build
- Migrate from `esdoc` to `typedoc` for auto generated type comments
- Fix nullable typings that can never be null
- Fix confirmation status types
- Fix buffer types
- Require `programId` to be set for `TransactionInstruction`

Fixes #


jstarry:
@oJshua and @bartosz-lipinski rejoice! web3 is nearly migrated to TS! Would really appreciate another set of eyes on this, the changes are massive but split into different commits for reviewability.

web3.js/src/fee-calculator.ts
```diff
@@ -1,17 +1,17 @@
-// @flow
+// @ts-ignore
```
oJshua, line 1: Assuming no type definitions or BufferLayout?

web3.js/src/system-program.ts
```diff
@@ -670,7 +690,7 @@ export class SystemProgram {
   ): TransactionInstruction {
     let data;
     let keys;
-    if (params.basePubkey) {
+    if ('basePubkey' in params) {
```
oJshua, line 693: Minor nit: maybe prettier should convert single quotes to doubles like in our React projects (for extra consistency).

oJshua (dismissed):
This looks good to me. Thanks for taking this on!

web3.js/src/fee-calculator.ts
jstarry: Yeah, unfortunately not. There are some other packages with generated typings that we could switch to later though!

Issue #13168, opened 2020-10-26
explorer: SPL mint account should show token history


oJshua:
#### Problem
It would be nice if SPL mint accounts showed the token history, similar to how standard accounts do.

https://explorer.solana.com/address/ErF6ptYb3tKzSZjLKBoTqYJPUkMQtPUeAGo16kGLdGVt/tokens (account showing token history)

https://explorer.solana.com/address/SRMuApVNdxXokk5GT7XD5cUUgXMBCoAz2LHeuAoKWRt (SPL mint account without history)

#### Proposed Solution
Make token history available (may require additional indices in rocksdb/bigtable).

PR #15861, opened 2021-03-15
explorer: Token mint histories (token balance and token instructions tabs)

oJshua:
#### Problem
SPL mints currently don't display any token history information (such as instructions, token flow).

#### Summary of Changes
- Begin refactoring out functionality from existing TokenHistoryCard into common utilities and components
- Introduce two new tabs for token mints (Transfers, Instructions)
- This work will serve as basis for adding similar functionality to Token Accounts

Fixes https://github.com/solana-labs/solana/issues/13168


oJshua:
Explorer-Solana (16)

Explorer-Solana (17)

oJshua:
@jstarry This POC is ready for a quick review. I think ultimately we want to go the direction of etherscan, showing direct token flows, but for now balance deltas may do since we already have them.

jstarry:
@oJshua I personally don't find the "Token Balances" tab to be too informative. I think we could have a nice "Transfers" tab which displays all the "TransferChecked" instructions which are in transaction-level instructions and inner instructions, thoughts?

oJshua:
@jstarry I think I agree with you there! It would be nice if at some point we could specifically query for transactions that include TransferChecked. Or more generally, transactions that include X kind of instruction (possibly program ID too), but that of course would require some work.

jstarry:
> @jstarry I think I agree with you there! It would be nice if at some point we could specifically query for transactions that include TransferChecked. Or more generally, transactions that include X kind of instruction, but that of course would require some work.

@oJshua great, that would be cool! For now, shall we stick with the client side filtering approach in this PR?

oJshua:
@jstarry Yep!

oJshua:
@jstarry Here's some progress on the Transfers tab. I'm including non-checked Transfers as well.

001_mint_transfers

I also experimented with removing the 'Status' column and showing failed transactions with a subtle background + left border to save some space:

Explorer-Solana (32)

Thoughts welcome!

jstarry:
The transfers info is great! A few design suggestions:
- the ellipsis in the middle of the address and hash look a little odd, I think it would be better to put them at the end
- I think we can drop the slot number column
- I really like etherscan's relative time column (1hr ago) is much easier to read at a glance than a full UTC timestamp
- I like the order of etherscan's columns too, it feels weird to have the hash and timestamp at the end

I like the color coded errors!
- I think that adding a small icon would be nice too so that it's color-blind friendly
- Maybe the hash link can be pink as well?

oJshua:
Explorer-Solana (35)

Explorer-Solana (36)

Explorer-Solana (37)

jstarry:
@oJshua the error and copy icons don't look too great right next to each other, agreed that it's too tight. Also not a big fan of the linear gradient :/

oJshua:
@jstarry I'll try something else!

oJshua:
@jstarry How about something like this? Alert icon on the slot; still functions as a copy button on hover (all with CSS). Obviously mobile users won't realize that's a copy button.

Explorer-Solana (39)

Explorer-Solana (40)

I also removed the left $warning border to keep it simple.

oJshua:
history_tab

transfers_tab

instructions_tab

- reintroduce status column
- add relative times
- reorganize columns similar to Etherscan

explorer/src/components/account/history/TokenTransfersDetails.tsx
```diff
@@ -0,0 +1,213 @@
...
+  const { tokenRegistry } = useTokenRegistry();
+  const mintDetails = tokenRegistry.get(pubkey.toBase58());
+
+  React.useEffect(() => {
+    if (history?.data?.fetched) {
+      fetchDetailedAccountHistory(history.data.fetched);
+    }
+  }, [history]); // eslint-disable-line react-hooks/exhaustive-deps
+
+  const hasTimestamps = !!slotRows.find((element) => !!element.blockTime);
```
jstarry, line 51: nit: you can use `some(..)` here

explorer/src/components/account/history/TokenTransfersDetails.tsx
```diff
@@ -0,0 +1,213 @@
...
+          if (transfer) {
+            transfers.push(transfer);
+          }
+        });
+      }
+    );
+
+    // Filter out transfers not belonging to this mint
+    transfers = transfers.filter((transfer) => {
+      if ("tokenAmount" in transfer && transfer.mint !== pubkey) {
```
jstarry, line 102:
```suggestion
      if ("tokenAmount" in transfer && !transfer.mint.equals(pubkey)) {
```
This check is excluding all checked transfers

explorer/src/components/account/history/TokenTransfersDetails.tsx
```diff
@@ -0,0 +1,213 @@
...
+        }
+
+        amountString = new Intl.NumberFormat("en-US", {
+          minimumFractionDigits: decimals,
+          maximumFractionDigits: decimals,
+        }).format(normalizeTokenAmount(transfer.amount, decimals));
+      }
+
+      detailsList.push(
+        <tr key={signature + transfer.source + transfer.destination}>
```
jstarry, line 147: One transaction could have multiple transfers between the same accounts. Maybe each transfer could have a unique id from signature + instruction / inner indexes?

explorer/src/components/account/history/TokenTransfersDetails.tsx
```diff
@@ -0,0 +1,213 @@
...
+        <tbody className="list">{detailsList}</tbody>
+      </table>
+    </div>
+  );
+}
+
+function getTransfer(
+  instruction: ParsedInstruction | PartiallyDecodedInstruction
+): Transfer | TransferChecked | undefined {
+  if ("parsed" in instruction && instruction.program === "spl-token") {
```
jstarry, line 202: Might want to wrap this in a try catch and report any errors

explorer/src/components/account/history/TokenInstructionsDetails.tsx
```diff
@@ -0,0 +1,105 @@
...
+          index
+        );
+
+        const programId = ix.programId;
+
+        if (instructionType) {
+          detailsList.push(
+            <tr key={signature + index}>
+              <td>
+                <Signature signature={signature} link truncateChars={48} />
```
jstarry, line 57: I think we can truncate this to 24 across the board to help with mobile responsiveness, thoughts?

explorer/src/components/common/Address.tsx
```diff
@@ -35,6 +37,14 @@ export function Address({
     truncate = true;
   }
 
+  let addressLabel = raw
+    ? address
+    : displayAddress(address, cluster, tokenRegistry);
+
+  if (truncateChars && addressLabel === address) {
+    addressLabel = addressLabel.slice(0, truncateChars) + "...";
```
jstarry, line 45:
```suggestion
    addressLabel = addressLabel.slice(0, truncateChars) + "…";
```
nit: the ellipsis character takes up less space

explorer/src/components/common/InstructionDetails.tsx
```diff
@@ -0,0 +1,48 @@
...
+  tx,
+}: {
+  instructionType: InstructionType;
+  tx: ConfirmedSignatureInfo;
+}) {
+  const [expanded, setExpanded] = React.useState(false);
+
+  let instructionTypes = instructionType.innerInstructions
+    .map((ix) => {
+      if ("parsed" in ix && ix.program === "spl-token") {
```
jstarry, line 16: I think these should be filtered by mint right?

explorer/src/providers/accounts/detailed-history.tsx
```diff
@@ -0,0 +1,156 @@
...
+): Map<string, ParsedConfirmedTransaction> {
+  const context = React.useContext(StateContext);
+
+  if (!context) {
+    throw new Error(`useAccountHistory must be used within a AccountsProvider`);
+  }
+
+  return (
+    context.entries[address]?.data?.transactionMap ||
+    new Map<string, ParsedConfirmedTransaction>()
```
jstarry, line 77: Can you memoize this to prevent downstream re-renders?

explorer/src/components/instruction/serum/types.ts
```diff
@@ -301,7 +301,8 @@ export function isSerumInstruction(instruction: TransactionInstruction) {
     SERUM_PROGRAM_IDS.includes(instruction.programId.toBase58()) ||
     MARKETS.some(
       (market) =>
-        market.programId && market.programId.equals(instruction.programId)
+        market.programId &&
+        market.programId.toBase58() === instruction.programId.toBase58()
```
jstarry, line 305: Any reason not to use the pubkey equals method? Using `toBase58` is pretty expensive

explorer/src/components/account/history/TokenTransfersDetails.tsx
```diff
@@ -0,0 +1,213 @@
...
+      if ("tokenAmount" in transfer && transfer.mint !== pubkey) {
+        return false;
+      } else if (
+        mintMap.has(transfer.source.toBase58()) &&
+        mintMap.get(transfer.source.toBase58())?.mint !== pubkey.toBase58()
+      ) {
+        return false;
+      } else if (
+        mintMap.has(transfer.destination.toBase58()) &&
+        mintMap.get(transfer.destination.toBase58())?.mint !== pubkey.toBase58()
```
jstarry, line 111: Calling `toBase58` is pretty expensive, it would be nice to precompute `pubkey.toBase58()`. I'm seeing lag when creating the "transfers" and "instructions" view

explorer/src/components/account/history/TokenTransfersDetails.tsx
```diff
@@ -0,0 +1,213 @@
...
+          const transfer = getTransfer(instruction);
+          if (transfer) {
+            transfers.push(transfer);
+          }
+        });
+      }
+    );
+
+    // Filter out transfers not belonging to this mint
+    transfers = transfers.filter((transfer) => {
```
jstarry, line 101: If you flip the logic to return true when conditions are met, I think the performance would be quite a bit better

jstarry (commented):
Yeah, I think that we can revisit design in the future. The transfers and instructions tables look a lot nicer now in my opinion! I noticed that rendering was a little laggy on my macbook pro so I think it deserves another look to speed that up

explorer/src/components/account/history/TokenTransfersDetails.tsx
oJshua: That makes sense! I'll add that.

explorer/src/components/instruction/serum/types.ts
oJshua: Not sure why this was introduced. Thanks!

explorer/src/components/account/history/TokenInstructionsDetails.tsx
oJshua: Good call

jstarry:
@oJshua are you still working on this? I haven't looked since because of failing CI

oJshua:
@jstarry ahh, I'm sorry about that! I was being a little shy about clicking the Re-Request, as I was still spinning on one thing. I've since pushed / rebased -- I'll check that it passes CI.

oJshua:
Summary: 
- I moved parsed transaction map over to history context / cache reducer. This seemed cleaner than having a separate context.
- One goal I had was preventing the default History tab from using the batch parsed transaction request, as I didn't want the default behavior stressing the RPC nodes. The additional transactions are only fetched if the user visits the Transfers or Instructions tab.
- Instructions now display only outer and inner instructions associated with a mint -- one per row.
- I left the default signatures (on the History tab) the full length to space out the table better. The truncate media query seems to handle truncation well for these rather than limiting to X characters.

explorer/src/providers/accounts/history.tsx
```diff
@@ -4,19 +4,26 @@ import {
...
   TransactionSignature,
   Connection,
+  ParsedConfirmedTransaction,
 } from "@solana/web3.js";
 import { useCluster, Cluster } from "../cluster";
 import * as Cache from "providers/cache";
 import { ActionType, FetchStatus } from "providers/cache";
 import { reportError } from "utils/sentry";
 
+const MAX_TRANSACTION_BATCH_SIZE = 30;
```
jstarry, line 14: I think this is too high, the queries are really slow. Maybe 10 instead?

explorer/src/providers/accounts/history.tsx
```diff
@@ -4,19 +4,26 @@ import {
...
+
+type TransactionMap = Map<string, ParsedConfirmedTransaction>;
+
 type AccountHistory = {
   fetched: ConfirmedSignatureInfo[];
+  transactionMap?: TransactionMap;
```
jstarry, line 20: Why not use the transactions cache to store these?

explorer/src/components/account/HistoryCardComponents.tsx
```diff
@@ -0,0 +1,116 @@
...
+        statusText = "Success";
+      }
+      transactionRows.push({
+        slot,
+        signature: slotTransaction.signature,
+        err: slotTransaction.err,
+        blockTime: slotTransaction.blockTime,
+        statusClass,
+        statusText,
+        signatureInfo: transactions[i],
```
jstarry, line 110: This is passing the incorrect `signatureInfo` for this row. It looks like this field isn't needed downstream anyways, I think you can delete it

jstarry (dismissed):
Looks really solid, great work!

explorer/src/components/account/HistoryCardComponents.tsx
oJshua: Good catch, thank you.

explorer/src/providers/accounts/history.tsx
oJshua: Let me think about that. That might be a better approach.

PR #15863, opened 2021-03-15
web3: remove deprecated bpf-sdk, localnet, and examples

jstarry:
#### Problem
Doesn't make sense for the BPF SDK and test validator scripts to live inside web3 anymore now that they are provided with the solana cli tools

#### Summary of Changes
- Removed localnet and bpf-sdk scripts
- Fix release notes issue links (should go to the solana monorepo instead of web3)
- Removed outdated examples (we need actually useful examples)
- Rather than building test fixtures, checked them into source
- Removed unnecessary files (doc, examples, etc) from release artifacts

Fixes #


jackcmay:
Looks good overall, might want to leave the test fixture source and build scripts so we can recreate the test .so files for the deploy tests.

jstarry:
> might want to leave the test fixture source and build scripts so we can recreate the test .so files

@jackcmay good call, any objection to just keeping the Rust fixtures? Maintaining the C fixtures is a bit more bothersome

mvines:
This is great but I'm worried about how many people we're going to break by yanking the BPF toolchain setup.  This definitely deserves a major version bump for web3.js (time to wave 0.x good bye!).

I think web3.js is the only consumer for [ci/publish-bpf-sdk.sh](https://github.com/solana-labs/solana/blob/85b4578b140967c4af1ed4f80f7c8e7f5f38fe44/ci/buildkite-secondary.yml#L8) right?  We can rip out all this plumbing as well here as well.

jackcmay:
> > might want to leave the test fixture source and build scripts so we can recreate the test .so files
> 
> @jackcmay good call, any objection to just keeping the Rust fixtures? Maintaining the C fixtures is a bit more bothersome

Fine by me

jstarry:
> This definitely deserves a major version bump for web3.js (time to wave 0.x good bye!).

Sounds good, afaict prepending `BREAKING CHANGE` should do the trick here

> We can rip out all this plumbing as well here as well

Cool, will do today

jstarry:
Haven't figured out how to test that semantic release will work correctly yet, will look again tomorrow

jstarry:
> Haven't figured out how to test that semantic release will work correctly yet, will look again tomorrow

Alright, tested it. As long as you have a `BREAKING CHANGE: blah blah` in the commit message body, the major version will bump

PR #15870, opened 2021-03-15
[wip] maintain spl-token exchange integration docs

ryoqun:
#### Problem

This is still a wip...

#### Summary of Changes

Fixes #


docs/src/integrations/exchange.md
```diff
@@ -695,7 +708,7 @@ lamports).
 
 Template `spl-token transfer` command for a withdrawal:
```
ryoqun, line 709: Also add some best practices like

- check owner address having more than 0 lamports
- check given withdrawal address isn't token holding account, and actually system account.

PR #15885, opened 2021-03-15
sdk: Improve Instruction::new deprecation warning

mvines:


sdk/program/src/instruction.rs
```diff
@@ -216,7 +216,7 @@ pub struct Instruction {
 impl Instruction {
     #[deprecated(
         since = "1.6.0",
-        note = "Please use another `Instruction constructor instead"
+        note = "Please use another Instruction constructor instead, such as `Instruction::new_with_bincode`"
```
jackcmay, line 219: Do you want to direct folks to use `bincode`?

mvines: My thinking here was that most likely all `Instruction::new()` usage in the wild today is now `Instruction::new_with_bincode()`, so recommending the most likely course of action to resolve the warning seems like the most friendly thing to do

jackcmay: Maybe we want to go a step further, rather not encourage folks continue or start down the bincode path

jackcmay: Maybe instead of the deprecated message we put more info into documenting that function

mvines: From the SPL perspective, an upgrade to 1.6.0 throws a bunch of these warnings.  There's no path other than changing to `Instruction::new_with_bincode()` since we're not going to rewrite the existing programs to use Borsh.  I imagine this'll be the case with virtually everybody else who goes from 1.5.x to 1.6.x.

I think dealing with not going down the bincode path to begin with is probably better addressed by ensuring all our examples and tutorials don't use it.

jackcmay: Agreed, both would help

PR #15887, opened 2021-03-15
Close buffer accounts

jackcmay:
#### Problem

Once a buffer account is created there is no way to recover the lamports from it except to deploy it.

#### Summary of Changes

- Add a close instruction to the upgradeable loader that will close buffer accounts.
- Add cli tooling to close accounts
- Add cli tooling to show a list of buffer accounts

Fixes #

### DIFF
--- a/cli-output/src/cli_output.rs
+++ b/cli-output/src/cli_output.rs
@@ -1392,7 +1392,7 @@ impl fmt::Display for CliAccountBalances {
         writeln!(
             f,
             "{}",
-            style(format!("{:<44}  {}", "Address", "Balance",)).bold()
+            style(format!("{:<44}  {}", "Address", "Balance")).bold()
         )?;
         for account in &self.accounts {
             writeln!(
@@ -1684,6 +1684,8 @@ pub struct CliUpgradeableBuffer {
     pub address: String,
     pub authority: String,
     pub data_len: usize,
+    pub lamports: u64,
+    pub use_lamports_unit: bool,
 }
 impl QuietDisplay for CliUpgradeableBuffer {}
 impl VerboseDisplay for CliUpgradeableBuffer {}
@@ -1694,9 +1696,51 @@ impl fmt::Display for CliUpgradeableBuffer {
         writeln_name_value(f, "Authority:", &self.authority)?;
         writeln_name_value(
             f,
+            "Balance:",
+            &build_balance_message(self.lamports, self.use_lamports_unit, true),
+        )?;
+        writeln_name_value(
+            f,
             "Data Length:",
             &format!("{:?} ({:#x?}) bytes", self.data_len, self.data_len),
         )?;
+
+        Ok(())
+    }
+}
+
+#[derive(Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct CliUpgradeableBuffers {
+    pub buffers: Vec<CliUpgradeableBuffer>,
+    pub use_lamports_unit: bool,
+}
+impl QuietDisplay for CliUpgradeableBuffers {}
+impl VerboseDisplay for CliUpgradeableBuffers {}
+impl fmt::Display for CliUpgradeableBuffers {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        writeln!(f)?;
+        writeln!(
+            f,
+            "{}",
+            style(format!(
+                "{:<44} | {:<44} | {}",
+                "Buffer Address", "Authority", "Balance"
+            ))
+            .bold()
+        )?;
+        for buffer in self.buffers.iter() {
+            writeln!(
+                f,
+                "{}",
+                &format!(
+                    "{:<44} | {:<44} | {}",
+                    buffer.address,
+                    buffer.authority,
+                    build_balance_message(buffer.lamports, self.use_lamports_unit, true)
+                )
+            )?;
+        }
         Ok(())
     }
 }
--- a/cli/src/program.rs
+++ b/cli/src/program.rs
@@ -10,15 +10,22 @@ use bincode::serialize;
 use bip39::{Language, Mnemonic, MnemonicType, Seed};
 use clap::{App, AppSettings, Arg, ArgMatches, SubCommand};
 use log::*;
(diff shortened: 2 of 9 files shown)


cli/src/program.rs
```diff
@@ -487,19 +568,87 @@ pub fn parse_program_subcommand(
...
+            } else {
+                pubkey_of(matches, "account")
+            };
+
+            let authority_pubkey = if let Some(authority_pubkey) =
+                pubkey_of_signer(matches, "buffer_authority", wallet_manager)?
+            {
+                authority_pubkey
+            } else if let (_, Some(authority_pubkey)) =
+                signer_of(matches, "buffer_authority", wallet_manager)?
```
jstarry, line 583: I'm probably wrong but isn't this doing the same thing as `pubkey_of_signer` above?

cli/src/program.rs
```diff
@@ -1103,6 +1331,162 @@ fn process_dump(
...
+    let mut tx = Transaction::new_unsigned(Message::new(
+        &[bpf_loader_upgradeable::close(
+            &account_pubkey,
+            &recipient_pubkey,
+            &authority_signer.pubkey(),
+        )],
+        Some(&config.signers[0].pubkey()),
+    ));
+
+    tx.try_sign(&[config.signers[0], authority_signer], blockhash)?;
```
jstarry, line 1352: will this fail if `config.signers[0]` and `authority_signer` are the same?

cli/src/program.rs
```diff
@@ -1103,6 +1331,162 @@ fn process_dump(
...
+                    {
+                        buffers.push(CliUpgradeableBuffer {
+                            address: account_pubkey.to_string(),
+                            authority: authority_address
+                                .map(|pubkey| pubkey.to_string())
+                                .unwrap_or_else(|| "none".to_string()),
+                            data_len: 0,
+                            lamports: account.lamports,
+                            use_lamports_unit,
+                        });
```
jstarry, line 1420: I'm not sure I understand what this is for. Is the intent to show that the buffer was closed?

cli-output/src/cli_output.rs
```diff
@@ -1692,11 +1694,53 @@ impl fmt::Display for CliUpgradeableBuffer {
         writeln!(f)?;
         writeln_name_value(f, "Buffer Address:", &self.address)?;
         writeln_name_value(f, "Authority:", &self.authority)?;
+        writeln_name_value(
+            f,
+            "Lamports:",
```
jstarry, line 1699: Balance?

jstarry (commented):
Looks really solid, thanks for doing this!

cli/src/program.rs
jackcmay: Correct, to show the caller which accounts were closed

cli/src/program.rs
jackcmay: Yeah, I think I copypasta'd that pattern from somewhere else, will fix all those cases in a followup

docs/src/cli/deploy-a-program.md
```diff
@@ -124,28 +124,83 @@ call to `deploy`.
...
-To resume a failed deploy, recover the ephemeral keypair file with
-`solana-keygen recover` and the following 12-word seed phrase,
-then pass it as the [BUFFER_SIGNER] argument to `solana deploy` or `solana write-buffer`
-=======================================================================
-spy axis cream equip bonus daring muffin fish noise churn broken diesel
-=======================================================================
+```
+==================================================================================
+Recover the intermediate account's ephemeral keypair file with
+`solana-keygen recover` and the following 12-word seed phrase:
```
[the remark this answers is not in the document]
jackcmay: The first time they `program deploy` they need to provide a signer (since the account is being created), but for resume you just need the address.  We could simplify this so they don't have to recreate the keypair for a resume

cli/src/program.rs
jackcmay: Doesn't fail if same

cli-output/src/cli_output.rs
```diff
@@ -1684,6 +1684,8 @@ pub struct CliUpgradeableBuffer {
     pub address: String,
     pub authority: String,
     pub data_len: usize,
+    pub lamports: u64,
+    pub use_lamports_unit: bool,
```
CriesofCarrots, line 1688: I think we should tag this field with `#[serde(skip_serializing)]` to avoid including it in the json obj

cli-output/src/cli_output.rs
```diff
@@ -1692,11 +1694,53 @@ impl fmt::Display for CliUpgradeableBuffer {
...
+
+        Ok(())
+    }
+}
+
+#[derive(Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct CliUpgradeableBuffers {
+    pub buffers: Vec<CliUpgradeableBuffer>,
+    pub use_lamports_unit: bool,
```
CriesofCarrots, line 1716: Ditto

docs/src/cli/deploy-a-program.md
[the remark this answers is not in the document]
jstarry: I think that would be a really nice simplification!

programs/bpf_loader/src/lib.rs
```diff
@@ -634,12 +634,43 @@ fn process_loader_upgradeable_instruction(
...
+                    ic_logger_msg!(logger, "Incorrect buffer authority provided");
+                    return Err(InstructionError::IncorrectAuthority);
+                }
+                if authority.signer_key().is_none() {
+                    ic_logger_msg!(logger, "Buffer authority did not sign");
+                    return Err(InstructionError::MissingRequiredSignature);
+                }
+
+                recipient_account.try_account_ref_mut()?.lamports += close_account.lamports()?;
+                close_account.try_account_ref_mut()?.lamports = 0;
```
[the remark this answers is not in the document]
jstarry: Don't think so, seems safe, but happy you added the zero'ing out anyways 👍🏻

cli/src/program.rs
jstarry: Ok, I was confused why you deserialize the account state a second time without refetching the account in between.

jackcmay: Ha, blind spot

PR #15893, opened 2021-03-16
Cli: better estimate of epoch time elapsed/remaining

### LINKED ISSUE #14616: cli: Epoch-info remaining time differs from Explorer
#### Problem
In Explorer, the remaining time for an epoch is computed using the hourly average slot time. In the cli, this computation is done using a static ~400ms. This causes the cli remaining time to differ from Explorer.

```
$ solana epoch-info

Block height: 56159319
Slot: 60786828
Epoch: 140
Transaction Count: 9344364216
Epoch Slot Range: [60480000..60912000)
Epoch Completed Percent: 71.025%
Epoch Completed Slots: 306828/432000 (125172 remaining)
Epoch Completed Time: 1day 10h 5m 31s/2days (13h 54m 28s remaining)
```

#### Proposed Solution
Using the performance samples rpc, compute the hourly average slot time in order to calculate epoch remaining time. See: https://github.com/solana-labs/solana/blob/master/explorer/src/providers/stats/solanaDashboardInfo.tsx#L87-L90

CriesofCarrots:
#### Problem
Solana-cli's remaining time for an epoch is pretty far off. This is because the computation is done using a static ~400ms slot time; real slots have been much slower, and can be estimated using the `getRecentPerformanceSamples` rpc.

#### Summary of Changes
- Add rpc-client api for `getRecentPerformanceSamples`
- Use ^ in `solana epoch-info` to estimate epoch elapsed/remaining times based on the number of slots in the last hour's worth of samples. Rounded to secs.
- Also simplifies `solana block-height` which is building a CliEpochInfo for no reason.

Fixes #14616


cli-output/src/cli_output.rs
```diff
@@ -231,12 +231,8 @@ pub struct CliSlotStatus {
...
     #[serde(flatten)]
     pub epoch_info: EpochInfo,
-}
-
-impl From<EpochInfo> for CliEpochInfo {
-    fn from(epoch_info: EpochInfo) -> Self {
-        Self { epoch_info }
-    }
+    #[serde(skip)]
+    pub avg_slot_time_ms: u64,
```
mvines, line 235: My eye would prefer to see `avg` spelled out.  
```suggestion
    pub average_slot_time_ms: u64,
```

mvines (approved):

PR #15900, opened 2021-03-16
fix: compute pre/post token balances on all accounts if token program present

oJshua:
#### Problem
SPL Token balance changes from inner instructions are not included in {pre,post}TokenBalances.

#### Summary of Changes
Previously, account balance queries were performed only if token program was involved at the instruction level. This change queries all accounts when the token program is present in the transaction.

Fixes https://github.com/solana-labs/solana/issues/15760

### DIFF
--- a/transaction-status/src/token_balances.rs
+++ b/transaction-status/src/token_balances.rs
@@ -59,25 +59,16 @@ pub fn collect_token_balances(
 
     for (_, transaction) in OrderedIterator::new(batch.transactions(), batch.iteration_order()) {
         let account_keys = &transaction.message.account_keys;
-        let mut fetch_account_hash: HashMap<u8, bool> = HashMap::new();
-        for instruction in transaction.message.instructions.iter() {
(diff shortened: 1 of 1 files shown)


transaction-status/src/token_balances.rs
```diff
@@ -59,25 +59,20 @@ pub fn collect_token_balances(
...
-            }
-        }
+        let has_token_program = account_keys.iter().any(|p| is_token_program(p));
 
         let mut transaction_balances: Vec<TransactionTokenBalance> = vec![];
-        for index in fetch_account_hash.keys() {
-            if let Some(account_id) = account_keys.get(*index as usize) {
+        if has_token_program {
+            for (index, account_id) in account_keys.iter().enumerate() {
+                if is_token_program(account_id) {
```
CriesofCarrots, line 67: Ah, nice, but we can go even further: we can also skip checking addresses in Message::program_ids().

oJshua: Ah, good plan!

CriesofCarrots (approved):
lgtm, thx!

PR #15920, opened 2021-03-16
Remove need for mutex for PohRecorder.record calls

jeffwashington:
#### Summary of Changes

Fixes #


core/src/banking_stage.rs
```diff
@@ -640,10 +654,7 @@ impl BankingStage {
 
             let mut poh_record = Measure::start("record::poh_record");
             // record and unlock will unlock all the successful transactions
-            let res = poh
-                .lock()
-                .unwrap()
-                .record(bank_slot, hash, processed_transactions);
+            let res = recorder.record(bank_slot, hash, processed_transactions);
```
jeffwashington, line 657: @sakridge here's an example client call

core/src/poh_service.rs
```diff
@@ -107,51 +113,165 @@ impl PohService {
...
+        let mut try_again_mixin = None;
         loop {
-            num_hashes += hashes_per_batch;
-            if poh.lock().unwrap().hash(hashes_per_batch) {
+            let should_tick = {
+                let mixin = if let Some(record) = try_again_mixin {
+                    try_again_mixin = None;
+                    Ok(record)
+                } else {
+                    record_receiver.try_recv()
```
jeffwashington, line 145: @sakridge here's where the 'record' request is discovered.

PR #15929, opened 2021-03-16
eliminate lock on record

jeffwashington:
#### Problem
PohRecorder.Record requires a PohRecorder mutex lock. This results in contention with poh. Poh locking delays cause drift.

#### Summary of Changes
Create a new PohRecorder 'Recorder' which can use non-blocking channels to record transactions in poh.

Fixes #

### DIFF
--- a/core/src/banking_stage.rs
+++ b/core/src/banking_stage.rs
@@ -2451,6 +2451,8 @@ mod tests {
         Arc<Bank>,
         Arc<Mutex<PohRecorder>>,
         Receiver<WorkingBankEntry>,
+        JoinHandle<()>,
+        Arc<AtomicBool>,
     ) {
         Blockstore::destroy(&ledger_path).unwrap();
         let genesis_config_info = create_genesis_config(10_000);
@@ -2462,7 +2464,7 @@ mod tests {
         let blockstore =
             Blockstore::open(&ledger_path).expect("Expected to be able to open database ledger");
         let bank = Arc::new(Bank::new(&genesis_config));
-        let (poh_recorder, entry_receiver, _record_receiver) = PohRecorder::new(
+        let (poh_recorder, entry_receiver, record_receiver) = PohRecorder::new(
             bank.tick_height(),
             bank.last_blockhash(),
             bank.slot(),
@@ -2484,14 +2486,23 @@ mod tests {
             system_transaction::transfer(&mint_keypair, &pubkey1, 1, genesis_config.hash()),
             system_transaction::transfer(&mint_keypair, &pubkey2, 1, genesis_config.hash()),
         ];
-        (transactions, bank, poh_recorder, entry_receiver)
+        let (poh_simulator, stop) = simulate_poh(record_receiver, &poh_recorder);
+
+        (
+            transactions,
+            bank,
+            poh_recorder,
+            entry_receiver,
+            poh_simulator,
+            stop,
+        )
     }
 
     #[test]
     fn test_consume_buffered_packets() {
         let ledger_path = get_tmp_ledger_path!();
         {
-            let (transactions, bank, poh_recorder, _entry_receiver) =
+            let (transactions, bank, poh_recorder, _entry_receiver, poh_simulator, stop) =
                 setup_conflicting_transactions(&ledger_path);
             let recorder = poh_recorder.lock().unwrap().recorder();
             let num_conflicting_transactions = transactions.len();
@@ -2545,6 +2556,8 @@ mod tests {
                     assert_eq!(buffered_packets[0].1.len(), num_expected_unprocessed);
                 }
             }
+            stop.store(true, Ordering::Relaxed);
+            let _ = poh_simulator.join();
         }
         Blockstore::destroy(&ledger_path).unwrap();
     }
@@ -2553,7 +2566,7 @@ mod tests {
     fn test_consume_buffered_packets_interrupted() {
         let ledger_path = get_tmp_ledger_path!();
         {
-            let (transactions, bank, poh_recorder, _entry_receiver) =
+            let (transactions, bank, poh_recorder, _entry_receiver, poh_simulator, stop) =
                 setup_conflicting_transactions(&ledger_path);
             let num_conflicting_transactions = transactions.len();
             let packets_vec = to_packets_chunked(&transactions, 1);
@@ -2630,6 +2643,8 @@ mod tests {
             }
 
             t_consume.join().unwrap();
+            stop.store(true, Ordering::Relaxed);
+            let _ = poh_simulator.join();
         }
         Blockstore::destroy(&ledger_path).unwrap();
     }
--- a/core/src/banking_stage.rs
+++ b/core/src/banking_stage.rs
@@ -4,7 +4,7 @@
 use crate::{
     cluster_info::ClusterInfo,
     packet_hasher::PacketHasher,
-    poh_recorder::{PohRecorder, PohRecorderError, Recorder, WorkingBankEntry},
+    poh_recorder::{PohRecorder, PohRecorderError, TransactionRecorder, WorkingBankEntry},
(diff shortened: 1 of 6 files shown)


jeffwashington:
Here are some rough metrics:
With Stephen's test on the colo machine, throttled to ~3k tps requested, 107600 hashes/tick (should be 100% of machine capability), 5s sample time, I get:
master: 17% of the time spent not hashing (mutexes, sending record entry on channel, etc.). 672 ticks (800 is ideal)
my branch: 6% of the time not hashing. 751 ticks (800 is ideal)

core/src/poh_recorder.rs
```diff
@@ -51,6 +51,80 @@ type Result<T> = std::result::Result<T, PohRecorderError>;
...
+            transactions,
+            bank_slot,
+            self.result_sender.clone(),
+        ));
+        if res.is_err() {
+            return Err(PohRecorderError::MaxHeightReached); // TODO what error here
+        }
+        let res = self
+            .result_receiver
+            .recv_timeout(std::time::Duration::from_millis(2000)); //TODO: consider timeout? _timeout(Duration::from_millis(4000));
```
jeffwashington, line 118: @sakridge not sure what to do about a timeout here. This should only get hit on test conditions.

core/src/poh_recorder.rs
```diff
@@ -51,6 +51,80 @@ type Result<T> = std::result::Result<T, PohRecorderError>;
...
+        match res {
+            Err(_err) => {
+                Err(PohRecorderError::MaxHeightReached) // TODO - what error here
```
jeffwashington, line 121: @sakridge and I discussed and it looks like MaxHeightReached is the correct behavior in all callers. This should only occur in real error conditions or in shutdown of tests. We could add a new error code for PohRecorderError indicating a different kind of failure but it looks like we want the same behavior as MaxHeightReached, presumably so caller tries again.

core/src/poh_recorder.rs
sakridge: This looks good to me.

core/src/poh_service.rs
```diff
@@ -121,18 +163,66 @@ impl PohService {
...
+                    record_receiver.try_recv()
+                };
+                if let Ok(mut record) = mixin {
+                    let mut lock_time = Measure::start("lock");
+                    let mut poh_recorder_l = poh_recorder.lock().unwrap();
+                    lock_time.stop();
+                    total_lock_time_ns += lock_time.as_ns();
+                    loop {
+                        let mut temp = Vec::new();
+                        std::mem::swap(&mut temp, &mut record.transactions);
```
sakridge, line 182: `std::mem::take()` will be a little cleaner here I think

core/src/poh_service.rs
```diff
@@ -121,18 +163,68 @@ impl PohService {
...
-                let mut lock_time = Measure::start("lock");
-                let mut poh_l = poh.lock().unwrap();
-                lock_time.stop();
-                total_lock_time_ns += lock_time.as_ns();
-                let mut hash_time = Measure::start("hash");
-                let r = poh_l.hash(hashes_per_batch);
-                hash_time.stop();
-                total_hash_time_ns += hash_time.as_ns();
-                r
+                let mixin = if let Some(record) = try_again_mixin {
```
sakridge, line 169: The logic inside this got quite complicated. Maybe warrants it's own function to calculate this should_tick result now?

core/src/banking_stage.rs
```diff
@@ -2117,11 +2122,40 @@ mod tests {
...
 
+    use crate::poh_recorder::Record;
+    use std::sync::atomic::AtomicBool;
+    use std::sync::mpsc::Receiver;
+
+    fn simulate_poh(
+        record_receiver: Receiver<Record>,
+        poh_recorder: &Arc<Mutex<PohRecorder>>,
+    ) -> (JoinHandle<()>, Arc<AtomicBool>) {
+        let stop = Arc::new(AtomicBool::new(false));
```
sakridge, line 2141: Can we call this `exit`? That matches the rest of the code.

core/src/poh_recorder.rs
```diff
@@ -51,6 +51,78 @@ type Result<T> = std::result::Result<T, PohRecorderError>;
...
+        transactions: Vec<Transaction>,
+    ) -> Result<()> {
+        let res = self.record_sender.send(Record::new(
+            mixin,
+            transactions,
+            bank_slot,
+            self.result_sender.clone(),
+        ));
+        if res.is_err() {
+            return Err(PohRecorderError::MaxHeightReached); // re-use an error that gives appropriate behavior
```
sakridge, line 114: I would say: "If the channel is dropped, then the validator is shutting down so return that we are hitting the max tick height to stop transaction processing and flush any transactions in the pipeline."

core/src/poh_recorder.rs
```diff
@@ -51,6 +51,78 @@ type Result<T> = std::result::Result<T, PohRecorderError>;
...
+    pub result_receiver: Receiver<Result<()>>,
+}
+
+impl Clone for Recorder {
+    fn clone(&self) -> Self {
+        Recorder::new(self.record_sender.clone())
+    }
+}
+
+impl Recorder {
```
sakridge, line 90: Recorder is too generic. How about EntryRecorder or TransactionRecorder

core/src/banking_stage.rs
```diff
@@ -2117,11 +2122,40 @@ mod tests {
...
+            stop.store(true, Ordering::Relaxed);
+            let _ = poh_simulator.join();
+
             assert_eq!(bank.get_balance(&pubkey), 1);
         }
         Blockstore::destroy(&ledger_path).unwrap();
     }
 
+    use crate::poh_recorder::Record;
+    use std::sync::atomic::AtomicBool;
```
sakridge, line 2134: put at the top of the file/module

sakridge (commented):
Can we also add performance information from a gcp bench-tps run and see where we are in terms of slot time under load and just the TPS. Can run it on gcp manually or through the same configuration as #performance-results

sakridge:
Want to mark ready-for-review if that's the case?

core/src/poh_service.rs
```diff
@@ -105,78 +195,124 @@ impl PohService {
...
+            let should_tick = Self::record_or_hash(
+                &mut next_record,
+                &poh_recorder,
+                &mut timing,
+                &record_receiver,
+                hashes_per_batch,
+                &poh,
+            );
             if should_tick {
                 // Lock PohRecorder only for the final hash...
```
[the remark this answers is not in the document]
jeffwashington: it is still trueish. we (still) don't lock pohrecorder for individual hashes - only for the hash that completes the tick. We do also now lock pohrecorder when we call record. I can clarify

jeffwashington:
> Can we also add performance information from a gcp bench-tps run and see where we are in terms of slot time under load and just the TPS. Can run it on gcp manually or through the same configuration as #performance-results
[last perf run:](https://solanalabs.slack.com/archives/CP2L2S4KV/p1616376415024200)
mean_tps: 15067
max_tps: 28837
mean_confirmation_ms: 1586
max_confirmation_ms: 3513
99th_percentile_confirmation_ms: 2461
max_tower_distance: 31
last_tower_distance: 31
slots_per_second: 1.273

[this pr:](https://solanalabs.slack.com/archives/CP2L2S4KV/p1616442727000100)
mean_tps: 24424
max_tps: 70562
mean_confirmation_ms: 1777
max_confirmation_ms: 12530
99th_percentile_confirmation_ms: 5267
max_tower_distance: 64
last_tower_distance: 31
slots_per_second: 1.523

sakridge:
Seems to make poh_service.elapsed_ms a bit worse, but probably because it's going a lot faster.

poh-service-elapsed-ms

sakridge (approved):
lgtm

PR #15930, opened 2021-03-17
drop poh lock after record

jeffwashington:
#### Problem
Keeping poh lock longer than we need to after a record. @sakridge found this.

#### Summary of Changes
drop it
Fixes #


sakridge (approved):
lgtm

PR #15931, opened 2021-03-17
add metrics for tick producer and poh_recorder

jeffwashington:
#### Problem
We need visibility into poh metrics due to lag.
#### Summary of Changes
Add metrics. This is work from @sakridge 
Fixes #


sakridge (approved):
lgtm

PR #15946, opened 2021-03-17
feat: add support for signature received notifications

jstarry:
#### Problem
Lack of support for signature received notifications which indicate that a transaction landed in a slot

#### Summary of Changes
- Added new `onSignatureWithOptions` method which accepts an options object which can enable the received notification.

Fixes #

### DIFF
--- a/web3.js/src/connection.ts
+++ b/web3.js/src/connection.ts
@@ -403,6 +403,11 @@ const SignatureStatusResult = pick({
   err: TransactionErrorResult,
 });
 
+/**
+ * Transaction signature received notification
+ */
+const SignatureReceivedResult = literal('receivedSignature');
+
 export type Version = {
   'solana-core': string;
   'feature-set'?: number;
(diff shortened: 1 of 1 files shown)


web3.js/src/connection.ts
```diff
@@ -1460,13 +1467,38 @@ export type SignatureResultCallback = (
...
+/**
+ * Callback function for signature notifications
+ */
+export type SignatureSubscriptionCallback = (
+  notification: SignatureStatusNotification | SignatureReceivedNotification,
+  context: Context,
+) => void;
+
+/**
+ * Callback function for signature notifications
```
CriesofCarrots, line 1488: c/p comment. Update?

CriesofCarrots (commented):
lgtm (thanks for your patience!); one nit

web3.js/src/connection.ts
jstarry: Ah, thanks!

PR #15953, opened 2021-03-17
Bump bpf-tools to v1.2 and get rid of xargo

dmakarov:
Xargo is used to build Rust custom standard libraries. The updated Rust BPF toolchain bundles the pre-built standard libraries in the tarball. This makes the rustc compiling for bpfel-unknown-unknown target find the libraries at a known location, eliminating the need for xargo and custom sysroot path.


sdk/cargo-build-bpf/src/main.rs
```diff
@@ -142,9 +142,46 @@ fn build_bpf_package(config: &Config, target_directory: &Path, package: &cargo_m
...
+    rust_flags.push_str(linker_script.to_str().unwrap());
+    rust_flags.push_str(" -C link-arg=--Bdynamic");
+    rust_flags.push_str(" -C link-arg=-shared");
+    rust_flags.push_str(" -C link-arg=--threads=1");
+    rust_flags.push_str(" -C link-arg=--entry=entrypoint");
+    rust_flags.push_str(" -C linker=");
+    rust_flags.push_str(linker.to_str().unwrap());
+    env::set_var("RUSTFLAGS", rust_flags);
+
+    let xargo_build = PathBuf::from("cargo");
```
mvines, line 177: So happy we're able to kill xargo, let's scrub all references to it at the same time!
```suggestion
    let cargo_build = PathBuf::from("cargo");
```

sdk/cargo-build-bpf/src/main.rs
```diff
@@ -142,9 +142,46 @@ fn build_bpf_package(config: &Config, target_directory: &Path, package: &cargo_m
...
+        .join("dependencies")
+        .join("bpf-tools")
+        .join("llvm")
+        .join("bin");
+    env::set_var("CC", llvm_bin.join("clang"));
+    env::set_var("AR", llvm_bin.join("llvm-ar"));
+    env::set_var("OBJDUMP", llvm_bin.join("llvm-objdump"));
+    env::set_var("OBJCOPY", llvm_bin.join("llvm-objcopy"));
+    // Use the SDK's version of Rust to build for BPF
+    env::set_var("RUSTUP_TOOLCHAIN", "bpf");
```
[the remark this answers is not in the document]
dmakarov: It seems this can be removed at all, because cargo runs with `+bpf`. I'll remove it. Thank you for catching this.

dmakarov:
Is there some file cache on a CI runner? Can it be cleared?

sdk/cargo-build-bpf/src/main.rs
```diff
@@ -142,9 +142,46 @@ fn build_bpf_package(config: &Config, target_directory: &Path, package: &cargo_m
         println!("Legacy program feature detected");
     }
 
-    let xargo_build = config.bpf_sdk.join("rust").join("xargo-build.sh");
-    let mut xargo_build_args = vec![];
+    // instal the BPF toolchain
+    let install = config.bpf_sdk.join("scripts").join("install.sh");
```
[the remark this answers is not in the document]
dmakarov: I removed the dependency on `install.sh` in cargo-build-bpf.

sdk/cargo-build-bpf/src/main.rs
```diff
@@ -56,18 +57,125 @@ where
...
+            fs::remove_file(&target_path).map_err(|err| err.to_string())?;
+        }
+        let mut url = String::from(url);
+        url.push('/');
+        url.push_str(version);
+        url.push('/');
+        url.push_str(file.to_str().unwrap());
+        download_file(&url.as_str(), &file, false)?;
+        fs::create_dir_all(&target_path).map_err(|err| err.to_string())?;
+        let tar = PathBuf::from("tar");
```
mvines, line 111: The `tar` crate is nice and will avoid a system `tar` dependency.  Example usage at https://github.com/solana-labs/solana/blob/4bbeb9c033e4618e8be0ce9d9ad5e7edbc90f48b/runtime/src/snapshot_utils.rs#L739-L740

sdk/cargo-build-bpf/src/main.rs
```diff
@@ -56,18 +57,125 @@ where
...
+        };
+        if invalid_link {
+            fs::remove_file(&source_path).map_err(|err| err.to_string())?;
+        }
+        invalid_link
+    } else {
+        true
+    };
+    if missing_source {
+        std::os::unix::fs::symlink(target_path, source_path).map_err(|err| err.to_string())?;
```
mvines, line 145: We build `cargo-build-bpf` for windows too :)

sdk/cargo-build-bpf/src/main.rs
```diff
@@ -142,26 +250,89 @@ fn build_bpf_package(config: &Config, target_directory: &Path, package: &cargo_m
         println!("Legacy program feature detected");
     }
 
-    let xargo_build = config.bpf_sdk.join("rust").join("xargo-build.sh");
-    let mut xargo_build_args = vec![];
+    let criterion_version = if cfg!(target_os = "macos") {
```
mvines, line 253: Can we remove the Criterion dependency entirely here?  It's only used for C right?

dmakarov: I don't know. If it's not needed, sure, easy to remove.

sdk/cargo-build-bpf/src/main.rs
```diff
@@ -142,26 +250,89 @@ fn build_bpf_package(config: &Config, target_directory: &Path, package: &cargo_m
...
+    };
+    install_if_missing(
+        &config,
+        "criterion",
+        criterion_version,
+        "https://github.com/Snaipe/Criterion/releases/download",
+        &PathBuf::from(criterion_filename),
+    )
+    .expect("Failed to install criterion");
+    let bpf_tools_filename = if cfg!(target_os = "macos") {
```
mvines, line 271: For windows we can just error at this point

dmakarov: Can a BPF toolchain exist on Windows in principle (from some other source, not Solana)?  If it exists and we only need to link to it, maybe it's better to error at the point where it's determined that we need to download the bpf-tools package for Windows, which we can't.

mvines: It probably can't exist yet.  We build `cargo-build-bpf` for windows right now just to help Windows users discover that they need a Linux or macOS machine faster.  Otherwise `cargo build-bpf` fails in more mysterious ways.

So `cargo-build-bpf` successfully building for windows but failing at runtime is just fine for now

sdk/cargo-build-bpf/src/main.rs
dmakarov: done

sdk/cargo-build-bpf/src/main.rs
dmakarov: fixed

sdk/cargo-build-bpf/src/main.rs
dmakarov: removed

sdk/cargo-build-bpf/src/main.rs
dmakarov: added an exit on windows at the beginning of main()

sdk/cargo-build-bpf/src/main.rs
```diff
@@ -56,18 +60,125 @@ where
...
     }
+    output
+        .stdout
+        .as_slice()
+        .iter()
+        .map(|&c| c as char)
+        .collect::<String>()
+}
+
+// Check whether a package is installed and install it if it is missing.
```
mvines, line 84:
```suggestion
// Check whether a package is installed and install if missing.
```

mvines (approved):
Looks great to me!  @jackcmay, you care to review as well?

Do we want to bring this into v1.6?  I'm thinking yes, as it's still very early in the v1.6 stabilization cycle

sdk/cargo-build-bpf/src/main.rs
dmakarov: corrected

jackcmay (approved):
This is great work, nice to get rid of cargo!

dmakarov:
@mvines is it possible to clean .cache/solana on a buildkite agent? This error looks like there is something wrong in .cache/ https://buildkite.com/solana-labs/solana/builds/42475#7ed2c3ec-71b1-4ac5-a4d5-c8fbe3b4fbae/79-147

PR #15962, opened 2021-03-17
Sdk: remove integer arithmetic

CriesofCarrots:
#### Problem
Integer arithmetic in sdk modules

#### Summary of Changes
Switch to saturating math

Toward #15368

### DIFF
--- a/sdk/src/secp256k1_instruction.rs
+++ b/sdk/src/secp256k1_instruction.rs
@@ -1,4 +1,3 @@
-#![allow(clippy::integer_arithmetic)]
 #![cfg(feature = "full")]
 
 use crate::instruction::Instruction;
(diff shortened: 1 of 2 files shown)


CriesofCarrots:
@sakridge , can you please look at secp256k1_instruction and let me know what you think? There are certainly some additions/indexes that will never overflow, but I opted to use saturating math everywhere, rather than ignore clippy on a case-by-case basis. Open to other options, though.

sakridge:
> @sakridge , can you please look at secp256k1_instruction and let me know what you think? There are certainly some additions/indexes that will never overflow, but I opted to use saturating math everywhere, rather than ignore clippy on a case-by-case basis. Open to other options, though.

Yea. I'll look.

sakridge (dismissed):

t-nelson (approved):
lgtm!

PR #15968, opened 2021-03-18
feat: add support for blockTime on confirmed blocks

oJshua:
#### Problem
ConfirmedBlock types do not have blockTime, which would be helpful now that these are available. 

#### Summary of Changes
Add blockTime to ConfirmedBlock.


oJshua:
@jstarry making sure I've covered everything here with this new TypeScript version.

jstarry (approved):
Yup, looks correct!

PR #15979, opened 2021-03-18
Signature ellipsis for transaction history

### LINKED ISSUE #14975: Explorer: transaction history doesn't fit horizontally
#### Problem
With the addition of block times in transaction history, the rows no longer fit horizontally and require a scroll bar.

Explorer-Solana (4)

#### Proposed Solution
Either shorten the timestamp display or truncate transaction signatures.

Arrowana:
#### Problem
Horizontal scrollbar on history looks horrible, page too heavy when overlooking an account.

#### Summary of Changes
Trying out some stuff to make it look like this

image

Who reads the full signature anyway?

I imagine we would even truncate more of it on small media, as we still want to be able to browse without a horizontal scrollbar

I can't quite get how to do that cleanly though. Also this frees some space, if the web rpc ends up giving fee and sol balance change per transaction we can make something close to etherscan

Fixes #14975

### DIFF
--- a/explorer/src/scss/_solana.scss
+++ b/explorer/src/scss/_solana.scss
@@ -353,3 +353,10 @@ pre.data-wrap, pre.json-wrap {
 pre.json-wrap {
   max-width: 36rem;
 }
+
+td {
+  max-width: 300px;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+}
--- a/explorer/src/components/account/TransactionHistoryCard.tsx
+++ b/explorer/src/components/account/TransactionHistoryCard.tsx
@@ -121,8 +121,8 @@ export function TransactionHistoryCard({ pubkey }: { pubkey: PublicKey }) {
         </button>
       </div>
 
-      <div className="table-responsive mb-0">
-        <table className="table table-sm table-nowrap card-table">
+      <div className="table-responsive-sm mb-0">
+        <table className="table table-sm table-nowrap">
           <thead>
             <tr>
               <th className="text-muted w-1">Slot</th>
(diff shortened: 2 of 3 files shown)


oJshua:
I believe we do truncate most transaction signatures on smaller media, but I fully agree this could be handled better altogether. Scrolling isn't so horrible on a touch screen, but on a desktop you might have to find the scrollbar at the bottom of the component.

We do have a similar ticket: https://github.com/solana-labs/solana/issues/14975

@Arrowana thanks for you efforts here. We can keep spinning on it. I'll close for now since we're already tracking, but please feel free to resubmit a PR if you come up with a solution.

oJshua:
We do have this snippet of SCSS too if you wanted to expand on it:
https://github.com/solana-labs/solana/blob/master/explorer/src/scss/_solana.scss#L205-L227

Address components have a truncate parameter.

oJshua (changes requested):
Comments in main thread. We can keep this draft open for now!

explorer/src/scss/_solana.scss
```diff
@@ -353,3 +353,10 @@ pre.data-wrap, pre.json-wrap {
 pre.json-wrap {
   max-width: 36rem;
 }
+
+td {
```
oJshua, line 357: We should avoid having a global rule for table columns. See the snippet I posted.

Arrowana:
> We do have this snippet of SCSS too if you wanted to expand on it:
> https://github.com/solana-labs/solana/blob/master/explorer/src/scss/_solana.scss#L205-L227
> 
> Address/Signature components have a truncate parameter.

Yes, but it doesn't seem to kick in because the parent container is `table-responsive`, i passed `truncate` to Signature but it was still not truncated. I will try this force-truncate that i missed though.

Agree that this draft was a hack, I just presented that to show how it could look. I will investigate how to achieve it properly

oJshua:
You're right about it not kicking in without it hitting that media query! I've been trying out a different truncation method as seen in the screenshot below. Let me know if you have any thoughts. Thanks again @Arrowana! Feel free to reopen this or submit a new PR if you come up with anything else.

001_mint_transfers

PR #16001, opened 2021-03-18
web3.js: add support for batch getParsedConfirmedTransactions

oJshua:
#### Problem
There is not currently support for fetching many confirmed transactions at once.

#### Summary of Changes
This solution uses batched JSON-RPC requests to query getConfirmedTransaction.

### DIFF
--- a/web3.js/src/connection.ts
+++ b/web3.js/src/connection.ts
@@ -63,6 +63,13 @@ export const BLOCKHASH_CACHE_TIMEOUT_MS = 30 * 1000;
 
 type RpcRequest = (methodName: string, args: Array<any>) => any;
 
+type RpcBatchRequest = (requests: RpcParams[]) => any;
+
+type RpcParams = {
+  methodName: string,
+  args: Array<any>,
+};
+
 export type TokenAccountsFilter =
   | {
       mint: PublicKey;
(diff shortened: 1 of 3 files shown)


web3.js/src/connection.ts
```diff
@@ -63,6 +63,13 @@ export const BLOCKHASH_CACHE_TIMEOUT_MS = 30 * 1000;
 
 type RpcRequest = (methodName: string, args: Array<any>) => any;
 
+type RpcBatchRequest = (requests: RpcParams[]) => any;
+
+export type RpcParams = {
```
jstarry, line 68: Can you please add an `@internal` comment tag to this type to avoid exposing it on the doc site?
Screen Shot 2021-03-22 at 2 16 34 PM

web3.js/src/connection.ts
```diff
@@ -2465,6 +2498,33 @@ export class Connection {
     return res.result;
   }
 
+  /**
+   * Fetch parsed transaction details for a batch of confirmed transactions
+   */
+  async getParsedConfirmedTransactions(
+    signatures: TransactionSignature[],
+  ): Promise<any> {
```
jstarry, line 2506: Please add a type for the return value

jstarry (approved):
Looks good after fixing up the docs / types!

web3.js/src/connection.ts
oJshua: Oops, thanks for catching!

PR #16002, opened 2021-03-18
program-test: Fix warp and staking issue

joncinque:
#### Problem

Since program-test creates a test genesis and then modifies fees and rent,
some of the genesis accounts get rent-collected after warping.  Most
notably, `StakeConfig` gets rent-collected, causing any stake operations
to fail after warp.

#### Summary of Changes

Create genesis with the `Rent` and `FeeRateGovernor` actually used by the bank, which automatically puts enough lamports in the genesis accounts.

Fixes #

cc @ysavchenko


mvines (approved):
sneaky

PR #16045, opened 2021-03-20
feat: add logs subscription

armaniferrante:


web3.js/test/connection.test.ts
```diff
@@ -2002,6 +2003,32 @@ describe('Connection', () => {
       await connection.removeRootChangeListener(subscriptionId);
     });
 
+    it('logs notification', async () => {
+      let listener: number | undefined;
+      const owner = new Account();
+      const [logsRes, ctx] = await new Promise(resolve => {
+        const filter = 'all' as LogsFilter;
```
armaniferrante, line 2010: This is awkward, but I'm not sure how to make TypeScript happy without the cast.

oJshua (approved):
Looks great!

web3.js/test/connection.test.ts
jstarry: Haha, TypeScript was trying to tell you that your logs filter typedef has an issue: `'all '` (note the extra space)

web3.js/src/connection.ts
```diff
@@ -3106,6 +3174,55 @@ export class Connection {
     }
   }
 
+  /**
+   * Registers a callback to be invoked whenever logs are emitted.
```
jstarry, line 3178: Can you please add a comment saying that this API is currently unstable? Or @mvines can you comment on whether the logs pubsub api can be stable now?

web3.js/src/connection.ts
```diff
@@ -1524,6 +1524,54 @@ type RootSubscriptionInfo = {
...
+ */
+const LogsNotificationResult = pick({
+  result: notificationResultAndContext(LogsResult),
+  subscription: number(),
+});
+
+/**
+ * Filter for log subscriptions.
+ */
+export type LogsFilter = {mentions: string[]} | 'all ' | 'allWithVotes';
```
jstarry, line 1558: Currently the logs pubsub api [only supports one mention](https://github.com/solana-labs/solana/blob/2c4ff000b9dc28b5e9dc89c016e67aa1f3440825/core/src/rpc_pubsub.rs#L344), so I think making that explicit for now would be better (we can expand on the type union later to allow an array). Also, can you change the mention type from string to `PublicKey` and then handle the base58 conversion internally?

jstarry (commented):
Thanks @armaniferrante!

web3.js/src/connection.ts
mvines: I'm game for declaring it stable.

web3.js/test/connection.test.ts
```diff
@@ -2002,6 +2002,27 @@ describe('Connection', () => {
       await connection.removeRootChangeListener(subscriptionId);
     });
 
+    it('logs notification', async () => {
+      let listener: number | undefined;
+      const owner = new Account();
+      const [logsRes, ctx] = await new Promise(resolve => {
+        listener = connection.onLogs('all', (logs, ctx) => {
```
jstarry, line 2009: You'll want to add a commitment to avoid the test timeout here. By default the "finalized" commitment is used so the logs won't be delivered for awhile. I think "processed" would be good here

web3.js/test/connection.test.ts
```diff
@@ -2002,6 +2002,27 @@ describe('Connection', () => {
...
+      });
+      expect(ctx.slot).to.be.greaterThan(0);
+      expect(logsRes.logs.length).to.eq(2);
+      expect(logsRes.logs[0]).to.eq(
+        'Program 11111111111111111111111111111111 invoke [1]',
+      );
+      expect(logsRes.logs[1]).to.eq(
+        'Program 11111111111111111111111111111111 success',
+      );
+      await connection.removeOnLogsListener(listener!);
```
jstarry, line 2023: what is that bang for?

armaniferrante: To tell TypeScript `listener` is not undefined, though it's not necessary it seems.

jstarry: That's super useful, thanks for the tip!

web3.js/test/connection.test.ts
armaniferrante: The test still intermittently fails with `processed`. Investigating.

web3.js/test/connection.test.ts
armaniferrante: Oops I was incorrect. It is indeed necessary, here.

web3.js/test/connection.test.ts
armaniferrante: Only the browser tests have the problem (with or without the commitment param).

armaniferrante: I believe the problem is a race condition between setting up the subscription in `onLogs` and subsequently executing the the transaction to trigger the logs.

If the transaction to trigger the log gets executed before the subscription is setup, the event listener is never fired and so the promise is never resolved. 

I've added a `sleep` to work around this race condition.

jstarry: Ah makes sense, thanks for investigating!

jstarry (approved):

PR #16051, opened 2021-03-22
Handle blockstore insert dup checks

carllin:
#### Problem
should_insert_data_shred() should check just_inserted_data_shreds() for certain shreds

#### Summary of Changes
Add `get_data_shred_from_just_inserted_or_db()` to fetch data shreds safely

Fixes #

### DIFF
--- a/ledger/src/blockstore.rs
+++ b/ledger/src/blockstore.rs
@@ -1266,78 +1266,90 @@ impl Blockstore {
         leader_schedule: Option<&Arc<LeaderScheduleCache>>,
         is_recovered: bool,
     ) -> bool {
-        let shred_index = u64::from(shred.index());
-        let slot = shred.slot();
-        let last_in_slot = if shred.last_in_slot() {
-            debug!("got last in slot");
-            true
-        } else {
-            false
-        };
+        loop {
+            let shred_index = u64::from(shred.index());
+            let slot = shred.slot();
+            let last_in_slot = if shred.last_in_slot() {
+                debug!("got last in slot");
+                true
+            } else {
+                false
+            };
 
-        // Check that we do not receive shred_index >= than the last_index
-        // for the slot
-        let last_index = slot_meta.last_index;
-        if shred_index >= last_index {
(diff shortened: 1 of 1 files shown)


ledger/src/blockstore.rs
```diff
@@ -1262,82 +1264,110 @@ impl Blockstore {
...
+                        self.get_data_shred(slot, last_index)
+                            .unwrap()
+                            .map(Cow::Owned)
+                    };
+
+                if ending_shred.is_none() {
+                    // The slot data might have been cleared between us fetching the SlotMeta and
+                    // retrieving the `ending_shred`. Slot data might have been cleared by
+                    // `Blockstore::clear_unconfirmed_slot()` due to events like duplicate block detection.
+                    // In this case, try to get the updated SlotMeta and try again
```
jstarry, line 1303: Looks like this is going to need a call to `get_slot_meta_entry` or something similar to actually get updated `SlotMeta`, right? Looping again isn't going to change anything

jstarry: We should probably have an exit hatch for if `SlotMeta` ever gets corrupted such that its `last_index` is still set but the shred doesn't exist

jstarry: Maybe for now we can just have this return an error? This might actually be an invalid shred

carllin: ha shoot forgot to purge this, it's no longer necessary

jstarry (dismissed):

sakridge (dismissed):
lgtm

carllin:
Ah dang, there's an annoying case here for duplicate blocks as well.

If somebody sends a batch with two shreds both last_index set to true, say for two shred indexes in the order `10, 9`. Then the shred_index `9` won't be inserted because it fails the should_insert check. If shred `9` is part of the confirmed version of this slot, then this slot will never finish replay, and moreover, the slot is not marked dead because it's never replayed because the entries are never successfully deserialized (blockstore just sees a hole for shred `9` and is waiting for a consecutive set of shreds to deserialize)

Issue #16048, opened 2021-03-22
rpc: no way to limit the response size of getLeaderSchedule


jstarry:
#### Problem
The leader schedule response is very large (~2.3MB currently for mainnet-beta) because it always returns the full schedule for the current epoch. There's no way to limit the size of the response if a client just wants to fetch the schedule for the next N slots.

#### Proposed Solution
Add an option for requesting the leaders of the next N slots and if the response will cross over into the next epoch, return those leaders as well.

PR #16057, opened 2021-03-22
Add GetSlotLeaders RPC method

jstarry:
#### Problem
No way to get the upcoming leaders without fetching the entire leader schedule. It's also a pain for clients to get upcoming leaders across epoch boundaries.

#### Summary of Changes
- Add new `getSlotLeaders` method which is similar to `getSlotLeader` but fetches leaders for a given slot range
- Use leader schedule cache to avoid expensive leader schedule computation and reduce memory overhead

Fixes https://github.com/solana-labs/solana/issues/16048

### DIFF
--- a/core/src/rpc.rs
+++ b/core/src/rpc.rs
@@ -30,13 +30,17 @@ use solana_client::{
         TokenAccountsFilter, DELINQUENT_VALIDATOR_SLOT_DISTANCE, MAX_GET_CONFIRMED_BLOCKS_RANGE,
         MAX_GET_CONFIRMED_SIGNATURES_FOR_ADDRESS2_LIMIT,
         MAX_GET_CONFIRMED_SIGNATURES_FOR_ADDRESS_SLOT_RANGE, MAX_GET_PROGRAM_ACCOUNT_FILTERS,
-        MAX_GET_SIGNATURE_STATUSES_QUERY_ITEMS, MAX_MULTIPLE_ACCOUNTS, NUM_LARGEST_ACCOUNTS,
+        MAX_GET_SIGNATURE_STATUSES_QUERY_ITEMS, MAX_GET_SLOT_LEADERS, MAX_MULTIPLE_ACCOUNTS,
+        NUM_LARGEST_ACCOUNTS,
     },
     rpc_response::Response as RpcResponse,
     rpc_response::*,
 };
 use solana_faucet::faucet::request_airdrop_transaction;
-use solana_ledger::{blockstore::Blockstore, blockstore_db::BlockstoreError, get_tmp_ledger_path};
+use solana_ledger::{
+    blockstore::Blockstore, blockstore_db::BlockstoreError, get_tmp_ledger_path,
+    leader_schedule_cache::LeaderScheduleCache,
+};
 use solana_metrics::inc_new_counter_info;
 use solana_perf::packet::PACKET_DATA_SIZE;
 use solana_runtime::{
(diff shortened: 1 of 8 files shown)


jstarry:
@mvines since the leader schedule response returns epoch slot indices, I decided not to support cross-epoch responses. I don't think it's too awful for the client to handle when they are near an epoch boundary themselves.

mvines:
> @mvines since the leader schedule response returns epoch slot indices, I decided not to support cross-epoch responses. I don't think it's too awful for the client to handle when they are near an epoch boundary themselves.

Yeah, ok.   I foresee a small stream of epoch-boundary bugs due to this but I see how the current interface makes doing better troublesome

core/src/rpc.rs
```diff
@@ -1963,37 +1970,45 @@ pub mod rpc_minimal {
...
-            commitment: Option<CommitmentConfig>,
+            config: Option<RpcLeaderScheduleConfig>,
         ) -> Result<Option<RpcLeaderSchedule>> {
-            let bank = meta.bank(commitment);
-            let slot = slot.unwrap_or_else(|| bank.slot());
-            let epoch = bank.epoch_schedule().get_epoch(slot);
-
-            debug!("get_leader_schedule rpc request received: {:?}", slot);
+            let limit = config.as_ref().and_then(|c| c.limit).unwrap_or_default();
+            if limit > DEFAULT_SLOTS_PER_EPOCH as usize {
```
sakridge, line 1976: Can it be based on actual slots_per_epoch. get it from the bank below.

docs/src/developing/clients/jsonrpc-api.md
```diff
@@ -1583,7 +1583,10 @@ Returns the leader schedule for an epoch
 #### Parameters:
 
 - `<u64>` - (optional) Fetch the leader schedule for the epoch that corresponds to the provided slot. If unspecified, the leader schedule for the current epoch is fetched
-- `<object>` - (optional) [Commitment](jsonrpc-api.md#configuring-state-commitment)
+- `<object>` - (optional) Configuration object containing the following optional fields:
+  - (optional) [Commitment](jsonrpc-api.md#configuring-state-commitment)
+  - (optional) `limit: <number>` - Limit the number of upcoming slots to fetch from the queried leader schedule. (Max: 432,000)
```
mvines, line 1588: Hmm, I'm not liking now how `limit` causes the change in starting slot index from 0 to the current slot number.

What if we added something new instead, `getLeaderScheduleWithLimit`, with params `start slot` and `limit` (same params as [getConfirmedBlocksWithLimit](https://docs.solana.com/developing/clients/jsonrpc-api#getconfirmedblockswithlimit))

`limit` should less than 5,000 or so (so max ~250kb response) and just returns a vector of pubkeys for the slots (no dictionary), which makes it easy to return results across an epoch boundary as well

jstarry: That's fair. The client experience of getting pubkeys instead of slot indexes would be much better. I'll mark the pre-existing `getLeaderSchedule` method as deprecated while I'm at it

mvines: For larger ranges, like what's needed to figure block production over an epoch,  the existing `getLeaderSchedule` is still nice because it avoids an RPC response of ~19MB.  So I think there's room for both to exist

jstarry: Cool, left it alone. I went with a new `GetSlotLeaders` method which fits better with your idea I think

mvines (approved):
nice!

mvines:
(adding v1.5 and v1.6, I don't want to wait for v1.7 to use this)

jstarry:
> (adding v1.5 and v1.6, I don't want to wait for v1.7 to use this)

Slipped my mind, thanks

PR #16060, opened 2021-03-22
solana transfer now requires --allow-unfunded-recipient if the recipient doesn't exist

mvines:
#### Problem
Users are losing funds because we never implemented https://github.com/solana-labs/solana/issues/6970

#### Summary of Changes
```
$ solana-keygen new -o ./new-account.json
$ solana transfer ./new-account.json 1
Error: The recipient address (9WtBoGqU1PwAvZwRGiS17zEhr8srDw8KdbSMxHkBpQuk) is not funded. Add `--allow-unfunded-recipient` to complete the transfer 
$ solana transfer ./new-account.json 1 --allow-unfunded-recipient
Signature: 4Z8xMnFqQKmHepZg6cotwHXdKJyTAtgv7DdKGWK4Y9aTAfJKTZku1RE56VWH41fx6enwVDaZQ6tgcMU8REmdPzVu
```

This now aligns with how `spl-token transfer` works as well.

#### Migration plan:
* Backport the `--allow-unfunded-recipient` flag to v1.5 as a hidden/ignored argument (https://github.com/solana-labs/solana/pull/16060)
* Update Exchange docs to suggest `--allow-unfunded-recipient` be now used, and update known users
* Backport this full PR to v1.6

This is a breaking change to the CLI, but a loss of funds is worse.

### DIFF
--- a/cli/src/cli.rs
+++ b/cli/src/cli.rs
@@ -360,6 +360,7 @@ pub enum CliCommand {
         from: SignerIndex,
         sign_only: bool,
         dump_transaction_message: bool,
+        allow_unfunded_recipient: bool,
         no_wait: bool,
         blockhash_query: BlockhashQuery,
         nonce_account: Option<Pubkey>,
@@ -865,6 +866,7 @@ pub fn parse_command(
             let (fee_payer, fee_payer_pubkey) =
                 signer_of(matches, FEE_PAYER_ARG.name, wallet_manager)?;
             let (from, from_pubkey) = signer_of(matches, "from", wallet_manager)?;
+            let allow_unfunded_recipient = matches.is_present("allow_unfunded_recipient");
 
             let mut bulk_signers = vec![fee_payer, from];
             if nonce_account.is_some() {
@@ -886,6 +888,7 @@ pub fn parse_command(
                     to,
                     sign_only,
                     dump_transaction_message,
+                    allow_unfunded_recipient,
                     no_wait,
                     blockhash_query,
                     nonce_account,
@@ -1139,6 +1142,7 @@ fn process_transfer(
     from: SignerIndex,
     sign_only: bool,
     dump_transaction_message: bool,
+    allow_unfunded_recipient: bool,
     no_wait: bool,
     blockhash_query: &BlockhashQuery,
     nonce_account: Option<&Pubkey>,
@@ -1152,6 +1156,21 @@ fn process_transfer(
 
     let (recent_blockhash, fee_calculator) =
         blockhash_query.get_blockhash_and_fee_calculator(rpc_client, config.commitment)?;
+
+    if !allow_unfunded_recipient {
+        let recipient_balance = rpc_client
+            .get_balance_with_commitment(to, config.commitment)?
+            .value;
+        if recipient_balance == 0 {
+            return Err(format!(
+                "The recipient address ({}) is not funded. \
+                                Add `--allow-unfunded-recipient` to complete the transfer \
+                               ",
+                to
+            )
+            .into());
+        }
+    }
 
     let nonce_authority = config.signers[nonce_authority];
     let fee_payer = config.signers[fee_payer];
@@ -1822,6 +1841,7 @@ pub fn process_command(config: &CliConfig) -> ProcessResult {
             from,
             sign_only,
             dump_transaction_message,
+            allow_unfunded_recipient,
             no_wait,
             ref blockhash_query,
             ref nonce_account,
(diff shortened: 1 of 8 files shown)


t-nelson:
Should we frame this in terms of the account not existing rather than by balance? The notion of "empty" doesn't seem to communicate much w/o context. "Recipient does not exist, pass --allow-create-recipient" on the other hand is quite clear

mvines:
> Should we frame this in terms of the account not existing rather than by balance? The notion of "empty" doesn't seem to communicate much w/o context. "Recipient does not exist, pass --allow-create-recipient" on the other hand is quite clear

The recipient likely does exist though, that is the keypair exists.  There's just no on-chain balance held by the account.

I feel like saying recipient doesn't exist would be more confusing.  "It does exist!  I `solana-keygen new`ed it"

t-nelson:
> The recipient likely does exist though, that is the keypair exists. There's just no on-chain balance held by the account.
> 
> I feel like saying recipient doesn't exist would be more confusing. "It does exist! I `solana-keygen new`ed it"

Hmm... I've been trying to correct this notion of "my keypair is my account" whenever I come across it (usually new validators who botched vote account creation).  Do you think this is a wide-spread view of how things work?

Regardless, my main gripe is with "empty", I would probably allow something more explicit "zero-balance"

mvines:
> Regardless, my main gripe is with "empty", I would probably allow something more explicit "zero-balance"

`--alllow-unfunded-recipient`?

Here's what sollet presents in this case:
image

t-nelson:
> `--allow-unfunded-recipient`?
> 
> Here's what sollet presents in this case:
> image
> 
> or maybe just: `--unfunded-recipient`

I can get on board with that!

Part of me wants to suggest that it's safe to suppress this warning in the case of valid 44-char base58 encoded pubkeys, but someone mentioned recently that a user didn't realize the pubkeys were case-sensitive and burned funds that way :facepalm:

mvines:
> Part of me wants to suggest that it's safe to suppress this warning in the case of valid 44-char base58 encoded pubkeys, but someone mentioned recently that a user didn't realize the pubkeys were case-sensitive and burned funds that way 🤦

Not safe!  Copy-and-paste an address and missing the first or last character and you likely have a valid 43-char base58 encoded pubkey.  😭

t-nelson:
> ...  valid 44-char ...

But yeah, lots of ways to screw this up.  Once we invent time travel I'm going back to push harder for checksums, breakage be damned

cli/tests/transfer.rs
```diff
@@ -423,6 +434,53 @@ fn test_transfer_all() {
...
+        blockhash_query: BlockhashQuery::All(blockhash_query::Source::Cluster),
+        nonce_account: None,
+        nonce_authority: 0,
+        fee_payer: 0,
+        derived_address_seed: None,
+        derived_address_program_id: None,
+    };
+
+    // Expect failure due to unfunded recipient and the lack of the `allow_unfunded_recipient` flag
+    process_command(&config).unwrap_err();
```
mvines, line 481: new test

cli/src/cli.rs
```diff
@@ -2945,11 +2974,12 @@ mod tests {
             }
         );
 
-        // Test Transfer no-wait
+        // Test Transfer no-wait and --allow-unfunded-recipient
         let test_transfer = test_commands.clone().get_matches_from(vec![
             "test",
             "transfer",
             "--no-wait",
+            "--allow-unfunded-recipient",
```
mvines, line 2982: new test

mvines:
@ryoqun - heads up that this'll affect at least one exchange

mvines:
Travis is red due to:
> anyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit

Ugh

mvines:
@t-nelson - does this seem ok?  You r+'ed the other support PRs but not the most important one, this one :)

t-nelson:
> @t-nelson - does this seem ok? You r+'ed the other support PRs but not the most important one, this one :)

Whoops sorry! I thought I already reviewed this one. Sec..

t-nelson (dismissed):
lgtm with the suggestion applied (feel free to DeMorgan)

cli/src/cli.rs
```diff
@@ -1153,6 +1157,21 @@ fn process_transfer(
     let (recent_blockhash, fee_calculator) =
         blockhash_query.get_blockhash_and_fee_calculator(rpc_client, config.commitment)?;
 
+    if !allow_unfunded_recipient {
```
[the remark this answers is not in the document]
mvines: ah, cool

PR #9698, opened 2020-04-24
Setup ReplayStage confirmation scaffolding for duplicate slots

carllin:
#### Problem
Duplicate versions of a slot potentially partition and stall the cluster

#### Summary of Changes
Here's how I imagine the full fix working:

What's in this PR (goal is to prevent voting on duplicate slots asap):
1) Duplicate slots are detected in WindowService and sent to ReplayStage through a channel.  ReplayStage processes them here: https://github.com/solana-labs/solana/compare/master...carllin:FixReplayStage?expand=1#diff-5984b6b0429f857c13a2a362669ab37cR533. Two cases:

  * a) If the duplicate fork is already confirmed, we (eventually) signal the cluster here which version of this slot we have: https://github.com/solana-labs/solana/compare/master...carllin:FixReplayStage?expand=1#diff-5984b6b0429f857c13a2a362669ab37cR543. This is because only 1 fork should be confirmed unless > 1/3 of the cluster is malicious. 

 * b) If ReplayStage processes and marks those duplicates/entire fork descended from those duplicate slots here: https://github.com/solana-labs/solana/compare/master...carllin:FixReplayStage?expand=1#diff-5984b6b0429f857c13a2a362669ab37cR559-R563

2) In 1b) after a fork has been marked duplicate, we remove it as a candidate for voting here: 
https://github.com/solana-labs/solana/compare/master...carllin:FixReplayStage?expand=1#diff-5984b6b0429f857c13a2a362669ab37cR1326-R1331. However, if you've already voted on this fork, and you can't generate a switching prof, then you will continue to generate banks on this fork 
 to avoid liveness issues (more details here: https://github.com/solana-labs/solana/compare/master...carllin:FixReplayStage?expand=1#diff-5984b6b0429f857c13a2a362669ab37cR1271-R1278)

3) If a duplicate slot is confirmed, the entire fork is added back into the candidate set by clearing the duplicate flag here: https://github.com/solana-labs/solana/compare/master...carllin:FixReplayStage?expand=1#diff-fb925e9cb1a6c2044ceaa55aa7c8f255R393

What's in follow-up PR:

1) Need to detect if cluster has confirmed some alternate version `V` of a duplicate slot (Imagine a validator has a dead slot, or a valid, playable version of a slot, but rest of the cluster confirmed a different slot). Can this be done by having another version of EpochSlots, but for confirmed slots instead of completed slots?

[...]

PR #16071, opened 2021-03-23
Purge slots greater than new last index

carllin:
#### Problem
Problem discovered here (copy and pasted):  https://github.com/solana-labs/solana/pull/16051#issuecomment-804458083

In the new dup block handling logic here: https://github.com/solana-labs/solana/pull/9698, dup blocks are not marked dead on detection, they are marked as invalid candidates for fork choice.

If somebody sends a batch with two shreds both last_index set to true, say for two shred indexes in the order `10, 9`. Then the shred_index `9` won't be inserted because it fails the should_insert check. If shred `9` is part of the confirmed version of this slot, then this slot will never finish replay, and moreover, the slot is not marked dead and thus can't be dumped because it's never replayed (the entries are never successfully deserialized blockstore just sees a hole for shred `9` and is waiting for a consecutive set of shreds to deserialize.

#### Summary of Changes

Fixes #

### DIFF
--- a/core/src/window_service.rs
+++ b/core/src/window_service.rs
@@ -270,13 +270,6 @@ where
                                     }
                                 };
                                 if shred_filter(&shred, last_root) {
-                                    // Mark slot as dead if the current shred is on the boundary
-                                    // of max shreds per slot. However, let the current shred
-                                    // get retransmitted. It'll allow peer nodes to see this shred
-                                    // and trigger them to mark the slot as dead.
-                                    if shred.index() >= (MAX_DATA_SHREDS_PER_SLOT - 1) as u32 {
-                                        let _ = blockstore.set_dead_slot(shred.slot());
-                                    }
                                     packet.meta.slot = shred.slot();
                                     packet.meta.seed = shred.seed();
                                     Some((shred, repair_info))
--- a/ledger/src/blockstore.rs
+++ b/ledger/src/blockstore.rs
@@ -1180,6 +1180,11 @@ impl Blockstore {
         let slot_meta = &mut slot_meta_entry.new_slot_meta.borrow_mut();
 
         if !is_trusted {
+            if Self::is_data_shred_present(&shred, slot_meta, &index_meta.data()) {
+                handle_duplicate(shred);
+                return Err(InsertDataShredError::Exists);
+            }
+
             if shred.last_in_slot() && shred_index < slot_meta.received && !slot_meta.is_full() {
                 // We got a last shred < slot_meta.received, which signals there's an alternative,
                 // shorter version of the slot. Because also `!slot_meta.is_full()`, then this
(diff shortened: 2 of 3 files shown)


ledger/src/blockstore.rs
```diff
@@ -1366,7 +1412,17 @@ impl Blockstore {
...
                     )
                 );
-            return false;
+
+            // If consumed <= shred_index, then We haven't gotten all the shreds up to `shred_index` yet.
+            // In this case, there's a chance hat since `shred_index` < the current last index, this shred
+            // index is necessary to complete the entry for replay. Worst case it's the wrong version,
+            // replay marks the slot as dead. However, if we don't insert here, then there's a chance we never
+            // get all the shreds < the current last index, never replay, and make no progress (for instance if
+            // a leader sends an additional detached "last index" shred with a very high index, but none of the
```
jstarry, line 1421: Isn't it still possible to get stuck if a node received a very high shred index for an unrepairable block and they already consumed more shreds than the *real* block? It feels to me that the block should be marked dead if it ever gets a last shred that's less than its current last shred. Then your dupe handling could repair the full slot if it detects a dupe?

core/src/window_service.rs
```diff
@@ -228,13 +228,6 @@ where
                                     }
                                 };
                                 if shred_filter(&shred, last_root) {
-                                    // Mark slot as dead if the current shred is on the boundary
```
jstarry, line 231: I opted to just drop this check since it doesn't provide much value. Think it deserves a feature switch?

ledger/src/blockstore.rs
```diff
@@ -1182,6 +1182,19 @@ impl Blockstore {
...
+                // We got a last shred < slot_meta.received, which signals there's an alternative,
+                // shorter version of the slot. Because also `!slot_meta.is_full()`, then this
+                // means, for the current version of the slot, we might never get all the
+                // shreds < the current last index, never replaythis slot, and make no
+                // progress (for instance if a leader sends an additional detached "last index"
+                // shred with a very high index, but none of the intermediate shreds). Ideally, we would
+                // just purge all shreds > the new last index slot, but because replay may have already
+                // replayed entries past the newly deteced "last" shred, then mark the slot as dead
+                // and wait for replay to dump and repair the correct version.
+                write_batch.put::<cf::DeadSlots>(slot, &true).unwrap();
```
jstarry, line 1195: I added https://github.com/solana-labs/solana/pull/16071/commits/e8f58cae6871e79336717f1b35b18ec0f2f31e15 to prevent a potential race condition where a slot is marked dead after `get_slot_entries` checks if the slot is dead and before it actually fetches entries (which could be full). Race condition was detected and fixed in this test: https://github.com/solana-labs/solana/pull/16071/commits/5b49ba6605268d1dd851a5baa587fc8ba0e183d7

PR #16091, opened 2021-03-24
make AccountSharedData.data private to abstract storage

jeffwashington:
#### Problem
All the work has been done to allow abstracting/hiding the implementation of the data on an account. Making it private now will retain the ability to change the storage format of an account in the (near?) future.

#### Summary of Changes
make data private. Callers use existing accessor methods.

Fixes #


sakridge (approved):
lgtm

PR #16092, opened 2021-03-24
poh record metrics

jeffwashington:
#### Problem
now that we do record from poh, we need to track the time spent in record.

#### Summary of Changes
track the time.

Fixes #


sakridge (approved):
lgtm

Issue #14475, opened 2021-01-07
explorer: stake breakdown details


aeyakovenko:
1. show circulating supply stake totals and distribution
2. show non circulating supply stake totals and distribution

PR #16095, opened 2021-03-24
Explorer: introduce circulating supply, active stake, and price on cluster stats page

oJshua:
#### Problem
Explorer doesn't show any information about active stake within relation to circulating supply. Also price information and deliquent stake.

#### Summary of Changes
Explorer-Solana (23)

Fixes https://github.com/solana-labs/solana/issues/14475


jstarry (approved):
This looks great! Added a few suggestions. Another thing we have to start thinking about is bandwidth bloat. Adding `getVoteAccounts` means an additional 690KB of data that gets fetched every time someone loads the main page.

Current breakdown is:
`getVoteAccounts`: 690,215
`getSupply`: 155,305
`getRecentPerformanceSamples`: 28,436
`getAccountInfo` (Stake history): 19,778

JavaScript files and the token list comparatively come in at 500KB and 30.7KB respectively but can be cached by the browser and so clients won't get the bandwidth hit each time. I think that a stake stats endpoint is in order to cut out that huge `getVoteAccounts` request and also maybe an option to remove the "non-circulating accounts" list from `getSupply`? Or at least cache that?

oJshua:
Noted on bandwidth considerations. I think a new stake stats endpoint is in store.

PR #16096, opened 2021-03-24
set target hash/sec to half of capability for auto

jeffwashington:
#### Problem
Working on poh, slot drift, high tps. In tests like bench-tps, config says auto hashes_per_tick. For auto, we want to use 50% of the ability so we have time for the poh thread to be interrupted with lock contention, waiting on record, tick, etc.
@sakridge wrote this code initially as an experiment.
#### Summary of Changes
set auto hashes/tick to 50% of max
Fixes #


sakridge (approved):
lgtm

PR #16098, opened 2021-03-24
sdk: Add try_from_slice_unchecked for Borsh

joncinque:
#### Problem

If we're using variable-sized structures on-chain, for example `Vec<>`, and we allocate 1000 bytes and don't immediately use all of them, Borsh fails deserialization because it doesn't read the entire slice.

#### Summary of Changes

Add a helper function.  This is a copy of https://github.com/near/borsh-rs/blob/3223286bae05d29162eaf37512966b9fa7d9cfdb/borsh/src/de/mod.rs#L30 which just removes the `is_empty()` check.

Fixes #

### DIFF
--- a/sdk/program/src/borsh.rs
+++ b/sdk/program/src/borsh.rs
@@ -1,7 +1,13 @@
 #![allow(clippy::integer_arithmetic)]
 //! Borsh utils
-use borsh::schema::{BorshSchema, Declaration, Definition, Fields};
-use std::collections::HashMap;
+use {
+    borsh::{
+        maybestd::io::Error,
+        schema::{BorshSchema, Declaration, Definition, Fields},
+        BorshDeserialize,
+    },
+    std::collections::HashMap,
+};
 
 /// Get packed length for the given BorchSchema Declaration
 fn get_declaration_packed_len(
@@ -53,4 +59,11 @@ fn get_declaration_packed_len(
 pub fn get_packed_len<S: BorshSchema>() -> usize {
     let schema_container = S::schema_container();
     get_declaration_packed_len(&schema_container.declaration, &schema_container.definitions)
+}
+
+/// Deserializes without checking that the entire slice has been consumed
+pub fn try_from_slice_unchecked<T: BorshDeserialize>(data: &[u8]) -> Result<T, Error> {
+    let mut data_mut = data;
+    let result = T::deserialize(&mut data_mut)?;
+    Ok(result)
 }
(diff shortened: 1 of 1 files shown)


sdk/program/src/borsh.rs
```diff
@@ -54,3 +60,10 @@ pub fn get_packed_len<S: BorshSchema>() -> usize {
     let schema_container = S::schema_container();
     get_declaration_packed_len(&schema_container.declaration, &schema_container.definitions)
 }
+
+/// Deserializes without checking that the entire slice has been consumed
+pub fn try_from_slice_unchecked<T: BorshDeserialize>(data: &[u8]) -> Result<T, Error> {
```
mvines, line 65: How about a little test for this guy?

joncinque: Certainly

mvines (approved):
lgtm!

sdk/program/src/borsh.rs
jackcmay: Unchecked sounds potentially dangerous, should folks be cautious using this function and if so what should they pay attention to?

joncinque: Hm, that's a good point.  The main difference is that this function eliminates a check that the slice has been totally read, so it lets you work with overallocated buffers.  The normal `try_from_slice` returns an error if there's data left in the buffer after deserialization.  How could we clarify that?

jackcmay: Is there a downside to always using the "unchecked" version?

joncinque: The length check ensures that whatever you're expecting to deserialize uses up all of the bytes in the buffer, so you could potentially run into an issue if you always use unchecked.  Any buffer that's big enough could work to deserialize your type, so if you pass the wrong buffer with unchecked, the error wouldn't get caught.

jackcmay: Any buffer greater than or equal to the expected size would work, correct?  Maybe call it "unsized", "unchecked" is fine but should probably include something to effect of the description you have above as what the "unchecked" means practically to the developer

joncinque: Correct -- I'll rename the function and add the description then

jackcmay: With the good description you have provided I think "unchecked" makes sense, I can go either way but "unchecked" is more consistent with rust norms

joncinque: It's also consistent with the PR that I put in for Borsh JS :-D

jackcmay (approved):

PR #16104, opened 2021-03-24
potentially poll for new records per hash batch

jeffwashington:
#### Problem
We want to improve tps, reduce excessive slot time.
#### Summary of Changes
After every hash batch, check to see if we are running ahead of schedule sufficiently. If so, release the poh lock and poll the record receiver until we are back on schedule (instead of busy waiting). This leaves poh with ticks available throughout the slot time and should greatly reduce any waiting at the end of a slot where we are unable to accept a record. It also means we can respond immediately when a record occurs.
Fixes #

### DIFF
--- a/core/src/poh_service.rs
+++ b/core/src/poh_service.rs
@@ -263,13 +263,10 @@ impl PohService {
                     }
                     // check to see if a record request has been sent
                     let get_again = record_receiver.try_recv();
-                    match get_again {
-                        Ok(record) => {
-                            // remember the record we just received as the next record to occur
-                            *next_record = Some(record);
-                            break;
-                        }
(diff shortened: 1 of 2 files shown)


jeffwashington:
Highest TPS: 140688.05 sampling period 1s max transactions: 9070238 clients: 1 drop rate: 0.54
[2021-03-24T21:37:58.960710645Z INFO  solana_bench_tps::bench] 	Average TPS: 100386.95
slot high of 471
This is with carl’s change as it is in master, ie. should be working.
My metrics from before are not comparable now that we’ve cut auto hashes/tick in half.
Master without changes I’ve made:
tps: 96422
slot high of 542

PR #16105, opened 2021-03-25
Remove stake-o-matic

mvines:
stake-o-matic now lives at https://github.com/solana-labs/stake-o-matic


mvines:
Flaky CI red seems unrelated

Issue #16041, opened 2021-03-19
CI: test_cluster_partition_1_1_1 sometimes hangs


CriesofCarrots:
#### Problem
Occasionally, `test_cluster_partition_1_1_1` hangs, printing a bunch of`No next leader found` messages
eg: https://buildkite.com/solana-labs/solana/builds/42632#1df2179e-36fd-4add-bed5-0df1b61c4b08
This causes CI to eventually time out.

#### Proposed Solution
Triage and fix. At least make `test_cluster_partition_1_1_1` fail fast.

PR #16123, opened 2021-03-25
Add timeout for local cluster partition tests

jstarry:
#### Problem
Local cluster partition test times out sometimes and it's unclear why

#### Summary of Changes
- Add 1min timeout for partition test root checking phase
- Speed up local cluster setup by parallelizing verification
- Enable some log filters for more transparency into local cluster tests
- Un-ignore `test_cluster_partition_1_1` test

Fixes https://github.com/solana-labs/solana/issues/16041

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4710,6 +4710,7 @@ dependencies = [
  "itertools",
  "log 0.4.11",
  "rand 0.7.3",
+ "rayon",
  "serial_test",
  "solana-client",
  "solana-config-program",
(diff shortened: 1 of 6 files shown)


t-nelson (approved):
lgtm!  Anything holding this up?

I think I broken the bpf instruction counts with the `ShortU16` refactor.  I hit it locally, but thought it was related to some another issue that's been floating around Discord for the last couple days

CriesofCarrots (approved):
lgtm.
Some of those replay_stage warnings are pretty noisy. We might want to quiet those down in the future, but let's see how it goes.

jstarry:
Hoping that the cause of the local cluster flakiness is revealed with the extra logging, we can quiet it back down later if it gets too bad

PR #16128, opened 2021-03-25
makes test_pull_request_time_pruning smaller

behzadnouri:
#### Problem
`test_pull_request_time_pruning` seems flaky on ci.
Locally the test always passes, but probably on buildkite it takes longer time and some of the crds values hit their timeout, and so the first assert fails.

#### Summary of Changes
Made the test smaller.


sakridge (dismissed):
great

PR #16139, opened 2021-03-25
Throttle PoH ticks by cumulative slot time

jeffwashington:
Problem
poh can be reset or ticked by PohRecorder. The code in poh_service could be using inaccurate information for throttling.
Also, when a tick is late or early, we currently ignore that when throttling subsequent ticks.

Summary of Changes
poh_service and poh recorder use information in poh directly. poh is the source of truth.
This infrastructure allows us to calculate expected tick time based on the slot time, as well as consistently sleep from ticks that occur in poh recorder as a result of a record.

Fixes #

### DIFF
--- a/core/src/poh_recorder.rs
+++ b/core/src/poh_recorder.rs
@@ -10,6 +10,7 @@
 //! For Entries:
 //! * recorded entry must be >= WorkingBank::min_tick_height && entry must be < WorkingBank::max_tick_height
 //!
+use crate::poh_service::PohService;
 use solana_ledger::blockstore::Blockstore;
 use solana_ledger::entry::Entry;
 use solana_ledger::leader_schedule_cache::LeaderScheduleCache;
@@ -158,6 +159,7 @@ pub struct PohRecorder {
     send_us: u64,
     tick_lock_contention_us: u64,
     tick_overhead_us: u64,
+    total_sleep_us: u64,
     record_us: u64,
     ticks_from_record: u64,
     last_metric: Instant,
@@ -354,7 +356,7 @@ impl PohRecorder {
         let mut cache = vec![];
         let poh_hash = {
             let mut poh = self.poh.lock().unwrap();
-            poh.reset(blockhash, self.poh_config.hashes_per_tick);
+            poh.reset_slot(blockhash, self.poh_config.hashes_per_tick); // TODO: ticks per slot in the bank? can this change?
             poh.hash
         };
         info!(
@@ -462,7 +464,21 @@ impl PohRecorder {
 
     pub fn tick(&mut self) {
         let now = Instant::now();
-        let poh_entry = self.poh.lock().unwrap().tick();
+        let (poh_entry, target_time) = {
+            let mut poh_l = self.poh.lock().unwrap();
+            let poh_entry = poh_l.tick();
+            let target_time = if poh_entry.is_some() {
+                let target_ns_per_tick = PohService::target_ns_per_tick(
+                    self.ticks_per_slot(),
+                    self.poh_config.target_tick_duration.as_nanos() as u64,
+                );
+                Some(poh_l.target_poh_time(target_ns_per_tick))
+            } else {
+                None
+            };
+
+            (poh_entry, target_time)
+        };
         self.tick_lock_contention_us += timing::duration_as_us(&now.elapsed());
         let now = Instant::now();
         if let Some(poh_entry) = poh_entry {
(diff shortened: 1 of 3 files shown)


jeffwashington:
slot time vs slot
Screen Shot 2021-03-25 at 8 56 34 PM

jeffwashington:
Highest TPS: 131103.50 sampling period 1s max transactions: 9291696 clients: 1 drop rate: 0.52
[2021-03-26T02:19:37.213273892Z INFO  solana_bench_tps::bench] 	Average TPS: 102836.37
[2021-03-26T02:19:37.213097201Z INFO  solana_metrics::metrics] datapoint: bench-tps-lamport_balance balance=1000095000i

jeffwashington:
A place I built to go next is to potentially throttle after each batch of hashes where instead of busy waiting we can check for record requests over and over. This reduces locking, spreads out the waits, causes us to more often be available when a record request arrives, and greatly reduces the time we spend busy waiting and deaf to events in the system.

jeffwashington:
Another improvement is reworking the poh_service pipeline where we continue hashing while the long parts of record and tick are executing and we interleave in the has associated with the record and tick with a continuously running poh hasher. This would allow us to increase the current 50% threshold higher.

core/src/poh_recorder.rs
```diff
@@ -462,7 +464,21 @@ impl PohRecorder {
 
     pub fn tick(&mut self) {
         let now = Instant::now();
-        let poh_entry = self.poh.lock().unwrap().tick();
+        let (poh_entry, target_time) = {
+            let mut poh_l = self.poh.lock().unwrap();
+            let poh_entry = poh_l.tick();
+            let target_time = if poh_entry.is_some() {
+                let target_ns_per_tick = PohService::target_ns_per_tick(
+                    self.ticks_per_slot(),
```
carllin, line 472: `ticks_per_slot`, shouldn't ever change in the middle of a run, so might make sense to extract `target_ns_per_tick` into a field in `PohRecorder` as well

core/src/poh_recorder.rs
```diff
@@ -500,6 +525,7 @@ impl PohRecorder {
                 ("prepare_send_us", self.prepare_send_us, i64),
                 ("send_us", self.send_us, i64),
                 ("ticks_from_record", self.ticks_from_record, i64),
+                ("total_sleep_us", self.total_sleep_us, i64),
```
carllin, line 528: heh, unrelated to this PR,  but it seems like `report_metrics()` is only reported on `record()`, which means we won't get data from non leaders? Do we want to report on `tick()` as well?

ledger/src/poh.rs
```diff
@@ -18,23 +21,50 @@ pub struct PohEntry {
...
         }
     }
 
-    pub fn reset(&mut self, hash: Hash, hashes_per_tick: Option<u64>) {
-        let mut poh = Poh::new(hash, hashes_per_tick);
+    pub fn reset_no_slot(&mut self, hash: Hash, hashes_per_tick: Option<u64>) {
+        let mut poh =
+            Poh::new_with_slot_info(hash, hashes_per_tick, self.ticks_per_slot, self.tick_number);
+        std::mem::swap(&mut poh, self);
+    }
```
carllin, line 51: seems like this function is unused

core/src/poh_service.rs
```diff
@@ -276,13 +285,12 @@ impl PohService {
     fn tick_producer(
         poh_recorder: Arc<Mutex<PohRecorder>>,
         poh_exit: &AtomicBool,
-        target_tick_ns: u64,
+        _target_tick_ns: u64,
```
carllin, line 288: nit: can we remove this?

core/src/poh_recorder.rs
```diff
@@ -354,7 +356,7 @@ impl PohRecorder {
         let mut cache = vec![];
         let poh_hash = {
             let mut poh = self.poh.lock().unwrap();
-            poh.reset(blockhash, self.poh_config.hashes_per_tick);
+            poh.reset_slot(blockhash, self.poh_config.hashes_per_tick); // TODO: ticks per slot in the bank? can this change?
```
carllin, line 359: ticks per slot should not change without restarting the validator.

ledger/src/poh.rs
jeffwashington: we need tick_number for target_poh_time

core/src/poh_recorder.rs
jeffwashington: so in tick(), will bank_clone.slot() give me the accurate slot after a tick? slot is an input to report.

carllin: ah never mind, I think I see what it's trying to do. You're right, the purpose is to aggregate all the ticks, including virtual ticks into the metrics for the next slot. The non-specific slot version is done in `report_service::report`.

core/src/poh_service.rs
```diff
@@ -199,6 +202,7 @@ impl PohService {
         }
     }
 
+    // returns target time for tick if we need to tick, or None if we don't need to tick
```
sakridge, line 205: hmm, does it return through an argument? Doesn't this just return `bool`?

jeffwashington: yes, comment was stale

sakridge (approved):
lgtm

PR #16142, opened 2021-03-25
Rpc: enable getConfirmedBlock and getConfirmedTransaction to return confirmed (not yet finalized) data

CriesofCarrots:
#### Problem
Despite their names, `getConfirmedBlock` and `getConfirmedTransaction` only return finalized data, forcing users to wait 30+ slots longer than necessary to get tx or block details.

#### Summary of Changes
- Add Blockstore apis to return data for complete blocks and transactions, dropping the rooted requirement
- Because the TransactionStatusService is async, add a method to track the highest slot with completely written statuses to ensure confirmed blocks are complete (will also need in `getConfirmedSignaturesForAddress2`)
- Expose confirmed data in `getConfirmedBlock` and `getConfirmedTransaction` via the commitment parameter; default remains `finalized`
- Return confirmed data in `solana block <slot>` and `solana confirm <signature> -v`

**Follow-up work:**
- Support confirmed data in `getConfirmedSignaturesForAddress2`; handling in separate PR because blockstore api is so convoluted
- Consider supporting confirmed data in `getConfirmedBlocks` and `getConfirmedBlocksWithLimit`; the former will need some rework a la https://github.com/solana-labs/solana/pull/15827 to retain backward compatibility while adding a config parameter
- For v1.7, add `getX` methods and deprecate `getConfirmedX` methods

### DIFF
--- a/core/src/rpc.rs
+++ b/core/src/rpc.rs
@@ -928,7 +928,7 @@ impl JsonRpcRequestProcessor {
                 Some(status)
             } else if self.config.enable_rpc_transaction_history && search_transaction_history {
                 self.blockstore
-                    .get_transaction_status(signature)
+                    .get_transaction_status(signature, true)
                     .map_err(|_| Error::internal_error())?
                     .filter(|(slot, _status_meta)| {
                         slot <= &self
--- a/core/src/rpc.rs
+++ b/core/src/rpc.rs
@@ -84,7 +84,7 @@ use std::{
     net::SocketAddr,
     str::FromStr,
     sync::{
-        atomic::{AtomicBool, Ordering},
+        atomic::{AtomicBool, AtomicU64, Ordering},
         mpsc::{channel, Receiver, Sender},
         Arc, Mutex, RwLock,
     },
@@ -144,6 +144,7 @@ pub struct JsonRpcRequestProcessor {
     largest_accounts_cache: Arc<RwLock<LargestAccountsCache>>,
     max_slots: Arc<MaxSlots>,
     leader_schedule_cache: Arc<LeaderScheduleCache>,
+    max_complete_transaction_status_slot: Arc<AtomicU64>,
 }
 impl Metadata for JsonRpcRequestProcessor {}
 
@@ -229,6 +230,7 @@ impl JsonRpcRequestProcessor {
         largest_accounts_cache: Arc<RwLock<LargestAccountsCache>>,
         max_slots: Arc<MaxSlots>,
         leader_schedule_cache: Arc<LeaderScheduleCache>,
+        max_complete_transaction_status_slot: Arc<AtomicU64>,
     ) -> (Self, Receiver<TransactionInfo>) {
         let (sender, receiver) = channel();
         (
@@ -249,6 +251,7 @@ impl JsonRpcRequestProcessor {
                 largest_accounts_cache,
                 max_slots,
                 leader_schedule_cache,
+                max_complete_transaction_status_slot,
             },
             receiver,
         )
@@ -291,6 +294,7 @@ impl JsonRpcRequestProcessor {
             largest_accounts_cache: Arc::new(RwLock::new(LargestAccountsCache::new(30))),
             max_slots: Arc::new(MaxSlots::default()),
             leader_schedule_cache: Arc::new(LeaderScheduleCache::new_from_bank(bank)),
+            max_complete_transaction_status_slot: Arc::new(AtomicU64::default()),
         }
     }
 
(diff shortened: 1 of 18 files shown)


CriesofCarrots:
@carllin could you please review the first 3 commits, plus the logic identified in comments?

core/src/rpc.rs
```diff
@@ -729,40 +733,58 @@ impl JsonRpcRequestProcessor {
...
+                    }
+                }
+                self.check_slot_cleaned_up(&result, slot)?;
+                return Ok(result.ok().map(|confirmed_block| {
+                    confirmed_block.configure(encoding, transaction_details, show_rewards)
+                }));
+            } else if commitment.is_confirmed() {
+                // Check if block is confirmed
+                let confirmed_bank = self.bank(Some(CommitmentConfig::confirmed()));
+                if confirmed_bank.status_cache_ancestors().contains(&slot)
```
CriesofCarrots, line 774: @carllin is this sufficient to determine if a block between the optimistically_confirmed_bank and the highest_confirmed_root is confirmed?

mvines (dismissed):

core/src/rpc.rs
```diff
@@ -733,41 +733,58 @@ impl JsonRpcRequestProcessor {
...
-            self.check_blockstore_root(&result, slot)?;
-            if result.is_err() {
-                if let Some(bigtable_ledger_storage) = &self.bigtable_ledger_storage {
-                    let bigtable_result = self
-                        .runtime
-                        .block_on(bigtable_ledger_storage.get_confirmed_block(slot));
-                    self.check_bigtable_result(&bigtable_result)?;
-                    return Ok(bigtable_result.ok().map(|confirmed_block| {
+            {
+                let result = self.blockstore.get_confirmed_block(slot, true);
```
carllin, line 754: heh, `confirmed` is overloaded here, this should really be named `get_rooted_block` right?

core/src/rpc.rs
```diff
@@ -1574,6 +1591,15 @@ fn verify_token_account_filter(
     }
 }
 
+fn check_confirmed_commitment(commitment: CommitmentConfig) -> Result<()> {
```
carllin, line 1605: nit: How about a more explicit name like `check_is_at_least_confirmed_commitment()`?

core/src/rpc.rs
```diff
@@ -1574,6 +1591,15 @@ fn verify_token_account_filter(
...
+    if commitment.is_processed() {
```
carllin, line 1606: I think here it might be worth it here to explicitly match on all the possible commitment levels, in case we add a weaker confirmation level in the future, the compiler will catch it.

core/src/rpc.rs
```diff
@@ -733,41 +733,58 @@ impl JsonRpcRequestProcessor {
...
+                    confirmed_block.configure(encoding, transaction_details, show_rewards)
+                }));
+            } else if commitment.is_confirmed() {
+                // Check if block is confirmed
+                let confirmed_bank = self.bank(Some(CommitmentConfig::confirmed()));
+                if confirmed_bank.status_cache_ancestors().contains(&slot)
+                    && slot
+                        <= self
+                            .max_complete_transaction_status_slot
+                            .load(Ordering::Relaxed)
```
carllin, line 778: Here and in `transaction_status_service`, I would use `Ordering::SeqCst` to avoid any weird reorderings.

core/src/rpc.rs
carllin: yeah I think this should work! If it doesn't you can blame it on me for poorly thought out ideas 😄

core/src/rpc.rs
CriesofCarrots: Nice idea! Will do

core/src/rpc.rs
CriesofCarrots: Hehe, yes, exactly. Now that I'm reclaiming `confirmed`, there's no reason to keep this method name. Will rename, thanks!

core/src/rpc.rs
CriesofCarrots: I mocked this up in `sdk/src/commitment_config.rs`, but I thought it made the code a whole lot less readable. However, I think the implementation now is a bit better, as new commitment levels will be default be `!is_at_least_confirmed` and have to be explicitly added to `is_confirmed` or `is_finalized`

core/src/rpc.rs
CriesofCarrots: There is a whole rabbit hole to go down here with the confirmed/Confirmed prefix. I contemplated updating the name of the bigtable method as well, but I think I'll save auditing all of that for v1.7-only work.

PR #16150, opened 2021-03-26
read only account cache for executable accounts - improve replay

jeffwashington:
Problem
During replay, we load things like programs from accounts, sometimes many (375) times during 1 slot. If we don't load them from the cache, then we load them from a store on each load. Loading from a store always requires copying data from the store to new memory. An example data size on programs is 500k.

Summary of Changes
Loaded accounts are kept in the read only cache for potential future loads.

We brainstormed on this a few weeks ago and decided to keep the read only cache separate from the existing cache.

Fixes #

### DIFF
--- a/runtime/src/accounts.rs
+++ b/runtime/src/accounts.rs
@@ -215,7 +215,7 @@ impl Accounts {
                     } else {
                         let (account, rent) = self
                             .accounts_db
-                            .load(ancestors, key) // replace this, too
+                            .load(ancestors, key)
                             .map(|(mut account, _)| {
                                 if message.is_writable(i, demote_sysvar_write_locks) {
                                     let rent_due = rent_collector
@@ -342,7 +342,7 @@ impl Accounts {
 
             let program = match self
                 .accounts_db
-                .load(ancestors, &program_id) // read only cache
+                .load(ancestors, &program_id)
                 .map(|(account, _)| account)
             {
                 Some(program) => program,
--- a/runtime/src/accounts.rs
+++ b/runtime/src/accounts.rs
@@ -368,7 +368,6 @@ impl Accounts {
                     if let Some(program) = self
                         .accounts_db
                         .load_and_keep_in_read_only_cache(ancestors, &programdata_address)
-                        .map(|account| account)
                     {
                         accounts.insert(0, (programdata_address, program));
                     } else {
--- a/runtime/src/accounts.rs
+++ b/runtime/src/accounts.rs
@@ -215,7 +215,7 @@ impl Accounts {
                     } else {
                         let (account, rent) = self
                             .accounts_db
-                            .load(ancestors, key)
+                            .load(ancestors, key) // replace this, too
                             .map(|(mut account, _)| {
                                 if message.is_writable(i, demote_sysvar_write_locks) {
                                     let rent_due = rent_collector
@@ -342,7 +342,7 @@ impl Accounts {
 
             let program = match self
                 .accounts_db
-                .load(ancestors, &program_id)
+                .load(ancestors, &program_id) // read only cache
                 .map(|(account, _)| account)
             {
                 Some(program) => program,
@@ -367,7 +367,8 @@ impl Accounts {
                 {
                     if let Some(program) = self
                         .accounts_db
-                        .load_and_keep_in_read_only_cache(ancestors, &programdata_address)
+                        .load(ancestors, &programdata_address)
+                        .map(|(account, _)| account)
                     {
                         accounts.insert(0, (programdata_address, program));
                     } else {
--- a/runtime/src/accounts.rs
+++ b/runtime/src/accounts.rs
@@ -367,8 +367,8 @@ impl Accounts {
                 {
                     if let Some(program) = self
                         .accounts_db
-                        .load(ancestors, &programdata_address)
-                        .map(|(account, _)| account)
+                        .load_and_keep_in_read_only_cache(ancestors, &programdata_address)
+                        .map(|account| account)
                     {
                         accounts.insert(0, (programdata_address, program));
                     } else {
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -8595,14 +8595,6 @@ pub mod tests {
             .unwrap();
         assert_eq!(account.lamports, 0);
         assert_eq!(db.read_only_accounts_cache.cache_len(), 1);
-
-        // Flush, then clean again. Should not need another root to initiate the cleaning
-        // because `accounts_index.uncleaned_roots` should be correct
-        db.flush_accounts_cache(true, None);
-        db.clean_accounts(None);
-        assert!(db
-            .do_load(&Ancestors::default(), &account_key, Some(0))
-            .is_none());
     }
 
     #[test]
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -289,7 +289,7 @@ impl<'a> LoadedAccount<'a> {
         }
     }
 
-    pub fn from_cache(&self) -> bool {
+    pub fn is_cached(&self) -> bool {
         match self {
             LoadedAccount::Stored(_) => false,
             LoadedAccount::Cached(_) => true,
@@ -2288,18 +2288,30 @@ impl AccountsDb {
         }
 
         //TODO: thread this as a ref
-        let mut from_cache = false;
+        let mut is_cached = false;
         let loaded_account = self
             .get_account_accessor_from_cache_or_storage(slot, pubkey, store_id, offset)
             .get_loaded_account()
             .map(|loaded_account| {
-                from_cache = loaded_account.from_cache();
+                is_cached = loaded_account.is_cached();
                 (loaded_account.account(), slot)
             });
 
-        if self.caching_enabled && !from_cache {
+        if self.caching_enabled && !is_cached {
             match loaded_account {
                 Some((account, slot)) => {
+                    /*
+                    We show this store into the read-only cache for account 'A' and future loads of 'A' from the read-only cache are
+                    safe/reflect 'A''s latest state on this fork.
+                    This safety holds if during replay of slot 'S', we show we only read 'A' from the write cache,
+                    not the read-only cache, after it's been updated in replay of slot 'S'.
+                    Assume for contradiction this is not true, and we read 'A' from the read-only cache *after* it had been updated in 'S'.
+                    This means an entry '(S, A)' was added to the read-only cache after 'A' had been updated in 'S'.
+                    Now when '(S, A)' was being added to the read-only cache, it must have been true that  'is_cache == false',
+                    which means '(S', A)' does not exist in the write cache yet.
+                    However, by the assumption for contradiction above ,  'A' has already been updated in 'S' which means '(S, A)'
+                    must exist in the write cache, which is a contradiction.
+                    */
                     self.read_only_accounts_cache.store(pubkey, slot, &account);
                     Some((account, slot))
                 }
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -4364,6 +4364,16 @@ impl AccountsDb {
                     self.stats.store_total_data.swap(0, Ordering::Relaxed),
                     i64
                 ),
+                (
+                    "read_only_accounts_cache_entries",
+                    self.read_only_accounts_cache.cache_len(),
+                    i64
+                ),
+                (
+                    "read_only_accounts_cache_data_size",
+                    self.read_only_accounts_cache.data_size(),
+                    i64
+                ),
             );
 
             let recycle_stores = self.recycle_stores.read().unwrap();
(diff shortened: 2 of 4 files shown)


jeffwashington:
initial data on ledger-tool verify.
difference is on load_us. We are now loading executable accounts only once.
```
prior to arc data change
ledger processing timing ExecuteTimings load_us 23338539 execute_us 53565352 store_us 24995196 details ExecuteDetailsTimings serialize_us 6824620 create_vm_us 1540312 execute_us 7245592 deserialize_us 5997551 changed_account_count 57938 total_account_count 1113864 total_data_size 48123505204 data_size_changed 49975074
ledger processed in 45 seconds. 5281 MB allocated. root slot is 69585630 1 fork at 69585662 with 32 frozen banks

master:
ledger processing timing ExecuteTimings load_us 9461061 execute_us 54405801 store_us 21695541 details ExecuteDetailsTimings serialize_us 9229740 create_vm_us 1680809 execute_us 7459375 deserialize_us 13008733 changed_account_count 57938 total_account_count 1113864 total_data_size 48123505204 data_size_changed 49975074
ledger processed in 41 seconds. 5094 MB allocated. root slot is 69585630 1 fork at 69585662 with 32 frozen banks

read only account cache:
ledger processing timing ExecuteTimings load_us 6489180 execute_us 54815610 store_us 21725037 details ExecuteDetailsTimings serialize_us 9724446 create_vm_us 1683096 execute_us 7427180 deserialize_us 12862927 changed_account_count 57938 total_account_count 1113864 total_data_size 48123505204 data_size_changed 49975074
ledger processed in 40 seconds. 5080 MB allocated. root slot is 69585630 1 fork at 69585662 with 32 frozen banks

```

jeffwashington:
Some old notes on the thinking prior to the decision to keep this cache separate.
Note that we need to work out details of this. When do we flush these accounts from the cache? Can we avoid writing them out again when we flush the cache? Should we architect this some other way? Should all account loads cause the loaded item to remain in the cache? If we move to an account being stored in the cache as an arc/cow then storing a loaded account in the cache isn't holding any additional resources as long as the account is kept in memory by the caller anyway. Clearly, it will be more memory than we use if we don't keep anything in the cache, but there are mitigating factors and tradeoffs to be made. I implemented this so far with the caller being explicit suggesting that there are likely to be more requests for this account in the future. We have to understand whether how reliable this information is.

jeffwashington:
Original prototype for discusssion: #15633

jackcmay:
Is there any reason to flush the executable account's cache since there accounts are immutable?

Also, upgradeable program's won't benefit from this as much since the executable account is very small since it just links to the program data account which is not immutable and holds the actual program data.  Ideally we could cache those program data accounts and flush them if they changed

jeffwashington:
> Is there any reason to flush the executable account's cache since there accounts are immutable?
> 
> Also, upgradeable program's won't benefit from this as much since the executable account is very small since it just links to the program data account which is not immutable and holds the actual program data. Ideally we could cache those program data accounts and flush them if they changed

I am in search of the definitive answer on the lifetime of these accounts and also ignorant ;-) For any cache, I imagine we need to flush them due to incorrect, never used anymore, space, etc. This pr is in the stage of early feedback on the basic structure. I am happy to hear however we can constrain the requirements from the experts!

jackcmay:
Executable accounts are rent-exempt and immutable, so forever lifetime.  You could constrain the cache via LRU rules.

sakridge:
> Is there any reason to flush the executable account's cache since there accounts are immutable?
> 
> Also, upgradeable program's won't benefit from this as much since the executable account is very small since it just links to the program data account which is not immutable and holds the actual program data. Ideally we could cache those program data accounts and flush them if they changed

It's not about immutable or not, we care about how the access is i.e. if it's read-only or write-able. The program data accounts are write-able but we don't expect transactions to be changing them often. They would likely be read many more times than written, that's what we're optimizing for. In that case we can help them by putting them in this read-only cache and then flush them if we encounter a write operation.

The existing accounts cache is like a write-cache in that it needs to be at least flushed for snapshots, and then also flushed just to limit the number of accounts in-memory and the only way we have is flushing a slot or list of slots. In a perfect world both of those things wouldn't necessarily be blockers because you could stream the snapshot contents from some in-memory state and you could flush based on some access heuristics and the clean/shrink algorithms would have to deal with a mixed on-disk/in-memory state, but that would be a bigger long-term project.

jeffwashington:
with a few weeks old mainnet snapshot, ledger-tool verify loaded these accounts into the read-only cache:
```D9Kh6CzoNb2gj7EDsP7fNZpbhLq8r9LoLgambqPKoF4z size 1223863
Em7Qs81QaZLTDVCj39h3nSunH3wuPbnR9DYJyA3r482X size 759629
9yCkCD7VS2ZTYN3nZEof73ATj6jmPWbgB3LWcdqAtw1 size 758861
42V8QNnkPfMhsBxyNK4txzLiqnJYSi7eUbSp5gdz7KnJ size 780013
68ZYga21EEhp6wcWfCfxxdTfhyYPwKio3SLwf68uhhR size 276877
FV9WWFWYzWHDpAc8oxiFKV1LmFUyWw7HGdo4GK19MiZW size 619629
Hmoj1nTYsPirbmS2jeEQJVFmKJ67a2eVgsLgJ13SNNVB size 3161469
DTxcpNApaMLNfYgwQ99PmpCm8rjS7o1q2YdfzYsrYohB size 494866
```

jeffwashington:
@sakridge, I can imagine settable metrics for the read only account cache limits on the validator for metrics like:
1. max read only cache size (ie. remove oldest or largest items until read only cache is under x Mb)
2. max read only cache item size (ie. only cache items < 4M of data)

What default do you think is good? The above accounts are ~8M of data. You had mentioned that this cache is a tradeoff in size with the write cache and other users of memory. Seems like a max size of 50M is big enough for now and still very small. We could remove items that hadn't been used in 10s? Keeping it simple, we would just set a limit of 50M and throw items out when a new addition would make the size > 50M. We would through away LRU, presumably. Thoughts? Do we need a validator flag now or do we add it later?

jeffwashington:
lru is implemented

runtime/src/accounts.rs
```diff
@@ -362,8 +362,7 @@ impl Accounts {
                 {
                     if let Some(program) = self
                         .accounts_db
-                        .load(ancestors, &programdata_address)
-                        .map(|(account, _)| account)
+                        .load_and_keep_in_read_only_cache(ancestors, &programdata_address)
```
jeffwashington, line 365: @sakridge, is it possible for the read only cache to have an incorrect version of its contents due to an account coming from a different ancestors?

sakridge: Yes. The cache would need to take the ancestors into account.

jeffwashington:
I did not yet optimize the lru for large numbers of accounts. I was thinking small # of accounts. it is possible we waste some decent amount of time calculating lru.

runtime/src/accounts.rs
jeffwashington: it does now

jeffwashington:
metrics with all loads using the cache and a 300M cache, this pr:
ledger processing timing ExecuteTimings **load_us 6517599** execute_us 45003879 store_us 20475386 details ExecuteDetailsTimings serialize_us 8872259 create_vm_us 1469312 execute_us 7161474 deserialize_us 5165972 changed_account_count 57938 total_account_count 1113864 total_data_size 48123505204 data_size_changed 49975074
ledger processed in **35 seconds**. 4571 MB allocated. root slot is 69585630 1 fork at 69585662 with 32 frozen banks

master:
ledger processing timing ExecuteTimings **load_us 10513495** execute_us 54101092 store_us 21687964 details ExecuteDetailsTimings serialize_us 9334806 create_vm_us 1655802 execute_us 7375114 deserialize_us 12843527 changed_account_count 57938 total_account_count 1113864 total_data_size 48123505204 data_size_changed 49975074
ledger processed in **41 seconds**. 5089 MB allocated. root slot is 69585630 1 fork at 69585662 with 32 frozen banks

jeffwashington:
2s worth of metrics:
[2021-03-31T03:33:44.247326327Z INFO  solana_metrics::metrics] datapoint: accounts_db_store_timings hash_accounts=438519i store_accounts=24217i update_index=18328i handle_reclaims=4i append_accounts=0i find_storage=0i num_accounts=19993i total_data=1021116473i read_only_accounts_cache_entries=8392i read_only_accounts_cache_data_size=298875197i read_only_accounts_cache_hits=11840i read_only_accounts_cache_misses=39131i
[2021-03-31T03:33:45.248042478Z INFO  solana_metrics::metrics] datapoint: accounts_db_store_timings hash_accounts=421650i store_accounts=29014i update_index=21963i handle_reclaims=5i append_accounts=0i find_storage=0i num_accounts=24825i total_data=1023660516i read_only_accounts_cache_entries=8487i read_only_accounts_cache_data_size=299693032i read_only_accounts_cache_hits=12958i read_only_accounts_cache_misses=47172i

runtime/src/accounts_db.rs
```diff
@@ -285,6 +286,13 @@ impl<'a> LoadedAccount<'a> {
             },
         }
     }
+
+    pub fn from_cache(&self) -> bool {
```
carllin, line 290: nit: `from_` prefixes in Rust generally imply the argument is consumed and transformed into another type. How about `is_cached()`?

runtime/src/accounts_db.rs
```diff
@@ -4215,6 +4250,8 @@ impl AccountsDb {
                 Ordering::Relaxed,
             ) == Ok(last)
         {
+            let (read_only_cache_hits, read_only_cache_misses) =
```
carllin, line 4253: nice! always good to have observable metrics 😄

runtime/src/accounts_db.rs
```diff
@@ -2234,10 +2246,32 @@ impl AccountsDb {
             // `lock` released here
         };
 
+        let result = self.read_only_accounts_cache.load(pubkey, slot);
```
carllin, line 2249: Note that there's a known bug here:  that is currently reproducing very rarely, but may occur more often now that we've added extra logic between the `self.accounts_index.get()` and the load from cache/store: https://github.com/solana-labs/solana/pull/15454

runtime/src/accounts_db.rs
```diff
@@ -2234,10 +2246,32 @@ impl AccountsDb {
...
+        let loaded_account = self
+            .get_account_accessor_from_cache_or_storage(slot, pubkey, store_id, offset)
             .get_loaded_account()
-            .map(|loaded_account| (loaded_account.account(), slot))
+            .map(|loaded_account| {
+                from_cache = loaded_account.from_cache();
+                (loaded_account.account(), slot)
+            });
+
+        if self.caching_enabled && !from_cache {
```
carllin, line 2264: How about a comment about why the `!from_cache` is sufficient to make sure we aren't loading stale data from the `read-only` cache if there is more up-to-date state in the write cache. Something like:

`We show this store into the read-only cache for account `A` and future loads of `A` from the read-only cache are safe/reflect `A`'s latest state on this fork. 

This safety holds if during replay of slot `S`, we show we only read `A` from the write cache after it's been updated in slot `S`. 

Assume for contradiction this is not true, and we read `A` from the read-only cache *after* it had been updated in `S`. This means an entry `(S, A)` was added to the read-only cache after `A` had been updated in `S`.
 
Now when `(S, A)` was being added to the read-only cache, it must have been true that  `from_cache == false`, which means `(S', A)` does not exist in the write cache yet.  

However, by the assumption for contradiction above ,  `A` has already been updated in `S` which means `(S, A)` must exist in the write cache, which is a contradiction.`

runtime/src/accounts_db.rs
ryoqun: (nice way to ping me... lol, itoken and spl-token has been stealing my time... I try to fix #15454  tonight...)

jeffwashington:
I am working on a version that pre-loads in a bg thread all accounts in transactions we're about to replay. End results:
datapoint: accounts_db_store_timings <snip> read_only_accounts_cache_entries=7050i read_only_accounts_cache_data_size=298999087i **read_only_accounts_cache_hits=11263i read_only_accounts_cache_misses=9i**

load time is also down ~0.8s
ledger processing timing ExecuteTimings **load_us 5766218** execute_us 44693728 store_us 20233914 details ExecuteDetailsTimings serialize_us 8969690 create_vm_us 1459452 execute_us 7054692 deserialize_us 5124414 changed_account_count 57938 total_account_count 1113864 total_data_size 48123505204 data_size_changed 49975074
ledger processed in 35 seconds. 4632 MB allocated. root slot is 69585630 1 fork at 69585662 with 32 frozen banks

sakridge (approved):
lgtm

runtime/src/accounts_db.rs
carllin: > (nice way to ping me... lol, itoken and spl-token has been stealing my time... I try to fix #15454 tonight...)

haha thanks! I think it's really close, you and Jeff might have to race for the conflicts 👊

PR #16151, opened 2021-03-26
Explorer: account for custom endpoint possible conditions in stakehistory

oJshua:
#### Problem
I believe https://sentry.io/organizations/solana/issues/2299223711/?project=5390542&referrer=slack is happening where local clusters don't have stake history. 

#### Summary of Changes
The alternative to this fix would be to hide the stake column altogether, but for now we will show an error box.


jstarry:
Nice, was just about to PR this, thanks!

PR #16153, opened 2021-03-26
removes OrderedIterator and transaction batch iteration order

behzadnouri:
#### Problem
In `TransactionBatch`,
https://github.com/solana-labs/solana/blob/e50f59844/runtime/src/transaction_batch.rs#L4-L11
`lock_results[i]` is aligned with `transactions[iteration_order[i]]`:
https://github.com/solana-labs/solana/blob/e50f59844/runtime/src/bank.rs#L2414-L2424
https://github.com/solana-labs/solana/blob/e50f59844/runtime/src/accounts.rs#L788-L817

However `load_and_execute_transactions` is iterating over 
```
  lock_results[iteration_order[i]]
```
https://github.com/solana-labs/solana/blob/e50f59844/runtime/src/bank.rs#L2878-L2889
and then returning `i` as for the index of the retryable transaction.

If `iteratorion_order` is `[1, 2, 0]`, and `i` is `0`, then:
```
  lock_results[iteration_order[i]] = lock_results[1]
```
which corresponds to
```
  transactions[iteration_order[1]] = transactions[2]
```
so neither `i = 0`, nor `iteration_order[i] = 1` gives the correct index for the
corresponding transaction (which is 2).

#### Summary of Changes
This commit removes `OrderedIterator` and transaction batch iteration order
entirely. There is only one place in blockstore processor which the
iteration order is not ordinal:
https://github.com/solana-labs/solana/blob/e50f59844/ledger/src/blockstore_processor.rs#L269-L271
It seems like, instead of using an iteration order, that can shuffle a cloned
vector of transactions.


sakridge:
Perf impact?

sakridge:
The re-ordering has been kind of a headache, but this introduces some new copying of transactions.

behzadnouri:
> The re-ordering has been kind of a headache, but this introduces some new copying of transactions.

On the other hand, this code:
* has better cache locality.
* removes the extra if branches on each call to `OrderedIterator::next`:
  https://github.com/solana-labs/solana/blob/60ed8e289/runtime/src/transaction_utils.rs#L23-L39

I do not see a significant impact on the path which hits clone.

`replay-loop-timing-stats::replay_active_banks_elapsed`:
replay-loop-timing-stats replay_active_banks_elapsed

sakridge:
> > The re-ordering has been kind of a headache, but this introduces some new copying of transactions.
> 
> On the other hand, this code:
> 
> * has better cache locality.
> * removes the extra if branches on each call to `OrderedIterator::next`:
>   https://github.com/solana-labs/solana/blob/60ed8e289/runtime/src/transaction_utils.rs#L23-L39
> 
> I do not see a significant impact on the path which hits clone.
> 

What is the setup of the graph? I think `replay-slot-stats` would be a better one to look at. And testing the bench-tps run for both that and overall TPS.

behzadnouri:
> What is the setup of the graph? I think replay-slot-stats.replay_time and/or replay-slot-stats.execute_us would be a better one to look at. And testing the bench-tps run for both that and overall TPS.

The graphs are from a node on testnet. I also tested on gce and the results are the same.

replay-slot-stats-tds

sakridge:
> > What is the setup of the graph? I think replay-slot-stats.replay_time and/or replay-slot-stats.execute_us would be a better one to look at. And testing the bench-tps run for both that and overall TPS.
> 
> The graphs are from a node on testnet. I also tested on gce and the results are the same.
> 

Testnet has very few transactions, so this will be a poor test for measuring the performance impact. Mainnet is better, but probably even better is a maximal throughput test where you are testing under full load like bench-tps.

behzadnouri:
Updated the code to mutate entries in-place (so no more cloning).
Nevertheless, bellow benchmarks are with the earlier version of the code which was making copies, and actually shows even better avg tps.

This code:
```
    Highest TPS: 8062.21 sampling period 1s max transactions: 385694 clients: 1 drop rate: 0.92
[2021-03-30T14:57:57.973767916Z INFO  solana_bench_tps::bench]  Average TPS: 4246.783
```
master:
```
    Highest TPS: 8094.10 sampling period 1s max transactions: 357143 clients: 1 drop rate: 0.92
[2021-03-30T15:52:20.992950629Z INFO  solana_bench_tps::bench]  Average TPS: 3909.3313
```

sakridge:
> Updated the code to mutate entries in-place (so no more cloning).
> Nevertheless, bellow benchmarks are with the earlier version of the code which was making copies, and actually shows even better avg tps.
> 
> This code (earlier version which was cloning entries):
> 
> ```
>     Highest TPS: 8062.21 sampling period 1s max transactions: 385694 clients: 1 drop rate: 0.92
> [2021-03-30T14:57:57.973767916Z INFO  solana_bench_tps::bench]  Average TPS: 4246.783
> ```
> 
> master:
> 
> ```
>     Highest TPS: 8094.10 sampling period 1s max transactions: 357143 clients: 1 drop rate: 0.92
> [2021-03-30T15:52:20.992950629Z INFO  solana_bench_tps::bench]  Average TPS: 3909.3313
> ```

ok, I started a gce run as well: https://buildkite.com/solana-labs/system-performance-tests/builds/4279
just to double check, but looks good if that is similar.

sakridge (approved):
lgtm

decodedfile (commented):
Duplicate of #

jstarry:
@behzadnouri any objection to this getting backported to v1.5? It would make it easier for me to backport one of my changes.

jstarry:
I'm going to prep the PR in the meantime

behzadnouri:
> I'm going to prep the PR in the meantime

Thanks for doing the backport. It could have been easier if commits were not squashed so I could see where the merge conflicts were.
I was planning to let this soak on TDS for a bit. It seems like the backport is already merged by automerge. Probably no need to revert, but maybe double check this pr that nothing is left out.

jstarry:
Hey @behzadnouri, sorry I didn't notice that the bot added the label again after I removed it. Thanks for the context about wanting to let it soak, I noticed it's been released since v1.6.4 so it has soaked for awhile already.

As for the conflicts, they were basically all due to the fact that the `demote_sysvar_write_locks` change hasn't been backported yet

PR #16155, opened 2021-03-26
bpf: Add failing test for calling get_packed_len in BPF

joncinque:
#### Problem

Calling `solana_program::get_packed_len()` in a program causes the runtime to error out while loading the ELF. Before running any code, the runtime fails with:

```
thread 'assert_instruction_count' panicked at 'called `Result::unwrap()` on an `Err` value: ElfError(FailedToParse("read-write: bad offset 11139531056"))', tests/programs.rs:208:68
```

#### Summary of Changes

Add a failing test. 

Fixes #

### DIFF
--- a/programs/bpf/Cargo.lock
+++ b/programs/bpf/Cargo.lock
@@ -2788,6 +2788,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "solana-bpf-rust-packed-len"
+version = "1.7.0"
+dependencies = [
+ "borsh",
+ "solana-program 1.7.0",
+]
+
+[[package]]
 name = "solana-bpf-rust-panic"
 version = "1.7.0"
 dependencies = [
--- a/programs/bpf/Cargo.toml
+++ b/programs/bpf/Cargo.toml
@@ -63,6 +63,7 @@ members = [
     "rust/many_args_dep",
     "rust/mem",
     "rust/noop",
+    "rust/packed_len",
     "rust/panic",
     "rust/param_passing",
     "rust/param_passing_dep",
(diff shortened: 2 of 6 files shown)


jackcmay:
@joncinque This is good, I'll take this over as part of the fix and I think I'll bundle this into an existing program instead of creating a new program specifically for this test case

joncinque:
Sounds good, thanks @jackcmay !

jackcmay:
@joncinque I looked into this issue and the problem is that borsh is introducing rw data into the ELF via the HashMap hasher they are using.  They are using the default `RandomState` hasher which keeps a list of keys in in thread-local storage.  As for why it crashes, looks like the goblin elf parse is having some issues with the ELF.  I'll look into that next so that we can more clearly identify when rw data is the issue.

Going forward I don't have a clear solution to enable calling S::schema_container() (which is called by `get_packed_len()`.  I tried flipping borsh over to a different hasher that does not use rw data and the tests pass so forking borsh is one option I guess (or get them to change to a different hasher).

jackcmay:
This PR will let the ELF loader report the real error (that there is a .bss section in the ELF) https://github.com/solana-labs/solana/pull/16256

And this PR will improve the error message developers get:
https://github.com/solana-labs/rbpf/pull/149

joncinque:
@jackcmay Gotcha, that makes sense -- thanks for investigating this! Do you have a branch with those hasher changes? Maybe we can make a case for adding a feature which changes the hasher used by Borsh.  What do you think?

jackcmay:
I do but it's pretty ugly, looks like it needs some public type changes, I'll try to clean it up

joncinque:
I'm happy to help out or take over any bits if you like :smile:

PR #16157, opened 2021-03-26
Hash stored accounts in bg

jeffwashington:
Problem
While processing a slot, the same account can be written to many times.
Hashing accounts is not free and gets expensive with large data. Hashing an account that is not the final update is not usually necessary.

Summary of Changes
When an account is stored to the accounts_cache, don't hash the contents first - instead use None. When a caller requires a hash, and a hash hasn't been calculated, calculate it and store it. A background thread processes items that are added to the accounts_cache, calculates the hash and updates the cached item in accounts_cache.

Fixes #

### DIFF
--- a/runtime/src/accounts_cache.rs
+++ b/runtime/src/accounts_cache.rs
@@ -127,11 +127,9 @@ pub struct CachedAccountInner {
 
 impl CachedAccountInner {
     pub fn hash(&self) -> Hash {
-        error!("locking to read hash");
         let hash = self.hash.read().unwrap();
         match *hash {
             Some(hash) => {
-                error!("done locking to read hash");
                 hash
             },
             None => {
@@ -142,9 +140,7 @@ impl CachedAccountInner {
                     &self.pubkey,
                     &self.cluster_type,
                 );
-                error!("write locking to set hash");
                 *self.hash.write().unwrap() = Some(hash.clone());
-                error!("done write locking to set hash");
                 hash
             }
         }
--- a/runtime/src/accounts_cache.rs
+++ b/runtime/src/accounts_cache.rs
@@ -1,5 +1,4 @@
 use dashmap::DashMap;
-use log::error;
 use solana_sdk::{
     account::{AccountSharedData, ReadableAccount},
     clock::Slot,
@@ -296,7 +295,9 @@ pub mod tests {
             inserted_slot,
             &Pubkey::new_unique(),
             AccountSharedData::new(1, 0, &Pubkey::default()),
-            Hash::default(),
+            Some(Hash::default()),
+            inserted_slot,
+            ClusterType::Development,
         );
         // If the cache is told the size limit is 0, it should return the one slot
         let removed = cache.remove_slots_le(0);
@@ -314,7 +315,9 @@ pub mod tests {
             inserted_slot,
             &Pubkey::new_unique(),
             AccountSharedData::new(1, 0, &Pubkey::default()),
-            Hash::default(),
+            Some(Hash::default()),
+            inserted_slot,
+            ClusterType::Development,
         );
 
         // If the cache is told the size limit is 0, it should return nothing because there's only
--- a/runtime/src/accounts_cache.rs
+++ b/runtime/src/accounts_cache.rs
@@ -14,6 +14,7 @@ use std::{
         Arc, RwLock,
     },
 };
+use log::error;
 
 pub type SlotCache = Arc<SlotCacheInner>;
 
@@ -55,7 +56,7 @@ impl SlotCacheInner {
         hash: Option<Hash>,
         hash_slot: Slot,
         cluster_type: ClusterType,
-    ) {
+    ) -> CachedAccount {
         if self.cache.contains_key(pubkey) {
             self.same_account_writes.fetch_add(1, Ordering::Relaxed);
             self.same_account_writes_size
@@ -64,16 +65,18 @@ impl SlotCacheInner {
             self.unique_account_writes_size
                 .fetch_add(account.data().len() as u64, Ordering::Relaxed);
         }
+        let item = Arc::new(CachedAccountInner {
+            account,
+            hash: RwLock::new(hash),
+            hash_slot,
+            cluster_type,
+            pubkey: *pubkey,
+        });
         self.cache.insert(
             *pubkey,
-            Arc::new(CachedAccountInner {
-                account,
-                hash: RwLock::new(hash),
-                hash_slot,
-                cluster_type,
-                pubkey: *pubkey,
-            }),
+            item.clone(),
         );
+        item
     }
 
     pub fn get_cloned(&self, pubkey: &Pubkey) -> Option<CachedAccount> {
@@ -124,9 +127,13 @@ pub struct CachedAccountInner {
 
 impl CachedAccountInner {
     pub fn hash(&self) -> Hash {
+        error!("locking to read hash");
         let hash = self.hash.read().unwrap();
         match *hash {
-            Some(hash) => hash,
+            Some(hash) => {
+                error!("done locking to read hash");
+                hash
+            },
             None => {
                 let hash = crate::accounts_db::AccountsDb::hash_account(
                     self.hash_slot,
@@ -134,7 +141,9 @@ impl CachedAccountInner {
                     &self.pubkey,
                     &self.cluster_type,
                 );
+                error!("write locking to set hash");
                 *self.hash.write().unwrap() = Some(hash.clone());
+                error!("done write locking to set hash");
                 hash
             }
         }
@@ -182,7 +191,7 @@ impl AccountsCache {
         hash: Option<Hash>,
         hash_slot: Slot,
         cluster_type: ClusterType,
-    ) {
+    ) -> CachedAccount {
         let slot_cache = self.slot_cache(slot).unwrap_or_else(||
             // DashMap entry.or_insert() returns a RefMut, essentially a write lock,
             // which is dropped after this block ends, minimizing time held by the lock.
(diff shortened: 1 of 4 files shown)


jeffwashington:
this pr: (improvement is in store_us and overall time)
ledger processing timing ExecuteTimings load_us 10909296 execute_us 55929281 store_us 5725080 details ExecuteDetailsTimings serialize_us 10248310 create_vm_us 1681543 execute_us 7627294 deserialize_us 13642778 changed_account_count 57938 total_account_count 1113864 total_data_size 48123505204 data_size_changed 49975074
ledger processed in 34 seconds. 4571 MB allocated. root slot is 69585630 1 fork at 69585662 with 32 frozen banks

master:
ledger processing timing ExecuteTimings load_us 10516422 execute_us 54218813 store_us 21934006 details ExecuteDetailsTimings serialize_us 9357363 create_vm_us 1677496 execute_us 7424016 deserialize_us 12894874 changed_account_count 57938 total_account_count 1113864 total_data_size 48123505204 data_size_changed 49975074
ledger processed in 41 seconds. 5092 MB allocated. root slot is 69585630 1 fork at 69585662 with 32 frozen banks

jeffwashington:
first prototype: #15617

runtime/src/accounts_db.rs
```diff
@@ -691,6 +693,10 @@ pub struct AccountsDb {
 
     pub accounts_cache: AccountsCache,
 
+    sender_bg_hasher: Option<Mutex<Sender<CachedAccount>>>,
+
+    store_hasher: Option<Arc<JoinHandle<()>>>,
```
carllin, line 698: Doesn't seem like this is used anywhere, probably want a join on AccountsDb drop?

jeffwashington: Yes, it was intended to be used, but isn't necessary. bg hasher will now quit when sender_bg_hasher is destroyed when accountsdb is destroyed. I removed store_hasher.

runtime/src/accounts_db.rs
```diff
@@ -691,6 +693,8 @@ pub struct AccountsDb {
 
     pub accounts_cache: AccountsCache,
 
+    sender_bg_hasher: Option<Mutex<Sender<CachedAccount>>>,
```
carllin, line 696: I think if we switch to crossbeam channels which implement `Sync`, we can remove the `Mutex` here, i.e. like so: https://github.com/carllin/solana/commit/9d22ae16f54d1d85484e5fd848a54c595c820600

jeffwashington: yes, brilliant. done. I didn't like the mutex.

runtime/src/accounts_db.rs
```diff
@@ -230,10 +232,10 @@ impl<'a> LoadedAccount<'a> {
         }
     }
 
-    pub fn loaded_hash(&self) -> &Hash {
+    pub fn loaded_hash(&self) -> Hash {
         match self {
-            LoadedAccount::Stored(stored_account_meta) => &stored_account_meta.hash,
-            LoadedAccount::Cached((_, cached_account)) => &cached_account.hash,
+            LoadedAccount::Stored(stored_account_meta) => *stored_account_meta.hash,
+            LoadedAccount::Cached((_, cached_account)) => cached_account.hash(),
```
carllin, line 238: Ah clever, so if the hash isn't computed yet by the background service by this point, (for things like `bank.freeze()` that need the hash), it'll compute it here.

jeffwashington: Thanks. I am just implementing the required blocking, lazy behavior in a way that seems correct, consistent, performant. I think it fits Rust idioms sufficiently.

runtime/src/accounts_db.rs
```diff
@@ -1298,6 +1305,35 @@ impl AccountsDb {
...
 
+    fn background_hasher(receiver: Receiver<CachedAccount>) {
+        loop {
+            let result = receiver.recv();
+            match result {
+                Ok(account) => {
+                    // if we hold the only ref, then this account doesn't need to be hashed, we ignore this account and it will disappear
+                    if Arc::strong_count(&account) > 1 {
+                        // this will cause the hash to be calculated and store inside account if it needs to be calculated
+                        let _ = (*account).hash();
```
carllin, line 1316: Point of clarification, in the PR summary it was said that: ` I don't know enough yet about the race conditions possible in the cache to attempt to update the cached entry with a new hash. `

However, this will update the hash of that cached entry no? I think it should be safe because if anything relevant to the hash is touched (lamports, account data, slot, owner, etc.), then a new account object would be stored in the cache. I.e. if for a given slot X, we had:

1) Store account A into cache
2) Background thread hashes A
3) Store account A into cache
4) Background thread commits hash for account A into its `RwLock<Hash>`

Then this should be ok because 1) and 3) stored separate versions of `A` into the cache, so the update in 4) won't update the later version of `A`, it'll update the earlier version.

jeffwashington: yes, that pr comment was outdated. Good catch. I will update it.

runtime/src/accounts_db.rs
```diff
@@ -3351,16 +3387,44 @@ impl AccountsDb {
...
+                assert_eq!(hashes.len(), len);
+                hashes
+            }
+            None => &empty,
+        };
+
+        (0..len)
+            .into_iter()
+            .map(|i| {
+                let (meta, account) = &accounts_and_meta_to_store[i];
```
sakridge, line 3406: In general manual indexing is slightly slower in rust because the compiler cannot remove the bounds checks. Why we are using the iterator here originally.
Can use enumerate for the index for `hashes`

runtime/src/accounts_db.rs
```diff
@@ -3351,16 +3387,44 @@ impl AccountsDb {
...
+                let hash = if hashes.is_empty() {
+                    None
+                } else {
+                    Some(hashes[i])
```
sakridge, line 3410: would it be cleaner without the `empty` variable:
`let hash = hashes.map(|hashes| hashes[i]);`

runtime/src/accounts_cache.rs
```diff
@@ -128,7 +169,15 @@ impl AccountsCache {
...
     }
 
-    pub fn store(&self, slot: Slot, pubkey: &Pubkey, account: AccountSharedData, hash: Hash) {
+    pub fn store(
+        &self,
+        slot: Slot,
+        pubkey: &Pubkey,
+        account: AccountSharedData,
+        hash: Option<Hash>,
+        hash_slot: Slot,
```
sakridge, line 178: Does `hash_slot != slot` ever?

runtime/src/accounts_cache.rs
```diff
@@ -89,10 +105,35 @@ impl Deref for SlotCacheInner {
...
-#[derive(Debug, Clone)]
-pub struct CachedAccount {
+pub type CachedAccount = Arc<CachedAccountInner>;
+
+#[derive(Debug)]
+pub struct CachedAccountInner {
     pub account: AccountSharedData,
-    pub hash: Hash,
+    hash: RwLock<Option<Hash>>,
+    hash_slot: Slot,
```
sakridge, line 114: Why is it `hash_slot` and not just `slot` is there something about this slot that is specific to hashing? Or is it just the same slot that the account is stored at?

jeffwashington: I could call it 'slot'. My thinking was: I need the correct slot to compute the hash later. The slot is meant (by me) to be the same slot we would have used to calculate the hash had we done the hash calculation before adding the account to the cache. Thus, I called it 'hash_slot' in case there was confusion at some point later about why the slot is stored here.
I don't have enough system knowledge to know whether or when the slot could be different at a time when an account in the cache needs the slot for hashing or other potential purposes. Would you like me to rename it to 'slot'?

sakridge: `slot` is good. Just realized this data is kind of duplicated now. `cluster_type` never changes but now we store a copy for each cached account.

jeffwashington: > `cluster_type` never changes but now we store a copy for each cached account.

Yes, I do not like this. Do you know a good rust way to improve this?

sakridge: Well, there's standard plumbing it around from `AccountsDB::cluster_type`. The background service could have it's own copy which is move'ed in on initialization. I think the other places we need it already have `AccountsDB` reference?

Other than that there's `thread_local` or `lazy_static` but it's like global variables essentially and there is a perf hit for each access. Maybe it's a win overall from the original copy version as there would be less data polluting the caches. Would have to benchmark it to see.

sakridge (approved):
looks good and super cool to have higher performance

runtime/src/accounts_cache.rs
ryoqun: @jeffwashington @sakridge hi, rather late comment here. I think we can remove the cluster_type flag altogether because this ultimate code (https://github.com/solana-labs/solana/blob/b521f5027894f0ea4a07ac6dcc6c2419b7444868/runtime/src/accounts_db.rs#L2878-L2908) needing the flag is very old; we can remove the flag because we don't retain compatibility for the replayability since genesis. This will clean the code a lot across the call chain, assuming we don't ever change hash algorithm or its logic in near future ;)

related to this, I also want to clean this line of call-chain as well: https://github.com/solana-labs/solana/blob/b521f5027894f0ea4a07ac6dcc6c2419b7444868/runtime/src/accounts_db.rs#L3620-L3627
original pr: https://github.com/solana-labs/solana/pull/15763#discussion_r592080167


how about doing spring cleaning? :)

jeffwashington: yes, spring cleaning can occur.

jeffwashington: @ryoqun how do you want to get rid fo the cluster_type flag? Do we remove it from the [hash_stored_account function completely](https://github.com/solana-labs/solana/pull/16375)? is that what you're proposing?

ryoqun: @jeffwashington yeah, that's correct. from time to time, we completely remove old dead-code behidn feature-gate like this: #15763

PR #16161, opened 2021-03-26
Rpc: enable getConfirmedBlocks and getConfirmedBlocksWithLimit to return confirmed (not yet finalized) data

CriesofCarrots:
#### Problem
Similar to https://github.com/solana-labs/solana/pull/16142, `getConfirmedBlocks` and `getConfirmedBlocksWithLimit` do not included confirmed but not-yet-finalized blocks in the response list.

#### Summary of Changes
- Some plumbing to support optional commitment parameter in `getConfirmedBlocks`
- Expose confirmed data in `getConfirmedBlocks` and `getConfirmedBlocksWithLimit` via the commitment parameter
- Also, sneak in a couple fixes:
  - if no endSlot is provided to `getConfirmedBlocks` when startSlot is far in the past, return `MAX_GET_CONFIRMED_BLOCKS_RANGE` blocks instead of error
  - Check that blocks returned from Blockstore are actually finalized, not just local roots

### DIFF
--- a/client/src/rpc_config.rs
+++ b/client/src/rpc_config.rs
@@ -178,16 +178,16 @@ impl EncodingConfig for RpcConfirmedTransactionConfig {
 
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 #[serde(untagged)]
-pub enum EndSlotOrCommitment {
+pub enum RpcConfirmedBlocksConfig {
     EndSlotOnly(Option<Slot>),
     CommitmentOnly(Option<CommitmentConfig>),
 }
 
-impl EndSlotOrCommitment {
+impl RpcConfirmedBlocksConfig {
     pub fn unzip(&self) -> (Option<Slot>, Option<CommitmentConfig>) {
         match &self {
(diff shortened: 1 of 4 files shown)


CriesofCarrots:
@mvines can you give this a look?

client/src/rpc_config.rs
```diff
@@ -175,3 +175,19 @@ impl EncodingConfig for RpcConfirmedTransactionConfig {
         }
     }
 }
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum EndSlotOrCommitment {
```
mvines, line 181: naming nit: could this be something more like `RpcConfirmedBlocksConfig`

CriesofCarrots: Certainly. I'm inclined to go with `RpcConfimedBlocksWrapper`; if we need to add additional config args in the future, we'll want to drop in an `RpcConfirmedBlocksConfig` to replace CommitmentConfig.

mvines: sgtm. `EndSlotOrCommitment` just felt a little too generic

mvines (approved):

PR #16167, opened 2021-03-27
poll checking for new record in poh service after every batch of hashes instead of busy waiting

jeffwashington:
#### Problem
busy waiting with no ability to be interrupted can miss events.
Longer, less frequent busy waits causes hashes to not be uniformly spaced within a tick.
#### Summary of Changes
Check on progress after each batch of hashes in poh_service. If we are ahead, busy wait polling for record events. This spreads out the busy waiting and we can be productive during busy waits and respond more quickly to record events.
Fixes #

### DIFF
--- a/core/src/poh_service.rs
+++ b/core/src/poh_service.rs
@@ -267,34 +267,26 @@ impl PohService {
                         return true;
                     }
                     // check to see if a record request has been sent
-                    let get_again = record_receiver.try_recv();
-                    match get_again {
-                        Ok(record) => {
-                            // remember the record we just received as the next record to occur
-                            *next_record = Some(record);
-                            break;
-                        }
-                        Err(_) => (),
+                    if let Ok(record) = record_receiver.try_recv() {
+                        // remember the record we just received as the next record to occur
+                        *next_record = Some(record);
+                        break;
                     }
-
-                    // check to see if we need to busy wait to catch up to ideal
+                    // check to see if we need to wait to catch up to ideal
                     let wait_start = Instant::now();
                     if ideal_time <= wait_start {
+                        // no, keep hashing. We still hold the lock.
                         continue;
                     }
 
-                    // yes, busy wait, polling for new records and after dropping poh lock
+                    // busy wait, polling for new records and after dropping poh lock (reset can occur, for example)
                     drop(poh_l);
                     while ideal_time > Instant::now() {
                         // check to see if a record request has been sent
-                        let get_again = record_receiver.try_recv();
-                        match get_again {
-                            Ok(record) => {
-                                // remember the record we just received as the next record to occur
-                                *next_record = Some(record);
-                                break;
-                            }
-                            Err(_) => (),
+                        if let Ok(record) = record_receiver.try_recv() {
+                            // remember the record we just received as the next record to occur
+                            *next_record = Some(record);
+                            break;
                         }
                     }
                     timing.total_sleep_us += wait_start.elapsed().as_micros() as u64;
--- a/core/src/poh_service.rs
+++ b/core/src/poh_service.rs
@@ -122,6 +122,10 @@ impl PohService {
                         ticks_per_slot,
                         hashes_per_batch,
                         record_receiver,
+                        Self::target_ns_per_tick(
+                            ticks_per_slot,
+                            poh_config.target_tick_duration.as_nanos() as u64,
+                        ),
                     );
                 }
                 poh_exit_.store(true, Ordering::Relaxed);
@@ -210,6 +214,7 @@ impl PohService {
         record_receiver: &Receiver<Record>,
         hashes_per_batch: u64,
         poh: &Arc<Mutex<Poh>>,
+        target_ns_per_tick: u64,
     ) -> bool {
         match next_record.take() {
             Some(mut record) => {
@@ -254,6 +259,7 @@ impl PohService {
                     timing.num_hashes += hashes_per_batch;
                     let mut hash_time = Measure::start("hash");
                     let should_tick = poh_l.hash(hashes_per_batch);
+                    let ideal_time = poh_l.target_poh_time(target_ns_per_tick);
                     hash_time.stop();
                     timing.total_hash_time_ns += hash_time.as_ns();
                     if should_tick {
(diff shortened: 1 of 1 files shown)


jeffwashington:
Highest TPS: 136039.83 sampling period 1s max transactions: 9441365 clients: 1 drop rate: 0.51
[2021-03-27T03:19:52.654261485Z INFO  solana_bench_tps::bench] 	Average TPS: 104487.58

jeffwashington:
built off: #16139

jeffwashington:
bench-tps on a colo machine
Screen Shot 2021-03-26 at 10 47 13 PM

jeffwashington:
where we sleep:
datapoint: **poh-service** ticks=181i hashes=8876483i elapsed_us=354289i **total_sleep_us=411314i** total_tick_time_us=1234i total_lock_time_us=3494i total_hash_time_us=531635i total_record_time_us=13i
datapoint: **poh_recorder** slot=6381i tick_lock_contention=0i record_us=0i flush_cache_no_tick_us=0i flush_cache_tick_us=365i prepare_send_us=0i send_us=3i ticks_from_record=0i **total_sleep_us=738i** tick_overhead=0i record_lock_contention=0i

core/src/poh_service.rs
```diff
@@ -122,6 +122,10 @@ impl PohService {
                         ticks_per_slot,
                         hashes_per_batch,
                         record_receiver,
+                        Self::target_ns_per_tick(
```
carllin, line 125: unrelated to this PR, but it seems like since `target_ns_per_tick()` is just subtracting a constant number from `poh_config.target_tick_duration` every time, might make sense to just move this into the initialization of `poh_config.target_tick_duration`

core/src/poh_service.rs
```diff
@@ -254,24 +259,38 @@ impl PohService {
...
+                    drop(poh_l);
+                    while ideal_time > Instant::now() {
+                        // check to see if a record request has been sent
+                        if let Ok(record) = record_receiver.try_recv() {
                             // remember the record we just received as the next record to occur
                             *next_record = Some(record);
                             break;
                         }
-                        Err(_) => {
-                            continue;
```
carllin, line 272: Oh good catch, would this have looped forever if the `record_receiver` was dropped before (such that this always errored?)

core/src/poh_service.rs
```diff
@@ -254,24 +259,38 @@ impl PohService {
...
                         // nothing else can be done. tick required.
                         return true;
                     }
                     // check to see if a record request has been sent
-                    let get_again = record_receiver.try_recv();
-                    match get_again {
-                        Ok(record) => {
+                    if let Ok(record) = record_receiver.try_recv() {
+                        // remember the record we just received as the next record to occur
+                        *next_record = Some(record);
```
carllin, line 272: Separate thought: Because `PoHRecorder::record()` can tick(), without incurring the spinwait/throttle, if we had enough records, would that entirely circumvent the throttle/waits? Maybe this is highly unlikely unless we have a bunch of banking threads...

core/src/poh_service.rs
```diff
@@ -254,24 +259,38 @@ impl PohService {
...
+                        break;
+                    }
+                    // check to see if we need to wait to catch up to ideal
+                    let wait_start = Instant::now();
+                    if ideal_time <= wait_start {
```
carllin, line 277: is it ever possible to get into a situation where we're constantly behind and won't record? I.e. something like:

1. PoH is behind the ideal, so we don't record, and hash until the next tick, break.
2.  Now we're only slightly ahead so PoHService spin waits for `X` ms.
3. Due to some combination of scheduling/overhead/the fact that the sleep in 2., isn't exact, when we get to this Poh loop again, we are behind again, so once again we don't record,

carllin: Also might be good to discuss with @aeyakovenko, because it seems PoH is getting less and less strict here.

With these and previous changes to throttle PoH, you  need hardware that is able to run PoH much faster than the actual output tick rate of the cluster. I think @sakridge has estimated you need at least a `5-10x` faster hashing speed in order to generate our current PoH due to these throttles, but an attacker could turn off these throttles.

core/src/poh_service.rs
jeffwashington: we still throttle in pohrecorder::tick. So, if pohrecorder::record calls pohrecorder::tick(), we still throttle. The goal of this is to reduce the useless work spinning in pohrecorder::tick and check for records while waiting for time to elapse in pohservice.

core/src/poh_service.rs
jeffwashington: > is it ever possible to get into a situation where we're constantly behind and won't record? I.e. something like:
> 
> 1. PoH is behind the ideal, so we don't record, and hash until the next tick, break.
> 2. Now we're only slightly ahead so PoHService spin waits for `X` ms.
> 3. Due to some combination of scheduling/overhead/the fact that the sleep in `2.`, isn't exact, when we get to this Poh loop again, we are behind again, so once again we don't record.
> 
> Eventually maybe you finally are able to record, and you have a huge backlog of records

record requests always cause us to stop hashing and handle all queued up records before we hash any more. look for record_receiver.try_recv()

core/src/poh_service.rs
jeffwashington: I don't think this change significantly changes what happens on an error - especially if all record_senders are dropped. I think we'll always get a try_recv() error and we'll always just ignore the error and keep hashing until it is time to tick and then we notice pohservice is told to quit.

carllin: ah yeah, there's a return in the `if should_tick {`

core/src/poh_service.rs
sakridge: > `5-10x` faster hashing speed in order to generate our current PoH due to these throttles, but an attacker could turn off these throttles.

All these will increase the efficiency so that we can get closer to the peak PoH speed, do a ton of `record`'s and still hit the wallclock time of the slot. The current one was not great in that you needed a 5-10x over speed to go the speed you wanted at load.

core/src/poh_service.rs
carllin: ah yup you're right, that `spin_loop` is in the `tick()` function itself, not inside the body of  PoH service.

carllin (approved):

sakridge (approved):
lgtm

PR #16195, opened 2021-03-29
TransactionRecorder uses unique channel so we can use Recv instead of RecvTimeout

jeffwashington:
#### Problem
Get rid of sender.clone on TransactionRecorder to handle test shutdown cases better.
#### Summary of Changes
Allocate a new channel for each record. This means when the sender is dropped, the receiver will fail. Previously, we had a clone, and there was still a sender in existence, so the receiver never finished during shutdown. This necessitated a timeout. Timeouts can cause trouble.
I measured the cost to create channels, and it mesured low. ~60ns (I tested creating 1M of them).

Fixes #

### DIFF
--- a/core/src/poh_recorder.rs
+++ b/core/src/poh_recorder.rs
@@ -105,6 +105,20 @@ impl TransactionRecorder {
         mixin: Hash,
         transactions: Vec<Transaction>,
     ) -> Result<()> {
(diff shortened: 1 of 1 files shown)


carllin (approved):

sakridge (approved):
lgtm

PR #16199, opened 2021-03-29
Add handling to close accounts to many-accounts bench

CriesofCarrots:
#### Problem
In the wild, accounts db sometimes sees issues due to emptied accounts. Our many-accounts client currently only creates accounts, but could be extended to also close them.

#### Summary of Changes
- Add parameter to accounts-cluster-bench to close every nth account created; uses derived account addresses to make this work


sakridge (approved):
nice!

PR #16210, opened 2021-03-30
get rid of data copy

jeffwashington:
#### Problem
Code migration left a location where we unnecessarily cause a data copy to be made.
#### Summary of Changes
Don't cause a copy to be made.
Fixes #


sakridge (approved):
lgtm

PR #16211, opened 2021-03-30
Cleanup nits

jackcmay:
#### Problem

Misc stuff that I didn't want to bundle in a feature pr

#### Summary of Changes

Fixes #


jackcmay:
@mvines Looks like some of the program error to instruction error mapping was neglected and now downstream tests are failing because instead of `invalidError` with this pr we are returning the real error.  The errors are in stake pool, we could fix those but others may have issues.  Also, looks like we might need to featurize these since they result in a different `InstructionError`.  Any ideas of a good way to solve this?

PR #16218, opened 2021-03-30
Return sysvars via syscalls

### LINKED ISSUE #15755: Consider mapping sysvars to syscalls
Sysvars are read-only accounts the the user must remember to pass into each instruction that needs access.   This has produced problems:
1. such as https://github.com/solana-labs/solana/pull/15497, where the runtime now must mitigate transactions that incorrectly acquire a write-lock on the read-only sysvars, and   
2. at more basic level, sysvars clog up the transaction with extra addresses

Consider mapping them to syscalls, such as clock() or rent(), which could enable direct access to sysvars from programs without going through the accounts database

jackcmay:
#### Problem

Requiring developers to include sysvars in their instructions complicates composibility and requires programs to pass the sysvar accounts between programs when an intermediate might not actually need them.

Also, because of the bincode serialization of the sysvar data and the lack of a C parser we currently don't support sysvars in C BPF programs.

#### Summary of Changes

- Allow a program to call a syscall to obtain the sysvar.
- Do so in a way that does NOT require the bincode deserialization in-program

Needed by: https://github.com/solana-labs/solana/pull/15927

Fixes #15755

### DIFF
--- a/program-test/src/lib.rs
+++ b/program-test/src/lib.rs
@@ -16,8 +16,8 @@ use {
     solana_sdk::{
         account::{Account, AccountSharedData, ReadableAccount},
         account_info::AccountInfo,
-        clock::Slot,
-        entrypoint::ProgramResult,
+        clock::{Clock, Slot},
+        entrypoint::{ProgramResult, SUCCESS},
         feature_set::demote_sysvar_write_locks,
         fee_calculator::{FeeCalculator, FeeRateGovernor},
         genesis_config::{ClusterType, GenesisConfig},
@@ -28,12 +28,13 @@ use {
         message::Message,
         native_token::sol_to_lamports,
         process_instruction::{
-            stable_log, BpfComputeBudget, InvokeContext, ProcessInstructionWithContext,
+            stable_log, BpfComputeBudget, InvokeContext, ProcessInstructionWithContext, Sysvars,
         },
-        program_error::ProgramError,
+        program_error::{ProgramError, ACCOUNT_BORROW_FAILED, UNSUPPORTED_SYSVAR},
         pubkey::Pubkey,
         rent::Rent,
         signature::{Keypair, Signer},
+        sysvar::{clock, rent},
     },
     solana_vote_program::vote_state::{VoteState, VoteStateVersions},
     std::{
@@ -42,7 +43,7 @@ use {
         convert::TryFrom,
         fs::File,
         io::{self, Read},
-        mem::transmute,
+        mem::{size_of, transmute},
         path::{Path, PathBuf},
         rc::Rc,
         sync::{
@@ -332,6 +333,55 @@ impl solana_sdk::program_stubs::SyscallStubs for SyscallStubs {
 
         stable_log::program_success(&logger, &program_id);
         Ok(())
+    }
+
+    fn sol_get_sysvar(&self, id: &Pubkey, var_addr: *mut u8) -> u64 {
+        let invoke_context = get_invoke_context();
+
+        ic_msg!(invoke_context, "sol_get_sysvar");
+
+        let sysvars = match invoke_context.get_sysvar(id).ok_or_else(|| {
+            ic_msg!(invoke_context, "Unable to get Sysvar {}", id);
+            UNSUPPORTED_SYSVAR
+        }) {
+            Ok(sysvar) => sysvar,
+            Err(err) => return err,
+        };
+
+        let size = if *id == clock::id() {
+            match *sysvars {
+                Sysvars::Clock(clock) => {
+                    let var = var_addr as *mut _ as *mut Clock;
+                    unsafe { *var = clock }
+                }
+                _ => return UNSUPPORTED_SYSVAR,
+            }
+            size_of::<Clock>()
+        } else if *id == rent::id() {
+            match *sysvars {
+                Sysvars::Rent(rent) => {
+                    let var = var_addr as *mut _ as *mut Rent;
+                    unsafe { *var = rent }
+                }
+                _ => return UNSUPPORTED_SYSVAR,
+            }
+            size_of::<Rent>()
+        } else {
+            return UNSUPPORTED_SYSVAR;
+        };
+
+        if invoke_context
+            .get_compute_meter()
+            .try_borrow_mut()
+            .map_err(|_| ACCOUNT_BORROW_FAILED)
+            .unwrap()
+            .consume(invoke_context.get_bpf_compute_budget().sysvar_base_cost + size as u64)
+            .is_err()
+        {
+            panic!("Exceeded compute budget");
+        }
+
+        SUCCESS
     }
 }
 
--- a/runtime/src/message_processor.rs
+++ b/runtime/src/message_processor.rs
@@ -250,11 +250,6 @@ impl ComputeMeter for ThisComputeMeter {
         self.remaining
     }
 }
-#[derive(Default)]
-pub struct SysvarsHolder {
-    pub clock: Option<Clock>,
-    pub rent: Option<Rent>,
-}
 pub struct ThisInvokeContext<'a> {
     program_ids: Vec<Pubkey>,
     rent: Rent,
--- a/runtime/src/message_processor.rs
+++ b/runtime/src/message_processor.rs
@@ -1,6 +1,6 @@
 use crate::{
-    instruction_recorder::InstructionRecorder, log_collector::LogCollector,
-    native_loader::NativeLoader, rent_collector::RentCollector,
+    accounts::Accounts, accounts_index::Ancestors, instruction_recorder::InstructionRecorder,
+    log_collector::LogCollector, native_loader::NativeLoader, rent_collector::RentCollector,
 };
 use log::*;
 use serde::{Deserialize, Serialize};
@@ -8,6 +8,7 @@ use solana_sdk::{
     account::{AccountSharedData, ReadableAccount, WritableAccount},
     account_utils::StateMut,
     bpf_loader_upgradeable::{self, UpgradeableLoaderState},
+    clock::Clock,
     feature_set::{
         cpi_share_ro_and_exec_accounts, demote_sysvar_write_locks, instructions_sysvar_enabled,
         FeatureSet,
@@ -19,11 +20,12 @@ use solana_sdk::{
     native_loader,
     process_instruction::{
         BpfComputeBudget, ComputeMeter, Executor, InvokeContext, Logger,
-        ProcessInstructionWithContext,
+        ProcessInstructionWithContext, Sysvars,
     },
     pubkey::Pubkey,
     rent::Rent,
     system_program,
+    sysvar::{clock, rent},
     transaction::TransactionError,
 };
 use std::{
@@ -248,6 +250,11 @@ impl ComputeMeter for ThisComputeMeter {
         self.remaining
     }
 }
+#[derive(Default)]
+pub struct SysvarsHolder {
+    pub clock: Option<Clock>,
+    pub rent: Option<Rent>,
+}
 pub struct ThisInvokeContext<'a> {
     program_ids: Vec<Pubkey>,
     rent: Rent,
@@ -262,6 +269,9 @@ pub struct ThisInvokeContext<'a> {
     instruction_recorder: Option<InstructionRecorder>,
     feature_set: Arc<FeatureSet>,
     pub timings: ExecuteDetailsTimings,
+    account_db: Arc<Accounts>,
+    ancestors: &'a Ancestors,
+    sysvars: Vec<(Pubkey, Option<Rc<Sysvars>>)>,
 }
 impl<'a> ThisInvokeContext<'a> {
     #[allow(clippy::too_many_arguments)]
@@ -277,6 +287,8 @@ impl<'a> ThisInvokeContext<'a> {
         executors: Rc<RefCell<Executors>>,
         instruction_recorder: Option<InstructionRecorder>,
         feature_set: Arc<FeatureSet>,
+        account_db: Arc<Accounts>,
+        ancestors: &'a Ancestors,
     ) -> Self {
         let mut program_ids = Vec::with_capacity(bpf_compute_budget.max_invoke_depth);
         program_ids.push(*program_id);
@@ -296,6 +308,28 @@ impl<'a> ThisInvokeContext<'a> {
             instruction_recorder,
             feature_set,
             timings: ExecuteDetailsTimings::default(),
+            account_db,
+            ancestors,
+            sysvars: vec![],
+        }
+    }
+    fn load_sysvar(&self, pubkey: &Pubkey) -> Option<Rc<Sysvars>> {
+        let account = self
+            .account_db
+            .load_slow(self.ancestors, pubkey)
+            .map(|(account, _)| Rc::new(RefCell::new(account)))?;
+        let borrowed = &account.try_borrow_mut().ok()?;
+
+        if &clock::id() == pubkey {
+            Some(Rc::new(Sysvars::Clock(
+                bincode::deserialize(&borrowed.data()[..]).ok()?,
+            )))
+        } else if &rent::id() == pubkey {
+            Some(Rc::new(Sysvars::Rent(
+                bincode::deserialize(&borrowed.data()[..]).ok()?,
+            )))
+        } else {
+            None
         }
     }
 }
@@ -414,7 +448,19 @@ impl<'a> InvokeContext for ThisInvokeContext<'a> {
         self.timings.execute_us += execute_us;
         self.timings.deserialize_us += deserialize_us;
     }
+    fn get_sysvar(&mut self, pubkey: &Pubkey) -> Option<Rc<Sysvars>> {
+        let mut result =
+            self.sysvars
+                .iter()
+                .find_map(|(key, sysvar)| if pubkey == key { sysvar.clone() } else { None });
+        if result.is_none() {
+            result = Some(self.load_sysvar(pubkey)?);
+            self.sysvars.push((*pubkey, result.clone()));
+        }
+        result
+    }
 }
+
 pub struct ThisLogger {
     log_collector: Option<Rc<LogCollector>>,
 }
@@ -1044,6 +1090,8 @@ impl MessageProcessor {
         bpf_compute_budget: BpfComputeBudget,
         timings: &mut ExecuteDetailsTimings,
         demote_sysvar_write_locks: bool,
+        account_db: Arc<Accounts>,
+        ancestors: &Ancestors,
     ) -> Result<(), InstructionError> {
         // Fixup the special instructions key if present
         // before the account pre-values are taken care of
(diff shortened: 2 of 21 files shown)


jackcmay:
@joncinque @jstarry What do you think of the syscall API?

runtime/src/accounts.rs
```diff
@@ -299,6 +300,14 @@ impl Accounts {
                                 )
                             })
                             .collect::<Result<TransactionLoaders>>()?;
+
+                        // Load sysvar accounts that will be made available to programs via syscalls
```
jackcmay, line 304: @sakridge Is there a better way to load the sysvar accounts in order to make them available to programs via a syacall API?

joncinque:
This looks good to me!  I'm not sure if this was already in your plans, but could you also add a stub for `sol_get_sysvar` in `program-test` at https://github.com/solana-labs/solana/blob/5eff23db0c01e6c976bc732510771865747bd17b/program-test/src/lib.rs#L210 ?  How would that work in this case?  Would you need to get the clock / rent using `get_account()` from the `InvokeContext`?

sdk/program/src/sysvar/mod.rs
```diff
@@ -26,6 +26,9 @@ pub fn is_sysvar_id(id: &Pubkey) -> bool {
         || instructions::check_id(id)
 }
 
+/// Sysvars that can be obtained via a syscall using `.get()`
+pub static SYSCALL_SUPPORTED_SYSVARS: &[Pubkey] = &[clock::ID, rent::ID];
```
mvines, line 30: Any reason to not include all sysvars?

sdk/program/src/program_error.rs
```diff
@@ -43,6 +43,8 @@ pub enum ProgramError {
     BorshIoError(String),
     #[error("An account does not have enough lamports to be rent-exempt")]
     AccountNotRentExempt,
+    #[error("Unsupported operations")]
+    Unsupported,
```
mvines, line 47: Specializing this error seems nicer: perhaps `UnsupportedSysvar`

runtime/src/accounts.rs
mvines: Incurring this cost for every single transaction doesn't feel grate.   Seems like we need a new on-demand interface to load read-only accounts from within the syscall handler

sdk/program/src/sysvar/mod.rs
jackcmay: In order to make these sysvars available via a syscall the runtime has to load their corresponding accounts for all transactions because it does not know whether they will be needed or not.  I figured initially we can limit the sysvars to the ones we know we need and are most used by folks, and expand as needed as the account caching stuff gets better.

sdk/program/src/program_error.rs
jackcmay: Yeah, I was on the fence about that, making a specialized error for every use case seemed messy for a case where unsupported can be a pretty obvious generic response, like in this case where you call a function that takes no arguments and it returns unsupported

runtime/src/accounts.rs
jackcmay: Yeah, not sure which is better, grabbing the lock whenever someone asks for a sysvar or loading the sysvar account as readonly from cache while the lock is already held...

jackcmay: If we put `bank.ancestors` behind `Rc` we could share the bank's accounts and ancestors with invoke_context to load the sysvars on demand, we could load on the initial request in the instruction.  @sakridge any thoughts on this and general direction?

sakridge: Loading on-demand is preferred I think and `bank.ancestors` in `Rc` also sounds reasonable.

jackcmay: Another option is we create a distinct external call for each sysvar which will result in unresolved external symbols for each syscall the program calls.  With that, we know what syscalls the program is planning to call and we can load just those once as a bundle with a single lock.  If some major restructuring, we could make that information known during the initial account load and do all the loading in the initial load lock.

mvines: > create a distinct external call for each sysvar

That's a cool idea.  One downside I see is that this'll cause sysvar  account loads for all instructions, even if the sysvar is never used

sakridge: It does sound attractive to have them be able to be scanned in the program and to be able to load them in this way. I'm thinking about a GPU implementation and like a thunk-style implementation will be kind of hard to do. It would be better to know it upfront and load them before getting into the interpreter itself. This scan could be done on program use and cached somewhere as well similar to the JIT binary.

jackcmay: @mvines Are you referring to the case where the transaction fails before a particular syscall is called?

mvines: Just because there's a unresolved sysvar symbol doesn't mean that all program instructions will require that sysvar, maybe the sysvar is only used one in a million instructions.

Although I hear sakridge's point about planning for future GPU support, where loading then all once up front to avoid a GPU stall is probably better than optimizing the low TPS case

jackcmay: I'm planning on loading on demand for now and exporting separate symbols for each sysvar so that we have the flexibility to optimize in the future.  Any objections to that direction?

mvines: sgtm, on-demand loading for cpu case seems great, with the option to batch if/when there's GPU offload available

sakridge: sounds good

sdk/program/src/sysvar/mod.rs
jackcmay: The latest changes also only support Clock and Rent, but it is easy to add the rest.  If the direction looks good I can add additional sysvars if folks feel like they should be

sdk/program/src/program_error.rs
jackcmay: Switched to `UnsupportedSysvar`

jackcmay:
> This looks good to me! I'm not sure if this was already in your plans, but could you also add a stub for `sol_get_sysvar` in `program-test` at
> 
> https://github.com/solana-labs/solana/blob/5eff23db0c01e6c976bc732510771865747bd17b/program-test/src/lib.rs#L210
> 
> ? How would that work in this case? Would you need to get the clock / rent using `get_account()` from the `InvokeContext`?

Add support to `ProgramTest`, check it out

jackcmay:
@joncinque @mvines Thoughts on adding more/all of the sysvars (just clock and rent right now)?

mvines:
> @joncinque @mvines Thoughts on adding more/all of the sysvars (just clock and rent right now)?

stake_history is needed by spl-stake-pool.  fees, epoch_schedule seems easy to add.  But then why not just add them all now so we don't need to deal with one-off features later to add them as people want them.

rewards is the only one I think we shouldn't add for sure

joncinque:
> stake_history is needed by spl-stake-pool. fees, epoch_schedule seems easy to add. But then why not just add them all now so we don't need to deal with one-off features later to add them as people want them.

:+1: on those.  Would `Instructions` be tough to add? I could see some neat use-cases for that too

program-test/src/lib.rs
```diff
@@ -333,6 +334,55 @@ impl solana_sdk::program_stubs::SyscallStubs for SyscallStubs {
         stable_log::program_success(&logger, &program_id);
         Ok(())
     }
+
+    fn sol_get_sysvar(&self, id: &Pubkey, var_addr: *mut u8) -> u64 {
+        let invoke_context = get_invoke_context();
+
+        ic_msg!(invoke_context, "sol_get_sysvar");
+
+        let sysvars = match invoke_context.get_sysvar(id).ok_or_else(|| {
```
joncinque, line 343: Oh nice, this is much simpler than I expected, thanks for providing the example!

mvines:
(Adding v1.6 label since I'm pretty sure we'll want this on mainnet sooner than v1.7 will hit mainnet, for iTokens)

sdk/src/process_instruction.rs
```diff
@@ -23,6 +25,13 @@ pub type LoaderEntrypoint = unsafe extern "C" fn(
 pub type ProcessInstructionWithContext =
     fn(&Pubkey, &[KeyedAccount], &[u8], &mut dyn InvokeContext) -> Result<(), InstructionError>;
 
+// Possible sysvar values returned by get_sysvar
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum Sysvars {
```
mvines, line 30:
```suggestion
pub enum Sysvar {
```

runtime/src/message_processor.rs
```diff
@@ -296,6 +303,28 @@ impl<'a> ThisInvokeContext<'a> {
...
+        let borrowed = &account.try_borrow_mut().ok()?;
+
+        if &clock::id() == pubkey {
+            Some(Rc::new(Sysvars::Clock(
+                bincode::deserialize(&borrowed.data()[..]).ok()?,
+            )))
+        } else if &rent::id() == pubkey {
+            Some(Rc::new(Sysvars::Rent(
+                bincode::deserialize(&borrowed.data()[..]).ok()?,
+            )))
```
mvines, line 325: I think the `bincode::deserialize()` here is unnecessary.   The data ends up getting unsafely cast back to a buffer for transport across the syscall boundary anyway, so it looks like the `Sysvars` struct could be removed entirely leaving us with `sysvars: Vec<(Pubkey, Option<Rc<Vec<u8>>>)>,`.

mvines (commented):
I'm not a fan of how many places need to be touched, all over the tree, to add a new syscall sysvar.  I think the code could be generalized to support all sysvars

sdk/program/src/sysvar/mod.rs
```diff
@@ -70,6 +78,22 @@ pub trait Sysvar:
...
         bincode::serialize_into(&mut account_info.data.borrow_mut()[..], self).ok()
     }
+    fn get() -> Result<Self, ProgramError> {
+        let mut var = Self::default();
+        let var_addr = &mut var as *mut _ as *mut u8;
+
+        #[cfg(target_arch = "bpf")]
+        let result = Self::call_syscall(var_addr);
+        #[cfg(not(target_arch = "bpf"))]
+        let result = crate::program_stubs::sol_get_sysvar(&Self::id(), var_addr);
```
mvines, line 88: Can we make the program_stubs signature match the BPF signature?  It's unfortunate to have to muddy the code with more `target_arch` conditionals

runtime/src/message_processor.rs
jackcmay: I originally was just generically passing the data back across the syscall boundary as `u8` but wanted to avoid doing the `bincode::deserialize()` in BPF program space.  Was much simpler codewise passing u8 but performance-wise not so much.  Also, the current approach allows us to cache the deserialized value.

sdk/program/src/sysvar/mod.rs
jackcmay: Stubs follow this model in general (https://github.com/solana-labs/solana/blob/1359bceb5d6c73a7b4852156d4c8676034e2e446/sdk/program/src/program.rs#L53). What did you have in mind for redirecting non-BPF program usage of the sysvar?

jackcmay:
> I'm not a fan of how many places need to be touched, all over the tree, to add a new sysvar syscall. I think the code could be generalized to support all sysvars

Agreed, I wrested with generics quite a bit in this pr, open to suggestions.  Rust isn't too kind to attempts to return generics in a trait returned by a trait.  The original implementation (which I still have) returns the data as u8 to the program space and deserializes it via rust generics and also had a mechanism to ensure that if a new sysvar was added the build would break until the sysvar was updated.  But comes with the trade-off of doing the bincode in BPF.  Though this attempt is complicated by the `history` sysvars which deserialize to vectors and thus are not able to be passed directly back to the program, so maybe the point is mute.  Have we concluded that borsh is binary compatible with bincode?  If so, maybe we borsh deserialize in program space.

jackcmay:
Just for reference, deserializing 11 slot hashes from the account data with bincode is taking 38_697 compute units  and that seems to scale "nicely" where 20 slot hashes takes 69_963 units :-(

jackcmay:
Maybe the better answer is the runtime deserializes the data and maps the final sysvar types into the program's address space.  This will require constructing the type in a way that will fit into a contiguous block of memory, for example SlotHashes is a `vec` so buffer might look like (ptr to data, len, cap, data, data data data...).  This would eliminate the computation and heap pressure reading the sysvar.

jackcmay:
@joncinque  looks like SPL Stake Pool is using bincode to deserialize StakeHistory, I expect that to start failing due to the growing size of StakeHistory sysvar

runtime/src/message_processor.rs
mvines: > I originally was just generically passing the data back across the syscall boundary as u8 but wanted to avoid doing the bincode::deserialize() in BPF program space. 

The program needs to `bincode::deserialize()` if the sysvar is passed in as an account though right, so we're not making things worse.  Keeping the interface consistent  between the now two paths to obtain a sysvar seems desirable and, better, should be much simpler code wise

>  Also, the current approach allows us to cache the deserialized value.

But the program really shouldn't be calling the syscall multiple times (except for perhaps some library code that a program imported)

sdk/program/src/sysvar/mod.rs
mvines: ugh ok :)

mvines:
> Maybe the better answer is the runtime deserializes the data and maps the final sysvar types into the program's address space. This will require constructing the type in a way that will fit into a contiguous block of memory, for example SlotHashes is a `vec` so buffer might look like (ptr to data, len, cap, data, data data data...). This would eliminate the computation and heap pressure reading the sysvar.

I really think raw bytes from the sysvar is fine.  Keep the runtime slim and make the program deserialize it.  Just like it works for account data.
 
The SlotHash sysvar is another example of where bincode was a poor choice (cc: https://github.com/mvines/solana-bpf-program-template/pull/4), but I think we improve the situation by deprecating the expensive sysvar and replacing it with something that's encoded in a form that's more usable from BPF when needed

joncinque:
@jackcmay There's a bit of bincode usage through the stake pool program, so I'd be very interested in getting Borsh to work with some of these types.  I haven't done tests to see if they're compatible unfortunately.

In the specific case of `StakeHistory` we could actually get rid of the deserialization and allow the stake program to do the required checks, but it's definitely a concern.  I retested this, and one stake account deserialization took roughly 15k BPF instructions.

mvines:
> Have we concluded that borsh is binary compatible with bincode? If so, maybe we borsh deserialize in program space.

It's not compatible.  However I think [speedy](https://crates.io/crates/speedy) can be made to decode most (all?) bincode encoded structures.   S

jackcmay:
> > Maybe the better answer is the runtime deserializes the data and maps the final sysvar types into the program's address space. This will require constructing the type in a way that will fit into a contiguous block of memory, for example SlotHashes is a `vec` so buffer might look like (ptr to data, len, cap, data, data data data...). This would eliminate the computation and heap pressure reading the sysvar.
> 
> I really think raw bytes from the sysvar is fine. Keep the runtime slim and make the program deserialize it. Just like it works for account data.
> 
> The SlotHash sysvar is another example of where bincode was a poor choice (cc: [mvines/solana-bpf-program-template#4](https://github.com/mvines/solana-bpf-program-template/pull/4)), but I think we improve the situation by deprecating the expensive sysvar and replacing it with something that's encoded in a form that's more usable from BPF when needed

Yeah, I'm leaning toward the model the Stephen used for `Instructions` where there are helpers provided to pull out specific instructions from the serialized (in hist case manually) data.  

`SlotHashes, `SlotHistory`, and `StakeHistory` are examples of sysvar accounts that can become (or are already) quite large which poses in-program allocation issues.  For example if a program were to iterate through the list of slots, we could have a helper that operates on the sterilized data but we wouldn't want to do any allocations if possible.  We should probably take these sysvar accounts case by case because, small accounts could be copied into program space, some might need to be mapped (SlotHistory is currently huge).

I'm fine going back to passing back u8 data and I think we can support clock, epoch_schedule, rent, fees, instructions for now (though we might want to provide some kind of low cost manual deserialization rather than generic bincode.  For the large ones, we are going to have to do something different and for now the sysvar will return unsupported.

runtime/src/message_processor.rs
jackcmay: The problem here is that `bincode::deserialize()` doesn't work for the large sysvars and `StakeHistory` is slippy by right now because the history isn't very long.  Pretty soon it is going to grow too large to deserrialize in program.

I see two options:
- map the sysvar into the program space and provide helpers that interact with it (`get_num_stakes` and `get_stake_at_index` come to mind).
- export syscalls on a per-syscall basis that does the same as the helpers I just mentioned

mvines:
> I'm fine going back to passing back u8 data and I think we can support clock, epoch_schedule, rent, fees for now (though we might want to provide some kind of low cost manual deserialization rather than generic bincode). For the large ones, we are going to have to do something different and for now the sysvar will return unsupported.

Sounds great to me

sakridge:
> > Have we concluded that borsh is binary compatible with bincode? If so, maybe we borsh deserialize in program space.
> 
> It's not compatible. However I think [speedy](https://crates.io/crates/speedy) can be made to decode most (all?) bincode encoded structures.

Yes, I tried this with borsh. The `Clock` sysvar is compatible because it's just 4 u64 in a struct, but there doesn't seem to be any speed benefit from using borsh in that case either. Something like SlotHashes is definitely not compatible.

jackcmay:
@mvines, gonna push one more time against returning serialized u8 data across the syscall boundary for the clock, epoch-schedule, frees, rent, and any other simple structure sysvars since we are creating distant syscalls for each anyway.  :-) 
These are small `repr(c)` structures and are very efficient if we load, deserialize, and cache these in the runtime (would ideally like to also across the whole tx).  I foresee a program calling `get()` multiple times once we provide this API and also inner instructions might call `get()` on the same sysvar.

For the larger and complex sysvars I'm still working on a method to communicate those as u8 data and then do surgical manual deserialization as we do currently with `Instructions`.  In these cases, I'm thinking of returning something like type`Instructions<'a>(&'a [u8]) where the slice has been mapped into the program's address space.  Users would then call `instruction.get_instruction()` to deserialize.  Each of these sysvars would have its own set of associated helper functions

mvines:
> @mvines, gonna push one more time against returning serialized u8 data across the syscall boundary for the clock, epoch-schedule, frees, rent, and any other simple structure sysvars since we are creating distant syscalls for each anyway. :-)
> These are small `repr(c)` structures and are very efficient if we load, deserialize, and cache these in the runtime (would ideally like to also across the whole tx). I foresee a program calling `get()` multiple times once we provide this API and also inner instructions might call `get()` on the same sysvar.

🤔 so the ABI boundary would be `repr(c)` things that are natively representable in C, Rust?  I do see your point. 

I think my initial recoil was when I saw a bunch of `match` statements in various different crates, implying that to add a new sysvar syscall in the future one would need to make edits in multiple places.  For example having to edit program-test, ugh.

jackcmay:
We'llll probably have to edit program-test anyway to add the new sysvar's function if a new sysvar is added.  It was either add a new function to program-stubs or add a new if case in the common stubbs function.  Besides the fact that program-test is in serious need of being combined with the runtime, did you have an idea of how to simplify the stubs?

PR #16275, opened 2021-03-31
update timings only on outer call

jeffwashington:
#### Problem
we are double counting metrics for executing transactions if there is are inner programs
#### Summary of Changes
only count metrics on outer calls
Fixes #


sakridge (approved):
lgtm

PR #16276, opened 2021-03-31
patches bug in banking stage where buffered packets are never retained

behzadnouri:
#### Problem
banking_stage::handle_forwarding is retaining buffered packets with
empty index, so nothing is held:
https://github.com/solana-labs/solana/blob/6f3926b64/core/src/banking_stage.rs#L520

#### Summary of Changes
* Changed to retain items with non-empty index.

### DIFF
--- a/core/src/banking_stage.rs
+++ b/core/src/banking_stage.rs
@@ -503,29 +503,28 @@ impl BankingStage {
         socket: &UdpSocket,
         hold: bool,
     ) {
-        if enable_forwarding {
-            let next_leader = poh_recorder
-                .lock()
-                .unwrap()
-                .leader_after_n_slots(FORWARD_TRANSACTIONS_TO_LEADER_AT_SLOT_OFFSET);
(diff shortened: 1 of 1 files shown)


core/src/banking_stage.rs
```diff
@@ -503,29 +503,28 @@ impl BankingStage {
...
-                            b.2 = true;
-                        }
-                    } else {
-                        buffered_packets.clear();
-                    }
-                })
-            })
+        if !enable_forwarding {
+            buffered_packets.clear(); // TODO: What if hold is true?
+            return;
```
sakridge, line 508: These will be votes that came through gossip. Maybe we should not clear here if hold==true. I think `cluster_info_vote_listener` would not re-send them. The votes might be somewhat old by the time the leader slot comes though and replaced by newer ones. Might have to enable some metrics on the real network to see if it's worth it to keep them.

behzadnouri: updated

sakridge (approved):

Issue #16077, opened 2021-03-23
rpc: introduce getInflationReward RPC call for total rewards


oJshua:
#### Problem
There is currently no efficient way to compute total earnings of a stake address without a bunch of separate RPC calls.

#### Proposed Solution
Wrap up this functionality into an RPC call, and use inputs of stake/vote account address and epoch. 

Here is how the CLI does it: https://github.com/solana-labs/solana/blob/4f82b897bc1756b6e0123c6c13762bb4b5eadea4/cli/src/stake.rs#L1653-L1754

PR #16278, opened 2021-03-31
Rpc: introduce get_inflation_reward rpc call

oJshua:
#### Problem
There is currently no way in the RPC to get the reward for a given epoch. 

#### Summary of Changes
- introduce rpc call for fetching reward for a single epoch
- if epoch is omitted, the most recent reward is fetched

Fixes https://github.com/solana-labs/solana/issues/16077


oJshua:
Screenshot from 2021-03-31 16-38-12

oJshua:
Testnet

```
curl --location --request POST 'http://35.230.78.5:8899' \
--header 'Content-Type: application/json' \
--data-raw '{
    "jsonrpc": "2.0",
    "id": 1,
    "method": "getInflationReward",
    "params": [
       "HMjrM2gMkxWPNVRKcqa98ptjgaMvaUxfyTKeZDws2KTE",
       170
    ]
}
```

```
{
    "jsonrpc": "2.0",
    "result": [
        {
            "amount": 14775098379,
            "apr": 219.88176982966198,
            "effectiveSlot": 70076256,
            "epoch": 174,
            "percentChange": 1.852936262609511,
            "postBalance": 812163474717
        },
        {
            "amount": 14529235940,
            "apr": 220.23578818237914,
            "effectiveSlot": 69644256,
            "epoch": 173,
            "percentChange": 1.8559195633346557,
            "postBalance": 797388376338
        },
        {
            "amount": 14342547065,
            "apr": 221.46330560436715,
            "effectiveSlot": 69212256,
            "epoch": 172,
            "percentChange": 1.866263811272757,
            "postBalance": 782859140398
        },
        {
            "amount": 14042378280,
            "apr": 220.8631866366892,
            "effectiveSlot": 68780260,
            "epoch": 171,
            "percentChange": 1.8612138095419943,
            "postBalance": 768516593333
        }
    ],
    "id": 1
}
```

core/src/rpc.rs
```diff
@@ -3107,6 +3218,20 @@ pub mod rpc_full {
             meta.get_stake_activation(&pubkey, config)
         }
 
+        fn get_inflation_reward(
```
mvines, line 3221: What do you think about just supporting a single epoch here?  Yes this means that the client would need to make multiple calls to fetch older epochs but I think that's better than a single very large RPC call

core/src/rpc.rs
```diff
@@ -382,6 +383,94 @@ impl JsonRpcRequestProcessor {
...
+                            rate_change * epochs_per_year
+                        });
+
+                        all_epoch_rewards.push(RpcEpochReward {
+                            epoch,
+                            effective_slot,
+                            amount: reward.lamports.abs() as u64,
+                            post_balance: reward.post_balance,
+                            percent_change: rate_change * 100.0,
+                            apr: apr.map(|r| r * 100.0),
```
mvines, line 458: The client can compute `percent_change` and `apr` if they want, let's keep it out of the API

client/src/rpc_response.rs
```diff
@@ -359,6 +359,17 @@ pub struct RpcPerfSample {
     pub sample_period_secs: u16,
 }
 
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct RpcEpochReward {
```
mvines, line 364: Maybe this name instead?
```suggestion
pub struct RpcInflationReward {
```

core/src/rpc.rs
oJshua: Sounds good!

core/src/rpc.rs
oJshua: Ah, maybe I misinterpreted original reason for adding this. Sure, that'll be a quick change!

CriesofCarrots: +1 for the single epoch. Default to most recent if epoch parameter not included?

oJshua: Good call. I'll default to the most recent.

core/src/rpc.rs
```diff
@@ -382,6 +382,60 @@ impl JsonRpcRequestProcessor {
...
+        pubkey: &Pubkey,
+        epoch: Epoch,
+    ) -> Result<Option<RpcInflationReward>> {
+        let epoch_schedule = self.get_epoch_schedule();
+        let first_available_block = self.get_first_available_block();
+
+        // Rewards for this epoch are found in the first confirmed block of the next epoch
+        let first_slot_in_epoch = epoch_schedule.get_first_slot_in_epoch(epoch + 1);
+        if first_slot_in_epoch < first_available_block {
+            return Err(Error::internal_error());
```
oJshua, line 396: This should probably be RpcCustomError::BlockNotAvailable?

oJshua: @CriesofCarrots any idea what error would make the most sense here?

core/src/rpc.rs
```diff
@@ -759,15 +813,12 @@ impl JsonRpcRequestProcessor {
...
-                        }));
+                        return bigtable_result
+                            .map_err(|_| RpcCustomError::BlockNotAvailable { slot }.into());
                     }
                 }
                 self.check_slot_cleaned_up(&result, slot)?;
-                return Ok(result.ok().map(|confirmed_block| {
-                    confirmed_block.configure(encoding, transaction_details, show_rewards)
-                }));
+                return result.map_err(|_| RpcCustomError::BlockNotAvailable { slot }.into());
```
CriesofCarrots, line 821: These and the following diff cause a behavior change in `getConfirmedBlock`, which I'd strongly prefer to avoid, due to strict partner expectations about this method.
Can we drop this new method, and just call `get_confirmed_block` from `get_inflation_reward`?

Something like:
```
self
    .get_confirmed_block(
        first_confirmed_block_in_epoch,
        Some(RpcEncodingConfigWrapper::Current(Some(RpcConfirmedBlockConfig::rewards_only()))),
    )
```
We could add a helper or a From impl to clean up the ugly Wrapper::Current part

core/src/rpc.rs
CriesofCarrots: It seems like the correct errors are either `LongTermStorageSlotSkipped`, if bigtable was queried by `first_available_block()`, or `SlotCleanedUp` if only blockstore. You'll probably have to use:
```
if self.bigtable_ledger_storage.is_some()
```

core/src/rpc.rs
oJshua: Yeah, factoring that out didn't feel quite right, plus I didn't like the name. Will revert

core/src/rpc.rs
```diff
@@ -382,6 +382,66 @@ impl JsonRpcRequestProcessor {
...
+        epoch: Option<Epoch>,
+    ) -> Result<Option<RpcInflationReward>> {
+        let epoch_schedule = self.get_epoch_schedule();
+        let first_available_block = self.get_first_available_block();
+        let epoch = epoch.unwrap_or_else(|| {
+            epoch_schedule.get_epoch(self.get_slot(Some(CommitmentConfig::finalized()))) - 1
+        });
+
+        // Rewards for this epoch are found in the first confirmed block of the next epoch
+        let first_slot_in_epoch = epoch_schedule.get_first_slot_in_epoch(epoch + 1);
```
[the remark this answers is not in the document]
oJshua: Oh cool, I was wondering if there was something like this.

core/src/rpc.rs
```diff
@@ -382,6 +382,77 @@ impl JsonRpcRequestProcessor {
...
+    pub fn get_inflation_reward(
+        &self,
+        pubkey: &Pubkey,
+        epoch: Option<Epoch>,
+    ) -> Result<Option<RpcInflationReward>> {
+        let epoch_schedule = self.get_epoch_schedule();
+        let first_available_block = self.get_first_available_block();
+        let epoch = epoch.unwrap_or_else(|| {
+            epoch_schedule
+                .get_epoch(self.get_slot(Some(CommitmentConfig::finalized())))
```
CriesofCarrots, line 394: A nit, but you could set `let commitment = Some(CommitmentConfig::finalized())` at the start of this method, since you pass it twice. Also would make commitment a little more obvious, if we ever have questions about results around epoch boundaries.

oJshua:
@CriesofCarrots was waiting for this to pass CI, but it might be good for another glance. Thanks much!

core/src/rpc.rs
```diff
@@ -382,6 +382,77 @@ impl JsonRpcRequestProcessor {
...
+            .get(0)
+            .ok_or(RpcCustomError::BlockNotAvailable {
+                slot: first_slot_in_epoch,
+            })?;
+
+        let first_confirmed_block = if let Ok(Some(first_confirmed_block)) = self
+            .get_confirmed_block(
+                first_confirmed_block_in_epoch,
+                Some(RpcConfirmedBlockConfig::rewards_only().into()),
+            ) {
```
CriesofCarrots, line 430: This turned out nice 🙂

CriesofCarrots (dismissed):
One nit, otherwise lgtm! Thanks for all the polish

mvines:
Can you please also add a section in the JSON RPC API docs, as well as support in RpcClient for the new method

oJshua:
@mvines on it.

mvines:
Oh as a follow-up PR, can you please use this new method in the cli as well.  To avoid backwards compatibility issues for now, the cli changes could only go into master and 1.6

CriesofCarrots:
> Oh as a follow-up PR, can you please use this new method in the cli as well. To avoid backwards compatibility issues for now, the cli changes could only go into master and 1.6

What do you think about limiting the cli to some number of previous epochs? We could add a separate command or arg to include complete rewards history. But it would be rather nice if `solana stake-account` could return quicker for users who just want to check the current status of an account.

mvines:
> > Oh as a follow-up PR, can you please use this new method in the cli as well. To avoid backwards compatibility issues for now, the cli changes could only go into master and 1.6
> 
> What do you think about limiting the cli to some number of previous epochs? We could add a separate command or arg to include complete rewards history. But it would be rather nice if `solana stake-account` could return quicker for users who just want to check the current status of an account.

Yes.  I hate so much how slow it is.  It worked great for the first couple epochs after pico-inflation was enabled though.

How about if on `solana stake-account` and `solana vote-account`, there
1. were no rewards displayed by default
2. `--rewards` by default shows the the latest epoch with an optional `EPOCH` arg to `--rewards` to display a previous epoch.  No ability  

Alternatively move rewards to `solana inflation-rewards`?   `solana inflation-rewards --epoch X address1 address2 ...`

core/src/rpc.rs
```diff
@@ -382,6 +382,74 @@ impl JsonRpcRequestProcessor {
         Ok(result)
     }
 
+    pub fn get_inflation_reward(
+        &self,
+        pubkey: &Pubkey,
```
mvines, line 387: Oh, we can easily make this `Vec<Pubkey>`.    This'll provide additional utility to the caller, and the work in this method is loading the block from BigTable/rocksdb after which we have all the rewards in hand already

CriesofCarrots:
> How about if on `solana stake-account` and `solana vote-account`, there
> 
>     1. were no rewards displayed by default
> 
>     2. `--rewards` by default shows the the latest epoch with an optional `EPOCH` arg to `--rewards` to display a previous epoch.  No ability
> 
> 
> Alternatively move rewards to `solana inflation-rewards`? `solana inflation-rewards --epoch X address1 address2 ...`

How about both? Too much? I can see different use-cases for each.

mvines:
> How about both? Too much? I can see different use-cases for each.

Yeah I think probably both too

core/src/rpc.rs
oJshua: Sure, I'll change that.

client/src/rpc_response.rs
```diff
@@ -362,6 +362,7 @@ pub struct RpcPerfSample {
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct RpcInflationReward {
+    pub address: String,
```
mvines, line 365: Is this needed?  Instead return `RpcInflationReward`s in the same order as the request?

oJshua: Yeah, we can do that instead.

CriesofCarrots: A queried address could be absent from `block.rewards`, right? I'm not sure an `RpcInflationReward { amount: 0, .. }` makes sense in these cases, since we don't know `post_balance`

mvines: `Vec<Option<RpcInflationReward>>` as a return type?

CriesofCarrots: > `Vec<Option<RpcInflationReward>>` as a return type?

Wfm!

oJshua: Yep, that's what I went with!

core/src/rpc.rs
```diff
@@ -382,6 +382,76 @@ impl JsonRpcRequestProcessor {
         Ok(result)
     }
 
+    pub fn get_inflation_reward(
+        &self,
+        pubkeys: Vec<Pubkey>,
```
mvines, line 387:
```suggestion
        addresses: Vec<Pubkey>,
```

core/src/rpc.rs
```diff
@@ -2161,6 +2231,14 @@ pub mod rpc_full {
             commitment: Option<CommitmentConfig>,
         ) -> Result<u64>;
 
+        #[rpc(meta, name = "getInflationReward")]
+        fn get_inflation_reward(
+            &self,
+            meta: Self::Metadata,
+            pubkey_strs: Vec<String>,
```
mvines, line 2238:
```suggestion
            address_strs: Vec<String>,
```

core/src/rpc.rs
```diff
@@ -382,6 +382,84 @@ impl JsonRpcRequestProcessor {
...
+        };
+
+        let mut reward_hash: HashMap<String, Reward> = HashMap::new();
+        for reward in first_confirmed_block
+            .rewards
+            .unwrap_or_default()
+            .into_iter()
+        {
+            reward_hash.insert(reward.clone().pubkey, reward);
+        }
```
CriesofCarrots, line 443: Can we filter this iterator to just the addresses of interest so we don't have to reallocate all the rewards?

core/src/rpc.rs
```diff
@@ -382,6 +382,89 @@ impl JsonRpcRequestProcessor {
...
+        let first_available_block = self.get_first_available_block();
+        let epoch = epoch.unwrap_or_else(|| {
+            epoch_schedule
+                .get_epoch(self.get_slot(commitment))
+                .saturating_sub(1)
+        });
+        let addresses: Vec<String> = addresses
+            .into_iter()
+            .map(|pubkey| pubkey.to_string())
+            .collect();
```
CriesofCarrots, line 401: Move down before `first_confirmed_block.rewards` handling?

core/src/rpc.rs
```diff
@@ -382,6 +382,89 @@ impl JsonRpcRequestProcessor {
...
+
+        let mut reward_hash: HashMap<String, Reward> = HashMap::new();
+        for reward in first_confirmed_block
+            .rewards
+            .unwrap_or_default()
+            .into_iter()
+            .filter(|reward| addresses.contains(&reward.pubkey))
+        {
+            reward_hash.insert(reward.clone().pubkey, reward);
+        }
```
CriesofCarrots, line 448: Can this block be rewritten as `into_iter().filter_map().collect()` for consistency?

core/src/rpc.rs
```diff
@@ -382,6 +382,89 @@ impl JsonRpcRequestProcessor {
...
+
+        let rewards = addresses
+            .iter()
+            .map(|address| {
+                if let Some(reward) = reward_hash.get(&address.to_string()) {
```
CriesofCarrots, line 453: I believe `to_string()` is redundant now

core/src/rpc.rs
oJshua: It can. Just learned how.

CriesofCarrots (dismissed):
:shipit:
Thanks for all the polish!

client/src/rpc_client.rs
```diff
@@ -764,6 +764,18 @@ impl RpcClient {
         self.send(RpcRequest::GetInflationRate, Value::Null)
     }
 
+    pub fn get_inflation_reward(
+        &self,
+        addresses: &[Pubkey],
+        epoch: Epoch,
```
mvines, line 770: The epoch is optional right?
```suggestion
        epoch: Option<Epoch>,
```

docs/src/developing/clients/jsonrpc-api.md
```diff
@@ -1468,6 +1469,56 @@ Result:
...
+- `<u64>` - (optional) An epoch for which the reward occurs (if omitted, the most recent epoch reward will be returned)
+
+#### Results
+
+The result field will be a JSON array with the following fields:
+
+- `epoch: <u64>`, epoch
+- `effective_slot: <u64>`, the slot in which the rewards are effective
+- `amount: <u64>`, reward amount in lamports
+- `post_balance: <u64>`, post balance in lamports
```
mvines, line 1487:
```suggestion
- `post_balance: <u64>`, post balance of the account in lamports
```

docs/src/developing/clients/jsonrpc-api.md
```diff
@@ -1468,6 +1469,56 @@ Result:
 {"jsonrpc":"2.0","result":{"epoch":100,"foundation":0.001,"total":0.149,"validator":0.148},"id":1}
 ```
 
+### getInflationReward
+
+Returns the inflation reward for a list of addresses and an epoch
+
+#### Parameters:
+- `<array>` - An array of addresses to query, as base-58 encoded strings
+- `<u64>` - (optional) An epoch for which the reward occurs (if omitted, the most recent epoch reward will be returned)
```
mvines, line 1478:
```suggestion
- `<u64>` - (optional) An epoch for which the reward occurs.  If omitted, the previous epoch will be used
```

docs/src/developing/clients/jsonrpc-api.md
```diff
@@ -1468,6 +1469,56 @@ Result:
 {"jsonrpc":"2.0","result":{"epoch":100,"foundation":0.001,"total":0.149,"validator":0.148},"id":1}
 ```
 
+### getInflationReward
+
+Returns the inflation reward for a list of addresses and an epoch
```
mvines, line 1474:
```suggestion
Returns the inflation reward for a list of addresses for an epoch
```

mvines (approved):
lgtm, just a couple final nits

client/src/rpc_client.rs
```diff
@@ -764,6 +764,18 @@ impl RpcClient {
         self.send(RpcRequest::GetInflationRate, Value::Null)
     }
 
+    pub fn get_inflation_reward(
+        &self,
+        addresses: &[Pubkey],
+        config: Option<RpcEpochConfig>,
```
CriesofCarrots, line 770: Wdyt about keeping this input `epoch: Option<Epoch>`? Can construct the config object using `self.commitment_config`, like `get_stake_activation()` above

core/src/rpc.rs
```diff
@@ -382,6 +382,91 @@ impl JsonRpcRequestProcessor {
         Ok(result)
     }
 
+    pub fn get_inflation_reward(
+        &self,
+        addresses: Vec<Pubkey>,
+        config: Option<RpcEpochConfig>,
+    ) -> Result<Vec<Option<RpcInflationReward>>> {
+        let config = config.unwrap_or_default();
+        let commitment = Some(config.commitment.unwrap_or(CommitmentConfig::finalized()));
```
CriesofCarrots, line 391: I think we can delete this line now, and just pass `config.commitment` where `commitment` is used below

core/src/rpc.rs
```diff
@@ -382,6 +382,91 @@ impl JsonRpcRequestProcessor {
...
+            .into_iter()
+            .map(|pubkey| pubkey.to_string())
+            .collect();
+
+        let reward_hash: HashMap<String, Reward> = first_confirmed_block
+            .rewards
+            .unwrap_or_default()
+            .into_iter()
+            .filter_map(|reward| {
+                (reward.reward_type? != RewardType::Rent && addresses.contains(&reward.pubkey))
```
CriesofCarrots, line 447: Just Voting and Staking? https://discord.com/channels/428295358100013066/476811830145318912/829071641841631244
Might make `match reward.reward_type?` cleaner

docs/src/developing/clients/jsonrpc-api.md
```diff
@@ -1468,6 +1469,56 @@ Result:
 {"jsonrpc":"2.0","result":{"epoch":100,"foundation":0.001,"total":0.149,"validator":0.148},"id":1}
 ```
 
+### getInflationReward
+
+Returns the inflation reward for a list of addresses for an epoch
+
+#### Parameters:
+- `<array>` - An array of addresses to query, as base-58 encoded strings
+- `<u64>` - (optional) An epoch for which the reward occurs. If omitted, the previous epoch will be used
```
CriesofCarrots, line 1478: Needs update, can refer to `getStakeActivation` for format

core/src/rpc.rs
CriesofCarrots: This will 🔪 the clippy error too :)

client/src/rpc_client.rs
oJshua: Ah, yes. I should have looked closer at this. Thank you.

PR #16281, opened 2021-03-31
Rpc: enable getConfirmedSignaturesForAddress2 to return confirmed (not yet finalized) data

CriesofCarrots:
#### Problem
As in #16142, `getConfirmedSignaturesForAddress2` does not included confirmed but not-yet-finalized blocks in the response list.

#### Summary of Changes
- Update Blockstore api to return data for any address signature newer than provided slot, dropping the rooted requirement
- Expose confirmed data in `getConfirmedSignaturesForAddress2` via the commitment parameter; default remains finalized.
  - Look up block time for not-yet-finalized signatures
- Return confirmed data in `solana transaction-history <address>`
- Add confirmation_status to `getConfirmedSignaturesForAddress2` return type to display real statuses in `solana transaction-history`

### DIFF
--- a/cli/src/cluster_query.rs
+++ b/cli/src/cluster_query.rs
@@ -1843,9 +1843,13 @@ pub fn process_transaction_history(
                     Some(block_time) =>
                         format!("timestamp={} ", unix_timestamp_to_string(block_time)),
                 },
-                match result.err {
-                    None => "Confirmed".to_string(),
-                    Some(err) => format!("Failed: {:?}", err),
+                if let Some(err) = result.err {
+                    format!("Failed: {:?}", err)
+                } else {
+                    match result.confirmation_status {
+                        None => "Finalized".to_string(),
+                        Some(status) => format!("{:?}", status),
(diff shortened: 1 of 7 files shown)


cli/src/cluster_query.rs
```diff
@@ -1842,9 +1843,13 @@ pub fn process_transaction_history(
...
                         format!("timestamp={} ", unix_timestamp_to_string(block_time)),
                 },
-                match result.err {
-                    None => "Confirmed".to_string(),
-                    Some(err) => format!("Failed: {:?}", err),
+                if let Some(err) = result.err {
+                    format!("Failed: {:?}", err)
+                } else {
+                    match result.confirmation_status {
+                        None => "Finalized".to_string(),
```
CriesofCarrots, line 1850: The reason for this None handling: `confirmation_status.is_none()` should only be returned from nodes running older code, which can only return finalized transaction signatures

mvines (approved):
looks great, as always :)

PR #16304, opened 2021-04-01
reduce locking on get_rooted_entries

jeffwashington:
#### Problem
get_rooted_entries is called during clean and perhaps other times. The current implementation acquires and releases the lock with each check.
#### Summary of Changes
Get the read lock and check all slots as tightly as possible.
Fixes #

### DIFF
--- a/runtime/src/accounts_index.rs
+++ b/runtime/src/accounts_index.rs
@@ -685,9 +685,11 @@ impl<T: 'static + Clone + IsCached + ZeroLamport> AccountsIndex<T> {
     }
 
     pub fn get_rooted_entries(&self, slice: SlotSlice<T>, max: Option<Slot>) -> SlotList<T> {
+        let max = max.unwrap_or(Slot::MAX);
+        let lock = &self.roots_tracker.read().unwrap().roots;
         slice
             .iter()
-            .filter(|(slot, _)| self.is_root(*slot) && max.map_or(true, |max| *slot <= max))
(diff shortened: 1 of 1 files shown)


runtime/src/accounts_index.rs
```diff
@@ -685,9 +685,11 @@ impl<T: 'static + Clone + IsCached + ZeroLamport> AccountsIndex<T> {
     }
 
     pub fn get_rooted_entries(&self, slice: SlotSlice<T>, max: Option<Slot>) -> SlotList<T> {
+        let max = max.unwrap_or(Slot::MAX);
+        let lock = &self.roots_tracker.read().unwrap().roots;
         slice
             .iter()
-            .filter(|(slot, _)| self.is_root(*slot) && max.map_or(true, |max| *slot <= max))
+            .filter(|(slot, _)| *slot <= max && lock.contains(slot))
```
sakridge, line 692: could also check `slot <= roots_tracker.max_root` before checking the `.contains()`. Not sure if that would be faster or not.

sakridge (approved):
Logic seems sound. Evidence of improvement would be nice.

runtime/src/accounts_index.rs
jeffwashington: contains will shortly be low constant time, so I think the change wouldn't help and would (likely very slightly) slow down the case when contains returns true.

jeffwashington:
[perf metrics](https://github.com/solana-labs/solana/pull/16306#issuecomment-813447911)

Issue #14523, opened 2021-01-10
BPF program serialization/deserialization may no longer be necessary


jackcmay:
Now that programs no longer support returning errors "gracefully" (https://github.com/solana-labs/solana/pull/14500), the way data flows to and through a cross-program invocation call chain also changes and may no longer require serialization or extra copies.

There is no longer any roll-back or intermediate discarding of data.  Account data now flows through the call chain linearly from one program to the next, either to the end of the chain upon success or up until an error where the data is discarded.  Because of this, rather than copying data in and out of each program's environment, the data can be passed (and translated) via pointers.  Each program in the call chain incrementally operates on the data, and the runtime verifies and snapshots in-between and at the end.  

Doing so provides the following advantages:
- No serialization or deserialization of program parameters.  Instead, the parameters undergo a deep pointer translation into and out of the program's virtual address space.
- No extra copying is required to update the caller's state between calls since the called program operates directly on the shared data.  Translating the data types into and out of the program is a fast operation.
- Could support a single entrypoint between native and BPF programs
- Operating on shared data potentially makes it easier to support account data reallocations
- In Rust, account data would be more naturally handled and understood by the compiler.  Right now we have an issue where account data is modified by the runtime on data values the program thinks is immutable (account data slice size for example).  Depending on what assumptions the compiler makes some updated values may be stale in registers upon return.

But comes with the following challenges:
- Rust does not have a stable ABI so any types translated should be done on stable ABI types (`repr(C)`).  We already have this issue when passing `Instruction` and `AccountIInfo` via `Invoke`.  New `repr(C)` types should be used for all data passed into and out of the programs.  This means the current native entrypoint (using `KeyedAccounts` is not well suited.
- Reallocating account data from the program space may be possible but would require coordination between the program's memory allocator and the native allocator in order to translate and route allocation requests to the native allocator.
- Potentially lots of churn in the runtime and sdk, for example, if we replace program parameters (native and BPF) with something that is `repr(C)` we should probably make that change all the way down to `accounts` so that we don't incur any extra copies of the account data then are absolutely necessary.

Overall I think this proposal does a lot for both performance, stability, and clarity of the data being passed to and between programs.

Lichtso:
Here is the plan:
[...]

PR #16306, opened 2021-04-01
reduce lock contention on latest_root

jeffwashington:
#### Problem
latest_root is called with a lock on the index for every account load. This code could be tighter. We could get the read lock once and keep it if we are able to reduce the amount of time we need to hold it open. We could also check ancestors, which is a smaller set of slots first, before checking the more expensive, 400k roots hashset.

#### Summary of Changes
rework the loop.
Fixes #

### DIFF
--- a/runtime/src/accounts_index.rs
+++ b/runtime/src/accounts_index.rs
@@ -750,30 +750,35 @@ impl<T: 'static + Clone + IsCached + ZeroLamport> AccountsIndex<T> {
     ) -> Option<usize> {
         let mut current_max = 0;
         let mut rv = None;
-        for (i, (slot, _t)) in slice.iter().rev().enumerate() {
-            if *slot >= current_max && self.is_ancestor_or_root(*slot, ancestors, max_root) {
-                rv = Some((slice.len() - 1) - i);
-                current_max = *slot;
+        if let Some(ancestors) = ancestors {
+            if !ancestors.is_empty() {
+                for (i, (slot, _t)) in slice.iter().rev().enumerate() {
+                    if *slot >= current_max && ancestors.contains_key(slot) {
+                        rv = Some(i);
+                        current_max = *slot;
+                    }
+                }
             }
         }
 
-        rv
-    }
+        let max_root = max_root.unwrap_or(Slot::MAX);
+        let mut tracker = None;
 
-    // Checks that the given slot is either:
-    // 1) in the `ancestors` set
-    // 2) or is a root
-    fn is_ancestor_or_root(
-        &self,
-        slot: Slot,
-        ancestors: Option<&Ancestors>,
-        max_root: Option<Slot>,
-    ) -> bool {
-        ancestors.map_or(false, |ancestors| ancestors.contains_key(&slot)) ||
-        // If the slot is a root, it must be less than the maximum root specified. This
-        // allows scans on non-rooted slots to specify and read data from
-        // ancestors > max_root, while not seeing rooted data update during the scan
-        (max_root.map_or(true, |max_root| slot <= max_root) && (self.is_root(slot)))
+        for (i, (slot, _t)) in slice.iter().rev().enumerate() {
+            if *slot >= current_max && *slot <= max_root {
+                let lock = match tracker {
+                    Some(inner) => inner,
+                    None => self.roots_tracker.read().unwrap(),
+                };
+                if lock.roots.contains(&slot) {
+                    rv = Some(i);
+                    current_max = *slot;
+                }
+                tracker = Some(lock);
+            }
+        }
+
+        rv.map(|index| slice.len() - 1 - index)
     }
 
     /// Get an account


runtime/src/accounts_index.rs
```diff
@@ -750,30 +750,35 @@ impl<T: 'static + Clone + IsCached + ZeroLamport> AccountsIndex<T> {
...
         let mut current_max = 0;
         let mut rv = None;
-        for (i, (slot, _t)) in slice.iter().rev().enumerate() {
-            if *slot >= current_max && self.is_ancestor_or_root(*slot, ancestors, max_root) {
-                rv = Some((slice.len() - 1) - i);
-                current_max = *slot;
+        if let Some(ancestors) = ancestors {
+            if !ancestors.is_empty() {
+                for (i, (slot, _t)) in slice.iter().rev().enumerate() {
+                    if *slot >= current_max && ancestors.contains_key(slot) {
```
jeffwashington, line 756: @sakridge isn't *slot > current_max ok here? What conditions would cause = to be useful? If we already found a root or ancestor at a slot = to the current slot we're looking for, isn't that good enough? Do we expect multiple items in the slice to have the same slot #?

sakridge: `slot > current_max` should be ok as well. I don't believe there should be duplicates in the list.

sakridge:
Believe it does the right thing. It would be nice to see how fast it is and in what scenarios, like if we have a large ancestors list vs. small and/or a large slot_list set vs. 1 or 2. Also, how contended the roots_tracker lock is. Because I think we just write to that once per slot to update the root, although we do probably read it a lot from banking/clean/rpc

jeffwashington:
I measure read().unwrap().contains to be 5x slower than already_unlocked.contains().

[Replacing HashSet with BitVec](https://github.com/solana-labs/solana/pull/16310) AND getting re-using the read lock measures 10x faster.

ryoqun:
@jeffwashington (cc: @sakridge ) oh, you're tackling lock contention? I don't have much context here, but I generally tried to fix all the slow lock contentions by superior third-party lock library than `std` ones: https://github.com/solana-labs/solana/pull/14578 `std` ones are kind of slow. That pr got pushed out from my work queue sadly, happy to help if you're interested in taking over. :) I wonder  we could gain a cheap win from just the library switch.

jeffwashington:
One line of thinking is that when we use a BitVec instead of HashSet, then .contains() will be small constant time. Ancestors is lock free. So, any call that results in us finding the slot in the ancestors list will mean we never have to get the read lock to the roots. If we have a very long slot list, we will check the ancestors list for each slot. But, with that check being lock free and small constant time, this should be negligible. Then, when we lock the roots, we keep it locked and we again have small constant time checks for each item in the slot list. With BitVec, the # of items in the Ancestors list or Roots list is irrelevant. So, worst case due to this change is long slot list, > 0 ancestors, but no matches in ancestors. We effectively waste time checking ancestors for each slot. Since this is lock free, low constant time per check, this seems negligible.

jeffwashington:
> @jeffwashington (cc: @sakridge ) oh, you're tackling lock contention? I don't have much context here, but I generally tried to fix all the slow lock contentions by superior third-party lock library than `std` ones: #14578 `std` ones are kind of slow. That pr got pushed out from my work queue sadly, happy to help if you're interested in taking over. :) I wonder we could gain a cheap win from just the library switch.

I'm finding things that make poh, hashing, or replay slower than it needs to be ;-) I ran into these 2 code paths. I also added a read only accounts cache last week. With that cache populated, account load time became dominated by this code checking for is_root.

jeffwashington:
oh, and the longer this code path takes, the longer we hold the index locks open, as far as I can tell. And, I believe we do have index lock contention.

jeffwashington:
> @jeffwashington (cc: @sakridge ) oh, you're tackling lock contention? I don't have much context here, but I generally tried to fix all the slow lock contentions by superior third-party lock library than `std` ones: #14578 `std` ones are kind of slow. That pr got pushed out from my work queue sadly, happy to help if you're interested in taking over. :) I wonder we could gain a cheap win from just the library switch.

I tagged myself on that pr. I will be glad to add that to my queue.

ryoqun:
> I ran into these 2 code paths. I also added a read only accounts cache last week. With that cache populated, account load time became dominated by this code checking for is_root.

sounds cool with pretty finished progress. so, you aren't seeing large portion of memcopies are gone now?: https://github.com/solana-labs/solana/issues/14523#issuecomment-765800757

runtime/src/accounts_index.rs
jeffwashington: I worried about this... A problem is a test that uses slot 0. I thought this might be a corner case with unsigned and tests. -1 is such a convenient number but shockingly missing from unsigned types. ;-)

sakridge: I see, yea it should find slot=0 if it exists.

runtime/src/accounts_index.rs
```diff
@@ -750,30 +750,35 @@ impl<T: 'static + Clone + IsCached + ZeroLamport> AccountsIndex<T> {
...
+                        rv = Some(i);
+                        current_max = *slot;
+                    }
+                }
             }
         }
 
-        rv
-    }
+        let max_root = max_root.unwrap_or(Slot::MAX);
```
sakridge, line 764: I think we could return early here if `rv.is_some()`

sakridge (approved):

runtime/src/accounts_index.rs
jeffwashington: https://github.com/solana-labs/solana/pull/16378

PR #16310, opened 2021-04-01
Replace RootsTracker HashSet

jeffwashington:
#### Problem
with a readonly accounts cache (meaning the store has previously been loaded), large cpu time is spent looking in 400k hashsets millions of times. We could do better.
#### Summary of Changes
Replace RootsTracker HashSet with BitVec based lookup.
Fixes #

### DIFF
--- a/program-test/tests/warp.rs
+++ b/program-test/tests/warp.rs
@@ -55,7 +55,7 @@ async fn clock_sysvar_updated_from_warp() {
     );
 
     let mut context = program_test.start_with_context().await;
-    let expected_slot = 5_000_000;
+    let expected_slot = 100_000;
     let instruction = Instruction::new_with_bincode(
         program_id,
         &expected_slot,
--- a/runtime/src/accounts_index.rs
+++ b/runtime/src/accounts_index.rs
@@ -200,13 +200,9 @@ impl RollingBitField {
         }
     }
 
-    fn get_address(&self, key: &u64) -> u64 {
-        Self::calc_address(self.max_width, key)
-    }
-
     // find the array index
-    fn calc_address(max_width: u64, key: &u64) -> u64 {
-        key % max_width
+    fn get_address(&self, key: &u64) -> u64 {
+        key % self.max_width
     }
 
     fn check_range(&self, key: u64) {
@@ -240,7 +236,7 @@ impl RollingBitField {
         let address = self.get_address(key);
         let value = self.bits.get(address);
         if value {
-            self.count -= 1; // this will underflow if it tries to go below 0. That is like an assert.
+            self.count -= 1;
             self.bits.set(address, false);
             self.purge(key);
         }
@@ -282,8 +278,8 @@ impl RollingBitField {
     pub fn get_all(&self) -> Vec<u64> {
         let mut all = Vec::with_capacity(self.count);
         for key in self.min..self.max {
-            if self.contains(&(key as u64)) {
-                all.push(key as u64);
+            if self.contains(&key) {
+                all.push(key);
                 if all.len() == self.count {
                     break;
                 }
--- a/runtime/src/accounts_index.rs
+++ b/runtime/src/accounts_index.rs
@@ -246,6 +246,7 @@ impl RollingBitField {
         }
     }
 
+    // after removing 'key' where 'key' = min, make min the correct new min value
     fn purge(&mut self, key: &u64) {
         if key == &self.min && self.count > 0 {
             let start = self.min;
@@ -261,10 +262,8 @@ impl RollingBitField {
     pub fn contains(&self, key: &u64) -> bool {
         let address = self.get_address(key);
         let result = self.bits.get(address);
-        if result && (self.count == 0 || key < &self.min || key >= &self.max) {
-            return false;
-        }
-        result
+        // a contains call outside min and max is allowed. The answer will be false.
+        result && (key >= &self.min && key < &self.max)
     }
 
     pub fn len(&self) -> usize {
@@ -1350,7 +1349,7 @@ pub mod tests {
         let mut bitfield = RollingBitField::new(2097152);
         let mut hash = HashSet::new();
 
-        let min = 101_000.clone();
+        let min = 101_000;
         let width = 400_000;
         let dead = 19;
 
--- a/runtime/src/accounts_index.rs
+++ b/runtime/src/accounts_index.rs
@@ -1370,9 +1370,8 @@ pub mod tests {
             if slot % dead == 0 {
                 continue;
             }
-            let s2 = slot.clone();
-            hash.insert(s2);
-            bitfield.insert(s2);
+            hash.insert(slot);
+            bitfield.insert(slot);
         }
 
         let max = slot + 1;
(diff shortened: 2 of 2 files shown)


runtime/src/accounts_index.rs
```diff
@@ -174,13 +174,157 @@ impl<T: 'static + Clone + IsCached> WriteAccountMapEntry<T> {
 }
 
 #[derive(Debug, Default)]
+pub struct RollingBitField {
+    max_width: usize,
+    min: u64,
+    max: u64, // exclusive
+    bits: Vec<u64>,
```
sakridge, line 181: We use the `BitVec` crate in other places, maybe this could just use that?

runtime/src/accounts_index.rs
```diff
@@ -174,13 +174,157 @@ impl<T: 'static + Clone + IsCached> WriteAccountMapEntry<T> {
...
+    fn calc_address(max_width: usize, key: u64) -> RollingBitFieldAddress {
+        let key = key as usize;
+        let bits_in_u64 = 64;
+        let array_index = (key % max_width) / bits_in_u64;
+        let bit_index = key % bits_in_u64;
+        let mask: u64 = 1u64 << bit_index;
+        RollingBitFieldAddress { array_index, mask }
+    }
+
+    pub fn insert(&mut self, key: &u64) {
```
sakridge, line 224: any reason not to just take `u64` directly?

jeffwashington: to match the api that I am replacing. hashset uses &key

jeffwashington: I agree it is ugly. This was the low impact change to measure, try it out, etc.

runtime/src/accounts_index.rs
jeffwashington: looks like that handles the low level math and mapping, Hopefully it is fast. I have that refit to use BitVec.

runtime/src/accounts_index.rs
jeffwashington: oops - insert is not a ref

jeffwashington:
I measure the contains call on a set of 400k items to be ~7x faster using a BitVec. This matches the data size of what we expect RootsTracker to hold. I measure a custom implemented bit vec to be slightly slower, which is great news. BItVec is faster than a simple custom implementation.

jeffwashington:
with 33 or 2 items (like ancestors) and a size of 2M, I get ~4x faster for BitVec implementation.

program-test/tests/warp.rs
```diff
@@ -55,7 +55,7 @@ async fn clock_sysvar_updated_from_warp() {
     );
 
     let mut context = program_test.start_with_context().await;
-    let expected_slot = 5_000_000;
```
jeffwashington, line 58: discussed this with @joncinque

joncinque (approved):
Thanks for confirming

runtime/src/accounts_index.rs
```diff
@@ -173,14 +173,154 @@ impl<T: 'static + Clone + IsCached> WriteAccountMapEntry<T> {
...
+            self.count += 1;
+        }
+    }
+
+    pub fn remove(&mut self, key: &u64) {
+        self.check_range(*key);
+        let address = self.get_address(key);
+        let value = self.bits.get(address);
+        if value {
+            self.count -= 1; // this will underflow if it tries to go below 0. That is like an assert.
```
sakridge, line 243: In release mode it would just wrap, is that ok behavior?

runtime/src/accounts_index.rs
```diff
@@ -173,14 +173,154 @@ impl<T: 'static + Clone + IsCached> WriteAccountMapEntry<T> {
...
+        }
+    }
+
+    fn get_address(&self, key: &u64) -> u64 {
+        Self::calc_address(self.max_width, key)
+    }
+
+    // find the array index
+    fn calc_address(max_width: u64, key: &u64) -> u64 {
+        key % max_width
```
sakridge, line 209: Function seems a bit superfluous

runtime/src/accounts_index.rs
```diff
@@ -173,14 +173,154 @@ impl<T: 'static + Clone + IsCached> WriteAccountMapEntry<T> {
...
+    pub fn clear(&mut self) {
+        let mut n = Self::new(self.max_width);
+        std::mem::swap(&mut n, self);
+    }
+
+    pub fn get_all(&self) -> Vec<u64> {
+        let mut all = Vec::with_capacity(self.count);
+        for key in self.min..self.max {
+            if self.contains(&(key as u64)) {
+                all.push(key as u64);
```
sakridge, line 286: does it need the casts? isn't key u64 here?

Also, seems like it could just get away with `self.bits.get(key)` here, it should be able to not do the range checks and the mod here.

runtime/src/accounts_index.rs
```diff
@@ -173,14 +173,154 @@ impl<T: 'static + Clone + IsCached> WriteAccountMapEntry<T> {
...
+            self.purge(key);
+        }
+    }
+
+    // after removing 'key' where 'key' = min, make min the correct new min value
+    fn purge(&mut self, key: &u64) {
+        if key == &self.min && self.count > 0 {
+            let start = self.min;
+            for key in start..self.max {
+                if self.bits.get(key) {
```
sakridge, line 254: Can it use this:
https://docs.rs/bitvec/0.22.1/bitvec/slice/struct.BitSlice.html#method.leading_zeros

runtime/src/accounts_index.rs
jeffwashington: the comment was left over. The function had already been changed to only decrement count if the value was previously set.
I did not realize that the assertion I had seen on wrapping was only present in debug builds. That will be the first new thing I learned today.

runtime/src/accounts_index.rs
jeffwashington: yep, was left over from when it was more complicated and the static version was called prior to construction

runtime/src/accounts_index.rs
jeffwashington: It does not need the casts. Hmm. I assumed clippy would fuss about that if it was happening.

runtime/src/accounts_index.rs
jeffwashington: These docs are from the [bitvec crate](https://crates.io/crates/bitvec). We use [this crate](https://tov.github.io/bv-rs/bv/struct.BitVec.html) elsewhere in the repo. bv does not appear to support leading_zeros.

runtime/src/accounts_index.rs
jeffwashington: I reworked the code to get rid of the range checks (and the early exit). I think some version of the mod or 2 loops will be required to handle the case where the range wraps over the end of the bitvec size. To keep the code easy to read, I kept get_all from accessing bitvec directly.

sakridge (approved):
lgtm

runtime/src/accounts_index.rs
sakridge: I see. In that case https://tov.github.io/bv-rs/bv/trait.Bits.html#method.get_block combined with ffs might be more efficient.

PR #16327, opened 2021-04-02
Cleanup use

jackcmay:
#### Summary of Changes

- Consolidate use in program-test
- Use default for MockInvokeContext to avoid requiring changing instantiation if new members are added

Fixes #

### DIFF
--- a/program-test/src/lib.rs
+++ b/program-test/src/lib.rs
@@ -7,20 +7,6 @@ use {
     log::*,
     solana_banks_client::start_client,
     solana_banks_server::banks_server::start_local_server,
-    solana_program::{
-        account_info::AccountInfo,
-        entrypoint::ProgramResult,
-        fee_calculator::{FeeCalculator, FeeRateGovernor},
-        hash::Hash,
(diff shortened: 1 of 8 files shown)


program-test/src/lib.rs
```diff
@@ -7,20 +7,6 @@ use {
     log::*,
     solana_banks_client::start_client,
     solana_banks_server::banks_server::start_local_server,
-    solana_program::{
```
mvines, line 10: FTR: The reason I originally explicitly used `solana_program` here instead of the `solana_sdk` re-imports is that I was envisioning a future release where we rip out the `solana_sdk` re-imports entirely.  Maybe that's a pipe dream though

jackcmay: Yeah, I figured it was something like that, it was a nit pain point when I would add new exports and then find they were already added in the other.  Also, because we aren't at the point where we can use `solana_program` it was confusing why they were split and if there was some larger reason to include both.  We can easily switch to `solana_program` later if/when we are able to.

PR #16332, opened 2021-04-03
Add faucet per-ip rate limiter

CriesofCarrots:
#### Problem
The solana-faucet ip_cache was never used for anything, but a per-IP rate limit would be pretty useful for testnet.

#### Summary of Changes
- Remove dead code
- Add per-ip rate limiter with `--ip-rate-limit` parameter; default behavior is no ip rate limiting. Current SOL-based limits (per single request and total requests in time slice) remain unchanged.


faucet/src/bin/faucet.rs
```diff
@@ -52,6 +54,13 @@ async fn main() {
                 .takes_value(true)
                 .help("Request limit for a single request, in SOL"),
         )
+        .arg(
```
mvines, line 57: Couldn't we just plug ip addresses into the existing  --cap/--slice parameters.  Currently these are global but in my mind, earlier in the week, these parameters were per IP

CriesofCarrots: You mean replace the global altogether?

mvines: Yes

PR #16340, opened 2021-04-03
Wait for 90 percent of stake before starting

sakridge:
#### Problem

80% is problematic because it leaves very little margin for a ~40/40 split on two forks when starting the network. If the network cannot obtain at least 38% on one of the forks, then one 35% partition could stall waiting for the other to reach the desired level.

#### Summary of Changes

Increase wait-for-supermajority to 90% stake to give more margin on restart.

Fixes #


mvines:
(adding v1.5 as well, so we keep the restart procedure consistent across testnet and mainnet)

Issue #16339, opened 2021-04-03
rpc: Log subscriptions should only receive notifications from processed transactions


jstarry:
#### Problem
Log subscriptions are sent notifications for non-processing related transaction errors which are not valuable to the client. This commonly includes "AccountInUse" errors because even transactions which don't obtain account locks are being iterated over when sending log notifications. This is causing unexpected error types as reported in this issue: https://github.com/solana-labs/solana-web3.js/issues/1019

#### Proposed Solution
Only notify log events for transactions that were processed

jstarry:
cc @mvines

PR #16349, opened 2021-04-04
Remove unprocessed transactions from log notifications

jstarry:
#### Problem
Log subscriptions are sent notifications for non-processing related transaction errors which are not valuable to the client. This commonly includes "AccountInUse" errors because even transactions which don't obtain account locks are being iterated over when sending log notifications. This is causing unexpected error types as reported in this issue: solana-labs/solana-web3.js#1019

#### Summary of Changes
Only notify log events for transactions that were processed

Fixes https://github.com/solana-labs/solana/issues/16339

### DIFF
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -3000,7 +3000,10 @@ impl Bank {
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


mvines (approved):
lgtm, v1.6 and v1.5 backport?

PR #16357, opened 2021-04-04
Fixup AncestorIterator method

CriesofCarrots:
#### Problem
Unnecessary AncestorIterator visits are slowing down `getConfirmedSignaturesForAddress2` rpc call.

#### Summary of Changes
Fixup iterator method to minimize ancestor visits

### DIFF
--- a/ledger/src/blockstore.rs
+++ b/ledger/src/blockstore.rs
@@ -2055,7 +2055,7 @@ impl Blockstore {
         );
         let confirmed_unrooted_slots: Vec<_> =
             AncestorIterator::new_inclusive(highest_confirmed_slot, self)
-                .filter(|&slot| slot > self.last_root())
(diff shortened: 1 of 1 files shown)


ledger/src/blockstore.rs
```diff
@@ -2055,7 +2055,7 @@ impl Blockstore {
         );
         let confirmed_unrooted_slots: Vec<_> =
             AncestorIterator::new_inclusive(highest_confirmed_slot, self)
-                .filter(|&slot| slot > self.last_root())
+                .take_while(|&slot| slot > self.last_root())
```
sakridge, line 2058: hoisting `self.last_root()` outside the loop might be better even

CriesofCarrots: Good call, will follow with that change; thanks!

PR #16375, opened 2021-04-05
remove cluster_type from hash_stored_account

jeffwashington:
#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -2878,33 +2878,20 @@ impl AccountsDb {
     pub fn hash_stored_account(
         slot: Slot,
         account: &StoredAccountMeta,
-        cluster_type: &ClusterType,
+        _cluster_type: &ClusterType,
     ) -> Hash {
-        let include_owner = Self::include_owner(cluster_type, slot);
+        let include_owner = true;
 
-        if slot > Self::get_blake3_slot(cluster_type) {
-            Self::blake3_hash_account_data(
-                slot,
-                account.account_meta.lamports,
-                &account.account_meta.owner,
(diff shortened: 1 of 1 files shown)


runtime/src/accounts_db.rs
```diff
@@ -2878,33 +2878,20 @@ impl AccountsDb {
     pub fn hash_stored_account(
         slot: Slot,
         account: &StoredAccountMeta,
-        cluster_type: &ClusterType,
+        _cluster_type: &ClusterType,
```
ryoqun, line 2881: just making sure: you can stop passing this altogether from the caller (assuming we don't have any immediate plan to change the hashing.).

runtime/src/accounts_db.rs
```diff
@@ -2878,33 +2878,20 @@ impl AccountsDb {
...
-        }
+        Self::blake3_hash_account_data(
+            slot,
+            account.account_meta.lamports,
+            &account.account_meta.owner,
+            account.account_meta.executable,
+            account.account_meta.rent_epoch,
+            account.data,
+            &account.meta.pubkey,
+            include_owner,
```
ryoqun, line 2893: also, you can remove this flag. blake3 and `include_owner` thing even predate the feature activation @mvines brought about. blame @carllin and @ryoqun for cleaning them, respectively. xD

runtime/src/accounts_db.rs
jeffwashington: yep, phase 2. ;-) roughing out this is what you meant.

jeffwashington:
taken over by @brooksprumo

PR #16377, opened 2021-04-05
Adjust tokio version to just "1"

mvines:
version="1" seems to be more idiomatic in the tokio subcommunity


CriesofCarrots:
Out of curiosity, did this cause conflicts somewhere?

mvines:
> Out of curiosity, did this cause conflicts somewhere?

No, I just continue to see `tokio = "1"` everywhere in tokio land so wanted to align

PR #16378, opened 2021-04-05
short circuit largest_slot

jeffwashington:
#### Problem
avoid work when finding slot in ancestors
#### Summary of Changes
^
Fixes #

### DIFF
--- a/runtime/src/accounts_index.rs
+++ b/runtime/src/accounts_index.rs
@@ -763,20 +763,23 @@ impl<T: 'static + Clone + IsCached + ZeroLamport> AccountsIndex<T> {
             }
         }
 
-        let max_root = max_root.unwrap_or(Slot::MAX);
-        let mut tracker = None;
+        // if we find a slot in ancestors, we don't need to look in roots at all
+        if rv.is_none() {
+            let max_root = max_root.unwrap_or(Slot::MAX);
+            let mut tracker = None;
 
-        for (i, (slot, _t)) in slice.iter().rev().enumerate() {
-            if *slot >= current_max && *slot <= max_root {
-                let lock = match tracker {
-                    Some(inner) => inner,
-                    None => self.roots_tracker.read().unwrap(),
-                };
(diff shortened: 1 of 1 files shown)


jeffwashington:
with this short circuit, [it doesn't help to change >= to >](https://github.com/solana-labs/solana/pull/16374)

jeffwashington:
Some tests fail. Here's one bad line:
test_accounts_db_serialize1
rv: Some(1), current_max: 1, slot: 2
So, we find a root at slot 2, but we previously found an ancestor at slot 1 so we tried to skip looking at roots.

jeffwashington:
all 3 failing tests from coverage:

accounts_db::tests::test_accounts_db_serialize1
bank::tests::test_bank_get_account_in_parent_after_squash2
bank::tests::test_purge_empty_accounts

jeffwashington:
@sakridge Here are 3 tests which fail if we find a slot in ancestors but then don't try looking in the roots. I imagine that these tests are producing artificial data which is inconsistent with the way the data behaves in a real running system.

I can work on reworking these tests. Do you think that is valuable?

sakridge:
> @sakridge Here are 3 tests which fail if we find a slot in ancestors but then don't try looking in the roots. I imagine that these tests are producing artificial data which is inconsistent with the way the data behaves in a real running system.
> 
> I can work on reworking these tests. Do you think that is valuable?

Yes, I believe so.

jeffwashington:
This change would result in a change of behavior.

> https://github.com/solana-labs/solana/blob/92f4018b07524ef6a0c44492309bde04205a4b5a/runtime/src/bank.rs#L8139
This request to get an account from bank0 results in ancestors = {(0,0)} and roots = {0, 1}. So, we find the account at slot 0 due to ancestors and then if we don’t check roots, we don’t find the account in slot 1 of roots. So, we return the account as of slot 0. Is it legal to be in this situation? Perhaps if we are calculating hashes in accounts bg service on an older bank?

PR #16382, opened 2021-04-05
simplifies PinnedVec implementation

behzadnouri:
#### Summary of Changes
* removed `PinnedIter` and `PinnedIterMut` in favor of using `std::slice::{Iter,IterMut}`.
* removed manual implementation of `Default`, and redundant `IntoIterator` implementation.
* removed the extra `Option` wrapper for recycler field.
* simplified `Drop` implementation.


utsl42 (approved):
LGTM

PR #16383, opened 2021-04-05
Remove old function: account_balance_for_capitalization

steviez:
#### Problem
This function currently returns one of its' parameters and is thus useless.

#### Summary of Changes
Removed all instances of function, and replaced with the parameter instead. Ryo called this out here:
https://github.com/solana-labs/solana/pull/16157#discussion_r607047216

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -3617,15 +3617,6 @@ impl AccountsDb {
         AccountsHash::checked_cast_for_capitalization(balances.map(|b| b as u128).sum::<u128>())
     }
 
-    // remove this by inlining and remove extra unused params upto all callchain
-    pub fn account_balance_for_capitalization(
-        lamports: u64,
-        _owner: &Pubkey,
-        _executable: bool,
-    ) -> u64 {
-        lamports
-    }
-
     fn calculate_accounts_hash(
         &self,
         slot: Slot,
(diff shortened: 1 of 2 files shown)


runtime/src/accounts_db.rs
```diff
@@ -8246,69 +8227,6 @@ pub mod tests {
...
-    }
-
-    #[test]
-    fn test_account_balance_for_capitalization_sysvar() {
-        let normal_sysvar = solana_sdk::account::create_account_for_test(
-            &solana_sdk::slot_history::SlotHistory::default(),
-        );
-        assert_eq!(
-            AccountsDb::account_balance_for_capitalization(
-                normal_sysvar.lamports,
```
jeffwashington, line 8274: I don't know if elsewhere we test normal_sysvar.lamports. You should probably replace the function call with normal_sysvar.lamports and leave the assert. The next one where we pass the 10 is obviously now useless.

runtime/src/accounts_db.rs
```diff
@@ -8246,69 +8227,6 @@ pub mod tests {
...
-        );
-    }
-
-    #[test]
-    fn test_account_balance_for_capitalization_native_program() {
-        let normal_native_program =
-            solana_sdk::native_loader::create_loadable_account_for_test("foo");
-        assert_eq!(
-            AccountsDb::account_balance_for_capitalization(
-                normal_native_program.lamports,
```
jeffwashington, line 8294: same thing here - replace the function call with normal_native_program.lamports.

runtime/src/accounts_db.rs
steviez: Good catch - I agree there is value to keeping the first half her in `test_account_balance_for_capitalization_sysvar()` and `test_account_balance_for_capitalization_native_program`

jeffwashington (approved):

ryoqun (approved):
LGTM thanks for cleaning!

PR #16385, opened 2021-04-05
RpcClient no longer panics in a tokio multi-threaded runtime

mvines:
Post-tokio 1 upgrade (Solana 1.6.2), `RpcClient` would panic if run from within a tokio runtime.   Fix it by shuttling the blocking requests off to a different tokio thread

One glorious day we'll have an async RpcClient (blocked on the jsonrpc crates turning into [jsonrpsee](https://github.com/paritytech/jsonrpsee) crates for a tokio  1 update here), after which we can autogen the RpcClient bindings.  For now `RpcClient` working in async is better than nothing though.

### DIFF
--- a/client/src/http_sender.rs
+++ b/client/src/http_sender.rs
@@ -8,12 +8,20 @@ use {
     },
     log::*,
     reqwest::{self, header::CONTENT_TYPE, StatusCode},
-    std::{thread::sleep, time::Duration},
+    std::{
+        sync::{
+            atomic::{AtomicU64, Ordering},
+            Arc,
+        },
+        thread::sleep,
+        time::Duration,
+    },
 };
 
 pub struct HttpSender {
-    client: reqwest::blocking::Client,
+    client: Arc<reqwest::blocking::Client>,
     url: String,
+    request_id: AtomicU64,
 }
 
 impl HttpSender {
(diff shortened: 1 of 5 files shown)


client/src/http_sender.rs
```diff
@@ -40,20 +58,26 @@ struct RpcErrorObject {
 
 impl RpcSender for HttpSender {
     fn send(&self, request: RpcRequest, params: serde_json::Value) -> Result<serde_json::Value> {
-        // Concurrent requests are not supported so reuse the same request id for all requests
-        let request_id = 1;
-
-        let request_json = request.build_request_json(request_id, params);
+        let request_id = self.request_id.fetch_add(1, Ordering::Relaxed);
```
mvines, line 61: Freebie!

client/src/http_sender.rs
```diff
@@ -122,3 +148,22 @@ impl RpcSender for HttpSender {
...
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn http_sender_on_tokio_multi_thread() {
+        let http_sender = HttpSender::new("http://localhost:1234".to_string());
+        let _ = http_sender.send(RpcRequest::GetVersion, serde_json::Value::Null);
+    }
+
+    #[tokio::test(flavor = "current_thread")]
+    #[should_panic(expected = "can call blocking only when running on the multi-threaded runtime")]
+    async fn http_sender_ontokio_current_thread_should_panic() {
```
mvines, line 164: This isn't great, but I don't see a way around.  Fortunately `[tokio::main]` creates a `multi_thread` runtime, so most users won't see this (but `#[tokio::test]`, sans flavour, is `current_thread`)

CriesofCarrots: I see, obv not ideal but okay for now. I haven't thought of anything better.

CriesofCarrots (approved):

PR #16401, opened 2021-04-06
docker: Expose all ports in Dockerfile, add back localnet.sh

joncinque:
#### Problem

Even though we don't recommend it for production use, sometimes people want to run a validator through Docker in development.  Since we got rid of `localnet.sh` in web3.js, it's more difficult to use Docker.

#### Summary of Changes

A few niceties to help:
* Add localnet.sh back in, without the parts specific to a package file in Node
* Expose all of the required ports in the `Dockerfile`
* Link to localnet.sh and Docker Hub in the documentation

I wasn't sure of the best way to make localnet.sh usable to outside people, so linking it from the documentation seemed to be the best option.  I dug around some of the Docker documentation and couldn't find any conventions about how to publish things like required ports, so any suggestions are welcome!

Fixes #


sdk/docker-solana/localnet.sh
```diff
@@ -0,0 +1,149 @@
+#!/usr/bin/env bash
```
jstarry, line 1: I personally feel like we shouldn't need to maintain this and we don't need to make it too easy to use docker for development. The docker file is probably sufficient for those who really want to use docker

sdk/docker-solana/Dockerfile
```diff
@@ -1,7 +1,35 @@
...
+# tpu_forwards
+EXPOSE 8004/udp
+# retransmit
+EXPOSE 8005/udp
+# repair
+EXPOSE 8006/udp
+# serve_repair
+EXPOSE 8007/udp
+# broadcast
+EXPOSE 8008/udp
```
jstarry, line 32: These changes are great, devs can use `docker run -P` to easily publish all ports

docs/src/running-validator/validator-reqs.md
```diff
@@ -51,10 +51,14 @@ especially for the case of running staked validators.
...
 not recommended and generally not supported. This is due to concerns of general
-docker's containerzation overhead and resultant performance degradation unless
+Docker's containerzation overhead and resultant performance degradation unless
 specially configured.
 
-We use docker only for development purpose.
+We use Docker only for development purposes. Docker Hub contains images for all
+releases at [solanalabs/solana](https://hub.docker.com/r/solanalabs/solana). We
+also provide a
+[convenience script](https://github.com/solana-labs/solana/blob/master/sdk/docker-solana/localnet.sh)
```
jstarry, line 60: I don't think we want to make it too easy to use docker, it's generally not something we want devs to do. Linking to the images sounds good though!

jstarry (commented):
Added my 2 cents, @mvines and @ryoqun may have different opinions

sdk/docker-solana/Dockerfile
joncinque: Yeah I hope it'll be useful!  The weird part is that `-P` publishes the ports to random other ports -- from https://docs.docker.com/engine/reference/builder/#expose :

> In this case, if you use -P with docker run, the port will be exposed once for TCP and once for UDP. Remember that -P uses an ephemeral high-ordered host port on the host, so the port will not be the same for TCP and UDP.

But it seems to work nicely if you're using a full Docker-based solution

docs/src/running-validator/validator-reqs.md
joncinque: That's fair -- how do you propose to give people access to that script?  It'll be tough to run with just the container without knowledge of all the ports required.  I tried using `-P` locally and can confirm that it maps to random ports on your machine, which to me defeats the purpose of making it available for development.  Unless we want the docker images to just be available for pure docker setups, and always have people use `solana-test-validator` locally.

On another note, `solana-test-validator` could use some docs.

mvines:
Do we have a good idea why Docker is being preferred over solana-test-validator? I’m not super excited to see this script be resurrected TBH, and I don’t really want to support it

joncinque:
@mvines this is specifically related to a request from Civic -- they run a k8 cluster so for them it's much easier to spin up a container and have applications use that.  That's the most (only?) valid use-case for these containers in my opinion.  Otherwise, what are they for?

mvines:
Ah, ok.  Let's just give them this script then and they can maintain it :)

mvines:
The API provided by our command-line tools, `solana-validator`, `solana`, `solana-keygen`, are sufficiently stable now that there shouldn't be much churn with this script out of our tree.  If/when we make incompatible chanes, we need to announce well in advance these days anyway to avoid breaking all downstream validators

joncinque:
@mvines heh, all right, that's two votes against this script.  I've removed it and references to it

sdk/docker-solana/Dockerfile
jstarry: Ah yeah that is unfortunate. This is still nice though!

jstarry (approved):

PR #16402, opened 2021-04-06
[easy, cleanup] Simplify some pattern-matches

huitseeker:
When those match an exact combinator on Option / Result.

Tool-aided by [comby-rust](https://github.com/huitseeker/comby-rust).

### DIFF
--- a/core/src/duplicate_shred.rs
+++ b/core/src/duplicate_shred.rs
@@ -226,10 +226,7 @@ pub fn into_shreds(
         chunk_index,
         chunk,
         ..
-    } = match chunks.next() {
-        None => return Err(Error::InvalidDuplicateShreds),
-        Some(chunk) => chunk,
-    };
+    } = chunks.next().ok_or(Error::InvalidDuplicateShreds)?;
     let slot_leader = leader(slot).ok_or(Error::UnknownSlotLeader)?;
     let check_chunk = check_chunk(slot, shred_index, shred_type, num_chunks);
     let mut data = HashMap::new();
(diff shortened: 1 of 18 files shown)


core/src/duplicate_shred.rs
```diff
@@ -226,10 +226,7 @@ pub fn into_shreds(
         chunk_index,
         chunk,
         ..
-    } = match chunks.next() {
-        None => return Err(Error::InvalidDuplicateShreds),
-        Some(chunk) => chunk,
-    };
+    } = chunks.next().ok_or_else(|| Error::InvalidDuplicateShreds)?;
```
mvines, line 229: Looks like clippy doesn't like this one!  The closure is superfluous

mvines:
Cool tool @huitseeker!

net-utils/src/ip_echo_server.rs
```diff
@@ -29,8 +29,8 @@ impl IpEchoServerMessage {
         assert!(tcp_ports.len() <= msg.tcp_ports.len());
         assert!(udp_ports.len() <= msg.udp_ports.len());
 
-        msg.tcp_ports[..tcp_ports.len()].copy_from_slice(tcp_ports);
-        msg.udp_ports[..udp_ports.len()].copy_from_slice(udp_ports);
+        msg.tcp_ports[..].copy_from_slice(tcp_ports);
+        msg.udp_ports[..].copy_from_slice(udp_ports);
```
CriesofCarrots, line 33: @huitseeker , looks like this change causes some CI tests to fail with the panic `source slice length does not match destination slice length` because `msg.tcp_ports.len()` may be greater than `tcp_ports.len()` (ditto for udp_ports).
Am I missing something?

huitseeker: 🤦

CriesofCarrots:
Thanks for the tidy up and sharing your tool, @huitseeker !
No change in functionality, so adding v1.5/v1.6 backports to prevent future backport merge conflicts

PR #16421, opened 2021-04-07
Track gossip vote updates per hash for replay stage

carllin:
#### Problem
Fork choice and switching votes from replay stage should include votes observed in gossip

#### Summary of Changes
First step,  add ability in cluster info vote listener to  track new votes by hash 
fyi @aeyakovenko 

Fixes #


sakridge (approved):
Looks pretty straightforward. lgtm

PR #16422, opened 2021-04-07
Return sysvars via syscalls

### LINKED ISSUE #15755: Consider mapping sysvars to syscalls

jackcmay:
#### Problem

Requiring developers to include sysvars in their instructions complicates composibility and requires programs to pass the sysvar accounts between programs when an intermediate might not actually need them.

Also, because of the bincode serialization of the sysvar data and the lack of a C parser we currently don't support sysvars in C BPF programs.

#### Summary of Changes

- Allow a program to call a syscall to obtain the sysvar.
- Do so in a way that does NOT require the bincode deserialization in-program
- Sysvars clock, epoch_schedule, fees, and rent are currently supported and new simple `repr(C)` sysvars can use this same mechanism in the future.  Large or complex sysvar types will need a different mechanism to map the sysvar memory into the program's memory space.

Needed by: https://github.com/solana-labs/solana/pull/15927

Fixes #15755

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4986,10 +4986,13 @@ version = "1.7.0"
 dependencies = [
  "async-trait",
  "base64 0.12.3",
+ "bincode",
  "chrono",
  "chrono-humanize",
  "log 0.4.11",
  "mio 0.7.6",
+ "serde",
+ "serde_derive",
  "solana-banks-client",
  "solana-banks-server",
  "solana-bpf-loader-program",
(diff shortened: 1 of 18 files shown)


mvines (commented):
Really nice!  

We should backport to v1.6 at least

PR #16447, opened 2021-04-08
Add memory operation syscalls

jackcmay:
#### Problem

Large repetitive memory operations (memcpy) are expensive in BPF

#### Summary of Changes

- Provide syscalls to do those operations natively
- This PR exposes syscalls for memcpy, memmove, memset, and memcmp.
- Consolidate the memory op tests into a trait/function combo so we can run the memory tests on multiple implementations of the above functions.  Currently testing builtins and syscalls
- Once these are mainlined we can update the toolchain to take advantage of these syscalls inherently.

Fixes #


dmakarov:
Is this going to be merged soon? Once this is merged I can release a new version of bpf-tools that uses the memory operations syscalls

dmakarov (approved):

PR #16461, opened 2021-04-09
feat: support commitment in getConfirmed methods

CriesofCarrots:
#### Problem
Various solana rpc methods that used to only return finalized data now support confirmed commitment. But solana-web3 does not expose a commitment parameter for these methods.

#### Summary of Changes
Add commitment to methods that target `getConfirmedBlock`, `getConfirmedTransaction`, and `getConfirmedSignaturesForAddress2` endpoints.

### DIFF
--- a/web3.js/src/connection.ts
+++ b/web3.js/src/connection.ts
@@ -220,6 +220,15 @@ export type Commitment =
   | 'max'; // Deprecated as of v1.5.5
 
 /**
+ * A subset of Commitment levels, which are at least optimistically confirmed
+ * <pre>
+ *   'confirmed': Query the most recent block which has reached 1 confirmation by the cluster
+ *   'finalized': Query the most recent block which has been finalized by the cluster
+ * </pre>
+ */
+export type Finality = 'confirmed' | 'finalized';
+
+/**
  * Filter for largest accounts query
  * <pre>
  *   'circulating':    Return the largest accounts that are part of the circulating supply
--- a/web3.js/src/connection.ts
+++ b/web3.js/src/connection.ts
@@ -2596,8 +2596,12 @@ export class Connection {
    * Fetch a list of Transactions and transaction statuses from the cluster
    * for a confirmed block
    */
-  async getConfirmedBlock(slot: number): Promise<ConfirmedBlock> {
-    const unsafeRes = await this._rpcRequest('getConfirmedBlock', [slot]);
+  async getConfirmedBlock(
+    slot: number,
+    commitment?: Finality,
+  ): Promise<ConfirmedBlock> {
+    const args = this._buildArgsAtLeastConfirmed([slot], commitment);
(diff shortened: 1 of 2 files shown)


CriesofCarrots:
These rpc endpoints do not support `processed` commitment, so when that commitment is passed (or a `processed` Connection is used), the web3 methods will throw:
```
Error: failed to get confirmed signatures for address: Method does not support commitment below `confirmed`
```
That seems okay to me, but we could check the commitment before submitting the request. Wdyt?

jstarry:
Thanks for adding this!

> so when that commitment is passed (or a processed Connection is used), the web3 methods will throw

I would prefer if it was harder to misuse this API. How about typing the `commitment` directly as `"confirmed" | "finalized"`?

Another thing is that `_buildArgs` falls back to the commitment set on the `Connection` class. If it's set to `processed` these methods would use that commitment by default. For these methods I think we might want to use something different from `_buildArgs`

CriesofCarrots:
> I would prefer if it was harder to misuse this API. How about typing the `commitment` directly as `"confirmed" | "finalized"`?

Sure, that makes sense.

> Another thing is that `_buildArgs` falls back to the commitment set on the `Connection` class. If it's set to `processed` these methods would use that commitment by default. For these methods I think we might want to use something different from `_buildArgs`

Well, I do think the api should work without an explicit commitment, if the Connection commitment is confirmed or finalized. Are you thinking something like a `_buildArgsIfAtLeastConfirmed` to handle the commitment check and throw a client error? (as opposed the the rpc error above, which currently gets thrown)

jstarry:
> Well, I do think the api should work without an explicit commitment, if the Connection class-defined commitment is confirmed or finalized. Are you thinking something like a _buildArgsIfAtLeastConfirmed to handle the commitment check and throw a client error? (as opposed the the rpc error above, which currently gets thrown)

Yeah, that sounds like a good solution to me

CriesofCarrots:
@jstarry , I think this is ready for another look. Will squash before merge.
Alternate, better names for the `_buildArgsAtLeastConfirmed` method very welcome!
Also, I waffled about a new CommitmentAtLeastConfirmed type to make the methods less brittle, should we add additional at-least-confirmed commitment variants upstream. Wdyt?

jstarry:
> @jstarry , I think this is ready for another look. Will squash before merge.

Looks great!

> Alternate, better names for the `_buildArgsAtLeastConfirmed` method very welcome!

It's descriptive and internal, seems fine to me!

> Also, I waffled about a new CommitmentAtLeastConfirmed type to make the methods less brittle, should we add additional at-least-confirmed commitment variants upstream. Wdyt?

Hmm what about calling it `Finality` where `Finality` is a stricter subset of `Commitment` which just includes optimistic confirmation ("confirmed") and finalized?

jstarry (approved):

PR #16477, opened 2021-04-12
replace Ancestors HashMap for performance

jeffwashington:
#### Problem
We grab an account index lock and look up slots as ancestors or roots. The ancestor lookup is slower than it could be, given the constraints on the type of lookup it is.
An Ancestors instance also almost never changes during its lifetime.

#### Summary of Changes
Replace the HashMap with a variable min/max range vector lookup.

Fixes #


jeffwashington:
measurements: this pr is 15-26x faster to do a lookup than the HashMap code.

runtime/src/accounts_index.rs
```diff
@@ -32,7 +32,164 @@ pub const ITER_BATCH_SIZE: usize = 1000;
...
+                    result.slots[slot as usize] = Some(size);
+                });
+            }
+        }
+
+        result
+    }
+}
+
+impl From<&HashMap<Slot, usize>> for Ancestors {
```
jeffwashington, line 81: vec and hashmap are very similar code. I spent some time with iterators, generics, traits, etc. trying to get a cleaner solution than duplicating the code. So far, my efforts have not worked.

jeffwashington: hashmap is needed for persistence, vector is the way all production code initializes a vector (as well as most test code).

runtime/src/accounts_index.rs
```diff
@@ -32,7 +32,167 @@ pub const ITER_BATCH_SIZE: usize = 1000;
 
 pub type SlotList<T> = Vec<(Slot, T)>;
 pub type SlotSlice<'s, T> = &'s [(Slot, T)];
-pub type Ancestors = HashMap<Slot, usize>;
+
+pub type AncestorsForSerialization = HashMap<Slot, usize>;
+
+#[derive(Debug, Default, Clone, PartialEq, Serialize, Deserialize, AbiExample)]
+pub struct Ancestors {
```
sakridge, line 39: How about Ancestors in it's own module

sakridge:
Implementation seems good

runtime/src/accounts_index.rs
jeffwashington: Yep, I was thinking about its own module earlier today. Seems appropriate.
I just went over all this with @carllin, too.

PR #16484, opened 2021-04-12
expands number of erasure coding shreds in the last batch in slots

behzadnouri:
#### Problem
Number of parity coding shreds is always less than the number of data shreds in FEC blocks:
https://github.com/solana-labs/solana/blob/6907a2366/ledger/src/shred.rs#L719

Data shreds are batched in chunks of 32 shreds each:
https://github.com/solana-labs/solana/blob/6907a2366/ledger/src/shred.rs#L714

However the very last batch of data shreds in a slot can be small, in which case the loss rate can be exacerbated.

#### Summary of Changes
This commit expands the number of coding shreds in the last FEC block in slots to: 64 - number of data shreds; so that FEC blocks are always 64 data and parity coding shreds each.

As a consequence of this, the last FEC block has more parity coding shreds than data shreds. So for some shred indices we will have a coding shred but no data shreds. This should not cause any kind of overlapping FEC blocks as in: https://github.com/solana-labs/solana/pull/10095 since this is done only for the very last batch in a slot, and the next slot will reset the shred index.

### DIFF
--- a/ledger/src/shred.rs
+++ b/ledger/src/shred.rs
@@ -141,7 +141,6 @@ pub const DATA_SHRED: u8 = 0b1010_0101;
 pub const CODING_SHRED: u8 = 0b0101_1010;
 
 pub const MAX_DATA_SHREDS_PER_FEC_BLOCK: u32 = 32;
-pub const RECOMMENDED_FEC_RATE: f32 = 1.0;
 
 pub const SHRED_TICK_REFERENCE_MASK: u8 = 0b0011_1111;
 const LAST_SHRED_IN_SLOT: u8 = 0b1000_0000;
@@ -568,7 +567,6 @@ pub struct Shredder {
     pub slot: Slot,
     pub parent_slot: Slot,
     version: u16,
-    fec_rate: f32,
     keypair: Arc<Keypair>,
     pub signing_coding_time: u128,
     reference_tick: u8,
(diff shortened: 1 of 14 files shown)


aeyakovenko (approved):

ledger/src/shred.rs
```diff
@@ -775,100 +761,53 @@ impl Shredder {
...
+        } = data.first().unwrap().common_header;
+        assert_eq!(fec_set_index, index);
+        assert!(data.iter().all(|shred| shred.common_header.slot == slot
+            && shred.common_header.version == version
+            && shred.common_header.fec_set_index == fec_set_index));
+        let num_data = data.len();
+        let num_coding = if is_last_in_slot {
+            (2 * MAX_DATA_SHREDS_PER_FEC_BLOCK as usize)
+                .saturating_sub(num_data)
+                .max(num_data)
```
carllin, line 781: sanity check, `num_data` should always be `<= MAX_DATA_SHREDS_PER_FEC_BLOCK`,  due to chunking by the caller right? Should we add this as an assert?

carllin (approved):

ledger/src/shred.rs
behzadnouri: I do not think we need an `assert`, since this function may be called directly and larger batches are totally valid.

carllin: yeah, was wondering if we wanted to ensure that `num_data + num_coding < 2 * MAX_DATA_SHREDS_PER_FEC_BLOCK`, but if we are planning for supporting larger batches as well, then that makes sense!.

Issue #1830, opened 2018-11-15
SDK: remove requestAirdrop from the JSON RPC API 


mvines:
The JSON `requestAirdrop` API is really just a thin wrapper around a drone, and makes unfortunate assumptions about where the drone is running and it's port.  With #1754 and the potential for many drones running around, we should fully decouple the drone from the RPC API and make the users fully aware of the drone they decide to use.  

Note that the drone itself may choose to implement something that looks very much like the existing JSON `requestAirdrop` API as a convenience for web-based users.

garious:
@mvines, is this still an issue?

mvines:
Yeah

PR #16487, opened 2021-04-12
Faucet: Add address_cache and exclude loopback from ip limit

CriesofCarrots:
#### Problem
Faucet requests via the `requestAirdrop` rpc endpoint all appear to have the same (loopback) IP; they will be overly limited by faucets running with a per-time-cap.

#### Summary of Changes
Add address_cache to faucet and also limit on request recipient address
Exclude loopback IP from ip limiting

Note, this is a quick stop-gap. Real solution is: https://github.com/solana-labs/solana/issues/1830


mvines (approved):

CriesofCarrots:
I'll manually backport

PR #16495, opened 2021-04-12
Add llvm feature option to compile for Solana BPF target

dmakarov:
#### Summary of Changes

Fixes #


dmakarov:
The testing on buildkite hosts is not deterministic. At one run the number of instructions is 901, at another run it's 869
```
  solana_bpf_rust_sanity              901    869  -32
```

dmakarov:
I'm not sure this needs to be backported to 1.6. The toolchain is essentially the same, except Solana LLVM extensions are enabled by `-march=bpfel+solana`

PR #16502, opened 2021-04-13
Rpc: deprecate getConfirmed endpoints

CriesofCarrots:
#### Problem
`getConfirmedX` endpoint names are overly verbose, and not very helpful since we've: (1) clarified the meaning of "confirmed"; and (2) updated these methods to support the `commitment` parameter.

#### Summary of Changes
- Move `getConfirmedX` endpoints to `rpc_deprecated_v1_7` for future removal
- Add `getX` analog endpoints

Follow-up work:
- [ ] Update solana-client: use new endpoints, and rename/deprecate client methods

~~Needs rebase on https://github.com/solana-labs/solana/pull/16500~~

### DIFF
--- a/docs/src/developing/clients/jsonrpc-api.md
+++ b/docs/src/developing/clients/jsonrpc-api.md
@@ -26,11 +26,6 @@ gives a convenient interface for the RPC methods.
 - [getBlocksWithLimit](jsonrpc-api.md#getblockswithlimit)
 - [getBlockTime](jsonrpc-api.md#getblocktime)
 - [getClusterNodes](jsonrpc-api.md#getclusternodes)
-- [getConfirmedBlock](jsonrpc-api.md#getconfirmedblock)
-- [getConfirmedBlocks](jsonrpc-api.md#getconfirmedblocks)
-- [getConfirmedBlocksWithLimit](jsonrpc-api.md#getconfirmedblockswithlimit)
-- [getConfirmedSignaturesForAddress2](jsonrpc-api.md#getconfirmedsignaturesforaddress2)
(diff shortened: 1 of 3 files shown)


CriesofCarrots:
@mvines quick pre-review question: in the past, we've been inconsistent about docs handling for deprecated endpoints. In some cases, we've marked them as deprecated; in others, we've removed them from docs altogether. I slightly prefer the latter, so that the docs can help encourage people to upgrade, but it's less obvious for new users which methods to use. How would you feel about a Deprecated Methods section?

mvines:
> How would you feel about a Deprecated Methods section?

Yep I agree.  I was thinking about such a thing as well

docs/src/developing/clients/jsonrpc-api.md
```diff
@@ -79,6 +79,14 @@ gives a convenient interface for the RPC methods.
   - [slotSubscribe](jsonrpc-api.md#slotsubscribe)
   - [slotUnsubscribe](jsonrpc-api.md#slotunsubscribe)
 
+### Deprecated Methods
```
mvines, line 82: For each of the deprecated methods, how about adding a statement indicating the release that we expect to remove them in

mvines (approved):

mvines:
I guess RpcClient needs the new function names as well with `#[deprecated]` attributes for the old ones

CriesofCarrots:
> I guess RpcClient needs the new function names as well with `#[deprecated]` attributes for the old ones

Yep, definitely. That's the follow-up work noted in the description. Separate PR okay?

PR #10264, opened 2020-05-27
Track outstanding nonces in repair

carllin:
#### Problem
Arbitrary repairs can be sent to repair ports

#### Summary of Changes
Last changes from: https://github.com/solana-labs/solana/pull/9903
Add a nonce to repair/responses that needs to match in order for response to be accepted.
Fixes #

PR #16512, opened 2021-04-13
Add non-default repair nonce values

sakridge:
#### Problem

Repair nonce is set to const default value and not checked. 

#### Summary of Changes

Add non-default nonce values and check that when window service receives repairs, the nonce is matched.

Fixes #


core/src/repair_service.rs
```diff
@@ -273,8 +277,20 @@ impl RepairService {
...
                 }
             });
+            {
+                let mut nonces = nonces.write().unwrap();
+                let now = timestamp();
+                for nonce in new_nonces {
+                    nonces.insert(nonce, now);
+                }
+                nonces
+                    .retain(|_nonce, timestamp| now.saturating_sub(*timestamp) >= MAX_NONCE_AGE_MS);
```
behzadnouri, line 292: this is retaining old nonces

core/src/repair_service.rs
```diff
@@ -345,6 +359,14 @@ impl RepairService {
         }
     }
 
+    fn insert_new_and_filter_old_nonces(new_nonces: Vec<Nonce>, nonces: &mut HashMap<Nonce, u64>) {
+        let now = timestamp();
+        for nonce in new_nonces {
+            nonces.insert(nonce, now);
+        }
+        nonces.retain(|_nonce, timestamp| now.saturating_sub(*timestamp) < MAX_NONCE_AGE_MS);
```
behzadnouri, line 367: The `retain` thing is traversing the entire hashmap for pruning, which might be slow.
Another idea is to use `LruCache<Nonce, u64>` which prunes by itself.
`verify_repair` can also check that if the nonce has expired, and if so remove it from the cache as well.

sakridge: True, good idea.

carllin:
whoo finally getting this patched 😄 

I also had an outstanding PR for this: https://github.com/solana-labs/solana/pull/10264

A few ideas I thought were good to address/incorporate:

1) Keeping around a timeout and number of expected responses per request: https://github.com/solana-labs/solana/pull/10264/files#diff-d0fb1321af935f265f21af26229db3ace2fda6083c96a53f1d4ef58d2deb5b76R82-R86, validated here: https://github.com/solana-labs/solana/pull/10264/files#diff-d0fb1321af935f265f21af26229db3ace2fda6083c96a53f1d4ef58d2deb5b76R44-R58

2) Implementing specific verification logic for each type of repair response: https://github.com/solana-labs/solana/pull/10264/files#diff-687eeb3a159097761026218d39b62bd6269e3163b67b9983314abb97f2d39774R57-R77

3) Tracking each outstanding request by validator: https://github.com/solana-labs/solana/pull/10264/files#diff-d0fb1321af935f265f21af26229db3ace2fda6083c96a53f1d4ef58d2deb5b76R99,  and ensuring responses came from those specific validators: https://github.com/solana-labs/solana/pull/10264/files#diff-d0fb1321af935f265f21af26229db3ace2fda6083c96a53f1d4ef58d2deb5b76R105

sakridge:
> whoo finally getting this patched 😄
> 
> I also had an outstanding PR for this: #10264.
> 
> A few ideas I thought were good to address/incorporate from there:
> 
> 1. Keeping around a timeout and number of expected responses per request: https://github.com/solana-labs/solana/pull/10264/files#diff-d0fb1321af935f265f21af26229db3ace2fda6083c96a53f1d4ef58d2deb5b76R82-R86, validated here: https://github.com/solana-labs/solana/pull/10264/files#diff-d0fb1321af935f265f21af26229db3ace2fda6083c96a53f1d4ef58d2deb5b76R44-R58
> 2. Implementing specific verification logic for each type of repair response: https://github.com/solana-labs/solana/pull/10264/files#diff-687eeb3a159097761026218d39b62bd6269e3163b67b9983314abb97f2d39774R57-R77
> 3. In the nonce discussions, we had planned originally to track each outstanding request by validator, done here: https://github.com/solana-labs/solana/pull/10264/files#diff-d0fb1321af935f265f21af26229db3ace2fda6083c96a53f1d4ef58d2deb5b76R99,  and ensuring responses came from those specific validators: https://github.com/solana-labs/solana/pull/10264/files#diff-d0fb1321af935f265f21af26229db3ace2fda6083c96a53f1d4ef58d2deb5b76R105

Oh cool. I didn't see that PR. This PR does require a timeout, but it doesn't do the request-specific checking.

For 3, I'm not sold on doing anything by IP. IP can vary legitimately based on network topology and routing and also can be spoofed. What is the case where someone would have the valid nonce to return to you and wouldn't be able to spoof the traffic coming to your node?

sakridge:
@carllin what is pending from the PR besides rebasing?

carllin:
> @carllin what is pending from the #10264 PR besides rebasing?

Should just require a rebase 😃 

> For 3, I'm not sold on doing anything by IP. IP can vary legitimately based on network topology and routing and also can be spoofed. What is the case where someone would have the valid nonce to return to you and wouldn't be able to spoof the traffic coming to your node?

Hah yeah for more robust safety we would need  encryption: https://discord.com/channels/428295358100013066/478692221441409024/705546482765266994. I think tracking the IP was just part of the "limitation" we described here: https://discord.com/channels/428295358100013066/478692221441409024/705544223792824383 to make things a little more difficult, but definitely not critical to have.

core/src/outstanding_requests.rs
```diff
@@ -0,0 +1,166 @@
...
+        nonce
+    }
+
+    pub fn register_response(&mut self, nonce: u32, response: &S, now: u64) -> bool {
+        let (is_valid, should_delete) = self
+            .requests
+            .get_mut(&nonce)
+            .map(|status| {
+                if status.num_expected_responses > 0
+                    && now < status.expire_timestamp
```
behzadnouri, line 42: shouldn't it also remove the key if `now > expire_timestamp`?

also this looks like it keeps the nonce alive as long as the sender is sending invalid responses because those will not decrease `num_expected_responses`. shouldn't invalid responses also count towards `num_expected_responses`?

core/src/outstanding_requests.rs
```diff
@@ -0,0 +1,166 @@
...
+{
+    // Returns boolean indicating whether sufficient time has passed for a request with
+    // the given timestamp to be made
+    pub fn add_request(&mut self, request: T) -> Nonce {
+        let num_expected_responses = request.num_expected_responses();
+        let nonce = thread_rng().gen_range(0, Nonce::MAX);
+        self.requests.put(
+            nonce,
+            RequestStatus {
+                expire_timestamp: timestamp() + DEFAULT_REQUEST_EXPIRATION_MS,
```
behzadnouri, line 28: maybe like the other function this one also better take `now: u64` as an argument rather than calling `timestamp()` itself.

behzadnouri (approved):
lgtm

carllin (dismissed):

core/src/outstanding_requests.rs
sakridge: I think it does remove the key if `now > expire_timestamp` no? If `now < status.expire_timestamp` is false, then it will go into the `else` which will set `should_delete=true`.

See this test:
https://github.com/solana-labs/solana/pull/16512/files#diff-d0fb1321af935f265f21af26229db3ace2fda6083c96a53f1d4ef58d2deb5b76R94-R107

sakridge: > shouldn't invalid responses also count towards num_expected_responses

Maybe. I can't really think of an attack from it. Is there a downside?

behzadnouri: > I think it does remove the key if now > expire_timestamp no?

true, my mistake

PR #16514, opened 2021-04-13
Document shreds

steviez:
No functionality changes from this commit

#### Problem
The shreds module has some intricacies that weren't immediately obvious; hopefully the additions from this commit make things more clear for anyone trying to figure these out from a clean slate.

#### Summary of Changes
- Add some additional explanation / documentation of shreds at the top
- Rename several variables with more meaningful (in my opinion :) ) names
- Some misc. comment cleanup

### DIFF
--- a/ledger/src/shred.rs
+++ b/ledger/src/shred.rs
@@ -1,4 +1,54 @@
-//! The `shred` module defines data structures and methods to pull MTU sized data frames from the network.
+//! The `shred` module defines data structures and methods to pull MTU sized data frames from the
+//! network. There are two types of shreds: data and coding. Data shreds contain entry information
+//! while coding shreds provide redundancy to protect against dropped network packets (erasures).
+//!
+//! +---------------------------------------------------------------------------------------------+
+//! | Data Shred                                                                                  |
+//! +---------------------------------------------------------------------------------------------+
+//! | common       | data       | payload                                                         |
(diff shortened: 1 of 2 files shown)


ledger/src/shred.rs
```diff
@@ -1,4 +1,54 @@
...
+//!
+//! +---------------------------------------------------------------------------------------------+
+//! | Data Shred                                                                                  |
+//! +---------------------------------------------------------------------------------------------+
+//! | common       | data       | payload                                                         |
+//! | header       | header     |                                                                 |
+//! |+---+---+---  |+---+---+---|+----------------------------------------------------------+----+|
+//! || s | s | .   || p | f | s || data (ie ledger entries)                                 | r  ||
+//! || i | h | .   || a | l | i ||                                                          | e  ||
+//! || g | r | .   || r | a | z || See notes below for explanation of restricted            | s  ||
```
carllin, line 13: Maybe wrap "restricted' in quotes to signal that it's a field in the struct. Also, maybe reference the specific function where this explanation is available?

carllin (dismissed):
Thanks for the documentation, looks great!

ledger/src/shred.rs
steviez: Done - put in quotes and mentioned that the "notes" are immediately after the shred diagrams

PR #16516, opened 2021-04-13
Deprecate RpcClient methods, RpcRequest variants

CriesofCarrots:
#### Problem
5 rpc endpoints have been deprecated for a while, and are now obsolete and slated for removal (https://github.com/solana-labs/solana/pull/16500). A couple of these are still supported in RpcRequest, and have associated RpcClient methods.

#### Summary of Changes
Use deprecated attribute to alert downstream users about `getTotalSupply` and `getConfirmedSignaturesForAddress`.
The v1.5/v1.6 backports are the main thrust of this pr; plan to remove these variants/methods in v1.7

PR #16520, opened 2021-04-13
Deprecate "confirmed" RpcClient methods

CriesofCarrots:
#### Problem
https://github.com/solana-labs/solana/pull/16502 deprecated `getConfirmed` rpc endpoints; however, they are still being used in solana-client

#### Summary of Changes
- Remove obsolete RpcClient methods (follows #16516 )
- Deprecate "confirmed" RpcClient methods and replace with new ones
- Rename Confirmed rpc_config structs, using deprecated versions for backward compatibility. I packaged these in a separate module to make them easy to remove with the rpc and client methods in v1.8

Probably easiest to review by commit, as the 3rd commit has a bit of churn.

Completes #16502

### DIFF
--- a/cli/src/cli.rs
+++ b/cli/src/cli.rs
@@ -28,7 +28,7 @@ use solana_client::{
     nonce_utils,
     rpc_client::RpcClient,
     rpc_config::{
-        RpcConfirmedTransactionConfig, RpcLargestAccountsFilter, RpcSendTransactionConfig,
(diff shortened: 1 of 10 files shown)


mvines (dismissed):

CriesofCarrots:
I think we'll want clients on this code to support at least back to v1.6 rpc, so RpcClient needs a bit more work, a la `get_node_version()` before this is ready to go.

PR #16534, opened 2021-04-14
web3.js: deprecate getTotalSupply and getConfirmedSignaturesForAddress

CriesofCarrots:
#### Problem
The rpc endpoints `getTotalSupply` and `getConfirmedSignaturesForAddress` have been deprecated for quite a stretch, and we are removing them from solana master: https://github.com/solana-labs/solana/pull/16500
But they have not yet been deprecated in solana-web3.js

#### Summary of Changes
- Deprecate `getTotalSupply` by using `getSupply` under the hood
- Deprecate `getConfirmedSignaturesForAddress` by getting blocks to find before/until signatures, and using `getConfirmedSignaturesForAddress2`
  - As part of this, add `Connection.getConfirmedBlockSignatures` helper using `getConfirmedBlock` config options to save having to get all transaction and rewards data in the response. This is a bit of a shortcut; `Connection.getConfirmedBlock` ought to be updated to support the extended config options. But that needs to be done carefully, since making the `transactions` field optional breaks type assumptions.

### DIFF
--- a/web3.js/src/connection.ts
+++ b/web3.js/src/connection.ts
@@ -2308,15 +2308,16 @@ export class Connection {
 
   /**
    * Fetch the current total currency supply of the cluster in lamports
+   * @deprecated Deprecated since v1.2.8. Use `Connection.getSupply()` instead.
    */
   async getTotalSupply(commitment?: Commitment): Promise<number> {
     const args = this._buildArgs([], commitment);
-    const unsafeRes = await this._rpcRequest('getTotalSupply', args);
-    const res = create(unsafeRes, jsonRpcResult(number()));
+    const unsafeRes = await this._rpcRequest('getSupply', args);
+    const res = create(unsafeRes, GetSupplyRpcResult);
     if ('error' in res) {
       throw new Error('failed to get total supply: ' + res.error.message);
     }
-    return res.result;
+    return res.result.value.total;
   }
 
   /**
--- a/web3.js/src/connection.ts
+++ b/web3.js/src/connection.ts
@@ -932,13 +932,6 @@ const StakeActivationResult = pick({
 });
 
 /**
- * Expected JSON RPC response for the "getConfirmedSignaturesForAddress" message
- */
-const GetConfirmedSignaturesForAddressRpcResult = jsonRpcResult(
-  array(string()),
-);
-
-/**
  * Expected JSON RPC response for the "getConfirmedSignaturesForAddress2" message
  */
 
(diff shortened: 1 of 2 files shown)


web3.js/src/connection.ts
```diff
@@ -2554,17 +2601,33 @@ export class Connection {
...
-      );
+    let options: any = {};
+
+    if (startSlot > 1) {
+      let start = startSlot - 1;
+      let startSignatures: Array<string> = [];
+      while (startSignatures.length == 0) {
+        const block = await this.getConfirmedBlockSignatures(start);
+        startSignatures = block.signatures;
+        start++;
```
jstarry, line 2612: Should be searching back further right? If `start` goes to 0 then we can probably omit `until`?

web3.js/src/connection.ts
```diff
@@ -2554,17 +2601,33 @@ export class Connection {
...
-      throw new Error(
-        'failed to get confirmed signatures for address: ' + res.error.message,
-      );
+    let options: any = {};
+
+    if (startSlot > 1) {
+      let start = startSlot - 1;
+      let startSignatures: Array<string> = [];
+      while (startSignatures.length == 0) {
+        const block = await this.getConfirmedBlockSignatures(start);
```
jstarry, line 2610: If `start` is a skipped block, this method will throw

web3.js/src/connection.ts
```diff
@@ -2554,17 +2601,33 @@ export class Connection {
...
+      options.until = startSignatures[startSignatures.length - 1].toString();
     }
-    return res.result;
+
+    let end = endSlot + 1;
+    let endSignatures: Array<string> = [];
+    while (endSignatures.length == 0) {
+      const block = await this.getConfirmedBlockSignatures(end);
+      endSignatures = block.signatures;
+      end++;
```
jstarry, line 2622: If end is larger than the latest confirmed block, this will throw

web3.js/src/connection.ts
CriesofCarrots: Ugh, yes, that was my intention. Thanks for catching!

CriesofCarrots:
@jstarry can you take a look at the latest commit? (I'll squash before merge)

web3.js/src/connection.ts
jstarry: @CriesofCarrots this is still an issue

CriesofCarrots: Sorry, I actually meant to ask you if you have recommendations for this, since the Connection methods don't distinguish between different kinds of errors. Is a string match the best we can do to distinguish skipped blocks (where we should continue the loop)?

jstarry: String match seems fine for now since we have no precedent for error codes yet. I think that can be pushed off until we adopt it across all of `Connection` methods

CriesofCarrots: Roger. https://github.com/solana-labs/solana/pull/16534/commits/917679ef92c80d1e00a1df8f77f8c6650fa7215a

jstarry (dismissed):

PR #16539, opened 2021-04-14
Correct gossip hostname

ConnorMcF:
#### Problem
Old hostname no longer works for gossip

#### Summary of Changes
Corrects hostname to devnet entrypoint


mvines:
Thanks for fixing this up!

PR #16541, opened 2021-04-14
prioritizes contact-infos in pull responses

behzadnouri:
#### Problem

Expired crds values where the contact-info does not exist are wasted:
https://github.com/solana-labs/solana/blob/f804ce63c/core/src/crds_gossip_pull.rs#L353-L378
and then are sent again over the next pull-request.

Also, the stake of the first response (which can be anything) is used to
weight all pull-responses to a node, while the rest of responses can
have different stake.
https://github.com/solana-labs/solana/blob/f804ce63c/core/src/cluster_info.rs#L2231

#### Summary of Changes
* Prioritize contact-infos in pull-responses.
* Patch the stake weight bug.

### DIFF
--- a/core/src/cluster_info.rs
+++ b/core/src/cluster_info.rs
@@ -27,9 +27,7 @@ use crate::{
     result::{Error, Result},
     weighted_shuffle::weighted_shuffle,
 };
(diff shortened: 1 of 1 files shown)


sakridge (approved):
cool, lgtm

PR #16549, opened 2021-04-14
uses timeouts based on stake for filtering pull responses

behzadnouri:
#### Problem
filter_pull_responses is using default timeout when discarding pull
responses (except for ContactInfo):
https://github.com/solana-labs/solana/blob/f804ce63c/core/src/crds_gossip_pull.rs#L349-L350

But purging code uses timeouts based on stake:
https://github.com/solana-labs/solana/blob/f804ce63c/core/src/cluster_info.rs#L1867-L1870

So the crds value will not be purged from the sender's table and will be
sent again over the next pull request.

#### Summary of Changes
* Use timeouts based on stake when discarding pull responses.

### DIFF
--- a/core/src/cluster_info.rs
+++ b/core/src/cluster_info.rs
@@ -4765,7 +4765,8 @@ mod tests {
         })
         .take(NO_ENTRIES)
         .collect();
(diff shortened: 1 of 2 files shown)


sakridge (dismissed):
looks right

PR #16557, opened 2021-04-15
Cli: move airdrop to rpc requests

CriesofCarrots:
#### Problem
Devnet faucet requires users to specify `--faucet-host` explicitly in `solana airdrop` requests, because otherwise the default attempts to connect to the faucet directly at devnet.solana.com:9900, and the load balancer cannot forward correctly.

#### Summary of Changes
- Update solana cli to use the `requestAirdrop` rpc instead of connecting directly to the faucet, requiring:
  - The addition of `recentBlockhash` to `requestAirdrop` config object, so the client can track transaction expiration
  - Separating transaction confirmation w/ spinner for use by `solana airdrop` with that recent-blockhash
  - The addition of RpcClient methods, namely `request_airdrop_with_blockhash()`
  - Updating TestValidator to accept an optional faucet_addr so that airdrops still work in tests

### DIFF
--- a/client/src/rpc_client.rs
+++ b/client/src/rpc_client.rs
@@ -1407,7 +1407,7 @@ impl RpcClient {
         .map_err(|_| {
             RpcError::ForUser(
                 "airdrop request failed. \
-                This can happen when the address rate limit is reached."
+                This can happen when the rate limit is reached."
                     .to_string(),
             )
             .into()
--- a/client/src/rpc_client.rs
+++ b/client/src/rpc_client.rs
@@ -1557,6 +1557,24 @@ impl RpcClient {
         commitment: CommitmentConfig,
         config: RpcSendTransactionConfig,
     ) -> ClientResult<Signature> {
+        let recent_blockhash = if uses_durable_nonce(transaction).is_some() {
+            self.get_recent_blockhash_with_commitment(CommitmentConfig::processed())?
+                .value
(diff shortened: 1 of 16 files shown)


client/src/rpc_client.rs
```diff
@@ -1356,6 +1356,64 @@ impl RpcClient {
...
+        )
+        .and_then(|signature: String| {
+            Signature::from_str(&signature).map_err(|err| {
+                ClientErrorKind::Custom(format!("signature deserialization failed: {}", err)).into()
+            })
+        })
+        .map_err(|_| {
+            RpcError::ForUser(
+                "airdrop request failed. \
+                This can happen when the address rate limit is reached."
```
mvines, line 1410: Maybe we'll add more kinds of rate limits one day:
```suggestion
                This can happen when the rate limit is reached."
```

mvines (approved):
v1.5 too?  I guess it's almost time to upgrade devnet to v1.6, so if a v1.5 backport is too bad then I'm game to use this as the excuse to bump devnet to v1.6

CriesofCarrots:
> v1.5 too? I guess it's almost time to upgrade devnet to v1.6, so if a v1.5 backport is too bad then I'm game to use this as the excuse to bump devnet to v1.6

I'm hopeful it won't be too bad; I'll give it a go

PR #16573, opened 2021-04-15
uses current local timestamp when recording purged values

behzadnouri:
#### Problem
`CrdsGossipPull.purged_values` is meant to record recently purged values
so that they are excluded from imminent pull requests, until the entire
cluster have synced to the updated value:
https://github.com/solana-labs/solana/blob/c826cddbb/core/src/crds_gossip_pull.rs#L449-L454

However, `VersionedCrdsValue.local_timestamp` represents the local time
when the value was last updated, and given that crds values may have
different timeouts based on stake, it does not necessarily represent how
recently the value was purged:
https://github.com/solana-labs/solana/blob/c826cddbb/core/src/crds.rs#L75-L76

As such, recording current local timestamp when purging values is more
appropriate. Additionally, `purge_purged` assumes that the `purge_values` is
sorted in timestamps when draining the old ones; which is not true if
those timestamps are `VersionedCrdsValue.local_timestamp`:
https://github.com/solana-labs/solana/blob/c826cddbb/core/src/crds_gossip_pull.rs#L563-L571

#### Summary of Changes
* use current local timestamp when recording purged values.

### DIFF
--- a/core/src/cluster_info.rs
+++ b/core/src/cluster_info.rs
@@ -1890,14 +1890,12 @@ impl ClusterInfo {
                 error!("crds table trim failed: {:?}", err);
             }
             Ok(purged_values) => {
+                let now = timestamp();
                 self.stats
                     .trim_crds_table_purged_values_count
                     .add_relaxed(purged_values.len() as u64);
-                gossip.pull.purged_values.extend(
-                    purged_values
-                        .into_iter()
-                        .map(|v| (v.value_hash, v.local_timestamp)),
-                );
+                let purged_values = purged_values.into_iter().map(|v| (v.value_hash, now));
+                gossip.pull.purged_values.extend(purged_values);
             }
         }
     }
--- a/core/src/crds.rs
+++ b/core/src/crds.rs
@@ -73,7 +73,7 @@ pub struct VersionedCrdsValue {
     /// local time when inserted
     pub insert_timestamp: u64,
     /// local time when updated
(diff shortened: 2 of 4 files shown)


carllin:
Good catches!

So to summarize for my understanding. 

The issue is the timestamps stored for the "purged" values in the purged set are currently based on the `local_timestamp`, i.e. when those values were last updated in gossip. Because we keep around values from staked validators for an extended period of time, this means it could have been a long time since that value was updated.

The importance of this is that "purged" values timestamps are also used for purging the purged set in `purge_purged()`. Thus if the timestamps used for expiring those values from the purge set are old, then we'll immediately remove those values from the purged set. This opens us up to getting redundant entries again via pull requests shortly after.

carllin (approved):

behzadnouri:
> Good catches!
> 
> So to summarize for my understanding.
> 
> The issue is the timestamps stored for the "purged" values in the purged set are currently based on the `local_timestamp`, i.e. when those values were last updated in gossip. Because we keep around values from staked validators for an extended period of time, this means it could have been a long time since that value was updated.
> 
> The importance of this is that "purged" values timestamps are also used for purging the purged set in `purge_purged()`. Thus if the timestamps used for expiring those values from the purge set are old, then we'll immediately remove those values from the purged set. This opens us up to getting redundant entries again via pull requests shortly after.

yes, that is basically it

PR #16576, opened 2021-04-15
retains crds values if the origin is still active

behzadnouri:
#### Problem
Local timestamps are updated for records associated with a pubkey if the
origin is still active:
https://github.com/solana-labs/solana/blob/c8ed14c64/core/src/crds.rs#L301-L311

However this is done inconsistently on some gossip paths (pull requests
and pull responses) but not all (e.g. push messages). Additionally
update_record_timestamp is inefficient since there can be ~800 values
associated with each pubkey.

This commit updates records timestamps only on contact-infos; and,
instead utilizes origin's timestamp when purging old values.

#### Summary of Changes
* Update record timestamp only on contact-info if it exists.
* Utilize origin's timestamp when purging old values.
* Retain crds values if the origin is still active.


carllin (approved):
Nice!

Small disambiguation on the problem description: by "origin timestamp" we don't mean the timestamp marked by the original sender, we mean the timestamp from the last refresh on the origin's `CrdsValueLabel::ContactInfo`, which is now refreshed on every message update from that origin.

behzadnouri:
> Small disambiguation on the problem description: by "origin timestamp" we don't mean the timestamp marked by the original sender, we mean the timestamp from the last refresh on the origin's `CrdsValueLabel::ContactInfo`, which is now refreshed on every message update from that origin.

yeah, that is correct

PR #16581, opened 2021-04-15
improve failing assert

jeffwashington:
#### Problem
assert text wasn't helpful. And, I didn't know you could match tests on substrings of assert text.
#### Summary of Changes
parameterize assert text
Fixes #


sakridge (approved):
lgtm

PR #16582, opened 2021-04-15
fix an assertion that might cause panic in debug build

taozhu-chicago:
#### Problem
`banking-bench` panics in debug build when running on macBook: 
```% cargo run
   Compiling solana-banking-bench v1.7.0 (/Users/taozhu/Solana/repo/solana/banking-bench)
    Finished dev [unoptimized + debuginfo] target(s) in 31.76s
     Running `/Users/taozhu/Solana/repo/solana/target/debug/solana-banking-bench`
thread 'main' panicked at 'assertion failed: txs_processed < bank.transaction_count()', banking-bench/src/main.rs:305:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

It runs fine in release build on same laptop. 

#### Summary of Changes
The issue is likely due to debug build runs much slower, resulting no new transactions being added to bank during some iteration. Should add this as acceptable condition.

Fixes #
To assert processed transaction count should be less *or equal* to bank transaction count.

### DIFF
--- a/banking-bench/src/main.rs
+++ b/banking-bench/src/main.rs
@@ -148,7 +148,7 @@ fn main() {
             Arg::with_name("num_threads")
                 .long("num-threads")
                 .takes_value(true)
-                .help("Number of iterations"),
+                .help("Number of threads"),
         )
         .get_matches();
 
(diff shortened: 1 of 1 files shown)


taozhu-chicago:
kudos to @sakridge

Issue #16570, opened 2021-04-15
tip of master panic!()-ed with mainnet-beta's snapshot


ryoqun:
#### Problem

maybe @jeffwashington 's #16310?

```
thread 'main' panicked at 'out of range', runtime/src/accounts_index.rs:208:9
stack backtrace:
   0: std::panicking::begin_panic
             at /home/ryoqun/.rustup/toolchains/nightly-2021-02-18-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/panicking.rs:519:12
   1: solana_runtime::accounts_index::RollingBitField::check_range
             at /home/ryoqun/work/solana/solana/runtime/src/accounts_index.rs:208:9
   2: solana_runtime::accounts_index::RollingBitField::insert
             at /home/ryoqun/work/solana/solana/runtime/src/accounts_index.rs:217:9
   3: solana_runtime::accounts_index::AccountsIndex<T>::add_root
             at /home/ryoqun/work/solana/solana/runtime/src/accounts_index.rs:1174:9
   4: solana_runtime::accounts_db::AccountsDb::generate_index
             at /home/ryoqun/work/solana/solana/runtime/src/accounts_db.rs:4911:13
   5: solana_runtime::serde_snapshot::reconstruct_accountsdb_from_fields
             at /home/ryoqun/work/solana/solana/runtime/src/serde_snapshot.rs:365:5
   6: solana_runtime::serde_snapshot::reconstruct_bank_from_fields
             at /home/ryoqun/work/solana/solana/runtime/src/serde_snapshot.rs:244:27
   7: solana_runtime::serde_snapshot::bank_from_stream
             at /home/ryoqun/work/solana/solana/runtime/src/serde_snapshot.rs:156:30
   8: solana_runtime::snapshot_utils::rebuild_bank_from_snapshots::{{closure}}
             at /home/ryoqun/work/solana/solana/runtime/src/snapshot_utils.rs:801:40
   9: solana_runtime::snapshot_utils::deserialize_snapshot_data_file_capped
             at /home/ryoqun/work/solana/solana/runtime/src/snapshot_utils.rs:488:15
  10: solana_runtime::snapshot_utils::deserialize_snapshot_data_file
             at /home/ryoqun/work/solana/solana/runtime/src/snapshot_utils.rs:436:5
  11: solana_runtime::snapshot_utils::rebuild_bank_from_snapshots
[...]

PR #16583, opened 2021-04-15
increase rolling slot width to 4M

jeffwashington:
#### Problem
#16570

#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/accounts_index.rs
+++ b/runtime/src/accounts_index.rs
@@ -302,9 +302,9 @@ pub struct RootsTracker {
 impl Default for RootsTracker {
     fn default() -> Self {
         // we expect to keep a rolling set of 400k slots around at a time
-        // 2M gives us plenty of extra room to handle a width 5x what we should need.
-        // cost is 2M bits of memory
-        RootsTracker::new(2097152)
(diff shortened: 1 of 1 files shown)


jeffwashington:
I confirmed I could start a mainnet-beta validator using ryo's snapshot with this increase to 4M.

sakridge:
Do we know that 4m will be enough?

jeffwashington:
> Do we know that 4m will be enough?

I don't yet know if this is enough. This is harmless and gets around the current situation. But, I'm not ready to submit this yet. I'm debugging shortly the actual situation in the snapshot. I also have a full revert pr ready in case we want to go that route.

sakridge:
> > Do we know that 4m will be enough?
> 
> I don't yet know if this is enough. This is harmless and gets around the current situation. But, I'm not ready to submit this yet. I'm debugging shortly the actual situation in the snapshot. I also have a full revert pr ready in case we want to go that route.

ok, sounds good.

jeffwashington:
@carllin is running into this, too. 4M is enough to get around the current issues and still well within the performance envelope while we figure out why old stores/accounts aren't getting cleaned.

PR #16602, opened 2021-04-16
Trim extra shred bytes in blockstore

### LINKED ISSUE #16236: Data shreds are stored with 0 padding in rocksdb
#### Problem

Data shreds are stored padded with 0s in rocksdb which makes rocksdb larger than it should be

#### Proposed Solution

Don't store the 0-padded shreds. Use the packet meta data to see how large the shreds are.

steviez:
#### Problem
Data shred bytestreams are currently inserted into the blockstore with padded 0's. Every data shred has at least some padded 0's due to there being a "restricted" section at the end of a data shred's payload:
https://github.com/solana-labs/solana/blob/master/ledger/src/shred.rs#L44-L50

Furthermore, data shreds that are not filled to capacity with data will have additional 0-padding to get to fill up the packet.

The result of these two items is that the blockstore is getting bloated with extraneous bytes.

#### Summary of Changes
Trim the extra padding bytes off of the shred on insertion; "re-expand" the bytestream when retrieving from the blockstore to avoid breaking the erasure coding algorithm (which requires that coding and data shreds are of same length).

Fixes #16236

### DIFF
--- a/core/src/serve_repair.rs
+++ b/core/src/serve_repair.rs
@@ -939,6 +939,7 @@ mod tests {
             assert_eq!(shreds[0].slot(), 1);
             assert_eq!(shreds[0].index(), 0);
             shreds[0].payload.push(10);
+            shreds[0].data_header.size = shreds[0].payload.len() as u16;
             blockstore
                 .insert_shreds(shreds, None, false)
                 .expect("Expect successful ledger write");
--- a/core/src/serve_repair.rs
+++ b/core/src/serve_repair.rs
@@ -611,9 +611,7 @@ mod tests {
     use solana_ledger::{
         blockstore::make_many_slot_entries,
         blockstore_processor::fill_blockstore_slot_with_ticks,
-        shred::{
-            max_ticks_per_n_shreds, CodingShredHeader, DataShredHeader, Shred, ShredCommonHeader,
-        },
+        shred::{max_ticks_per_n_shreds, Shred},
     };
     use solana_perf::packet::Packet;
     use solana_sdk::{hash::Hash, pubkey::Pubkey, timing::timestamp};
@@ -726,23 +724,10 @@ mod tests {
                 nonce,
             );
             assert!(rv.is_none());
-            let common_header = ShredCommonHeader {
-                slot,
-                index: 1,
-                ..ShredCommonHeader::default()
-            };
-            let data_header = DataShredHeader {
-                parent_offset: 1,
-                ..DataShredHeader::default()
-            };
-            let shred_info = Shred::new_empty_from_header(
-                common_header,
-                data_header,
-                CodingShredHeader::default(),
-            );
+            let shred = Shred::new_from_data(slot, 1, 1, None, false, false, 0, 2, 0);
 
             blockstore
-                .insert_shreds(vec![shred_info], None, false)
+                .insert_shreds(vec![shred], None, false)
                 .expect("Expect successful ledger write");
 
             let index = 1;
--- a/ledger/src/blockstore.rs
+++ b/ledger/src/blockstore.rs
@@ -1281,6 +1281,7 @@ impl Blockstore {
         leader_schedule: Option<&Arc<LeaderScheduleCache>>,
         is_recovered: bool,
     ) -> bool {
+        use crate::shred::SHRED_PAYLOAD_SIZE;
         let shred_index = u64::from(shred.index());
         let slot = shred.slot();
         let last_in_slot = if shred.last_in_slot() {
@@ -1291,6 +1292,9 @@ impl Blockstore {
         };
 
         if shred.data_header.size == 0 {
+            return false;
+        }
+        if shred.payload.len() > SHRED_PAYLOAD_SIZE {
             return false;
         }
 
@@ -1381,7 +1385,6 @@ impl Blockstore {
         shred: &Shred,
         write_batch: &mut WriteBatch,
     ) -> Result<Vec<(u32, u32)>> {
-        use crate::shred::SHRED_PAYLOAD_SIZE;
         let slot = shred.slot();
         let index = u64::from(shred.index());
 
@@ -1415,14 +1418,6 @@ impl Blockstore {
 
         // Commit step: commit all changes to the mutable structures at once, or none at all.
         // We don't want only a subset of these changes going through.
-        if shred.payload.len() > SHRED_PAYLOAD_SIZE {
-            return Err(BlockstoreError::InvalidShredData(Box::new(
-                bincode::ErrorKind::Custom(format!(
-                    "Data shred with slot {}, index {} has payload that is too large to be valid.",
-                    slot, index
-                )),
-            )));
-        }
         write_batch.put_bytes::<cf::ShredData>(
             (slot, index),
             // Payload will be padded out to SHRED_PAYLOAD_SIZE
@@ -5302,6 +5297,23 @@ pub mod tests {
             blockstore
                 .insert_shreds(shreds[0..5].to_vec(), None, false)
                 .unwrap();
+
+            let slot_meta = blockstore.meta(0).unwrap().unwrap();
+            // Corrupt shred by making it too large
+            let mut shred5 = shreds[5].clone();
+            shred5.payload.push(10);
+            shred5.data_header.size = shred5.payload.len() as u16;
+            assert_eq!(
+                blockstore.should_insert_data_shred(
+                    &shred5,
+                    &slot_meta,
+                    &HashMap::new(),
+                    &last_root,
+                    None,
+                    false
+                ),
+                false
+            );
 
             // Trying to insert another "is_last" shred with index < the received index should fail
             // skip over shred 7
--- a/ledger/src/blockstore.rs
+++ b/ledger/src/blockstore.rs
@@ -1290,6 +1290,10 @@ impl Blockstore {
             false
         };
 
+        if shred.data_header.size == 0 {
+            return false;
+        }
+
         // Check that we do not receive shred_index >= than the last_index
         // for the slot
         let last_index = slot_meta.last_index;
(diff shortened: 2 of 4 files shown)


steviez:
@sakridge - I ended up cherry-picking these commits into my own branch since they were failing the CI before; it appears those failures were either unrelated issues that were fixed or a non-deterministic failure.

Regardless of RocksDB vs. AccountsDB, I think we still want this change to keep our backend datastore as slim as possible, right?

Assuming so, thoughts on any additional testing that should be done for validation of this PR? The `.resize()` on the vector seems like the only thing that could affect performance

sakridge:
> @sakridge - I ended up cherry-picking these commits into my own branch since they were failing the CI before; it appears those failures were either unrelated issues that were fixed or a non-deterministic failure.
> 
> Regardless of RocksDB vs. AccountsDB, I think we still want this change to keep our backend datastore as slim as possible, right?
> 
> Assuming so, thoughts on any additional testing that should be done for validation of this PR? The `.resize()` on the vector seems like the only thing that could affect performance
> 
> Edit: I'm pushing in one more commit to add a comment or two so CI will re-run, but everything had passed:

Yep, we still want it, exactly as you said to keep the db as small as possible.

steviez:
Here are the performance run results:
https://solanalabs.slack.com/archives/CP2L2S4KV/p1618955757009100

```
mean_tps: 21020
max_tps: 50685.5
mean_confirmation_ms: 2244
max_confirmation_ms: 20425
99th_percentile_confirmation_ms: 10463
max_tower_distance: 164
last_tower_distance: 63.5
slots_per_second: 2.041
```
Does seem to be worse than the recent runs on master's latest (which are hitting 55k+ max TPS)

sakridge:
> Here are the performance run results:
> https://solanalabs.slack.com/archives/CP2L2S4KV/p1618955757009100
> 
> ```
> mean_tps: 21020
> max_tps: 50685.5
> mean_confirmation_ms: 2244
> max_confirmation_ms: 20425
> 99th_percentile_confirmation_ms: 10463
> max_tower_distance: 164
> last_tower_distance: 63.5
> slots_per_second: 2.041
> ```
> 
> Does seem to be worse than the recent runs on master's latest (which are hitting 55k+ max TPS)

I think it might be within the noise of this measurement. You could compare the specific `insert_shred` or the `fetch_entries` part of replay stats. Those correspond to when we store or read shreds from the db.

ledger/src/blockstore.rs
```diff
@@ -1294,6 +1294,10 @@ impl Blockstore {
             false
         };
 
+        if shred.data_header.size == 0 {
```
sakridge, line 1297: @carllin think we need cutoff for this? I don't think we create a len=0 shred in the normal code paths.

sakridge (dismissed):
lgtm

steviez:
> I think it might be within the noise of this measurement. You could compare the specific `insert_shred` or the `fetch_entries` part of replay stats. Those correspond to when we store or read shreds from the db.

So I dug a little deeper on this and did a comparison of [my run](https://metrics.solana.com:3000/d/monitor-edge/cluster-telemetry-edge?orgId=2&var-testnet=gce-perf-cpu-only&from=1618951216752&to=1618955757115) with the previous [nightly run](https://metrics.solana.com:3000/d/monitor-edge/cluster-telemetry-edge?orgId=2&var-testnet=gce-perf-cpu-only&from=1618877577220&to=1618882199624). For the next two graphs, the first two humps are the previous two nightly runs; the third hump is my run.

Here is insert:
image
and here is fetch:
image

Digging deeper on the graphana graphs, the data is pretty "jagged" so think I would agree that any observed difference is within the noise, so I'm happy if you are

carllin:
nit: `Summary of Changes` section seems to have been cut off 😃

ledger/src/blockstore.rs
```diff
@@ -1442,7 +1451,16 @@ impl Blockstore {
...
 
     pub fn get_data_shred(&self, slot: Slot, index: u64) -> Result<Option<Vec<u8>>> {
-        self.data_shred_cf.get_bytes((slot, index))
+        use crate::shred::SHRED_PAYLOAD_SIZE;
+        self.data_shred_cf.get_bytes((slot, index)).map(|data| {
+            data.map(|mut d| {
+                // Only data_header.size bytes stored in the blockstore so
+                // pad the payload out to SHRED_PAYLOAD_SIZE so that the
+                // erasure recovery works properly.
+                d.resize(cmp::max(d.len(), SHRED_PAYLOAD_SIZE), 0);
```
carllin, line 1460: should there also be a check on insert that the size doesn't exceed `SHRED_PAYLOAD_SIZE`?

ledger/src/blockstore.rs
carllin: Yeah, you're thinking it might actually be useful to store the zero length shred for duplicate detection? 

If we store it most likely I think replay will call `get_slot_entries()`, the best case scenario is if the validator then can't recreate the entries and mark the slot as dead. 

This brings up an interesting point though. I wonder if it's possible to use zero-length shreds as filler to recreate the same entries, but with a different set of shreds, i.e. instead of splitting an entry amount `N` shreds, you spread it out between `N + Z` shreds where `Z` is some number of empty shreds (could also do this without zero-length shreds I think).

I'm guessing an attacker could do this with just the last couple of shreds to optimize their chances that some validators getting all of one version or another (you don't want a mix of versions otherwise they won't deserialize properly).

Replay might actually play that alternate version of the shreds without a hitch. As long as you end up with the right state though, it should be fine? It might make repair a little wonky if some people are serving one version of a slot, and others serving another version though.

ledger/src/blockstore.rs
steviez: Yeah, I could see that being a reasonable check to make. Although, I think we'd actually want to compare to `SIZE_OF_DATA_SHRED_PAYLOAD` on insert; because we know we there shouldn't be anything in the "restricted" area at the end

steviez: Oops, CI helped me see you were right :). `SIZE_OF_DATA_SHRED_PAYLOAD` excludes the headers, confused myself there a bit ... thank goodness for unit tests

core/src/serve_repair.rs
```diff
@@ -940,66 +925,6 @@ mod tests {
...
-                &socketaddr_any!(),
-                nonce,
-            )
-            .unwrap()];
-            assert_eq!(rv, expected);
-        }
-
-        Blockstore::destroy(&ledger_path).expect("Expected successful database destruction");
-    }
-
```
steviez, line 1002: @carllin - This test previously "corrupted" a shred by add an extra byte which made it unable to fit in a packet. `repair_response_pakcet(...)` would return `None` only when the `repair_response_packet(...)` eventually makes it to this snippet:
```
pub fn repair_response_packet_from_shred(
    shred: Vec<u8>,
    dest: &SocketAddr,
    nonce: Nonce,
) -> Option<Packet> {
    let mut packet = Packet::default();
    packet.meta.size = shred.len() + SIZE_OF_NONCE;
    if packet.meta.size > packet.data.len() {
        return None;
    }
```

With the addition of the size check on insertion into the blockstore, this is no longer possible (and so the test was failing). With that in mind, this tests appears to have been redundant to the test above it, but let me know what you think

carllin: hmmm by "size check", I think you mean the newly added truncation of the data shred here: https://github.com/solana-labs/solana/pull/16602/files#diff-01b72272fd1033dcbb8fed13821a6232d35913119af6be20a7c2d1e9c772feb9R1422 ?

With these changes, the shred is successfully inserted because its truncated down to `shred.data_header.size`, regardless of the actual data size, and when we fetch from the blockstore later we get a "correctly" sized shred. 

I think if we 
1) Did this https://github.com/solana-labs/solana/pull/16602/files#r617935210,
2) Expand the `shred.data_header.size` of the corrupted shred in this test to trigger that above check

Then this test would succeed because the corrupted shred would never be inserted into Blockstore. That way we can keep this test that ensures corrupted shreds don't make it into repair. Wdyt?

steviez: > hmmm by "size check", I think you mean the newly added truncation of the data shred here: https://github.com/solana-labs/solana/pull/16602/files#diff-01b72272fd1033dcbb8fed13821a6232d35913119af6be20a7c2d1e9c772feb9R1422 ?

Yes, apologies for the loose wording on my side

> I think if we
> 1. Did this https://github.com/solana-labs/solana/pull/16602/files#r617935210,
> 2. Expand the shred.data_header.size of the corrupted shred in this test to trigger that above check
>
> Then this test would succeed because the corrupted shred would never be inserted into Blockstore. That way we can keep this test that ensures corrupted shreds don't make it into repair. Wdyt?

I had it in my mind that we "had to" insert the shred in this instance. But, truncating a corrupted shred to make it "fit" would do more harm than good. I see what you meant now (I think) with your original comment. See:
https://github.com/solana-labs/solana/pull/16602/commits/35b4da4f5b39990dc3fa3d429161315c633eaca2#diff-01b72272fd1033dcbb8fed13821a6232d35913119af6be20a7c2d1e9c772feb9R1418-R1425

steviez:
> Would it make more sense to fold this into `test_should_insert_data_shred`? That way we can also add a test case in `test_should_insert_data_shred`

Yeah, I think you're right. We already have another similar validity check in `should_insert_data_shred` and makes sense to check this as early as possible

carllin (approved):

ledger/src/blockstore.rs
behzadnouri: I am not sure this change was backward compatible.
broadcast has to create zero length shreds when the previous slot was interrupted:
https://github.com/solana-labs/solana/blob/ca7b36ad8/core/src/broadcast_stage/standard_broadcast_run.rs#L79

`shred data == None` will result in zero length data shred.

steviez: Per some back and forth on Discord, we believe this is a non-issue as the the headers will take up some space. Behzad suggested adding a unit test to protect the behavior we rely on; that test is added here:
https://github.com/solana-labs/solana/pull/16955

carllin: awesome, thanks for reasoning through it guys 😃

PR #16603, opened 2021-04-16
Move derivation path into sdk

CriesofCarrots:
#### Problem
In order to support bip32 and bip44 Keypair derivation, we will need to be able parse derivation path inputs of Signers beyond RemoteWallets.
This will also make it easier to feature-gate remote-wallet/hidapi support, when desirable.

#### Summary of Changes
Move DerivationPath stucts and helpers into sdk
Also update remote-wallet format to single `use` statement

Toward #5246


t-nelson (approved):
lgtm!

PR #16610, opened 2021-04-17
read only accounts cache uses accurate size representation

jeffwashington:
#### Problem
Large #s of accounts with small data will end up using far more memory than the budget for the read only accounts cache.

#### Summary of Changes
Account for true storage size per account in the read only cache.

Fixes #

### DIFF
--- a/runtime/src/read_only_accounts_cache.rs
+++ b/runtime/src/read_only_accounts_cache.rs
@@ -50,6 +50,7 @@ impl ReadOnlyAccountsCache {
     }
 
     fn per_account_size() -> usize {
+        // size_of(arc(x)) does not return the size of x, so we have to add the size of RwLock...
         std::mem::size_of::<ReadOnlyAccountCacheEntry>() + std::mem::size_of::<RwLock<Instant>>()
     }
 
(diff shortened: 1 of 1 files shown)


jeffwashington:
@carllin I'd like to do the same for write cache, but I haven't looked hard enough yet to make sure I found the right place(s)

runtime/src/read_only_accounts_cache.rs
```diff
@@ -44,9 +45,14 @@ impl ReadOnlyAccountsCache {
             hits: AtomicU64::new(0),
             misses: AtomicU64::new(0),
             lru: Arc::new(RwLock::new(Vec::new())),
+            per_account_size: Self::per_account_size(),
         }
     }
 
+    fn per_account_size() -> usize {
+        std::mem::size_of::<ReadOnlyAccountCacheEntry>() + std::mem::size_of::<RwLock<Instant>>()
```
carllin, line 53: Could we add a comment about why we need to add and extra `RwLock<Instant>` size to `ReadOnlyAccountCacheEntry>`? Looks like `ReadOnlyAccountCacheEntry` already has a `last_used: RwLock<Instant>` field?

carllin (approved):

PR #16622, opened 2021-04-17
replay_stage: send votes to next leader's TPU instead of our TPU

mvines:
ReplayStage currently sends votes to the node's TPU, which in almost all cases guarantees that the vote will need to be forwarded by BankingStage to the next leader.  Avoid this hop and send votes directly at the next leader from ReplayStage

### DIFF
--- a/core/src/banking_stage.rs
+++ b/core/src/banking_stage.rs
@@ -566,15 +566,8 @@ impl BankingStage {
             }
             return;
         }
-        let next_leader = match poh_recorder
-            .lock()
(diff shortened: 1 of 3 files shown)


mvines:
@sakridge - wdyt?  I have no proof this helps anything.  But it just seems to be the right thing to do

sakridge:
> @sakridge - wdyt? I have no proof this helps anything. But it just seems to be the right thing to do

Yea, it should be better.

PR #16623, opened 2021-04-18
setup-from-mainnet-beta: Update snapshot download URL

mvines:
snapshot download doesn't work from http://api.mainnet-beta.solana.com anymore, probably due to the new load balancer config.


mvines:
🤔 a better solution might be to rip out all the `curl`ing and instead use `solana-validator init` to download a mainnet-beta snapshot

PR #16635, opened 2021-04-19
`solana validators` now shows current epoch credits instead of lifetime credits

mvines:
Lifetime credits are a fun factoid that nobody really cares about.  What you really care about is the credits earned in the current epoch, as this is what influences staking rewards.

### DIFF
--- a/cli-output/src/cli_output.rs
+++ b/cli-output/src/cli_output.rs
@@ -338,7 +338,7 @@ impl fmt::Display for CliValidators {
 
             writeln!(
                 f,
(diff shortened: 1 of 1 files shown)


brianlong:
@mvines This is a breaking change for me. Can you get this into both v1.5 & v1.6 at the same time? That way I won't need conditional logic in validators.app

mvines:
> @mvines This is a breaking change for me. Can you get this into both v1.5 & v1.6 at the same time? That way I won't need conditional logic in validators.app

Oh thanks!  Would you prefer that I back this out?  Or enable lifetime credits with a flag?  Otherwise sure I can backport to v1.5.

Does https://github.com/solana-labs/solana/pull/16639 cause you any concern?

PR #16636, opened 2021-04-19
log roots range metric

jeffwashington:
#### Problem
On some validators, old roots are not getting cleaned. We now track min root, so we can calculate a range. The range of roots could be useful information.

#### Summary of Changes
Log the range of roots that are currently in memory on a validator.

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -950,6 +950,7 @@ struct LatestAccountsIndexRootsStats {
     roots_len: AtomicUsize,
     uncleaned_roots_len: AtomicUsize,
     previous_uncleaned_roots_len: AtomicUsize,
+    roots_range: AtomicU64,
 }
 
 impl LatestAccountsIndexRootsStats {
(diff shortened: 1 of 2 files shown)


runtime/src/accounts_index.rs
```diff
@@ -218,6 +218,11 @@ impl RollingBitField {
         );
     }
 
+    pub fn range_loose(&self) -> u64 {
```
carllin, line 221: nit, rename to range_width?

runtime/src/accounts_db.rs
```diff
@@ -984,6 +987,11 @@ impl LatestAccountsIndexRootsStats {
                 self.previous_uncleaned_roots_len.load(Ordering::Relaxed) as i64,
                 i64
             ),
+            (
+                "roots_range",
```
carllin, line 991: nit: rename to`roots_range_width`

carllin (approved):

jeffwashington:
some log entries:
datapoint: accounts_index_roots_len roots_len=391303i uncleaned_roots_len=1i previous_uncleaned_roots_len=391302i roots_range_width=1791086i not_a_root_cleaned=0i cleaned_roots=69674i
datapoint: accounts_index_roots_len roots_len=391270i uncleaned_roots_len=1i previous_uncleaned_roots_len=0i roots_range_width=1791086i not_a_root_cleaned=0i cleaned_roots=33i

PR #16642, opened 2021-04-19
cargo-install-all.sh: on Mac OS X, use greadlink instead of readlink

cdrappi:
#### Problem
Installing the Solana CLI from source on Mac fails because Mac's version of `readlink` does not support the `-f` flag

#### Summary of Changes
On Macs, use `greadlink`, which is a tool that directly mirrors Linux `readlink` functionality on Mac. Building `v1.6.6` works locally using the script after this change

Fixes #
No issues, this is just a minor thing I came across when building from source


mvines:
Thanks

PR #16646, opened 2021-04-19
removes first_coding_index from erasure recovery code

behzadnouri:
#### Problem
`first_coding_index` is the same as the `set_index` and is so redundant:
https://github.com/solana-labs/solana/blob/37b8587d4/ledger/src/blockstore_meta.rs#L49-L60

#### Summary of Changes
* removed `first_coding_index`

### DIFF
--- a/ledger/src/blockstore_meta.rs
+++ b/ledger/src/blockstore_meta.rs
@@ -51,8 +51,9 @@ pub struct ShredIndex {
 pub struct ErasureMeta {
     /// Which erasure set in the slot this is
     pub set_index: u64,
-    /// First coding index in the FEC set
-    pub first_coding_index: u64,
+    /// Deprecated field.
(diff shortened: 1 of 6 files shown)


ledger/src/blockstore_meta.rs
```diff
@@ -51,8 +51,6 @@ pub struct ShredIndex {
 pub struct ErasureMeta {
     /// Which erasure set in the slot this is
     pub set_index: u64,
-    /// First coding index in the FEC set
-    pub first_coding_index: u64,
```
carllin, line 55: hmmm since these are stored in the database, is this going to cause compatibility issues if we try to fetch old ErasureMeta's?

behzadnouri: you are right. fixed

carllin (approved):

Issue #16596, opened 2021-04-16
RPC: pubsub defaults to confirmed but rpc http defaults to finalized


jstarry:
#### Problem
The default commitment for pubsub and http requests are different which can cause consistency issues in downstream clients.

This problem manifests itself in the web3 sdk. When a commitment is not specified for a `Connection`, an account info request will default to finalized but confirming a transaction will default to confirmed because it uses pubsub to subscribe to confirmation. This causes consistency problems in apps that may immediately fetch account state after confirming a transaction. The out of the box configuration shouldn't be error prone like this.

#### Proposed Solution
Default everything to confirmed?

Problem is that this is a breaking change and some clients may be surprised about the defaults changing. However, one could argue that any application requiring full finalization, should already be explicit about that in commitment parameters and not rely on defaults.

tag @mvines @CriesofCarrots

ryoqun:
oh seems like another read consistency issue... allow me to tag.. #15705

mvines:
> Default everything to confirmed?

I think we still want to be conservative and default to finalized for everything, 100% safety by default

>  However, one could argue that any application requiring full finalization, should already be explicit about that in commitment parameters and not rely on defaults.

Pretty sure this'll break some exchanges that already assume finalized by default

jstarry:
Ok, by being conservative, developers opt into lower levels of safety if they wish but by default get a slow but safe experience. We should consolidate in the other direction then and change the Pubsub default to finalized

mvines:
> We should consolidate in the other direction then and change the Pubsub default to finalized

100%

PR #16659, opened 2021-04-20
RPC: use finalized as default pubsub commitment level

jstarry:
#### Problem
The default commitment for pubsub and http requests are different which can cause consistency issues in downstream clients.

#### Summary of Changes
Change the default commitment of pubsub subscriptions from confirmed to finalized

Fixes https://github.com/solana-labs/solana/issues/16596

### DIFF
--- a/core/src/rpc_subscriptions.rs
+++ b/core/src/rpc_subscriptions.rs
@@ -630,9 +630,7 @@ impl RpcSubscriptions {
         subscriber: Subscriber<Response<UiAccount>>,
     ) {
         let config = config.unwrap_or_default();
-        let commitment = config
-            .commitment
-            .unwrap_or_else(CommitmentConfig::confirmed);
+        let commitment = config.commitment.unwrap_or_default();
 
         let slot = if commitment.is_finalized() {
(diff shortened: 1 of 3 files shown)


mvines:
Minor wrinkle here I guess is that we explicitly document the default as "confirmed" :-/
https://github.com/solana-labs/solana/search?q=%22For+subscriptions%2C+if+commitment+is+unspecified%2C+the+default+value+is%22

I think it's fair to call this a bug, but it should be called out as a breaking change in the release notes.  Does web3.js have anything to say about the default in comments anywhere?

jstarry:
> Minor wrinkle here I guess is that we explicitly document the default as "confirmed" :-/

Dang, will fix

> I think it's fair to call this a bug, but it should be called out as a breaking change in the release notes. 

Agreed on both points

> Does web3.js have anything to say about the default in comments anywhere?

Nope, it even specifies that the "max" is the default here: https://solana-labs.github.io/solana-web3.js/modules.html#sendandconfirmtransaction

mvines (approved):

PR #16676, opened 2021-04-20
Allow SetUpgradeAuthority instruction in CPI calls

SebastianBor:
#### Problem

For Governance program we want to set the upgrade authority of the program we take under governance.
The upgrade authority should be transferred to the Governance Program Derived Address to let it upgrade the governed program and prevent anybody else but governance from doing so.
It also works as a security check ensuring the governance instruction is signed by the current program upgrade authority and prevents from setting governance for somebody's else program.

It's currently not possible to change upgrade authority via CPI and the following error is returned: 
`"Program failed to complete: Program BPFLoaderUpgradeab1e11111111111111111111111 not supported by inner instructions"`

#### Summary of Changes

Add SetAuthority to instructions authorised for CPI invocation

### DIFF
--- a/programs/bpf/tests/programs.rs
+++ b/programs/bpf/tests/programs.rs
@@ -2130,7 +2130,7 @@ fn test_program_bpf_set_upgrade_authority_via_cpi() {
         "solana_bpf_rust_upgradeable",
     );
 
-    // Set program upgrade authority via CPI
+    // Set program upgrade authority instruction to invoke via CPI
     let new_upgrade_authority_key = Keypair::new().pubkey();
     let mut set_upgrade_authority_instruction = bpf_loader_upgradeable::set_upgrade_authority(
         &program_id,
--- a/programs/bpf/tests/programs.rs
+++ b/programs/bpf/tests/programs.rs
@@ -2087,6 +2087,65 @@ fn test_program_bpf_upgrade_self_via_cpi() {
 
 #[cfg(feature = "bpf_rust")]
 #[test]
+fn test_program_bpf_set_upgrade_authority_via_cpi() {
+    solana_logger::setup();
+
+    let GenesisConfigInfo {
+        genesis_config,
+        mint_keypair,
+        ..
+    } = create_genesis_config(50);
+    let mut bank = Bank::new(&genesis_config);
+    let (name, id, entrypoint) = solana_bpf_loader_program!();
+    bank.add_builtin(&name, id, entrypoint);
+    let (name, id, entrypoint) = solana_bpf_loader_upgradeable_program!();
+    bank.add_builtin(&name, id, entrypoint);
+    let bank_client = BankClient::new(bank);
+    let invoke_and_return = load_bpf_program(
+        &bank_client,
+        &bpf_loader::id(),
+        &mint_keypair,
+        "solana_bpf_rust_invoke_and_return",
+    );
+
+    // Deploy upgradeable program
+    let buffer_keypair = Keypair::new();
+    let program_keypair = Keypair::new();
+    let program_id = program_keypair.pubkey();
+    let authority_keypair = Keypair::new();
+    load_upgradeable_bpf_program(
+        &bank_client,
+        &mint_keypair,
+        &buffer_keypair,
+        &program_keypair,
+        &authority_keypair,
+        "solana_bpf_rust_upgradeable",
+    );
+
+    // Set program upgrade authority via CPI
+    let new_authority_keypair = Keypair::new();
+    let mut set_upgrade_authority_instruction = bpf_loader_upgradeable::set_upgrade_authority(
+        &program_id,
+        &authority_keypair.pubkey(),
+        Some(&new_authority_keypair.pubkey()),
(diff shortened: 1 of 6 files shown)


sdk/program/src/bpf_loader_upgradeable.rs
```diff
@@ -192,6 +192,14 @@ pub fn is_upgrade_instruction(instruction_data: &[u8]) -> bool {
     3 == instruction_data[0]
 }
 
+pub fn is_set_authority_instruction(instruction_data: &[u8]) -> bool {
+    4 == instruction_data[0]
+}
+
+pub fn is_cpi_authorized_instruction(instruction_data: &[u8]) -> bool {
```
jackcmay, line 199: Get rid of this and move this logic to syscalls.rs, would like to keep the decision logic in one place

programs/bpf_loader/src/syscalls.rs
```diff
@@ -1684,7 +1684,7 @@ fn check_authorized_program(
         || bpf_loader::check_id(program_id)
         || bpf_loader_deprecated::check_id(program_id)
         || (bpf_loader_upgradeable::check_id(program_id)
-            && !bpf_loader_upgradeable::is_upgrade_instruction(instruction_data))
+            && !bpf_loader_upgradeable::is_cpi_authorized_instruction(instruction_data))
```
jackcmay, line 1687: Needs to be feature gated

jackcmay:
How about some tests that exercise the `upgrade_authority` instruction via CPI?

sdk/program/src/bpf_loader_upgradeable.rs
SebastianBor: Done in 454931c

programs/bpf/tests/programs.rs
```diff
@@ -2085,6 +2085,65 @@ fn test_program_bpf_upgrade_self_via_cpi() {
...
+        .accounts
+        .insert(0, AccountMeta::new(bpf_loader_upgradeable::id(), false));
+
+    let message = Message::new(
+        &[set_upgrade_authority_instruction],
+        Some(&mint_keypair.pubkey()),
+    );
+    bank_client
+        .send_and_confirm_message(&[&mint_keypair, &authority_keypair], message)
+        .unwrap();
```
jackcmay, line 2144: Could query the account and verify that the authority has been set successfully?

programs/bpf_loader/src/syscalls.rs
SebastianBor: Done in 05caf3b8f35ca993b31f3610d4f1902718056639 please check if that's correct

programs/bpf_loader/src/syscalls.rs
```diff
@@ -1679,12 +1679,16 @@ fn check_account_infos(
...
     program_id: &Pubkey,
     instruction_data: &[u8],
+    invoke_context: &Ref<&mut dyn InvokeContext>,
 ) -> Result<(), EbpfError<BpfError>> {
     if native_loader::check_id(program_id)
         || bpf_loader::check_id(program_id)
         || bpf_loader_deprecated::check_id(program_id)
         || (bpf_loader_upgradeable::check_id(program_id)
-            && !bpf_loader_upgradeable::is_upgrade_instruction(instruction_data))
+            && !(bpf_loader_upgradeable::is_upgrade_instruction(instruction_data)
```
SebastianBor, line 1688: Is this cascading condition ok should I extract it into a helper function?

jackcmay: I'm fine with this, still very readable in my opinion

jackcmay: After the feature is active for > 2 epochs on master we can remove the featurization

sdk/src/feature_set.rs
```diff
@@ -75,6 +75,10 @@ pub mod bpf_loader_upgradeable_program {
     solana_sdk::declare_id!("FbhK8HN9qvNHvJcoFVHAEUCNkagHvu7DTWzdnLuVQ5u4");
 }
 
+pub mod set_upgrade_authority_via_cpi_enabled {
```
jackcmay, line 78: These aren't ordered in any way, it's just been convention to add it to the end

sdk/src/feature_set.rs
```diff
@@ -152,6 +156,7 @@ lazy_static! {
         (rewrite_stake::id(), "rewrite stake"),
         (filter_stake_delegation_accounts::id(), "filter stake_delegation_accounts #14062"),
         (bpf_loader_upgradeable_program::id(), "upgradeable bpf loader"),
+        (set_upgrade_authority_via_cpi_enabled::id(), "set upgrade authority instruction via cpi calls for upgradable programs"),
```
jackcmay, line 159: Same here

jackcmay (dismissed):

sdk/src/feature_set.rs
SebastianBor: Moved to the end

PR #16682, opened 2021-04-20
use cheaper account get

jeffwashington:
#### Problem
We don't need to check roots or ancestors for cleaning stores.

#### Summary of Changes
Just get the slot list from the index.

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -1879,7 +1879,7 @@ impl AccountsDb {
             stored_accounts
                 .iter()
                 .filter(|(pubkey, stored_account)| {
-                    if let Some((locked_entry, _)) = self.accounts_index.get(pubkey, None, None) {
+                    if let Some(locked_entry) = self.accounts_index.get_account_read_entry(pubkey) {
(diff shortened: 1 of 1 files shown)


runtime/src/accounts_db.rs
```diff
@@ -5117,7 +5117,7 @@ impl AccountsDb {
                         offset,
                         ..
                     } = stored_account;
-                    if let Some((locked_entry, _)) = self.accounts_index.get(pubkey, None, None) {
```
jeffwashington, line 5120: @carllin Assuming this is right, should we change v1 or leave it alone?

carllin: yeah, v1 is toggled via a flag, so as long as we're supporting that flag, I think we should port these changes as well.

carllin (dismissed):

runtime/src/accounts_db.rs
carllin: Seems safe!

sakridge (approved):
lgtm

jeffwashington:
This now checks for accounts in stores that aren’t in the index in a case where it would not have before besides also being faster (due to no root/ancestor search) in the common case where pubkeys are in the index.

PR #16688, opened 2021-04-20
some ReadableAccount changes

jeffwashington:
#### Problem
There will likely be new implementations of an account that can implement ReadableAccount without data copies required to create an AccountSharedData.

#### Summary of Changes
A few code paths now support the more general ReadableAccount trait instead of specifically AccountSharedData. This allows other implementations of ReadableAccount to share the code path with no changes.

Fixes #


jeffwashington:
@jackcmay and I discussed

PR #16689, opened 2021-04-20
Reduce account index lookups during clean

jeffwashington:
#### Problem
For some code paths in clean, we end up searching the account index for the same pubkey twice. This is not duplicate pubkeys in the stores - this is due to: if the key isn't doesn't match the fork then we call contains_key to search again.

#### Summary of Changes
Return more information from get. Specifically, whether the pubkey was found in the index at all - even if it wasn't in a rooted or ancestor slot less than the specified max.
Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -1561,7 +1561,7 @@ impl AccountsDb {
                         for pubkey in pubkeys {
                             let remove_from_zero_lamports_list =
                                 match self.accounts_index.get(pubkey, None, max_clean_root) {
-                                    AccountIndexGetResult::Success(locked_entry, index) => {
+                                    AccountIndexGetResult::Found(locked_entry, index) => {
                                         let slot_list = locked_entry.slot_list();
                                         let (slot, account_info) = &slot_list[index];
                                         if account_info.lamports == 0 {
@@ -1594,11 +1594,11 @@ impl AccountsDb {
                                         }
                                         remove_from_zero_lamports_list
                                     }
-                                    AccountIndexGetResult::NotFoundOnFork(_locked_entry) => {
+                                    AccountIndexGetResult::NotFoundOnFork => {
                                         // do nothing - pubkey is in index, but not found in a root slot
                                         false
                                     }
-                                    AccountIndexGetResult::Missing() => {
+                                    AccountIndexGetResult::Missing => {
                                         // pubkey is missing from index, so remove from zero_lamports_list
                                         true
                                     }
@@ -2340,12 +2340,12 @@ impl AccountsDb {
         clone_in_lock: bool,
     ) -> Option<(Slot, AppendVecId, usize, Option<LoadedAccountAccessor<'a>>)> {
         let (lock, index) = match self.accounts_index.get(pubkey, Some(ancestors), max_root) {
-            AccountIndexGetResult::Success(lock, index) => (lock, index),
+            AccountIndexGetResult::Found(lock, index) => (lock, index),
             // we bail out pretty early for missing.
-            AccountIndexGetResult::NotFoundOnFork(_) => {
+            AccountIndexGetResult::NotFoundOnFork => {
                 return None;
             }
-            AccountIndexGetResult::Missing() => {
+            AccountIndexGetResult::Missing => {
                 return None;
             }
         };
@@ -3896,7 +3896,7 @@ impl AccountsDb {
                         let result: Vec<Hash> = pubkeys
                             .iter()
                             .filter_map(|pubkey| {
-                                if let AccountIndexGetResult::Success(lock, index) =
+                                if let AccountIndexGetResult::Found(lock, index) =
                                     self.accounts_index.get(pubkey, Some(ancestors), Some(slot))
                                 {
                                     let (slot, account_info) = &lock.slot_list()[index];
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -1569,8 +1569,7 @@ impl AccountsDb {
                                             self.accounts_index
                                                 .roots_and_ref_count(&locked_entry, max_clean_root),
                                         );
-                                    }
-                                    else {
+                                    } else {
                                         // prune zero_lamport_pubkey set which should contain all 0-lamport
                                         // keys whether rooted or not. A 0-lamport update may become rooted
                                         // in the future.
@@ -1596,7 +1595,6 @@ impl AccountsDb {
                                 }
                                 AccountIndexGetResult::NotFoundOnFork => {
                                     // do nothing - pubkey is in index, but not found in a root slot
-                                    ()
                                 }
                                 AccountIndexGetResult::Missing(lock) => {
                                     // pubkey is missing from index, so remove from zero_lamports_list
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -1559,53 +1559,49 @@ impl AccountsDb {
                         let mut purges_in_root = Vec::new();
                         let mut purges = HashMap::new();
                         for pubkey in pubkeys {
-                            let remove_from_zero_lamports_list =
-                                match self.accounts_index.get(pubkey, None, max_clean_root) {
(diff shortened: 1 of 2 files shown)


jeffwashington:
@sakridge , @carllin , please comment on this idea and implementation. I was unsure whether to add an enum (which I did) or make 2 apis or otherwise duplicate code. This seemed like the clearest longer term solution that is most Rusty. Note that I have a few other very narrow prs related to other parts of this codepath.

runtime/src/accounts_index.rs
```diff
@@ -936,13 +944,18 @@ impl<T: 'static + Clone + IsCached + ZeroLamport> AccountsIndex<T> {
...
         ancestors: Option<&Ancestors>,
         max_root: Option<Slot>,
-    ) -> Option<(ReadAccountMapEntry<T>, usize)> {
-        self.get_account_read_entry(pubkey)
-            .and_then(|locked_entry| {
-                let found_index =
-                    self.latest_slot(ancestors, &locked_entry.slot_list(), max_root)?;
-                Some((locked_entry, found_index))
-            })
+    ) -> AccountIndexGetResult<T> {
```
jeffwashington, line 947: Here's the api change where we return an enum instead of an Option.

jeffwashington: Note that we want 3 states:
1. found in index and matching a root/ancestor (previously was Some(...)
2. found in index, NOT matching a root/ancestor (previously was None)
3. not found in index, (previously was None)

carllin: Yeah an enum for multiple options makes the most sense to me 😃

runtime/src/accounts_db.rs
```diff
@@ -2345,8 +2352,13 @@ impl AccountsDb {
         max_root: Option<Slot>,
         clone_in_lock: bool,
     ) -> Option<(Slot, AppendVecId, usize, Option<LoadedAccountAccessor<'a>>)> {
-        let (lock, index) = self.accounts_index.get(pubkey, Some(ancestors), max_root)?;
-        // Notice the subtle `?` at previous line, we bail out pretty early for missing.
+        let (lock, index) = match self.accounts_index.get(pubkey, Some(ancestors), max_root) {
+            AccountIndexGetResult::Success(lock, index) => (lock, index),
+            // we bail out pretty early for missing.
+            _ => {
```
carllin, line 2358: Might be worth explicitly listing out the options here, in case we add something else in the future

runtime/src/accounts_index.rs
```diff
@@ -85,6 +85,12 @@ impl<T> AccountMapEntryInner<T> {
     }
 }
 
+pub enum AccountIndexGetResult<T: 'static> {
+    Success(ReadAccountMapEntry<T>, usize),
+    NoRootFound(ReadAccountMapEntry<T>),
+    Missing(),
```
carllin, line 91: nit: parens unnecessary?

runtime/src/accounts_index.rs
```diff
@@ -85,6 +85,12 @@ impl<T> AccountMapEntryInner<T> {
     }
 }
 
+pub enum AccountIndexGetResult<T: 'static> {
+    Success(ReadAccountMapEntry<T>, usize),
+    NoRootFound(ReadAccountMapEntry<T>),
```
carllin, line 90: nit: Rename `NoRootFound` to `NotFoundOnFork`, as this also encapsulates not matching non-rooted ancestors

runtime/src/accounts_db.rs
```diff
@@ -1559,47 +1559,52 @@ impl AccountsDb {
...
-                                // Release the lock
-                                let slot = *slot;
-                                drop(locked_entry);
-
-                                if self.accounts_index.is_uncleaned_root(slot) {
-                                    // Assertion enforced by `accounts_index.get()`, the latest slot
-                                    // will not be greater than the given `max_clean_root`
-                                    if let Some(max_clean_root) = max_clean_root {
-                                        assert!(slot <= max_clean_root);
+                            let remove_from_zero_lamports_list =
```
carllin, line 1562: Could we remove the redundant pubkeys instead on construction in `construct_candidate_clean_keys()` by using something like a `DashMap` instead of returning a vector with potential duplicates?

carllin: Oh hehe I see this is solving a different problem. When reading the problem description I thought the duplicate lookup was for duplicate pubkeys, but it's for the extra `contains_key()` call below.

runtime/src/accounts_db.rs
```diff
@@ -1559,47 +1559,52 @@ impl AccountsDb {
...
+                                    AccountIndexGetResult::NoRootFound(_locked_entry) => {
+                                        // do nothing - pubkey is in index, but not found in a root slot
+                                        false
+                                    }
+                                    AccountIndexGetResult::Missing() => {
+                                        // pubkey is missing from index, so remove from zero_lamports_list
+                                        true
+                                    }
+                                };
+                            if remove_from_zero_lamports_list {
```
carllin, line 1606: hmmm I think the issue here is the lock is not held for this pubkey during the remove.  This opens the window for the pubkey to be reinserted by some ongoing slot processing  between when we checked for emptiness, dropped the lock, and then delete here

runtime/src/accounts_db.rs
jeffwashington: I added the other 2 options.

runtime/src/accounts_index.rs
jeffwashington: done

runtime/src/accounts_db.rs
jeffwashington: I updated the description.

runtime/src/accounts_db.rs
carllin: Hmmm in the original code there are two calls to `remove_from_zero_lamports_list`:
1) The first https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_db.rs#L1577-L1583 is safe because the slot list lock for that particular pubkey is being held, so it cannot be modified between the check: https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_db.rs#L1578-L1580 and the subsequent remove.

2) This path here https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_db.rs#L1598-L1602 holds the `account_maps` lock between the check and delete

This new change does not hold the lock through either of these checks anymore, i.e., 

1) the slot list lock is dropped here: https://github.com/solana-labs/solana/pull/16689/files#diff-1090394420d51617f3233275c2b65ed706b35b53b115fe65f82c682af8134a6fR1585 after checking for any zero lamport accounts, so that check can be outdated by the time the delete is called
2) If the `NotFoundOnFork` is triggered, we don't grab the lock again as described in 2) above in the original code to check for any changes, so this check can be outdated by the time the delete is called

jeffwashington: yes, I agree with your conclusions and arrived at them as well. I should have taken the time to be more precise and complete in my answer earlier. I was demonstrating that there are 2 different locks that are being held for the different code paths. I haven't done the work yet to figure out the other writers of this list and under what conditions they would be allowed to write to it with no synchronization enforced on it. I meant to do that work before you had read my earlier comment, but I was distracted all day so far. I was not suggesting that the code as it is in this pr is thus correct - merely trying to understand and make sure that the old and new code path are well understood and correct. You raised a very good point initially that there could be synchronization problems on changes to the zero key list code. It is obvious there are subtle synchronization issues here that I don't understand yet. I can fix this by doing the list removes within the lock. I will also figure out how synchronization and access to this list actually works.

jeffwashington: Ah. zero_lamport_pubkeys is a dashset. And modifications to it are designed to be governed indirectly by one of (at least) 2 different external (external to the dashset) locks.

carllin: 🦶 💣

sakridge (approved):
Nice, this should help reduce the accounts_map lock contention from clean. Sorry for the tricky locking. Good job @carllin and @jeffwashington for handling it.

PR #16691, opened 2021-04-20
simplify do_shrink_slot_stores, delay/reduce account clone

jeffwashington:
#### Problem
during do_shrink_slot_stores, we make unnecessary copies of many fields including an account and its data. This work is a precursor to creating a ReadableAccount trait impl for StoredAccountMeta such that we never have to allocate a new account structure and a new data: Arc<Vec<u8>> when we have a StoredAccountMeta instance.
#### Summary of Changes
eliminate completely unnecessary copies of accounts and data. Delay account creation during slot shrink operation.
Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -1840,13 +1840,10 @@ impl AccountsDb {
     where
         I: Iterator<Item = &'a Arc<AccountStorageEntry>>,
     {
-        struct FoundStoredAccount {
-            account: AccountSharedData,
-            account_hash: Hash,
-            account_size: usize,
-            store_id: AppendVecId,
-            offset: usize,
-            write_version: u64,
+        struct FoundStoredAccount<'a> {
+            pub account: StoredAccountMeta<'a>,
+            pub store_id: AppendVecId,
+            pub account_size: usize,
         }
         debug!("do_shrink_slot_stores: slot: {}", slot);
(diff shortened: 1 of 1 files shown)


runtime/src/accounts_db.rs
```diff
@@ -1938,15 +1922,20 @@ impl AccountsDb {
...
+            let mut stored_accounts = Vec::with_capacity(alive_accounts.len());
             let mut hashes = Vec::with_capacity(alive_accounts.len());
             let mut write_versions = Vec::with_capacity(alive_accounts.len());
 
-            for (pubkey, alive_account) in alive_accounts {
-                accounts.push((pubkey, &alive_account.account));
-                hashes.push(alive_account.account_hash);
-                write_versions.push(alive_account.write_version);
+            for (_pubkey, alive_account) in alive_accounts.iter() {
+                stored_accounts.push(alive_account.account.clone_account());
```
carllin, line 1930: Cool, out of curiosity, what's the benefit of delaying the clone until here? Maybe some future change where we'll need a `StoredAccountMeta` early on instead of an `AccountSharedData`?

carllin (approved):

runtime/src/accounts_db.rs
jeffwashington: If we find a duplicate of an account, then we only clone the last one instead of all of them. Shortly, we will not clone the account at all - we will implement ReadableAccount on StoredAccountMeta. This prevents us from having to make copies of anything, including the data.

Issue #5395, opened 2019-08-01
Experiment with skipping bank processing on the leader


sakridge:
#### Problem

Banking stage takes a while.

#### Proposed Solution

The leader could do a reduced check of just checking that the payer for the transaction can pay for the fee, then including the transaction in the blob/broadcast set. Validators would then execute the entire transaction and check the error status/etc. The replay stage of the leader would have to be setup to not skip the block.

garious:
@sakridge, any updates here? A lot has changed since you created this issue.

sakridge:
@garious no, I think it's still a pretty good avenue of optimization.

jeffwashington:
@taozhu-chicago

taozhu-chicago:
created #16641 for `cost model` feature

Issue #16641, opened 2021-04-19
Transaction cost model for bankless leader


taozhu-chicago:
#### Problem

Needed by #5395 to limit number of transaction in a block.

#### Proposed Solution
1. new `cost model` 
    * is configured with two limits: max block cost, and max same account transaction cost.
    * integrated into `banking_stage`, shared amount threads. 
    * is reset when new bank is introduced.
    * Keeps tracking current bank's costs, when a transaction would breach either block or account cost limit, it will be placed into leader's retry queue.

2. the cost limits used to configure `cost model` (described above) are collected from mainnet, they are to encourage parallelizeble transactions as well as to limit block size;

3. determine cost of an transaction:
    * transaction cost = account access cost + instruction execution cost, where
    * account access cost is the sum of accounts instruction accesses, writable is more expansive than read
    * instruction execution cost is maintained in a table, known instrument execution time are collected from cluster; whereas new on-chain program is given a high default value when it is added to the table; `replay stage` will feed the actual execution cost back to `cost model`.

taozhu-chicago:
visit the approach with @sakridge , updated description regarding account access cost and instruction cost

PR #16694, opened 2021-04-20
Cost Model to limit transactions which are not parallelizeable

taozhu-chicago:
#### Problem
Needed by #5395 to limit number of transaction in a block.

#### Summary of Changes
1. added `cost model` that calculates each transaction's "executing cost"; if a system_transfer instruction is cost of '1', a BPF program maybe configured to have cost of 1_000.
2. pack transactions to comply with two limits: 
    2.1 the total cost of un-parallelizeable transactions (eg transactions need to lock and write to same account ) must be less than configured `chain_max_cost`;
    2.2 total cost of a blob of transaction should be less than configured `block_max_cost`.

Fixes #

### DIFF
--- a/core/src/banking_stage.rs
+++ b/core/src/banking_stage.rs
@@ -1056,9 +1056,7 @@ impl BankingStage {
         // Making a snapshot of shared cost_tracker by clone(), drop lock immediately.
         // Local copy `cost_tracker` is used to filter transactions by cost.
         // Shared cost_tracker is updated later by processed transactions confirmed by bank.
-        let guard = cost_tracker.lock().unwrap();
-        let mut cost_tracker = guard.clone();
-        drop(guard);
+        let mut cost_tracker = cost_tracker.lock().unwrap().clone();
 
         let mut retryable_transaction_packet_indexes: Vec<usize> = vec![];
         let (filtered_transactions, filter_transaction_packet_indexes) = transaction_indexes
--- a/core/src/banking_stage.rs
+++ b/core/src/banking_stage.rs
@@ -2,6 +2,8 @@
 //! to contruct a software pipeline. The stage uses all available CPU cores and
 //! can do its processing in parallel with signature verification on the GPU.
 use crate::{
+    cost_model::{CostModel, ACCOUNT_MAX_COST, BLOCK_MAX_COST},
+    cost_tracker::CostTracker,
     packet_hasher::PacketHasher,
     poh_recorder::{PohRecorder, PohRecorderError, TransactionRecorder, WorkingBankEntry},
     poh_service::{self, PohService},
@@ -231,6 +233,35 @@ impl BankingStage {
         transaction_status_sender: Option<TransactionStatusSender>,
         gossip_vote_sender: ReplayVoteSender,
     ) -> Self {
+        Self::new_with_cost_limit(
+            cluster_info,
+            poh_recorder,
+            verified_receiver,
+            verified_vote_receiver,
+            transaction_status_sender,
+            gossip_vote_sender,
+            ACCOUNT_MAX_COST,
+            BLOCK_MAX_COST,
+        )
+    }
+
+    pub fn new_with_cost_limit(
+        cluster_info: &Arc<ClusterInfo>,
+        poh_recorder: &Arc<Mutex<PohRecorder>>,
+        verified_receiver: CrossbeamReceiver<Vec<Packets>>,
+        verified_vote_receiver: CrossbeamReceiver<Vec<Packets>>,
+        transaction_status_sender: Option<TransactionStatusSender>,
+        gossip_vote_sender: ReplayVoteSender,
+        account_cost_limit: u32,
+        block_cost_limit: u32,
+    ) -> Self {
+        // shared immutable 'cost_model' that calcuates transaction costs
+        // shared mutex guarded 'cost_tracker' tracks bank's cost against configured limits.
+        let cost_model = Arc::new(CostModel::new(account_cost_limit, block_cost_limit));
+        let cost_tracker = Arc::new(Mutex::new(CostTracker::new(
+            cost_model.get_account_cost_limit(),
+            cost_model.get_block_cost_limit(),
+        )));
         Self::new_num_threads(
             cluster_info,
             poh_recorder,
@@ -239,6 +270,8 @@ impl BankingStage {
             Self::num_threads(),
             transaction_status_sender,
             gossip_vote_sender,
+            &cost_model,
+            &cost_tracker,
         )
     }
 
@@ -250,6 +283,8 @@ impl BankingStage {
         num_threads: u32,
         transaction_status_sender: Option<TransactionStatusSender>,
         gossip_vote_sender: ReplayVoteSender,
+        cost_model: &Arc<CostModel>,
+        cost_tracker: &Arc<Mutex<CostTracker>>,
     ) -> Self {
         let batch_limit = TOTAL_BUFFERED_PACKETS / ((num_threads - 1) as usize * PACKETS_PER_BATCH);
         // Single thread to generate entries from many banks.
@@ -276,6 +311,8 @@ impl BankingStage {
                 let transaction_status_sender = transaction_status_sender.clone();
                 let gossip_vote_sender = gossip_vote_sender.clone();
                 let duplicates = duplicates.clone();
+                let cost_model = cost_model.clone();
+                let cost_tracker = cost_tracker.clone();
                 Builder::new()
                     .name("solana-banking-stage-tx".to_string())
                     .spawn(move || {
@@ -291,6 +328,8 @@ impl BankingStage {
                             transaction_status_sender,
                             gossip_vote_sender,
                             &duplicates,
+                            &cost_model,
+                            &cost_tracker,
                         );
                     })
                     .unwrap()
@@ -338,6 +377,11 @@ impl BankingStage {
         has_more_unprocessed_transactions
     }
 
+    fn reset_cost_tracker_if_new_bank(cost_tracker: &Arc<Mutex<CostTracker>>, bank_slot: Slot) {
+        cost_tracker.lock().unwrap().reset_if_new_bank(bank_slot);
+    }
+
+    #[allow(clippy::too_many_arguments)]
     pub fn consume_buffered_packets(
         my_pubkey: &Pubkey,
         max_tx_ingestion_ns: u128,
@@ -348,6 +392,8 @@ impl BankingStage {
         test_fn: Option<impl Fn()>,
         banking_stage_stats: &BankingStageStats,
         recorder: &TransactionRecorder,
+        cost_model: &Arc<CostModel>,
+        cost_tracker: &Arc<Mutex<CostTracker>>,
     ) {
         let mut rebuffered_packets_len = 0;
         let mut new_tx_count = 0;
@@ -365,6 +411,8 @@ impl BankingStage {
                     &original_unprocessed_indexes,
                     my_pubkey,
                     *next_leader,
+                    cost_model,
+                    cost_tracker,
                 );
                 Self::update_buffered_packets_with_new_unprocessed(
                     original_unprocessed_indexes,
@@ -373,6 +421,7 @@ impl BankingStage {
             } else {
                 let bank_start = poh_recorder.lock().unwrap().bank_start();
                 if let Some((bank, bank_creation_time)) = bank_start {
+                    Self::reset_cost_tracker_if_new_bank(cost_tracker, bank.slot());
                     let (processed, verified_txs_len, new_unprocessed_indexes) =
                         Self::process_packets_transactions(
                             &bank,
@@ -383,6 +432,8 @@ impl BankingStage {
                             transaction_status_sender.clone(),
                             gossip_vote_sender,
                             banking_stage_stats,
+                            cost_model,
+                            cost_tracker,
                         );
                     if processed < verified_txs_len
                         || !Bank::should_bank_still_be_processing_txs(
@@ -485,6 +536,8 @@ impl BankingStage {
         gossip_vote_sender: &ReplayVoteSender,
         banking_stage_stats: &BankingStageStats,
         recorder: &TransactionRecorder,
+        cost_model: &Arc<CostModel>,
+        cost_tracker: &Arc<Mutex<CostTracker>>,
     ) -> BufferedPacketsDecision {
         let bank_start;
         let (
@@ -495,6 +548,9 @@ impl BankingStage {
         ) = {
             let poh = poh_recorder.lock().unwrap();
             bank_start = poh.bank_start();
+            if let Some((ref bank, _)) = bank_start {
+                Self::reset_cost_tracker_if_new_bank(cost_tracker, bank.slot());
+            };
             (
                 poh.leader_after_n_slots(FORWARD_TRANSACTIONS_TO_LEADER_AT_SLOT_OFFSET),
                 PohRecorder::get_bank_still_processing_txs(&bank_start),
@@ -525,6 +581,8 @@ impl BankingStage {
                     None::<Box<dyn Fn()>>,
                     banking_stage_stats,
                     recorder,
+                    cost_model,
+                    cost_tracker,
                 );
             }
             BufferedPacketsDecision::Forward => {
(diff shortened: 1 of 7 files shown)


core/src/banking_stage.rs
```diff
@@ -899,12 +900,23 @@ impl BankingStage {
...
+        /* TODO - 
+        // instead of using `MAX_NUM_TRANSACTIONS_PER_BATCH` to get chunks, invoking cost_model here to
+        // breaks `transactions` into chunks of re-organized transaction, each complies with cost
+        // model
+        let cost_model = CostModel::new();
+        let chunks = cost_model.pack_transactions_by_cost( transactions );
+        for &chunk in chunks  {
+            reuse below code
+        }
+        // */
```
taozhu-chicago, line 914: Is this right spot to insert new 'cost model' logic - to replace the static chunking with cost_model.

core/src/cost_model.rs
```diff
@@ -0,0 +1,136 @@
...
+
+use solana_sdk::{
+    bpf_loader, bpf_loader_deprecated, bpf_loader_upgradeable, pubkey::Pubkey, system_program,
+};
+use solana_runtime::{
+    hashed_transaction::HashedTransaction,
+};
+use std::{collections::HashMap, str::FromStr};
+
+// TODO  revisit these hardcoded numbers, better get from mainnet log
```
taozhu-chicago, line 13: below are just some made-up numbers, will be replace with stats collected from mainnet later

core/src/cost_model.rs
```diff
@@ -0,0 +1,136 @@
...
+const COST_UNIT: u32 = 1;
+const DEFAULT_PROGRAM_COST: u32 = COST_UNIT * 500;
+const CHAIN_MAX_COST: u32 = COST_UNIT * 100_000;
+const BLOCK_MAX_COST: u32 = COST_UNIT * 100_000_000;
+
+use Pubkey as ProgramKey;
+use u32 as Cost;
+
+#[derive(Debug)]
```
taozhu-chicago, line 22: The code model assigns `cost` to instructions, which is identified by `program id`. I am not sure if this is correct, i am assuming since a transaction can have multiple instructions, the cost_of_transaction = sum( cost_of_instructions ); in order to figure out the cost of transaction, need to iterate through all instructions in it. 

`chain_max_cost` is the configured number of transactions has same `writable accounts`. chained transactions are not parallelizeable. I have question how to define `chained transaction`. I understand transaction have multiple accounts Read-write accounts. Which read-write account should be used to group transactions together as 'chained'?  

`block_max_cost` is the configured total `cost` a block should have. (does using word `block` here cause confusion?)

core/src/cost_model.rs
```diff
@@ -0,0 +1,136 @@
...
+    }}
+}
+
+impl CostModel {
+    pub fn new() -> Self {
+        // NOTE: message.rs has following lazy_static program ids. Can probably use them to define
+        // `cost` for each type.
+        // NOTE: since each instruction has `compute budget`, possible to derive `cost` from `budget`?
+        let parse = |s| Pubkey::from_str(s).unwrap();
+        Self {
```
taozhu-chicago, line 43: stole the list of program IDs from `message.rs`. Is this a comprehensive list for all program IDs? for BPF programs, can they be identified (and cost assigned) by `bpf_loader` IDs? 

In below `find_instruction_cost` function, I added logic that is a instrument's program ID is not found in this metrics, it'd be assigned to default cost. Does this make sense?

core/src/cost_model.rs
```diff
@@ -0,0 +1,136 @@
...
+        transaction: &HashedTransaction
+    ) -> &Cost {
+        // TODO - iter through instructions in `transaction`, return sum( cost(instruction) )
+        //self.find_instruction_cost( &system_program::id() )
+        &DEFAULT_PROGRAM_COST
+    }
+
+    // NOTE - main function that breaks a collection of transactions into several smaller
+    //        collection of transaction, each complies the cost limits (on both Chain_cost and
+    //        block_cost).
```
taozhu-chicago, line 85: this is the function `banking_stage` calls, does the signature make sense?

mvines:
> "[WIP] initial review" 

Please give your PRs a helpful title for easier scanning 🙏🏼

taozhu-chicago:
> > "[WIP] initial review"
> 
> Please give your PRs a helpful title for easier scanning 🙏🏼

Oops, meant for a private review. In any case, always a good practice to supply a meaningful title and description. Thanks for pointing out

core/src/banking_stage.rs
sakridge: This could be a good spot, then any transactions that don't meet the model are pushed into unprocessed_txs to retry in the next slot. I don't think it really needs to mess with the batching though. It can pass however many transactions meet the model and then the MAX_NUM_TRANSACTIONS_PER_BATCH can split them into multiple batches if necessary.

core/src/cost_model.rs
sakridge: Yes, that's correct about iterating all instructions in the transaction.

All read-write accounts in the transaction should be used to group, since all of those will prevent parallel execution.

block seems reasonable to me.

core/src/banking_stage.rs
sakridge: Actually I think in `transactions_from_packets` itself might be a good place. In there we are constructing the `Transaction` `Vec` so it's easier to just skip adding a Transaction to the list. In here we only have the reference to the Transaction list and we cannot change it, just copy it or use a partial view of it.

taozhu-chicago: Brilliant! I do feel it is bit too tight in regarding we can do with the transaction list here. Will look into move this to `transactions_to_packets`. Thanks!

core/src/cost_model.rs
taozhu-chicago: In current version, I 'chain' transactions by signer's account - `&tx.message().account_keys[0..tx.message().header.num_required_signatures as usize];`
What is a right to find out all writable accounts?

taozhu-chicago:
@sakridge Implemented your suggestion - making cost_model a filter when banking_stage gets transactions from packages. It dramatically simplified the code since `cost model` no longer needs to catch and organize transactions, instead it only needs to keep tracking cost for one block. 
What is in right now is the final version of draft :D Can you review when you have chance? 
I plan to run bench_tps as next step.

core/src/cost_model.rs
sakridge: There is a helper function on Message `is_writable`

core/src/cost_model.rs
```diff
@@ -0,0 +1,427 @@
...
+use std::{collections::HashMap, str::FromStr};
+
+// TODO  revisit these hardcoded numbers, better get from mainnet log
+const COST_UNIT: u32 = 1;
+const DEFAULT_PROGRAM_COST: u32 = COST_UNIT * 500;
+const CHAIN_MAX_COST: u32 = COST_UNIT * 100_000;
+const BLOCK_MAX_COST: u32 = COST_UNIT * 100_000_000;
+
+use u32 as Cost;
+use Pubkey as ProgramKey;
```
sakridge, line 19: It looks a bit weird as `use` I think we would generally use `type Cost = u32;` for a type alias.

taozhu-chicago:
@sakridge, this PR is force pushed again :D This time to include changes:

taozhu-chicago:
Hardcoded cost model config based on data collected from mainnet ledger using ledger-tool. The idea is to set block cost limit much higher than what's in current ledger to allow rapid growth, while setting chain cost limit (eg cost per writable account) to close to current average to encourage adding parallelizable transactions in to bank.

sakridge:
> Hardcoded cost model config based on data collected from mainnet ledger using ledger-tool. The idea is to set block cost limit much higher than what's in current ledger to allow rapid growth, while setting chain cost limit (eg cost per writable account) to close to current average to encourage adding parallelizable transactions in to bank.

Hmm. So if a new program is uploaded to the chain, how will the cost model treat it? I think we need to handle this before we can incorporate this change.

taozhu-chicago (commented):
> > Hardcoded cost model config based on data collected from mainnet ledger using ledger-tool. The idea is to set block cost limit much higher than what's in current ledger to allow rapid growth, while setting chain cost limit (eg cost per writable account) to close to current average to encourage adding parallelizable transactions in to bank.
> 
> Hmm. So if a new program is uploaded to the chain, how will the cost model treat it? I think we need to handle this before we can incorporate this change.

Fair point. The hardcoded instruction to cost matrix is not future approve. How about we estimate cost of instrument by accounts it is going to read and write? Something like to read an account cost 1 unit, and write to an account cost 10? Are similar methods out there to gauge instruments' cost I can borrow here?

taozhu-chicago:
update `banking_stage` to reduce lock scope, the bench comparisons are:
### banking-bench

stats | master | branch 
--- | --- | ---
banking_bench_total | 10,691.61 | 10,740.34 
banking_bench_tx_total | 10,977.33 | 11,027.94 
banking_bench_success_tx_total | 10,365.51 | 10,349.21 

### bench-tps

stats | master | branch 
--- | --- | ---
max tps | 58,521.46 |  47,856.24
average tps | 30,495.34 | 28,866.98
total transactions | 2,779,373.00 |  2,629,776.00
drop rate | 0.77 | 0.78

### system-performance-test
results in slack channel.

If there are considerable regression in these stats, one area we can reduce impact by adding cost_model logic is at https://github.com/solana-labs/solana/pull/16694/files#diff-bb8464c2f02c8863fb60e6e29f023f3e6da93a93fbb9199a70a44c9c1b41c7f3R1010 to instead of acquiring lock for each transaction, it can take snapshot of cost_model data first, then iterate through transaction while determine if a transaction can be added to bank based on snapshot of cost model, submit changes to cost model after finish iteration. This should significantly reduce locking, and get us back to close to original performance; However, the drawback is snapshot does not reflecting what other threads were doing, therefore the cost application will not be as precise, but should be acceptable. My main concern is if the extra logic and complicity worth the gain.

core/src/cost_model.rs
```diff
@@ -0,0 +1,466 @@
...
+            chain_max, block_max
+        );
+
+        // NOTE: message.rs has following lazy_static program ids.
+        let parse = |s| Pubkey::from_str(s).unwrap();
+        Self {
+            cost_metrics: costmetrics![
+                parse("Config1111111111111111111111111111111111111") => COST_UNIT,
+                parse("Feature111111111111111111111111111111111111") => COST_UNIT * 2,
+                parse("NativeLoader1111111111111111111111111111111") => COST_UNIT * 2,
```
sakridge, line 123: Can we not use the hardcoded ones?

There should be an `id(()` for each of these.

core/src/cost_model.rs
```diff
@@ -0,0 +1,466 @@
...
+use crate::cost_tracker::CostTracker;
+use log::*;
+use solana_sdk::{
+    bpf_loader, bpf_loader_deprecated, bpf_loader_upgradeable, clock::Slot, message::Message,
+    pubkey::Pubkey, system_program, transaction::Transaction,
+};
+use std::{collections::HashMap, str::FromStr};
+
+// from mainnet-beta data, taking `vote program` as 1 COST_UNIT to load and execute
+// amount all type programs, the costs are:
```
sakridge, line 14: it would be useful to have the hardware that produced this cost.

taozhu-chicago: That'd be a solid approach. updated issue page (https://github.com/solana-labs/solana/issues/16641) regarding instruction cost table, will get pipe work done in this PR, then create a PR to implement have `replay stage` to feedback realtime value.

core/src/cost_model.rs
taozhu-chicago: pushed the change ... thanks for pointing it out

taozhu-chicago:
force pushed some more refactoring to PR, includs:
1. added Account Access Cost to be part of Transaction Cost; 
2. instruction_execution_cost_table for execution cost, it is read-only for now; in next PR, Replay will write realtime data into it.
3. Separated cost_model and cost_tracker in banking_stage, former is outside of any lock to only estimate transaction cost; later is Mutex guarded.
4. replace hardcoded program ID with `id()`
5. Some more unit tests for new functions.
@sakridge can you review when you have chance? I'll run Banking-bench, bench-tps and system-test on it to compare with master.

core/src/banking_stage.rs
```diff
@@ -985,23 +1049,48 @@ impl BankingStage {
...
         secp256k1_program_enabled: bool,
-    ) -> (Vec<HashedTransaction<'static>>, Vec<usize>) {
-        transaction_indexes
+        cost_model: &Arc<CostModel>,
+        cost_tracker: &Arc<Mutex<CostTracker>>,
+    ) -> (Vec<HashedTransaction<'static>>, Vec<usize>, Vec<usize>) {
+        // make a copy of cost_tracker to statically check limit for this batch of packets
+        let guard = cost_tracker.lock().unwrap();
+        let mut cost_tracker = guard.clone();
+        drop(guard);
```
sakridge, line 1058: Is there any difference between doing this and just doing:
`let mut cost_tracker = cost_tracker.lock().unwrap();`?

core/src/banking_stage.rs
```diff
@@ -985,23 +1049,48 @@ impl BankingStage {
...
                 let message_bytes = Self::packet_message(p)?;
                 let message_hash = Message::hash_raw_message(message_bytes);
                 Some((
                     HashedTransaction::new(Cow::Owned(tx), message_hash),
                     tx_index,
                 ))
             })
-            .unzip()
+            .unzip();
+
```
sakridge, line 1088: I think we might want to just do another loop after doing the hash and secp256k1, so we do those outside the Mutex lock.

core/src/cost_model.rs
```diff
@@ -0,0 +1,554 @@
...
+// Median: 12 COST_UNIT
+// Average: 19 COST_UNIT
+const COST_UNIT: u32 = 1;
+const DEFAULT_PROGRAM_COST: u32 = COST_UNIT * 100;
+// re-adjust these numbers if needed
+const SIGNED_WRITABLE_ACCOUNT_ACCESS_COST: u32 = COST_UNIT * 10;
+const SIGNED_READONLY_ACCOUNT_ACCESS_COST: u32 = COST_UNIT * 2;
+const NON_SIGNED_WRITABLE_ACCOUNT_ACCESS_COST: u32 = COST_UNIT * 5;
+const NON_SIGNED_READONLY_ACCOUNT_ACCESS_COST: u32 = COST_UNIT;
+// running 'ledger-tool compute-cost' over mainnet ledger, the alrgest block cost
```
sakridge, line 31: "largest block cost"

core/src/cost_tracker.rs
```diff
@@ -0,0 +1,356 @@
...
+        Ok(self.block_cost)
+    }
+
+    fn would_fit(&self, keys: &[Pubkey], cost: &u32) -> Result<bool, &'static str> {
+        // check against the total package cost
+        if self.block_cost + cost > self.block_cost_limit {
+            return Err("would exceed block cost limit");
+        }
+
+        // chech if the transaction itself is more costly than the account_cost_limit
```
sakridge, line 51: "check if the transaction itself"

core/src/cost_tracker.rs
```diff
@@ -0,0 +1,356 @@
...
+                        return Err("would exceed account cost limit");
+                    } else {
+                        continue;
+                    }
+                }
+                None => continue,
+            }
+        }
+
+        Ok(true)
```
sakridge, line 70: Does anything use this value? If not, just use the empty return `()`

core/src/banking_stage.rs
sakridge: it could be `let mut cost_tracker = cost_tracker.lock().unwrap().clone();`

taozhu-chicago: add comment to document the intent

core/src/banking_stage.rs
taozhu-chicago: added comment to document the use of local copy of tracker, no lock here

core/src/cost_model.rs
taozhu-chicago: fixed

core/src/cost_tracker.rs
taozhu-chicago: fixed

core/src/cost_tracker.rs
taozhu-chicago: good call! changed return type to `Result<(), str>` and returning `Ok(())`

sakridge (approved):
lgtm

PR #16696, opened 2021-04-21
Add port and gossip options to solana-test-validator

CriesofCarrots:
#### Problem
`solana-test-validator` is a convenient way to spin up a node, but can only realistically be used locally. Users testing more complete workflows need to be able to do things like deploy programs from remote nodes.

#### Summary of Changes
Add `--gossip-port`, `--gossip-host`, `--dynamic-port-range`, and `--bind-address` options to `solana-test-validator`

### DIFF
--- a/core/src/cluster_info.rs
+++ b/core/src/cluster_info.rs
@@ -3358,6 +3358,7 @@ impl Node {
             },
         }
     }
+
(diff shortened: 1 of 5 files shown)


mvines (dismissed):
lgtm overall

PR #16708, opened 2021-04-21
limits to data_header.size when combining shreds' payloads

behzadnouri:
#### Problem
deshred is ignoring `data_header.size` when combining shreds' payloads:
https://github.com/solana-labs/solana/blob/37b8587d4/ledger/src/shred.rs#L940-L961

#### Summary of Changes
* limit the payload to `data_header.size`.
* also more sanity checks on the given shreds indices.

### DIFF
--- a/ledger/src/shred.rs
+++ b/ledger/src/shred.rs
@@ -786,7 +786,7 @@ impl Shredder {
             .iter()
             .map(|shred| &shred.payload[..PAYLOAD_ENCODE_SIZE])
             .collect();
-        let mut parity = vec![vec![0; PAYLOAD_ENCODE_SIZE]; num_coding];
+        let mut parity = vec![vec![0u8; PAYLOAD_ENCODE_SIZE]; num_coding];
         Session::new(num_data, num_coding)
             .unwrap()
             .encode(&data, &mut parity[..])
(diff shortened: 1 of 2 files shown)


ledger/src/shred.rs
```diff
@@ -938,37 +938,36 @@ impl Shredder {
...
-                0
-            };
-
-            if num_data.saturating_add(first_index) != last_index.saturating_add(1) {
-                return Err(reed_solomon_erasure::Error::TooFewDataShards);
-            }
-
-            shreds.iter().map(|shred| &shred.payload).collect()
+        let index = shreds.first().ok_or(TooFewDataShards)?.index();
+        let aligned = shreds.iter().zip(index..).all(|(s, i)| s.index() == i);
```
carllin, line 945: If the indexes don't align, that would imply something is disastrously wrong with the blockstore insertion code/corruption might have occurred. In these cases, can we hard panic?

carllin (approved):

ledger/src/shred.rs
```diff
@@ -938,37 +938,36 @@ impl Shredder {
...
             })
-            .cloned()
-            .collect()
+            .copied()
+            .collect();
+        if data.is_empty() {
+            // For backward compatibility. This is needed when the data shred
+            // payload is None, so that deserializing to Vec<Entry> results in
+            // an empty vector.
+            Ok(vec![0u8; SIZE_OF_DATA_SHRED_PAYLOAD])
```
carllin, line 967: So we know the input `shreds` isn't empty, otherwise `verify_consistent_shred_payload_sizes()` above would've errored out with `TooFewShardsPresent`. 

>This is needed when the data shred payload is None
Does this mean when all the `shred.payload[offset..size].iter()` above return empty, and if so can we add a test case for this?

ledger/src/shred.rs
```diff
@@ -938,37 +938,36 @@ impl Shredder {
...
+        let data: Vec<_> = shreds
             .iter()
-            .flat_map(|data| {
-                let offset = SIZE_OF_COMMON_SHRED_HEADER + SIZE_OF_DATA_SHRED_HEADER;
-                data[offset..valid_data_len].iter()
+            .flat_map(|shred| {
+                let size = shred.data_header.size as usize;
+                let size = shred.payload.len().min(size);
+                let offset = SHRED_DATA_OFFSET.min(size);
+                shred.payload[offset..size].iter()
```
carllin, line 959: When we reconstruct the data shred header here: https://github.com/solana-labs/solana/blob/master/ledger/src/shred.rs#L340-L341, are we guaranteed that the `shred.payload.len()` is greater than `offset`? (i.e. can a malicious validator plumb in a bad header size?)

PR #16710, opened 2021-04-21
Make metrics tests independent of RUST_LOG environment variable

ruuda:
#### Problem

Running the metrics counter tests with `RUST_LOG=none` would fail, because the env logger would set its filter level to reject all log messages, and incrementing a counter only happens if the global logger has at least the specified log level. Having the tests behave differently when `RUST_LOG` is set is surprising, they should be self-contained. Nix' `buildRustPackage` sets `RUST_LOG=""` by default (I suppose to make the build logs less verbose). This makes the tests fail, and that is a small hurdle in packaging Solana for Nix. 

#### Summary of Changes

Instead of setting up the `EnvLogger` with `from_env`, which reads `RUST_LOG`, configure the logger builder manually to ensure the result does not depend on any environment variables. Extract it into a function to avoid some duplication.

### DIFF
--- a/metrics/src/counter.rs
+++ b/metrics/src/counter.rs
@@ -226,12 +226,26 @@ mod tests {
         }
     }
 
+    /// Try to initialize the logger with a filter level of INFO.
+    ///
+    /// Incrementing a counter only happens if the logger is configured for the
+    /// given log level, so the tests need an INFO logger to pass.
+    fn try_init_logger_at_level_info() -> Result<(), log::SetLoggerError> {
+        // Use ::new() to configure the logger manually, instead of using the
+        // default of reading the RUST_LOG environment variable. Set is_test to
+        // print to stdout captured by the test runner, instead of polluting the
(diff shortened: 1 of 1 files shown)


ruuda:
The build logs say:

```
failures:
 
---- accounts_db::tests::test_load_account_and_purge_race_without_retry stdout ----
note: test did not panic as expected
---- accounts_db::tests::test_load_account_and_purge_race_with_retry stdout ----
note: test did not panic as expected
```

However I did not modify those, and locally these tests pass on this branch. They are in a different package too, so I don’t expect the global logger to matter here. How can I debug this?

CriesofCarrots:
> However I did not modify those, and locally these tests pass on this branch. They are in a different package too, so I don’t expect the global logger to matter here. How can I debug this?

Sadly, I think those tests are just flaky. I retried the job, and they both passed.

CriesofCarrots (approved):
Thank you! Much better.

PR #16711, opened 2021-04-21
Revert "read only accounts cache uses accurate size representation (#…

sakridge:
#### Problem

Performance regression from accounts cache change

#### Summary of Changes

This reverts commit 36e11998c743590035d1860479c12a3a6259698b.

Fixes #


sakridge:
Fixed another way.

PR #16722, opened 2021-04-21
clean read_only_cache in the bg

jeffwashington:
#### Problem
Several problems with read only cache lru cleaning.

#### Summary of Changes
Move lru cleaning to a background thread.

Fixes #

### DIFF
--- a/runtime/src/read_only_accounts_cache.rs
+++ b/runtime/src/read_only_accounts_cache.rs
@@ -211,6 +211,7 @@ impl ReadOnlyAccountsCache {
 
             // we didn't get enough, so calculate a new list and keep purging
             let (new_lru, new_size) = self.calculate_lru_list();
+            lru_index = 0;
             lru = new_lru;
             self.data_size.store(new_size, Ordering::Relaxed);
             lru.sort();
--- a/runtime/src/read_only_accounts_cache.rs
+++ b/runtime/src/read_only_accounts_cache.rs
@@ -11,7 +11,7 @@ use std::thread::sleep;
 use std::time::Duration;
 use std::{
     sync::{
-        atomic::{AtomicU64, AtomicUsize, Ordering},
+        atomic::{AtomicBool, AtomicU64, AtomicUsize, Ordering},
         Arc, RwLock,
     },
     time::Instant,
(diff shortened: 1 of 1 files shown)


runtime/src/read_only_accounts_cache.rs
```diff
@@ -27,25 +28,54 @@ pub struct ReadOnlyAccountCacheEntry {
...
     hits: AtomicU64,
     misses: AtomicU64,
-    lru: LruList,
     per_account_size: usize,
+    stop: Arc<AtomicBool>,
+}
+
+impl Drop for ReadOnlyAccountsCache {
+    fn drop(&mut self) {
+        self.stop.store(true, Ordering::Relaxed);
```
sakridge, line 42: generally good practice to join the thread as well

runtime/src/read_only_accounts_cache.rs
```diff
@@ -92,86 +140,86 @@ impl ReadOnlyAccountsCache {
...
-            list.sort();
-            new_size = self.purge_lru_list(&mut list, false, new_size);
-            // the list is stored in self so we use it to purge next time
+            // purge from the lru list we last made
+            if self.purge_lru_list(&lru, &mut lru_index) {
+                continue;
+            }
+
+            // we didn't get enough, so calculate a new list and keep purging
+            let (new_lru, new_size) = self.calculate_lru_list();
```
sakridge, line 208: What about `lru.clear()` here and then just recreate into it again?

runtime/src/read_only_accounts_cache.rs
jeffwashington: done

runtime/src/read_only_accounts_cache.rs
jeffwashington: done. bench tps shows a range of 1.3-1.8M entries in this list. With this new strategy, this lru list will keep growing and then remain at its largest size forever. Not a huge deal. I was imagining there were advantages to it using exactly what it needed each time the lru list is recalculated, which isn't too often. I can also add code to reallocate it smaller if we shrink size enough. I was trying to keep the code simple with less things to go wrong.

sakridge (approved):
lgtm

jeffwashington:
perf tests:
mean_tps: 21412
max_tps: 62900
mean_confirmation_ms: 2511
max_confirmation_ms: 22199
99th_percentile_confirmation_ms: 9994
max_tower_distance: 164
last_tower_distance: 51.8333333333
slots_per_second: 2.127

PR #16723, opened 2021-04-21
`getLeaderSchedule` now supports filtered results based on validator identity

mvines:
It's wasteful for `getLeaderSchedule` to return the entire schedule when the client only cares about one validator

### DIFF
--- a/core/src/rpc.rs
+++ b/core/src/rpc.rs
@@ -2110,7 +2110,7 @@ pub mod rpc_minimal {
             &self,
             meta: Self::Metadata,
             slot: Option<Slot>,
-            commitment: Option<CommitmentConfig>,
+            config: Option<RpcLeaderScheduleConfig>,
         ) -> Result<Option<RpcLeaderSchedule>>;
     }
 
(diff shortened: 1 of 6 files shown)


mvines:
This was something that Nyet raised a while ago, finally got around to it.

mvines:
`accounts::tests::test_load_largest_accounts` failed.  Strike 1

core/src/rpc.rs
```diff
@@ -2110,7 +2110,7 @@ pub mod rpc_minimal {
             &self,
             meta: Self::Metadata,
             slot: Option<Slot>,
-            commitment: Option<CommitmentConfig>,
+            config: Option<RpcLeaderScheduleConfig>,
```
CriesofCarrots, line 2113: This issue predates this PR, but the combo of 2 options here means that to use `config`, a user *must* include a `slot` (it can be `null`, as I see you discovered in tests below). For both params to be truly optional, the solution looks something like this: https://github.com/solana-labs/solana/blob/1a658c7f31e1e0d2d39d9efbc0e929350e2c2bcb/client/src/rpc_config.rs#L214-L228
Not super critical, I guess, since no one's complained so far. But I can follow up with that rework?

CriesofCarrots (approved):
lgtm overall

core/src/rpc.rs
mvines: > But I can follow up with that rework?

Sure!  I didn't feel too badly about just passing in a null slot but I don't disagree at all that it would be nicer for it to be truly optional

PR #16728, opened 2021-04-21
Clean up "APR" language around inflation rewards

### LINKED ISSUE #16718: RPC JSON API is not clear about how inflation rate is measured, leading to knock-on effects
#### Problem

Solana implements an inflation system whose rate is based on time periods measured in slots.  Because slots take a variable length of time to complete, the wall-clock measured rate of inflation is variable and not easy to represent in a way that people are used to thinking about.  This is not a problem per se, but when terminology normally associated with wall clock time such as APR, APY, etc, are used when describing the Solana inflation rate, it can lead to much confusion and misinformation.

The RPC JSON API does not provide any documentation to assist those using the getInflationRate API in understanding this.  Although it is not strictly necessary to document this, it would be obviously beneficial to do so just to give one more place that naturally familiarized developers with the concepts to reduce overall confusion.

The solana command line program itself seems to confuse this issue - labelling the information obtained from getInflationRate as 'APR' -- which stands for Annual Percentage Rate.  However, this is a misnomer as the inflation rate is not tied to an annual schedule.  It was set to a rate that was *expected* to, for a given slot completion rate, yield a specific APR, but that is not the same thing as the *actual* APR being achieved, due to differences in actual slot completion rate.

For example, at the moment (2021.04.21), epochs are running at closer to 3 days per epoch, instead of the expected 2.5 days or so, resulting in an APR much lower than that reported by the "solana inflation" command.  I don't know what the exact numbers are -- I'd have to go look up the number of slots per year that the inflation schedule is based on -- but I do know that it's significantly lower than the 7.84% APR that the command line reports.

It may seem unimportant to some people, but accuracy in communication, especially surrounding financial aspects of Solana that people may feel misled by when they are not accurate, is, in my opinion, very important.

#### Proposed Solution

First, a minimum of details describing what the rate represents should be added to the documentation for getInflationRate.  It could be a simple statement like, "inflation is measured by number of slots and may not correspond to values obtained over the course of a true year -- see <insert_link_here> for details".

I realize that this many be unpopular, but I think that a new term needs to be used other than "APR" in the command line client output.  I can think of one ugly acronym - "EAPR" -- Epoch-Annual Percentage Rate.  This would make it at least clearer that the rate is based on an Epoch-Year, which would be defined as a certain number of epochs, and that is well understood to likely deviate from the true length of a wall-clock year.

Finally, this probably goes beyond the scope of a github issue, but I think some effort should be put into making sure that those sources of information currently serving as the de-facto public face of Solana (solanabeach.io, validators.app, etc), use correct terminology as well.

mvines:
New `solana inflation` output, APR is now removed entirely as it's misleading.  Instead just use rate:
```
$ solana inflation 
Inflation Governor:
Initial rate:             8.00%
Terminal rate:            1.50%
Rate reduction per year: 15.00%
* Rate reduction is derived using the target slot time in genesis config

Inflation for Epoch 172:
Total rate:               7.84%
Staking rate:             7.84%
Foundation rate:          0.00%
```

The `getInflationGovernor` RPC method documentation now uses the same language.

Also updated the output from `solana-ledger-tool genesis` to include the hard-coded target tick duration and slots-per-year values that are used by the Bank to compute inflation. I briefly considered exposing those values over RPC as well, but ultimately decided not to for now since there's no client demand yet.

Fixes #16718

### DIFF
--- a/cli-output/src/cli_output.rs
+++ b/cli-output/src/cli_output.rs
@@ -1405,24 +1405,28 @@ impl fmt::Display for CliInflation {
         if (self.governor.initial - self.governor.terminal).abs() < f64::EPSILON {
             writeln!(
                 f,
-                "Fixed APR:               {:>5.2}%",
+                "Fixed rate:              {:>5.2}%",
                 self.governor.terminal * 100.
             )?;
         } else {
             writeln!(
                 f,
-                "Initial APR:             {:>5.2}%",
+                "Initial rate:            {:>5.2}%",
                 self.governor.initial * 100.
             )?;
             writeln!(
                 f,
-                "Terminal APR:            {:>5.2}%",
+                "Terminal rate:           {:>5.2}%",
                 self.governor.terminal * 100.
             )?;
             writeln!(
                 f,
                 "Rate reduction per year: {:>5.2}%",
                 self.governor.taper * 100.
+            )?;
+            writeln!(
+                f,
+                "* Rate reduction is derived using the target slot time in genesis config"
             )?;
         }
         if self.governor.foundation_term > 0. {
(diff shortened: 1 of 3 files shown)


bji:
Wow thank you for the fast action!

bji (approved):

mvines:
downstream-projects build failure is unrelated

PR #16735, opened 2021-04-22
Retry latest vote if expired

carllin:
#### Problem
Follow-up to problem as described in https://github.com/solana-labs/solana/pull/16560. The issue with that solution is 
if people selectively rewrite votes in gossip, they may be able to influence fork choice.

#### Summary of Changes
1. Periodically refresh the latest vote `V` if it's expired, using the recent blockhash of the heaviest bank on the same fork as that last vote `V`
2. Add ability in `cluster_info` to "refresh" a vote, i.e. emplace a vote for the same slot with a newer one (use the same vote index in gossip)
2. Re-enable local cluster test showing issue is resolved
Fixes #

### DIFF
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -98,6 +98,7 @@ pub(crate) struct ComputedBankState {
     // Tree of intervals of lockouts of the form [slot, slot + slot.lockout],
     // keyed by end of the range
     pub lockout_intervals: LockoutIntervals,
+    pub my_latest_landed_vote: Slot,
 }
 
 #[frozen_abi(digest = "Eay84NBbJqiMBfE7HHH2o6e51wcvoU79g8zCi5sw6uj3")]
@@ -108,6 +109,11 @@ pub struct Tower {
     threshold_size: f64,
     lockouts: VoteState,
     last_vote: Vote,
+    // The blockhash used in the last vote transaction, may or may not equal the
+    // blockhash of the voted block itself, depending if the vote slot was refreshed.
+    // For instance, a vote for slot 5, may be refreshed/resubmitted for inclusion in
+    //  block 10, in  which case `last_vote_tx_blockhash` equals the blockhash of 10, not 5.
+    last_vote_tx_blockhash: Hash,
     last_timestamp: BlockTimestamp,
     #[serde(skip)]
     path: PathBuf,
@@ -134,6 +140,7 @@ impl Default for Tower {
             lockouts: VoteState::default(),
             last_vote: Vote::default(),
             last_timestamp: BlockTimestamp::default(),
+            last_vote_tx_blockhash: Hash::default(),
             path: PathBuf::default(),
             tmp_path: PathBuf::default(),
             stray_restored_slot: Option::default(),
@@ -234,6 +241,7 @@ impl Tower {
         // Tree of intervals of lockouts of the form [slot, slot + slot.lockout],
         // keyed by end of the range
         let mut lockout_intervals = LockoutIntervals::new();
+        let mut my_latest_landed_vote = 0;
         for (key, (voted_stake, account)) in vote_accounts {
             if voted_stake == 0 {
                 continue;
@@ -261,6 +269,7 @@ impl Tower {
             }
 
             if key == *node_pubkey || vote_state.node_pubkey == *node_pubkey {
+                my_latest_landed_vote = vote_state.nth_recent_vote(0).map(|v| v.slot).unwrap_or(0);
                 debug!("vote state {:?}", vote_state);
                 debug!(
                     "observed slot {}",
@@ -348,6 +357,7 @@ impl Tower {
             total_stake,
             bank_weight,
             lockout_intervals,
+            my_latest_landed_vote,
         }
     }
 
@@ -368,7 +378,7 @@ impl Tower {
         slot: Slot,
         hash: Hash,
         last_voted_slot_in_bank: Option<Slot>,
-    ) -> (Vote, Vec<Slot> /*VoteState.tower*/) {
+    ) -> Vote {
         let mut local_vote_state = local_vote_state.clone();
         let vote = Vote::new(vec![slot], hash);
         local_vote_state.process_vote_unchecked(&vote);
@@ -388,7 +398,15 @@ impl Tower {
             slots,
             local_vote_state.votes
         );
-        (Vote::new(slots, hash), local_vote_state.tower())
+        Vote::new(slots, hash)
+    }
+
+    pub fn tower_slots(&self) -> Vec<Slot> {
+        self.lockouts.tower()
+    }
+
+    pub fn last_vote_tx_blockhash(&self) -> Hash {
+        self.last_vote_tx_blockhash
     }
 
     fn last_voted_slot_in_bank(bank: &Bank, vote_account_pubkey: &Pubkey) -> Option<Slot> {
@@ -397,22 +415,23 @@ impl Tower {
         slot
     }
 
+    pub fn refresh_last_vote_tx_blockhash(&mut self, new_vote_tx_blockhash: Hash) {
+        self.last_vote_tx_blockhash = new_vote_tx_blockhash;
+    }
+
     pub fn record_bank_vote(
         &mut self,
         bank: &Bank,
         vote_account_pubkey: &Pubkey,
-    ) -> (Option<Slot>, Vec<Slot> /*VoteState.tower*/) {
-        let (vote, tower_slots) = self.new_vote_from_bank(bank, vote_account_pubkey);
-
+    ) -> (Option<Slot>, Vote) {
+        let vote = self.new_vote_from_bank(bank, vote_account_pubkey);
         let new_root = self.record_bank_vote_update_lockouts(vote);
-        (new_root, tower_slots)
+        self.last_vote.timestamp =
+            self.maybe_timestamp(self.last_vote.last_voted_slot().unwrap_or(0));
+        (new_root, self.last_vote.clone())
     }
 
-    pub fn new_vote_from_bank(
-        &self,
-        bank: &Bank,
-        vote_account_pubkey: &Pubkey,
-    ) -> (Vote, Vec<Slot> /*VoteState.tower*/) {
+    pub fn new_vote_from_bank(&self, bank: &Bank, vote_account_pubkey: &Pubkey) -> Vote {
         let voted_slot = Self::last_voted_slot_in_bank(bank, vote_account_pubkey);
         Self::new_vote(&self.lockouts, bank.slot(), bank.hash(), voted_slot)
     }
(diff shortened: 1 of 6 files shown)


core/src/replay_stage.rs
```diff
@@ -1469,11 +1461,104 @@ impl ReplayStage {
...
+                .unwrap_or(false)
+            // In order to avoid voting on multiple forks all past MAX_PROCESSING_AGE that don't
+            // include the last voted blockhash
+            || last_vote_refresh_time.elapsed().as_millis() < 5000
+        {
+            return;
+        }
+
+        // TODO: check the timestamp in this vote is correct, i.e. it shouldn't
+        // have changed from the original timestamp of the vote.
```
carllin, line 1500: @CriesofCarrots just wanted to check that this is safe for the time adjustments. Revoting for the same slot with a different blockhash, but the `Vote.timestamp` remains unchanged

core/src/consensus.rs
```diff
@@ -108,6 +109,12 @@ pub struct Tower {
     threshold_size: f64,
     lockouts: VoteState,
     last_vote: Vote,
+    #[serde(skip)]
+    // The blockhash used in the last vote transaction, may or may not equal the
+    // blockhash of the voted block itself, depending if the vote slot was refreshed.
+    // For instance, a vote for slot 5, may be refreshed/resubmitted for inclusion in
+    //  block 10, in  which case `last_vote_tx_blockhash` equals the blockhash of 10, not 5.
+    last_vote_tx_blockhash: Hash,
```
jstarry, line 117: Use an `Option<Hash>` here otherwise a node that just loaded their tower from disk will think `Hash::default()` isn't in the recent blockhash queue and retry their latest recorded vote even if it was recently sent

jstarry: Also, does this need to be stored here? Could it be recovered from gossip instead?

core/src/consensus.rs
```diff
@@ -363,13 +374,70 @@ impl Tower {
...
+        let (_stake, vote_account) = bank.get_vote_account(vote_account_pubkey)?;
+        let slot = vote_account.vote_state().as_ref().ok()?.last_voted_slot();
+        slot
+    }
+
+    pub fn refresh_last_vote_tx_blockhash(&mut self, new_vote_tx_blockhash: Hash) {
+        self.last_vote_tx_blockhash = new_vote_tx_blockhash;
+    }
+
+    pub fn record_bank_vote(&mut self, bank: &Bank, vote_account_pubkey: &Pubkey) -> Option<Slot> {
```
jstarry, line 395: nit: can you rearrange the methods here to make the diff easier to read? It's hard to tell what changed

core/src/replay_stage.rs
```diff
@@ -1469,11 +1462,104 @@ impl ReplayStage {
...
+        // Refresh the vote if our latest vote hasn't landed, and the recent blockhash of the
+        // last attempt at a vote transaction has expired
+        let last_voted_slot = last_voted_slot.unwrap();
+        if my_latest_landed_vote >= last_voted_slot
+            || heaviest_bank_on_same_fork
+                .check_hash_age(&tower.last_vote_tx_blockhash(), MAX_PROCESSING_AGE)
+                .unwrap_or(false)
+            // In order to avoid voting on multiple forks all past MAX_PROCESSING_AGE that don't
+            // include the last voted blockhash
+            || last_vote_refresh_time.elapsed().as_millis() < MAX_VOTE_REFRESH_INTERVAL_MILLIS as u128
```
jstarry, line 1495: I don't think this is needed because you call `refresh_last_vote_tx_blockhash` whenever a vote is refreshed

core/src/replay_stage.rs
CriesofCarrots: Yes. As long as both the slot and the timestamp are the same as in the previous vote, the Vote program timestamp-check code should be happy.

core/src/consensus.rs
```diff
@@ -261,6 +270,7 @@ impl Tower {
             }
 
             if key == *node_pubkey || vote_state.node_pubkey == *node_pubkey {
+                my_latest_landed_vote = vote_state.nth_recent_vote(0).map(|v| v.slot).unwrap_or(0);
```
jstarry, line 273: Since nodes can switch vote accounts, this should take the max: `my_latest_landed_vote = my_latest_landed_vote.max(...)`

carllin: Switched to check by vote account key instead of reverse indexing by node pubkey

core/src/replay_stage.rs
carllin: Yeah, but the issue is you can hop between multiple forks, if they're all descended from your last vote. None of the forks' blockhashes will be included in the others, so checking using the latest refreshed blockhash may not be enough

core/src/consensus.rs
```diff
@@ -234,6 +242,7 @@ impl Tower {
         // Tree of intervals of lockouts of the form [slot, slot + slot.lockout],
         // keyed by end of the range
         let mut lockout_intervals = LockoutIntervals::new();
+        let mut my_latest_landed_vote = 0;
```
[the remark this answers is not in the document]
carllin: hmmm, I think the `refresh_last_vote()` will catch this case with the check

```
let last_voted_slot = tower.last_voted_slot();
        if last_voted_slot.is_none() {
            return;
        }
```

If there is not vote history, the tower last voted slot will be empty as well so it won't refresh.

But I can see how an `Option` would be clearer, I can make the change

core/src/consensus.rs
carllin: I think recovering from gossip will be a lot more complexity

It should be fine for them to retry their latest recorded vote, as it should only happen *once* on startup. Also, if the vote was pushed by this node and already exists in gossip, this check here: https://github.com/solana-labs/solana/pull/16735/files#diff-acba1816adaf11741b4cd19f39d543dbdb6a1af4726544a8b61a6c4db616c76cR1215-R1219 will ensure the node does not replace that vote in gossip before receiving it from the network first.

core/src/consensus.rs
carllin: Yeah for sure, let me clean it up 😄

core/src/consensus.rs
```diff
@@ -363,13 +374,24 @@ impl Tower {
             .unwrap_or(false)
     }
 
-    fn new_vote(
```
carllin, line 366: replaced with `apply_vote_and_generate_vote_diff`

core/src/consensus.rs
```diff
@@ -388,44 +410,48 @@ impl Tower {
...
-    ) -> (Option<Slot>, Vec<Slot> /*VoteState.tower*/) {
-        let (vote, tower_slots) = self.new_vote_from_bank(bank, vote_account_pubkey);
+    pub fn record_bank_vote(&mut self, bank: &Bank, vote_account_pubkey: &Pubkey) -> Option<Slot> {
+        let last_voted_slot_in_bank = Self::last_voted_slot_in_bank(bank, vote_account_pubkey);
 
-        let new_root = self.record_bank_vote_update_lockouts(vote);
-        (new_root, tower_slots)
-    }
-
-    pub fn new_vote_from_bank(
```
carllin, line 411: folded into `apply_vote_and_generate_vote_diff`, much like `new_vote` from above

core/src/replay_stage.rs
jstarry: Ah got it, sounds good!

core/src/consensus.rs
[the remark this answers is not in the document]
jstarry: Ah true, thanks for the `Option` change

core/src/consensus.rs
jstarry: > will ensure the node does not replace [ ... ]

Doesn't that ensure the node *does* replace the vote in gossip?

core/src/replay_stage.rs
```diff
@@ -1684,7 +1785,7 @@ impl ReplayStage {
 
     #[allow(clippy::too_many_arguments)]
     pub(crate) fn compute_bank_stats(
-        my_pubkey: &Pubkey,
+        my_vote_pubkey: &Pubkey,
```
jstarry, line 1788: The caller of `compute_bank_stats` should have changed from `my_pubkey` to `vote_account` as well, right?

core/src/consensus.rs
carllin: So if the vote doesn't already exist, `vote_index` will be `None`, so nothing new will be pushed

core/src/replay_stage.rs
carllin: ugh yeah, good catch!

core/src/consensus.rs
```diff
@@ -388,44 +410,48 @@ impl Tower {
...
-        Self::new_vote(&self.lockouts, bank.slot(), bank.hash(), voted_slot)
+        // Returns the new root if one is made after applying a vote for the given bank to
+        // `self.lockouts`
+        self.record_bank_vote_and_update_lockouts(bank.slot(), bank.hash(), last_voted_slot_in_bank)
     }
 
-    pub fn record_bank_vote_update_lockouts(&mut self, vote: Vote) -> Option<Slot> {
-        let slot = vote.last_voted_slot().unwrap_or(0);
-        trace!("{} record_vote for {}", self.node_pubkey, slot);
+    pub fn record_bank_vote_and_update_lockouts(
```
jstarry, line 430: nit: doesn't need to be public

jstarry (dismissed):
Looks good, just a few small nits

core/src/replay_stage.rs
```diff
@@ -480,6 +490,11 @@ impl ReplayStage {
                         .select_forks(&frozen_banks, &tower, &progress, &ancestors, &bank_forks);
                     select_forks_time.stop();
 
+                    if let Some(heaviest_bank_on_same_voted_fork) = heaviest_bank_on_same_voted_fork.as_ref() {
+                        if let Some(my_latest_landed_vote) = progress.my_latest_landed_vote(heaviest_bank_on_same_voted_fork.slot()) {
+                            Self::refresh_last_vote(&mut tower, &cluster_info, heaviest_bank_on_same_voted_fork, &poh_recorder, my_latest_landed_vote, &vote_account, &authorized_voter_keypairs.read().unwrap(), &mut voted_signatures, has_new_vote_been_rooted, &mut last_vote_refresh_time);
```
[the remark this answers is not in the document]
carllin: yeah should be on different lines, eh?

aeyakovenko:
Resigning the same vote seems dicey.  Want to sync up on this? @carllin

carllin:
> Resigning the same vote seems dicey

Seems relatively safe if we're not changing the votes, just changing the blockhash? 

>. Want to sync up on this? @carllin

Yeah sure!

carllin:
Discussed with @aeyakovenko offline, concerns should be resolved 😃

Issue #16076, opened 2021-03-23
CLI: Program deploy fetches full leader schedule which can fail for API's which limit response size


jstarry:
#### Problem
Program deploy fetches full leader schedule which can fail for API's which limit response size

#### Proposed Solution
Switch to new RPC method `getSlotLeaders` if using an API server which supports that method

Blocked on https://github.com/solana-labs/solana/pull/16057

PR #16736, opened 2021-04-22
Add TPU client for sending txs to the current leader tpu port

jstarry:
#### Problem
Program deploy fetches full leader schedule which can fail for API's which limit response size

#### Summary of Changes
- Switch to `getSlotLeaders` API which requires less bandwidth
- Refactor tpu sending into new `TpuClient` which can be utilized by other programs

Fixes https://github.com/solana-labs/solana/issues/16076


mvines:
(a v1.6 backport is probably ok, we don't recommend v1.5 tools for dev work anymore)

jstarry:
@mvines the pubsub client we have isn't great and not failure resistant. I think it's ok for deploys but might be an issue for stake-o-matic if it's long running and experiences occasional websocket disconnects. I decided to push that off but feel free to disagree there

mvines:
> @mvines the pubsub client we have isn't great and not failure resistant. I think it's ok for deploys but might be an issue for stake-o-matic if it's long running and experiences occasional websocket disconnects. I decided to push that off but feel free to disagree there

Should be ok for my desired usage as well, and at worst case an excuse to improve the pubsub client

client/src/tpu_client.rs
```diff
@@ -0,0 +1,393 @@
...
+            )?)
+        } else {
+            None
+        };
+
+        let t_leader_tpu_service = Some({
+            let recent_slots = recent_slots.clone();
+            let leader_tpu_cache = leader_tpu_cache.clone();
+            std::thread::Builder::new()
+                .name("solana-leader-tpu-service".to_string())
```
mvines, line 276: The cryptic `ldr-tpu-srv` is ironically probably a more descriptive linux thread name than `solana-leader-`. I believe we get 15 characters

mvines (approved):
looks great

PR #16740, opened 2021-04-22
explorer: downgrade nodejs version in travis

jstarry:
#### Problem
Travis CI latest Node.js version is v16 as of 4/20 and this is not supported by node-sass yet

#### Summary of Changes
Use LTS node

Fixes #


solana-grimes:

PR #16743, opened 2021-04-22
Readonlyaccounts

jeffwashington:
#### Problem
#### Summary of Changes
lamports -> lamports()
Fixes #

### DIFF
--- a/runtime/src/message_processor.rs
+++ b/runtime/src/message_processor.rs
@@ -125,13 +125,13 @@ impl PreAccount {
 
         // An account not assigned to the program cannot have its balance decrease.
         if *program_id != pre.owner // line coverage used to get branch coverage
-         && pre.lamports > post.lamports
+         && pre.lamports() > post.lamports
         {
             return Err(InstructionError::ExternalAccountLamportSpend);
         }
 
         // The balance of read-only and executable accounts may not change
-        let lamports_changed = pre.lamports != post.lamports;
+        let lamports_changed = pre.lamports() != post.lamports;
         if lamports_changed {
             if !is_writable {
                 return Err(InstructionError::ReadonlyLamportChange);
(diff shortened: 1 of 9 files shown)


runtime/src/message_processor.rs
```diff
@@ -125,13 +125,13 @@ impl PreAccount {
...
         if *program_id != pre.owner // line coverage used to get branch coverage
-         && pre.lamports > post.lamports
+         && pre.lamports() > post.lamports
         {
             return Err(InstructionError::ExternalAccountLamportSpend);
         }
 
         // The balance of read-only and executable accounts may not change
-        let lamports_changed = pre.lamports != post.lamports;
+        let lamports_changed = pre.lamports() != post.lamports;
```
brooksprumo, line 134: Does `post.lamports` need to be `post.lamports()`?

runtime/src/message_processor.rs
```diff
@@ -125,13 +125,13 @@ impl PreAccount {
 
         // An account not assigned to the program cannot have its balance decrease.
         if *program_id != pre.owner // line coverage used to get branch coverage
-         && pre.lamports > post.lamports
+         && pre.lamports() > post.lamports
```
brooksprumo, line 128: Does `post.lamports` need to be `post.lamports()`?

runtime/src/rent_collector.rs
```diff
@@ -158,7 +158,7 @@ mod tests {
         assert_ne!(existing_account.rent_epoch, old_epoch);
 
         // newly created account should be collected for less rent; thus more remaining balance
-        assert!(created_account.lamports > existing_account.lamports);
+        assert!(created_account.lamports() > existing_account.lamports);
```
brooksprumo, line 161: Does `existing_account.lamports` need to be `existing_account.lamports()`?

brooksprumo (commented):
I only looked at the changes via this PR, not the files in their entirety, so I may've missed other uses of bare `.lamports`.

runtime/src/message_processor.rs
jeffwashington: maybe, will pick these up on a future pr when I do another batch of stragglers.

brooksprumo (approved):

PR #16747, opened 2021-04-22
Set lamports

jeffwashington:
Problem

Summary of Changes
.lamports = X -> .set_lamports(X)
Fixes #


brooksprumo:
@jeffwashington Can you now remove `pub` from the `lamports` field in the `AccountSharedData` struct in `account.rs`?

jeffwashington:
> @jeffwashington Can you now remove `pub` from the `lamports` field in the `AccountSharedData` struct in `account.rs`?

not yet. that's the goal. There are many uses. I broke the replacements into easy to diff changes.

brooksprumo (approved):
Looks good to me!

PR #16748, opened 2021-04-22
validates gossip addresses before sending pull-requests

behzadnouri:
#### Problem
ip addresses need to be validated before sending packets to them.

#### Summary of Changes
* send a ping packet to ip addresses before any pull requests.
* only pull from ip addresses which have responded with the correct hash of their respective ping packet.


behzadnouri:
any objection to merging this?

sakridge (dismissed):
lgtm

PR #16749, opened 2021-04-22
Rpc: Rework `getLeaderSchedule` options

CriesofCarrots:
#### Problem
The parameters in `getLeaderSchedule` are not truly optional. In order to use `commitment` or `identity`, a slot must be provided.

#### Summary of Changes
Rework `getLeaderSchedule` options so that slot is truly optional, even with config options included

Fixes https://github.com/solana-labs/solana/pull/16723#discussion_r617960638


mvines (approved):
thanks, better!

PR #16751, opened 2021-04-22
Add minimumContiguousLedgerSlot RPC method

mvines:
[minimumLedgerSlot](https://docs.solana.com/developing/clients/jsonrpc-api#minimumledgerslot) doesn't consider ledger holes that might arise from restarting a node from snapshot, add a method that is ledger hole aware


ledger/src/blockstore.rs
```diff
@@ -2493,6 +2493,30 @@ impl Blockstore {
             .collect())
     }
 
+    pub fn get_lowest_connected_full_parent(&self, mut slot: Slot) -> Result<Slot> {
```
mvines, line 2496: TODO:
* [ ] Write a test

ledger/src/blockstore.rs
```diff
@@ -2493,6 +2493,30 @@ impl Blockstore {
...
+        let mut slot_meta = self.meta(slot)?.ok_or(BlockstoreError::SlotUnavailable)?;
+        if !slot_meta.is_full() {
+            return Err(BlockstoreError::SlotUnavailable);
+        }
+
+        loop {
```
CriesofCarrots, line 2502: I think this is going to get really slow when it's looping over 500k slots; I'd estimate at least 30s (AncestorIterator performs similarly). Should we maybe be caching this data in the Blockstore struct, similar to `lowest_cleanup_slot` (but updated on snapshot boots)?

mvines: Cool that's what I was thinking of doing next.  

Overall wdyt about this new method?  I have a couple use cases in mind:
1. For the stake-o-matic, I want to confirm the node I'm talking with is not missing data for the previous epoch
2. bare-metal-rpc folks could use this method as an additional test for bringing backend nodes into the pool by comparing the responses from `minimumLedgerSlot` and `minimumContiguousLedgerSlot`

🤔 perhaps for (2), `minimumContiguousLedgerSlot` should also return the `minimumLedgerSlot` value too?  Deprecate `minimumLedgerSlot` and create a `getMinimumLedgerSlot` that aligns with the naming of all the other RPC getters and returns both slots?

CriesofCarrots: > Deprecate `minimumLedgerSlot` and create a `getMinimumLedgerSlot` that aligns with the naming of all the other RPC getters and returns both slots?

I think this sounds great ^

For stake-o-matic, does Blockstore need to have all the data? Or can it fall back to BigTable? Just thinking about this method vis-a-vis to `getFirstAvailableBlock` (which includes BigTable, by turns helpful and annoying).

mvines: Folding getFirstAvailableBlock into all this seems nice too.

For SOM though, I’m planning to rewrite it to use the SlotHistory sysvar instead of fetching the confirmed block list.  In all normal cases, SlotHistory will have the slot bitmap from the previous epoch on it.  I just need this PR to tell me that the SlotHistory bitmap isn’t missing any data

mvines:
Nope, I don't need this after all.

PR #16777, opened 2021-04-23
refactor complicated len calculation

jeffwashington:
#### Problem
This code is about to be refactored. This duplicate lookup of len is complicated.
#### Summary of Changes
Calculate it once and use local.
Fixes #


brooksprumo (approved):
Looks good to me (assuming checks pass)!

PR #16783, opened 2021-04-23
owner -> owner()

jeffwashington:
Problem

Summary of Changes
.owner -> .owner()

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -280,7 +280,7 @@ impl<'a> LoadedAccount<'a> {
     pub fn owner(&self) -> &Pubkey {
         match self {
             LoadedAccount::Stored(stored_account_meta) => &stored_account_meta.account_meta.owner,
-            LoadedAccount::Cached((_, cached_account)) => &cached_account.account.owner,
+            LoadedAccount::Cached((_, cached_account)) => &cached_account.account.owner(),
         }
     }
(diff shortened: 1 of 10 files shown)


runtime/src/accounts_db.rs
```diff
@@ -3254,7 +3254,7 @@ impl AccountsDb {
         let mut hasher = Hasher::default();
 
         hasher.hash(&account.data());
-        hasher.hash(&account.owner.as_ref());
+        hasher.hash(&account.owner().as_ref());
```
brooksprumo, line 3257: Doesn't `.owner()` already return a ref?

jeffwashington: as_ref on the pubkey causes it to return a &[u8], which is what hasher wants.
The leading & appears to be unnecessary but acceptable.

jeffwashington: in pubkey.rs:
impl AsRef<[u8]> for Pubkey {
    fn as_ref(&self) -> &[u8] {
        &self.0[..]
    }
}

jeffwashington: This was a great question, though.

brooksprumo: Gotcha, thanks!

brooksprumo (approved):

Issue #16580, opened 2021-04-15
Dropping non-rooted banks does not clean up keys


brooksprumo:
@sakridge found an issue where non-rooted banks do not clean up their keys when dropped (see PR #15106).

`Bank::clean_accounts()` should be fixed, and new tests should be added to confirm the fix.

PR #16786, opened 2021-04-23
Collect uncleaned pubkeys from all slots

brooksprumo:
While working on another issue (#16580), the list of uncleaned pubkeys
returned from `remove_uncleaned_slots_and_collect_pubkeys_up_to_slot()`
did not include unrooted slots.  This meant that during cleaning,
unrooted slots would not have their pubkeys cleaned up properly.

Now, return all uncleaned pubkeys, regardless if the slot is rooted or
not.  Additionally, update the tests to have unrooted slots to ensure
this behavior.

This is part two of PR #16879, and originally based on PR #15106.

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -1479,25 +1479,18 @@ impl AccountsDb {
 
     /// Remove `slots` from `uncleaned_pubkeys` and collect all pubkeys
     ///
-    /// For each slot in the list of slots, remove it from the `uncleaned_pubkeys` Map and collect
-    /// all the pubkeys from root slots to return.
+    /// For each slot in the list of uncleaned slots, remove it from the `uncleaned_pubkeys` Map
+    /// and collect all the pubkeys to return.
     fn remove_uncleaned_slots_and_collect_pubkeys(
         &self,
         uncleaned_slots: Vec<Slot>,
     ) -> Vec<Vec<Pubkey>> {
         uncleaned_slots
             .into_iter()
-            .filter_map(|slot| {
-                let maybe_slot_keys = self.uncleaned_pubkeys.remove(&slot);
-                if self.accounts_index.is_root(slot) {
-                    // Safe to unwrap on rooted slots since this is called from clean_accounts
-                    // and only clean_accounts operates on rooted slots. purge_slots only
-                    // operates on uncleaned_pubkeys
-                    let (_slot, keys) = maybe_slot_keys.expect("Root slot should exist");
-                    Some(keys)
-                } else {
-                    None
-                }
+            .filter_map(|uncleaned_slot| {
+                self.uncleaned_pubkeys
+                    .remove(&uncleaned_slot)
+                    .map(|(_removed_slot, removed_pubkeys)| removed_pubkeys)
             })
             .collect()
     }
@@ -10163,7 +10156,7 @@ pub mod tests {
         db.store_uncached(slot3, &[(&pubkey3, &account3)]);
 
         db.add_root(slot1);
-        db.add_root(slot2);
+        // slot 2 is _not_ a root on purpose
         db.add_root(slot3);
 
         db.uncleaned_pubkeys.insert(slot1, vec![pubkey1]);
@@ -10220,7 +10213,7 @@ pub mod tests {
         db.store_uncached(slot2, &[(&pubkey2, &account2)]);
         db.store_uncached(slot3, &[(&pubkey3, &account3)]);
 
-        db.add_root(slot1);
+        // slot 1 is _not_ a root on purpose
         db.add_root(slot2);
         db.add_root(slot3);
 


runtime/src/accounts_db.rs
```diff
@@ -1464,39 +1464,42 @@ impl AccountsDb {
...
-            .collect();
-        (
-            slots
-                .into_iter()
-                .filter_map(|slot| {
-                    let maybe_slot_keys = self.uncleaned_pubkeys.remove(&slot);
-                    if self.accounts_index.is_root(slot) {
-                        // Safe to unwrap on rooted slots since this is called from clean_accounts
-                        // and only clean_accounts operates on rooted slots. purge_slots only
-                        // operates on uncleaned_pubkeys
```
brooksprumo, line 1490: You'll notice that I took out the check for `is_root()`. Shouldn't _all_ the pubkeys be collected from slot removal, instead of just the pubkeys that came from rooted slots?

I was thinking that either (1) all the slots in `uncleaned_pubkeys` are rooted slots anyway, or (2) if there are _non-rooted_ slots, then those pubkeys should definitely be cleaned up.

Am I missing something here?

runtime/src/accounts_db.rs
```diff
@@ -1515,7 +1518,7 @@ impl AccountsDb {
 
         let mut collect_delta_keys = Measure::start("key_create");
         let max_slot = max_clean_root.unwrap_or_else(|| self.accounts_index.max_root());
-        let (delta_keys, _max_slot) = self.collect_uncleaned_pubkeys_to_slot(max_slot);
```
brooksprumo, line 1518: Since `max_slot` is never used, I figured I'd remove it from here, and therefore remove it from the functions. I'm assuming that's OK, yes?

brooksprumo:
@jeffwashington @ryoqun @carllin @sakridge Can y'all take a look at this PR and my comments to see if (1) this is OK, and (2) if I missed anything? Also, do I need to add new tests for something like this?

runtime/src/accounts_db.rs
ryoqun: @brooksprumo yeah, it looks just an oversight from quick look. I think you can check `git log` of relevant lines to make informed verdict for that. :) And could you write short findings from the historian work here?

runtime/src/accounts_db.rs
```diff
@@ -1464,39 +1464,42 @@ impl AccountsDb {
...
+            .collect()
+    }
+
+    /// Remove uncleaned slots, up to a maximum slot, and return the collected pubkeys
+    ///
+    fn remove_uncleaned_slots_and_collect_pubkeys_up_to_slot(
+        &self,
+        max_slot: Slot,
+    ) -> Vec<Vec<Pubkey>> {
+        let slots = self.collect_uncleaned_slots_up_to_slot(max_slot);
```
ryoqun, line 1501: nits: how about renaming `slots` => `uncleaned_slots` to be consistent with method names?

runtime/src/accounts_db.rs
```diff
@@ -1464,39 +1464,42 @@ impl AccountsDb {
...
             .filter_map(|entry| {
-                let slot = entry.key();
-                max_slot_in_uncleaned_pubkeys = max_slot_in_uncleaned_pubkeys.max(*slot);
-                if *slot <= max_slot {
-                    Some(*slot)
-                } else {
-                    None
-                }
+                let slot = *entry.key();
+                (slot <= max_slot).then(|| slot)
```
ryoqun, line 1475: i like this `.then()` rusty idiomatic.

runtime/src/accounts_db.rs
ryoqun: @brooksprumo (Okay, I was forgetting everything about AccountsDb and remembered it to some extent..)

you have a good questions from the newcomer's eyes. :) And, I think this is good exercise for you to read through the AccountsDb code.

> (1) all the slots in uncleaned_pubkeys are rooted slots anyway

I don't think this is true. And I think this is easy to prove. With that hint given, could you read the code to find the _why_?

> (2) if there are non-rooted slots, then those pubkeys should definitely be cleaned up.

as you say, the need of cleaning is right. but non-rooted slots should be handled carefully. and that's fun part of AccountsDB. Unless rooted, banks are like snapshots/immutable collection for the whole accounts state. So, each bank must maintain delta of account updates. That's why we pass around `Ancestors`. So, new minor fork (child bank) can be created from any banks (parent banks). And then those un-rooted banks are finally promoted to rooted banks by  `squash()` and its caller. I think these code are also helpful to understand the high-level workings.

sorry for a bit being teasing. no intention to be mean to you. I just want you to be levelled up for AccountsDB. Code reading to fill the questions with hints might be good excises for you. :)

sakridge: @ryoqun Yes, these will included rooted slots, but because we only call `clean_acounts` on a rooted bank and/or we only iterate up to a rooted slot, these slots will be unrooted slots below the root and thus are dead and cleanable. Is there any issue you see?

We call `clean_accounts` on rooted banks:
https://github.com/solana-labs/solana/blob/8f56c116d71e5349b8a63e5d478c37033cfbdc2c/runtime/src/accounts_background_service.rs#L305

Max slot here will either be highest root or root -1:
https://github.com/solana-labs/solana/blob/8f56c116d71e5349b8a63e5d478c37033cfbdc2c/runtime/src/bank.rs#L4720

```
                // Note this only cleans up the storage entries. The accounts index cleaning
                // (removing from the slot list, decrementing the account ref count), is handled in
                // clean_accounts() -> purge_older_root_entries()
```
https://github.com/solana-labs/solana/blob/8f56c116d71e5349b8a63e5d478c37033cfbdc2c/runtime/src/accounts_db.rs#L3034-L3036

`clean_accounts -> purge_older_root_entries` won't be operating on the keys touched in these unrooted slots if it does not look at the list of delta keys.

runtime/src/accounts_db.rs
brooksprumo: OK, looks like the `max_slot` was used in the original commit `ca7914aa3d73f17217de498a4e5dcec3f0b9ce0d`, but was then reverted the next day in commit `ad2e10e17b521d6a880a5e4958787d1cbd55247a`.

Two days after that, in commit `0d32a0e0f44f53ba13189c10ee4b85da2cc8ca68`, `collect_uncleaned_pubkeys_to_slot()` was re-added via a re-revert (un-revert?), but the use of the `max_slot` return parameter was removed. Inside `construct_candidate_clean_keys()`, when `collect_uncleaned_pubkeys_to_slot()` is called, the `max_slot` was always ignored, so my guess is the re-revert should've removed `max_slot` here, but wasn't.

runtime/src/accounts_db.rs
brooksprumo: Will do! (Implemented in PR #16879)

runtime/src/accounts_db.rs
ryoqun: i see. perfect answer. thanks for the write up.

runtime/src/accounts_db.rs
ryoqun: @sakridge yeah, it looks like you're right. those `slots` are actually `roots` proven by the current codepath. I was short-sighted...

@brooksprumo sorry for confusion. So, it looks like `if self.accounts_index.is_root(slot) {` has been redundant check to begin with. Then removing this check sounds ok to me according to @sakridge 's reasonsing.

> > (1) all the slots in `uncleaned_pubkeys` are rooted slots anyway

> I don't think this is true.

Here's my thinking regarding the above statement:

(I failed once but if I read the code correctly,) `uncleaned_pubkeys` itself should be mixed with rooted and unrooted slots at the time of cleaning because the below is the sole code `insert()`-ing into `uncleaned_pubkeys` and it's called for both the cases of rooted and unrooted at the time of the insertion (immediately before `Bank.freeze()`):

https://github.com/solana-labs/solana/blob/1c95e2bbee0a221f7ff4a4666605363aabc26a74/runtime/src/accounts_db.rs#L4269

Does this code reading align with your reading?

sakridge: @ryoqun 
>> I don't think this is true.

> Here's my thinking regarding the above statement:

yes, I believe this statement is correct. The slots in `uncleaned_pubkeys` will be all rooted and unrooted slots.

I believe the origination of this is that previously the `clean_accounts` only operated on rooted slots, so we kept that behavior. Although the caching code may have added new assumptions about where unrooted slots were being cleaned up in the index.

sakridge (approved):

PR #16791, opened 2021-04-23
account.lamports +=

jeffwashington:
Problem

Summary of Changes
Modify code to return Results and use checked_add_lamports.
Fixes #


t-nelson (approved):
lgtm!

PR #16800, opened 2021-04-24
records hash of values purged by expired pull-responses

behzadnouri:
#### Problem
process_pull_responses should record hash of values purged by expired
responses (as well as unexpired ones):
https://github.com/solana-labs/solana/blob/c1829dd00/core/src/crds_gossip_pull.rs#L385-L387

otherwise, these values are not excluded from following pull-requests
(from likely different nodes):
https://github.com/solana-labs/solana/blob/c1829dd00/core/src/crds_gossip_pull.rs#L447-L452

and would waste bandwidth should they be included in subsequent
pull-responses.

#### Summary of Changes
* record hash of values purged by expired pull-responses.


carllin (approved):
Nice, good idea!

PR #16803, opened 2021-04-24
experiment with clean_rooted_entries

jeffwashington:
#### Problem
we have a situation where old 0 lamport accounts are not getting cleaned. I have a theory on tracking it down to at least 1 issue. When we clean_rooted_entries, we remove items from the index's slot list, but we do not deref the refcount. The result is accounts/stores are never getting cleaned. This 'fixes' the specific problem I'm working on (and does not hit the assert). After cleaning several times, the range of roots goes from 2M to 800k, which is still too large, but I'm not done.

I'm observing that if an account is loaded with 2 refcounts initially, then the refcount remains 2, even if we clean older roots and throw away those stores. Since the refcount is at 2, the store remains alive, the accounts in it remain unpurged, the roots that match the slot remain active, ...

#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/accounts_index.rs
+++ b/runtime/src/accounts_index.rs
@@ -1158,16 +1158,19 @@ impl<T: 'static + Clone + IsCached + ZeroLamport> AccountsIndex<T> {
         reclaims: &mut SlotList<T>,
         max_clean_root: Option<Slot>,
         account_indexes: &HashSet<AccountIndex>,
-    ) {
+    ) -> usize {
+        let mut count = 0;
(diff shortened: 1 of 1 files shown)


jeffwashington:
root stats:
roots: 460977, min: 71500402, max: 73790273, range: 2289871 (loaded from snapshot)
roots: 391303, min: 71999188, max: 73790273, range: 1791085 (after clean)
roots: 388002, min: 72921034, max: 73790273, range: 869239 (after another clean)
further cleans do nothing

PR #16806, opened 2021-04-24
removes delayed crds inserts when upserting gossip table

behzadnouri:
#### Problem
It is crucial that `VersionedCrdsValue::insert_timestamp` does not go
backward in time:
https://github.com/solana-labs/solana/blob/ec37a843a/core/src/crds.rs#L67-L79

Otherwise methods such as `get_votes` and `get_epoch_slots_since` will
break, which will break their downstream flow, including vote-listener
and optimistic confirmation:

For that, `Crds::new_versioned` is intended to be called "atomically" with
`Crds::insert_verioned` (as the comment already says so):
https://github.com/solana-labs/solana/blob/ec37a843a/core/src/crds.rs#L126-L129

However, currently this is violated in the code. For example,
`filter_pull_responses` creates `VersionedCrdsValue`s (with the current
timestamp), then acquires an exclusive lock on gossip, then
`process_pull_responses` writes those values to the crds table:
https://github.com/solana-labs/solana/blob/ec37a843a/core/src/cluster_info.rs#L2375-L2392

Depending on the workload and lock contention, the insert_timestamps may
well be in the past when these values finally are inserted into gossip.

#### Summary of Changes
To avoid such scenarios, this commit:
  * removes `Crds::new_versioned` and `Crd::insert_versioned`.
  * makes `VersionedCrdsValue` constructor private, only invoked in
    `Crds::insert`, so that `insert_timestamp` is populated right before
    insert.

This will improve `insert_timestamp` monotonicity as long as `Crds::insert`
is not called with a stalled timestamp. Following commits may further
improve this by calling `timestamp()` inside `Crds::insert`, and/or
switching to `std::time::Instant` which guarantees monotonicity.

also related: https://github.com/solana-labs/solana/pull/16808

### DIFF
--- a/core/benches/crds_shards.rs
+++ b/core/benches/crds_shards.rs
@@ -3,30 +3,34 @@
 extern crate test;
 
 use rand::{thread_rng, Rng};
-use solana_core::contact_info::ContactInfo;
-use solana_core::crds::VersionedCrdsValue;
-use solana_core::crds_shards::CrdsShards;
-use solana_core::crds_value::{CrdsData, CrdsValue};
-use solana_sdk::pubkey;
+use solana_core::{
+    crds::{Crds, VersionedCrdsValue},
+    crds_shards::CrdsShards,
+    crds_value::CrdsValue,
+};
 use solana_sdk::timing::timestamp;
+use std::iter::repeat_with;
 use test::Bencher;
 
 const CRDS_SHARDS_BITS: u32 = 8;
 
-fn new_test_crds_value() -> VersionedCrdsValue {
-    let data = CrdsData::ContactInfo(ContactInfo::new_localhost(&pubkey::new_rand(), timestamp()));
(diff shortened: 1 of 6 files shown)


behzadnouri:
this is on top of https://github.com/solana-labs/solana/pull/16800

carllin (approved):
Good stuff! 

Was going to ask if it was reasonable to take the timestamp inside the `insert()` here: https://github.com/solana-labs/solana/pull/16806/files#diff-16b34c14163de65972a6481cd8823161ab4d9faeb34e1c89b15bf103a5957475R140, but I see your PR description has already addressed this 😄. 

>This will improve insert_timestamp monotonicity as long as Crds::insert
is not called with a stalled timestamp.

For now, as long as we're calling `insert()` with a timestamp grabbed after acquiring the gossip write lock, hopefully this shouldn't corrupt the ordering right?

PR #16807, opened 2021-04-24
web3.js: introduce getInflationReward to connection

oJshua:
#### Problem
Web3.js needs to support the `getInflationReward` RPC call.

#### Summary of Changes
Add `getInflationReward` support to Connection object.

### DIFF
--- a/web3.js/test/connection.test.ts
+++ b/web3.js/test/connection.test.ts
@@ -545,6 +545,51 @@ describe('Connection', () => {
     }
   });
 
+  it('get inflation reward', async () => {
+    await mockRpcResponse({
+      method: 'getInflationReward',
+      params: [
+        [
+          '7GHnTRB8Rz14qZQhDXf8ox1Kfu7mPcPLpKaBJJirmYj2',
+          'CrinLuHjVGDDcQfrEoCmM4k31Ni9sMoTCEEvNSUSh7Jg',
+        ],
+        {
+          epoch: 0,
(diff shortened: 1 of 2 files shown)


web3.js/test/connection.test.ts
```diff
@@ -545,6 +545,51 @@ describe('Connection', () => {
...
+        [
+          new PublicKey('7GHnTRB8Rz14qZQhDXf8ox1Kfu7mPcPLpKaBJJirmYj2'),
+          new PublicKey('CrinLuHjVGDDcQfrEoCmM4k31Ni9sMoTCEEvNSUSh7Jg'),
+        ],
+        0,
+      );
+    } catch (error) {
+      // a live server needs to have slots up to 432000
+      // in order to have rewards for epoch 0
+      if (!mockServer && error.message.match(/Block not available/)) {
```
jstarry, line 583: how about only running this test as a mock then?

jstarry (approved):

web3.js/test/connection.test.ts
CriesofCarrots: +1
However, solana-test-validator now supports `--slots-per-epoch`, so if there are similar tests in the future that are important to run live, we could boot the CI test validator with shorter epochs.

oJshua: Nice, good to know @CriesofCarrots.

PR #16808, opened 2021-04-24
uses current timestamp when flushing local pending push queue

behzadnouri:
#### Problem
`local_message_pending_push_queue` is recording timestamps at the time the
value is created, and uses that when the pending values are flushed:
https://github.com/solana-labs/solana/blob/ec37a843a/core/src/cluster_info.rs#L321
https://github.com/solana-labs/solana/blob/ec37a843a/core/src/crds_gossip.rs#L96-L102

which is then used as the `insert_timestamp` when inserting values in the
crds table:
https://github.com/solana-labs/solana/blob/ec37a843a/core/src/crds_gossip_push.rs#L183

The flushing may happen 100ms after the values are created (or even
later if there is a lock contention). This will cause non-monotone
`insert_timestamp`s in the crds table (where time goes backward),
hindering the usability of insert_timestamps for other computations.

For example both `ClusterInfo::get_votes` and `get_epoch_slots_since` rely
on monotone `insert_timestamp`s when values are inserted into the table:

#### Summary of Changes
This commit removes timestamps from `local_message_pending_push_queue` and
uses current timestamp when flushing the queue.

also related: https://github.com/solana-labs/solana/pull/16806

### DIFF
--- a/core/src/crds_gossip.rs
+++ b/core/src/crds_gossip.rs
@@ -62,15 +62,12 @@ impl CrdsGossip {
         values
             .into_iter()
             .filter_map(|val| {
-                let res = self
+                let old = self
                     .push
-                    .process_push_message(&mut self.crds, from, val, now);
-                if let Ok(Some(val)) = res {
-                    self.pull.record_old_hash(val.value_hash, now);
-                    Some(val)
-                } else {
-                    None
-                }
+                    .process_push_message(&mut self.crds, from, val, now)
+                    .ok()?;
+                self.pull.record_old_hash(old.as_ref()?.value_hash, now);
+                old
             })
             .collect()
     }
@@ -93,8 +90,12 @@ impl CrdsGossip {
         prune_map
     }
 
-    pub fn process_push_messages(&mut self, pending_push_messages: Vec<(CrdsValue, u64)>) {
-        for (push_message, timestamp) in pending_push_messages {
+    pub(crate) fn process_push_messages(
+        &mut self,
+        pending_push_messages: Vec<CrdsValue>,
+        timestamp: u64,
+    ) {
+        for push_message in pending_push_messages {
             let _ =
                 self.push
                     .process_push_message(&mut self.crds, &self.id, push_message, timestamp);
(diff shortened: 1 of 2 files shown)


core/src/crds_gossip.rs
```diff
@@ -103,10 +104,10 @@ impl CrdsGossip {
 
     pub fn new_push_messages(
         &mut self,
-        pending_push_messages: Vec<(CrdsValue, u64)>,
+        pending_push_messages: Vec<CrdsValue>,
         now: u64,
     ) -> HashMap<Pubkey, Vec<CrdsValue>> {
-        self.process_push_messages(pending_push_messages);
+        self.process_push_messages(pending_push_messages, now);
```
carllin, line 110: Will using the same "now" be a problem if there are two messages with the same key sitting in the push queue?

carllin:
> The flushing may happen 100ms after the values are created (or even
> later if there is a lock contention). This will cause non-monotone
> `insert_timestamp`s in the crds table (where time goes backward),
> hindering the usability of insert_timestamps for other computations.

Great catch!

So for clarification, is the problem that the `local_message_pending_push_queue` isn't necessarily ordered by the insert timestamps? I.e., an earlier entry in the vector could have a later timestamp, because each thread is taking its own wallclock timestamps on insert before grabbing the `local_message_pending_push_queue` lock?

Could we also add a test, maybe simulating a push for votes, then a push for epoch slots, that shows:

1) The ordering is now consistent after the fix
2) The `ClusterInfo::get_votes` and `get_epoch_slots_since()` now return the right ordering?

core/src/crds_gossip.rs
behzadnouri: I think it would be fine. The 2nd one will overwrite the 1st one since the vector is processed in order and that is consistent with how values are pushed to this queue.

behzadnouri (commented):
> So for clarification, is the problem that currently, the `local_message_pending_push_queue` isn't necessarily ordered by the insert timestamps? I.e., an earlier entry in the vector could have a later timestamp, because each thread is taking its own wallclock timestamps on insert before grabbing the `local_message_pending_push_queue` lock?

The final goal is to ensure that when values are inserted into the gossip table their `insert_timestamp`s are monotone. So it does not go backward. e.g. we insert a value in the crds table with `insert_timestamp = 2` and then the next value is inserted with `insert_timestamp = 1`. This should be across **all** code paths which insert values into the crds table, not just pending-push messages. 

This commit by itself is not enough to guarantee that. It only reduces scenarios where `insert_timestamp`s may go backward. We also need https://github.com/solana-labs/solana/pull/16806 to get merged, and probably more patches to rework the `insert_timestamp`s as mentioned in the bottom of https://github.com/solana-labs/solana/pull/16806 description.

carllin (approved):

PR #16809, opened 2021-04-24
indexes crds values by their insert order

behzadnouri:
#### Problem
Having an ordinal index on crds values based on insert order allows to
efficiently filter values using a cursor. In particular
`CrdsGossipPush::push_messages` hash-map can be replaced with a cursor,
saving on the bookkeepings, purging, etc

#### Summary of Changes
* added an index for crds values on their insert order.
* replace `CrdsGossipPush::push_messages` with a cursor.

### DIFF
--- a/core/src/crds.rs
+++ b/core/src/crds.rs
@@ -56,6 +56,8 @@ pub struct Crds {
     epoch_slots: BTreeMap<u64 /*insert order*/, usize /*index*/>,
     // Indices of all crds values associated with a node.
     records: HashMap<Pubkey, IndexSet<usize>>,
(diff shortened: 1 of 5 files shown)


behzadnouri:
`cluster_info_stats2::purge`: this code on the left, vs current code on the right:

cluster_info_stats2 purge

behzadnouri:
This is on top of https://github.com/solana-labs/solana/pull/16952

carllin (commented):
Nice!

For bookkeeping, I think the PR title could be updated to be "Use ordinal indexes for gossip push messages"

carllin (approved):

PR #16812, opened 2021-04-25
Annotate clock constants with computed values for quick reference

jstarry:
#### Problem
I always forget how many slots are in an epoch

#### Summary of Changes
Annotated clock consts with computed values thru compile time assertions so that they never fall out of date

Fixes #


mvines (approved):
Helpful!

PR #16824, opened 2021-04-26
Make test_hash_stored_account pass in release mode

### LINKED ISSUE #16823: accounts_db::test_hash_stored_account fails in release mode
#### Problem

The following fails:

```
$ cargo test --package solana-runtime --release test_hash_stored_account
```
```
thread 'accounts_db::tests::test_hash_stored_account' panicked at 'assertion failed: `(left == right)`
  left: `33ruy7m3Xto7irYfsBSN74aAzQwCQxsfoZxXuZy2Rra3`,
 right: `4StuvYHFd7xuShVXB94uHHvpqGMCaacdZnYB74QQkPA1`: StoredAccountMeta's data layout might be changed; update hashing if needed.', runtime/src/accounts_db.rs:7394:9
```

#### Proposed Solution

It is possible to add a second `expected_account_hash` in the test with `if cfg!(debug_assertions) { ... } else { ... }`. This makes the test pass, but I wonder if this is the proper solution. Is it expected that the hash differs between debug and release mode?

ruuda:
#### Problem

In the `accounts_db::test_hash_stored_account` test, the hash differs between debug and release mode, which makes the test fail when running under `cargo test --release`.

#### Summary of Changes

Use `cfg!` to include a different expected hash in release mode. This does make the test pass, but I’m not sure if it is a proper solution — is the hash expected to differ across debug/release?

Fixes #16823

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -7538,8 +7538,12 @@ pub mod tests {
             hash: &hash,
         };
         let account = stored_account.clone_account();
-        let expected_account_hash =
-            Hash::from_str("4StuvYHFd7xuShVXB94uHHvpqGMCaacdZnYB74QQkPA1").unwrap();
+
+        let expected_account_hash = if cfg!(debug_assertions) {
+            Hash::from_str("4StuvYHFd7xuShVXB94uHHvpqGMCaacdZnYB74QQkPA1").unwrap()
+        } else {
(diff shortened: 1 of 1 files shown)


ruuda:
@stale does this comment count as “further activity”?

ruuda:
@stale further activity

ruuda:
@stale further activity

ruuda:
@stale further activity

mvines:
Thanks for the change, sorry it took so long to merge!

PR #16826, opened 2021-04-26
cli: Add ALL support to withdraw-stake subcommand

mvines:
Cleaning up a pile of random testnet stake is pretty rough without the ability to use `ALL` with `solana withdraw-stake`


CriesofCarrots:
Beat me to it :) I just ran into this on Friday

PR #16829, opened 2021-04-26
clean_dead_slot grabs roots lock once

jeffwashington:
Problem
on clean_dead_slot, we get a read lock to check for root, then we get a write lock to remove the root. If we are a high% of the time finding that the item is a root, then we can improve performance and reduce lock contention by just getting the write lock and removing the root the first time.
#16849 adds metrics to determine the reality.

Summary of Changes
Get root lock once.
Fixes #

### DIFF
--- a/runtime/src/accounts_index.rs
+++ b/runtime/src/accounts_index.rs
@@ -1235,30 +1235,28 @@ impl<T: 'static + Clone + IsCached + ZeroLamport> AccountsIndex<T> {
     /// Remove the slot when the storage for the slot is freed
     /// Accounts no longer reference this slot.
     pub fn clean_dead_slot(&self, slot: Slot) -> Option<AccountsIndexRootsStats> {
-        if self.is_root(slot) {
-            let (roots_len, uncleaned_roots_len, previous_uncleaned_roots_len, roots_range) = {
-                let mut w_roots_tracker = self.roots_tracker.write().unwrap();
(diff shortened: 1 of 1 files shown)


jeffwashington:
validator against mnb after 33 clean cycles:
unrooted_cleaned_count: 54
rooted_cleaned_count: 4973
99% of slots are roots

runtime/src/accounts_index.rs
```diff
@@ -1235,30 +1235,28 @@ impl<T: 'static + Clone + IsCached + ZeroLamport> AccountsIndex<T> {
...
-            })
-        } else {
-            None
-        }
+        let (roots_len, uncleaned_roots_len, previous_uncleaned_roots_len, roots_range) = {
+            let mut w_roots_tracker = self.roots_tracker.write().unwrap();
+            if !w_roots_tracker.roots.remove(&slot) {
+                return None;
+            }
+            w_roots_tracker.uncleaned_roots.remove(&slot);
```
jeffwashington, line 1243: should we go ahead and remove this slot in case it does exist in these?

carllin (approved):

runtime/src/accounts_index.rs
carllin: I think that integrity check is not a bad idea, to collect metrics to see if this is actually happening

jeffwashington: will add a metric separately

PR #16830, opened 2021-04-26
stretchy roots tracker

jeffwashington:
#### Problem
I've been chasing a problem where really old (-2m slots) 0-lamport accounts are not getting cleaned. There is at least 1 validator on mnb that is generating snapshots like this. My research indicates that the problem is that we are cleaning, but after removing a store, we aren't updating the accounts index to deref the accounts that are present in the store.

We can load a snapshot with this data:
roots: 460977, min: 71500402, max: 73790273, range: 2289871
the first non-zero lamport account is:
slot: 73358273, previous # of ZERO lamport accounts: 156880

This panics because a range of 2M is too large for the previously hardcoded range.
There is a problem apparently where old accounts are not being cleaned correctly. I'm working on that separately.

This PR will prevent us from asserting and instead we will silently handle the behavior with some minor performance penalty. The previous implementation of RootsTracker WAS a HashSet.

#### Summary of Changes
Adapt the RootsTracker to allow it to stretch to accommodate really old roots while maintaining high performance lookups on the most recent set of roots. Code inspection seems to indicate minor performance impact. We call is_root many times during account load and many other operations.

Fixes #

### DIFF
--- a/runtime/src/accounts_index.rs
+++ b/runtime/src/accounts_index.rs
@@ -1675,16 +1675,16 @@ pub mod tests {
         for width in [16, 4194304].iter() {
             let width = *width;
             let mut tester = setup_empty(width);
-            for start in [0, width * 5].iter().map(|x| *x) {
+            for start in [0, width * 5].iter().cloned() {
                 // recreate means create empty bitfield with each iteration, otherwise re-use
-                for recreate in [false, true].iter().map(|x| *x) {
+                for recreate in [false, true].iter().cloned() {
                     let max = start + 3;
                     // first root to add
                     for slot in start..max {
                         // subsequent roots to add
                         for slot2 in (slot + 1)..max {
                             // reverse_slots = 1 means add slots in reverse order (max to min). This causes us to add second and later slots to excess.
-                            for reverse_slots in [false, true].iter().map(|x| *x) {
+                            for reverse_slots in [false, true].iter().cloned() {
                                 let maybe_reverse = |slot| {
                                     if reverse_slots {
                                         max - slot
--- a/runtime/src/accounts_index.rs
+++ b/runtime/src/accounts_index.rs
@@ -187,6 +187,10 @@ pub struct RollingBitField {
     max: u64, // exclusive
     bits: BitVec,
     count: usize,
+    // These are items that are true and lower than min.
+    // They would cause us to exceed max_width if we stored them in our bit field.
+    // We only expect these items in conditions where there is some other bug in the system.
+    excess: HashSet<u64>,
 }
 // functionally similar to a hashset
 // Relies on there being a sliding window of key values. The key values continue to increase.
@@ -202,6 +206,7 @@ impl RollingBitField {
             count: 0,
             min: 0,
             max: 0,
+            excess: HashSet::new(),
         }
     }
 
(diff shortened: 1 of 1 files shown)


jeffwashington:
logging:
#16636

jeffwashington:
I plan to add some more tests. Before doing so, I thought I would make sure this seems like a reasonable idea.

sakridge:
> I plan to add some more tests. Before doing so, I thought I would make sure this seems like a reasonable idea.

Seems reasonable to me.

runtime/src/accounts_index.rs
```diff
@@ -1465,9 +1510,14 @@ pub mod tests {
...
+    fn setup_empty(width: u64) -> (RollingBitField, HashSet<u64>) {
+        let bitfield = RollingBitField::new(width);
+        let hash = HashSet::new();
+        (bitfield, hash)
+    }
+
     fn setup_wide(width: u64, start: u64) -> (RollingBitField, HashSet<u64>) {
-        let mut bitfield = RollingBitField::new(width);
-        let mut hash = HashSet::new();
+        let (mut bitfield, mut hash) = setup_empty(width);
```
sakridge, line 1520: Can we call it `hash_set`. `hash` kind of confusing with a sha2 hash value.

runtime/src/accounts_index.rs
```diff
@@ -1562,7 +1621,142 @@ pub mod tests {
...
+        for start in [0, width * 5].iter() {
+            let start = *start;
+            // recreate = 1 means create empty bitfield with each iteration, otherwise re-use
+            for recreate in 0..2 {
+                let max = start + 3;
+                // first root to add
+                for slot in start..max {
+                    // subsequent roots to add
+                    for slot2 in (slot + 1)..max {
+                        // revers_slots = 1 means add slots in reverse order (max to min). This causes us to add second and later slots to excess.
```
sakridge, line 1690: reverse_slots here in the comment

runtime/src/accounts_index.rs
```diff
@@ -1562,7 +1621,142 @@ pub mod tests {
     }
 
     #[test]
-    #[should_panic(expected = "out of range")]
+    fn test_bitfield_excess2() {
+        solana_logger::setup();
+        let width = 16;
```
sakridge, line 1626: maybe at least one test where `width != 16`

runtime/src/accounts_index.rs
```diff
@@ -1574,24 +1768,33 @@ pub mod tests {
...
-            max = std::cmp::max(max, item);
-            assert!(hashset.contains(&item));
-        }
-        assert!(
-            bitfield.range_width()
-                >= if bitfield.is_empty() {
+        if !bitfield.is_empty() {
+            let mut min = Slot::MAX;
+            let mut max = Slot::MIN;
+            for item in bitfield.get_all() {
```
sakridge, line 1774: does it make sense to also assert this: `assert_eq!(bitfield.get_all().len(), hashset.len())`

runtime/src/accounts_index.rs
```diff
@@ -1562,7 +1621,142 @@ pub mod tests {
...
 
     #[test]
-    #[should_panic(expected = "out of range")]
+    fn test_bitfield_excess2() {
+        solana_logger::setup();
+        let width = 16;
+        let (mut bitfield, mut hash) = setup_empty(width);
+        let slot = 100;
+        // insert 1st slot
+        bitfield.insert(slot);
```
sakridge, line 1630: Seems like the tests could be simplified with a kind of wrapper object with both bitfield and hashset where it has `insert` and `compare` functions. `compare` can test that bitfield.contains is tested with what exists in the reference hashset.

runtime/src/accounts_index.rs
```diff
@@ -1562,7 +1621,142 @@ pub mod tests {
...
+    #[test]
+    fn test_bitfield_excess() {
+        solana_logger::setup();
+        let width = 16;
+        let (mut bitfield, mut hash) = setup_empty(width);
+        // start at slot 0 or a separate, higher slot
+        for start in [0, width * 5].iter() {
+            let start = *start;
+            // recreate = 1 means create empty bitfield with each iteration, otherwise re-use
+            for recreate in 0..2 {
```
sakridge, line 1684: I think this works: `for recreate in &[false, true] {`

sakridge (approved):
lgtm

jeffwashington:
ran validator against mnb. Seems healthy. Roots range is back to normal 400k, though.
Some metrics:
[2021-04-29T13:58:31.278072015Z INFO  solana_metrics::metrics] datapoint: accounts_index_roots_len roots_len=400696i uncleaned_roots_len=1i previous_uncleaned_roots_len=400695i roots_range_width=852590i
[2021-04-29T14:01:01.641614676Z INFO  solana_metrics::metrics] datapoint: accounts_index_roots_len roots_len=400675i uncleaned_roots_len=1i previous_uncleaned_roots_len=0i roots_range_width=432001i
[2021-04-29T14:02:29.072682319Z INFO  solana_metrics::metrics] datapoint: accounts_index_roots_len roots_len=400830i uncleaned_roots_len=1i previous_uncleaned_roots_len=100i roots_range_width=432170i
[2021-04-29T14:03:08.340413551Z INFO  solana_metrics::metrics] datapoint: accounts_index_roots_len roots_len=400898i uncleaned_roots_len=1i previous_uncleaned_roots_len=100i roots_range_width=432257i
[2021-04-29T14:05:03.109012997Z INFO  solana_metrics::metrics] datapoint: accounts_index_roots_len roots_len=400901i uncleaned_roots_len=1i previous_uncleaned_roots_len=500i roots_range_width=432243i

PR #16831, opened 2021-04-26
rootstracker.remove returns previous state

jeffwashington:
#### Problem
RootsTracker was designed to replace HashSet. HashSet.remove() returns true if the item was removed.
This is needed for future changes that are coming.
#### Summary of Changes
RootsTracker.remove() returns true if the item was removed.
Fixes #


taozhu-chicago:
Changes look good

steviez (approved):
Looks good to me!

PR #16838, opened 2021-04-26
reclaim unrefs accounts from index

jeffwashington:
#### Problem
I've been chasing a problem where really old (-2m slots) 0-lamport accounts are not getting cleaned. There is at least 1 validator on mnb that is generating snapshots like this. My research indicates that the problem is that we are cleaning, but after removing a store, we aren't updating the accounts index to deref the accounts that are present in the store.

Without this change, we load a snapshot with this data:
roots: 460977, min: 71500402, max: 73790273, range: 2289871
the first non-zero lamport account is:
slot: 73358273, previous # of ZERO lamport accounts: 156880

after 1st clean, we end up with:
roots: 460977, min: 71500402, max: 73790273, range: 2,289,871 < in snapshot, before clean
roots: 391303, min: 71999188, max: 73790273, range: 1,791,085 < after 1st clean

subsequent cleans do nothing to remove additional accounts.

WITH this change, the following progression is observed:
roots: 460977, min: 71500402, max: 73790273, range: 2,289,871 < in snapshot, before clean
roots: 391303, min: 71999188, max: 73790273, range: 1,791,085 < after 1st clean
roots: 387969, min: 73358273, max: 73790273, range: 432,000 < after 2nd clean (no further changes after this)

I believe this is now expected behavior for range of active slots.

It is very possible there is a better way to fix this problem. I am unsure yet if reclaim_result needs to be used for further processing. This finding seemed to be a good stopping point to circle back and discuss the demonstrated problem and a potential 'solution'.

#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -1724,7 +1724,9 @@ impl AccountsDb {
         // Don't reset from clean, since the pubkeys in those stores may need to be unref'ed
         // and those stores may be used for background hashing.
         let reset_accounts = false;
-        self.handle_reclaims(&reclaims, None, false, None, reset_accounts);
+        let mut reclaim_result = ReclaimResult::default();
+        let reclaim_result = Some(&mut reclaim_result);
+        self.handle_reclaims(&reclaims, None, false, reclaim_result, reset_accounts);
 
         reclaims_time.stop();
 


jeffwashington:
logging:
#16636

jeffwashington:
For discussion with:
@brooksprumo , @sakridge , @carllin 

I need help knowing whether this is a reasonable way to get the desired behavior without introducing some other negative side effect, race condition, etc.

jeffwashington:
doh - my initial creation of this commit had the modifications on the wrong call to handle_reclaims. That led @carllin and me to be speaking different languages!

runtime/src/accounts_db.rs
```diff
@@ -1724,7 +1724,9 @@ impl AccountsDb {
         // Don't reset from clean, since the pubkeys in those stores may need to be unref'ed
         // and those stores may be used for background hashing.
         let reset_accounts = false;
-        self.handle_reclaims(&reclaims, None, false, None, reset_accounts);
+        let mut reclaim_result = ReclaimResult::default();
+        let reclaim_result = Some(&mut reclaim_result);
+        self.handle_reclaims(&reclaims, None, false, reclaim_result, reset_accounts);
```
carllin, line 1729: ah shoot, this is because the final `finalize_dead_slot_removal` only unrefs the pubkey if `purged_stored_account_slot.is_some()`. This seems like the likely bug.

Looks like I introduced it here: https://github.com/solana-labs/solana/pull/15534/files#diff-1090394420d51617f3233275c2b65ed706b35b53b115fe65f82c682af8134a6fR3958, ugh sorry!

jeffwashington:
Sounds good. I gave it a some so that code would run. ;)

runtime/src/accounts_db.rs
carllin: A good thing to enforce in `handle_reclaims()` might be that anytime that `no_dead_slot == false,`, then `reclaim_result.is_some()` should also be true, because if a slot can be dead, then unrefing must be able to happen.

sakridge: Yea, this seems right. Good catch @jeffwashington

jeffwashington: #15534 is the introduction of this problem.
Unless I'm mistaken, I see this commit in 1.5 and 1.6.
@mvines @sakridge  what do you recommend?

carllin:
@jeffwashington can we also add or expand one of the existing tests to cover this case?

jeffwashington:
@carllin will help me add tests. we may consider moving this to 1.6/1.5

PR #16840, opened 2021-04-26
Update bpf loader info on native-programs docs

CriesofCarrots:
#### Problem
Native-program docs refer to BPF deprecated loader

#### Summary of Changes
Update to upgradeable loader


jackcmay (dismissed):

jackcmay:
> @jackcmay , I did a super simple update here. Let me know if you think we should add more detail about upgrading and such in this area, or link to somewhere else with more upgradeable loader details.

Yeah, a  link to the program deployment docs would be helpful

PR #16849, opened 2021-04-26
stats on clean for removing roots

jeffwashington:
#### Problem
on clean_dead_slot, we get a read lock to check for root, then we get a write lock to remove the root. If we are a high% of the time finding that the item is a root, then we can improve performance and reduce lock contention by just getting the write lock and [removing the root the first time](https://github.com/solana-labs/solana/pull/16829).

#### Summary of Changes
log metrics on how often we try to clean_dead_slot on a slot that isn't a root.
Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -951,8 +951,8 @@ struct LatestAccountsIndexRootsStats {
     uncleaned_roots_len: AtomicUsize,
     previous_uncleaned_roots_len: AtomicUsize,
     roots_range: AtomicU64,
-    cleaned_roots: AtomicUsize,
-    not_a_root_cleaned: AtomicUsize,
+    rooted_cleaned_count: AtomicUsize,
+    unrooted_cleaned_count: AtomicUsize,
 }
 
 impl LatestAccountsIndexRootsStats {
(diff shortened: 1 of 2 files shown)


runtime/src/accounts_db.rs
```diff
@@ -4402,17 +4420,25 @@ impl AccountsDb {
...
                     accounts_index_root_stats = latest;
+                } else {
+                    not_a_root_count += 1;
                 }
                 *slot
             })
             .collect();
         info!("finalize_dead_slot_removal: slots {:?}", dead_slots);
 
+        accounts_index_root_stats.cleaned_roots += dead_slots_count;
```
carllin, line 4439: hmm this is called `cleaned_roots`, but `dead_slots_count` also includes unrooted slots, it's incremented outside of the `if let Some(latest) = self.accounts_index.clean_dead_slot(*slot) {` check above. 

How about either:

1) Renaming `cleaned_roots` to `cleaned_slots_count`, or
2)  If we actually want to track the `cleaned_roots`, renaming `dead_slots_count -> dead_roots_count`, and moving the increment inside the `if` check

runtime/src/accounts_db.rs
```diff
@@ -951,6 +951,8 @@ struct LatestAccountsIndexRootsStats {
     uncleaned_roots_len: AtomicUsize,
     previous_uncleaned_roots_len: AtomicUsize,
     roots_range: AtomicU64,
+    cleaned_roots: AtomicUsize,
+    not_a_root_cleaned: AtomicUsize,
```
carllin, line 955: how about and `rooted_cleaned_count` and `unrooted_cleaned_count` to communicate we're specifying a tally of the total cleans

carllin (dismissed):

jeffwashington:
first metrics, on snapshot load:
datapoint: accounts_index_roots_len roots_len=391303i uncleaned_roots_len=1i previous_uncleaned_roots_len=391302i roots_range_width=1791086i unrooted_cleaned_count=0i rooted_cleaned_count=69674i
datapoint: accounts_index_roots_len roots_len=387969i uncleaned_roots_len=1i previous_uncleaned_roots_len=0i roots_range_width=432001i unrooted_cleaned_count=0i rooted_cleaned_count=3334i

PR #16855, opened 2021-04-27
Add getVoteAccounts RPC method parameter to restrict results to a single vote account

mvines:
Sometimes users only care about a single validator vote account, which can now be supplied as an input to the `getVoteAccounts` RPC method

### DIFF
--- a/client/src/rpc_client.rs
+++ b/client/src/rpc_client.rs
@@ -496,10 +496,17 @@ impl RpcClient {
         &self,
         commitment_config: CommitmentConfig,
     ) -> ClientResult<RpcVoteAccountStatus> {
-        self.send(
-            RpcRequest::GetVoteAccounts,
-            json!([self.maybe_map_commitment(commitment_config)?]),
-        )
+        self.get_vote_accounts_with_config(RpcGetVoteAccountsConfig {
+            commitment: Some(self.maybe_map_commitment(commitment_config)?),
+            ..RpcGetVoteAccountsConfig::default()
+        })
+    }
+
+    pub fn get_vote_accounts_with_config(
+        &self,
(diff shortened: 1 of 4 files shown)


client/src/rpc_config.rs
```diff
@@ -49,6 +49,14 @@ pub struct RpcLeaderScheduleConfig {
     pub commitment: Option<CommitmentConfig>,
 }
 
+#[derive(Debug, Default, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct RpcGetVoteAccountsConfig {
+    pub vote_pubkey: Option<String>, // validator vote address, as a base-58 encoded string
```
mvines, line 55: Breaks my heart a little to use `vote_pubkey` instead of `vote_address`, but I wanted to be consistent with the returned object.

I also only added support to filter by vote account address instead of validator identity because this method is about getting vote accounts.

CriesofCarrots: 😢 yeah. I suppose we could deprecate the `x_pubkey` fields and add `x_address` fields instead

client/src/rpc_client.rs
```diff
@@ -884,15 +891,12 @@ impl RpcClient {
...
-        self.send(
-            RpcRequest::GetLeaderSchedule,
-            json!([
-                slot,
-                RpcLeaderScheduleConfig {
-                    commitment: Some(self.maybe_map_commitment(commitment_config)?),
-                    ..RpcLeaderScheduleConfig::default()
-                }
-            ]),
+        self.get_leader_schedule_with_config(
```
CriesofCarrots, line 894: Thanks

CriesofCarrots (approved):
lgtm!

PR #16872, opened 2021-04-27
explorer: add Bonfida Bots instructions

dr497:
#### Summary of Changes

Fixes #

### DIFF
--- a/explorer/src/components/instruction/bonfida-bot/types.ts
+++ b/explorer/src/components/instruction/bonfida-bot/types.ts
@@ -1,0 +1,429 @@
+/* eslint-disable @typescript-eslint/no-redeclare */
+import { PublicKey, TransactionInstruction } from "@solana/web3.js";
+import {
+  enums,
+  number,
+  type,
+  Infer,
+  create,
+  array,
+  string,
+  optional,
+} from "superstruct";
+import {
+  Instruction,
+  decodeInstruction,
+  BONFIDABOT_PROGRAM_ID,
+} from "bonfida-bot";
+
+export const SERUM_DECODED_MAX = 6;
+
+export type Side = Infer<typeof Side>;
+export const Side = enums([0, 1]);
+
+export type OrderType = Infer<typeof OrderType>;
+export const OrderType = enums([0, 1, 2]);
+
+export type InitializeBot = {
+  systemProgramAccount: PublicKey;
+  rentSysvarAccount: PublicKey;
+  splTokenProgramAccount: PublicKey;
+  poolAccount: PublicKey;
+  mintAccount: PublicKey;
+  payerAccount: PublicKey;
+
+  poolSeed: string;
+  maxNumberOfAsset: number | undefined | null;
+  numberOfMarkets: number;
+  programId: PublicKey;
+};
+
+export const InitializeBotDecode = type({
+  poolSeed: string(),
+  maxNumberOfAsset: optional(number()),
+  numberOfMarkets: number(),
+});
+
+export const decodeInitializeBot = (
+  ix: TransactionInstruction
+): InitializeBot => {
+  const decoded = create(
+    decodeInstruction(ix.data, Instruction.Init),
+    InitializeBotDecode
+  );
+  const initializeBot: InitializeBot = {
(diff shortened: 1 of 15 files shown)


dr497:
List of example tx for each instruction:

- Initialize Bot:  

`xLXBE2nWTpfXgHQgWRAxuDD1bep7d8EGqugJZW776Lj5RSRi1R2Pa64aTjfBZ7rXGMLUdouojDajVZiyUn6d2mT`

- Create Bot:

`5Byxx2WH6dEBDVcSqJGpoMgNEEtwmv1qBsSdVstu5DJPJQvNBg9YztcxdRucKPFRLADS2ensRDLcYmy2i5K4QPZi`

- Deposit:

`32WXim2F58ibZUbBoHZH9ezsieGENTeWKi52TZYUFQ8BMhwcaeDshxLHrngxrsLzcZ8kVPESbHUbakXYAARvhx9g`

- Create Order:

`2FYLUab7XwDnANCAmri6tUn32VHprR3XxhXLDWhVRYjcLRSWWoo51nXjuUjRYVRmeETR27omKvxGdzSReh7wBanA`

- Cancel Order: Not in use currently, but will be in the future so no tx id for the moment.

- Settle Funds:

`LMig4jdMipdZvg4XtP3gbaHpSTbqShcDFL2ZDpiWfWcD2foU8QpBALN23dw1KSsWufZgoppvxM9QKJszrQv8c3e` ?

- Redeem:

`4YezJkmDRH8iA3LrLEx7qBfA66X2BHX7BAAQf4hvLTXinKU9Bpi3eZUVYj85cruMtbL31yV31i1SZdXhMJEAWkoU`

- Collect Fees:

`5eELZ9NvBKEDLLBCyCNdEDpsfyVMNipuikinUmniDHSt2AreDWq8XeeZg22Thi9qQbYnz5ck5hnpGju81sMGt53a`

explorer/src/components/instruction/bonfida-bot/types.ts
```diff
@@ -0,0 +1,429 @@
...
+  feeRatio: number;
+  programId: PublicKey;
+};
+
+export const CreateBotDecode = type({
+  poolSeed: string(),
+  feeCollectionPeriod: number(),
+  feeRatio: number(),
+  depositAmounts: array(number()),
+  markets: array(string()),
```
oJshua, line 91: This `CreateBotDecode` decoder is failing on the markets array. Instead of an array of strings being returned from `decodeInstruction(ix.data, Instruction.Create)`, an array of `PublicKey`s are being returned.

One thing you could do here is write a custom `PublicKey` coercer that returns a string:
```
const PublicKeyToString = coerce(
  string(),
  any(), // usually this would be instance(PublicKey), but there appears to be a compatibility issue between Explorer web3 and bonfida-bot web3 PublicKeys
  value => value.toBase58()
);
```

Then your decoder would look like:
```
export const CreateBotDecode = type({
  poolSeed: string(),
  feeCollectionPeriod: number(),
  feeRatio: number(),
  depositAmounts: array(number()),
  markets: array(PublicKeyToString),
});
```

explorer/src/components/instruction/bonfida-bot/types.ts
```diff
@@ -0,0 +1,429 @@
...
+  splTokenProgramId: PublicKey;
+  dexProgramKey: PublicKey;
+  srmReferrerKey: PublicKey | null;
+  poolSeed: string;
+};
+
+export const SettleFundsDecode = type({
+  poolSeed: string(),
+  pcIndex: number(),
+  orderId: number(),
```
oJshua, line 302: `orderId` appears to be coming back `undefined` here, causing the Settle cards to render as a raw instruction:

settlecardraw

oJshua (dismissed):
This looks really good -- just a couple things needing fixing! Mostly, it appears some of the superstruct decoders are failing, which cause the associated cards to render as Raw instructions. I believe it's just Create Bot and Settle, but I will make another pass tomorrow.

dr497:
Should be good now!

oJshua (approved):
Thank you for the changes! Looks great.

PR #16690, opened 2021-04-20
Readable account trait for stored account meta

jeffwashington:
#### Problem
Remove copies of account and its data during a few code paths.
#### Summary of Changes

Fixes #

PR #16874, opened 2021-04-27
write Option<AccountSharedData>

jeffwashington:
#### Problem
Working towards abstracting AccountSharedData to other implementations. Moving callers to use Readable/WritableAccount methods. There will [soon be a new ReadableAccount impl for a stored account](https://github.com/solana-labs/solana/pull/16690/commits/95261a4f74b5aed97e96cf11a0f9b791565a3c23#r621561972). This does not have a good default. So, AccountSharedData::default() doesn't work well.

#### Summary of Changes
Make the write path that uses AccountSharedData::default() instead use Option<AccountSharedData>

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -3749,6 +3749,9 @@ impl AccountsDb {
             .iter()
             .map(|(pubkey, account)| {
                 self.read_only_accounts_cache.remove(pubkey, slot);
+                // this is the source of Some(Account) or None.
+                // Some(Account) = store 'Account'
+                // None = store a default/empty account with 0 lamports
                 let (account, data_len) = if account.lamports() == 0 {
                     (None, 0)
                 } else {
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -3310,7 +3310,7 @@ impl AccountsDb {
         slot: Slot,
         hashes: &[impl Borrow<Hash>],
         mut storage_finder: F,
-        accounts_and_meta_to_store: &[(StoredMeta, &AccountSharedData)],
+        accounts_and_meta_to_store: &[(StoredMeta, Option<&AccountSharedData>)],
     ) -> Vec<AccountInfo> {
         assert_eq!(hashes.len(), accounts_and_meta_to_store.len());
         let mut infos: Vec<AccountInfo> = Vec::with_capacity(accounts_and_meta_to_store.len());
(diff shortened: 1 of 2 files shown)


runtime/src/accounts_db.rs
```diff
@@ -3310,15 +3310,18 @@ impl AccountsDb {
...
+        accounts_and_meta_to_store: &[(StoredMeta, Option<&AccountSharedData>)],
     ) -> Vec<AccountInfo> {
         assert_eq!(hashes.len(), accounts_and_meta_to_store.len());
         let mut infos: Vec<AccountInfo> = Vec::with_capacity(accounts_and_meta_to_store.len());
         let mut total_append_accounts_us = 0;
         let mut total_storage_find_us = 0;
         while infos.len() < accounts_and_meta_to_store.len() {
             let mut storage_find = Measure::start("storage_finder");
-            let data_len = accounts_and_meta_to_store[infos.len()].1.data().len();
+            let data_len = accounts_and_meta_to_store[infos.len()]
```
carllin, line 3321: if `data_len` is zero, does that mean this account wasn't touched at all? Can we early exit to avoid the storage lookup overhead below?

runtime/src/accounts_db.rs
```diff
@@ -3310,15 +3310,18 @@ impl AccountsDb {
         slot: Slot,
         hashes: &[impl Borrow<Hash>],
         mut storage_finder: F,
-        accounts_and_meta_to_store: &[(StoredMeta, &AccountSharedData)],
+        accounts_and_meta_to_store: &[(StoredMeta, Option<&AccountSharedData>)],
```
carllin, line 3313: What do you think about an enum `CommitAccountSharedData` that has two variants, `Dirty` and `Unmodified`?

runtime/src/accounts_db.rs
jeffwashington: I'm not sure. Looking up the call chain, we decide to store a default AccountSharedData if lamports are 0. This code was pre-existing to my change. I don't know if we allow non-zero lamport, valid accounts with data.len==0. I imagine it is an optimization in the code up the call chain to pass a default account so that we don't store the data[] with a zero-lamport account?

runtime/src/accounts_db.rs
jeffwashington: I think the 2 options here mean:
1. zero lamport account intended only to hide earlier zero lamport accounts (None)
2. every other account we want to write to storage or cache

runtime/src/accounts_db.rs
```diff
@@ -3738,17 +3745,18 @@ impl AccountsDb {
...
-        let default_account = AccountSharedData::default();
-        let accounts_and_meta_to_store: Vec<(StoredMeta, &AccountSharedData)> = accounts
+        let accounts_and_meta_to_store: Vec<(StoredMeta, Option<&AccountSharedData>)> = accounts
             .iter()
             .map(|(pubkey, account)| {
                 self.read_only_accounts_cache.remove(pubkey, slot);
-                let account = if account.lamports() == 0 {
-                    &default_account
+                // this is the source of Some(Account) or None.
+                // Some(Account) = store 'Account'
```
jeffwashington, line 3753: @carllin comment describing some/none at the introduction of None

carllin (approved):

steviez (approved):
Ship It!

PR #16879, opened 2021-04-27
Refactor `collect_uncleaned_pubkeys_to_slot()`

brooksprumo:
While working on another issue (#16580), I came across
`collect_unclean_pubkeys_to_slot()` and had difficulty understanding it.
Since the function does a few logically separate things, I split the
function up.  I also added documentation, removed an usused return value,
and renamed the functions to be more specific.

This commit is to split up an existing PR (#16786), where I had both this
aesthetic change _and_ a behavioral change.

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -1464,39 +1464,52 @@ impl AccountsDb {
         }
     }
 
-    fn collect_uncleaned_pubkeys_to_slot(&self, max_slot: Slot) -> (Vec<Vec<Pubkey>>, Slot) {
-        let mut max_slot_in_uncleaned_pubkeys = 0;
-        let slots: Vec<Slot> = self
-            .uncleaned_pubkeys
+    /// Collect all the uncleaned slots, up to a max slot
+    ///
+    /// Search through the uncleaned Pubkeys and return all the slots, up to a maximum slot.
+    fn collect_uncleaned_slots_up_to_slot(&self, max_slot: Slot) -> Vec<Slot> {
+        self.uncleaned_pubkeys
             .iter()
             .filter_map(|entry| {
-                let slot = entry.key();
-                max_slot_in_uncleaned_pubkeys = max_slot_in_uncleaned_pubkeys.max(*slot);
-                if *slot <= max_slot {
-                    Some(*slot)
+                let slot = *entry.key();
+                (slot <= max_slot).then(|| slot)
+            })
+            .collect()
+    }
+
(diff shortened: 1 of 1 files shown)


runtime/src/accounts_db.rs
```diff
@@ -1515,7 +1528,7 @@ impl AccountsDb {
 
         let mut collect_delta_keys = Measure::start("key_create");
         let max_slot = max_clean_root.unwrap_or_else(|| self.accounts_index.max_root());
-        let (delta_keys, _max_slot) = self.collect_uncleaned_pubkeys_to_slot(max_slot);
```
brooksprumo, line 1518: For more discussion about removing the max_slot return value, refer to https://github.com/solana-labs/solana/pull/16786#pullrequestreview-643591378

runtime/src/accounts_db.rs
```diff
@@ -1464,39 +1464,52 @@ impl AccountsDb {
...
+    fn collect_uncleaned_slots_up_to_slot(&self, max_slot: Slot) -> Vec<Slot> {
+        self.uncleaned_pubkeys
             .iter()
             .filter_map(|entry| {
-                let slot = entry.key();
-                max_slot_in_uncleaned_pubkeys = max_slot_in_uncleaned_pubkeys.max(*slot);
-                if *slot <= max_slot {
-                    Some(*slot)
+                let slot = *entry.key();
+                (slot <= max_slot).then(|| slot)
```
jeffwashington, line 1475: maybe: .then(|| *entry.key()) and delete line 1474

jeffwashington:
I suggest you add a simple test that at least calls each of these functions. Someone may come along later and refactor again. Would be nice to know things still at least basically work in that situation.

runtime/src/accounts_db.rs
brooksprumo: I think I could do
```
(entry.key() <= max_slot).then(|| *entry.key())
```
to get it onto a single line. I thought it was a bit easier to read with the `let slot = ...` so that (1) I didn't have to duplicate `entry.key()`, and (2) so that it is clear what `entry.key()` is. Thoughts?

jeffwashington: yep, I agree. I didn't realize it was used twice.

brooksprumo:
> I suggest you add a simple test that at least calls each of these functions. Someone may come along later and refactor again. Would be nice to know things still at least basically work in that situation.

@jeffwashington OK, simple tests have been added!

jeffwashington (commented):
lgtm

jeffwashington (approved):

PR #16882, opened 2021-04-27
AccountSharedData.executable becomes private

jeffwashington:
Problem
Working towards abstracting AccountSharedData to other implementations. Moving callers to use Readable/WritableAccount methods.

Summary of Changes
Make executable private to prevent further uses from appearing.
Fixes #

### DIFF
--- a/sdk/src/account.rs
+++ b/sdk/src/account.rs
@@ -37,7 +37,7 @@ pub struct AccountSharedData {
     /// the program that owns this account. If executable, the program that loads this account.
     owner: Pubkey,
     /// this account's data contains a loaded program (and is now read-only)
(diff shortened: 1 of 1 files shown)


sdk/src/account.rs
```diff
@@ -37,7 +37,7 @@ pub struct AccountSharedData {
     /// the program that owns this account. If executable, the program that loads this account.
     owner: Pubkey,
     /// this account's data contains a loaded program (and is now read-only)
-    pub executable: bool,
+    executable: bool,
```
brooksprumo, line 40: 🙌

brooksprumo (approved):

PR #16891, opened 2021-04-27
Add allowed-IP list to faucet

CriesofCarrots:
#### Problem
Follow-up to https://github.com/solana-labs/solana/pull/16487
Airdrop requests sent to the same rpc node appear to the faucet with the same source IP address, regardless of the IP that sent the rpc request. If this IP is not the loopback address (as in the case of load-balanced api nodes remote from the faucet), airdrop requests will be overly limited by a faucet running with a per-time-cap, limiting all users together instead of per IP as intended.

#### Summary of Changes
Add allowed-IP list to faucet; IPs listed are excluded from ip limiting (address limits still checked/applied)

cc @joeaba

### DIFF
--- a/faucet/src/bin/faucet.rs
+++ b/faucet/src/bin/faucet.rs
@@ -1,5 +1,5 @@
 use {
-    clap::{crate_description, crate_name, App, Arg},
+    clap::{crate_description, crate_name, values_t, App, Arg},
     log::*,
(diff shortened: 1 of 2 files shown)


mvines (commented):
lgtm, let's just replace the term whitelist with something like always allowed list.

CriesofCarrots:
@mvines I went with just `allowed_ips`. Look okay?

mvines (approved):

PR #16911, opened 2021-04-28
Clean unrooted dropped banks (#16580)

brooksprumo:
Clean unrooted dropped banks (#16580)

In a scenario where a bank is unrooted and dropped, any keys that exist
_only_ in that bank are now cleaned up.

This work was originally based on PR #15106.

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -1269,8 +1269,8 @@ impl AccountsDb {
             .expect("Cluster type must be set at initialization")
     }
 
-    // Reclaim older states of rooted accounts for AccountsDb bloat mitigation
-    fn clean_purgeable_accounts(
+    /// Reclaim older states of accounts older than max_clean_root for AccountsDb bloat mitigation
+    fn clean_accounts_older_than_root(
         &self,
         purges: Vec<Pubkey>,
         max_clean_root: Option<Slot>,
@@ -1642,7 +1642,7 @@ impl AccountsDb {
         let mut clean_old_rooted = Measure::start("clean_old_roots");
         let purges = [purges_rooted, purges_unrooted].concat();
         let (purged_account_slots, removed_accounts) =
-            self.clean_purgeable_accounts(purges, max_clean_root);
+            self.clean_accounts_older_than_root(purges, max_clean_root);
 
         if self.caching_enabled {
             self.do_reset_uncleaned_roots(max_clean_root);
@@ -1671,7 +1671,7 @@ impl AccountsDb {
                     return false;
                 }
                 // Check if this update in `slot` to the account with `key` was reclaimed earlier by
-                // `clean_purgeable_accounts()`
+                // `clean_accounts_older_than_root()`
                 let was_reclaimed = removed_accounts
                     .get(&account_info.store_id)
                     .map(|store_removed| store_removed.contains(&account_info.offset))
@@ -2501,7 +2501,7 @@ impl AccountsDb {
         //        purge_slot_cache_pubkeys()      | (removes existing store_id, offset for caches)
         //      OR                                |
         //    clean_accounts()/                   |
-        //        clean_purgeable_accounts()      | (removes existing store_id, offset for stores)
+        //        clean_accounts_older_than_root()| (removes existing store_id, offset for stores)
         //                                        V
         //
         // Remarks for purger: So, for any reading operations, it's a race condition
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -1564,14 +1564,13 @@ impl AccountsDb {
         let total_keys_count = pubkeys.len();
         let mut accounts_scan = Measure::start("accounts_scan");
         // parallel scan the index.
-        let (mut purges_zero_lamports, purges_rooted, purges_unrooted) = {
+        let (mut purges_zero_lamports, purges_old_accounts) = {
             self.thread_pool_clean.install(|| {
                 pubkeys
                     .par_chunks(4096)
                     .map(|pubkeys: &[Pubkey]| {
                         let mut purges_zero_lamports = HashMap::new();
-                        let mut purges_rooted = Vec::new();
-                        let mut purges_unrooted = Vec::new();
+                        let mut purges_old_accounts = Vec::new();
                         for pubkey in pubkeys {
                             match self.accounts_index.get(pubkey, None, max_clean_root) {
                                 AccountIndexGetResult::Found(locked_entry, index) => {
@@ -1604,7 +1603,7 @@ impl AccountsDb {
                                         if let Some(max_clean_root) = max_clean_root {
                                             assert!(slot <= max_clean_root);
                                         }
-                                        purges_rooted.push(*pubkey);
+                                        purges_old_accounts.push(*pubkey);
                                     }
                                 }
                                 AccountIndexGetResult::NotFoundOnFork => {
@@ -1614,7 +1613,7 @@ impl AccountsDb {
                                     // Also, this pubkey must have been touched by some slot since
                                     // it was in the dirty list, so we assume that the slot it was
                                     // touched in must be unrooted.
-                                    purges_unrooted.push(*pubkey);
+                                    purges_old_accounts.push(*pubkey);
                                 }
                                 AccountIndexGetResult::Missing(lock) => {
                                     // pubkey is missing from index, so remove from zero_lamports_list
@@ -1623,15 +1622,14 @@ impl AccountsDb {
                                 }
                             };
                         }
-                        (purges_zero_lamports, purges_rooted, purges_unrooted)
+                        (purges_zero_lamports, purges_old_accounts)
                     })
                     .reduce(
-                        || (HashMap::new(), Vec::new(), Vec::new()),
+                        || (HashMap::new(), Vec::new()),
                         |mut m1, m2| {
                             // Collapse down the hashmaps/vecs into one.
                             m1.0.extend(m2.0);
                             m1.1.extend(m2.1);
-                            m1.2.extend(m2.2);
                             m1
                         },
                     )
@@ -1640,9 +1638,8 @@ impl AccountsDb {
         accounts_scan.stop();
 
         let mut clean_old_rooted = Measure::start("clean_old_roots");
-        let purges = [purges_rooted, purges_unrooted].concat();
         let (purged_account_slots, removed_accounts) =
-            self.clean_accounts_older_than_root(purges, max_clean_root);
+            self.clean_accounts_older_than_root(purges_old_accounts, max_clean_root);
 
         if self.caching_enabled {
             self.do_reset_uncleaned_roots(max_clean_root);
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -1565,20 +1565,21 @@ impl AccountsDb {
         let total_keys_count = pubkeys.len();
         let mut accounts_scan = Measure::start("accounts_scan");
         // parallel scan the index.
-        let (mut purges, purges_in_root) = {
+        let (mut purges_zero_lamports, purges_in_root, purges_unrooted) = {
             self.thread_pool_clean.install(|| {
                 pubkeys
                     .par_chunks(4096)
                     .map(|pubkeys: &[Pubkey]| {
+                        let mut purges_zero_lamports = HashMap::new();
                         let mut purges_in_root = Vec::new();
-                        let mut purges = HashMap::new();
+                        let mut purges_unrooted = Vec::new();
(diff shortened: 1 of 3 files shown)


runtime/src/bank.rs
```diff
@@ -12468,4 +12468,54 @@ pub(crate) mod tests {
             ))
         );
     }
+
+    #[test]
+    fn test_clean_unrooted_dropped_banks() {
```
carllin, line 12473: Note for other reviewers that there is also a special case for when the unrooted bank is also not frozen, in which case the slot pubkey diff will not be added to the dirty key list for clean. 

After some discussion, the decision was there will be an incoming fix in `Bank::drop()` to handle this case

brooksprumo: Yes! That fix will be in the next PR.

sakridge (dismissed):
lgtm

runtime/src/accounts_index.rs
```diff
@@ -1161,18 +1161,18 @@ impl<T: 'static + Clone + IsCached + ZeroLamport> AccountsIndex<T> {
         account_indexes: &HashSet<AccountIndex>,
     ) {
         let roots_tracker = &self.roots_tracker.read().unwrap();
-        let max_root = Self::get_max_root(&roots_tracker.roots, &list, max_clean_root);
+        let max_present_root = Self::get_max_root(&roots_tracker.roots, &list, max_clean_root);
```
carllin, line 1164: nits: can we rename `max_present_root` to `max_root_in_slot_list`, and `get_max_root()` to `get_max_root_in_slot_list()` to distinguish these are different than the absolute max root?

runtime/src/accounts_index.rs
```diff
@@ -1194,12 +1195,21 @@ impl<T: 'static + Clone + IsCached + ZeroLamport> AccountsIndex<T> {
...
+                if x.slot_list.read().unwrap().is_empty() {
+                    w_maps.remove(pubkey);
+                }
+            }
+        }
     }
 
-    pub fn can_purge(max_root: Slot, slot: Slot) -> bool {
-        slot < max_root
+    pub fn can_purge(max_clean_root: Slot, max_present_root: Slot, slot: Slot) -> bool {
```
carllin, line 1211: nits: I think we can remove the `pub` and also rename `can_purge()` to `can_purge_older_entries()` since it's only specifically used by `purge_older_root_entries()`

runtime/src/accounts_db.rs
```diff
@@ -1608,7 +1609,9 @@ impl AccountsDb {
                                     }
                                 }
                                 AccountIndexGetResult::NotFoundOnFork => {
-                                    // do nothing - pubkey is in index, but not found in a root slot
+                                    // pubkey is in the index but not in a root slot, so clean it
```
carllin, line 1612: An additional clarifying comment is that also this pubkey must have been touched by some slot since it was in the dirty list, so we assume that the slot it was touched in must be unrooted

runtime/src/accounts_index.rs
carllin: Also nit: rename `max_present_root` to `max_cleanable_root_in_slot_list`

runtime/src/accounts_index.rs
```diff
@@ -1194,12 +1195,21 @@ impl<T: 'static + Clone + IsCached + ZeroLamport> AccountsIndex<T> {
...
+        slot < max_clean_root && slot != max_present_root
```
carllin, line 1212: I think for clarifying future readers, I think it would be useful to:

1) Add an assert that `max_cleanable_root_in_slot_list < max_clean_root`
2) Add clarifying comment here  along the lines of:

"If we get a slot update `s  != max_cleanable_root_in_slot_list ` for an account where `s <  max_clean_root`, then we know it's safe to delete because:
a)  If `s < max_cleanable_root_in_slot_list`, then we know the update is outdated by a later rooted update, namely the one in `max_cleanable_root_in_slot_list`
b) If `s > max_cleanable_root_in_slot_list`, then because `s < max_clean_root` and we know there are no roots in the slot list between `max_cleanable_root_in_slot_list` and `max_clean_root`, (otherwise there would be a bigger `max_cleanable_root_in_slot_list`, which is a contradiction), then we know `s` must be an unrooted slot less than `max_clean_root` and thus safe to clean as well.

runtime/src/accounts_index.rs
brooksprumo: I like it! Will do.

runtime/src/accounts_db.rs
```diff
@@ -1633,8 +1637,9 @@ impl AccountsDb {
         accounts_scan.stop();
 
         let mut clean_old_rooted = Measure::start("clean_old_roots");
+        let purges = [purges_in_root, purges_unrooted].concat();
```
carllin, line 1640: I can see how differentiating `purges_in_root` and `purges_unrooted` would be more self documenting, but this does eat an extra allocation. A potential alternative is to just use `purges` but rely on comments in the various places `purges` is added to for documentation

runtime/src/accounts_index.rs
```diff
@@ -1185,6 +1185,7 @@ impl<T: 'static + Clone + IsCached + ZeroLamport> AccountsIndex<T> {
         max_clean_root: Option<Slot>,
         account_indexes: &HashSet<AccountIndex>,
     ) {
+        let mut empty_key = false;
```
carllin, line 1188: nit: prefix booleans with a predicate like 'is`, i.e. rename `empty_key` -> `is_empty_key` or `is_slot_list_empty`

runtime/src/accounts_index.rs
```diff
@@ -1194,12 +1195,21 @@ impl<T: 'static + Clone + IsCached + ZeroLamport> AccountsIndex<T> {
                     max_clean_root,
                     account_indexes,
                 );
+                empty_key = slot_list.is_empty();
             });
         }
+        if empty_key {
+            let mut w_maps = self.account_maps.write().unwrap();
+            if let Some(x) = w_maps.get(pubkey) {
+                if x.slot_list.read().unwrap().is_empty() {
```
carllin, line 1204: As a good exercise, could we add a comment about why this additional check is necessary? What race are we trying to prevent?

carllin:
Just had some nits, good stuff 😃!

runtime/src/accounts_index.rs
brooksprumo: For the `can_purge()` function, there's not a slot list, so should the rename still be `max_cleanable_root_in_slot_list`? I also don't think it should be `max_cleanable_root`, since the other parameter is called `max_clean_root` and that may be confusing (it is to me, at least!).

runtime/src/accounts_db.rs
brooksprumo: Done!

runtime/src/accounts_db.rs
brooksprumo: Which would you prefer?

Another options is to call `clean_old_rooted_accounts()` twice, once for `purges_in_root` and once for `purges_unrooted`.

runtime/src/accounts_index.rs
brooksprumo: Done!

runtime/src/accounts_index.rs
brooksprumo: Done!

runtime/src/accounts_index.rs
```diff
@@ -978,7 +978,7 @@ impl<T: 'static + Clone + IsCached + ZeroLamport> AccountsIndex<T> {
     }
 
     // Get the maximum root <= `max_allowed_root` from the given `slice`
-    fn get_max_root(
+    fn get_max_root_in_slot_list(
```
ryoqun, line 981: here

runtime/src/accounts_index.rs
```diff
@@ -1194,12 +1199,21 @@ impl<T: 'static + Clone + IsCached + ZeroLamport> AccountsIndex<T> {
...
+                    w_maps.remove(pubkey);
+                }
+            }
+        }
     }
 
-    pub fn can_purge(max_root: Slot, slot: Slot) -> bool {
-        slot < max_root
+    fn can_purge_older_entries(max_clean_root: Slot, max_cleanable_root: Slot, slot: Slot) -> bool {
+        slot < max_clean_root && slot != max_cleanable_root
```
ryoqun, line 1216: this line isn't so self-explanatory.. I mean there is little aid for readers to understand the difference of those max clean root variables, only with or without `able`.. how about introducing `newest_` to avoid overloading `max_` to mean the slot is something most recent in the AccountsIndex list. this will yet again change this fn name... https://github.com/solana-labs/solana/pull/16911/files#r623961443 ;)

Then, this line would read like this:
```rust
slot < max_clean_root && slot != newest_root_in_slot_list`
```

runtime/src/accounts_index.rs
brooksprumo: Done! (I used your text pretty much verbatim, hope that's OK!)

runtime/src/accounts_index.rs
brooksprumo: I like this idea. I personally found it confusing to understand the different "max" types, so I'm onboard!

runtime/src/accounts_db.rs
```diff
@@ -1633,8 +1641,9 @@ impl AccountsDb {
...
         let (purged_account_slots, removed_accounts) =
-            self.clean_old_rooted_accounts(purges_in_root, max_clean_root);
+            self.clean_old_rooted_accounts(purges, max_clean_root);
```
ryoqun, line 1646: nits: how about renaming: `self.clean_accounts_older_than_root()` because we're now handling unrooted slots? more better naming is welcome :)

runtime/src/accounts_db.rs
ryoqun: or, some `iter::chain()`?

runtime/src/accounts_db.rs
```diff
@@ -1565,20 +1565,21 @@ impl AccountsDb {
...
         let mut accounts_scan = Measure::start("accounts_scan");
         // parallel scan the index.
-        let (mut purges, purges_in_root) = {
+        let (mut purges_zero_lamports, purges_in_root, purges_unrooted) = {
             self.thread_pool_clean.install(|| {
                 pubkeys
                     .par_chunks(4096)
                     .map(|pubkeys: &[Pubkey]| {
+                        let mut purges_zero_lamports = HashMap::new();
                         let mut purges_in_root = Vec::new();
```
ryoqun, line 1574: nits: another rename `purges_rooted` to be paired with `purges_unrooted`?

runtime/src/accounts_index.rs
brooksprumo: OK, this was changed to `newest_root_in_slot_list`, as per @ryoqun 's suggestion below.

runtime/src/accounts_db.rs
brooksprumo: Hah, yes totally agree!

runtime/src/accounts_db.rs
```diff
@@ -3056,9 +3065,6 @@ impl AccountsDb {
             // It should not be possible that a slot is neither in the cache or storage. Even in
             // a slot with all ticks, `Bank::new_from_parent()` immediately stores some sysvars
             // on bank creation.
-
-            // Remove any delta pubkey set if existing.
-            self.uncleaned_pubkeys.remove(remove_slot);
```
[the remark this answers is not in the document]
brooksprumo: This line flies under the radar, but is quite important. Without this change, unrooted slots are removed from the `uncleaned_pubkeys` list. So then when `clean_accounts()` runs, unrooted slots would not be considered for cleanup.

runtime/src/accounts_db.rs

runtime/src/accounts_index.rs
brooksprumo: Done!

runtime/src/accounts_db.rs

runtime/src/accounts_db.rs
carllin: > or, some iter::chain()?
Considered this as well, but it needs to be able to support par_iter() in `clean_old_rooted_accounts`

> I've collapsed these two vecs down into a single vec called purges_old_accounts now
this makes the most sense to me since we're handing this off to `par_iter()`

runtime/src/accounts_db.rs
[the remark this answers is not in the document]
carllin: @ryoqun, Yeah, the fear here I think is Bank::drop() of an unrooted bank would drop the `pubkeys` diff

carllin (approved):

PR #16917, opened 2021-04-28
bank deposit checked_add error

jeffwashington:
#### Problem

#### Summary of Changes
Modify code to return Results and use checked_add_lamports.

Fixes #

### DIFF
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -51,6 +51,7 @@ use solana_sdk::{
     incinerator,
     inflation::Inflation,
     instruction::CompiledInstruction,
+    lamports::LamportsError,
     message::Message,
     native_loader,
     native_token::sol_to_lamports,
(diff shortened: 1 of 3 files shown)


runtime/src/bank.rs
```diff
@@ -2001,23 +2002,31 @@ impl Bank {
...
+                            RewardInfo {
+                                reward_type: RewardType::Fee,
+                                lamports: unburned as i64,
+                                post_balance,
+                            },
+                        ));
+                    }
+                }
+                Err(_) => {
+                    error!("Incinerated {} fee lamports", unburned);
```
jeffwashington, line 2026: @jackcmay @t-nelson this one is special here. It follows the pattern from the other incinerator pr which we haven't closed on yet. @mvines 
I'm not entirely sure these are right and if we really need the incinerator account vs decrementing cap as the code already does in some conditions.

jackcmay, line 2026: Add `collector_id` to error message

jackcmay: Seems fine, be nice to have @mvines eyes on it since he added this

mvines: This looks fine to me.  The `unburned` variable name is now super confusing though: `burned += unburned;` 🙃 
Consider renaming `unburned` to something like `collector_fee_after_burn`

runtime/src/bank.rs
```diff
@@ -2001,25 +2002,36 @@ impl Bank {
         let collector_fees = self.collector_fees.load(Relaxed) as u64;
 
         if collector_fees != 0 {
-            let (unburned, burned) = self.fee_rate_governor.burn(collector_fees);
+            let (deposit, mut burn) = self.fee_rate_governor.burn(collector_fees);
```
jeffwashington, line 2005: @mvines better?

runtime/src/bank.rs
```diff
@@ -2001,25 +2002,36 @@ impl Bank {
...
+                                reward_type: RewardType::Fee,
+                                lamports: deposit as i64,
+                                post_balance,
+                            },
+                        ));
+                    }
+                }
+                Err(_) => {
+                    error!(
+                        "Incinerated {} fee lamports instead of sending to {}",
```
mvines, line 2027: nit:
```suggestion
                        "Burning {} fee instead of crediting {}",
```

runtime/src/bank.rs
```diff
@@ -2001,25 +2002,36 @@ impl Bank {
...
+                        deposit, self.collector_id
+                    );
+                    inc_new_counter_error!("bank-incinerated_fee_lamports", deposit as usize);
```
mvines, line 2030: nit:
```suggestion
                    inc_new_counter_error!("bank-burned_fee_lamports", deposit as usize);
```

mvines (dismissed):
lgtm!  My nits are just about removing the implication that we're using the Incinerator  address here since we aren't

t-nelson (approved):

jackcmay (approved):

PR #16925, opened 2021-04-28
Add rocksdb compression option

### LINKED ISSUE #16235: Plumb rocksdb flag to enable compression
#### Problem

The ledger data can take up a lot of space

#### Proposed Solution

rocksdb supports different compression types for different levels, but nobody has really experimented with how it affects the size and performance of the database.

A validator option could be added to give validators the option to enable compression and see how it performs.

sakridge:
#### Problem

rocksdb has compression modes that are not being used but could be helpful to the size of the rocksdb database.

#### Summary of Changes

Plumb through optional compression option.

Fixes #16235

### DIFF
--- a/ledger/src/blockstore_db.rs
+++ b/ledger/src/blockstore_db.rs
@@ -5,8 +5,8 @@ use log::*;
 use prost::Message;
 pub use rocksdb::Direction as IteratorDirection;
 use rocksdb::{
-    self, ColumnFamily, ColumnFamilyDescriptor, DBIterator, DBRawIterator, DBRecoveryMode,
-    IteratorMode as RocksIteratorMode, Options, WriteBatch as RWriteBatch, DB,
+    self, ColumnFamily, ColumnFamilyDescriptor, DBCompressionType, DBIterator, DBRawIterator,
+    DBRecoveryMode, IteratorMode as RocksIteratorMode, Options, WriteBatch as RWriteBatch, DB,
 };
 use serde::de::DeserializeOwned;
 use serde::Serialize;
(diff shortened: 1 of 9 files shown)


ledger/src/blockstore_db.rs
```diff
@@ -227,6 +259,18 @@ impl Rocks {
...
+            let mode = compression_mode.into();
+            let compression = [
+                DBCompressionType::None,
+                DBCompressionType::None,
+                mode,
+                mode,
+                mode,
+                mode,
+            ];
+            db_options.set_compression_per_level(&compression);
```
sakridge, line 272: @carllin @ryoqun I tried `set_compression_per_level` and also `db_options.set_compression_type`. Neither of them actually seem to work when I check the `rocksdb/LOG` file. When I add the `snappy` feature, then it uses `snappy` as the enabled compression option regardless of the setting here. Does it seem right?

carllin: hmmmm because each column family has its own set of levels, I think you may have to add the `set_compression_per_level` to  the column family specific options returned from `get_cf_options`

sakridge: ah, that was it. Also had to read the log correctly. First it lists a "default" column family options, then each individually. I thought it was like a global setting.

carllin: nice, glad it was a simple fix!

steviez (approved):
Looks good to me!

Out of curiosity, was there any performance hit? It looks like we keep no compression in first two levels, so maybe the the compression is negligible overhead at these higher levels?

PR #16899, opened 2021-04-28
cli: Add skip rate to `solana validators`

mvines:

PR #16928, opened 2021-04-28
Add getBlockProduction RPC method

mvines:
There's no way to easily obtain block production information over RPC.  However the validator does have this information in hand for the current and previous epoch.

Prerequisite for #16899


client/src/rpc_response.rs
```diff
@@ -211,6 +211,21 @@ pub struct RpcContactInfo {
...
+pub struct RpcBlockProductionRange {
+    pub first_slot: Slot,
+    pub last_slot: Slot,
+}
+
+#[derive(Serialize, Deserialize, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct RpcBlockProduction {
+    /// Map of leader base58 identity pubkeys to a tuple of `(number of leader slots, number of blocks produced)`
+    pub by_identity: HashMap<String, (usize, usize)>,
```
mvines, line 225: I'm not a super fan of the unnamed tuple as a return type.  I mostly did this to avoid a bunch of repetitive object file names (`leaderSlots`, `blocksProducted`) in the JSON RPC response.  Not worth it?

mvines:
Turns out this method is pretty zippy!  

I was tempted to roll block production into `getVoteAccounts`, but `getVoteAccounts` can (and does!) return multiple vote accounts for the same validator identity and I didn't want to deal with that complexity

core/src/rpc.rs
```diff
@@ -3154,6 +3174,94 @@ pub mod rpc_full {
...
+                    "endSlot, {}, is too large; max {}",
+                    last_slot,
+                    slot_history.newest()
+                )));
+            }
+
+            let slot_leaders = meta.get_slot_leaders(
+                config.commitment,
+                first_slot,
+                last_slot.saturating_sub(first_slot) as usize + 1,
```
CriesofCarrots, line 3229:
```suggestion
                last_slot.saturating_sub(first_slot) as usize + 1, // +1 because last_slot is inclusive
```

CriesofCarrots (dismissed):
Nice, lgtm! Just a couple comment suggestions

PR #16938, opened 2021-04-29
Docs/staking with exodus wallet

remjx:
#### Problem

Exodus wallet was not on the list of Supported Wallets for Staking

#### Summary of Changes

Added it with link to their FAQ.

### DIFF
--- a/docs/src/staking.md
+++ b/docs/src/staking.md
@@ -68,7 +68,9 @@ Check out our [guide to using SolFlare](wallet-guide/solflare.md) for details.
  Ledger Nano.
  [Staking commands using the Solana Command Line Tools](cli/delegate-stake.md).
 
- - Exodus makes the process very simple. See their [FAQ](https://support.exodus.com/article/1551-solana-staking-faq) for details.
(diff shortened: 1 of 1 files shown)


docs/src/staking.md
```diff
@@ -68,6 +68,8 @@ Check out our [guide to using SolFlare](wallet-guide/solflare.md) for details.
  Ledger Nano.
  [Staking commands using the Solana Command Line Tools](cli/delegate-stake.md).
 
+ - Exodus makes the process very simple. See their [FAQ](https://support.exodus.com/article/1551-solana-staking-faq) for details.
```
jstarry, line 71: I think we should add a note that it's not possible to select a validator to stake to with this wallet. Similar to how Binance and FTX work (which also aren't listed here)

remjx: Sure. I will update.

remjx: @jstarry updated,

jstarry: Thanks!

PR #16952, opened 2021-04-29
implements cursor for gossip crds table queries

behzadnouri:
#### Problem
`VersionedCrdsValue.insert_timestamp` is used for fetching crds values
inserted since last query:

So it is crucial that insert_timestamp does not go backward in time when
new values are inserted into the table. However `std::time::SystemTime` is
**not** monotonic, or due to workload, lock contention, thread scheduling,
etc, ... new values maybe inserted with a stalled timestamp way in the
past. Additionally, reading system time for above purpose is
inefficient/unnecessary.

#### Summary of Changes
This commit adds an ordinal index to crds value indicating their insert
order. Additionally, it implements a new `Cursor` type for fetching values
inserted since last query.

### DIFF
--- a/core/src/cluster_slots.rs
+++ b/core/src/cluster_slots.rs
@@ -1,5 +1,5 @@
 use crate::{
-    cluster_info::ClusterInfo, contact_info::ContactInfo, epoch_slots::EpochSlots,
+    cluster_info::ClusterInfo, contact_info::ContactInfo, crds::Cursor, epoch_slots::EpochSlots,
     serve_repair::RepairType,
 };
 use itertools::Itertools;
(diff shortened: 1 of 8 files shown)


core/src/crds.rs
```diff
@@ -153,33 +168,30 @@ impl Crds {
...
-                            .insert((value.insert_timestamp, entry_index));
+                        self.epoch_slots.insert(value.ordinal, entry_index);
                     }
                     _ => (),
                 };
                 self.records
                     .entry(value.value.pubkey())
                     .or_default()
                     .insert(entry_index);
+                self.cursor.consume(value.ordinal);
```
carllin, line 179: ah nice! this prevents us from losing values in calls like `get_votes(since)` if a new value with the same `timestamp == since` arrived since the last fetch.

carllin (approved):

core/src/cluster_slots.rs
```diff
@@ -22,22 +19,26 @@ pub type SlotPubkeys = HashMap<Pubkey, u64>;
 #[derive(Default)]
 pub struct ClusterSlots {
     cluster_slots: RwLock<BTreeMap<Slot, Arc<RwLock<SlotPubkeys>>>>,
-    since: AtomicU64,
     validator_stakes: RwLock<Arc<NodeIdToVoteAccounts>>,
     epoch: RwLock<Option<u64>>,
+    cursor: Mutex<Cursor>,
```
carllin, line 24: Would be nice if we had an `AtomicCursor` as well that supported a backing AtomicU64

behzadnouri: yeah, that makes sense

PR #16955, opened 2021-04-29
Add test to ensure data shreds with empty data would be inserted

steviez:
#### Problem
Per discussion linked below, it isn't immediately obvious that empty shreds (headers but no payload) would be inserted into blockstore
https://github.com/solana-labs/solana/pull/16602#discussion_r623232699

#### Summary of Changes
Add test to ensure shreds created with no data are treated as valid in the eyes of the blockstore

### DIFF
--- a/ledger/src/blockstore.rs
+++ b/ledger/src/blockstore.rs
@@ -5303,17 +5303,46 @@ pub mod tests {
             let mut shred5 = shreds[5].clone();
             shred5.payload.push(10);
             shred5.data_header.size = shred5.payload.len() as u16;
-            assert_eq!(
-                blockstore.should_insert_data_shred(
-                    &shred5,
-                    &slot_meta,
-                    &HashMap::new(),
-                    &last_root,
-                    None,
-                    false
-                ),
+            assert!(!blockstore.should_insert_data_shred(
+                &shred5,
+                &slot_meta,
+                &HashMap::new(),
+                &last_root,
+                None,
                 false
+            ));
+
+            // Ensure that an empty shred (one with no data) would get inserted. Such shreds
+            // may be used as signals (broadcast does so to indicate a slot was interrupted)
+            // Reuse shred5's header values to avoid a false negative result
+            let mut empty_shred = Shred::new_from_data(
+                shred5.common_header.slot,
+                shred5.common_header.index,
+                shred5.data_header.parent_offset,
+                None, // data
+                true, // is_last_data
+                true, // is_last_in_slot
+                0,    // reference_tick
+                shred5.common_header.version,
+                shred5.common_header.fec_set_index,
             );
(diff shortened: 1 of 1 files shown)


ledger/src/blockstore.rs
```diff
@@ -5315,6 +5315,32 @@ pub mod tests {
...
+                shred5.common_header.index,
+                shred5.data_header.parent_offset,
+                None, // data
+                true, // is_last_data
+                true, // is_last_in_slot
+                0,    // reference_tick
+                shred5.common_header.version,
+                shred5.common_header.fec_set_index,
+            );
+            assert_eq!(
```
behzadnouri, line 5332: `assert!` instead of `assert_eq!(..., true)`

ledger/src/blockstore.rs
```diff
@@ -5315,6 +5315,32 @@ pub mod tests {
...
+                blockstore.should_insert_data_shred(
+                    &empty_shred,
+                    &slot_meta,
+                    &HashMap::new(),
+                    &last_root,
+                    None,
+                    false
+                ),
+                true
+            );
```
behzadnouri, line 5342: I think after this assert, we also need to set `shred.data_header.size` to 0, and confirm that it then returns false.

Currently, I think this is short circuiting at this line:
https://github.com/solana-labs/solana/blob/63c56b57a/ledger/src/blockstore.rs#L1287-L1289

and does not even hit `if shred.data_header.size == 0 {` check.

steviez: So I thought the same thing the first time I looked at that snippet; however, that statement is only assigning a value to `last_in_slot` (with the added debug statement), not returning from the entire function.
```
        let last_in_slot = if shred.last_in_slot() {
            debug!("got last in slot");
            true
        } else {
            false
        };
```

behzadnouri: oh, you are right.
so maybe just set the `shred.data_header.size` to 0 after this assert, and confirm that it then returns false.

ledger/src/blockstore.rs
```diff
@@ -1292,9 +1292,39 @@ impl Blockstore {
         };
 
         if shred.data_header.size == 0 {
+            let leader_pubkey = leader_schedule
+                .map(|leader_schedule| leader_schedule.slot_leader_at(slot, None))
+                .unwrap_or(None);
```
behzadnouri, line 1297: instead of `map(...).unwrap_or(None)` you can use `.and_then`:
https://doc.rust-lang.org/std/option/enum.Option.html#method.and_then

similarly for the other case below

behzadnouri (approved):
lgtm

ledger/src/blockstore.rs
steviez: Thanks for the tip! Didn't know about `.and_then()`

PR #16958, opened 2021-04-29
Refactor `AccountsDb::get_accounts_delta_hash()`

brooksprumo:
The function `get_accounts_delta_hash()` does quite a lot of work.
While addressing another issue (#16580), I found that I need to redo some
of the same work that this function does.  So I've refactored the impl
of `get_accounts_delta_hash()` into sub-functions, that I'll then be
able to use later.

This commit introduces _zero_ behavior changes.

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -4252,69 +4252,110 @@ impl AccountsDb {
         }
     }
 
-    pub fn get_accounts_delta_hash(&self, slot: Slot) -> Hash {
-        let mut scan = Measure::start("scan");
-
-        let scan_result: ScanStorageResult<(Pubkey, Hash), DashMapVersionHash> = self
-            .scan_account_storage(
-                slot,
-                |loaded_account: LoadedAccount| {
-                    // Cache only has one version per key, don't need to worry about versioning
-                    Some((*loaded_account.pubkey(), loaded_account.loaded_hash()))
-                },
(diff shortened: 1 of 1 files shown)


runtime/src/accounts_db.rs
```diff
@@ -4252,69 +4252,110 @@ impl AccountsDb {
...
+    /// uncleaned pubkeys
+    ///
+    /// This function is called in Bank::drop() when the bank is _not_ frozen, so that its pubkeys
+    /// are considered for cleanup
+    pub fn scan_slot_and_insert_dirty_pubkeys_into_uncleaned_pubkeys(&self, slot: Slot) {
+        let scan_result = self.scan_slot_for_dirty_pubkeys_and_hashes(slot);
+        let mapped_result = self.map_scan_result_to_pubkeys_and_hashes(scan_result);
+        let dirty_pubkeys = mapped_result
+            .into_iter()
+            .map(|(pubkey, _hash)| pubkey)
```
carllin, line 4319: Hmmm instead of dumping the `hash` here after cloning it in the scan, could we pass a custom filter `map_account: impl Fn(LoadedAccount) -> T` into `scan_slot_for_dirty_pubkeys_and_hashes` to get the exact types we want

brooksprumo: Oh, that sounds cool! I don't actually know how to do that. Can you help me with a little more hand-holding?

brooksprumo:
Based on reviews, I've absorbed these changes into another PR. Please head over to PR #17003

PR #16960, opened 2021-04-30
Delete old budget program

jeffwashington:
#### Problem
This program is not necessary anymore.
#### Summary of Changes
Delete it.
Fixes #


jeffwashington:
@garious, you are git blame on 2 of the 3 places so far where we are using things from the budget program in other tests. Can you please let me know how we should proceed?
https://buildkite.com/solana-labs/solana/builds/45490#359db7b4-ce8f-473c-93db-4574c17dbe74

garious:
Looks like those tests could point to a different program, like stake

PR #16984, opened 2021-04-30
Comprehensive fees proposal

jackcmay:
#### Problem

The current fee structure lacks a comprehensive account of the work required by
a validator to process a transaction

#### Summary of Changes

Initial thoughts on how to improve the fee structure, feedback please

Fixes #

### DIFF
--- a/docs/src/proposals/comprehensive-compute-fees.md
+++ b/docs/src/proposals/comprehensive-compute-fees.md
@@ -1,0 +1,88 @@
+---
+title: Comprehensive Compute Fees
+---
+
+## Motivation
+
+The current fee structure lacks a comprehensive account of the work required by
+a validator to process a transaction.  The fee structure is only based on the
+number of signatures in a transaction but is meant to account for the work that
+the validator must perform to validate each transaction.  The validator performs
+a lot more user-defined work than just signature verification.  The work to
+process a transaction typically includes signature verifications, account
+locking, account loading, and instruction processing.
+
+## Proposed Solution
+
+### New fee structure
+
+In addition to signature checking, the other work required to process an
+instruction should be taken into account.  To do this, the total fee should be
+an accumulated cost of all the work.  Each piece of work can be
+measured/accounted for in a way that makes sense for the type of work involved.
+
+A new fee structure could include:
+1. flat fee per signature
+2. flat fee for each write lock
+3. per-byte fee for the amount of data in each loaded account
+4. per-compute-unit fee for the number of compute units used to process an
+   instruction
+
+Fees 1-3 can be determined upfront before the message is processed.  #4 would
+need to be finally accounted for after the message is processed, but the payer's
+balance could be pre-deducted against a compute budget cap to ensure they have
+enough lamports to cover a max fee, and then any unused compute budget could be
+credited back after the message is fully processed.
+
+The goal of the fees is to cover the computation cost of processing a
+transaction.  Each of the above fee categories could be represented as a compute
+unit cost that, when added together, encompasses the entire cost of processing
+the transaction.  By calculating the total cost of the transaction the runtime
+can make better decisions on what transactions to process and when, as well as
+throw out transactions that exceed a cap.
+
+The per-compute unit fee doesn't have to be linear; developers could be
+non-linearly incentivized to reduce compute costs by optimizing in any of the 4
+fee categories listed above.
+
+To give developers some control over fees and the compute cap, a new built-in
+instruction could be introduced that requests a specific transaction-wide
+compute budget cap.  The instruction can be used by a developer to reduce the
+cap and thus fees they expect to pay, or to request a higher than the default
+cap.  The runtime could in-turn, use these instructions to determine how to
+schedule these more expensive transactions.
+
+### Transaction compute caps
+
+The current compute caps are independently applied to individual instructions.
+This means the overall transaction cap varies depending on how many instructions
+are in the transaction.  Instead, a transaction-wide cap is probably more
+appropriate.  One challenge of the transaction-wide cap is that each instruction
+(program) could no longer expect to be given an equal amount of compute units
+since the number of compute units will be based on the number of units already
+consumed by earlier instructions in the message.  This will provide some
+additional tuning and composability challenges for developers.
+
+## Proposed steps to implement
+
+- Apply the compute budget cap across the entire transaction rather than
+  per-instruction
+- Convert the per-sig fee to a per-sig compute cost and incorporate it into the
+  transaction-wide cap.  Initially, keep charging the current per-sig fee.
+- Add compute costs for write locks and account data loading and incorporate
+  those into the transaction-wide cap
+- Remove the per-sig centric fee and start charging fees based on the
+  transaction compute budget
+- Rework how the bank decides which transactions to process and when based on
+  the compute budget expectations
+- Add a built-in instruction that requests the amount of compute budget up front
+
+  ## Things to ponder
+
+  - Can things like write locks be meaningfully converted into a compute
+    cost/cap?
+  - Should account data size be accounted for in the compute cost/cap
+  - Calculating a transaction's fee upfront becomes more difficult, would
+    probably have to be reworded to be max fee based on the compute cap.  Actual
+    fee would be equal to or less than that.
+  - "Compute cost" is a lame name, ideas for something better?
--- a/docs/src/proposals/comprehensive-compute-fees.md
+++ b/docs/src/proposals/comprehensive-compute-fees.md
@@ -100,6 +100,21 @@ each builtin program's instructions.
   be equal to or less than that.
 - "Compute cost" is a lame name, ideas for something better?
 
+## Mainnet beta transaction wide compute measurements
+
+Over 4 1/2 hours of Mainnet beta monitoring with a 200k instruction cap:
+675kPX9MHTjS2zt1qfr1NYHuzeLXfQM9H24wFSUt1Mp8 (Radium v3?)
+  - exceeded 79669 times
+  - exceeded by between 1 and 800k units
+5fNfvyp5czQVX77yoACa3JJVEhdRaWjPuazuWgjhTqEH (Mango!)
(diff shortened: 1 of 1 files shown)


jackcmay:
@sakridge Can you take a look at this?

docs/src/proposals/comprehensive-compute-fees.md
```diff
@@ -0,0 +1,88 @@
...
+To give developers some control over fees and the compute cap, a new built-in
+instruction could be introduced that requests a specific transaction-wide
+compute budget cap.  The instruction can be used by a developer to reduce the
+cap and thus fees they expect to pay, or to request a higher than the default
+cap.  The runtime could in-turn, use these instructions to determine how to
+schedule these more expensive transactions.
+
+### Transaction compute caps
+
+The current compute caps are independently applied to individual instructions.
```
jackcmay, line 57: @sakridge planning to start with applying a transaction-wide cap, thoughts?

docs/src/proposals/comprehensive-compute-fees.md
```diff
@@ -0,0 +1,88 @@
...
+fee categories listed above.
+
+To give developers some control over fees and the compute cap, a new built-in
+instruction could be introduced that requests a specific transaction-wide
+compute budget cap.  The instruction can be used by a developer to reduce the
+cap and thus fees they expect to pay, or to request a higher than the default
+cap.  The runtime could in-turn, use these instructions to determine how to
+schedule these more expensive transactions.
+
+### Transaction compute caps
```
jackcmay, line 55: @sakridge, @mvines @jstarry  Planning to start with applying a transaction-wide cap, thoughts?

docs/src/proposals/comprehensive-compute-fees.md
```diff
@@ -0,0 +1,97 @@
...
+  transaction-wide cap.  Initially, keep charging the current per-sig fee.
+- Add compute costs for write locks and account data loading and incorporate
+  those into the transaction-wide cap
+- Remove the per-sig centric fee and start charging fees based on the
+  transaction compute budget
+- Rework how the bank decides which transactions to process and when based on
+  the compute budget expectations
+- Add a built-in instruction that requests the amount of compute budget up front
+
+  ## Things to ponder
```
mvines, line 89: I think devs would need an option into the `simulateTransaction` RPC method to figure the cost of a transaction, since part of the cost is now based on what code paths get executed

docs/src/proposals/comprehensive-compute-fees.md
mvines: (not sure I have much to add here at the moment)

jackcmay: I'm looking for opinions on how much of an effect this would have on developers and if applying a tx wide cap makes sense where each instruction charges against a common meter (rather then what we have now where each instruction has its own budget)

docs/src/proposals/comprehensive-compute-fees.md
```diff
@@ -0,0 +1,97 @@
...
+an accumulated cost of all the work.  Each piece of work can be
+measured/accounted for in a way that makes sense for the type of work involved.
+
+A new fee structure could include:
+1. flat fee per signature
+2. flat fee for each write lock
+3. per-byte fee for the amount of data in each loaded account
+4. per-compute-unit fee for the number of compute units used to process an
+   instruction
+
```
mvines, line 30: There's one more here: there are some instructions that require an extra fee, currently just the [secp256k1_program](https://github.com/solana-labs/solana/blob/7cea2c446656510da7cbe943d3e87c18e794a76e/sdk/program/src/fee_calculator.rs#L24).  We consider that to be an extra signature fee right now but generalizing into the concept of there exists programs that cost X to use them seem worthwhile if we're redoing the fee structure

docs/src/proposals/comprehensive-compute-fees.md
mvines: I mean I think we need to do it.  What impact it has could be determined by running a hacked validator against mainnet to see how many existing transactions would start failing when we do this.

sakridge:
> @sakridge Can you take a look at this?

Nice. I think this makes a lot of sense.

The per-byte account loading fee would be good, but I do also see the downside of it being more difficult to compute the fee.

sakridge:
cc @taozhu-chicago who is working on the transaction cost model which relates to this.

docs/src/proposals/comprehensive-compute-fees.md
sakridge: +1 impact would depend on what transactions are going into the system now and then old vs. new costs.

docs/src/proposals/comprehensive-compute-fees.md
```diff
@@ -0,0 +1,97 @@
...
+credited back after the message is fully processed.
+
+The goal of the fees is to cover the computation cost of processing a
+transaction.  Each of the above fee categories could be represented as a compute
+unit cost that, when added together, encompasses the entire cost of processing
+the transaction.  By calculating the total cost of the transaction the runtime
+can make better decisions on what transactions to process and when, as well as
+throw out transactions that exceed a cap.
+
+The per-compute unit fee doesn't have to be linear; developers could be
```
sakridge, line 44: I think we would want to like buckets with a fee rate for each one. The Transaction can request 100k compute units for fee `X`, then 200k for fee `2 * X`, etc. Then the bank-less leader can still rate the transactions accurately without execution.

jackcmay: Yeah, for bank-less leader, we would want to know the caps upfront.  There could be a default cap as well as a requested cap (one of the instructions in the tx specifies it).  One challenge with the bank-less leader is that instructions using the default cap will probably be over-charged and I suspect relatively highly since they would be charged on a worst-case basis.  Also I was thinking that we might want to charge non-linearly so that higher cap buckets require relatively higher per-unit cost to encourage developers to optimize.  These kinds of questions probably require input from @ericlwilliams

taozhu-chicago: the cost model PR (https://github.com/solana-labs/solana/pull/16694) currently assigns fixed 'cost' artificially to instructions by type; implementing proposed fee structure there would be great

jackcmay: Each bucket would have its own target and each compute unit range would be its own bucket, kinda like tax brackets

jon-chuang: > One challenge with the bank-less leader is that instructions using the default cap will probably be over-charged and I suspect relatively highly since they would be charged on a worst-case basis.

For the particular problem of page faults, which probably have the worst difference between average and worst case, I think allowing certain accounts to pay a steeper rent to be "Ram-Resident" could make sense, and accessing those would require smaller fees. Practically, the mechanism could be to never flush these accounts from the read only cache.

jackcmay:
I took a look a program execution metrics on mainnet beta, you can find some of my observations here: https://github.com/solana-labs/research/tree/master/compute-budget
One goal was to establish what we should set for a proposed transaction-wide compute unit cap.  Looks like except Radium v3 all the programs I saw running in a 5 minute window on mainnet were well under the current 200k max instruction-wide cap.  I’m thinking we should initially set the transaction-wide cap to 200k units.
Another goal  was to establish a target compute unit/us rate we can use to calibrate the cost of new program operations.  Our current rate varies quite a bit which to me means that current cost rates vary from operation to operation.  The current measured rate is also very low, and if all programs maxed out their compute budget we could see around 125 tps.  We constantly face pressure to allow programs to do more and more computation, we have a challenge to reconcile this with our high tps promises.  For example, cybercore has proposed an ecrecover syscall that they’ve measured at ~248us and proposed a cost of 25k compute units.  That results in a ~100 unit/us rate which at the current cap of 200k instructions results in a max of ~500 tps.  That’s a big discrepancy between where we want to be and where we are.

docs/src/proposals/comprehensive-compute-fees.md
```diff
@@ -0,0 +1,190 @@
...
+EhhTKczWMGQt46ynNeRX1WfeagwwJd7ufHvCDjRxjo5Q                 634    1557  111  292  294   29485   6100  11707  8519   53  47 101
+FjJ5mhdRWeuaaremiHjeQextaAw1sKWDqr3D7pXjgztv                  86    2484  951  968  226   40565  40565  40565     0   43  42 179
+J21zqcffYQU2NUJrvAKhqpKZLQK1YaB9dY5kmcxkMQvQ                  72     952  332  334   96   10645  10315  10331   128   32  31 108
+SwaPpA9LAaLfeLi3a68M4DjnLqgtticKg6CnyNwgAC8                  203    9947 1076 1198  896   66633  58706  57275  4676   53  47  61
+TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA                72286    8453  101  112   98    4429   3531   3026   683   29  27  31
+Vote111111111111111111111111111111111111111                        39278  121  160  400
+11111111111111111111111111111111                                      24   11   11    3
+```
+
+- Handy compute unit conversions
```
jon-chuang, line 143: I guess this is with the caveat that the min TPS here is a limit per contract/account write lock.

Parallelised execution within a `TransactionBatch` of 128 txs, using work-stealing or other load balancing method is still a meaningful way of hiding the worst-case execution latency and catering to the median, both in cu/us, and median cu overall.

So even assuming a moderate amount of worker threads (16) and targetting the median cu of 5000, let's say, at 10cu/us, one could still achieve an average of 500us per tx at a 4+ms batch time. That's still a moderate 32K TPS. This will only scale with num_cores, although perhaps we may want to increase the batch sizes so that there are enough average time txs to hide the worst case time txs.

Also for this reason, I'd prefer to see average rather than median execution time.

This analysis depends on the average execution time staying low. I think this is fairly reasonable since raydium is already a pretty complex contract that has multiple inner instructions, but seems to have a reasonable median execution time. I think with more contracts/asset pairs, the parallelisability would only increase.

PR #16987, opened 2021-05-01
Improve readability of vote lockout processing

jstarry:
#### Problem
Wasn't clear that `expiration_slot` is inclusive (meaning that the vote is still locked out at that slot) which could lead to off-by-one issues.

#### Summary of Changes
Added some comments and renamed some methods / variables

Fixes #

### DIFF
--- a/programs/vote/src/vote_state/mod.rs
+++ b/programs/vote/src/vote_state/mod.rs
@@ -91,9 +91,8 @@ impl Lockout {
         self.slot + self.lockout()
     }
 
-    // Check if a lockout has expired before checked slot.
-    pub fn has_expired_by_slot(&self, slot: Slot) -> bool {
-        self.last_locked_out_slot() < slot
+    pub fn is_locked_out_at_slot(&self, slot: Slot) -> bool {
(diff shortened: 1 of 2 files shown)


programs/vote/src/vote_state/mod.rs
```diff
@@ -84,13 +84,16 @@ impl Lockout {
...
+    // vote on a slot in another fork which is less than or equal to this slot
+    // to avoid having their stake slashed.
+    pub fn last_locked_out_slot(&self) -> Slot {
         self.slot + self.lockout()
     }
-    pub fn is_expired(&self, slot: Slot) -> bool {
-        self.expiration_slot() < slot
+
+    // Check if a lockout has expired before checked slot.
+    pub fn has_expired_by_slot(&self, slot: Slot) -> bool {
```
carllin, line 95: how about `is_locked_out_at_slot` and flipping this to `>=`

carllin (dismissed):

programs/vote/src/vote_state/mod.rs
jstarry: switched

PR #16988, opened 2021-05-01
add option --priority-trusted-validator

diman-io:
#### Problem
Sometimes I get this (see picture). And I saw many messages about this problem many times in telegram chat.
submitting 28 points

#### Summary of Changes

I have added new option `--primary-trusted-validator`. First 100 times (in loop) validator wait only primary trusted validators for download snapshot (if they presented).

### DIFF
--- a/core/src/validator.rs
+++ b/core/src/validator.rs
@@ -100,7 +100,8 @@ pub struct ValidatorConfig {
     pub fixed_leader_schedule: Option<FixedSchedule>,
     pub wait_for_supermajority: Option<Slot>,
     pub new_hard_forks: Option<Vec<Slot>>,
-    pub trusted_validators: Option<HashSet<Pubkey>>, // None = trust all
+    pub trusted_validators: Option<HashSet<Pubkey>>, // None = trust all, priority included
+    pub priority_trusted_validators: Option<HashSet<Pubkey>>, // None = all - common trusted
     pub repair_validators: Option<HashSet<Pubkey>>,  // None = repair from all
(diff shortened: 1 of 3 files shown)


mvines:
Why not just remove the `--trusted-validator`s that you don't want to use?   When deciding on what validator to download from, it would be cool either:
1. issue a simple `HTTP GET` was issued to all candidates and the one that responded fastest was used.
2. incorporate geoip information (from somewhere?) and contact the closest node

Of course this doesn't mean the overall download throughput will be fastest but at least it's better than what we have now.

diman-io:
I want to use `--trusted-validator` (and I think many node operators want, before the team say don't use). 

Yes, you are completely right. Your both cases are better, and I thinked about them too.

First, I don't have enough experience in Rust to develop them 🤦‍♀️😭

And  second, in fact I (and many others node operators) have second  reserve machine in local network (where first main machine)/same rack/same provider, on which I can run no-voting node for reserve snapshots. And this node will be sole primary-trusted-validator. I think this is the main case of using this option.

But I totally agree with you, my solution looks like a crutch without above explanation 🤣🤣🤣

PS. If you don't want this option, then close this PR, I will understand, it's really looks as a crutch.

smediainfo:
I have this problem too.

mvines:
We have some improvements in v1.7.0 here actually, where the download will abort if it's too slow and try another RPC endpoint.  I think I'd rather go down that path and see how it performs in the real world.  But thanks for the idea @diman-io, definitely appreciate you proposing the PR!

PR #17003, opened 2021-05-03
Clean unrooted unfrozen banks (#16580)

brooksprumo:
Clean unrooted unfrozen banks (#16580)

Cleanup pubkeys when an unrooted, unfrozen bank is dropped.  This is a
continuation of PR #16911.

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -4272,7 +4272,7 @@ impl AccountsDb {
         }
     }
 
-    /// Perform the scan for dirty pubkeys in a slot
+    /// Perform the scan for pubkeys that were written to in a slot
     fn do_scan_slot_for_dirty_pubkeys(
         &self,
         slot: Slot,
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -4279,9 +4279,7 @@ impl AccountsDb {
     ) -> ScanStorageResult<Pubkey, DashSet<Pubkey>> {
         self.scan_account_storage(
             slot,
-            |loaded_account: LoadedAccount| {
-                Some(*loaded_account.pubkey())
-            },
+            |loaded_account: LoadedAccount| Some(*loaded_account.pubkey()),
             |accum: &DashSet<Pubkey>, loaded_account: LoadedAccount| {
                 accum.insert(*loaded_account.pubkey());
             },
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -4272,6 +4272,52 @@ impl AccountsDb {
         }
     }
 
+    /// Perform the scan for dirty pubkeys in a slot
+    fn do_scan_slot_for_dirty_pubkeys(
+        &self,
+        slot: Slot,
+    ) -> ScanStorageResult<Pubkey, DashSet<Pubkey>> {
+        self.scan_account_storage(
+            slot,
+            |loaded_account: LoadedAccount| {
+                // Cache only has one version per key, don't need to worry about versioning
+                Some(*loaded_account.pubkey())
+            },
+            |accum: &DashSet<Pubkey>, loaded_account: LoadedAccount| {
+                accum.insert(*loaded_account.pubkey());
+            },
+        )
+    }
+
+    /// Reduce the scan result of dirty pubkeys after calling `scan_account_storage()` into a
+    /// single vec of Pubkeys.
+    fn do_reduce_scan_slot_for_dirty_pubkeys(
+        scan_result: ScanStorageResult<Pubkey, DashSet<Pubkey>>,
+    ) -> Vec<Pubkey> {
+        match scan_result {
+            ScanStorageResult::Cached(cached_result) => cached_result,
+            ScanStorageResult::Stored(stored_result) => {
+                stored_result.into_iter().collect::<Vec<_>>()
+            }
+        }
+    }
+
+    /// Scan a slot for dirty pubkeys
+    fn scan_slot_for_dirty_pubkeys(&self, slot: Slot) -> Vec<Pubkey> {
+        let dirty_pubkeys = self.do_scan_slot_for_dirty_pubkeys(slot);
+        Self::do_reduce_scan_slot_for_dirty_pubkeys(dirty_pubkeys)
+    }
+
+    /// Scan a slot in the account storage for dirty pubkeys and insert them into the list of
+    /// uncleaned pubkeys
+    ///
+    /// This function is called in Bank::drop() when the bank is _not_ frozen, so that its pubkeys
+    /// are considered for cleanup.
+    pub fn scan_slot_and_insert_dirty_pubkeys_into_uncleaned_pubkeys(&self, slot: Slot) {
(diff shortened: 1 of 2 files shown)


runtime/src/accounts_db.rs
```diff
@@ -4272,6 +4272,52 @@ impl AccountsDb {
...
 
+    /// Perform the scan for dirty pubkeys in a slot
+    fn do_scan_slot_for_dirty_pubkeys(
+        &self,
+        slot: Slot,
+    ) -> ScanStorageResult<Pubkey, DashSet<Pubkey>> {
+        self.scan_account_storage(
+            slot,
+            |loaded_account: LoadedAccount| {
+                // Cache only has one version per key, don't need to worry about versioning
```
jeffwashington, line 4283: stale comment

brooksprumo: Good catch!

runtime/src/accounts_db.rs
```diff
@@ -4272,6 +4272,52 @@ impl AccountsDb {
         }
     }
 
+    /// Perform the scan for dirty pubkeys in a slot
```
carllin, line 4275: nit: `dirty` ->`pubkeys that were written to` to be more explicit

runtime/src/accounts_db.rs
carllin: nice! yeah we don't care about versioning, just that the pubkey was touched!

runtime/src/bank.rs
```diff
@@ -5044,20 +5044,42 @@ impl Bank {
...
-                // 1. Tests
-                // 2. At startup when replaying blockstore and there's no
-                // AccountsBackgroundService to perform cleanups yet.
-                self.rc.accounts.purge_slot(self.slot());
-            }
+        if self.skip_drop.load(Relaxed) {
+            return;
+        }
+
+        self.cleanup();
```
carllin, line 5073: nice! I think this is the right place to put this in relation to the `Some(drop_callback)` check below, subtle!

runtime/src/accounts_db.rs

runtime/src/accounts_db.rs

runtime/src/bank.rs
brooksprumo: Yeah, I was wondering about where exactly to call `cleanup()`. Since the drop triggers `clean_accounts()`, and `uncleaned_pubkeys` needs to have all the pubkeys beforehand, it seems like `cleanup()` needs to be called first.

runtime/src/bank.rs
```diff
@@ -12550,4 +12572,54 @@ pub(crate) mod tests {
...
+        assert_eq!(
+            bank2
+                .rc
+                .accounts
+                .accounts_db
+                .accounts_index
+                .ref_count_from_storage(&pubkey1),
+            0
+        );
+    }
```
carllin, line 12624: Nice, I had a few additional test cases in mind, interested in your thoughts if you think they're necessary!

1) An additional case as well that if there was a `pubkey3` touched in *both* (current test case is separate pubkeys being touched) `bank1` and `bank2`, that the reference from the unfrozen `bank1` is cleaned up as well

2) If there was a zero-lamport key `K` in *only* the unfrozen `bank1`, that it is accurately cleaned up.

3)  If there was a zero-lamport key  `K` in *both* the unfrozen `bank1`, and the frozen `bank2` that it is accurately cleaned up.

runtime/src/bank.rs
carllin: yes also, regardless of whether there is a `drop_callback()`, those dirty pubkeys are still in storage, so they need to be added to the list!

runtime/src/bank.rs
brooksprumo: @carllin I like these additional tests. Would it be OK to address them in a subsequent PR, or would you like them in this one?

carllin: I think it makes sense to combine them under this same PR for tracking purposes and historical reference, but definitely not essential.

brooksprumo: @carllin Gotcha! I've added the tests now. Let me know if they look like what you were expecting!

runtime/src/bank.rs
```diff
@@ -12525,101 +12525,133 @@ pub(crate) mod tests {
...
         let collector = Pubkey::new_unique();
-        let pubkey1 = Pubkey::new_unique();
-        let pubkey2 = Pubkey::new_unique();
 
-        bank0.transfer(2, &mint_keypair, &pubkey2).unwrap();
+        let key1 = Keypair::new(); // only touched in bank1
+        let key2 = Keypair::new(); // only touched in bank2
+        let key3 = Keypair::new(); // touched in both bank1 and bank2
+        let key4 = Keypair::new(); // in only bank1, and has zero lamports
+        let key5 = Keypair::new(); // in both bank1 and bank2, and has zero lamports
```
carllin, line 12576: thanks for all the clarifying comments!

runtime/src/bank.rs
```diff
@@ -12525,101 +12525,133 @@ pub(crate) mod tests {
...
 
         let slot = 1;
         let bank1 = Bank::new_from_parent(&bank0, &collector, slot);
-        bank1.transfer(3, &mint_keypair, &pubkey1).unwrap();
-        bank1.freeze();
+        bank1.transfer(3, &mint_keypair, &key1.pubkey()).unwrap();
+
+        // make key4 have zero lamports
+        bank1.transfer(8, &mint_keypair, &key4.pubkey()).unwrap();
+        bank1.transfer(8, &key4, &key3.pubkey()).unwrap();
```
carllin, line 12587: heh, creative. Here's also a more straightforward way to store a zero lamport account https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_db.rs#L6583-L6587

runtime/src/bank.rs
```diff
@@ -12525,101 +12525,133 @@ pub(crate) mod tests {
...
+                .ref_count_from_storage(&key3.pubkey()),
+            expected_ref_count_for_keys_only_in_slot_2
+        );
+        assert_eq!(
+            bank2
+                .rc
+                .accounts
+                .accounts_db
+                .accounts_index
+                .ref_count_from_storage(&key4.pubkey()),
```
carllin, line 12644: Checking the ref count to verify zero-lamport accounts is not a bad idea, but I think an additional top-level should also be made to make sure it was actually removed from the index. 

I recommend something like this check used in other zero lamport clean tests: https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_db.rs#L6626-L6629

runtime/src/bank.rs
brooksprumo: OK, I've updated the tests. I've used `Bank::store_account()` instead of `Bank::transfer()`. I'm assuming it's OK to use this version instead of `AccountsDb::store_uncached()`. Since the tests are in `bank.rs`, I wanted to only use `Bank` functions, if possible.

runtime/src/bank.rs
brooksprumo: Done!

carllin (approved):

PR #17004, opened 2021-05-03
removes position field in coding-shred-header

behzadnouri:
#### Problem
`CodingShredHeader.position` is equal to

    ShredCommonHeader.index - ShredCommonHeader.fec_set_index

and is so redundant. The extra position field can add bugs if not
consistent with `index` and `fec_set_index`.

#### Summary of Changes
Removed position field in coding-shred-header.

### DIFF
--- a/core/src/retransmit_stage.rs
+++ b/core/src/retransmit_stage.rs
@@ -787,7 +787,7 @@ mod tests {
         assert_eq!(check_if_already_received(&packet, &shreds_received), None);
         assert_eq!(check_if_already_received(&packet, &shreds_received), None);
 
-        let shred = Shred::new_empty_coding(slot, index, 0, 1, 1, 0, version);
+        let shred = Shred::new_empty_coding(slot, index, 0, 1, 1, version);
         shred.copy_to_packet(&mut packet);
         // Coding at (1, 5) passes
         assert_eq!(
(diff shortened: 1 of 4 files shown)


carllin (approved):
Checking the three places where the `position` field is used:
1. https://github.com/solana-labs/solana/blob/master/ledger/src/blockstore.rs#L1219
2. https://github.com/solana-labs/solana/blob/master/ledger/src/blockstore.rs#L1226
3. https://github.com/solana-labs/solana/blob/master/ledger/src/blockstore.rs#L1244

I think this should be safe to merge without any gating (i.e. it should be ok for a leader to broadcast position == 0 to validators who haven't upgraded to this change yet)

Thanks for the cleanup!

sakridge (approved):
lgtm, nice

carllin:
@mvines and @sakridge, fyi saw this nice reminder on compatibility for upgrade/downgrade for mainnet release + master: https://github.com/solana-labs/solana/issues/17136#issuecomment-836996089, and double checked this PR would not break that assumption, specifically the downgrade path.

@behzadnouri and I discussed this critical change: https://solanalabs.slack.com/archives/D019TRTQMA7/p1620673230006800 and decided this change should be safe even for downgrade because the 3 places this change would be detected: https://github.com/solana-labs/solana/pull/17004#pullrequestreview-650801004, should become no-ops with `position == 0`

PR #17006, opened 2021-05-03
More docs for bank.rs

sakridge:
#### Problem

Bank not documented that well.

#### Summary of Changes

Add some more docs

Fixes #


taozhu-chicago:
Thanks @sakridge it explains a lot.

taozhu-chicago (approved):

PR #17013, opened 2021-05-03
flatten_hash_intermediate sets capacity first

jeffwashington:
#### Problem
time for flatten_hash_intermediate grows linearly, but at a faster rate than other hash calculation code.
#### Summary of Changes
Pre-allocate flattened vector instead of just extend.
Fixes #


jeffwashington:
This flatten change executes in 75% of the time of the current algorithm against mnb. Not world changing, but should help. On the kin simulation, flatten alone is taking 35s at 340M accounts.

carllin (approved):

PR #17021, opened 2021-05-04
Update README.md

spindriftur:
These are instructions with an example command for minting an SPL token that is easy to understand for anyone going through the CLI.

#### Problem
Going through the default instructions and doing it myself yielded a lot of problems, so I made an example command with details to add to the README
#### Summary of Changes
Added more details to each bit of initiating the airdrop through the CLI
Fixes #
Easier to understand instructions :)

### DIFF
--- a/tokens/README.md
+++ b/tokens/README.md
@@ -20,7 +20,18 @@ CYRJWqiSjLitBAcRxPvWpgX3s5TvmN2SuRY3eEYypFvT,43.0,
 ```bash
 solana-tokens distribute-tokens --from <KEYPAIR> --input-csv <RECIPIENTS_CSV> --fee-payer <KEYPAIR>
 ```
+Example of command: 
+```
+solana-tokens distribute-spl-tokens --from <ENTER ADDRESS HERE> --owner /Users/<COMPUTERS NAME>/new-keypair.json --input-csv spl-token.csv --fee-payer /Users/<COMPUTERS NAME>/new-keypair.json --db-path airdrop2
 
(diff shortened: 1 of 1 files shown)


tokens/README.md
```diff
@@ -20,7 +20,18 @@ CYRJWqiSjLitBAcRxPvWpgX3s5TvmN2SuRY3eEYypFvT,43.0,
...
 ```
+Example of command: 
+```
+solana-tokens distribute-spl-tokens --from <ENTER ADDRESS HERE> --owner /Users/<COMPUTERS NAME>/new-keypair.json --input-csv spl-token.csv --fee-payer /Users/<COMPUTERS NAME>/new-keypair.json --db-path airdrop2
+
+Retrieve --from by typing spl-token accounts -v it will be the address under Account
+Retrieve --owner by entering the filepath to your new-keypair.json file (mine was in my home folder called "COMPUTERS NAME")
+For --input-csv I would recommend you putting the airdrop csv file (NOT .xlsx change from .xlsx to .csv) in your home folder 
+--fee-payer is going to be the same filepath as your --owner 
+--db-path can be named anything, I named it airdrop2
```
CriesofCarrots, line 31: I do think it would be helpful to have a generic statement at the top defining what ought to be in the `--input-csv` and what `--db-path` is for.

CriesofCarrots (commented):
Thanks. Overall, I think these recommendations are a bit too specific for this readme, which isn't intended to be a complete walkthrough. However, we do make some assumptions about users' familiarity with the solana command-line tools that may be incorrect. Interested in fleshing out those areas?

PR #17026, opened 2021-05-04
Document that Transaction::sign might panic

ruuda:
#### Problem

My program panicked, because somewhere in a third-party library, that library called `Transaction::new_signed_with_payer`, which can panic. I’m not sure whether the caller of `Transaction::new_signed_with_payer` was aware of that. Document it, so users can at least be aware of this.

#### Summary of Changes

Add `# Panics` sections. Where possible, include an [intra-doc-links](https://doc.rust-lang.org/nightly/rustdoc/linking-to-items-by-name.html) to the non-panicking `try_*` variants of these functions.

### DIFF
--- a/sdk/src/transaction.rs
+++ b/sdk/src/transaction.rs
@@ -234,6 +234,10 @@ impl Transaction {
     }
 
     /// Check keys and keypair lengths, then sign this transaction.
+    ///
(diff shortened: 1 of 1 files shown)


mvines:
Thanks!

PR #17034, opened 2021-05-04
Add bpf_loader_upgradeable to ProgramTest default builtins

SebastianBor:
#### Problem

bpf_loader_upgradeable program is not loaded by default by ProgramTest and has to be added manually when required.

#### Summary of Changes

Add bpf_loader_upgradeable program to default builtins for ProgramTest


mvines (approved):

mvines:
Merging, so this can get into v1.6.7

PR #17038, opened 2021-05-04
Test account index and store alignment

steviez:
#### Problem
Jeff tracked down a regression in https://github.com/solana-labs/solana/pull/16838 that was a result of the store and index falling out of sync. This test adds protection for the desired behavior of (keeping store and index in sync) in our unit tests.

#### Summary of Changes
Adds a unit test (and other very minor cleanup)

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -6695,6 +6695,44 @@ pub mod tests {
     }
 
     #[test]
+    fn test_clean_multiple_zero_lamport_store_matches_index() {
+        solana_logger::setup();
+
+        let accounts = AccountsDb::new(Vec::new(), &ClusterType::Development);
+        let pubkey1 = solana_sdk::pubkey::new_rand();
+        let pubkey2 = solana_sdk::pubkey::new_rand();
+        let zero_lamport_account =
+            AccountSharedData::new(0, 0, AccountSharedData::default().owner());
+
+        // Store account 1 & 2 in slot 0, then update account 1 in several more slots
+        accounts.store_uncached(0, &[(&pubkey1, &zero_lamport_account)]);
+        accounts.store_uncached(0, &[(&pubkey2, &zero_lamport_account)]);
+        accounts.store_uncached(1, &[(&pubkey1, &zero_lamport_account)]);
+        accounts.store_uncached(2, &[(&pubkey1, &zero_lamport_account)]);
+
+        accounts.add_root(0);
+        accounts.add_root(1);
+        accounts.add_root(2);
+
+        // Account 1 ref'd from all 3 slots, account 2 from only 1 slot
+        assert_eq!(accounts.accounts_index.ref_count_from_storage(&pubkey1), 3);
+        assert_eq!(accounts.accounts_index.ref_count_from_storage(&pubkey2), 1);
+
+        accounts.clean_accounts(None);
+        // Slots 0 and 1 should each have been cleaned because all of their
+        // accounts are zero lamports
+        assert!(accounts.storage.get_slot_stores(0).is_none());
+        assert!(accounts.storage.get_slot_stores(1).is_none());
+        // Slot 2 is still around because ...
+        assert!(accounts.storage.get_slot_stores(2).is_some());
+        // Check that the index has also been updated to unreference accounts
+        // and is in-sync with the store. Account 1 should be ref'd from slot 2,
+        // account 1 should have no remaining refs in the index
+        assert_eq!(accounts.accounts_index.ref_count_from_storage(&pubkey1), 1);
+        assert_eq!(accounts.accounts_index.ref_count_from_storage(&pubkey2), 0);
+    }
+
+    #[test]
     fn test_clean_zero_lamport_and_old_roots() {
         solana_logger::setup();
 
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -1314,7 +1314,7 @@ impl AccountsDb {
         max_clean_root: Option<Slot>,
     ) -> ReclaimResult {
         if purges.is_empty() {
-            return (HashMap::new(), HashMap::new());
+            return ReclaimResult::default();
         }
         // This number isn't carefully chosen; just guessed randomly such that
         // the hot loop will be the order of ~Xms.
@@ -1345,7 +1345,7 @@ impl AccountsDb {
         // and those stores may be used for background hashing.
         let reset_accounts = false;
 
-        let mut reclaim_result = (HashMap::new(), HashMap::new());
+        let mut reclaim_result = ReclaimResult::default();
         self.handle_reclaims(
             &reclaims,
             None,
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -6695,7 +6695,7 @@ pub mod tests {
     }
 
     #[test]
-    fn test_clean_multiple_zero_lamport_store_matches_index() {
+    fn test_clean_multiple_zero_lamport_decrements_index_ref_count() {
         solana_logger::setup();
 
         let accounts = AccountsDb::new(Vec::new(), &ClusterType::Development);
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -6704,17 +6704,18 @@ pub mod tests {
         let zero_lamport_account =
             AccountSharedData::new(0, 0, AccountSharedData::default().owner());
 
-        // Store account 1 & 2 in slot 0, then update account 1 in several more slots
+        // Store 2 accounts in slot 0, then update account 1 in two more slots
         accounts.store_uncached(0, &[(&pubkey1, &zero_lamport_account)]);
         accounts.store_uncached(0, &[(&pubkey2, &zero_lamport_account)]);
         accounts.store_uncached(1, &[(&pubkey1, &zero_lamport_account)]);
         accounts.store_uncached(2, &[(&pubkey1, &zero_lamport_account)]);
-
+        // Root all slots
         accounts.add_root(0);
         accounts.add_root(1);
         accounts.add_root(2);
 
-        // Account 1 ref'd from all 3 slots, account 2 from only 1 slot
+        // Account ref counts should match how many slots they were stored in
+        // Account 1 = 3 slots; account 2 = 1 slot
         assert_eq!(accounts.accounts_index.ref_count_from_storage(&pubkey1), 3);
         assert_eq!(accounts.accounts_index.ref_count_from_storage(&pubkey2), 1);
 
@@ -6723,13 +6724,19 @@ pub mod tests {
         // accounts are zero lamports
         assert!(accounts.storage.get_slot_stores(0).is_none());
         assert!(accounts.storage.get_slot_stores(1).is_none());
-        // Slot 2 is still around because ...
+        // Slot 2 only has a zero lamport account as well. But, calc_delete_dependencies()
+        // should exclude slot 2 from the clean due to changes in other slots
         assert!(accounts.storage.get_slot_stores(2).is_some());
-        // Check that the index has also been updated to unreference accounts
-        // and is in-sync with the store. Account 1 should be ref'd from slot 2,
-        // account 1 should have no remaining refs in the index
+        // Index ref counts should be consistent with the slot stores. Account 1 ref count
+        // should be 1 since slot 2 is the only alive slot; account 2 should have a ref
+        // count of 0 due to slot 0 being dead
         assert_eq!(accounts.accounts_index.ref_count_from_storage(&pubkey1), 1);
         assert_eq!(accounts.accounts_index.ref_count_from_storage(&pubkey2), 0);
+
+        accounts.clean_accounts(None);
+        // Slot 2 will now be cleaned, which will leave account 1 with a ref count of 0
+        assert!(accounts.storage.get_slot_stores(2).is_none());
+        assert_eq!(accounts.accounts_index.ref_count_from_storage(&pubkey1), 0);
     }
 
     #[test]


runtime/src/accounts_db.rs
```diff
@@ -6629,6 +6629,44 @@ pub mod tests {
...
+        assert_eq!(accounts.accounts_index.ref_count_from_storage(&pubkey1), 3);
+        assert_eq!(accounts.accounts_index.ref_count_from_storage(&pubkey2), 1);
+
+        accounts.clean_accounts(None);
+        // Slots 0 and 1 should each have been cleaned because all of their
+        // accounts are zero lamports
+        assert!(accounts.storage.get_slot_stores(0).is_none());
+        assert!(accounts.storage.get_slot_stores(1).is_none());
+        // Slot 2 is still around because ...
+        assert!(accounts.storage.get_slot_stores(2).is_some());
```
steviez, line 6662: I'm throwing this up here for now, but it is not yet clear to me why slot 2 isn't also cleaned as well. Slot 2 has only the same zero-lamport account as slot 1; need to step through the process a bit more to pinpoint this well enough to leave an intelligent comment.

runtime/src/accounts_db.rs
```diff
@@ -6629,6 +6629,44 @@ pub mod tests {
...
+        // Check that the index has also been updated to unreference accounts
+        // and is in-sync with the store. Account 1 should be ref'd from slot 2,
+        // account 1 should have no remaining refs in the index
+        assert_eq!(accounts.accounts_index.ref_count_from_storage(&pubkey1), 1);
```
steviez, line 6666: With the change from Jeff's PR (linked in this PR's description) removed, this assertion fails:
`accounts.accounts_index.ref_count_from_storage(&pubkey1)` returns `2`

runtime/src/accounts_db.rs
steviez: Updated the comment in new commit - figured out that a check in `AccountsDb::calc_delete_dependencies()` was the reason for slot 2 not being removed on the first clean

runtime/src/accounts_db.rs
```diff
@@ -6629,6 +6629,51 @@ pub mod tests {
...
+        // Account ref counts should match how many slots they were stored in
+        // Account 1 = 3 slots; account 2 = 1 slot
+        assert_eq!(accounts.accounts_index.ref_count_from_storage(&pubkey1), 3);
+        assert_eq!(accounts.accounts_index.ref_count_from_storage(&pubkey2), 1);
+
+        accounts.clean_accounts(None);
+        // Slots 0 and 1 should each have been cleaned because all of their
+        // accounts are zero lamports
+        assert!(accounts.storage.get_slot_stores(0).is_none());
+        assert!(accounts.storage.get_slot_stores(1).is_none());
```
carllin, line 6661: Oh this is interesting, the clean doesn't remove the zero lamport account here like I expected it would (there's no pubkey in an earlier storage entry sharing a storage to keep this one alive).

But somehow the next `clean_accounts()` does remove the zero lamport key. 

This seems like a bug 😃

runtime/src/accounts_db.rs
```diff
@@ -6629,6 +6629,51 @@ pub mod tests {
...
+        // Slot 2 only has a zero lamport account as well. But, calc_delete_dependencies()
+        // should exclude slot 2 from the clean due to changes in other slots
+        assert!(accounts.storage.get_slot_stores(2).is_some());
```
carllin, line 6664: Oh this is interesting, the clean doesn't remove the zero lamport account here as I expected it would (there's no pubkey in an earlier storage entry sharing a storage to keep this one alive).

But somehow the next `clean_accounts()` does remove the zero lamport key. 

This seems like a bug 😃

runtime/src/accounts_db.rs
```diff
@@ -6629,6 +6629,51 @@ pub mod tests {
         assert_eq!(accounts.alive_account_count_in_slot(1), 0);
     }
 
+    #[test]
+    fn test_clean_multiple_zero_lamport_store_matches_index() {
```
carllin, line 6633: nit: update test name to `test_clean_multiple_zero_lamport_decrements_index_ref_count()`

runtime/src/accounts_db.rs
jeffwashington: I saw similar behavior with the real snapshot that for some reason it took 2 passes, but 2 passes was an improvement over it never cleaning, so I stopped for the moment where I did.

runtime/src/accounts_db.rs
jeffwashington: Great! This is exactly what I hoped we would see!

jeffwashington (dismissed):

runtime/src/accounts_db.rs
steviez: > This seems like a bug worth investigating 😃

In `clean_accounts()`, the initial scan to find accounts that can be removed picks up both accounts and all of their associated account entries in `purges_old_accounts` and `purges_zero_lamports`, respectively.
<details>
<summary>https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_db.rs#L1605-L1676</summary>

```
        // Parallel scan the index.
        let (mut purges_zero_lamports, purges_old_accounts) = {
            self.thread_pool_clean.install(|| {
                pubkeys
                    .par_chunks(4096)
                    .map(|pubkeys: &[Pubkey]| {
                        let mut purges_zero_lamports = HashMap::new();
                        let mut purges_old_accounts = Vec::new();
                        for pubkey in pubkeys {
                            match self.accounts_index.get(pubkey, None, max_clean_root) {
                                AccountIndexGetResult::Found(locked_entry, index) => {
                                    let slot_list = locked_entry.slot_list();
                                    let (slot, account_info) = &slot_list[index];
                                    if account_info.lamports == 0 {
                                        purges_zero_lamports.insert(
                                            *pubkey,
                                            self.accounts_index
                                                .roots_and_ref_count(&locked_entry, max_clean_root),
                                        );
                                    } else {
                                        // prune zero_lamport_pubkey set which should contain all 0-lamport
                                        // keys whether rooted or not. A 0-lamport update may become rooted
                                        // in the future.
                                        if !slot_list
                                            .iter()
                                            .any(|(_slot, account_info)| account_info.lamports == 0)
                                        {
                                            self.accounts_index.remove_zero_lamport_key(pubkey);
                                        }
                                    }
                                    // Release the lock
                                    let slot = *slot;
                                    drop(locked_entry);

                                    if self.accounts_index.is_uncleaned_root(slot) {
                                        // Assertion enforced by `accounts_index.get()`, the latest slot
                                        // will not be greater than the given `max_clean_root`
                                        if let Some(max_clean_root) = max_clean_root {
                                            assert!(slot <= max_clean_root);
                                        }
                                        purges_old_accounts.push(*pubkey);
                                    }
                                }
                                AccountIndexGetResult::NotFoundOnFork => {
                                    // This pubkey is in the index but not in a root slot, so clean
                                    // it up by adding it to the to-be-purged list.
                                    //
                                    // Also, this pubkey must have been touched by some slot since
                                    // it was in the dirty list, so we assume that the slot it was
                                    // touched in must be unrooted.
                                    purges_old_accounts.push(*pubkey);
                                }
                                AccountIndexGetResult::Missing(lock) => {
                                    // pubkey is missing from index, so remove from zero_lamports_list
                                    self.accounts_index.remove_zero_lamport_key(pubkey);
                                    drop(lock);
                                }
                            };
                        }
                        (purges_zero_lamports, purges_old_accounts)
                    })
                    .reduce(
                        || (HashMap::new(), Vec::new()),
                        |mut m1, m2| {
                            // Collapse down the hashmaps/vecs into one.
                            m1.0.extend(m2.0);
                            m1.1.extend(m2.1);
                            m1
                        },
                    )
            })
        };
```
</details>

`purges_old_accounts` is just a vector with both accounts (pubkeys); `purges_old_accounts` is then passed into `clean_accounts_older_than_root` here.
<details>
<summary>https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_db.rs#L1679-L1680
</summary>

```
       let (purged_account_slots, removed_accounts) =
            self.clean_accounts_older_than_root(purges_old_accounts, max_clean_root);
```
</details>

The return values from this indicate that Account 1 was removed from both Slot 0 and Slot 1. Since Account 1 was the only account in Slot 1, Slot 1 gets removed (ie callstack hits `finalize_dead_slot_removal()` for Slot 1). It is worth noting that the entry for Account 1 in Slot 2 and the entry for Account 2 in Slot 0 are not marked for purge at this point due to the following function from accounts_index.rs where the newest rooted slot won't be purged
```
    fn can_purge_older_entries(
        max_clean_root: Slot,
        newest_root_in_slot_list: Slot,
        slot: Slot,
    ) -> bool {
        slot < max_clean_root && slot != newest_root_in_slot_list
    }
```

The next step runs through and calculates store counts for the items in `purges_zero_lamports` while also trimming account_info's that were already purged.
<details>
<summary>https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_db.rs#L1691-L1736</summary>

```
        // Calculate store counts as if everything was purged
        // Then purge if we can
        let mut store_counts: HashMap<AppendVecId, (usize, HashSet<Pubkey>)> = HashMap::new();
        for (key, (account_infos, ref_count)) in purges_zero_lamports.iter_mut() {
            if purged_account_slots.contains_key(&key) {
                *ref_count = self.accounts_index.ref_count_from_storage(&key);
            }
            account_infos.retain(|(slot, account_info)| {
                let was_slot_purged = purged_account_slots
                    .get(&key)
                    .map(|slots_removed| slots_removed.contains(slot))
                    .unwrap_or(false);
                if was_slot_purged {
                    // No need to look up the slot storage below if the entire
                    // slot was purged
                    return false;
                }
                // Check if this update in `slot` to the account with `key` was reclaimed earlier by
                // `clean_accounts_older_than_root()`
                let was_reclaimed = removed_accounts
                    .get(&account_info.store_id)
                    .map(|store_removed| store_removed.contains(&account_info.offset))
                    .unwrap_or(false);
                if was_reclaimed {
                    return false;
                }
                if let Some(store_count) = store_counts.get_mut(&account_info.store_id) {
                    store_count.0 -= 1;
                    store_count.1.insert(*key);
                } else {
                    let mut key_set = HashSet::new();
                    key_set.insert(*key);
                    let count = self
                        .storage
                        .slot_store_count(*slot, account_info.store_id)
                        .unwrap()
                        - 1;
                    debug!(
                        "store_counts, inserting slot: {}, store id: {}, count: {}",
                        slot, account_info.store_id, count
                    );
                    store_counts.insert(account_info.store_id, (count, key_set));
                }
                true
            });
        }
```
</details>

`slot_count` and `zero_lamport_purges` are then both passed into `calc_delete_dependencies()`. Looking closer at the state of things in the debugger, the following items returns true for account 1.
```
            let no_delete = if account_infos.len() as u64 != *ref_count_from_storage {
```
Namely, `account_infos.len() == 1` (seems correct since Slot 0 and Slot 1 accounts removed) and `*ref_count_from_storage == 2` (seems wrong). `ref_count_from_storage` is set in the previous loop via `self.accounts_index.ref_count_from_storage(&key);`.

From further inspection, I think the issue is that `remove_dead_accounts()` [removes accounts from storage](https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_db.rs#L4437), but as @carllin pointed out in Jeff's PR, [`unref_from_storage()` is called in `finalize_dead_slot_removal()`](https://github.com/solana-labs/solana/pull/15534/files#diff-1090394420d51617f3233275c2b65ed706b35b53b115fe65f82c682af8134a6fR3958). So, the reference count for Account 1 is decremented only once when it has two accounts removed, since only one of those accounts was in a dead slot. This mismatch is present when Slot 2 is being evaluated for clean; the mismatch prevents Slot 2 from being cleaned but is rectified when Slot 0 goes dead. Thus, the second clean makes things ok.

Going to snag lunch, but will start looking at how things could be shifted to fix this after that.

jeffwashington: Wow. Good job! I had dug up my instrumented/debug trace build to dump all this out and figure it out again but I hadn't gotten back to it yet. Your analysis matches approximately what I remembered from before about what was going on. Thanks for digging into it!

jeffwashington: I think this test is fine to put in as-is. That other fix will not be just a test fix - it will be code change.

jeffwashington (approved):

runtime/src/accounts_db.rs
steviez: > I think this test is fine to put in as-is. That other fix will not be just a test fix - it will be code change.

Fair enough - I'm going to submit this PR and made a Github issue to track fixing the `clean()` behavior
https://github.com/solana-labs/solana/issues/17089

carllin: @steviez omg, this triggered a memory. I think this may have been by design: https://github.com/solana-labs/solana/blob/master/runtime/src/bank.rs#L2116-L2119

carllin: ah great find so the crux of the issue here is that when you delete an outdated index entry from slot `S` `clean_accounts_older_than_root`, it doesn't necessarily decrement the refcount for that index entry unless the entire  slot `S` is also marked dead, at which all the refcounts are decremented.

Hmmm I'm trying to recall if there's anywhere important where we rely on the assumption that an individual account's refcount is not decremented until the entire slot is dead...might have to double check some of the code.

carllin: Ah great find! So the crux of the issue here is that when the latest version of an account `A` is a zero-lamport key, it triggers this check in `can_purge_older_entries ()`slot < max_clean_root && slot != newest_root_in_slot_list`. This means it doesn't get added to the reclaims list in `clean_accounts_older_than_root()`, and consequently, the slot `S`, even though all it contains is this zero-lamport key, is not marked dead by the time we take the refcount in the zero-lamport purging loop. Ultimately, this implies the refcount of `A` is not decremented before it hits the `calc_delete_dependencies()` logic.

Then, when the delete of the zero-lamport key `A` happens later after `calc_delete_dependencies()`, it does not cause a cascading deletion of later zero-lamport keys it was blocking from later slots.

PR #17039, opened 2021-05-04
prunes received-cache only once per unique owner's key

behzadnouri:
#### Problem
`prune_received_cache` is redundantly called with possibly duplicate origin keys:
https://github.com/solana-labs/solana/blob/a6a1355b8/core/src/cluster_info.rs#L2426-L2442

#### Summary of Changes
prune received-cache only once per unique owner's key

### DIFF
--- a/core/src/cluster_info.rs
+++ b/core/src/cluster_info.rs
@@ -2423,8 +2423,9 @@ impl ClusterInfo {
         self.stats
             .skip_push_message_shred_version
             .add_relaxed(num_crds_values - num_filtered_crds_values);
-        // Update crds values and obtain updated keys.
-        let updated_labels: Vec<_> = {
+        // Origins' pubkeys of updated crds values.
+        // TODO: Should this also include origins of new crds values?
+        let origins: HashSet<_> = {
             let mut gossip =
                 self.time_gossip_write_lock("process_push", &self.stats.process_push_message);
             let now = timestamp();
@@ -2433,13 +2434,13 @@ impl ClusterInfo {
                 .flat_map(|(from, crds_values)| {
                     gossip.process_push_message(&from, crds_values, now)
                 })
-                .map(|v| v.value.label())
+                .map(|v| v.value.pubkey())
                 .collect()
         };
         // Generate prune messages.
         let prunes = self
             .time_gossip_write_lock("prune_received_cache", &self.stats.prune_received_cache)
-            .prune_received_cache(updated_labels, stakes);
+            .prune_received_cache(origins, stakes);
         let prunes: Vec<(Pubkey /*from*/, Vec<Pubkey> /*origins*/)> = prunes
             .into_iter()
             .flat_map(|(from, prunes)| {
--- a/core/src/crds_gossip_push.rs
+++ b/core/src/crds_gossip_push.rs
@@ -53,7 +53,10 @@ pub struct CrdsGossipPush {
     /// bool indicates it has been pruned.
     /// This cache represents a lagging view of which validators
     /// currently have this node in their `active_set`
-    received_cache: HashMap<Pubkey, HashMap<Pubkey, (bool, u64)>>,
+    received_cache: HashMap<
+        Pubkey, // origin/owner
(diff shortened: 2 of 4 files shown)


behzadnouri:
`cluster_info_stats3::prune_received_cache`:

cluster_info_stats3 prune_received_cache

core/src/crds_gossip_push.rs
```diff
@@ -102,67 +105,58 @@ impl CrdsGossipPush {
...
-        rand::thread_rng().fill(&mut seed[..]);
-        let shuffle = weighted_shuffle(
-            &staked_peers.iter().map(|(_, stake)| *stake).collect_vec(),
-            seed,
-        );
-
+        let shuffled_staked_peers = {
+            let peers: Vec<_> = peers
+                .iter()
+                .filter(|(_, (pruned, _))| !pruned)
```
carllin, line 124: nit: `pruned -> is_pruned`

core/src/crds_gossip_push.rs
```diff
@@ -102,67 +105,58 @@ impl CrdsGossipPush {
     ) -> Vec<Pubkey> {
         let origin_stake = stakes.get(origin).unwrap_or(&0);
         let self_stake = stakes.get(self_pubkey).unwrap_or(&0);
-        let cache = self.received_cache.get(origin);
-        if cache.is_none() {
-            return Vec::new();
-        }
-        let peers = cache.unwrap();
-
+        let peers = match self.received_cache.get_mut(origin) {
```
carllin, line 108: I always have a bit of. trouble remembering what is being tracked in the `received_cache`

iirc, the `received_cache` is more like a `active_push_validators_by_origin`, where for each possible origin `X`, it tells us which set of validators `V`  have pushed messages `M` to us where `M.origin == X`. Could we add a quick comment to `received_cache` to clarify this? Thanks 😃!

core/src/crds_gossip_push.rs
```diff
@@ -102,67 +105,58 @@ impl CrdsGossipPush {
...
+            None => return Vec::default(),
+            Some(peers) => peers,
+        };
         let peer_stake_total: u64 = peers
             .iter()
-            .filter(|v| !(v.1).0)
-            .map(|v| stakes.get(v.0).unwrap_or(&0))
+            .filter(|(_, (pruned, _))| !pruned)
```
carllin, line 114: `nit: pruned -> is_pruned`

carllin: Thanks for cleaning this up, was unreadable before xD

core/src/cluster_info.rs
```diff
@@ -2423,8 +2423,9 @@ impl ClusterInfo {
         self.stats
             .skip_push_message_shred_version
             .add_relaxed(num_crds_values - num_filtered_crds_values);
-        // Update crds values and obtain updated keys.
-        let updated_labels: Vec<_> = {
+        // Origins' pubkeys of updated crds values.
+        // TODO: Should this also include origins of new crds values?
+        let origins: HashSet<_> = {
```
carllin, line 2428: Hmmm, so first I'm assuming:

If a validator `V` pushes us a new crds value with origin `O`, then validator `V` gets added to the `received_set` for `O`.

So now the question is:
If there were earlier push messages with a *different* label, but for the same origin `O`, should this qualify `V` for pruning? Put another way, should the prune set be per origin, or per origin + label.

From my naive understanding, it should be per origin because the push decisions are all made based on the origin of the message, regardless of the label? So it seems like new crds values should also be included here, but it might be good to also double check with @aeyakovenko

carllin (approved):

core/src/cluster_info.rs
behzadnouri: yeah, I agree with you

core/src/crds_gossip_push.rs
behzadnouri: Agree that it is confusing; It has some comments here:
https://github.com/solana-labs/solana/blob/0aa782488/core/src/crds_gossip_push.rs#L52-L56
I also added comments for what the fields are as in above.

PR #17040, opened 2021-05-04
insert accounts in parallel when building initial index

jeffwashington:
#### Problem
large numbers of accounts cause initial index creation to be slow. Secondary indexes are dashmaps. Primary index does have a single mutex. There are other pieces of code that can execute outside the locks that would benefit from paralleism.
#### Summary of Changes
Parallelize the validator startup indexing.
Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -4921,15 +4921,16 @@ impl AccountsDb {
                     let my_total_newly_processed_slots_since_last_report =
                         (index as u64) - my_last_reported_number_of_processed_slots;
                     my_last_reported_number_of_processed_slots = index as u64;
-                    let add = progress.fetch_add(
+                    let previous_total_processed_slots_across_all_threads = progress.fetch_add(
                         my_total_newly_processed_slots_since_last_report,
                         Ordering::Relaxed,
                     );
-                    was_first = was_first || 0 == add;
+                    was_first = was_first || 0 == previous_total_processed_slots_across_all_threads;
                     if was_first {
                         info!(
                             "generating index: {}/{} slots...",
-                            add + my_total_newly_processed_slots_since_last_report,
+                            previous_total_processed_slots_across_all_threads
+                                + my_total_newly_processed_slots_since_last_report,
                             outer_slots_len
                         );
                     }
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -4908,7 +4908,7 @@ impl AccountsDb {
         let mut slots = self.storage.all_slots();
         #[allow(clippy::stable_sort_primitive)]
         slots.sort();
-        let progress = AtomicU64::new(0);
+        let total_processed_slots_across_all_threads = AtomicU64::new(0);
         let outer_slots_len = slots.len();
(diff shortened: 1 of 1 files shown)


jeffwashington:
not sure if slots have to be in order in the slotlist. Tests will hopefully tell us.
Loading accounts on colo machine with mnb snapshot goes from 3 minutes to 2. This is with 2 indices, but not the spl index.

jeffwashington:
Also note this does not solve the out of memory problem.

runtime/src/accounts_db.rs
```diff
@@ -4908,70 +4908,86 @@ impl AccountsDb {
...
-                        };
-                        self.accounts_index.insert_new_if_missing(
-                            *slot,
-                            &pubkey,
-                            &stored_account.account_meta.owner,
-                            &stored_account.data,
-                            &self.account_indexes,
-                            account_info,
-                            &mut _reclaims,
+        let progress = AtomicU64::new(0);
```
carllin, line 4911: nit: `progress -> total_processed_slots_across_all_threads`

runtime/src/accounts_db.rs
```diff
@@ -4908,70 +4908,86 @@ impl AccountsDb {
...
+        let outer_slots_len = slots.len();
+        let chunk_size = (outer_slots_len / 7) + 1; // approximately 400k slots in a snapshot
+        slots.par_chunks(chunk_size).for_each(|slots| {
+            let mut last_log_update = Instant::now();
+            let mut last = 0;
```
carllin, line 4916: nit: `last` -> `this_thread_last_reported_number_of_processed_slots`

runtime/src/accounts_db.rs
```diff
@@ -4908,70 +4908,86 @@ impl AccountsDb {
...
+            let mut was_first = false;
+            for (index, slot) in slots.iter().enumerate() {
+                let now = Instant::now();
+                if now.duration_since(last_log_update).as_secs() >= 2 {
+                    let added = (index as u64) - last;
```
carllin, line 4921: nit `added` -> `newly_processed_slots_since_last_report_count`

runtime/src/accounts_db.rs
```diff
@@ -4908,70 +4908,86 @@ impl AccountsDb {
...
+                    last = index as u64;
+                    let add = progress.fetch_add(added, Ordering::Relaxed);
```
carllin, line 4923: nit: `add` -> `previous_total_processed_slots_across_all_threads`

carllin (approved):
Nice, we had the wrong level of parallelization here: https://github.com/solana-labs/solana/pull/14792/files#diff-1090394420d51617f3233275c2b65ed706b35b53b115fe65f82c682af8134a6fL4163-L4164, this seems more useful 😃

sakridge (approved):

jeffwashington:
> not sure if slots have to be in order in the slotlist. Tests will hopefully tell us.
> Loading accounts on colo machine with mnb snapshot goes from 3 minutes to 2. This is with 2 indices, but not the spl index.

@sakridge and I believe slots do not have to be in order and can be inserted out of order at runtime at least in fork situations. There may be a negligible penalty as we search the slot list in reverse for the latest version of an account. After an epoch, all slot lists will be in approximate slot order as the snapshot-generated entries are cleaned.

PR #17044, opened 2021-05-05
feature: make executor lfu cache discounted

jon-chuang:
#### Problem
Executor cache uses LFU with count maintained for as long as validator does not shut down (potentially all time). But `Executor`s that were popular in past may not be popular at present.

For that reason, I propose a decaying LFU counter with the following behaviour:
- Every epoch, the bank applies a discount factor `\gamma` to the atomicU64. 

- The `\gamma` should be chosen based on our intended half life of the cacheing. I propose that this half life is about 15 epochs or about 30 days. This sets `\gamma` to be 0.95. 

- Every epoch, we apply the following to all counts: `let c = count.get_mut(); c *= 19; c /= 20; drop(c);`

However, I need some help on where to invoke the `apply_lfu_discount` each epoch.

### DIFF
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -256,16 +256,26 @@ pub struct Builtins {
 
 const MAX_CACHED_EXECUTORS: usize = 100; // 10 MB assuming programs are around 100k
 
-/// LFU Cache of executors
+// We set the per-epoch discount factor to be ~0.95
+// This results in a executor usage count half life of about
+// 15 epochs ~ 30 days.
+const DEFAULT_DISCOUNT_NUMERATOR: u64 = 19;
+const DEFAULT_DISCOUNT_DENOMINATOR: u64 = 20;
+
+/// Discounted LFU Cache of executors
 #[derive(Debug)]
 struct CachedExecutors {
     max: usize,
+    discount_numerator: u64,
+    discount_denominator: u64,
     executors: HashMap<Pubkey, (AtomicU64, Arc<dyn Executor>)>,
 }
 impl Default for CachedExecutors {
     fn default() -> Self {
         Self {
             max: MAX_CACHED_EXECUTORS,
+            discount_numerator: DEFAULT_DISCOUNT_NUMERATOR,
+            discount_denominator: DEFAULT_DISCOUNT_DENOMINATOR,
             executors: HashMap::new(),
         }
     }
@@ -292,14 +302,18 @@ impl Clone for CachedExecutors {
         }
         Self {
             max: self.max,
+            discount_numerator: self.discount_numerator,
+            discount_denominator: self.discount_numerator,
             executors,
         }
     }
 }
 impl CachedExecutors {
-    fn new(max: usize) -> Self {
+    fn new(max: usize, discount_numerator: u64, discount_denominator: u64) -> Self {
         Self {
             max,
+            discount_numerator,
+            discount_denominator,
             executors: HashMap::new(),
         }
     }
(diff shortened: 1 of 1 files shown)


jon-chuang:
So I think the compiler is complaining now because I am being more cautious about mutability than I need to be. 

Essentially, I do not need to defensively program as the fact that `Atomic64::get_mut` requires `&mut self` means that we are already locking the `AtomicU64` even from reads while calling `put`. This does mean that `put` is a very expensive operation potentially. So I'm wondering when it's being called.

Waiting for review.

jon-chuang:
Requesting review: @sakridge @mvines

jackcmay:
What issue is this pr trying to solve?  Reduce the memory impact of potentially keeping cached executables around even though they are never used again?  I would like to see evidence that the cache is holding a bunch of stale  before we add changes like these.  I've been doing some mainnet analysis and can take a look at our cache usage to see if we are experiencing this issue.

jon-chuang:
I suppose this PR isn't based on any evidence of an issue, and may ultimately be a case of premature optimisation due to inexperience.

Some of the assumptions may not be correct. For instance, everytime a validator undergoes maintenance and restarts or upgrades their client version, the cache gets rebooted and bpf executables need to be jit recompiled, and thus usage frequency is not quite collected "for all time" as was the worry.

The second assumption which may be false is that programs do in fact go out of fashion after a brief period of popularity. I don't know enough to say if this could be the case in the future, and if for example ethereum experiences this, with fads like crypto kitties. But as long as time of fad ~ time for validator to reboot on average, the cache will reset in due course and crypto kitties won't take up cache space uneccesarily.

Further, I haven't figured out whether one attempts to `put` an executable into the cache every time it's called.

Could I check if an Executor is the entire contract or merely a single instruction? I was guessing the former, due to the quoted size (10mb/executor) and if so, the executor has a way to do dynamic linking or at the very least has an ABI similar to extern, however I don't know where to read up more about it.

jackcmay:
The executor cache is fixed size so new programs added to a full cache kick off the lfu.  Fads like kitties would fall off the end as new programs came in.

jon-chuang:
Hmm, yes, but imagine the following scenario:
cryptokitty fad: 1 month. number of prog invocations during that time: 100_000_000.

Say there are 5 fads a month. 

Then in 1 year, we would have accumulated 60 fads.

If the validator does not restart, programs that are used relatively frequently but don't quite reach the total program invocations of the fads may end up never living in the cache.

In a sense there are two extremes: fads that have limited life, but reach a very high count of program invocations over their lifetime.

Programs that are durable but keep getting evicted from the cache and thus never accrue any usage count.

I just want to guard against this scenario by letting the counts gracefully fall off over time.

jon-chuang:
Hmm, basing it off the slot sounds ok now, when there are a handful of programs, and TPS is 1k, but I'm not sure if it's a good idea going forward, when there could be 100s of programs and TPS is 30K say. If > 100 Executors are invoked per slot, this could lead to lots of churn.

One can always expand the size of the cache in that case, but one should probably avoid that and try to make sure the cache-eviction algorithm is efficient.

I what we could do is just make the cache evict those programs which haven't been used for a long time, even if they have a high count. We keep an additional `last_epoch_and_slot_used` field as you suggest, and evict all Executors if it exceeds that threshold. I think 2 epochs is probably good enough for our purposes so we can drop the slot field.

And really sorry but I can't seem to figure out for sure - does each executor correspond to a single instruction, or does it contain multiple instructions (the entire program)?

Issue #17000, opened 2021-05-03
RPC: Blocks go missing 


linuskendall:
#### Problem

We've noticed on several occasions that blocks goes missing from RPC.  This typically happens when the RPC node becomes overwhelemd (not necessarily crashes, but becomes unresponsive for sometime) or when there's an abrupt reboot or something like this.

In these cases, `minimumLedgerSlot` will report an old value, but there will be holes in the confirmed blocks between `minimumLedgerSlot` and the current slot. The RPC service will report that "Slot xxx was skipped, or missing due to ledger jump to recent snapshot". However, this occurs even when we haven't pulled a snapshot. Also if pulling snapshot would we not presume that `minimumLedgerSlot` would get updated? 

We've seen between 300 and 900 blocks missing typically. Most common (anecdotally) seems to be around300 blocks.

Sample output from a node that was missing blocks:

```
$ curl http://hk1.rpcpool.com:8899 -X POST  -H "Content-Type: application/json" -d '{"jsonrpc": "2.0","id":1,"method":"getConfirmedBlock","params":[75140922, {"encoding": "json","transactionDetails":"full","rewards":false}]}'
{"jsonrpc":"2.0","error":{"code":-32007,"message":"Slot 75140922 was skipped, or missing due to ledger jump to recent snapshot"},"id":1}
```

However, the slot clearly exists in the ledger directory as solana ledger tool can access it:
```
solana@hk1:~$ solana-ledger-tool -l /solana/ledger slot 75140922 >75140922.bak.slot
[2021-04-25T19:28:45.836799091Z INFO  solana_ledger_tool] solana-ledger-tool 1.6.6 (src:fe775a97; feat:3714435735)
[2021-04-25T19:28:45.836851189Z INFO  solana_ledger::blockstore] Maximum open file descriptors: 750000
[2021-04-25T19:28:45.836867639Z INFO  solana_ledger::blockstore] Opening database at "/solana/ledger/rocksdb"
[2021-04-25T19:28:45.862167110Z WARN  solana_ledger::blockstore_db] Error when opening as primary: IO error: While lock file: /solana/ledger/rocksdb/LOCK: Resource temporarily unavailable
[2021-04-25T19:28:45.862181937Z WARN  solana_ledger::blockstore_db] Trying as secondary at : "/solana/ledger/rocksdb/solana-secondary"
[...]

PR #17045, opened 2021-05-05
Add ledger-tool for restoring roots to the Roots CF

CriesofCarrots:
#### Problem
Sometimes a chunk of blocks appear to be "missing" from the ledger because they are no longer recorded as roots in the blockstore Root CF. The cause of this Roots-CF data issue is not clear.

However, we can restore the data by walking backward from a known root through its ancestors.

#### Summary of Changes
- Add `solana-ledger-tool repair-roots` that use an AncestorIterator from a provided (and verified) root and rewriting missing entries to the cf::Root

Stopgap for #17000

### DIFF
--- a/ledger-tool/src/main.rs
+++ b/ledger-tool/src/main.rs
@@ -1337,6 +1337,25 @@ fn main() {
             )
         )
         .subcommand(
+            SubCommand::with_name("repair-roots")
+                .about("Traverses the AncestorIterator backward from a last known root \
+                        to restore missing roots to the Root column")
+                .arg(
+                    Arg::with_name("start_root")
+                        .long("before")
+                        .value_name("NUM")
+                        .takes_value(true)
+                        .help("First good root after the range to repair")
+                )
+                .arg(
+                    Arg::with_name("end_root")
+                        .long("until")
+                        .value_name("NUM")
+                        .takes_value(true)
+                        .help("Last slot to check for root repair")
+                )
+        )
+        .subcommand(
             SubCommand::with_name("analyze-storage")
                 .about("Output statistics in JSON format about \
                         all column families in the ledger rocksdb")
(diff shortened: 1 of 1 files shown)


ledger-tool/src/main.rs
```diff
@@ -2796,6 +2815,41 @@ fn main() {
...
+                blockstore.lowest_slot()
+            };
+            assert!(start_root > end_root);
+            assert!(blockstore.is_root(start_root));
+            let ancestor_iterator =
+                AncestorIterator::new(start_root, &blockstore).take_while(|&slot| slot >= end_root);
+            let roots_to_fix: Vec<_> = ancestor_iterator
+                .filter(|slot| !blockstore.is_root(*slot))
+                .collect();
+            if !roots_to_fix.is_empty() {
```
ryoqun, line 2841: I usually like to print what actually done rather verbose for one-off yet dangerous ops like this:

```rust
eprintln!("{} slots to be rooted", roots_to_fix.len());
for linated_roots in roots_to_fix.chunk(100) {
    eprintln!("{:?}", lineated_roots);
}
```

So, we could be acertain what went wrong from 2nd disaster recovery ops....

ledger-tool/src/main.rs
```diff
@@ -2796,6 +2815,41 @@ fn main() {
...
+                blockstore.set_roots(&roots_to_fix).unwrap_or_else(|err| {
+                    eprintln!("Unable to set roots {:?}: {}", roots_to_fix, err);
+                    exit(1);
+                });
+            } else {
+                println!(
```
ryoqun, line 2847: nice. I like placing exceptional clause at `else` too even at the cost of extra `!` for the `if` condition. :)

ledger-tool/src/main.rs
```diff
@@ -2796,6 +2815,41 @@ fn main() {
...
+            };
+            assert!(start_root > end_root);
+            assert!(blockstore.is_root(start_root));
+            let ancestor_iterator =
+                AncestorIterator::new(start_root, &blockstore).take_while(|&slot| slot >= end_root);
+            let roots_to_fix: Vec<_> = ancestor_iterator
+                .filter(|slot| !blockstore.is_root(*slot))
+                .collect();
+            if !roots_to_fix.is_empty() {
+                blockstore.set_roots(&roots_to_fix).unwrap_or_else(|err| {
```
ryoqun, line 2842: What about sensible-yet-override-able limit to avoid too many slots to be rooted like `--repair-limit{=2000}` (default value is picked by me from the reported observation)?

ledger-tool/src/main.rs
CriesofCarrots: Hehe, well I originally wrote it with no `!`, ie. `if roots_to_fix.len() > 0`, but clippy prefers `is_empty`

ledger-tool/src/main.rs
CriesofCarrots: Done in https://github.com/solana-labs/solana/pull/17045/commits/bfc2a068cf019f2bbe4f7fd6be1970651e6495a3

CriesofCarrots:
@ryoqun hopefully this looks good to you now; post-merge amendments welcome :)

ryoqun (commented):
> @ryoqun hopefully this looks good to you now; post-merge amendments welcome :)

no problem. solid work. Sorry for late reply...

PR #17048, opened 2021-05-05
Remove bloat from secondary indexes

carllin:
#### Problem
Secondary indexes are bloated by slot information

#### Summary of Changes
Remove a lot of the fine grained slot tracking, and instead only purge secondary indexes when an entire pubkey is marked dead in `handle_dead_keys()`

This seems to reduce memory usage of secondary indexes by about 2/3, from 60GB to 20GB for the program id index for instance.

cc @brooksprumo @steviez 

Fixes #

### DIFF
--- a/runtime/src/accounts_index.rs
+++ b/runtime/src/accounts_index.rs
@@ -878,10 +878,9 @@ impl<T: 'static + Clone + IsCached + ZeroLamport> AccountsIndex<T> {
                     if index_entry.get().slot_list.read().unwrap().is_empty() {
                         index_entry.remove();
 
-                        // Note passing `None` to remove all the entries for this key
-                        // is only safe because we have the lock for this key's entry
-                        // in the AccountsIndex, so no other thread is also updating
-                        // the index
+                        // Note it's only safe to remove all the entries for this key
+                        // because we have the lock for this key's entry in the AccountsIndex,
+                        // so no other thread is also updating the index
                         self.purge_secondary_indexes_by_inner_key(key, account_indexes);
                     }
                 }
--- a/runtime/src/accounts_index.rs
+++ b/runtime/src/accounts_index.rs
@@ -2728,6 +2728,7 @@ pub mod tests {
         secondary_index: &SecondaryIndex<SecondaryIndexEntryType>,
         key_start: usize,
         key_end: usize,
+        secondary_indexes: &AccountSecondaryIndexes,
     ) {
         // No roots, should be no reclaims
         let slots = vec![1, 2, 5, 9];
@@ -2745,7 +2746,7 @@ pub mod tests {
                 // Make sure these accounts are added to secondary index
                 &inline_spl_token_v2_0::id(),
                 &account_data,
-                account_index,
+                secondary_indexes,
                 true,
                 &mut vec![],
             );
@@ -2765,7 +2766,7 @@ pub mod tests {
                 .read()
                 .unwrap()
                 .len(),
-            slots.len()
+            1
         );
 
         index.purge_exact(
@@ -2774,31 +2775,34 @@ pub mod tests {
             &mut vec![],
         );
 
+        index.handle_dead_keys(&[&account_key], secondary_indexes);
         assert!(secondary_index.index.is_empty());
         assert!(secondary_index.reverse_index.is_empty());
     }
 
     #[test]
     fn test_purge_exact_dashmap_secondary_index() {
-        let (key_start, key_end, _account_index) = create_dashmap_secondary_index_state();
+        let (key_start, key_end, secondary_indexes) = create_dashmap_secondary_index_state();
         let index = AccountsIndex::<bool>::default();
         run_test_purge_exact_secondary_index(
             &index,
             &index.spl_token_mint_index,
             key_start,
             key_end,
+            &secondary_indexes,
         );
     }
 
     #[test]
     fn test_purge_exact_rwlock_secondary_index() {
-        let (key_start, key_end, _account_index) = create_rwlock_secondary_index_state();
+        let (key_start, key_end, secondary_indexes) = create_rwlock_secondary_index_state();
         let index = AccountsIndex::<bool>::default();
         run_test_purge_exact_secondary_index(
             &index,
             &index.spl_token_owner_index,
             key_start,
             key_end,
+            &secondary_indexes,
         );
     }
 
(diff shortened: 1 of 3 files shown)


jeffwashington:
It looks like a sound idea, but I don't know enough of the usage yet. I can volunteer to fix tests and get this pr ready if @ryoqun, @sakridge, @t-nelson decide this seems reasonable.

runtime/src/accounts_index.rs
```diff
@@ -2989,61 +2909,66 @@ pub mod tests {
...
-        // Check correctness
-        check_secondary_index_unique(&secondary_index, slot, &secondary_key2, &account_key);
-        assert!(secondary_index.get(&secondary_key1).is_empty());
-        assert_eq!(secondary_index.get(&secondary_key2), vec![account_key]);
+        // Both pubkeys will now be present in the index
+        check_secondary_index_mapping_correct(
+            &secondary_index,
+            &[secondary_key1, secondary_key2],
+            &account_key,
+        );
```
carllin, line 2933: This section of the test highlights the critical tradeoff between this change and the existing change.

In this new change, we don't have slot-level granularity tracking in the SecondaryIndex anymore. 

So if for instance, if a pubkey `P` belonged to  program id `A` in slot 1, then program id `B` in slot 2, the mapping looks like `{A: P, B: P}`. If slot 2 is then rooted, making the update in slot 1 outdated this secondary mapping will remain unchanged, whereas the current code understands that because the mapping `{A: P}` was only relevant in slot 1, it will delete that mapping.

Now we only delete when  pubkey `A` is marked entirely dead. Hopefully this should be ok because a pubkey belonging to multiple secondary index keys should be rare...

runtime/src/secondary_index.rs
```diff
@@ -1,157 +1,95 @@
...
-    }
-
-    fn get<T>(&self, key: &Pubkey, f: &dyn Fn(Option<&RwLock<HashSet<Slot>>>) -> T) -> T {
-        let slot_set = self.pubkey_to_slot_set.get(key);
-
-        f(slot_set.as_ref().map(|entry_ref| entry_ref.value()))
+    fn insert_if_not_exists(&self, key: &Pubkey) {
+        self.account_keys
+            .get(key)
+            .unwrap_or_else(|| self.account_keys.entry(*key).or_default().downgrade());
```
sakridge, line 29: Maybe a stupid question, why do we need this vs. just `self.account_keys.entry(*key).or_default()`

runtime/src/accounts_index.rs
sakridge: Ok, does that then just cause an extra entry for the key which is ignored when the index scan operates?

carllin: So the scan will always find the latest account for that extra key, then it will be the responsibility of RPC to do the additional filter.

Heh, I was going to say yes because when I checked the RPC call for the `ProgramId` earlier it has this filter https://github.com/solana-labs/solana/blob/master/core/src/rpc.rs#L1714: `account.owner() == program_id `.

But it seems like the token owner and mint indexes are missing them: https://github.com/solana-labs/solana/blob/master/core/src/rpc.rs#L1788-L1794. @CriesofCarrots would it be safe to add that extra check in those places, or is that done somewhere earlier that I'm missing?

CriesofCarrots: @carllin , this extra check? https://github.com/solana-labs/solana/blob/ddfbae260fa23ec3e0fcbcb70b4e1a48eccc0fba/core/src/rpc.rs#L1789

CriesofCarrots: @carllin , I'm confused. That 2nd link you included contains the re-application of the mint-key filter here https://github.com/solana-labs/solana/blob/ddfbae260fa23ec3e0fcbcb70b4e1a48eccc0fba/core/src/rpc.rs#L1790-L1793
Am I misunderstanding the check you're referring to?

CriesofCarrots: > a pubkey belonging to multiple secondary index keys should be rare

The token-owner case is one that might be more common than the others, since it's easy to set a new owner on an spl-token account, and doesn't require emptying the entire account like the others do. That said, the extra rpc checks should prevent any erroneous data here; I'm happy to go with it until such time as this seems to be a problem.

runtime/src/secondary_index.rs
carllin: Heh it's only necessary because I didn't clean this up as much as I could have 😃.

`get` returns a `Ref` and `.entry()` returns a `RefMut`, and the `unwrap_or_else` enforces the results must be of the same type.

Changed to check for `is_none()` instead here: https://github.com/solana-labs/solana/pull/17048/commits/1c43c59361923354816b28af61ef81d1c35a2911

runtime/src/accounts_index.rs
carllin: Ah it's addressed by this check here which adds to the filter here: https://github.com/solana-labs/solana/blob/master/core/src/rpc.rs#L1778-L1782

carllin:
hmmmm noticed memory seems to leak with this change

I've been running this with the new metrics to track if there are some accounts that belong to multiple secondary keys that aren't being cleaned up, but the `num_inner_keys` and `num_reverse_index_keys` seem to be pretty close with each other which suggests close to a 1-1 mapping between the secondary key (mint index) and the account keys, despite memory usage going up:

Screen Shot 2021-05-08 at 5 52 28 PM

runtime/src/secondary_index.rs
```diff
@@ -160,163 +124,102 @@ impl<SecondaryIndexEntryType: SecondaryIndexEntry + Default + Sync + Send>
...
-        if is_key_empty {
+        if is_outer_key_empty {
             // Other threads may have interleaved writes to this `key`,
             // so double-check again for its emptiness
-            if let Occupied(key_entry) = self.index.entry(*key) {
+            if let Occupied(key_entry) = self.index.entry(*outer_key) {
                 if key_entry.get().is_empty() {
                     key_entry.remove();
                 }
             }
```
carllin, line 194: ah shoot, also I don't think this check for empty -> deletion triggered from  is safe because we're no longer accounting for slots.

For instance, if:

1) clean calls `handle_dead_keys(pubkey)` where the mint for `pubkey` is `M`,
2) a new block `B` inserts `pubkey` with mint also `M` into `AccountsIndex`, then calls `update_secondary_indexes()`: https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_index.rs#L1153, the new insert here only checks `M` exists in the secondary index set (doesn't add a unique entry for block `B` if `M` already exists)
3) Clean now finishes removing `pubkey` from the secondary mint index `M` and comes into this check, and removes this key entirely, even though block `B` is keeping this mapping alive.

I think if we pass the atomic `refcount` from the AccountsIndex here, it will solve this issue.

carllin, line 194: For future reference, the reason why this below situation can't happen:

1) clean calls `handle_dead_keys(pubkey)` where the mint for `pubkey` is `M`,
2) a new block `B` inserts `pubkey` with mint also `M` into `AccountsIndex`, then calls `update_secondary_indexes()`: https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_index.rs#L1153, the new insert here only checks `M` exists in the secondary index set (doesn't add a unique entry for block `B` if `M` already exists)
3) Clean now finishes removing `pubkey` from the secondary mint index `M` and comes into this check, and removes this key entirely, even though block `B` is keeping this mapping alive.

is because we hold the accounts index lock while calling `purge_secondary_indexes_by_inner_key` in `handle_dead_keys()`: https://github.com/solana-labs/solana/blob/a1df57a4eaa61ecb7c1fc1ddebbcd9506efc9206/runtime/src/accounts_index.rs#L825-L838

jeffwashington:
@carllin, I rebased to master and fixed the merge conflicts that came up because of a test fix and refactoring change I put in master this morning.

PR #17057, opened 2021-05-05
Don't recognize temp snapshots as possible snapshots to open

sakridge:
#### Problem

Validator still recognizes temp snapshot files as valid to open. When there is an error encountered with the download, it will try to open this broken archive and panic in a confusing way.

#### Summary of Changes

Don't recognize anything but what starts with `snapshot-<hash>-<slot>.tar.*` as a snapshot to look at.

Fixes #


behzadnouri (approved):
thanks for fixing this

mvines (approved):
dang, thanks

PR #17059, opened 2021-05-05
add comment

jeffwashington:
#### Problem
I was thinking I could get rid of a slower btree and replace it with something faster.
#### Summary of Changes
Comment why we need a btree.
Fixes #


carllin (approved):
Ha yeah it's just for the write version, a bit subtle. Thanks!

PR #16346, opened 2021-04-04
Apple M1 compatibility

cdrappi:
#### Problem
Solana does not build on computers with Mac M1 chips

#### Summary of Changes
- Upgrade `cc` crate
- Explicitly force `ring` to version `0.16.17`: It's required by [rustls](https://crates.io/crates/rustls/0.19.0/dependencies) which has some wiggle room on ring versions
- Remove memory profiling with `jemallocator`, now defaulting to the system allocator
- Exclude `target_os = "macOS"` from AVX check

Fixes #16072 

#### Current state
**Good**: It builds, runs (with `./run.sh`), and `cargo +nightly bench` works

**Bad**: `cargo test` and `cargo test --release` both fail different tests. After having run these many times each and getting a feel for their behavior, I've gotten similar but not exactly the same failures. I'd love to get some color on what you expect from these runs

##### `cargo test`
- `test_kill_heaviest_partition` fails every time with: `'assertion failed: loop_start.elapsed() < loop_timeout', local-cluster/src/cluster_tests.rs:284:9`
- `test_kill_partition_switch_threshold_no_progress` fails sometimes with a `'get_slot for 6iXMbNNczChSGvo4EAThk8udgApPK9Sk8vcsc1w4Nq3 failed', local-cluster/src/cluster_tests.rs:332:37`
- `test_future_tower_master_only` has failed but more rarely

##### `cargo test --release`
- `test_push_votes_with_tower` fails all the time, on either `assert_eq!(vote_slots.len(), MAX_LOCKOUT_HISTORY);` (with many different `vote_slots` lengths) or `assert!(vote_slot != 23);`. _From knowing almost nothing about Solana, I'm most concerned with the last type of failure case here_

PR #17070, opened 2021-05-06
crate-features: cc cargo bump: required for Apple Silicon support on solana_client

nixxholas:
https://github.com/ctz/sct.rs/pull/7/files
https://github.com/briansmith/ring/issues/1063

Signed-off-by: nixxholas <nixholas@icloud.com>

#### Problem

Adopt support for Apple Silicon on 1.6.x.

#### Summary of Changes

Simply bumping the cc dependency would resolve the issue.


mvines:
This should be simple, but it causes other dependencies to blow up :-/

nixxholas:
> This should be simple, but it causes other dependencies to blow up :-/

Actually, there are many more dependencies required if we scope beyond solana-client. num_cpus should be referenced from the independent num_cpus instead of via cc.

nixxholas:
Checking against the entire Solana suite, `reed-solomon-erasure` has a breaking dependency for arm as of now. I will carry on with develop on x86 for now until an update comes out.

https://github.com/darrenldl/reed-solomon-erasure

https://stackoverflow.com/questions/66661891/using-march-native-flag-with-clang-on-apple-m1
https://developer.apple.com/forums/thread/672654

nixxholas:
Irrelevant due to: https://github.com/solana-labs/solana/pull/16346

PR #16498, opened 2021-04-12
Add Keccak256 syscall and sdk support

jackcmay:
#### Problem

No support for Keccak hashig

#### Summary of Changes

Add Keccak256 hash support and a bpf syscall for it

Fixes #

PR #17082, opened 2021-05-06
Add BigNumber Type and syscalls for BPF programs

FrankC01:
@jackcmay  - As per your comments... for review...

#### Problem
BigNum operations are not available in vanilla Solana for BPF programs. For big and prime numbers it is desirable to expose these capabilities to BPF programs.

In addition, hashing to a prime number is valuable as well. 

#### Summary of Changes

- Added budget costing for use of BigNumber syscalls as well as hashing to prime number. Can be reviewed and modified.
- Added BigNumber operational syscalls and unit tests
- Added BigNumber type and implementation. When used in BPF programs, passing a BigNumb back and forth to the syscalls does so via Box raw pointers

Fixes #


FrankC01:
@jackcmay Second commit should address the sanity failure

jackcmay:
@joncinque @jstarry I think you guys have been using big numbers recently?  If so, do you see a syscall for these types of things as a good solution?

FrankC01:
@jackcmay From a PR completeness standpoint, is there anything else I should be adding to this?

programs/bpf_loader/Cargo.toml
```diff
@@ -21,6 +21,9 @@ solana-runtime = { path = "../../runtime", version = "=1.7.0" }
 solana-sdk = { path = "../../sdk", version = "=1.7.0" }
 solana_rbpf = "=0.2.8"
 thiserror = "1.0"
+hkdf = "0.11.0"
```
jackcmay, line 24: nit: nice to keep these in alphabetical order

programs/bpf_loader/src/syscalls.rs
```diff
@@ -1,5 +1,8 @@
 use crate::{alloc, BpfError};
 use alloc::Alloc;
+use blake3::traits::digest::Digest;
```
jackcmay, line 3: Are you using `cargo fmt` on these changes?

sdk/program/src/bignum.rs
```diff
@@ -0,0 +1,704 @@
...
+
+use borsh::{BorshDeserialize, BorshSchema, BorshSerialize};
+use std::fmt;
+
+#[repr(transparent)]
+#[derive(
+    BorshSerialize,
+    BorshDeserialize,
+    BorshSchema,
+    // Clone,
```
jackcmay, line 11: No?

sdk/program/src/bignum.rs
```diff
@@ -0,0 +1,704 @@
...
+    PartialOrd,
+    Hash,
+    AbiExample,
+)]
+
+pub struct BigNumber(u64);
+
+// Syscall interfaces
+#[cfg(target_arch = "bpf")]
+extern "C" {
```
jackcmay, line 25: Stick these in their representative functions below

sdk/program/src/bignum.rs
```diff
@@ -0,0 +1,704 @@
...
+        }
+        #[cfg(target_arch = "bpf")]
+        {
+            write!(f, "{}", 0)
+        }
+    }
+}
+
+impl AsRef<u64> for BigNumber {
+    fn as_ref(&self) -> &u64 {
```
jackcmay, line 620: This can't fail?

sdk/program/src/bignum.rs
```diff
@@ -0,0 +1,704 @@
...
+}
+
+impl From<u64> for BigNumber {
+    fn from(ptr: u64) -> Self {
+        Self(ptr)
+    }
+}
+
+/// Drop - removes the underlying BigNum
+impl Drop for BigNumber {
```
jackcmay, line 632: This means that if someone hacks thier program they could force a leaked bignum in the runtime?

sdk/program/src/bignum.rs
```diff
@@ -0,0 +1,704 @@
...
+                digest.as_mut_ptr() as *mut _ as *mut u8,
+                &mut digest_len as *mut _ as *mut u64,
+            );
+        };
+        digest[0..digest_len].to_vec()
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
jackcmay, line 679: Would like to see more comprehensive testing.  Also would need testing from within a bpf program, check out the sha program in this or for an example: https://github.com/solana-labs/solana/pull/16498

programs/bpf_loader/src/syscalls.rs
```diff
@@ -934,901 +1150,1892 @@ impl<'a> SyscallObject<BpfError> for SyscallSha256<'a> {
...
+    ($input:expr) => {
+        (BIGNUM_WORDCOST * ($input / BIGNUM_WORDCOST_DIVISOR).ceil()) as u64
+    };
 }
 
-/// Get a Clock sysvar
-struct SyscallGetClockSysvar<'a> {
-    invoke_context: Rc<RefCell<&'a mut dyn InvokeContext>>,
+/// BIGNUM sol_bignum_new
+pub struct SyscallBigNumNew<'a> {
```
jackcmay, line 1165: why pub?

programs/bpf_loader/src/syscalls.rs
```diff
@@ -934,901 +1150,1892 @@ impl<'a> SyscallObject<BpfError> for SyscallSha256<'a> {
     }
 }
 
-fn get_sysvar<T: std::fmt::Debug + Sysvar + SysvarId>(
```
jackcmay, line 937: Not sure why github is breaking the diff up like this but it's making it very hard to read

programs/bpf_loader/src/syscalls.rs
```diff
@@ -934,901 +1150,1892 @@ impl<'a> SyscallObject<BpfError> for SyscallSha256<'a> {
...
+        let big_number = question_mark!(
+            translate_type_mut::<u64>(memory_mapping, bn_addr, self.loader_id, true),
+            result
         );
+        let bbox = Box::new(BigNum::from_u32(u32_val as u32).unwrap());
+        let bytes = bbox.num_bytes() as f64;
+        let rwptr = Box::into_raw(bbox);
+        let bignum_ptr = rwptr as u64;
+        *big_number = bignum_ptr;
+        question_mark!(
```
jackcmay, line 1219: Should consume first so that the call fails before doing any work if the caller doesn't have enough units left

programs/bpf_loader/src/syscalls.rs
```diff
@@ -934,901 +1150,1892 @@ impl<'a> SyscallObject<BpfError> for SyscallSha256<'a> {
...
+            translate_type_mut::<u64>(memory_mapping, bn_addr, self.loader_id, true),
+            result
+        );
+        let byte_slice = question_mark!(
+            translate_slice::<u8>(memory_mapping, bytes_addr, bytes_len, self.loader_id, true),
+            result
+        );
+        let bytes: f64 = byte_slice.len() as f64;
+        let bbox = Box::new(BigNum::from_slice(byte_slice).unwrap());
+        let rwptr = Box::into_raw(bbox);
```
jackcmay, line 1253: This screems memory leak, is it possible to create a container in program space (above the syscall) and just have the syscall fill in the result?

jackcmay: Maybe pass back up a byte array and then call `from_slice()`?

sdk/src/process_instruction.rs
```diff
@@ -188,6 +226,25 @@ impl BpfComputeBudget {
             max_cpi_instruction_size: 1280, // IPv6 Min MTU size
             cpi_bytes_per_unit: 250,        // ~50MB at 200,000 units
             sysvar_base_cost: 100,
+            bignum_new_base_cost: 100,
```
jackcmay, line 229: I get the sense that these costs are too small.  By chance have you done any perf measurements of the big-num calls on native?

programs/bpf_loader/src/syscalls.rs
FrankC01: Moved consumes to early as possible depending on when size is available to calculate byte costs

programs/bpf_loader/src/syscalls.rs
FrankC01: May need clarity here.

programs/bpf_loader/Cargo.toml
FrankC01: Fixed in 03af7519e1ca7a29ed2ee25c2cd6b31e0d4e99fd

programs/bpf_loader/src/syscalls.rs
FrankC01: Fixed in 03af7519e1ca7a29ed2ee25c2cd6b31e0d4e99fd

sdk/program/src/bignum.rs
FrankC01: Fixed in 03af7519e1ca7a29ed2ee25c2cd6b31e0d4e99fd

sdk/program/src/bignum.rs
```diff
@@ -0,0 +1,704 @@
...
+}
+/// Clone - creates a new BigNumber replicating itself
+impl Clone for BigNumber {
+    fn clone(&self) -> Self {
+        BigNumber::from_bytes(self.to_bytes().as_slice())
+    }
+}
+
+/// Returns an array of bytes (big endian) of self
+pub fn blake3_digest(data: &[u8]) -> Vec<u8> {
```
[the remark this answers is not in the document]
FrankC01: @arthurgreef - Will review

programs/bpf_loader/src/syscalls.rs
FrankC01: Fixed in 03af7519e1ca7a29ed2ee25c2cd6b31e0d4e99fd

sdk/program/src/bignum.rs
```diff
@@ -0,0 +1,777 @@
...
+                    nonce as *const _ as *const u64,
+                    &mut bignum_ptr as *mut _ as *mut u64,
+                );
+            }
+            Self(bignum_ptr)
+        }
+    }
+
+    /// Hash to Prime
+    pub fn hash_to_prime(u: &BigNumber, a: &BigNumber, z: &BigNumber, nonce: &[u8]) -> Self {
```
arthurgreef, line 569: Consider moving hash operation to a hash.rs module.

sdk/program/src/bignum.rs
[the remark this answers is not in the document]
arthurgreef: We should probably create a blake3.rs module and submit as a separate PR.

programs/bpf_loader/src/syscalls.rs
```diff
@@ -1144,765 +1382,1727 @@ impl<'a> SyscallObject<BpfError> for SyscallKeccak256<'a> {
...
+        // hash_generator
+        let mut transcript = u_bn.as_ref().to_vec();
+        transcript.append(&mut a_bn.as_ref().to_vec());
+        // transcript.extend_from_slice(nonce_ptr.as_ref());
+        transcript.extend_from_slice(nonce_ptr);
+        // println!("t:{:?}",transcript);
+        let length = n_bn.num_bytes();
+        // println!("len:{}",length);
+        let h = Hkdf::<blake3::Hasher>::new(
+            Some(b"SST_SALT_HASH_TO_GENERATOR_"),
```
arthurgreef, line 1986: The salt needs to be an input argument to the hash operation.

sdk/program/src/bignum.rs
FrankC01: Added BPF program testing in #2b1f3e1f6ff9935bada2404e4c1c40064a8eeb71

jackcmay:
@FrankC01 Looks like CI is coughing on extra whitespace

programs/bpf_loader/src/syscalls.rs
jackcmay: The model here looks like it is allocating a bignum type in the syscall and then passing back a raw pointer, which in program space is converted into a bignum type.  This means the allocation is occurring in the runtime and relying on the program to trigger the drop.  A malicious program could skip the drop resulting in a memory leak in the runtime.

sdk/program/Cargo.toml
```diff
@@ -37,6 +37,9 @@ blake3 = "0.3.7"
 curve25519-dalek = "2.1.0"
 rand = "0.7.0"
 solana-logger = { path = "../../logger", version = "=1.7.0" }
+hkdf = "0.11.0"
```
jackcmay, line 40: nit: alphabetical

sdk/program/src/bignum.rs
jackcmay: Why is clone commented out?  Do you not want to support the default implementation of `clone`?  If not, remove this line.

programs/bpf_loader/src/syscalls.rs
jackcmay: Could you use `from_bytes` and `to_bytes` to pass a byte array between the two.  That does lead to local allocations on both ends which isn't ideal but better then a possible memory leak.

jackcmay:
Tweaking the compute costs to get one particular program to run is probably not the best generic assessment of what the compute cost of these operations should be.  The compute cost should be based on the amount of time these operations take compared to the time expected to consume the full compute budget.  We don't yet have a good conversion from microseconds to compute units, we are working on that at the moment.

sdk/program/Cargo.toml
FrankC01: Fixed in 01ed0b9d229749506e0065fed170cac1310741b1

programs/bpf_loader/src/syscalls.rs
FrankC01: @jackcmay - Wanting to grok the suggestion but haven't been working in rust for too long so bear with me...

1. Are you suggesting that instead of boxing, we pass arrays back and forth to syscalls and instantiate the openSSL BigNum in the syscall to perform the operations?
2. As the size of the arrays are variable, is there a way to pass mutable pointers to Vec<u8>, `translate` them to Vec<u8> in the syscall and push results? I'm having a darned time trying to figure this out.

Sorry for the n00b questions...

sdk/program/src/bignum.rs
```diff
@@ -0,0 +1,777 @@
...
+                    &self.0 as *const _ as *const u64,
+                    &exponent.0 as *const _ as *const u64,
+                    &modulus.0 as *const _ as *const u64,
+                );
+            }
+            Self(*bignum_ptr)
+        }
+    }
+    /// Hashed Generator
+    pub fn hashed_generator(&self, a: &BigNumber, n: &BigNumber, nonce: &[u8]) -> Self {
```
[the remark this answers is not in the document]
FrankC01: Removed `hashed_generator` from PR... see 01ed0b9d229749506e0065fed170cac1310741b1

sdk/program/src/bignum.rs
FrankC01: Removed `hash_to_prime` from PR... see 01ed0b9d229749506e0065fed170cac1310741b1

programs/bpf_loader/src/syscalls.rs
FrankC01: Removed `blake3 hash` from PR... see 01ed0b9d229749506e0065fed170cac1310741b1

programs/bpf_loader/src/syscalls.rs
FrankC01: @jackcmay So, remember I am a Rust n00b:
1. Instead of boxing and raw-pointers, send and receive byte arrays to/from syscalls and in the syscalls instantiate the BigNum's?
2. Is there a way to pass in a Vec<u8> and `translate` it to type in a syscall? I've had trouble and can't seem to make it work.
3. The reason for the Vec is that BigNum are ultimately variable length and can get quite large so knowing fixed buffers a priori would be daunting.

sdk/program/src/bignum.rs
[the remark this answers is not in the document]
FrankC01: Removed `hashed_generator, hash_to_prime and blake3_hash` from PR.... see 01ed0b9d229749506e0065fed170cac1310741b1

FrankC01:
Found defect in `to_bytes()` that can overflow buffer. Fix requires a new syscall `sol_bignum_size_in_bytes()` to first get the expected return size bytes buffer required to populate.

FrankC01:
@jackcmay Completed the transition out of using the Box'ing of system size BigNum. Take a look and let me know, The build is failing outside of anything I control I believe.

jackcmay:
@FrankC01 Does this test pass for you locally:
```
thread 'test_program_bpf_finalize' panicked at 'assertion failed: `(left == right)`
left: `InstructionError(0, MissingAccount)`,
right: `InstructionError(0, ProgramFailedToComplete)`', tests/programs.rs:2427:5
```

jackcmay:
Hmm, I also don't see that error when running the exact test that CI is running :-(

jackcmay:
> The solang compiler https://github.com/hyperledger-labs/solang could possibly use bignum operations if they are efficient enough for math of integer types > 64. However there are some issues with that.
> 
>     * Converting to/from BigNum is expensive, and needs to be done often. I have my doubts how well this will work out IRL.
> 
>     * Solang needs BigNum operations with overflow checks for BigNum of N bits width. This PR doesn't offer that.
> 
>     * The only relevant instruction not available in BPF is 64 bit mul with 128 bit result (imulq with result $rax/$rdax on x86). If BPF had that instruction then I don't know if this implementation is more efficient than a native one.

@seanyoung is the overflow checks something that can be added later or must it be part of these initial syscalls to meet Solang's requirements?

FrankC01:
@jackcmay It seems like a different error now (in coverage)... other PRs seem to have issues as well. Are any actually getting through?

arthurgreef:
@jackcmay The syscalls in this pull request do not provide fixed width integer arithmetic. For this you would need to use something like unit provided by parity. https://crates.io/crates/uint. These syscalls rely on OpenSSL to catch underflow and overflow errors.

jackcmay:
@FrankC01 Are you rebased up to the latest master?

FrankC01:
> @FrankC01 Are you rebased up to the latest master?

Before each checkin i do
`git fetch upstream`
`git merge upstream/master'

PR #17086, opened 2021-05-06
`solana-validator exit` now uses `process::exit()` to ensure prompt termination

mvines:


validator/src/admin_rpc_service.rs
```diff
@@ -56,12 +56,15 @@ pub struct AdminRpcImpl;
...
+    fn exit(&self, _meta: Self::Metadata) -> Result<()> {
         debug!("exit admin rpc request received");
         // Delay exit signal until this RPC request completes, otherwise the caller of `exit` might
         // receive a confusing error as the validator shuts down before a response is sent back.
         tokio::spawn(async move {
-            meta.validator_exit.write().unwrap().exit();
+            // TODO: Debug why ValidatorExit doesn't always cause the validator to fully exit
+            // (rocksdb background processing or some other stuck thread perhaps?)
+            std::process::exit(0);
+            // _meta.validator_exit.write().unwrap().exit();
```
sakridge, line 67: Use the clean way first? then fallback to `exit(0)` if it doesn't work after 5 seconds?

mvines: Yeah, not a bad idea.  I can do the same for the duplicate gossip exit code as well

PR #17090, opened 2021-05-06
async secondary indices

jeffwashington:
#### Problem
Secondary indices take a long time to generate on validator startup.

#### Summary of Changes
Generate secondary indices in the background. Really, move all access to the secondary index into an async channel model.

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -1190,8 +1190,9 @@ impl Default for AccountsDb {
 
         let mut bank_hashes = HashMap::new();
         bank_hashes.insert(0, BankHashInfo::default());
+        let account_indexes = AccountSecondaryIndexes::default();
         AccountsDb {
-            accounts_index: AccountsIndex::default(),
+            accounts_index: AccountsIndex::new(account_indexes.clone()),
             storage: AccountStorage::default(),
             accounts_cache: AccountsCache::default(),
             sender_bg_hasher: None,
@@ -1220,7 +1221,7 @@ impl Default for AccountsDb {
             shrink_stats: ShrinkStats::default(),
             stats: AccountsStats::default(),
             cluster_type: None,
-            account_indexes: AccountSecondaryIndexes::default(),
+            account_indexes,
             caching_enabled: false,
             #[cfg(test)]
             load_delay: u64::default(),
@@ -1251,6 +1252,7 @@ impl AccountsDb {
                 paths,
                 temp_paths: None,
                 cluster_type: Some(*cluster_type),
+                accounts_index: AccountsIndex::new(account_indexes.clone()),
                 account_indexes,
                 caching_enabled,
                 ..Self::default()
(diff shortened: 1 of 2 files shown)


jeffwashington:
This change is built on top of @carllin 's change.

[This](https://github.com/solana-labs/solana/pull/17090/commits/c8384d785f0b32ebdf0eafa18b5849f23e760b46) is the unique commit.

carllin:
I think the logic here is sound/safe and should guarantee consistent scans. Moreover, it's a really good idea to offload the updating of the secondary index to a background thread to prevent replay from blocking while a scan on that index is occurring. 

My concerns here are the scans in this approach are all serialized/performed by one thread, which may limit the capacity of each RPC replica to handle concurrent scans (RPC currently runs multiple threads serving requests). I have an idea here which might be interesting. Instead of serializing the lookup on the secondary indexes, we allow concurrent lookups, except we have to make a few extra checks:

First a little bit of context:
1) Scans should only occur on frozen banks
2) When RPC invokes a scan on a particular bank `B` happen, they pass in `ancestors` to the scan:  , and this ancestor list is inclusive of the bank itself: https://github.com/solana-labs/solana/blob/master/runtime/src/bank.rs#L4243. This means we scan through all state up to and including bank `B`.

Proposal:
The rough idea here is we can safely allow a scan on a bank  `B` if we know all the secondary index updates in the channel have been processed.
 
1) When a bank is frozen it sends a `AccountIndexAction::Frozen` signal into the channel
2) When the `background_secondary_indexer` sees the frozen signal, it marks this bank as safe for scan. Might need a new field in the bank here.
3)  When we get to `do_checked_scan_accounts()`, we need to know what the bank `B` being scanned is so we can check that it's safe for scan. We could for instance pass an additional argument `scan_bank` here: https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_index.rs#L490, or iterate and find the biggest slot in `ancestors`, https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_index.rs#L490

4)`Note:` After we find the bank `B` from 3), the current logic checks if this bank has already been outdated by a later root here: https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_index.rs#L567-L591. If so, we ignore all the ancestors passed in and just scan the rooted slots up to `max_root`

This means the bank we need to make our safety check on (checking all the secondary index updates are completed), the `safety_bank`, is equal to `if ancestors.is_empty() {B} else {max_root}`. 

So now this RPC thread blocks until `safety_bank` has completed all of its secondary index updates. While this RPC thread is waiting, it's safe from clean because we recorded `max_root` here: https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_index.rs#L511 and we won't clean past this `max_root` in the background.

5) Once the `safety_bank` is safe, we can continue with the scan!

carllin:
I think it may also be possible to have multiple `background_secondary_indexer` threads running reading from the channel, provided we have a peek operation on the channel, maybe something like: https://users.rust-lang.org/t/mpsc-channel-receiver-peek/24939

Then when the threads `peek()` and see a `AccountIndexAction::Remove(pubkey)` in the queue, they grab the lock in the `SecondaryIndex` dashmap for that pubkey before they pop from the channel, do the remove logic and then we continue onwards.

carllin:
fyi @CriesofCarrots trying to minimize contention between replay and scan 😃

jeffwashington:
discussed with @sakridge , @carllin :
This is likely not a 1.6 change. it is useful to have and consider going forward.

PR #17092, opened 2021-05-06
sdk: Add get_instance_packed_len for variable-size types

joncinque:
#### Problem

Borsh is great for (de)serialization of types in programs, and also supports variable-length types, such as `Vec` and `HashMap`.  Developers need to know the size of their type in order to allocate space for it, but it's not possible to know the size statically if there's a `Vec` or `HashMap`.

#### Summary of Changes

Add `get_instance_packed_len`, which uses the existing serialization logic to count how big the result would be, saving on the full serialization.  That way, a dapp user can create an instance with the maximum sizes they want to support, calculate the size, then allocate that much space in their account.

This is already being used for stake pools, and it could be useful for other projects using types with variable-length fields.

Fixes #

### DIFF
--- a/sdk/program/src/borsh.rs
+++ b/sdk/program/src/borsh.rs
@@ -2,9 +2,9 @@
 //! Borsh utils
 use {
     borsh::{
-        maybestd::io::Error,
+        maybestd::io::{Error, Write},
         schema::{BorshSchema, Declaration, Definition, Fields},
-        BorshDeserialize,
+        BorshDeserialize, BorshSerialize,
     },
     std::collections::HashMap,
 };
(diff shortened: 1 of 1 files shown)


sdk/program/src/borsh.rs
```diff
@@ -76,6 +76,36 @@ pub fn try_from_slice_unchecked<T: BorshDeserialize>(data: &[u8]) -> Result<T, E
...
+    }
+}
+
+/// Get the packed length for the serialized form of this object instance.
+///
+/// Useful when working with instances of types that contain a variable-length
+/// sequence, such as a Vec or HashMap.  Since it is impossible to know the packed
+/// length only from the type's schema, this can be used when an instance already
+/// exists, to figure out how much space to allocate in an account.
+pub fn get_instance_packed_len<T: BorshSerialize>(instance: &T) -> Result<usize, Error> {
```
jackcmay, line 103: Very cool, would be helpful to warn theapp  developers that the borsch `get_packed_len` isn't available from within bpf programs

joncinque: Oh, yes that's a great point.  I'll add a comment about that on `get_packed_len`

sdk/program/src/borsh.rs
```diff
@@ -159,4 +192,21 @@ mod tests {
                 + get_packed_len::<TestEnum>()
         );
     }
+
+    #[test]
+    fn instance_packed_len() {
```
jackcmay, line 197: Be nice to see a few more tests here, otherwise, besides this nit, I approve!

jackcmay (approved):

sdk/program/src/borsh.rs
joncinque: No problem, any other cases in particular that you'd like to see tested?

PR #17097, opened 2021-05-06
[refactor bank.rs] chore: refactor executors cache out from bank.rs

jon-chuang:
#### Problem
This is an incremental part of a planned big refactor of the 12 kloc `bank.rs`  with a directory structure similar to:
image

That's because as is, `bank.rs` is hard to read/cross-reference, slows down some people's IDEs, and in addition a refactor can lead to the following benefits:

**Benefits**. 
1. Easier for new core/contract devs to start reading/pinpointing needed sections
2. Logical refactoring means all relevant methods and tests are grouped up in the same file - so easier to reference & refresh on one particular component, even as an experienced dev. 
3. Further changes going can be forced to be more modular, now having to be aware of file/module boundaries. Can enforce greater separation between public, crate-level, and module-level methods.

**Negatives**. 
1. no git blame beyond point of refactor (but this is in some sense temporary)
2. cannot search single file (but can adapt to knowing which refactored component to search, and have a maximum 1-2 hops due to logical organisation).
3. giant refactor may be susceptible to bugs. But we can review incremental changes starting with the low hanging fruit. Some basic sanity checks and the simplicity - copy and pasting, means the probability of introducing an error is low.

**Summary**
The non-zero effort required to accurately refactor and review changes ought to pay off in the long-term for maintainability/modularity/developer productivity, especially getting new core/contract devs up to speed on the most central component of the runtime. 

Many (like @brooksprumo @leoluk @sakridge) have expressed support of this direction, so it's clearly a desire among some of the core devs/contributors, but this must be weighed against the cost, especially of review. But I am careful to check that all the tests are the same and that they run and pass.

#### Notes

wrt 3., just need to make sure that we don't run out of steam by the time it comes to larger components. Those may also be slightly more intertwined. We'll leave the discussion of those to later.

#### Estimated Cost
The cost is:
1. the refactor done by me, achieving 2-birds-1-stone of reading the codebase thoroughly and understanding the logical organisation as well as the concrete implementation of the runtime components
2. review, which would be about 15-20 PRs, one for each module, that takes about 5-10 mins to read through, copy and paste some code and check no diffs are introduced, and then rely on CI to make sure the tests run.
3. Time: the longer it drags out, the more of a chore it will be. 

#### Timeline
Target time for full refactor to be merged: overall review - 1 week, individual merges: 1 week (5 modules every 2 days). Total: 2 weeks.

Alternatively, as discussed on the discord, we split into two phases: v0, v1. v0 will refactor cleanly separable components that can be checked easily and have low risk. v1 will proceed if people feel comfortable and think the refactored version is much more ergonomic and pleasant to read so it's worth the effort (which I think it is).

Wrt to not dragging out, I suggest 
1. standardise the PR format and review procedure. 
2. put 3-5 PRs up at a time, based on individual complexity, to be reviewed and merged asynchronously. 

#### Actions
Within about 2 days, I will submit a draft PR which contains a blueprint of the final refactor, where ppl can clone and get a feel for (the very satisfying and neat) end result and give feedback on. The result is about 20 modules, ranging from about 200-1000 loc each, well-organised and providing a better factored picture of what the various functions of the bank are and how they are connected to one another.

The recommended reviewing style is to copy and paste destination code snippets into the source snippet in the master on your own IDE, and make sure there are no diffs introduced. 

Note that this is a different style from a typical feature/bug fix PR review. 

**The reviewer is not recommended to take a look at the `Files changed` tab directly, but needs to have an IDE open to check the diff in the recommended way.**

Also make sure that exactly `520` tests run in `solana/runtime`.

#### Review Actions
- [ ] Copied and pasted destination code snippets into source location in master and checked no diffs are introduced.
- [ ] Made sure 520 tests run in `solana/runtime`, 172 tests run for bank via `cargo test bank`.
- [ ] Run tests specifically for the module `cargo test bank::module_name`
- [ ] Do a once over to check the refactoring makes sense and that imports are correct.

#### Summary of Changes

We start with the lowest hanging fruit, the executor cache, a very isolated component of the bank.

Although barely making a dent, we have to start somewhere.

CC: @mvines @jackcmay


jon-chuang:
I have done a self-review as a sanity check. This took me 5 mins to execute and 10 mins to be sure the changes are isomorphic to the original.

- [x] Copied and pasted destination code snippets into source location in master and checked no diffs are introduced.
- [x]  Made sure 520 tests run in solana/runtime, 172 tests run for bank via cargo test bank.
- [x]  Run tests specifically for the module cargo test bank::module_name
- [x] Do a once over to check the refactoring makes sense and that imports are correct.

brooksprumo:
Indeed, I am on board for splitting up `bank.rs` into its smaller logical components! Thanks for doing this work!

I tested this out locally by copying `executor_cache.rs` from the PR in my local copy of `master`, I removed the executor duplicate code from `bank.rs` and added 
```
mod executor_cache;
use executor_cache::{CachedExecutors, CowCachedExecutors, MAX_CACHED_EXECUTORS};
```
to `bank.rs`. I also purposely left the tests in `bank.rs`, as I wanted to make sure existing tests passed. 

After these changes, I ran `cargo test` from `runtime/` and all tests passed.



@jon-chuang Going forward, would it be beneficial to split PRs like this one into two PRs? The first would move the impl code to the new .rs file, while leaving the original tests in place. Then a second PR would move the tests over. I think this way would make it easier to ensure no behavior was inadvertently changed, and thus easier to review.

runtime/src/bank/executor_cache.rs
```diff
@@ -0,0 +1,377 @@
+use std::{
```
brooksprumo, line 1: Oh, also I think it would be helpful to add some `//!` doc comments to the new files. Even if it's basic, this is where I look first to see if I can learn about intentions and interactions of the contained code.

runtime/src/bank/executor_cache.rs
```diff
@@ -0,0 +1,377 @@
...
+    cell::RefCell,
+    collections::HashMap,
+    rc::Rc,
+    sync::{
+        atomic::{AtomicU64, Ordering::Relaxed},
+        Arc, LockResult, RwLock, RwLockReadGuard, RwLockWriteGuard,
+    },
+};
+
```
jackcmay, line 10: get rid of the whitespace between `use` lines

runtime/src/bank/executor_cache.rs
```diff
@@ -0,0 +1,377 @@
...
+            .insert(*pubkey, (AtomicU64::new(0), executor));
+    }
+    pub(super) fn remove(&mut self, pubkey: &Pubkey) {
+        let _ = self.executors.remove(pubkey);
+    }
+}
+
+impl Bank {
+    /// Add executors back to the bank's cache if modified
+    pub(super) fn update_executors(&self, executors: Rc<RefCell<Executors>>) {
```
jackcmay, line 136: These functions could be implemented in their own module which the bank calls after grabbing the write/read lock.  But unlike the other two, the `update` function avoids taking a lock if the executor is not dirty.  To retain this behavior you could implement these functions for a locked exector as stand-alone functions or as part of the executor impl.

runtime/src/bank/executor_cache.rs
```diff
@@ -0,0 +1,377 @@
...
+        let _ = self
+            .executors
+            .insert(*pubkey, (AtomicU64::new(0), executor));
+    }
+    pub(super) fn remove(&mut self, pubkey: &Pubkey) {
+        let _ = self.executors.remove(pubkey);
+    }
+}
+
+impl Bank {
```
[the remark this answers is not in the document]
jon-chuang: Hmm why is this? 

Typically I don't like that as well, but given the Bank impl is pretty big, but can be refactored into logical groups that makes it easier to read and cross reference, I thought in this case it would be suitable.

PR #17108, opened 2021-05-07
type AccountSecondaryIndexes = HashSet

jeffwashington:
#### Problem
Account secondary indexes are about to be more complicated. Much code assumes that the account secondary indexes are defined only by a hashset. See [here](https://github.com/solana-labs/solana/issues/17019#issuecomment-834481000).
Follow up PR: #17110
#### Summary of Changes
Make a type and use type::default() so we can replace the implementation.
Fixes #


sakridge (dismissed):
straightforward. looks good

lijunwangs (approved):
Looks good to me

PR #17109, opened 2021-05-07
Move block-time caching earlier

CriesofCarrots:
#### Problem
The CacheBlockTimeService is triggered on bank root in ReplayStage. This means there is a gap where a block is complete and rooted in Blockstore, but its block time is not available to Rpc. This is silly, because the block time *is* available to Rpc (from the Bank) if queried before the block is rooted. This behavior is a vestige of the old code before the block time was available in the Bank for every block.

#### Summary of Changes
Move CacheBlockTimeService sends to immediately after a Bank is frozen: in ReplayStage, and multiple spots in blockstore_processor for genesis block and slots processed from snapshot
Also sneak in a little cleanup to pass transaction_status_sender by reference into the same places

### DIFF
--- a/ledger/src/blockstore.rs
+++ b/ledger/src/blockstore.rs
@@ -1760,10 +1760,13 @@ impl Blockstore {
     }
 
     pub fn cache_block_time(&self, slot: Slot, timestamp: UnixTimestamp) -> Result<()> {
-        if !self.is_root(slot) {
-            return Err(BlockstoreError::SlotNotRooted);
+        let slot_meta_cf = self.db.column::<cf::SlotMeta>();
+        if slot_meta_cf.get(slot)?.is_none() {
+            info!("SlotMeta not found for slot {}", slot);
+            Err(BlockstoreError::SlotUnavailable)
+        } else {
+            self.blocktime_cf.put(slot, &timestamp)
         }
-        self.blocktime_cf.put(slot, &timestamp)
     }
 
     pub fn get_first_available_block(&self) -> Result<Slot> {
--- a/ledger/src/blockstore.rs
+++ b/ledger/src/blockstore.rs
@@ -1760,13 +1760,7 @@ impl Blockstore {
     }
 
     pub fn cache_block_time(&self, slot: Slot, timestamp: UnixTimestamp) -> Result<()> {
-        let slot_meta_cf = self.db.column::<cf::SlotMeta>();
-        if slot_meta_cf.get(slot)?.is_none() {
-            info!("SlotMeta not found for slot {}", slot);
-            Err(BlockstoreError::SlotUnavailable)
-        } else {
-            self.blocktime_cf.put(slot, &timestamp)
-        }
+        self.blocktime_cf.put(slot, &timestamp)
     }
 
     pub fn get_first_available_block(&self) -> Result<Slot> {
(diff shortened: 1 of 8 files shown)


ledger/src/blockstore.rs
```diff
@@ -1760,10 +1760,13 @@ impl Blockstore {
     }
 
     pub fn cache_block_time(&self, slot: Slot, timestamp: UnixTimestamp) -> Result<()> {
-        if !self.is_root(slot) {
-            return Err(BlockstoreError::SlotNotRooted);
+        let slot_meta_cf = self.db.column::<cf::SlotMeta>();
+        if slot_meta_cf.get(slot)?.is_none() {
```
carllin, line 1764: Can't wait for this to be released: https://github.com/rust-rocksdb/rust-rocksdb/pull/518, so that we can check for existence without eating the fetch/deserialization cost of getting the value.

Is there anything bad that can happen if the `SlotMeta` doesn't exist but the block time does? 

Just curious because it seems like this case shouldn't happen because we only add to the cached block here after freeze from replay/blockstore replay on startup via `process_single_slot()`.

Even including the snapshot root bank which we may not have the ledger block for (I don't think we call the cache block method on that bank since it's not processed via `process_single_slot()). And clean hopefully shouldn't be running on recent slots you're still replaying.

ledger/src/blockstore_processor.rs
```diff
@@ -1171,6 +1196,26 @@ impl TransactionStatusSender {
...
+pub fn cache_block_time(
+    blockstore: &Blockstore,
+    bank: &Arc<Bank>,
+    cache_block_time_sender: Option<&CacheBlockTimeSender>,
+) {
+    if let Some(cache_block_time_sender) = cache_block_time_sender {
+        if blockstore
+            .get_block_time(bank.slot())
+            .unwrap_or_default()
+            .is_none()
```
carllin, line 1210: I think we can move this read into the background thread as well, maybe even directly in the `blockstore::cache_block_time()` method

ledger/src/blockstore.rs
CriesofCarrots: Good point! I think you're right; there doesn't seem to be any risk of a block time existing where `SlotMeta` does not. I'll just remove this check altogether.

CriesofCarrots: Done here: https://github.com/solana-labs/solana/pull/17109/commits/0f1741c81b00f8cccc1a2077bc76915cb019cbb8

ledger/src/blockstore_processor.rs
CriesofCarrots: Done here https://github.com/solana-labs/solana/pull/17109/commits/45d5815f3b10b144fd55999fbdedfedcbc0aad25

CriesofCarrots:
@carllin , anything else here?

carllin (approved):

ledger/src/blockstore.rs
ryoqun: @carllin @CriesofCarrots 

> Can't wait for this to be released: rust-rocksdb/rust-rocksdb#518

This is finally landed in upstream and available in our codebase starting from: https://github.com/solana-labs/solana/pull/18904

ryoqun: ... but it look like the new method incurs the possibility of false positive:

https://docs.rs/rocksdb/0.17.0/rocksdb/struct.DBWithThreadMode.html#method.key_may_exist_cf_opt:

> Returns false if the given key definitely doesn’t exist in the specified column family, otherwise returns true.

PR #17124, opened 2021-05-08
Fix syscalls in the C SDK failing at runtime when compiled as C++

machacekch:
#### Problem
Some syscalls are wrongly declared `static` in solana_sdk.h, which makes clang++ assume they are local to the compilation unit. It therefore ignores the `extern "C" {}` block and mangles their names. While that doesn't break C++ compilation, the syscall fails at runtime with something along the lines of `ELF error: Unresolved symbol (_ZL26sol_create_program_addressPK13SolSignerSeediPK9SolPubkeyS4_)`.

#### Summary of Changes
Cleanly separate syscalls from non-syscalls and ensure the former are no longer declared static.


mvines (approved):
Thanks!

PR #17128, opened 2021-05-08
Add a make target to run the readelf utility on a compiled program

machacekch:
#### Summary of Changes
The readelf utility (already shipped with the solana tools) shows meta-information about ELF files, such as symbol tables. It is useful for investigating "unresolved symbol" errors that crop up at runtime.

This commit also fixes the objdump flags (two dashes are required and there is no "color" option) as well as a few typos.


mvines (approved):

mvines:
CI failure is unrelated

PR #17147, opened 2021-05-10
Zero pad data shreds on fetch from blockstore

### LINKED ISSUE #17136: Shreds inserted by master causes panic when read by v1.6
#### Problem

Maybe, #16602 started to cause this problem? (CC: @steviez )

```
$ solana-ledger-tool --ledger bad-deploy-ledger-copy slot 77052081
[2021-05-10T06:40:35.791148527Z INFO  solana_ledger_tool] solana-ledger-tool 1.6.6 (src:fe775a97; feat:3714435735)
[2021-05-10T06:40:35.791215160Z INFO  solana_ledger::blockstore] Maximum open file descriptors: 500000
[2021-05-10T06:40:35.791223466Z INFO  solana_ledger::blockstore] Opening database at "/home/sol/bad-deploy-ledger-copy/rocksdb"
[2021-05-10T06:40:35.815238897Z INFO  solana_ledger::blockstore] "/home/sol/bad-deploy-ledger-copy/rocksdb" open took 24ms
Slot 77052081
thread 'thread 'thread 'blockstore_2blockstore_5' panicked at 'blockstore_3' panicked at '' panicked at 'assertion failed: payload.len() >= expected_data_sizethread 'assertion failed: payload.len() >= expected_data_sizeassertion failed: payload.len() >= expected_data_size', blockstore_0', ', ledger/src/shred.rs' panicked at 'ledger/src/shred.rsledger/src/shred.rs:assertion failed: payload.len() >= expected_data_size::279279:', 279:ledger/src/shred.rs9
:9:note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
9
279
:thread '9blockstore_6
' panicked at 'assertion failed: payload.len() >= expected_data_size', ledger/src/shred.rs:279:9
thread 'blockstore_7' panicked at 'assertion failed: payload.len() >= expected_data_size', ledger/src/shred.rs:279:9
thread 'blockstore_4' panicked at 'assertion failed: payload.len() >= expected_data_size', ledger/src/shred.rs:279:9
thread 'blockstore_1' panicked at 'assertion failed: payload.len() >= expected_data_size', ledger/src/shred.rs:279:9
thread 'blockstore_0' panicked at 'assertion failed: payload.len() >= expected_data_size', ledger/src/shred.rs:279:9
thread 'blockstore_3' panicked at 'assertion failed: payload.len() >= expected_data_size', ledger/src/shred.rs:279:9
thread 'thread 'blockstore_5blockstore_2' panicked at '' panicked at 'assertion failed: payload.len() >= expected_data_sizethread 'assertion failed: payload.len() >= expected_data_size', blockstore_4', ledger/src/shred.rs' panicked at 'ledger/src/shred.rs:assertion failed: payload.len() >= expected_data_size:thread '279', thread '279blockstore_7:thread ':ledger/src/shred.rsblockstore_1' panicked at '9assertion failed: payload.len() >= expected_data_sizeblockstore_69:' panicked at '
', ' panicked at 'ledger/src/shred.rsassertion failed: payload.len() >= expected_data_size
279assertion failed: payload.len() >= expected_data_sizethread ':thread '', :', blockstore_3279blockstore_0ledger/src/shred.rs9ledger/src/shred.rs' panicked at ':' panicked at ':
:assertion failed: payload.len() >= expected_data_size9assertion failed: payload.len() >= expected_data_size279279',
', ::ledger/src/shred.rsledger/src/shred.rs99thread '::

blockstore_2279279' panicked at '::assertion failed: payload.len() >= expected_data_size99',

ledger/src/shred.rsthread ':blockstore_5279' panicked at ':assertion failed: payload.len() >= expected_data_size9',
ledger/src/shred.rs:279:9
thread 'blockstore_4' panicked at 'assertion failed: payload.len() >= expected_data_size', ledger/src/shred.rs:279:9
thread 'blockstore_5' panicked at 'assertion failed: payload.len() >= expected_data_size', ledger/src/shred.rs:279thread ':9blockstore_1
' panicked at 'assertion failed: payload.len() >= expected_data_sizethread '', blockstore_7ledger/src/shred.rs' panicked at ':assertion failed: payload.len() >= expected_data_size279', thread 'blockstore_6' panicked at 'assertion failed: payload.len() >= expected_data_size:ledger/src/shred.rsthread '', thread '9:blockstore_3ledger/src/shred.rsblockstore_4
279' panicked at ':' panicked at '279thread ':assertion failed: payload.len() >= expected_data_sizeassertion failed: payload.len() >= expected_data_sizethread ':blockstore_29', ', blockstore_09thread '' panicked at '
ledger/src/shred.rsledger/src/shred.rs' panicked at '
blockstore_5assertion failed: payload.len() >= expected_data_sizethread '::assertion failed: payload.len() >= expected_data_size' panicked at '', blockstore_1279279', assertion failed: payload.len() >= expected_data_sizeledger/src/shred.rs' panicked at 'thread '::ledger/src/shred.rs', thread ':assertion failed: payload.len() >= expected_data_sizeblockstore_799:ledger/src/shred.rsblockstore_6279', ' panicked at '

279:' panicked at ':ledger/src/shred.rsassertion failed: payload.len() >= expected_data_size:279assertion failed: payload.len() >= expected_data_size9:', 9:',
279ledger/src/shred.rs:
9ledger/src/shred.rsthread '9
:blockstore_3:
279' panicked at '279::assertion failed: payload.len() >= expected_data_size99',

```

#### Proposed Solution

Support reading old format shreds from rocksdb?

steviez:
#### Problem
#16602 introduced a change where shreds would have zero-padding stripped off for insertion into the blockstore, and then re-appended upon fetch from the blockstore. This introduced a compatibility break with v1.6 since v1.6 doesn't have the logic to "extend" shreds. This is problematic for a scenario where a node is running the tip of master (or really, any state of master at/after the mentioned PR) and inserts shreds into the blockstore, and then reverts to v1.6 and attempts to retrieve/operate on those shreds.

#### Summary of Changes
Add logic to ensure shred payloads are a consistent length
- For shreds that were inserted with the zero-padding (v1.6), this should be a no-op
- For shreds that were inserted without zero-padding (latest on master), this will create compatibility for v1.6

Fixes #17136

### DIFF
--- a/ledger/src/shred.rs
+++ b/ledger/src/shred.rs
@@ -263,20 +263,18 @@ impl Shred {
     }
 
     pub fn new_from_serialized_shred(mut payload: Vec<u8>) -> Result<Self> {
+        // A shred can be deserialized in several cases; payload length will vary for these:
+        //   payload.len() <= SHRED_PAYLOAD_SIZE when payload is retrieved from the blockstore
+        //   payload.len() == SHRED_PAYLOAD_SIZE when payload is from a local shred (shred.payload)
+        //   payload.len()  > PACKET_DATA_SIZE when payload comes from a packet (window serivce)
+        // Resize here so the shreds always have the same length
+        payload.resize(SHRED_PAYLOAD_SIZE, 0);
+
         let mut start = 0;
         let common_header: ShredCommonHeader =
             Self::deserialize_obj(&mut start, SIZE_OF_COMMON_SHRED_HEADER, &payload)?;
 
         let slot = common_header.slot;
-        let expected_data_size = SHRED_PAYLOAD_SIZE;
-        // Safe because any payload from the network must have passed through
-        // window service,  which implies payload wll be of size
-        // PACKET_DATA_SIZE, and `expected_data_size` <= PACKET_DATA_SIZE.
-        //
-        // On the other hand, if this function is called locally, the payload size should match
-        // the `expected_data_size`.
-        assert!(payload.len() >= expected_data_size);
-        payload.truncate(expected_data_size);
         let shred = if common_header.shred_type == ShredType(CODING_SHRED) {
             let coding_header: CodingShredHeader =
                 Self::deserialize_obj(&mut start, SIZE_OF_CODING_SHRED_HEADER, &payload)?;
--- a/ledger/src/blockstore.rs
+++ b/ledger/src/blockstore.rs
@@ -1438,7 +1438,16 @@ impl Blockstore {
     }
 
     pub fn get_data_shred(&self, slot: Slot, index: u64) -> Result<Option<Vec<u8>>> {
-        self.data_shred_cf.get_bytes((slot, index))
+        use crate::shred::SHRED_PAYLOAD_SIZE;
+        self.data_shred_cf.get_bytes((slot, index)).map(|data| {
+            data.map(|mut d| {
+                // For forward compatibility, pad the payload out to
+                // SHRED_PAYLOAD_SIZE incase the shred was inserted
+                // with any padding stripped off.
+                d.resize(cmp::max(d.len(), SHRED_PAYLOAD_SIZE), 0);
+                d
+            })
+        })
     }
 
     pub fn get_data_shreds_for_slot(


carllin:
hmmm I think there actually may be a lot more issues than initially expected. Specifically, on 1.6 I'm seeing this assert trigger: https://github.com/solana-labs/solana/pull/16602/files#diff-b4053531f29bf62ea474f3434620c0fdeda978406ca267e7dc1a5197b4afd22fL328 in `new_from_serialized_shred()`, even with this change. I suspect this means that we've not properly accounted for every place `new_from_serialized_shred` can be called from, for example, one such place is `blockstore::get_entries_in_data_block()`

carllin:
> > hmmm I think there actually may be a lot more issues than initially expected. Specifically, on 1.6 even with this change I'm seeing this assert trigger: https://github.com/solana-labs/solana/pull/16602/files#diff-b4053531f29bf62ea474f3434620c0fdeda978406ca267e7dc1a5197b4afd22fL328 in `new_from_serialized_shred()`. I suspect this means that we've not properly accounted for every place `new_from_serialized_shred` can be called from, for example, one such place is `blockstore::get_entries_in_data_block()`
> 
> Thanks for chiming in - was working on getting a setup running to test this but that testing may not have uncovered this issue. But agreed, searching for `.get_bytes()` in that file, looks like there are several instances where we operate on the data_shred column family:
> 
> ```
> get_entries_in_data_block()
> get_recovery_data_shreds()
> get_data_shred()
> ```
> 
> So unfortunately, it looks like this is broken on master right now. Would be nice if all 3 of these funneled through one code-path; let me see if I can spin that up (for both master and v1.6)

Thanks! yeah perhaps a `new_from_serialized_blockstore_shred` wrapper

steviez:
@carllin - So I think there are two cases here; creating compatibility between tip of master (soon to be v1.7) and v1.6, and handling things consistently in latest.

Since this PR is targeted at v1.6, I think we keep things simple here and just make it such that the shreds are always the same length (`SHRED_PAYLOAD_SIZE`) when pulled out of RocksDB. In total, there are four locations that data shreds are pulled out of RocksDB:
```
        get_recovery_data_shreds()
        get_entries_in_data_block()
        get_data_shreds_for_slot()
        get_data_shred()
```
The first 3 of these all call `new_from_serialized_shred()`; `get_data_shred()`is the only that does not as it just returns a byte array. So, the following is the simplest path to consistency in my opinion:
- Drop the size check assertion from `new_from_serialized_shred()` and call `payload.resize(SHRED_PAYLOAD_SIZE)`
- Perform a resize to `SHRED_PAYLOAD_SIZE` bytes in `get_data_shred()`

This is consistent with what we saw work in testing, so unless there were any findings that I'm unaware of, I think this minimal set of changes creates compatibility for v1.6 code operating on shreds inserted by tip of master (where shreds have zero-padding dropped on insert)

ledger/src/shred.rs
```diff
@@ -263,20 +263,18 @@ impl Shred {
     }
 
     pub fn new_from_serialized_shred(mut payload: Vec<u8>) -> Result<Self> {
+        // A shred can be deserialized in several cases; payload length will vary for these:
+        //   payload.len() <= SHRED_PAYLOAD_SIZE when payload is retrieved from the blockstore
+        //   payload.len() == SHRED_PAYLOAD_SIZE when payload is from a local shred (shred.payload)
+        //   payload.len()  > PACKET_DATA_SIZE when payload comes from a packet (window serivce)
+        // Resize here so the shreds always have the same length
+        payload.resize(SHRED_PAYLOAD_SIZE, 0);
```
steviez, line 271: @jeffwashington - For the sake of more datapoints on this, did you encounter any issues after applying this change locally? This was the change to allow use of older stack (such as v1.6) on a ledger populated by master

carllin (approved):

ledger/src/shred.rs
steviez: From DM with Jeff:
> so, you’re trying to get more data points on your change on shreds?
> I have not run into any issues.

PR #17151, opened 2021-05-10
fix: Improve findProgramAddress error when MAX_SEED_LENGTH is exceeded

machacekch:
#### Problem
findProgramAddress calls createProgramAddress, which can throw for two reasons: a seed given is larger than MAX_SEED_LENGTH or because the address doesn't fall off the curve. Currently, both exceptions report "Unable to find a viable program address nonce", which in the first case doesn't call out the issue.

#### Summary of Changes
When MAX_SEED_LENGTH is exceeded, it doesn't help to try other nonces, so this error case is now handled by re-throwing the inner Error immediately. The error reported this way ("Max seed length exceeded") also points to the problem more directly than the previously reported "Unable to find a viable program address nonce".

### DIFF
--- a/web3.js/src/publickey.ts
+++ b/web3.js/src/publickey.ts
@@ -113,7 +113,7 @@ export class PublicKey {
     let buffer = Buffer.alloc(0);
     seeds.forEach(function (seed) {
       if (seed.length > MAX_SEED_LENGTH) {
-        throw new Error(`Max seed length exceeded`);
+        throw new TypeError(`Max seed length exceeded`);
       }
       buffer = Buffer.concat([buffer, toBuffer(seed)]);
     });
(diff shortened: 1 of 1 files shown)


machacekch:
Looks like travis doesn't like my commit message format. Can you fix it on your end or do I have to change something? If it's easier, feel free to just add the changes by hand - it's just 4 lines of code after all.

oJshua (changes requested):
Looks good @machacekch! 

Could you rename this commit: "fix: Improve findProgramAddress error when MAX_SEED_LENGTH is exceeded"? (and this should pass CI)

machacekch:
> Could you rename this commit: "fix: Improve findProgramAddress error when MAX_SEED_LENGTH is exceeded"? (and this should pass CI)

I suppose changing the title of the pull request is not going to cut it. Can the commit itself be renamed without sending a new one?

oJshua:
@machacekch right, you can do a quick:
 `git commit --amend -m "fix: Improve findProgramAddress error when MAX_SEED_LENGTH is exceeded"`
and then force push and it should straighten out.

machacekch:
Hmm, "subject must not be sentence-case, start-case, pascal-case, upper-case" - what's tripping it up this time? The all-caps of MAX_SEED_LENGTH?

oJshua:
Hmm, it must be the Sentence case. I can fix it when I squash the commit. Thanks for your work on this @machacekch

PR #16127, opened 2021-03-25
Proposal to handle duplicate blocks

carllin:
#### Problem
No design for slashing duplicate leaders and storing/handling multiple versions of a slot

#### Summary of Changes
Add such a design

Fixes #

PR #17168, opened 2021-05-11
Switch EpochSlots to be frozen slots, not completed slots

carllin:
#### Problem
If the cluster doesn't directly voted on one version of a slot `S` (maybe only voted on the descendants), and we marked slot `S` as dead, then currently there's no mechanism by which to detect that the cluster has frozen another version of this slot.

#### Summary of Changes
Switch EpochSlots to be for frozen slots instead of just completed slots, so we can later invoke dump/repair logic if a slot is frozen > duplicate threshold of the cluster. Design follows the proposal here: https://github.com/solana-labs/solana/pull/16127/commits/f79c0e803ea92b9eebc137c1675d54efb72f469d#diff-0bd0e50030bdee14082c678da5f3098df8156825fa86c59ad55ddbd2a3669585R82, which describes the context in more detail

Fixes #

### DIFF
--- a/core/src/cluster_slots_service.rs
+++ b/core/src/cluster_slots_service.rs
@@ -1,29 +1,32 @@
 use crate::cluster_slots::ClusterSlots;
+use crossbeam_channel::{Receiver, RecvTimeoutError, Sender};
 use solana_gossip::cluster_info::ClusterInfo;
-use solana_ledger::blockstore::{Blockstore, CompletedSlotsReceiver};
+use solana_ledger::blockstore::Blockstore;
 use solana_measure::measure::Measure;
 use solana_runtime::bank_forks::BankForks;
 use solana_sdk::{clock::Slot, pubkey::Pubkey};
 use std::{
     sync::{
         atomic::{AtomicBool, Ordering},
-        mpsc::RecvTimeoutError,
         {Arc, RwLock},
     },
     thread::{self, Builder, JoinHandle},
     time::{Duration, Instant},
 };
 
+pub type ClusterSlotsUpdateReceiver = Receiver<Vec<Slot>>;
+pub type ClusterSlotsUpdateSender = Sender<Vec<Slot>>;
+
 #[derive(Default, Debug)]
 struct ClusterSlotsServiceTiming {
     pub lowest_slot_elapsed: u64,
-    pub update_completed_slots_elapsed: u64,
+    pub process_cluster_slots_updates_elapsed: u64,
 }
 
 impl ClusterSlotsServiceTiming {
-    fn update(&mut self, lowest_slot_elapsed: u64, update_completed_slots_elapsed: u64) {
+    fn update(&mut self, lowest_slot_elapsed: u64, process_cluster_slots_updates_elapsed: u64) {
         self.lowest_slot_elapsed += lowest_slot_elapsed;
-        self.update_completed_slots_elapsed += update_completed_slots_elapsed;
+        self.process_cluster_slots_updates_elapsed += process_cluster_slots_updates_elapsed;
     }
 }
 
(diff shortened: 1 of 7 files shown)


sakridge:
Ok, so I guess the only downside might be that a node which is trying to repair a slot which is errored wouldn't have as much information to repair since it doesn't know who has the slot data.

I guess we don't need any kind of switch-over. The EpochSlot still has the same data and meaning for being able to repair a successfully frozen slot.

carllin:
> Ok, so I guess the only downside might be that a node which is trying to repair a slot which is errored wouldn't have as much information to repair since it doesn't know who has the slot data.
> 
> I guess we don't need any kind of switch-over. The EpochSlot still has the same data and meaning for being able to repair a successfully frozen slot.

Yeah people repairing an errored slot will not have all the details, and now people who want to repair non-errored slots will have to wait for people to replay everything which might cause a little more of a delay as well, usually 200-300ms.

jstarry (commented):
Look good to me. Did you consider having the leader push their slots into epoch slots too? If they are the one producing the dupe blocks it's a little weird but it would help cross the threshold. Seems useful for detecting descendant confirmed blocks too

carllin:
> Look good to me. Did you consider having the leader push their slots into epoch slots too? If they are the one producing the dupe blocks it's a little weird but it would help cross the threshold. Seems useful for detecting descendant confirmed blocks too

https://github.com/solana-labs/solana/pull/17168/files#diff-6d8458bb2e53158ac472a9ad4709e6a0a52b75d930c019013298e8acda133828R1731 This should run on the leader as well, when they freeze the bank, they'll also update their own EpochSlots

jstarry:
> This should run on the leader as well, when they freeze the bank, they'll also update their own EpochSlots

Ah, that's right. I forgot that replay stage is what advances the working bank in poh service for leader blocks, not banking stage itself.

carllin:
@behzadnouri, hmmm it's a possibility, but the number of skipped slots due to the leader propagation check which relies on these EpochSlots seems relatively unaffected:

https://metrics.solana.com:8888/sources/18/chronograf/data-explorer?query=SELECT%20count%28%22slot%22%29%20AS%20%22mean_slot%22%20FROM%20%22tds%22.%22autogen%22.%22replay_stage-skip_leader_slot%22%20WHERE%20time%20%3E%20now%28%29%20-%201h%20GROUP%20BY%20time%281h%29%20FILL%28null%29

ryoqun:
@behzadnouri @carllin well, it seems that there's definitely some repair related regression between v1.7.0 and v1.7.1:

this node is switched between v1.7.0 <=> v1.7.1. And only if v1.7.1 is running, there's some delayed repairs with same hw and network:

image

And catchup (ie node restart) is affected the most, that's because some missing blocks must be repaired at startup to catchup...

The odd part is that this is not happening to every node. I've received multiple similar reports from several node operators.

carllin:
@ryoqun, thanks for the update!

Is this consistent across multiple restarts with 1.7.0 and 1.7.1 on this machine? If toggling between the two releases can reliably repro this behavior, then it looks like a good machine to start bisecting changes 😃 . For example if we roll back this PR, does the behavior improve on this single node?

I took another look at this PR and can't find anything that would affect repair on a *single* node unless the entire cluster were to downgrade to 1.7 and revert their `EpochSlots` to the old behavior. 

If catchup is still flaky with this PR rolled back on a single node, this could point to this PR needing a cluster-wide rollback in order to fix the problem

carllin:
I've been running both 1.7.0 and 1.7.1 and haven't found much correlation between for the spikes in `get-best-shreds` on validator startup

1) The first three 3 runs starting at 21:45 here are on 1.7.1,
2) The next 2 starting at 22:45 and 23:15 are on 1.7.0

Screen Shot 2021-06-12 at 11 32 18 PM

As shown on the last two runs, even on 1.7.0 the repair times can fluctuate dramatically, presumably based on how fast the validator can detect which forks to repair based on gossip votes. The average slot sizes are also roughly the same across all the runs: 

Screen Shot 2021-06-12 at 11 35 42 PM


Now that's not to say there's not a regression between 1.7.0 and 1.7.1, it may just not be observable on a single node with this metric/unless the entire cluster rolls back.

ryoqun:
> Is this consistent across multiple restarts with 1.7.0 and 1.7.1 on this machine?

yeah.

Also, my testnode machine can't repro this.. very odd: https://metrics.solana.com:3000/d/monitor-edge/cluster-telemetry-edge?orgId=2&var-datasource=Solana%20Metrics%20(read-only)&var-testnet=tds&var-hostid=BgVN55yxGYEd99zwchcFzd2n7YATMf5zTMBBaj79hBP1&from=1623566258211&to=1623567158211
 ;)

image


Also, I got another report apparently from another node operator, v1.7.1 started to WARN this message: 

```
Ignored received repair requests from ME: {}
```

Sadly, accompanied metric `"serve_repair-handle-repair--eq"` is DEBUG...

PR #17171, opened 2021-05-11
Rocks def lz4

sakridge:
#### Problem

testing rocksdb with compression on

#### Summary of Changes

Fixes #


sakridge:
Just for testing

PR #17175, opened 2021-05-11
Resize shreds on all get methods

steviez:
#### Problem
In another PR, we discovered that our logic to properly resize shreds pulled from the blockstore is a little inconsistent.

#### Summary of Changes
Funnel code paths through a common function.

Fixes #

### DIFF
--- a/ledger/src/blockstore.rs
+++ b/ledger/src/blockstore.rs
@@ -549,8 +549,7 @@ impl Blockstore {
         (set_index..set_index + erasure_meta.config.num_data() as u64).for_each(|i| {
             if index.data().is_present(i) {
                 if let Some(shred) = prev_inserted_datas.remove(&(slot, i)).or_else(|| {
-                    let some_data = data_cf
-                        .get_bytes((slot, i))
+                    let some_data = Self::fetch_data_shred(slot, i, &data_cf)
                         .expect("Database failure, could not fetch data shred");
                     if let Some(data) = some_data {
                         Shred::new_from_serialized_shred(data).ok()
@@ -1471,9 +1470,13 @@ impl Blockstore {
         Ok(newly_completed_data_sets)
     }
 
-    pub fn get_data_shred(&self, slot: Slot, index: u64) -> Result<Option<Vec<u8>>> {
+    fn fetch_data_shred(
+        slot: Slot,
+        index: u64,
+        data_cf: &LedgerColumn<cf::ShredData>,
+    ) -> Result<Option<Vec<u8>>> {
         use crate::shred::SHRED_PAYLOAD_SIZE;
-        self.data_shred_cf.get_bytes((slot, index)).map(|data| {
+        data_cf.get_bytes((slot, index)).map(|data| {
             data.map(|mut d| {
                 // Only data_header.size bytes stored in the blockstore so
                 // pad the payload out to SHRED_PAYLOAD_SIZE so that the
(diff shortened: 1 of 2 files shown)


ledger/src/blockstore.rs
```diff
@@ -1484,11 +1487,16 @@ impl Blockstore {
...
+
     pub fn get_data_shreds_for_slot(
         &self,
         slot: Slot,
         start_index: u64,
     ) -> ShredResult<Vec<Shred>> {
+        use crate::shred::SHRED_PAYLOAD_SIZE;
         self.slot_data_iterator(slot, start_index)
             .expect("blockstore couldn't fetch iterator")
             .map(|data| Shred::new_from_serialized_shred(data.1.to_vec()))
```
steviez, line 1502: @carllin - Went back and forth on this for a little bit so figured I'd just open the PR to open up discussion.

The other entrypoints for getting shreds all funnel down nicely to one function. However, this entrypoint that uses an iterator doesn't so we'd potentially have to have repeated logic. Moreover, `slot_data_iterator()` itself is a public function, so someone could use that method directly (and incorrectly).

I'm kind of leaning towards duplicate logic to resize since we know this might be going away soon, but curious if you (or anyone else that happens to see this) have any thoughts on this

ledger/src/blockstore.rs
```diff
@@ -1484,14 +1487,26 @@ impl Blockstore {
...
+        use crate::shred::SHRED_PAYLOAD_SIZE;
         self.slot_data_iterator(slot, start_index)
             .expect("blockstore couldn't fetch iterator")
-            .map(|data| Shred::new_from_serialized_shred(data.1.to_vec()))
+            .map(|data| {
+                let mut payload = data.1.to_vec();
+                // Only data_header.size bytes stored in the blockstore so
+                // pad the payload out to SHRED_PAYLOAD_SIZE so that the
+                // erasure recovery works properly.
+                payload.resize(cmp::max(payload.len(), SHRED_PAYLOAD_SIZE), 0);
```
steviez, line 1507: I'm questioning why we use `cmp::max(payload.len(), SHRED_PAYLOAD_SIZE)` and not just `SHRED_PAYLOAD_SIZE`; we explicitly check for "oversized" shreds in `should_insert_data_shred()` so there `payload.len() > SHRED_PAYLOAD_SIZE` should be impossible
```
    fn should_insert_data_shred(...) -> bool {
    ...
        if shred.payload.len() > SHRED_PAYLOAD_SIZE {
            ...
            return false;
```

The check also occurs in `is_shred_duplicate()` and `fetch_data_shred()`

ledger/src/shred.rs
```diff
@@ -314,16 +314,16 @@ impl Shred {
...
 
         let slot = common_header.slot;
-        // Shreds should be padded out to SHRED_PAYLOAD_SIZE
-        // so that erasure generation/recovery works correctly
-        // But only the data_header.size is stored in blockstore.
-        payload.resize(SHRED_PAYLOAD_SIZE, 0);
+        // Data hreds should be padded out to SHRED_PAYLOAD_SIZE so that erasure
+        // generation/recovery works correctly. Check uses >= because a full packet
+        // payload (including the nonce) could be passed in
+        assert!(payload.len() >= SHRED_PAYLOAD_SIZE);
```
steviez, line 326: Looked a little bit more into what we should do here (resize / assertion / etc) after working on something else for a bit. As Carl pointed out, blockstore isn't the only caller of this function. One example of a [non blockstore caller is here](https://github.com/solana-labs/solana/blob/master/core/src/window_service.rs#L250-L255) in `window_service.rs`:
```
                            // shred fetch stage should be sending packets
                            // with sufficiently large buffers. Needed to ensure
                            let serialized_shred = packet.data.to_vec();
                            if let Ok(shred) = Shred::new_from_serialized_shred(serialized_shred) {
```
Quick refresher, `SHRED_PAYLOAD_SIZE == PACKET_DATA_SIZE - SIZE_OF_NONCE`.

In this case, the incoming buffer has the nonce (or at least the space for the nonce) included. If that shred is to subsequently be placed into the blockstore, we do need to drop the "nonce bytes" since we have a check in blockstore where shreds that are too large aren't inserted.

Going to mull on this in the back of my mind and come back to it a bit later - we need a resize / truncate for sure to handle the case I just mentioned, debatable whether we need the assertion.

For reference, here is what we had before:
```
        let expected_data_size = SHRED_PAYLOAD_SIZE;
        // Safe because any payload from the network must have passed through
        // window service,  which implies payload wll be of size
        // PACKET_DATA_SIZE, and `expected_data_size` <= PACKET_DATA_SIZE.
        //
        // On the other hand, if this function is called locally, the payload size should match
        // the `expected_data_size`.
        assert!(payload.len() >= expected_data_size);
        payload.truncate(expected_data_size);
```

PR #17192, opened 2021-05-12
metrics for generating index time

jeffwashington:
#### Problem
Validator startup takes a while to build indexes. It would be useful to have a metric on how long index generation takes. Developers can spend a lot of time generating indexes during daily work.
#### Summary of Changes
add a metric
Fixes #


carllin:
Do we care about a more fine grained breakdown between the

storage entry scan:
https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_db.rs#L5000-L5017

and the account index insert: https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_db.rs#L5022-L5043

jeffwashington:
> Do we care about a more fine grained breakdown

This could be useful. I modified the pr to add scan time as a separate metric.

jeffwashington:
from mnb on colo:
```datapoint: generate_index total_us=87897034i scan_stores_us=24172162i```

PR #17193, opened 2021-05-12
Return error for excluded secondary-index keys

CriesofCarrots:
#### Problem
As of https://github.com/solana-labs/solana/pull/17110, nodes running with secondary account indexes can explicitly exclude/include keys from the index. If a key is excluded, a full account scan will run. But since these keys are being excluded explicitly because there are too many indexed values for callers to handle, a full account scan is impractical -- it can only perform worse.

#### Summary of Changes
- Add custom rpc error for excluded keys
- Check keys for exclusion in Rpc before attempting indexed scan -- This operation could be done here: https://github.com/solana-labs/solana/blob/597373f5fa8037cee1a1241e6dc3bf1f9e1631d7/runtime/src/accounts_db.rs#L2283 where we are already doing the check. However this would require a bunch of churn to change type signatures from Vec to Result, and anyway it seems more reasonable for the rpc service to decide whether to support the full-accounts scan, instead of a hard block in the runtime.

### DIFF
--- a/core/src/rpc.rs
+++ b/core/src/rpc.rs
@@ -46,7 +46,7 @@ use solana_metrics::inc_new_counter_info;
 use solana_perf::packet::PACKET_DATA_SIZE;
 use solana_runtime::{
     accounts::AccountAddressFilter,
-    accounts_index::{AccountIndex, IndexKey},
+    accounts_index::{AccountIndex, AccountSecondaryIndexes, IndexKey},
     bank::Bank,
     bank_forks::{BankForks, SnapshotConfig},
     commitment::{BlockCommitmentArray, BlockCommitmentCache, CommitmentSlots},
@@ -125,7 +125,7 @@ pub struct JsonRpcConfig {
     pub enable_bigtable_ledger_storage: bool,
     pub enable_bigtable_ledger_upload: bool,
     pub max_multiple_accounts: Option<usize>,
-    pub account_indexes: HashSet<AccountIndex>,
+    pub account_indexes: AccountSecondaryIndexes,
     pub rpc_threads: usize,
     pub rpc_bigtable_timeout: Option<Duration>,
     pub minimal_api: bool,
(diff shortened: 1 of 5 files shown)


CriesofCarrots:
@mvines , if you wouldn't mind, does the rpc error seem satisfactory?

core/src/rpc.rs
```diff
@@ -1746,13 +1755,22 @@ impl JsonRpcRequestProcessor {
             .account_indexes
```
jeffwashington, line 1755: I think the way you did this is ok. However, the .account_indexes is specified in the config to rpc also from validator. I specifically had to chop out the include/exclude info. So, you could just keep a copy of the include/exclude info in rpc directly. There are 2 sources of truth then, but we already do that with account_indexes.

jeffwashington: [here](https://github.com/solana-labs/solana/blob/597373f5fa8037cee1a1241e6dc3bf1f9e1631d7/validator/src/main.rs#L2119) is where I exclude the include/exclude info

CriesofCarrots: Ah, I failed to track where the verbose/concise account_indexes were split. Yeah, let's do that; saves any runtime helpers needed.

CriesofCarrots: @jeffwashington done here: https://github.com/solana-labs/solana/pull/17193/commits/32118afe7511c600371a45f9b947727ec0dbc4b8
Can you take a look?

I considered changing the type to `Arc<AccountSecondaryIndexes>` to reduce to only one source of truth, but the level of churn was very high. Let me know if you prefer that, though.

jeffwashington: perfect.
I prefer not an arc myself. The secondary config is created once at startup and never changed and looked up on each modification to the accounts index. 2 copies is ok with me.

jeffwashington (approved):
lgtm

PR #17194, opened 2021-05-12
hold lock to speed up insert

jeffwashington:
#### Problem
#### Summary of Changes
Hold the accounts index lock while many inserts are done.
Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -5034,22 +5034,33 @@ impl AccountsDb {
                         let dirty_keys =
                             accounts_map.iter().map(|(pubkey, _info)| *pubkey).collect();
                         self.uncleaned_pubkeys.insert(*slot, dirty_keys);
-                        for (pubkey, account_infos) in accounts_map.into_iter() {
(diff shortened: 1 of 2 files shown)


jeffwashington:
10s faster to load validator
 master:INFO  solana_metrics::metrics] datapoint: generate_index total_us=74934823i scan_stores_us=21244497i
 this change: solana_metrics::metrics] datapoint: generate_index total_us=64532862i scan_stores_us=23246957i

jeffwashington:
@carllin still tests to write. What do you think of this idea?

carllin:
> @carllin still tests to write. What do you think of this idea?

I think it's a good one,  it stood out when looking at the results of your latest metrics as well 😃 

Can't argue with the results!

jeffwashington:
I reworked it again and ran it on lemond instead of wiggins. Different cpu, mem, snapshot, etc.

with change:
 generate_index total_us=47022182i scan_stores_us=22914989i

master:
generate_index total_us= 81614843i scan_stores_us=24490653i

sakridge (approved):
lgtm

PR #3361, opened 2019-03-19
Simplify bank_forks, add squashing and pruning

carllin:
#### Problem
1) Removing parents of new banks from bank_forks on insert() makes using bank_forks clumsy whenever we want to perform operations/searches on a set of banks (always have to search parents of the stored set).

2) Banks are not squashed after max_lockout is reached in locktower

#### Summary of Changes
1) Stop pruning parents in bank_forks() on insert
2) Add squashing and pruning logic to bank_forks after we reach max lockout on a slot.

Thanks @rob-solana for the discussion.

Fixes #

PR #17213, opened 2021-05-13
Add documentation for BankForks::prune_non_rooted()

brooksprumo:
No functional changes, just doc comments.

While in the process of learning about snapshots and reading through `BankForks::set_root()`, I wanted to understand `prune_non_rooted()`. I needed to draw out this diagram to visualize and understand it, so I figure it may be helpful for others too.

### DIFF
--- a/runtime/src/bank_forks.rs
+++ b/runtime/src/bank_forks.rs
@@ -294,6 +294,7 @@ impl BankForks {
     ///
     /// Given the following banks and slots...
     ///
+    /// ```text
     /// slot 6                   * (G)
     ///                         /
     /// slot 5        (F)  *   /
@@ -307,6 +308,7 @@ impl BankForks {
     /// slot 1          \  * (B)
     ///                  \ |
     /// slot 0             * (A)  <-- highest confirmed root [1]
+    /// ```
     ///
     /// ...where (D) is set as root, clean up (C) and (E), since they are not rooted.
     ///
@@ -321,6 +323,7 @@ impl BankForks {
     ///
     /// and in table form...
     ///
+    /// ```text
     ///       |          |  is root a  | is a descendant ||
     ///  slot | is root? | descendant? |    of root?     || keep?
     /// ------+----------+-------------+-----------------++-------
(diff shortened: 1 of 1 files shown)


brooksprumo:
@behzadnouri Added you as a reviewer since you were the last one to touch this code. Can you double check my doc comment is correct?

behzadnouri:
> @behzadnouri Added you as a reviewer since you were the last one to touch this code. Can you double check my doc comment is correct?

cc @carllin who I think added this function in https://github.com/solana-labs/solana/pull/3361

jeffwashington (approved):

runtime/src/bank_forks.rs
```diff
@@ -290,7 +290,33 @@ impl BankForks {
...
+    /// slot 3        |    * (D) <-- root, from set_root()
+    ///               |    |
+    /// slot 2    (C) *    |
+    ///                \   |
+    /// slot 1          \  * (B)
+    ///                  \ |
+    /// slot 0             * (A)  <-- highest confirmed root
+    ///
+    /// ...where (D) is set as root, clean up (C) and (E), since they are not rooted.
+    /// (A) is kept because it is a lower slot than (D), and (D) is one of its decendants
```
carllin, line 312: nit: decendants-> descendants

runtime/src/bank_forks.rs
```diff
@@ -290,7 +290,33 @@ impl BankForks {
...
+    /// (B) is kept for the same reason as (A)
+    /// (C) is pruned since it is a lower slot than (D), but (D) is _not_ one of its decendants
+    /// (D) is kept since it is the root
+    /// (E) is pruned since it is not a decendant of (D)
```
carllin, line 316: nit: decendant -> descendant

runtime/src/bank_forks.rs
```diff
@@ -290,7 +290,33 @@ impl BankForks {
...
+    /// (F) is kept since it is a decendant of (D)
```
carllin, line 317: nit: decendant-> descendant

runtime/src/bank_forks.rs
carllin: nit: Also add because `A, B` are greater than `highest confirmed root`. We keep these around because RPC has a commitment level that is based on the cluster-confirmed root, which is a stronger commitment level than just a local root

carllin (approved):

PR #17214, opened 2021-05-13
Fix clippy on newer +nightly

brooksprumo:
Newer clippy from +nightly have new errors about where annotations can and
cannot go.  This commit fixes two of them.


mvines (approved):
👮🏼

PR #17217, opened 2021-05-13
Fix a bug in input deserialization in the C SDK

machacekch:
#### Problem
When the input contains more accounts than the user has requested to be deserialized, and one of the excess ones is a dup, the input pointer is not adjusted correctly.

Compare the lines added by this commit to line 401: `input += 7; // padding`. Since the input data layout does not depend on the number of accounts the user wants to deserialize, this adjustment by 7 bytes must happen in both branches.

#### Summary of Changes
Adjust the `input` pointer by 7 bytes whether the user cares about the account or not.

### DIFF
--- a/sdk/bpf/c/inc/solana_sdk.h
+++ b/sdk/bpf/c/inc/solana_sdk.h
@@ -345,6 +345,8 @@ static bool sol_deserialize(
         input += MAX_PERMITTED_DATA_INCREASE;
         input = (uint8_t*)(((uint64_t)input + 8 - 1) & ~(8 - 1)); // padding
         input += sizeof(uint64_t);
+      } else {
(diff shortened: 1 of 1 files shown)


machacekch:
Looks like a general problem with the pipeline, unrelated to the change.

jackcmay:
What is the unrelated issue?

I'm working up a test case for this issue

jackcmay:
Ah, you are referring to CI, I kicked it

mvines:
lgtm, @jackcmay any concerns landing so we can get this into 1.6.9 (cc: @CriesofCarrots)

jackcmay:
I'd like to see some tests to verify it but that could follow, I have limited service today so won't get to it asap

PR #17227, opened 2021-05-14
Propagate dead slots up to replay

carllin:
#### Problem
1) The state machine `cluster_slot_state_verifier` in replay needs dead slots in the form of banks to make decisions about duplicate slot (detecting a confirmed version, dumping their descendants, repairing the correct version, etc.). However, currently `get_slots_since()` currently ignores dead slots if they were marked dead from a previous replay (and then the validator restarted for instance), so those slots never become banks.

2) Another issue is blockstore processor currently replays banks, but if they error/are marked dead, it continues here: https://github.com/solana-labs/solana/blob/6e9deaf1bd54b20d5ab07c90d6f484c2232c61cf/ledger/src/blockstore_processor.rs#L951, which means those banks are dropped and not added to `initial_forks ` via the following call https://github.com/solana-labs/solana/blob/6e9deaf1bd54b20d5ab07c90d6f484c2232c61cf/ledger/src/blockstore_processor.rs#L1027, and don't end up in the resulting  `BankForks`.  However, the call to `new_from_parent()` while constructing these dead banks *already* modified the account state by adding sysvars to the accounts state for that dead slot.

This inconsistency means if we then try to reconstruct the dead banks during replay via another call to `new_from_parent()`, we encounter errors/conflicts with that existing state. 

#### Summary of Changes
Blockstore processor now returns the tips of dead slots unless they have been outdated by a root, and then clean will sweep up all the dead state once a new root is set

Fixes #

### DIFF
--- a/ledger/src/blockstore.rs
+++ b/ledger/src/blockstore.rs
@@ -2755,12 +2755,7 @@ impl Blockstore {
         let result: HashMap<u64, Vec<u64>> = slots
             .iter()
             .zip(slot_metas)
-            .filter_map(|(height, meta)| {
-                meta.map(|meta| {
-                    let valid_next_slots: Vec<u64> = meta.next_slots.to_vec();
-                    (*height, valid_next_slots)
-                })
-            })
+            .filter_map(|(height, meta)| meta.map(|meta| (*height, meta.next_slots.to_vec())))
             .collect();
 
         Ok(result)
--- a/ledger/src/blockstore.rs
+++ b/ledger/src/blockstore.rs
@@ -2757,12 +2757,7 @@ impl Blockstore {
             .zip(slot_metas)
             .filter_map(|(height, meta)| {
                 meta.map(|meta| {
-                    let valid_next_slots: Vec<u64> = meta
-                        .next_slots
-                        .iter()
-                        .cloned()
-                        .filter(|s| !self.is_dead(*s))
-                        .collect();
+                    let valid_next_slots: Vec<u64> = meta.next_slots.to_vec();
                     (*height, valid_next_slots)
                 })
             })
--- a/ledger/src/blockstore_processor.rs
+++ b/ledger/src/blockstore_processor.rs
@@ -549,7 +549,6 @@ fn do_process_blockstore_from_root(
             ""
         },
     );
-    assert!(bank_forks.active_banks().is_empty());
 
     // We might be promptly restarted after bad capitalization was detected while creating newer snapshot.
     // In that case, we're most likely restored from the last good snapshot and replayed up to this root.
@@ -698,7 +697,6 @@ pub fn confirm_slot(
     allow_dead_slots: bool,
 ) -> result::Result<(), BlockstoreProcessorError> {
     let slot = bank.slot();
-
     let (entries, num_shreds, slot_full) = {
         let mut load_elapsed = Measure::start("load_elapsed");
         let load_result = blockstore
@@ -869,11 +867,6 @@ fn process_next_slots(
                     .unwrap(),
                 *next_slot,
             ));
-            trace!(
-                "New bank for slot {}, parent slot is {}",
-                next_slot,
-                bank.slot(),
-            );
             pending_slots.push((next_meta, next_bank, bank.last_blockhash()));
         }
     }
@@ -942,8 +935,7 @@ fn load_frozen_forks(
             }
 
             let mut progress = ConfirmationProgress::new(last_entry_hash);
-
-            if process_single_slot(
+            let process_result = process_single_slot(
                 blockstore,
                 &bank,
                 opts,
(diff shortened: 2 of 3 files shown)


ledger/src/blockstore.rs
```diff
@@ -2755,12 +2755,7 @@ impl Blockstore {
...
             .filter_map(|(height, meta)| {
                 meta.map(|meta| {
-                    let valid_next_slots: Vec<u64> = meta
-                        .next_slots
-                        .iter()
-                        .cloned()
-                        .filter(|s| !self.is_dead(*s))
-                        .collect();
+                    let valid_next_slots: Vec<u64> = meta.next_slots.to_vec();
                     (*height, valid_next_slots)
```
ryoqun, line 2759: nit: fold into `(*height, meta.next_slots.to_vec())`?

ledger/src/blockstore_processor.rs
```diff
@@ -945,17 +937,24 @@ fn load_frozen_forks(
...
+            if let Err(e) = process_result {
+                warn!("processing single slot {} error {:?}", slot, e);
+                initial_forks.insert(bank.slot(), bank.clone());
+                all_banks.insert(bank.slot(), bank.clone());
                 continue;
             }
             txs += progress.num_txs;
 
             // Block must be frozen by this point, otherwise `process_single_slot` would
             // have errored above
```
ryoqun, line 956: nits: `// Block must be frozen by this point because any errors should have been bailed out above.`

ledger/src/blockstore.rs
```diff
@@ -2755,12 +2755,7 @@ impl Blockstore {
             .zip(slot_metas)
```
ryoqun, line 2755: it looks like `get_slots_since` isn't used for rpc. So, it's only used by replay stage.

ledger/src/blockstore_processor.rs
```diff
@@ -945,17 +937,24 @@ fn load_frozen_forks(
...
+            // Insert even dead banks into the BankForks so that cleanup
+            // of account state created on Bank creation via `Bank::new_from_parent()`
+            // will occur when the bank is dropped. Otherwise, this state is not
+            // cleaned
+            all_banks.insert(bank.slot(), bank.clone());
+
+            if let Err(e) = process_result {
+                warn!("processing single slot {} error {:?}", slot, e);
+                initial_forks.insert(bank.slot(), bank.clone());
+                all_banks.insert(bank.slot(), bank.clone());
```
ryoqun, line 950: maybe, this `insert()` into `all_banks` is redundant because it's already done at line 945?

ledger/src/blockstore_processor.rs
```diff
@@ -945,17 +937,24 @@ fn load_frozen_forks(
                 cache_block_time_sender,
                 None,
                 timing,
-            )
-            .is_err()
-            {
+            );
+            // Insert even dead banks into the BankForks so that cleanup
+            // of account state created on Bank creation via `Bank::new_from_parent()`
+            // will occur when the bank is dropped. Otherwise, this state is not
```
ryoqun, line 943: So, `purge_slot()` called inside `Bank::drop()` isn't enough for `cleanup of account state`?

ledger/src/blockstore_processor.rs
```diff
@@ -934,8 +927,7 @@ fn load_frozen_forks(
             }
```
ryoqun, line 927: how about renaming this fn to `load_fork_tip_banks()` because it can return unfrozen banks now?

ryoqun: also, for `solana-ledger-tool`, I think it's better to retain the old behavior somehow?

local-cluster/tests/local_cluster.rs
```diff
@@ -2122,9 +2122,14 @@ fn test_optimistic_confirmation_violation_detection() {
...
-        // marking this voted slot as dead makes the saved tower garbage
-        // effectively. That's because its stray last vote becomes stale (= no
-        // ancestor in bank forks).
+        // Necessary otherwise tower will inform this validator that it's latest
+        // vote is on slot `prev_voted_slot`. This will then prevent this validator
+        // from resetting to the parent of `prev_voted_slot` to create an alternative fork because
+        // 1) Validator can't vote on earlier ancestor of last vote due to switch threshold (can't vote
+        // on ancestors of last vote)
+        // 2) Won't reset to this earlier ancestor becasue reset can only happen on same voted fork if
+        // it's for the last vote slot or later
```
ryoqun, line 2131: :+1:

ledger/src/blockstore_processor.rs
ryoqun: oh, maybe this?: https://github.com/solana-labs/solana/pull/17269

carllin: @ryoqun, hehe good eye. Yeah, it used to only purge the storage entries, not the accounts index. Maybe with this change here: https://github.com/solana-labs/solana/pull/17319, this change in blockstoreprocessor can now be avoided

carllin: @ryoqun with https://github.com/solana-labs/solana/pull/17319, I think this change is no longer necessary, so reverted the `blockstore_processor` changes!

jstarry (approved):

PR #17232, opened 2021-05-14
rework slot list update

jeffwashington:
#### Problem
In general, there is contention on account index updates.

#### Summary of Changes
Simplify accounts_index.update for clarity, reduced iterations, collapse redundant operations.

Fixes #


jeffwashington:
discussed with @carllin in context of all that is going on.

sakridge (approved):
looks good to me.

PR #17236, opened 2021-05-14
bumps up min number of bloom items in gossip pull requests

behzadnouri:
#### Problem
When a validator starts, it has an (almost) empty crds table and it only sends one pull-request to the entrypoint. The bloom filter in the pull-request targets 10% false rate given the number of items. So, if the `num_items` is very wrong, it makes a very small bloom filter with a very high false rate:
https://github.com/solana-labs/solana/blob/2ae57c172/runtime/src/bloom.rs#L70-L80
https://github.com/solana-labs/solana/blob/2ae57c172/core/src/crds_gossip_pull.rs#L48

As a result, it is very unlikely that the validator obtains entrypoint's contact-info in response. This exacerbates how long the validator will loop on:
> Waiting to adopt entrypoint shred version

https://github.com/solana-labs/solana/blob/ed51cde37/validator/src/main.rs#L390-L412

#### Summary of Changes
This commit increases the min number of bloom items when making gossip pull requests. Effectively this will break the entrypoint crds table into 64 shards, one pull-request for each, a larger bloom filter for each shard, and increases the chances that the response will include entrypoint's contact-info, which is needed for adopting shred version and validator start.

### DIFF
--- a/core/src/crds_gossip_pull.rs
+++ b/core/src/crds_gossip_pull.rs
@@ -13,7 +13,7 @@ use crate::{
     cluster_info::{Ping, CRDS_UNIQUE_PUBKEY_CAPACITY},
     contact_info::ContactInfo,
     crds::{Crds, CrdsError},
-    crds_gossip::{get_stake, get_weight, CRDS_GOSSIP_DEFAULT_BLOOM_ITEMS},
+    crds_gossip::{get_stake, get_weight},
     crds_gossip_error::CrdsGossipError,
     crds_value::CrdsValue,
     ping_pong::PingCache,
@@ -30,7 +30,6 @@ use solana_sdk::{
     signature::{Keypair, Signer},
 };
 use std::{
-    cmp,
     collections::{HashMap, HashSet, VecDeque},
     convert::TryInto,
     net::SocketAddr,
@@ -468,11 +467,10 @@ impl CrdsGossipPull {
         bloom_size: usize,
     ) -> Vec<CrdsFilter> {
         const PAR_MIN_LENGTH: usize = 512;
-        let num = cmp::max(
-            CRDS_GOSSIP_DEFAULT_BLOOM_ITEMS,
-            crds.len() + self.purged_values.len() + self.failed_inserts.len(),
-        );
-        let filters = CrdsFilterSet::new(num, bloom_size);
+        const MIN_NUM_BLOOM_ITEMS: usize = 65_536;
+        let num_items = crds.len() + self.purged_values.len() + self.failed_inserts.len();
+        let num_items = MIN_NUM_BLOOM_ITEMS.max(num_items);
+        let filters = CrdsFilterSet::new(num_items, bloom_size);
         thread_pool.install(|| {
             crds.par_values()
                 .with_min_len(PAR_MIN_LENGTH)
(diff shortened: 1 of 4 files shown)


carllin:
Hmmm, so if there are no entries in the `Crds` table, then only one filter will be built in the `CrdsFilterSet with `mask_bits = 0`. Then this gets made into a `CrdsFilter` here with `mask_bits = 0` and `mask = !0x0`. 

My question is, when the entrypoint gets this message, do they not respond with everything in their table because this mask with `mask_bits=0` shouldn't filter out anything?

behzadnouri:
> My question is, when the entrypoint gets this message, do they not respond with everything in their table because this mask with `mask_bits=0` shouldn't filter out anything?

So, the `mask_bits` effectively only breaks the entrypoint's crds table into a number of shards (and with `mask_bits = 0` there will be only one shard for the entire table). Then the bloom filters are checked to filter out what values the validator already has:
https://github.com/solana-labs/solana/blob/2ae57c172/core/src/crds_gossip_pull.rs#L525-L537

The bloom filter target 10% false rate given the number of items. So, if the `num_items` is very wrong, it makes a very small bloom filter with a very high false rate:
https://github.com/solana-labs/solana/blob/2ae57c172/runtime/src/bloom.rs#L70-L80
https://github.com/solana-labs/solana/blob/2ae57c172/core/src/crds_gossip_pull.rs#L48

The validator always have a couple of items in its crds table (its own contact-info, version, node-instance). So the bloom filter always have some bits set causing the false collisions.

I realize the description of this pull-request is not good; I will update the description as well.

carllin:
@behzadnouri Ah I see, the bloom is built for a 10% false rate given very little expected default capacity, so here we are bumping the expected default capacity, thanks for the explanation!

core/src/crds_gossip_pull.rs
```diff
@@ -468,11 +467,10 @@ impl CrdsGossipPull {
         bloom_size: usize,
     ) -> Vec<CrdsFilter> {
         const PAR_MIN_LENGTH: usize = 512;
-        let num = cmp::max(
-            CRDS_GOSSIP_DEFAULT_BLOOM_ITEMS,
-            crds.len() + self.purged_values.len() + self.failed_inserts.len(),
-        );
-        let filters = CrdsFilterSet::new(num, bloom_size);
+        const MIN_NUM_BLOOM_ITEMS: usize = 65_536;
```
carllin, line 470: Will this be a moving target as the number of validators on the network increases?

carllin (approved):

core/src/crds_gossip_push.rs
```diff
@@ -264,46 +263,44 @@ impl CrdsGossipPush {
...
+                break;
             }
+            if self.active_set.contains_key(&peer) || new_items.contains_key(&peer) {
+                continue;
+            }
+            let bloom = AtomicBloom::from(Bloom::random(
+                num_bloom_items,
+                BLOOM_FALSE_RATE,
+                BLOOM_MAX_BITS,
+            ));
```
carllin, line 301: thanks for cleaning this up!

core/src/crds_gossip_pull.rs
behzadnouri: We will see, but probably with https://github.com/solana-labs/solana/pull/17261 we should be fine

PR #17238, opened 2021-05-14
prioritizes more recent values in pull responses

behzadnouri:
#### Problem
On the receiving end, the outdated values are discarded, and they will
only waste bandwidth:
https://github.com/solana-labs/solana/blob/3f0480d06/core/src/crds_gossip_pull.rs#L385-L400

This is also exacerbating validator start, since the entrypoint is
returning old values in pull responses, and the validator immediately
discards those; resulting in huge delay until the validator obtains
contact-info of the entrypoint and is able to adopt shred-version and
fully start.

#### Summary of Changes
prioritize more recent values in pull responses

### DIFF
--- a/core/src/cluster_info.rs
+++ b/core/src/cluster_info.rs
@@ -70,8 +70,9 @@ use std::{
     fmt::Debug,
     fs::{self, File},
     io::BufReader,
+    iter::repeat,
     net::{IpAddr, Ipv4Addr, SocketAddr, TcpListener, UdpSocket},
-    ops::{Deref, DerefMut},
+    ops::{Deref, DerefMut, Div},
     path::{Path, PathBuf},
     sync::{
         atomic::{AtomicBool, Ordering},
@@ -426,17 +427,6 @@ impl Sanitize for Protocol {
             Protocol::PongMessage(pong) => pong.sanitize(),
         }
     }
-}
-
-// Rating for pull requests
-// A response table is generated as a
-// 2-d table arranged by target nodes and a
-// list of responses for that node,
-// to/responses_index is a location in that table.
-struct ResponseScore {
-    to: usize,              // to, index of who the response is to
-    responses_index: usize, // index into the list of responses for a given to
-    score: u64,             // Relative score of the response
 }
 
 // Retains only CRDS values associated with nodes with enough stake.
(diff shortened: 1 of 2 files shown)


core/src/cluster_info.rs
```diff
@@ -2089,56 +2080,42 @@ impl ClusterInfo {
...
-                        }
-                    })
-                    .collect::<Vec<ResponseScore>>()
+            .zip(pull_responses)
+            .flat_map(|(addr, responses)| repeat(addr).zip(responses))
+            .map(|(addr, response)| {
+                let age = now.saturating_sub(response.wallclock());
+                let score = DEFAULT_EPOCH_DURATION_MS
+                    .saturating_sub(age)
+                    .div(CRDS_GOSSIP_PULL_CRDS_TIMEOUT_MS)
```
carllin, line 2091: Double checking, are most Crds values expected to be refreshed within `CRDS_GOSSIP_PULL_CRDS_TIMEOUT_MS`, timeout?

carllin: Also, any concerns that somebody could finegle their wallclock to be further ahead and have a higher score?

behzadnouri: > Also, any concerns that somebody could finegle their wallclock to be further ahead and have a higher score?

Good point. If the value's wallclock exceeds caller's wallclock, it will be filtered out:
https://github.com/solana-labs/solana/blob/2ae57c172/core/src/crds_gossip_pull.rs#L529

and, the `.div(CRDS_GOSSIP_PULL_CRDS_TIMEOUT_MS)` in this change will mask out small differences in wallclocks. So I think it should be fine. Nodes are just better off to use the right wallclock.

> are all Crds values expected to be refreshed within CRDS_GOSSIP_PULL_CRDS_TIMEOUT_MS timeout, or are there some that are refreshed less often

Not sure what you mean by refresh. Only, contact-info and node-instance are refreshed periodically:
https://github.com/solana-labs/solana/blob/2ae57c172/core/src/cluster_info.rs#L546-L549

carllin: Oh I was mostly concerned if some older values from higher staked nodes would now be outprioritized by later messages by lower staked nodes. But I think you're right, the message from a staked validator would have to half an epoch old to be outprioritized  by a newer response from an unstaked node

carllin (approved):

core/src/cluster_info.rs
behzadnouri: There is also a score bump for values from staked nodes.
Potentially we can increase that if needed:
https://github.com/solana-labs/solana/blob/e8b35a4f7/core/src/cluster_info.rs#L2102-L2106

PR #17239, opened 2021-05-14
feat: add borsh utilities and public key support

joncinque:
#### Problem

People want to use borsh js for their projects, but the solana web3.js PublicKey isn't a supported Borsh type, which makes dapp integrations annoying.  They have to constantly reimplement the public key type in order to do anything.

#### Summary of Changes

Now that the new borsh js version is out, we can fully support Borsh!  The PublicKey is a `u256` BN under the hood, so we can now define it as a borsh type.  This PR adds a new constructor for PublicKey for Borsh support, and adds a schema that dapps can reuse as they add their own types.  The `Enum` and `Struct` types are meant for dapps, to easily redefine their types without having to dig through the borsh js source code.

Fixes #

### DIFF
--- a/web3.js/package-lock.json
+++ b/web3.js/package-lock.json
@@ -11,6 +11,7 @@
       "dependencies": {
         "@babel/runtime": "^7.12.5",
         "bn.js": "^5.0.0",
+        "borsh": "^0.4.0",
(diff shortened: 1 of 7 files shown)


oJshua (approved):
Looks great!

PR #17245, opened 2021-05-14
calculate account infos before lock

jeffwashington:
#### Problem

#### Summary of Changes
Calculate intermediate data before grabbing lock.

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -5034,15 +5034,29 @@ impl AccountsDb {
                         let dirty_keys =
                             accounts_map.iter().map(|(pubkey, _info)| *pubkey).collect();
                         self.uncleaned_pubkeys.insert(*slot, dirty_keys);
+
+                        let infos: Vec<_> = accounts_map
+                            .iter()
+                            .map(|(pubkey, account_infos)| {
+                                account_infos
+                                    .iter()
+                                    .map(|(_, (store_id, stored_account))| {
+                                        (
+                                            pubkey,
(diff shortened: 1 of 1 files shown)


jeffwashington:
```
generate_index total_us=42580823i scan_stores_us=22309550i (with this change)
generate_index total_us=47632492i scan_stores_us=22522726i (master)
```
lemond colo machine, snapshot-77779424-AP7QKgrw31e5tBhtPqxejbJbYU8rs6LA97L1Qq7gqpdX.tar.zst

runtime/src/accounts_db.rs
```diff
@@ -5034,15 +5034,29 @@ impl AccountsDb {
...
+                                                store_id: *store_id,
+                                                offset: stored_account.offset,
+                                                stored_size: stored_account.stored_size,
+                                                lamports: stored_account.account_meta.lamports,
+                                            },
+                                        )
+                                    })
+                                    .collect::<Vec<_>>()
+                            })
+                            .collect();
```
sakridge, line 5056: bit of a nit, but I thought of doing a `flatten` here. Might be slightly faster?

jeffwashington: It may be slightly faster with a flatten outside the lock here. But, it is about to be much faster with another modification in the next pr that builds on this. Each change should hopefully be easy to verify as correct.

sakridge: ah ok

sakridge (approved):

PR #17246, opened 2021-05-14
don't try to build secondary indexes if we don't need them

jeffwashington:
#### Problem

#### Summary of Changes
Don't try to build secondary indexes if we don't need to.

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -5054,14 +5054,16 @@ impl AccountsDb {
                             }
                         }
                         drop(lock);
-                        for (pubkey, account_infos) in accounts_map.into_iter() {
-                            for (_, (_store_id, stored_account)) in account_infos.iter() {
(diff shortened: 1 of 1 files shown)


carllin (approved):

runtime/src/accounts_db.rs
```diff
@@ -5054,14 +5054,16 @@ impl AccountsDb {
...
-                        for (pubkey, account_infos) in accounts_map.into_iter() {
-                            for (_, (_store_id, stored_account)) in account_infos.iter() {
-                                self.accounts_index.update_secondary_indexes(
-                                    &pubkey,
-                                    &stored_account.account_meta.owner,
-                                    &stored_account.data,
-                                    &self.account_indexes,
-                                );
+                        if !self.account_indexes.is_empty() {
+                            for (pubkey, account_infos) in accounts_map.into_iter() {
```
carllin, line 5058: nice, save us the loop iteration

PR #17247, opened 2021-05-14
generate_index inserts ideal initial data

jeffwashington:
#### Problem

#### Summary of Changes
When inserting new items, calculate the actual btreemap 'value' for initial insertion instead of blank values. Then, skip calling 'update' if the pubkey did not exist and we inserted the correct initial data.

Fixes #

### DIFF
--- a/runtime/src/accounts_index.rs
+++ b/runtime/src/accounts_index.rs
@@ -2430,7 +2430,6 @@ pub mod tests {
     fn test_insert_with_lock_no_ancestors() {
         let key = Keypair::new();
         let index = AccountsIndex::<bool>::default();
-        let mut gc = Vec::new();
         let slot = 0;
         let account_info = true;
 
@@ -2441,11 +2440,8 @@ pub mod tests {
             &mut w_account_maps,
             new_entry,
         );
-        let mut write = write.unwrap();
+        assert!(write.is_none());
         drop(w_account_maps);
-        write.update(slot, account_info, &mut gc);
-        assert!(gc.is_empty());
-        drop(write);
 
         let mut ancestors = Ancestors::default();
         assert!(index.get(&key.pubkey(), Some(&ancestors), None).is_none());
--- a/runtime/src/accounts_index.rs
+++ b/runtime/src/accounts_index.rs
@@ -508,7 +508,7 @@ pub trait ZeroLamport {
 }
 
 type MapType<T> = AccountMap<Pubkey, AccountMapEntry<T>>;
-type ReadWriteLockMapType<'a, T> = RwLockWriteGuard<'a, AccountMap<Pubkey, AccountMapEntry<T>>>;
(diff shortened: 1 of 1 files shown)


jeffwashington:
generate_index total_us=39113701i scan_stores_us=22296476i (this pr)
generate_index total_us=42341212i scan_stores_us=22534402i (master)

sakridge:
Is there test coverage that would catch a change to `update()` would require a corresponding change here?

sakridge:
I think it's questionable that the performance improvement justifies the extra code complexity/risk. Does this build on a further change? The performance increase seems to be ~8%, is that right?

jeffwashington:
> I think it's questionable that the performance improvement justifies the extra code complexity/risk. Does this build on a further change? The performance increase seems to be ~8%, is that right?

[this pr](https://github.com/solana-labs/solana/pull/17290#issuecomment-843189263) builds on the initial data change:
and is now approaching 2x faster overall from master right now

jeffwashington:
> Is there test coverage that would catch a change to `update()` would require a corresponding change here?

I can rework this to make sure that is tested well.

jeffwashington:
discussed with @carllin in context of other changes going on.

runtime/src/accounts_index.rs
```diff
@@ -842,47 +854,55 @@ impl<T: 'static + Clone + IsCached + ZeroLamport> AccountsIndex<T> {
...
-    fn insert_new_entry_if_missing(&self, pubkey: &Pubkey) -> (WriteAccountMapEntry<T>, bool) {
-        let new_entry = Self::new_entry();
-        let mut w_account_maps = self.get_account_maps_write_lock();
-        self.insert_new_entry_if_missing_with_lock(pubkey, &mut w_account_maps, new_entry)
+    fn insert_new_entry_if_missing(
+        &self,
+        pubkey: &Pubkey,
+        slot: Slot,
+        info: &T,
+        lock: Option<&mut ReadWriteLockMapType<T>>,
```
sakridge, line 862: maybe a better name than `lock` here. The name and type don't help understand what this is.

jeffwashington: updated to:
``` 
   fn insert_new_entry_if_missing(
        &self,
        pubkey: &Pubkey,
        slot: Slot,
        info: &T,
        w_account_maps: Option<&mut AccountMapsWriteLock<T>>,
    ) -> Option<WriteAccountMapEntry<T>> {
```

jeffwashington:
updated results:
```
this change:
generate_index total_us=40146488i scan_stores_us=11513893i

master:
generate_index total_us=45073105i scan_stores_us=11413174i
```

sakridge (approved):

jeffwashington:
@carllin fyi. reworked this extensively with stephen.

PR #17254, opened 2021-05-15
Remove duplicate std::net reference in docs

sebasibarguen:
#### Problem
Document for `/developing/on-chain-programs/developing-rust` contains duplicate `std::net` in list.

#### Summary of Changes
Remove one reference of `std::net`

Fixes #


CriesofCarrots:
Thanks

PR #17261, opened 2021-05-16
extends crds values timeouts if stakes are unknown

behzadnouri:
#### Problem
If stakes are unknown, then timeouts will be short, resulting in values
being purged from the crds table, and consequently higher pull-response
load when they are obtained again from gossip. In particular, this slows
down validator start where almost all values obtained from entrypoint
are immediately discarded.

#### Summary of Changes
extend crds values timeouts if stakes are unknown

### DIFF
--- a/core/src/crds_gossip_pull.rs
+++ b/core/src/crds_gossip_pull.rs
@@ -547,26 +547,27 @@ impl CrdsGossipPull {
         inc_new_counter_info!("gossip_filter_crds_values-dropped_values", total_skipped);
         ret
     }
-    pub fn make_timeouts_def(
-        &self,
-        self_id: &Pubkey,
-        stakes: &HashMap<Pubkey, u64>,
-        epoch_ms: u64,
-        min_ts: u64,
-    ) -> HashMap<Pubkey, u64> {
-        let mut timeouts: HashMap<Pubkey, u64> = stakes.keys().map(|s| (*s, epoch_ms)).collect();
-        timeouts.insert(*self_id, std::u64::MAX);
-        timeouts.insert(Pubkey::default(), min_ts);
-        timeouts
-    }
 
-    pub fn make_timeouts(
+    pub(crate) fn make_timeouts(
         &self,
-        self_id: &Pubkey,
+        self_pubkey: Pubkey,
         stakes: &HashMap<Pubkey, u64>,
         epoch_ms: u64,
     ) -> HashMap<Pubkey, u64> {
-        self.make_timeouts_def(self_id, stakes, epoch_ms, self.crds_timeout)
+        let default_timeout = if stakes.values().all(|stake| *stake == 0) {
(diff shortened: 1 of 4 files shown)


core/src/crds_gossip_pull.rs
```diff
@@ -549,26 +549,28 @@ impl CrdsGossipPull {
...
         &self,
-        self_id: &Pubkey,
+        self_pubkey: Pubkey,
         stakes: &HashMap<Pubkey, u64>,
-        epoch_ms: u64,
+        epoch_duration: Duration,
     ) -> HashMap<Pubkey, u64> {
-        self.make_timeouts_def(self_id, stakes, epoch_ms, self.crds_timeout)
+        let extended_timeout = self.crds_timeout.max(epoch_duration.as_millis() as u64);
+        let default_timeout = if stakes.values().all(|stake| *stake == 0) {
```
carllin, line 560: hmmm so the only time we default to the longer epoch timeout is when all the stakes are zero. From reading the PR description, I'm assuming this is supposed to be only at startup?

From what I'm reading, however, some of the stakes should always be nonzero for a validator on a real network because:

1) We always pass a `BankForks` here https://github.com/solana-labs/solana/blob/master/core/src/validator.rs#L600

2) Which passes into GossipService: https://github.com/solana-labs/solana/blob/master/core/src/gossip_service.rs#L66

3) Which means there should be some stake in this `staked_nodes` https://github.com/solana-labs/solana/blob/master/core/src/cluster_info.rs#L1848 which is then passed here to `make_timeouts()`

behzadnouri: Initially, to adopt the shred version from the entrypoint, the validator starts a gossip service with no bank-forks:
https://github.com/solana-labs/solana/blob/ed51cde37/validator/src/main.rs#L369

That is the primary motivation for this change, because then the validator assigns short timeouts to all crds values, and when it gets a response from the entrypoint, most often it discards it right away. (This exacerbates the start-up wait and bandwidth usage because the values previously fetched are not excluded from the next pull request to the entrypoint and they are likely redundantly sent again). 

But, besides that, I see non-zero values for `cluster_info-purge-no_working_bank` metric on mainnet, which then indicates that the stakes were unknown when computing timeouts:
https://github.com/solana-labs/solana/blob/2ae57c172/core/src/cluster_info.rs#L1750
https://github.com/solana-labs/solana/blob/2ae57c172/core/src/cluster_info.rs#L2528

So besides starting validator, I think for an already running validator we are currently hitting cases where stakes are unknown. (Presumably because of a no working bank?!) 

Please note that `cluster_info-purge-no_working_bank` is kind of a misnomer. There are 2 instances of the matric in the code and one is actually using the root bank. This change is updates both places to always consistently use the root-bank.

carllin: thanks for the explanation!

carllin (approved):

PR #17269, opened 2021-05-17
Support out of band dumping of unrooted slots in AccountsDb

carllin:
#### Problem
Follow-up to https://github.com/solana-labs/solana/pull/17319.

Bad versions of duplicate blocks can be replayed, and thus need to be cleared from Accounts state before replaying the good version.

#### Summary of Changes
Support safe block dumping by:

1) Making sure cache flushes for the same block are not happening concurrently via the `RemoveUnrootedSlots` synchronization mechanism which keeps a list of slots being flushed from cache OR being removed by this new dumping path. This dumping path blocks until a flush is completed, whereas the cache flushing path will ignore requests to flush slots that are currently/about to be dumped.
2) Removing the slot completely from both the AccountsIndex and removing the storage entries by calling `purge_slots_from_cache_and_store()`.

Follow-up PR:
Notifying relevant scans that their results may be outdated, and abort those results, noted here:https://github.com/solana-labs/solana/blob/3d0cacc1a617b4344ea44afdb0b51e5627bd2761/runtime/src/accounts_db.rs#L3411

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -843,7 +843,7 @@ pub struct AccountsDb {
     // Set of slots currently being flushed by `flush_slot_cache()` or removed
     // by `remove_unrooted_slot()`. Used to ensure `remove_unrooted_slots(slots)`
     // can safely clear the set of unrooted slots `slots`.
-    remove_unrooted_slots: RemoveUnrootedSlotsSynchronization,
+    remove_unrooted_slots_synchronization: RemoveUnrootedSlotsSynchronization,
 }
 
 #[derive(Debug, Default)]
@@ -1286,7 +1286,7 @@ impl Default for AccountsDb {
             #[cfg(test)]
             load_limit: AtomicU64::default(),
             is_bank_drop_callback_enabled: AtomicBool::default(),
-            remove_unrooted_slots: RemoveUnrootedSlotsSynchronization::default(),
+            remove_unrooted_slots_synchronization: RemoveUnrootedSlotsSynchronization::default(),
         }
     }
 }
@@ -3438,7 +3438,7 @@ impl AccountsDb {
         let RemoveUnrootedSlotsSynchronization {
             slots_under_contention,
             signal,
-        } = &self.remove_unrooted_slots;
+        } = &self.remove_unrooted_slots_synchronization;
 
         {
             // Slots that are currently being flushed by flush_slot_cache()
@@ -3964,25 +3964,25 @@ impl AccountsDb {
     ) -> Option<FlushStats> {
         self.accounts_cache.slot_cache(slot).and_then(|slot_cache| {
             let is_being_purged = {
-                let mut remove_unrooted_slots = self
-                    .remove_unrooted_slots
+                let mut slots_under_contention = self
+                    .remove_unrooted_slots_synchronization
                     .slots_under_contention
                     .lock()
                     .unwrap();
                 // If we're purging this slot, don't flush it here
-                if remove_unrooted_slots.contains(&slot) {
+                if slots_under_contention.contains(&slot) {
                     true
                 } else {
-                    remove_unrooted_slots.insert(slot);
+                    slots_under_contention.insert(slot);
                     false
                 }
             };
             if !is_being_purged {
                 let flush_stats = self.do_flush_slot_cache(slot, &slot_cache, should_flush_f);
                 // Nobody else should have been purging this slot, so should not have been removed
-                // from `self.remove_unrooted_slots`.
+                // from `self.remove_unrooted_slots_synchronization`.
                 assert!(self
-                    .remove_unrooted_slots
+                    .remove_unrooted_slots_synchronization
                     .slots_under_contention
                     .lock()
                     .unwrap()
@@ -3990,7 +3990,9 @@ impl AccountsDb {
 
                 // Signal to any threads blocked on `remove_unrooted_slots(slot)` that we have finished
                 // flushing
-                self.remove_unrooted_slots.signal.notify_all();
+                self.remove_unrooted_slots_synchronization
+                    .signal
+                    .notify_all();
                 Some(flush_stats)
             } else {
                 None
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -3501,7 +3501,7 @@ impl AccountsDb {
                         // Mark that we're about to delete this slot now
                         currently_contended_slots.insert(*flush_slot);
                     }
-                    !is_being_flushed
+                    is_being_flushed
                 });
             }
         }
@@ -10730,6 +10730,24 @@ pub mod tests {
                 .unwrap()
         };
 
+        let exit = Arc::new(AtomicBool::new(false));
+
+        let t_spurious_signal = {
+            let db = db.clone();
+            let exit = exit.clone();
+            std::thread::Builder::new()
+                .name("account-cache-flush".to_string())
+                .spawn(move || loop {
+                    if exit.load(Ordering::Relaxed) {
+                        return;
+                    }
+                    // Simulate spurious wake-up that can happen, but is too rare to
+                    // otherwise depend on in tests.
+                    db.remove_unrooted_slots_synchronization.signal.notify_all();
+                })
+                .unwrap()
+        };
+
         // Run multiple trials. Has the added benefit of rewriting the same slots after we've
         // dumped them in previous trials.
         for _ in 0..num_trials {
@@ -10800,8 +10818,10 @@ pub mod tests {
             }
         }
 
+        exit.store(true, Ordering::Relaxed);
         drop(new_trial_start_sender);
         t_flush_cache.join().unwrap();
+        t_spurious_signal.join().unwrap();
     }
 
     #[test]
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -754,6 +754,9 @@ impl RecycleStores {
     }
 }
 
+/// Removing unrooted slots in Accounts Background Service needs to be synchronized with flushing
+/// slots from the Accounts Cache.  This keeps track of those slots and the Mutex + Condvar for
+/// synchronization.
 #[derive(Debug, Default)]
 struct RemoveUnrootedSlotsSynchronization {
     // slots being flushed from the cache or being purged
@@ -840,9 +843,9 @@ pub struct AccountsDb {
 
     is_bank_drop_callback_enabled: AtomicBool,
 
-    // Set of slots currently being flushed by `flush_slot_cache()` or removed
-    // by `remove_unrooted_slot()`. Used to ensure `remove_unrooted_slots(slots)`
-    // can safely clear the set of unrooted slots `slots`.
+    /// Set of slots currently being flushed by `flush_slot_cache()` or removed
+    /// by `remove_unrooted_slot()`. Used to ensure `remove_unrooted_slots(slots)`
+    /// can safely clear the set of unrooted slots `slots`.
     remove_unrooted_slots_synchronization: RemoveUnrootedSlotsSynchronization,
 }
 
@@ -3424,6 +3427,15 @@ impl AccountsDb {
     }
 
     // TODO: This is currently unsafe with scan because it can remove a slot in the middle
+    /// Remove the set of slots in `remove_slots` from both the cache and storage. This requires
+    /// we know the contents of the slot are either:
+    ///
+    /// 1) Completely in the cache
+    /// 2) Have been completely flushed from the cache
+    ///
+    /// in order to guarantee that when this function returns, the contents of the slot have
+    /// been completely and not only partially removed. Thus synchronization with `flush_slot_cache()`
+    /// through `self.remove_unrooted_slots_synchronization` is necessary.
     pub fn remove_unrooted_slots(&self, remove_slots: &[Slot]) {
         let rooted_slots = self
             .accounts_index
@@ -3954,9 +3966,9 @@ impl AccountsDb {
         }
     }
 
-    // `should_flush_f` is an optional closure that determines whether a given
-    // account should be flushed. Passing `None` will by default flush all
-    // accounts
+    /// `should_flush_f` is an optional closure that determines whether a given
+    /// account should be flushed. Passing `None` will by default flush all
+    /// accounts
     fn flush_slot_cache(
         &self,
         slot: Slot,
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -841,7 +841,8 @@ pub struct AccountsDb {
     is_bank_drop_callback_enabled: AtomicBool,
 
     // Set of slots currently being flushed by `flush_slot_cache()` or removed
-    // by `remove_unrooted_slot()`
+    // by `remove_unrooted_slot()`. Used to ensure `remove_unrooted_slots(slots)`
+    // can safely clear the set of unrooted slots `slots`.
     remove_unrooted_slots: RemoveUnrootedSlots,
 }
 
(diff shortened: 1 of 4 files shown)


runtime/src/accounts_db.rs
```diff
@@ -3396,30 +3408,81 @@ impl AccountsDb {
...
-        // TODO: Handle if the slot was flushed to storage while we were removing the cached
-        // slot above, i.e. it's possible the storage contains partial version of the current
-        // slot. One way to handle this is to augment slots to contain a "version", That way,
-        // 1) We clean older versions via the natural clean() pipeline
-        // without having to call this function out of band.
-        // 2) This deletion doesn't have to block on scan
-        // Reads will then always read the latest version of a slot. Scans will also know
-        // which version their parents because banks will also be augmented with this version,
-        // which handles cases where a deletion of one version happens in the middle of the scan.
+    // TODO: This is currently unsafe with scan because it can remove a slot in the middle
```
jeffwashington, line 3411: What is your plan for this now? Does this unsafeness exist today and you're just calling it out here or does this PR add this unsafeness? Is the unsafeness worth whatever this is fixing? I'm missing some context.
I'm confused with 'currently'. Prior to your change, in master? Or, 'currently' because of your change?

carllin: Yeah, this unsafeness does exist today, but luckily this function isn't called anywhere by the validator 😃 

Yeah it's definitely worth fixing, and is addressed in this next PR here: https://github.com/solana-labs/solana/pull/17471, a peek into the future 👀

runtime/src/accounts_db.rs
```diff
@@ -3794,101 +3857,132 @@ impl AccountsDb {
...
-                        None
-                    }
-                })
-                .unzip();
-
-            let is_dead_slot = accounts.is_empty();
-            // Remove the account index entries from earlier roots that are outdated by later roots.
-            // Safe because queries to the index will be reading updates from later roots.
-            self.purge_slot_cache_pubkeys(
+        let iter_items: Vec<_> = slot_cache.iter().collect();
```
brooksprumo, line 3869: I know this wasn't changed, but is this copy still needed? Could the loop a few lines down use `slot_cache` directly, instead of `iter_items`?

brooksprumo (commented):
OK! I made it through!

I'm still not at a point where I understand all the uses and places when flushes/purges/cleans/removals happen, so I'm not sure if there are other _designs_ that would also work here. But _this_ design looks good to me.

I had one question about a variable copy, but that's it. I can give an Approve after that's resolved.

Also, nice test; pretty gnarly!

runtime/src/accounts_db.rs
```diff
@@ -753,6 +753,13 @@ impl RecycleStores {
     }
 }
 
+#[derive(Debug, Default)]
+struct RemoveUnrootedSlots {
```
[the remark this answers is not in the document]
carllin: Reminder: rename it to `RemoveUnrootedSlotsSynchronization`

runtime/src/accounts_db.rs
carllin: @brooksprumo yeah, I think you're right, it's the more efficient thing to do.

There's a few changes you would have to make though.  The iter() returns locked entries called `RefMulti` from within the `DashMap`. If we don't create a vector of these `RefMulti'`, then for the lifetimes to work, we can't create `let (accounts, hashes): (Vec<(&Pubkey, &AccountSharedData)>, Vec<Hash>)` anymore because the `&Pubkey` references would outlive the `RefMulti` from which they were created here: https://github.com/solana-labs/solana/blob/654918ab27000b71e8af024f374e5d6e3e00cce0/runtime/src/accounts_db.rs#L3813. Instead you would have to clone out the Pubkey.

This then means when we make. this call here: https://github.com/solana-labs/solana/blob/654918ab27000b71e8af024f374e5d6e3e00cce0/runtime/src/accounts_db.rs#L3852, we have have to rework the signature of `store_accounts_frozen` https://github.com/solana-labs/solana/blob/654918ab27000b71e8af024f374e5d6e3e00cce0/runtime/src/accounts_db.rs#L5011 to support taking something like `Borrow<Pubkey>` instead of just `&Pubkey`, so that it can accept both `Pubkey` and `&Pubkey`

Might be good to tackle outside of this PR if somebody wants to play around with this stuff 😃

lijunwangs (commented):
Just some questions.

runtime/src/accounts_db.rs
```diff
@@ -3396,30 +3408,81 @@ impl AccountsDb {
...
+                } else {
+                    // There are no slots being flushed to wait on, so it's safe to continue
+                    // to purging the slots we want to purge!
+                    break;
+                }
+
+                // For each slot the cache flush has finished, mark that we're about to start
+                // purging these slots by reserving it in `contended_slots`.
+                contended_cache_flush_slots.retain(|flushing_slot| {
+                    let is_being_flushed = contended_slots.contains(flushing_slot);
```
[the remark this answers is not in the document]
carllin: @lijunwangs, the `signal.wait(contended_slots).unwrap()` above releases the `contended_slots` and then regrabs the lock when the condition variable gets a signal, see documentation here: https://doc.rust-lang.org/std/sync/struct.Condvar.html#method.wait

>When it wakes up, how come the slot in `contended_cache_flush_slots` is not found in contended_slots
So the flush thread is the thread removing the slots from `contended_slots`, which is the shared variable under the lock. The `contended_cache_flush_slots` is a local snapshot of the slots in `contended_slots` at the time we checked it at the beginning of the function, and is never added to again after we filter from it, so it's always shrinking after each iteration of the `retain` above. We only add a slot to the `contended_slots` through`contended_slots.insert(*flushing_slot)` *after* we've confirmed the *retain* will remove the `slot ` because `!is_being_flushed` is true.

runtime/src/accounts_db.rs
```diff
@@ -3794,101 +3857,132 @@ impl AccountsDb {
...
+                // If we're purging this slot, don't flush it here
+                if remove_unrooted_slots.contains(&slot) {
+                    true
+                } else {
+                    remove_unrooted_slots.insert(slot);
+                    false
+                }
+            };
+            if !is_being_purged {
+                let flush_stats = self.do_flush_slot_cache(slot, &slot_cache, should_flush_f);
```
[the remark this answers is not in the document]
carllin: so we held the lock until we inserted here: `remove_unrooted_slots.insert(slot);`, then we dropped the lock. 

The thread running the `remove_unrooted_slots()` function will block until this flush thread removes the slot from the `remove_unrooted_slots` list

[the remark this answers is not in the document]
lijunwangs: Got it. Maybe rename let mut remove_unrooted_slots to let mut slots_under_contention make it more clear.

runtime/src/accounts_db.rs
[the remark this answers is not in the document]
lijunwangs: Thanks explaining. Still have a question. In line 3445, contended_cache_flush_slots only has the slots which is in the also in contended_slots. Doesn't line 3471 re-add it again?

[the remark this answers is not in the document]
carllin: ha yeah it does, I don't think it's strictly necessary, but I wanted to uphold the invariant that any slot that is *either* undergoing flush or `remove_unrooted_slots` are both present in `contended_slots`, so we always have the working set of slots observable by any thread.

runtime/src/accounts_db.rs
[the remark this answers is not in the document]
carllin: yup, will do!

lijunwangs (approved):
Looks good to me. Thanks for answering my questions.

runtime/src/accounts_db.rs
brooksprumo: Ahhh understood. Thanks for the info!

runtime/src/accounts_db.rs
```diff
@@ -753,6 +753,13 @@ impl RecycleStores {
     }
 }
```
brooksprumo, line 755:
```suggestion

/// Removing unrooted slots in Accounts Background Service needs to be synchronized with flushing slots from the Accounts Cache.  This keeps track of those slots and the Mutex + Condvar for synchronization.
```

runtime/src/accounts_db.rs
```diff
@@ -831,6 +838,11 @@ pub struct AccountsDb {
     load_limit: AtomicU64,
 
     is_bank_drop_callback_enabled: AtomicBool,
+
+    // Set of slots currently being flushed by `flush_slot_cache()` or removed
+    // by `remove_unrooted_slot()`. Used to ensure `remove_unrooted_slots(slots)`
+    // can safely clear the set of unrooted slots `slots`.
```
brooksprumo, line 844: I love the comments. Could they be `///` instead?

brooksprumo (dismissed):
Looks good! I think the doc comment updates would help future-Brooks too!

runtime/src/accounts_db.rs
carllin: Updated, thanks!

runtime/src/accounts_db.rs
carllin: done!

jeffwashington (dismissed):
lgtm

brooksprumo (dismissed):
Looks good to me!

runtime/src/accounts_db.rs
```diff
@@ -3396,30 +3411,90 @@ impl AccountsDb {
...
+    pub fn remove_unrooted_slots(&self, remove_slots: &[Slot]) {
+        let rooted_slots = self
+            .accounts_index
+            .get_rooted_from_list(remove_slots.iter());
+        if !rooted_slots.is_empty() {
+            panic!(
+                "Trying to remove accounts for rooted slots {:?}",
+                rooted_slots
+            );
+        }
```
ryoqun, line 3433: nits:

```rust
assert!(rooted_slots.is_empty(), "Trying to remove accounts for rooted slots {:?}", rooted_slots);
```

runtime/src/accounts_db.rs
```diff
@@ -3396,30 +3411,90 @@ impl AccountsDb {
...
+
+                // For each slot the cache flush has finished, mark that we're about to start
+                // purging these slots by reserving it in `contended_slots`.
+                contended_cache_flush_slots.retain(|flushing_slot| {
+                    let is_being_flushed = contended_slots.contains(flushing_slot);
+                    if !is_being_flushed {
+                        // Mark that we're about to delete this slot now
+                        contended_slots.insert(*flushing_slot);
+                    }
+                    !is_being_flushed
```
ryoqun, line 3485: maybe, this condition needs to be flipped?

i mean, we need to `retain()` elements of `contended_cache_flush_slots` which have yet to be flushed (i.e. `is_being_flushed`; slots that are still `contain()`-ed  in `contended_slots` (and will be `remove()`-ed by the flusher thread later)?

runtime/src/accounts_db.rs
```diff
@@ -3396,30 +3411,90 @@ impl AccountsDb {
...
+            // we want to remove in this function
+            let mut contended_cache_flush_slots: Vec<Slot> = remove_slots
+                .iter()
+                .filter(|remove_slot| {
+                    let is_cache_flushing_slot = contended_slots.contains(remove_slot);
+                    if !is_cache_flushing_slot {
+                        // Reserve the slots that we want to purge that aren't currently
+                        // being flushed to prevent cache from flushing those slots in
+                        // the future
+                        contended_slots.insert(**remove_slot);
```
ryoqun, line 3454: this exclusive control-flow logic assumes same (detected as duplicate) `remove_slot` won't be ever given to `remove_unrooted_slots()` from multiple replayingstage threads. Is this correct assumption? Maybe replayingstage always process a slot at each time? How about commenting about the underlying assumption somewhere?

runtime/src/accounts_db.rs
```diff
@@ -3396,30 +3411,90 @@ impl AccountsDb {
...
+                    contended_slots = signal.wait(contended_slots).unwrap();
+                } else {
+                    // There are no slots being flushed to wait on, so it's safe to continue
+                    // to purging the slots we want to purge!
+                    break;
+                }
+
+                // For each slot the cache flush has finished, mark that we're about to start
+                // purging these slots by reserving it in `contended_slots`.
+                contended_cache_flush_slots.retain(|flushing_slot| {
```
ryoqun, line 3479: nit: how about this naming?:

```rust
remaining_contended_flush_slots.retain(|flush_slot|
```

- `remaining_` is added to better describe this variable is `retained()` iteratively until `.is_empty()`.
- `flush_slot` is used to be consistent with `remove_slot` for english grammar.

runtime/src/accounts_db.rs
```diff
@@ -3396,30 +3411,90 @@ impl AccountsDb {
...
+        {
+            // Slots that are currently being flushed by flush_slot_cache()
+            let mut contended_slots = slots_under_contention.lock().unwrap();
+
+            // Slots that are currently being flushed by flush_slot_cache() AND
+            // we want to remove in this function
+            let mut contended_cache_flush_slots: Vec<Slot> = remove_slots
+                .iter()
+                .filter(|remove_slot| {
+                    let is_cache_flushing_slot = contended_slots.contains(remove_slot);
```
ryoqun, line 3449: nits: rename to `is_being_flushed` to be consistent with https://github.com/solana-labs/solana/pull/17269/files#r641874795 (`let is_being_flushed = contended_slots.contains(flushing_slot);`)

imo, semantically same expression should result in same name unless there is good reason (like state transition between the occurences). In this case, I think its usage is same for both occurrences in that it's used to reserve slot with `contended_slots` for removal.

runtime/src/accounts_db.rs
```diff
@@ -3396,30 +3411,90 @@ impl AccountsDb {
...
+        }
+
+        let RemoveUnrootedSlotsSynchronization {
+            slots_under_contention,
+            signal,
+        } = &self.remove_unrooted_slots_synchronization;
+
+        {
+            // Slots that are currently being flushed by flush_slot_cache()
+            let mut contended_slots = slots_under_contention.lock().unwrap();
```
ryoqun, line 3442: nits: rename to `currently_contended_slots` to indicate volatility or mutation of this variable via `CondVar`?

runtime/src/accounts_db.rs
```diff
@@ -753,6 +753,16 @@ impl RecycleStores {
...
 }
 
+/// Removing unrooted slots in Accounts Background Service needs to be synchronized with flushing
+/// slots from the Accounts Cache.  This keeps track of those slots and the Mutex + Condvar for
+/// synchronization.
+#[derive(Debug, Default)]
+struct RemoveUnrootedSlotsSynchronization {
+    // slots being flushed from the cache or being purged
+    slots_under_contention: Mutex<HashSet<Slot>>,
+    signal: Condvar,
```
ryoqun, line 763: fyi, our first usage of `Condvar` in the codebase. :)

jeffwashington: I know. I keep running into situations where this has almost been the piece I needed. Glad to know it is there.

runtime/src/accounts_db.rs
carllin: Yeah, there is only one replay thread, and one version of the `remove_slot` needs to be removed before the next version can be replayed. Added a comment!

runtime/src/accounts_db.rs
ryoqun: perfect. :)

runtime/src/accounts_db.rs
```diff
@@ -3396,30 +3411,94 @@ impl AccountsDb {
...
+                    let is_being_flushed = currently_contended_slots.contains(remove_slot);
+                    if !is_being_flushed {
+                        // Reserve the slots that we want to purge that aren't currently
+                        // being flushed to prevent cache from flushing those slots in
+                        // the future.
+                        //
+                        // Note that the single replay thread has to remove a specific slot `N`
+                        // before another version of the same slot can be replayed. This means
+                        // multiple threads should not call `remove_unrooted_slots()` simultaneously
+                        // with the same slot.
```
ryoqun, line 3457: :100:

ryoqun (approved):
lgtm; nice solid work!

I like the logically chunked series of prs towards the duplicate slot implementation. :)

Issue #17268, opened 2021-05-17
Unable to build it in a docker container using rust image


r4881t:
#### Problem

I am trying to build this repo inside a docker container and it fails. It's very simple and similar to the one at [https://github.com/solana-labs/solana/blob/master/ci/docker-rust/Dockerfile](https://github.com/solana-labs/solana/blob/master/ci/docker-rust/Dockerfile). 

```

# Based on the work at https://github.com/solana-labs/solana/blob/master/ci/docker-rust/Dockerfile


# Note: when the rust version is changed also modify
# ci/rust-version.sh to pick up the new image tag
FROM rust:1.51.0

RUN set -x \
 && apt update \
 && apt-get install apt-transport-https \
 && echo deb https://apt.buildkite.com/buildkite-agent stable main > /etc/apt/sources.list.d/buildkite-agent.list \
 && apt-key adv --no-tty --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 32A37959C2FA5C3C99EFBC32A79206696452D198 \
 && apt update \
 && apt install -y \
      buildkite-agent \
      clang-7 \
      cmake \
      lcov \
      libudev-dev \
      libclang-common-7-dev \
      mscgen \
      net-tools \
      rsync \
      sudo \
      golang \
      unzip

WORKDIR /root

# Install Solana itself

RUN git clone https://github.com/solana-labs/solana.git solana

RUN cd /root/solana && cargo build


```

Upon running this, I get error as 

```
   ....
   ....
   ....
   Compiling solana-merkle-root-bench v1.7.0 (/root/solana/merkle-root-bench)
   Compiling solana-stake-accounts v1.7.0 (/root/solana/stake-accounts)
   Compiling solana-watchtower v1.7.0 (/root/solana/watchtower)
   Compiling solana-cargo-build-bpf v1.7.0 (/root/solana/sdk/cargo-build-bpf)
   Compiling rocksdb v0.15.0
   Compiling solana-genesis v1.7.0 (/root/solana/genesis)
   Compiling solana-poh-bench v1.7.0 (/root/solana/poh-bench)
   Compiling solana-ledger-tool v1.7.0 (/root/solana/ledger-tool)
error[E0658]: use of unstable library feature 'stdsimd'
  --> poh-bench/src/main.rs:87:12
   |
87 |         if is_x86_feature_detected!("avx2") && entry::api().is_some() {
   |            ^^^^^^^^^^^^^^^^^^^^^^^
   |
[...]

PR #17270, opened 2021-05-17
Fix the error for cross compile architecture

r4881t:
#### Problem

The problem is mentioned at https://github.com/solana-labs/solana/issues/17268

#### Summary of Changes

I have simply followed what the compiler was saying, and I was able to get it compiled

Fixes #

- Added `#[cfg(any(target_arch = "x86", target_arch = "x86_64"))] {` before the use of `is_x86_feature_detected!("avx512f")`

### DIFF
--- a/poh-bench/src/main.rs
+++ b/poh-bench/src/main.rs
@@ -117,7 +117,6 @@ fn main() {
                 );
             }
         }
-        
 
         if perf_libs::api().is_some() {
             let mut time = Measure::start("time");
--- a/poh-bench/src/main.rs
+++ b/poh-bench/src/main.rs
@@ -84,35 +84,40 @@ fn main() {
             time.as_us() / iterations as u64
         );
 
-        if is_x86_feature_detected!("avx2") && entry::api().is_some() {
-            let mut time = Measure::start("time");
-            for _ in 0..iterations {
-                assert!(ticks[..num_entries]
-                    .verify_cpu_x86_simd(&start_hash, 8)
-                    .finish_verify());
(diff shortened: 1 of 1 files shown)


r4881t:
I see the following error in testing, but I am not sure if they are due to my changes. i am not even a rust developer currently so can't really help with fixing them. 


```
$ cargo test --release

...
...
...

failures:

---- serialization::tests::test_serialize_parameters stdout ----
thread 'serialization::tests::test_serialize_parameters' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidArgument', programs/bpf_loader/src/serialization.rs:410:10
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- syscalls::tests::test_syscall_sol_alloc_free stdout ----
thread 'syscalls::tests::test_syscall_sol_alloc_free' panicked at 'assertion failed: `(left != right)`
  left: `0`,
 right: `0`', programs/bpf_loader/src/syscalls.rs:2763:17


failures:
    serialization::tests::test_serialize_parameters
    syscalls::tests::test_syscall_sol_alloc_free

test result: FAILED. 30 passed; 2 failed; 1 ignored; 0 measured; 0 filtered out; finished in 0.22s

error: test failed, to rerun pass '-p solana-bpf-loader-program --lib'

```

Lichtso:
The thing is that compiling the JIT compiler should bail if the target arch is not `x86_64` and it currently does not check that. (I will patch that in the next release of RBPF).

Anyway, can you check what the output of `rustc --print cfg` is? Because it seems like you are not building for a `x86_64` target.

PR #17276, opened 2021-05-17
Refactor genesis download/load/check functions

sakridge:
#### Problem

Genesis download code could be useful for another client that wants to download genesis, load it and check the hash.

#### Summary of Changes

Refactor into genesis-utils crate from validator/src/main.rs

Fixes #


ryoqun (approved):
LGTM! thanks for cleaning up!

PR #17278, opened 2021-05-17
fixed getProgramAccounts fields list

dragmz:
#### Problem

Response fields list is missing a bullet for the data field

#### Summary of Changes

Added the missing bullet for the data field


mvines (approved):
Thanks

PR #17289, opened 2021-05-17
Explorer: add token market prices from Coingecko

oJshua:
#### Problem
Tokens don't currently list a market price.

#### Summary of Changes
Now that token-list has CoingeckoIds, we can query for price data.

Explorer-Solana (4)


bartosz-lipinski (approved):
Looks awesome 🚀

PR #17292, opened 2021-05-17
simplify generate_index

jeffwashington:
#### Problem
Generate_index has some outdated comments and data structures. We don't need old write version entries during generate_index.
#### Summary of Changes
Simplify, remove old comments.
Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -53,7 +53,7 @@ use solana_vote_program::vote_state::MAX_LOCKOUT_HISTORY;
 use std::{
     borrow::{Borrow, Cow},
     boxed::Box,
-    collections::{hash_map::Entry, BTreeMap, BTreeSet, HashMap, HashSet},
+    collections::{hash_map::Entry, BTreeSet, HashMap, HashSet},
     convert::TryFrom,
     io::{Error as IoError, Result as IoResult},
     ops::{Range, RangeBounds},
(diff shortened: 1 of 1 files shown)


runtime/src/accounts_db.rs
```diff
@@ -4953,10 +4953,8 @@ impl AccountsDb {
     }
 
     pub fn generate_index(&self) {
-        // BTreeMap because we want in-order traversal of oldest write_version to newest.
-        // Thus, all instances of an account in a store are added to the index in oldest to newest
-        // order and we update refcounts and track reclaims correctly.
-        type AccountsMap<'a> = HashMap<Pubkey, BTreeMap<u64, (AppendVecId, StoredAccountMeta<'a>)>>;
+        type AccountsMap<'a> =
+            HashMap<Pubkey, (StoredMetaWriteVersion, AppendVecId, StoredAccountMeta<'a>)>;
```
carllin, line 4957: I think the old versioned accounts still need to be inserted into the index to properly update the refcount, https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_index.rs#L204. This is necessary as some of the older entries in old storage entries may be preventing zero lamport keys from being cleaned.

carllin: Discussed with @jeffwashington offline, this should be ok as the refcount is only incremented once per pubkey per slot anyways

carllin (approved):

PR #17293, opened 2021-05-17
StoredMetaWriteVersion

jeffwashington:
#### Problem
Make a named type for StoredMetaWriteVersion instead of using u64. This helps code readability for what a u64 could mean.
#### Summary of Changes

Fixes #


carllin (approved):

jeffwashington:
@brooksprumo fyi for the Lamport type(?) you were going to make. This is lower budget, but has some readability improvements.

brooksprumo:
> @brooksprumo fyi for the Lamport type(?) you were going to make. This is lower budget, but has some readability improvements.

I definitely agree! I like this for its readability improvements.

PR #17294, opened 2021-05-17
Add C Serialization Tests for #17217

jackcmay:
#### Problem

#17217 fix went in but didn't contain a verifying test

#### Summary of Changes

Add test

Fixes #

### DIFF
--- a/programs/bpf/c/src/ser/ser.c
+++ b/programs/bpf/c/src/ser/ser.c
@@ -1,0 +1,34 @@
+/**
+ * @brief Example C-based BPF sanity rogram that prints out the parameters
+ * passed to it
+ */
+#include <solana_sdk.h>
+
+extern uint64_t entrypoint(const uint8_t *input) {
+  SolAccountInfo ka[1];
+  SolParameters params = (SolParameters){.ka = ka};
+
+  sol_log(__FILE__);
+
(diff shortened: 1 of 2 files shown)


programs/bpf/c/src/ser/ser.c
```diff
@@ -0,0 +1,34 @@
...
+    return ERROR_INVALID_ARGUMENT;
+  }
+
+  char ka_data[] = {1, 2, 3};
+  SolPubkey ka_owner;
+  sol_assert(params.ka_num == 2);
+  for (int i = 0; i < 2; i++) {
+    sol_assert(*params.ka[i].lamports == 42);
+    sol_assert(!sol_memcmp(params.ka[i].data, ka_data, 4));
+    sol_assert(SolPubkey_same(params.ka[i].owner, &ka_owner));
```
dmakarov, line 23: This test seems to fail consistently at this line.  Why should it work if local struct `ka_owner` is not initialized?  What am I missing?..

jackcmay:
Test is probably assuming zero initialized memory

dmakarov:
but the `params.ka[i].owner` is not zero.

jackcmay:
Fixed in: https://github.com/solana-labs/solana/pull/17310

PR #17302, opened 2021-05-18
Chore: remove dead code

hrls:
#### Problem
Error type case variant is unnecessary. There is no way to construct argument of Infallible type.


hrls:
Nice

t-nelson (approved):
Sorry!  I mixed this one up with a similarly titled "remove unused code" PR that was merged around the same time.

Much cleaner.  Thanks for the contribution

PR #17305, opened 2021-05-18
tests: cleanup test

jeffwashington:
#### Problem
Use local in test to make code more straightforward. Test is about to change elsewhere and this change is noisy.

#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/accounts_index.rs
+++ b/runtime/src/accounts_index.rs
@@ -2203,6 +2203,7 @@ pub mod tests {
     #[test]
     fn test_insert_new_with_lock_no_ancestors() {
         let key = Keypair::new();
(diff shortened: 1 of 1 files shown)


AshwinSekar (dismissed):
https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_index.rs#L2214 looks like we can change this one as well

jeffwashington:
> https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_index.rs#L2214 looks like we can change this one as well

yep, 1 more, too. These are the lines that are changing in the other pr that this pr facilitates, so they didn't make the transition.

AshwinSekar (approved):

PR #17308, opened 2021-05-18
shink all slots in parallel on startup

jeffwashington:
#### Problem
validator startup time is painful - especially as # accounts grows. There will soon be an issue filed for this.
#### Summary of Changes
From verify_snapshot_bank, shrink slots in parallel.
Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -2170,12 +2170,22 @@ impl AccountsDb {
         num_candidates
     }
 
-    pub fn shrink_all_slots(&self) {
-        for slot in self.all_slots_in_storage() {
-            if self.caching_enabled {
-                self.shrink_slot_forced(slot);
-            } else {
-                self.do_shrink_slot_forced_v1(slot);
+    pub fn shrink_all_slots(&self, is_startup: bool) {
+        if is_startup && self.caching_enabled {
+            let slots = self.all_slots_in_storage();
+            let chunk_size = std::cmp::max(slots.len() / 8, 1); // approximately 400k slots in a snapshot
+            slots.par_chunks(chunk_size).for_each(|slots| {
+                for slot in slots {
+                    self.shrink_slot_forced(*slot);
+                }
+            });
+        } else {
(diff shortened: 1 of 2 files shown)


jeffwashington:
```
verify_snapshot_bank clean_us=57710861i shrink_all_slots_us=1097525i verify_bank_hash_us=19499188i verify_hash_us=2205i (master)
verify_snapshot_bank clean_us=57490783i shrink_all_slots_us= 622640i verify_bank_hash_us=19254628i verify_hash_us=2209i (this pr)
```

runtime/src/bank.rs
```diff
@@ -4815,8 +4815,8 @@ impl Bank {
         self.rc.accounts.accounts_db.clean_accounts(max_clean_slot);
     }
 
-    pub fn shrink_all_slots(&self) {
-        self.rc.accounts.accounts_db.shrink_all_slots();
+    pub fn shrink_all_slots(&self, startup: bool) {
```
carllin, line 4818: nit `startup -> is_startup`

runtime/src/accounts_db.rs
```diff
@@ -2170,12 +2170,22 @@ impl AccountsDb {
...
     }
 
-    pub fn shrink_all_slots(&self) {
-        for slot in self.all_slots_in_storage() {
-            if self.caching_enabled {
-                self.shrink_slot_forced(slot);
-            } else {
-                self.do_shrink_slot_forced_v1(slot);
+    pub fn shrink_all_slots(&self, startup: bool) {
+        if startup && self.caching_enabled {
```
carllin, line 2174: any particular reason we can't do the parallel shrink when caching is not enabled?

jeffwashington: I don't want to test that code path or be worried about it ;-)

jeffwashington: We are actively maintaining and improving the caching code path. I have more confidence we will get that codepath right and keep it right.

carllin (approved):

PR #17316, opened 2021-05-18
move Ancestors to its own module

jeffwashington:
#### Problem
Ancestors will soon change and become worthy of its own module.
#### Summary of Changes
Give it some space to grow.
Fixes #


jeffwashington:
Fyi, here is what happens when you move something to its own 'module'. The 'Ancestors' type used to be HashMap<...> until Ryo turned it into a type Ancestors = HashMap<...> a year ago. Today, [I'm ready to turn it into a real struct, impl](https://github.com/solana-labs/solana/pull/16477), etc. Stephen asked it to go in its own module.

lijunwangs (approved):
Looks good to me.

AshwinSekar (approved):

Issue #16638, opened 2021-04-19
Racy test test_load_account_and_purge_race_without_retry


sakridge:
#### Problem

`test_load_account_and_purge_race_without_retry` is racy
https://buildkite.com/solana-labs/solana/builds/44388#9e5bfd29-91b7-4304-969c-2a8aff9e5b72/125-1941

#### Proposed Solution

debug/fix

sakridge:
cc @ryoqun

mvines:
Please revert https://github.com/solana-labs/solana/pull/16793 when this issue is fixed

ryoqun:
this is indirectly fixed by https://github.com/solana-labs/solana/pull/17319

PR #17319, opened 2021-05-19
Refactor purge_slots_from_cache_and_store() and handle_reclaims()

carllin:
#### Problem
To support the dumping logic in https://github.com/solana-labs/solana/pull/17269, `remove_unrooted_slot()` https://github.com/solana-labs/solana/blob/a3c0833a1cb52236340fd04984fea3e08d8aa9ba/runtime/src/accounts_db.rs#L3230 needs to for an unrooted slot `S`:

1) Remove all the accounts index entries for `S`
2) Remove all the storage entries `S`

To this end, it would be nice if `remove_unrooted_slot()` and other purge logic like `purge_slots()`: https://github.com/solana-labs/solana/blob/a3c0833a1cb52236340fd04984fea3e08d8aa9ba/runtime/src/accounts_db.rs#L3209-L3221 (called on `Bank::drop()` for unrooted banks), could share the same core purging logic by. both calling into `purge_slots_from_cache_and_store()`: https://github.com/solana-labs/solana/blob/a3c0833a1cb52236340fd04984fea3e08d8aa9ba/runtime/src/accounts_db.rs#L3068-L3072 to handle all the associated purging logic.

Currently however `purge_slots_from_cache_and_store()`, if the slot to be purged is not in the cache, it does not purge the AccountsIndex entries, only the storage entries themselves: https://github.com/solana-labs/solana/blob/a3c0833a1cb52236340fd04984fea3e08d8aa9ba/runtime/src/accounts_db.rs#L3098-L3101

#### Summary of Changes
1. In order to introduce accounts index purging into `purge_slots_from_cache_and_store(slot)` (see Problem 1 above), we would ideally like to reuse the `purge_exact() -> handle_reclaims()` (similar to in clean_accounts) flow to completely remove the `slot`. <br/><br/>However, the current problem is `handle_reclaims()` itself calls into `purge_slots_from_cache_and_store()` to delete the storage entries, via the path `handle_reclaims() -> process_dead_slots() -> purge_storage_slots() -> purge_slots_from_cache_and_store()`, so we can't currently call `handle_reclaims()` inside `purge_slots_from_cache_and_store()`.<br/><br/>To fix this first we first note that `handle_reclaims()` only ever reclaims entries not in the cache. This means if we factor out the storage deletion logic from `purge_slots_from_cache_and_store()` into a separate `purge_slot_storage_entries()`, then we can break. this cycle  by having both `purge_slots_from_cache_and_store()` and `handle_reclaims() -> purge_storage_slots()` call into `purge_slot_storage_entries()`: https://github.com/solana-labs/solana/compare/master...carllin:RefactorPurge?expand=1#diff-1090394420d51617f3233275c2b65ed706b35b53b115fe65f82c682af8134a6fR3112

2.  Now that `purge_slots_from_cache_and_store()` completely removes the accounts index entry for unrooted banks, we no longer need the logic in `Bank::drop()` to add the dirty keys: https://github.com/solana-labs/solana/compare/master...carllin:RefactorPurge?expand=1#diff-ed47b4a0198313377e091bb3957bbbc63d937805426d1b2b6de39d0a50d32a0cL5120-L5130
@brooksprumo 

Fixes #


runtime/src/bank.rs
```diff
@@ -12724,7 +12698,7 @@ pub(crate) mod tests {
                 .accounts_db
                 .accounts_index
                 .ref_count_from_storage(&key5.pubkey()),
-            expected_ref_count_for_keys_only_in_slot_2
+            expected_ref_count_for_keys_in_both_slot1_and_slot2,
```
carllin, line 12701: @brooksprumo these tests were super helpful for debugging 😄 

For this test here there was a peculiar case. I think the correct ref count here should be `1`, since `pubkey5` was added in both `bank1` and `bank1` and then we dropped/purged `bank1`. The old test expected `expected_ref_count_for_keys_only_in_slot_2 == 2` references, which seems wrong in retrospect. With this change, it seems to be correctly reporting a refcount of 1.

runtime/src/accounts_db.rs
```diff
@@ -10431,76 +10431,6 @@ pub mod tests {
...
-                    // Simulate purge_slots()
-                    db.purge_slot(slot);
-                    sleep(Duration::from_millis(RACY_SLEEP_MS));
-                })
-                .unwrap()
-        };
-
-        let ancestors: Ancestors = vec![(slot, 0)].into_iter().collect();
-        let t_do_load =
-            start_load_thread(with_retry, ancestors, db, exit.clone(), pubkey, move |_| {
```
carllin, line 10382: @ryoqun from what I remember, these tests were essentially testing that if we try to load on a purged note, we panic.

This was because the behavior before was we only removed the storage entry in `purge_slot()` and left the accounts index entry untouched. This meant we would fail to retrieve the account, retry the get from the index, see the same index entry again and panic when we got the same index entry: https://github.com/solana-labs/solana/blob/305d9dd3f4679c8026bdd20b18ec9d84e3b770f0/runtime/src/accounts_db.rs#L2674-L2703

With the current set of changes where the `purge_slot()` call above actually removes the account index entry, which means the `load thread` here will realize the account is deleted and will not retry the account index get, returning `None`. This causes the load thread to panic here: https://github.com/solana-labs/solana/blob/305d9dd3f4679c8026bdd20b18ec9d84e3b770f0/runtime/src/accounts_db.rs#L10184 since the pubkey no longer exists in the index.

Essentially it is no longer possible for the panic to happen because the account index entry is removed before the storage entry is removed, so I have removed these tests.

ryoqun: @carllin your reasoning sounds good at first glance. let me think on it more thoughly though. btw, another pretty mind-bending pr from you. :)

runtime/src/accounts_db.rs
```diff
@@ -10339,76 +10431,6 @@ pub mod tests {
...
-    #[test]
-    #[should_panic(expected = "assertion failed: load_hint == LoadHint::Unspecified")]
-    fn test_load_account_and_purge_race_with_retry() {
-        // this tests impossible situation in the wild, so panic is expected
-        // Conversely, we show that we're preventing this race condition from occurring
-        do_test_load_account_and_purge_race(true);
-    }
-
-    #[test]
-    #[ignore]
```
ryoqun, line 10402: yay! less work (ref #16638) :)

runtime/src/bank.rs
brooksprumo: Glad they helped!

Yeah, I was confused about the ref count of 2 as well. I took a step back in this particular test case to figure out what I _really_ wanted to test out and avoid testing other components (if possible). That's why I compared key5 here to key2, because I knew that if key2 had the right value, then key5 should have that same value, whatever it actually is.

I think there is/was another issue that caused the ref count to be 2 instead of 1, which looks like you've found and fixed (yay!). To make sure I understand right, shouldn't the original logic still work in the assert too? Specifically:

```
let expected_ref_count_for_keys_in_both_slot1_and_slot2 = bank2
    .rc
    .accounts
    .accounts_db
    .accounts_index
    .ref_count_from_storage(&key2.pubkey());
```

runtime/src/accounts_db.rs
```diff
@@ -3065,56 +3107,148 @@ impl AccountsDb {
         recycle_stores_write_elapsed.as_us()
     }
 
+    // Purges every slot in `removed_slots` from both the cache and storage. This includes
+    // entries in the accounts index, cache entries, and any backing storage entries.
```
brooksprumo, line 3111: nit: Thanks for adding comments! I can't overstate how helpful they are. Can you make these doc (`///`) comments? Then they get rendered with the docs and in some editors too.

runtime/src/accounts_db.rs
```diff
@@ -3065,56 +3107,148 @@ impl AccountsDb {
...
+        purge_stats
+            .num_cached_slots_removed
+            .fetch_add(num_cached_slots_removed, Ordering::Relaxed);
+        purge_stats
+            .total_removed_cached_bytes
+            .fetch_add(total_removed_cached_bytes, Ordering::Relaxed);
+    }
+
+    // Purge the backing storage entries for the given slot, does not purge from
+    // the cache!
```
brooksprumo, line 3215: nit: Same `///` comment here

brooksprumo (commented):
There's a lot of changes in there that I don't fully grok yet, so I'll defer to others on approval. I've added tiny nits, mostly comments because I think they are helpful. I'll revisit this PR and add additional comments if I find anything else.

runtime/src/bank.rs
carllin: @brooksprumo, you'll be surprised to find that the original logic you pointed out returns `2` instead of `1` for `expected_ref_count_for_keys_in_both_slot1_and_slot2`. This is because `key2` is touched in both `bank0` and `bank1`.

Now you may that that because the only *explicit* transaction we did in `bank0` was a transfer to `key2`, then when that update is outdated by a transfer on a later *rooted* bank `bank2.transfer(4, &mint_keypair, &key2.pubkey()).unwrap();`, then `clean_accounts()` should mark slot `0` as dead and unref `key2`. However, `bank0` actually has a lot of additional keys initialized from genesis that have not been outdated yet, namely sysvars and various program accounts, here's a list!

```
updating slot 0, pubkey 3ywG1DkxsBYvS6oq4cmeCyzPJzDYeX2eLhEwyJjJaocj
updating slot 0, pubkey SysvarFees111111111111111111111111111111111
updating slot 0, pubkey 11111111111111111111111111111111
updating slot 0, pubkey Vote111111111111111111111111111111111111111
updating slot 0, pubkey Stake11111111111111111111111111111111111111
updating slot 0, pubkey Config1111111111111111111111111111111111111
updating slot 0, pubkey So11111111111111111111111111111111111111112
updating slot 0, pubkey SysvarStakeHistory1111111111111111111111111
updating slot 0, pubkey SysvarC1ock11111111111111111111111111111111
updating slot 0, pubkey SysvarRent111111111111111111111111111111111
updating slot 0, pubkey SysvarEpochSchedu1e111111111111111111111111
updating slot 0, pubkey SysvarRecentB1ockHashes11111111111111111111
updating slot 0, pubkey 3ywG1DkxsBYvS6oq4cmeCyzPJzDYeX2eLhEwyJjJaocj
updating slot 0, pubkey DMfWhsGGtwa4otGUrAx1fNMRbSGZ3pXdxtuw67k2G6wr
```

Updated in later slots:

```
updating slot 1, pubkey SysvarS1otHashes111111111111111111111111111
updating slot 1, pubkey SysvarC1ock11111111111111111111111111111111
updating slot 1, pubkey SysvarFees111111111111111111111111111111111
updating slot 1, pubkey 3ywG1DkxsBYvS6oq4cmeCyzPJzDYeX2eLhEwyJjJaocj
updating slot 1, pubkey 9dG5h4dsfz3vCy9C8c4Fhod4F96YtJcHe9okwht8ouit
updating slot 1, pubkey 36nj3aidYASj5kndMV2GsjcGB8egiEzGLiBAZFXcGhs8
updating slot 1, pubkey Hu8UuzSyQStcsfMFsxNepzG7PN2yKDvH8Qez32WDR5D8
updating slot 2, pubkey SysvarS1otHashes111111111111111111111111111
updating slot 2, pubkey SysvarC1ock11111111111111111111111111111111
updating slot 2, pubkey SysvarFees111111111111111111111111111111111
updating slot 2, pubkey 3ywG1DkxsBYvS6oq4cmeCyzPJzDYeX2eLhEwyJjJaocj
updating slot 2, pubkey DMfWhsGGtwa4otGUrAx1fNMRbSGZ3pXdxtuw67k2G6wr
updating slot 2, pubkey 3ywG1DkxsBYvS6oq4cmeCyzPJzDYeX2eLhEwyJjJaocj
updating slot 2, pubkey 44y4dbRFM2Pijb6nRFJjbTaU96ut62tbGhiFnFjGY5Zs
updating slot 2, pubkey Hu8UuzSyQStcsfMFsxNepzG7PN2yKDvH8Qez32WDR5D8
updating slot 2, pubkey SysvarS1otHistory11111111111111111111111111
```

runtime/src/accounts_db.rs
```diff
@@ -3065,56 +3107,148 @@ impl AccountsDb {
...
+                    Some(&mut ReclaimResult::default()),
+                    false,
+                );
+                handle_reclaims_elapsed.stop();
+                purge_stats
+                    .handle_reclaims_elapsed
+                    .fetch_add(handle_reclaims_elapsed.as_us(), Ordering::Relaxed);
+                // After handling the reclaimed entries, this slot's
+                // storage entries should be purged from self.storage
+                assert!(self.storage.get_slot_stores(*remove_slot).is_none());
```
carllin, line 3195: ick, this can actually race with `clean_accounts()`. If clean removes some of the the accounts index entries first, then we wont remove them here, and the `handle_reclaims()` call here may not remove all the storage entries (clean may have the remaining reclaims from the index necessary to mark the slot as dead)

ryoqun:
so, new callgraph looks like this..

image

runtime/src/accounts_db.rs
```diff
@@ -3176,24 +3303,6 @@ impl AccountsDb {
             .fetch_add(recycle_stores_write_elapsed, Ordering::Relaxed);
     }
 
-    fn purge_storage_slots(&self, removed_slots: &HashSet<Slot>) {
```
ryoqun, line 3179: memo: `purge_slot_storage_entries` is the new equivalent fn called in `process_dead_slots`.

runtime/src/accounts_db.rs
```diff
@@ -3089,56 +3131,148 @@ impl AccountsDb {
...
+            .num_cached_slots_removed
+            .fetch_add(num_cached_slots_removed, Ordering::Relaxed);
+        purge_stats
+            .total_removed_cached_bytes
+            .fetch_add(total_removed_cached_bytes, Ordering::Relaxed);
+    }
+
+    /// Purge the backing storage entries for the given slot, does not purge from
+    /// the cache!
+    fn purge_slot_storage_entries<'a>(
```
ryoqun, line 3240: here

runtime/src/accounts_db.rs
```diff
@@ -1890,7 +1932,7 @@ impl AccountsDb {
         clean_dead_slots.stop();
 
         let mut purge_removed_slots = Measure::start("reclaims::purge_removed_slots");
-        self.purge_storage_slots(&dead_slots);
+        self.purge_slot_storage_entries(dead_slots.iter(), purge_stats);
```
ryoqun, line 1935: call https://github.com/solana-labs/solana/pull/17319/files#r636980951

runtime/src/accounts_db.rs
```diff
@@ -3176,24 +3303,6 @@ impl AccountsDb {
             .fetch_add(recycle_stores_write_elapsed, Ordering::Relaxed);
     }
```
ryoqun, line 3304: memo: end of `purge_slots_from_cache_and_store` in old code.

runtime/src/accounts_db.rs
```diff
@@ -3089,56 +3131,148 @@ impl AccountsDb {
...
+                    &reclaims,
+                    expected_dead_slot,
+                    Some(purge_stats),
+                    Some(&mut ReclaimResult::default()),
+                    false,
+                );
+                handle_reclaims_elapsed.stop();
+                purge_stats
+                    .handle_reclaims_elapsed
+                    .fetch_add(handle_reclaims_elapsed.as_us(), Ordering::Relaxed);
```
ryoqun, line 3216: how about extracting this code into new fn like `purge_slot_storage()`. This is to align with `purge_slot_cache()` above.

runtime/src/accounts_db.rs
ryoqun: > Luckily right now AccountsBackgroundService runs all Bank::drop() serially with clean_accounts()

With the upcoming duplicate slot thing, how about introducing a runtime check to really validate this?

I think we can signal AccountsDb here at the identical timing to setup callbacks for all banks: https://github.com/solana-labs/solana/blob/44831c18d29f5c43feb0ffdb0abe258be34d4a99/core/src/tvu.rs#L237-L240

so that all subsequent `purge_slot()` must be coming from the ABS thread, not directly from `Bank::drop()`-ed thread.

ryoqun: i mean, I think this property could tend to be broken in the future by other code juggling. :)

runtime/src/accounts_db.rs
```diff
@@ -3089,56 +3131,148 @@ impl AccountsDb {
...
-                if !can_exist_in_cache {
-                    panic!("The removed slot must alrady have been flushed from the cache");
-                }
                 num_cached_slots_removed += 1;
                 total_removed_cached_bytes += slot_cache.total_bytes();
                 self.purge_slot_cache(*remove_slot, slot_cache);
-            } else if let Some((_, slot_removed_storages)) = self.storage.0.remove(&remove_slot) {
+                remove_cache_elapsed.stop();
+                remove_cache_elapsed_across_slots += remove_cache_elapsed.as_us();
+            } else {
```
ryoqun, line 3156: maybe time to update a part of my masterpiece? lol: 

https://github.com/solana-labs/solana/blob/44831c18d29f5c43feb0ffdb0abe258be34d4a99/runtime/src/accounts_db.rs#L2563-L2581

runtime/src/accounts_db.rs
ryoqun: this your comment is finally making sense... what a brain-teasing pr... lol

ryoqun:
> Problem
> 
> To support the dumping logic in #17269, remove_unrooted_slot() 

btw, this pr description is finally making sense too. and indeed, it turns out very crisp and consumable explanation. your efforts are well deserved for writing words, not code for human. :)

runtime/src/accounts_index.rs
```diff
@@ -1377,6 +1377,19 @@ impl<T: 'static + Clone + IsCached + ZeroLamport> AccountsIndex<T> {
         slot < max_clean_root && slot != newest_root_in_slot_list
     }
```
brooksprumo, line 1379:
```suggestion

/// Given a list of slots, return a new list of only the slots that are rooted
```

runtime/src/accounts_db.rs
brooksprumo, line 3240: 1. What about calling this function `purge_dead_slots_from_storage()`, since it is called from `process_dead_slots()`?

2. I know that I'd personally find it helpful to note in the doc comment that slots in `removed_slots` must be non-rooted.

runtime/src/accounts_db.rs
```diff
@@ -4523,7 +4601,27 @@ impl AccountsDb {
         dead_slots
     }
 
-    fn finalize_dead_slot_removal<'a>(
+    fn remove_dead_slot_metadata<'a>(
```
brooksprumo, line 4604: nit: Should `slot` be `slots`?

brooksprumo (commented):
OK, this is making a lot more sense now; thanks for explaining it in our call! Again, my comments are just around naming and documenting. Everything else is looks good to me!

I also like @ryoqun's suggestion of extracting out [the code](https://github.com/solana-labs/solana/pull/17319/files?file-filters%5B%5D=&w=1#r636998213) into a new function like `purge_slot_storage()`.

runtime/src/accounts_db.rs
carllin: @brooksprumo, actually, when called from the `clean_accounts-> handle_reclaims()` path, items in `removed_slots()` can be rooted if the rooted slot is dead, i.e. all of its updates have been outdated by later roots

runtime/src/accounts_db.rs
carllin: updated, have to keep this beauty fresh!

runtime/src/accounts_db.rs
carllin: @ryoqun, hmm so currently `Bank::new_from_parent()`, already guarantees that all child banks will have the proper callback.

>I think we can signal AccountsDb here at the identical timing to setup callbacks

So something to verify from `purge_slot()` that the caller is ABS thread, i.e. something like checking the thread id is equal to the ABS thread?

runtime/src/accounts_db.rs
brooksprumo: Oh, interesting. I think I misunderstood this part of the code:

https://github.com/solana-labs/solana/pull/17319/files/a3508c0af00c209c556b7ecf7b5120eff105bc6c#diff-1090394420d51617f3233275c2b65ed706b35b53b115fe65f82c682af8134a6fR3245-R3250

Maybe the comment can be updated?

carllin: @brooksprumo I think I see where the confusion is :). The slot being deleted is rooted, but by the time it reaches here, the root has already been deleted from the rooted set, so that assert won't trigger. The equivalent in the old code path:

`clean_stored_dead_slots() -> finalize_dead_slot_removal() -> clean_dead_slot()` removes the roots
https://github.com/solana-labs/solana/blob/d41266e4e915cc70fdec4a3ef54d5f1289f1afaa/runtime/src/accounts_db.rs#L1889 

and then we call the equivalent `purge_storage_slots`: https://github.com/solana-labs/solana/blob/d41266e4e915cc70fdec4a3ef54d5f1289f1afaa/runtime/src/accounts_db.rs#L1893 and make the assertion check here: https://github.com/solana-labs/solana/blob/d41266e4e915cc70fdec4a3ef54d5f1289f1afaa/runtime/src/accounts_db.rs#L3198-L3200

carllin: Updated comment to reflect this

brooksprumo: Ah! I get it now. Thanks for the explanation and the comments.

https://github.com/solana-labs/solana/commit/e602ebf159f0de2e72e7338ac36b4d4cf2117f03#diff-1090394420d51617f3233275c2b65ed706b35b53b115fe65f82c682af8134a6fR3190

Looks like there's a minor typo on 'alrady' in the link above.

carllin: thanks, fixed up!

runtime/src/accounts_db.rs
ryoqun: > So something to verify from purge_slot() that the caller is ABS thread, i.e. something like checking the thread id is equal to the ABS thread?

@carllin Oh, I was imaging more simpler way than thread id like this:

```rust
let callback = accounts_db.create_drop_callback(); // internally set `.drop_callback_installed` to `true`.
// Before replay starts, set the callbacks in each of the banks in BankForks 
for bank in bank_forks.read().unwrap().banks().values() { 
    bank.set_callback(callback)
}


impl AccountsDb {
   fn purge_slot(from_abs: bool) {
     if self.drop_callback_installed && !from_abs {
        panic!("bad drop callpath detected; bank.drop must be serialized")
     }
   }
}

impl Drop for Bank {
  fn drop() {
     accounts_db.purge_slot(..., false)
  }
}

impl ABS {
    accounts_db.purge_slot(..., true);
}
```

runtime/src/accounts_db.rs
```diff
@@ -4946,9 +5043,11 @@ impl AccountsDb {
         //    a) this slot  has at least one account (the one being stored),
         //    b)From 1) we know no other slots are included in the "reclaims"
         //
-        // From 1) and 2) we guarantee passing Some(slot), true is safe
+        // From 1) and 2) we guarantee passing `purge_stats` == None, which is
+        // equivalent to asserting there will be no dead slots, is safe.
+        let purge_stats = None;
```
ryoqun, line 5048: nits: `no_purge_stats`.

runtime/src/accounts_db.rs
```diff
@@ -1890,7 +1932,7 @@ impl AccountsDb {
         clean_dead_slots.stop();
 
         let mut purge_removed_slots = Measure::start("reclaims::purge_removed_slots");
-        self.purge_storage_slots(&dead_slots);
```
ryoqun, line 1893: memo: so, this is the pivotal code change to cut the circular problem.

runtime/src/accounts_db.rs
ryoqun: anyway this is optional nicety.

ryoqun:
so, this is the updated call tree:

image

ryoqun (dismissed):
LGTM with nits!

runtime/src/bank.rs
```diff
@@ -5145,16 +5127,14 @@ impl Drop for Bank {
...
-
         if let Some(drop_callback) = self.drop_callback.read().unwrap().0.as_ref() {
             drop_callback.callback(self);
         } else {
             // Default case
             // 1. Tests
             // 2. At startup when replaying blockstore and there's no
             // AccountsBackgroundService to perform cleanups yet.
-            self.rc.accounts.purge_slot(self.slot());
+            self.rc.accounts.purge_slot(self.slot(), false);
```
ryoqun, line 5137: :+1:

ryoqun (approved):
lgtm again!

PR #17320, opened 2021-05-19
Create solana-rpc crate and move subscriptions

CriesofCarrots:
#### Problem
There are several different aspects of validator operation that are enmeshed in solana-core, even though their functions -- rpc-services, gossip, and consensus -- are essentially distinct. This prevents us handling these functionalities independently, especially problematic in the case of rpc, where we might like to run the service separately from a voting validator or swap out the service altogether.

#### Summary of Changes
Create a solana-rpc crate and move rpc-subscriptions to the new crate
This is the first step toward ultimately splitting both rpc and gossip out of core.

There is no functionality change in this PR, with the exception of removing unnecessary core dependencies from the `test_vote_subscribe` unit test.
The last commit is purely cosmetic.


mvines (approved):
noice

CriesofCarrots:
Wasn't planning to backport these org changes to v1.6, but if other backports get too painful in the future, I can.

mvines:
> Wasn't planning to backport these org changes to v1.6, but if other backports get too painful in the future, I can.

I was assuming no backport to v1.6 too

PR #17332, opened 2021-05-19
removes manual trait impl for contact-info

behzadnouri:
#### Problem
The current implementations only use the `id` and disregard other fields,
in particular `wallclock`. This can lead to bugs where an outdated
contact-info shadows or overrides a current one because they compare
equal.

#### Summary of Changes
Removed manual implementation in favor of derive macro.


CriesofCarrots (approved):
New to this area, but makes sense to me.

PR #17340, opened 2021-05-20
indexes crds votes by insert order

behzadnouri:
#### Problem
Crds::get_votes is scanning over all votes in the crds table only to
return those inserted since the given cursor:
https://github.com/solana-labs/solana/blob/2ae57c172/core/src/crds.rs#L250-L266

Having votes indexed by insert order avoids the table scan and will be
more efficient.

#### Summary of Changes
* Index crds votes by insert order.

### DIFF
--- a/core/src/crds.rs
+++ b/core/src/crds.rs
@@ -48,7 +48,8 @@ pub struct Crds {
     cursor: Cursor, // Next insert ordinal location.
     shards: CrdsShards,
     nodes: IndexSet<usize>, // Indices of nodes' ContactInfo.
-    votes: IndexSet<usize>, // Indices of Vote crds values.
+    // Indices of Votes keyed by insert order.
+    votes: BTreeMap<u64 /*insert order*/, usize /*index*/>,
     // Indices of EpochSlots keyed by insert order.
     epoch_slots: BTreeMap<u64 /*insert order*/, usize /*index*/>,
     // Indices of all crds values associated with a node.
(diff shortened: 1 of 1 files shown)


behzadnouri:
`cluster_info_stats::get_votes`: left chunk is this code, vs current code on the right:

cluster_info_stats get_votes_2

core/src/crds.rs
```diff
@@ -172,7 +173,7 @@ impl Crds {
                         self.nodes.insert(entry_index);
                     }
                     CrdsData::Vote(_, _) => {
-                        self.votes.insert(entry_index);
+                        self.votes.insert(value.ordinal, entry_index);
```
carllin, line 176: For clarification, so even though the `IndexSet` tracks the order of the insertions, there's no way to look up all keys inserted *after* a specific key `k`, which is where the `BtreeMap` has the advantage?

carllin (approved):

core/src/crds.rs
behzadnouri: Current code only uses `IndexSet` because of faster sequential iterations (due to cache locality). It does not actually preserve insertion order. For example the `swap_remove` in these lines:
https://github.com/solana-labs/solana/blob/8aec50a27/core/src/crds.rs#L410
https://github.com/solana-labs/solana/blob/8aec50a27/core/src/crds.rs#L444
and, also in the case a value is updated, `self.votes` does not change:
https://github.com/solana-labs/solana/blob/8aec50a27/core/src/crds.rs#L188-L203

So, ultimately current index does not maintain insertion orders.

behzadnouri:
`cluster_info_stats::get_votes` on `mainnet` with so far ~50% upgrading to this patch (1.6.10):

get_votes_mainnet

carllin:
> `cluster_info_stats::get_votes` on `mainnet` with so far ~50% upgrading to this patch (1.6.10):
> 
> get_votes_mainnet

Whoo awesome 😃

PR #17351, opened 2021-05-20
adds metric for turbine retransmit tree mismatch

behzadnouri:
#### Problem
In order to remove port-based forwarding logic in turbine, we need to first track how often the turbine retransmit/broadcast trees mismatch across nodes. One consistency condition is that if the node is on the critical path (i.e. the first node in each neighborhood), then we expect that the packet arrives at tvu socket as opposed to tvu-forwards.

#### Summary of Changes
* 1st commit adds a metric to track how often above consistency condition is not met.
* 2nd commit just removes the nested for loop. No code logic change.

### DIFF
--- a/core/src/retransmit_stage.rs
+++ b/core/src/retransmit_stage.rs
@@ -65,6 +65,7 @@ struct RetransmitStats {
     retransmit_total: AtomicU64,
     last_ts: AtomicU64,
     compute_turbine_peers_total: AtomicU64,
+    retransmit_tree_mismatch: AtomicU64,
     packets_by_slot: Mutex<BTreeMap<Slot, usize>>,
     packets_by_source: Mutex<BTreeMap<String, usize>>,
 }
@@ -83,6 +84,7 @@ fn update_retransmit_stats(
     packets_by_source: HashMap<String, usize>,
     epoch_fetch: u64,
     epoch_cach_update: u64,
+    retransmit_tree_mismatch: u64,
 ) {
     stats.total_time.fetch_add(total_time, Ordering::Relaxed);
     stats
@@ -105,6 +107,9 @@ fn update_retransmit_stats(
     stats
         .epoch_cache_update
         .fetch_add(epoch_cach_update, Ordering::Relaxed);
+    stats
+        .retransmit_tree_mismatch
+        .fetch_add(retransmit_tree_mismatch, Ordering::Relaxed);
     {
         let mut stats_packets_by_slot = stats.packets_by_slot.lock().unwrap();
         for (slot, count) in packets_by_slot {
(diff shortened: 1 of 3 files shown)


core/src/retransmit_stage.rs
```diff
@@ -315,15 +325,15 @@ fn retransmit(
     first_shreds_received: &Mutex<BTreeSet<Slot>>,
     rpc_subscriptions: &Option<Arc<RpcSubscriptions>>,
 ) -> Result<()> {
-    let timer = Duration::new(1, 0);
+    const RECV_TIMEOUT: Duration = Duration::from_secs(1);
     let r_lock = r.lock().unwrap();
-    let packets = r_lock.recv_timeout(timer)?;
+    let packets = r_lock.recv_timeout(RECV_TIMEOUT)?;
```
CriesofCarrots, line 330:
```suggestion
    let packets = r_lock.recv_timeout(Duration::from_secs(1))?;
```
Could this just be a one-liner? Or do you expect to use the const elsewhere?

core/src/retransmit_stage.rs
```diff
@@ -368,91 +378,97 @@ fn retransmit(
...
+        let (my_index, shuffled_stakes_and_index) = ClusterInfo::shuffle_peers_and_index(
+            &my_id,
+            &r_epoch_stakes_cache.peers,
+            &stakes_and_index,
+            packet.meta.seed,
+        );
+        // If the node is on the critical path (i.e. the first node in each
+        // neighborhood), then we expect that the packet arrives at tvu socket
+        // as opposed to tvu-forwards. If this is not the case, then the
+        // turbine broadcast/retransmit tree mismatch across nodes.
```
CriesofCarrots, line 429:
```suggestion
        // turbine broadcast/retransmit tree is mismatched across nodes.
```
?

CriesofCarrots (approved):
Seems straightforward. Couple nits.

core/src/retransmit_stage.rs
```diff
@@ -368,91 +378,97 @@ fn retransmit(
...
+        );
+        let (my_index, shuffled_stakes_and_index) = ClusterInfo::shuffle_peers_and_index(
+            &my_id,
+            &r_epoch_stakes_cache.peers,
+            &stakes_and_index,
+            packet.meta.seed,
+        );
+        // If the node is on the critical path (i.e. the first node in each
+        // neighborhood), then we expect that the packet arrives at tvu socket
+        // as opposed to tvu-forwards. If this is not the case, then the
```
carllin, line 428: This is a clever way to detect this, nice!

carllin (approved):

core/src/retransmit_stage.rs
behzadnouri: done, thanks

core/src/retransmit_stage.rs
behzadnouri: I guess I picked up this habit from previous jobs, to use consts to avoid magic numbers in the code as a means of (mini) documentation and maybe more readability. I guess here it does not matter much either case.

PR #17354, opened 2021-05-20
remove flatten in storage scan

jeffwashington:
#### Problem
unnecessary flatten in hash calculation store scan
#### Summary of Changes
remove flatten
Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -4099,8 +4099,8 @@ impl AccountsDb {
             .map(|storages: &[Vec<Arc<AccountStorageEntry>>]| {
                 let mut retval = B::default();
 
-                for storage in storages.iter() {
-                    for storage in storage {
+                for sub_storages in storages {
+                    for storage in sub_storages {
                         let accounts = storage.accounts.accounts(0);
                         accounts.into_iter().for_each(|stored_account| {
                             scan_func(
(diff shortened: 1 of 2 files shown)


jeffwashington:
metrics on mnb snapshot:
```
with this change: (before I removed the pre_scan_flatten metric completely)
datapoint: calculate_accounts_hash_without_index accounts_scan=2210702i eliminate_zeros=443083i hash=235583i sort=1851620i hash_total=61837974i flatten=1701824i unreduced_entries=62270664i num_snapshot_storage=372887i pre_scan_flatten=0i total=6442812i
master: pending
```

runtime/src/accounts_db.rs
```diff
@@ -4085,35 +4085,31 @@ impl AccountsDb {
...
-                for storage in storages {
-                    let accounts = storage.accounts.accounts(0);
-                    accounts.into_iter().for_each(|stored_account| {
-                        scan_func(
-                            LoadedAccount::Stored(stored_account),
-                            &mut retval,
-                            storage.slot(),
-                        )
-                    });
+                for storage in storages.iter() {
```
brooksprumo, line 4102: Is the `.iter()` needed in this case?

(Legitimately asking because I don't know, and I was under the impression that Rust had syntactic sugar for this case.)

jeffwashington: Turns out it is not required. IIRC, it was when I was initially getting it to compile, but no longer is... curious.

brooksprumo (approved):
Looks good to me!

PR #17363, opened 2021-05-20
ancestors is set instead of map

jeffwashington:
#### Problem
It turns out Ancestors doesn't need to be HashMap and keep track of the a usize value per slot (except we need to persist to/from for bank serialization format).

#### Summary of Changes
Change Ancestors api to be like a HashSet instead of HashMap. Next, I will refactor Ancestors to internally be a bit vector instead of the more complicated implementation we have now.

Fixes #

### DIFF
--- a/runtime/src/ancestors.rs
+++ b/runtime/src/ancestors.rs
@@ -83,7 +83,7 @@ impl From<&Ancestors> for HashMap<Slot, usize> {
     fn from(source: &Ancestors) -> HashMap<Slot, usize> {
         let mut result = HashMap::with_capacity(source.len());
         source.keys().iter().for_each(|slot| {
-            result.insert(*slot, *source.get(slot).unwrap());
+            result.insert(*slot, 0);
         });
         result
     }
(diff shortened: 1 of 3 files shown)


runtime/src/ancestors.rs
```diff
@@ -83,7 +83,7 @@ impl From<&Ancestors> for HashMap<Slot, usize> {
     fn from(source: &Ancestors) -> HashMap<Slot, usize> {
         let mut result = HashMap::with_capacity(source.len());
         source.keys().iter().for_each(|slot| {
-            result.insert(*slot, *source.get(slot).unwrap());
+            result.insert(*slot, 0);
```
carllin, line 86: For future reference,  this `From` implementation is used for serializing this `ancestor` set into the snapshot.

Before this change, the `value` represented the distance between the ancestor and the bank being serialized, here we're just setting it to 0.

This is fine to do because currently that distance between the root and the ancestor is not being used anywhere (double checking with @jeffwashington I think we only use the keys)?

carllin (approved):

runtime/src/ancestors.rs
jeffwashington: I don't see any code that uses the values.

jeffwashington: In 1.5, I also do not see anyone using the value of the hashmap.

PR #17366, opened 2021-05-21
Prevent withrawing Initialized stake account to rent-exempt reserve

CriesofCarrots:
#### Problem
It is possible to end up with a Delegated stake account with 0 delegated stake by withdrawing an Initialized account to exactly the rent-exempt reserve and then delegating. This is confusing to users (is it activating? can it be redelegated?), and defies our intention that every stake account have more lamports than the rent-exempt reserve.

#### Summary of Changes
Bump `reserve` for Initialized stake accounts by 1 lamport
Also flesh out testing of withdrawals from Initialized accounts

### DIFF
--- a/programs/stake/src/stake_state.rs
+++ b/programs/stake/src/stake_state.rs
@@ -881,6 +881,7 @@ pub trait StakeAccount {
         stake_history: &StakeHistory,
         withdraw_authority: &KeyedAccount,
         custodian: Option<&KeyedAccount>,
+        prevent_withdraw_to_zero: bool,
     ) -> Result<(), InstructionError>;
 }
 
@@ -1219,6 +1220,7 @@ impl<'a> StakeAccount for KeyedAccount<'a> {
         stake_history: &StakeHistory,
         withdraw_authority: &KeyedAccount,
         custodian: Option<&KeyedAccount>,
+        prevent_withdraw_to_zero: bool,
     ) -> Result<(), InstructionError> {
         let mut signers = HashSet::new();
         let withdraw_authority_pubkey = withdraw_authority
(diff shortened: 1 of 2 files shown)


CriesofCarrots:
@t-nelson , I did examine all the places in the Stake program where we ensure resulting accounts have balances > rent-exempt-reserve, as we discussed. But I didn't see a great way to unify the logic. This seemed low churn and the most consistent with the withdraw handling for Delegated account.

t-nelson (approved):
Nice!  lgtm

jon-chuang:
I think a similar problem is faced with splitting accounts. 

To combat this - rather than go for the somewhat confusing reliance on credits_observed as a indirect proxy to check if balances are too small, one can ensure that split results in accounts with at least 1 SOL staked.

Having at least 1 SOL staked would combat stake fragmentation too. Not sure if this should be applied uniformly and would cause churn in current deployment.

CriesofCarrots:
> I think a similar problem is faced with splitting accounts.

@jon-chuang I do not think it is possible to generate a 0-stake account via `StakeAccount::split()`, but if you have a test or STR that proves me wrong, I would love to see it 🙏

jon-chuang:
I think I've probably made a mistake - the 1 in the code is 1 lamport, not 1 SOL - there are plenty of stakers presently w/ < 1 SOL. 

Also you're probably right but I'll go and double check.

joncinque:
Nice! I had been wondering about this for some time

programs/stake/src/stake_state.rs
```diff
@@ -1248,8 +1250,13 @@ impl<'a> StakeAccount for KeyedAccount<'a> {
             StakeState::Initialized(meta) => {
                 meta.authorized
                     .check(&signers, StakeAuthorize::Withdrawer)?;
+                let reserve = if prevent_withdraw_to_zero {
+                    checked_add(meta.rent_exempt_reserve, 1)? // stake accounts must have a balance > rent_exempt_reserve
```
ryoqun, line 1254: here

PR #17367, opened 2021-05-21
Update getrandom bpf dependency

jackcmay:
#### Problem

Newer version of the `rand` crate do not include the `dummy` feature.

#### Summary of Changes

Document the way to include `rand = 0.2`

Fixes #


mvines:
lgtm

Issue #17088, opened 2021-05-06
Incremental Snapshots


brooksprumo:
# Problem

Startup time for nodes is slow when having to download a large snapshot.

# Proposed Solution

Incremental snapshots! Instead of always having to download large, full snapshots, download a full snapshot once/less often, then download a small incremental snapshot. The expectation/hope/thought is that only a small number of accounts are touched often, so use incremental snapshots to optimize for that behavior. At startup, now a node with an existing full snapshot only needs to download a new, small incremental snapshot.

- Create full snapshots much less often (?)
  - every 100,000 slots? at epoch? `SLOTS_PER_EPOCH / 2 - 1`?
- Create incremental snapshots every 100 slots (?)
- Each incremental snapshot is the difference from the last _full_ snapshot
- Old incremental snapshots can be cleaned up, but save at least one extra as fallback
- Add a new snapshot field to gossip to differentiate between full and incremental snapshots
  - The gossip info for incremental snapshots will need to include the slot of the full snapshot that this incremental snapshot is based on

### Example

```
slot 100,000: full snapshot (A)
slot 100,100: incremental snapshot (B)
slot 100,200: incremental snapshot (C)
...
slot 1xx,x0: incremental snapshot (D)
...
slot 200,000: full snapshot (E)
```

- Incremental snapshot (ISS for short) B is the diff between full snapshot (FSS) A and ISS B. Similarly, `ISS C = diff(A, C)`, and so on.
- The latest snapshot is still the valid snapshot. If the latest snapshot is an incremental snapshot, replay the FSS then the ISS.
- Incremental snapshots _older_ than a full snapshot can be deleted (i.e. FSS E supersedes FSS A, and ISS B, C, and D).
- When ISS D is created, ISS B can be deleted.
- If at a slot between D and E, a new node would query for FSS A and then ISS D through gossip.

#### Storing an Incremental Snapshot

1. Get the slot from the last full snapshot
2. Snapshot the bank (same as for FSS)
3. Snapshot the status cache (slot deltas) (same as FSS)
4. Package up the storages (AppendVecs) from _after_ the FSS
5. Make archive

#### Loading from an Incremental Snapshot

1. Get the highest full snapshot as done now
2. Get the highest incremental snapshot based on the full snapshot from above
3. Extract full snapshot
4. Extract incremental snapshot
5. Rebuild the AccountsDb from the storages in both FSS and ISS
6. Rebuild the Bank from the ISS

#### Validator 

[...]

PR #17369, opened 2021-05-21
Flatten SnapshotStorages

brooksprumo:
`AccountDb::SnapshotStorages` is a `Vec<Vec<>>` around
`Arc<AccountStorageEntry>`.  It is created in
`AccountsDb::get_snapshot_storages()`, but pretty much every usage of
`SnapshotStorages` immediately flattened the `Vec<Vec<>>`, since the
multidimensionality was not useful nor necessary.

This commit changes `get_snapshot_storages()` to output the flattened
Vec directly (i.e. `Vec<Arc<AccountStorageEntry`).  This simplifies the
callers usage of this function, as they no longer have to flatten
themselves.

However, this change does come at a cost.  The snapshot format changes.
The ABI is changed because the datatype of the snapshot storages
changed.

So, I've put up this Draft PR to get some feedback. I do not want to go through the trouble of creating a new snapshot format if this change is not worthwhile. What do you think?

@ryoqun @carllin @sakridge

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -164,8 +164,8 @@ impl ZeroLamport for AccountInfo {
 
 /// An offset into the AccountsDb::storage vector
 pub type AppendVecId = usize;
-pub type SnapshotStorage = Vec<Arc<AccountStorageEntry>>;
-pub type SnapshotStorages = Vec<SnapshotStorage>;
+pub type SnapshotStorageEntry = Arc<AccountStorageEntry>;
+pub type SnapshotStorage = Vec<SnapshotStorageEntry>;
 
 // Each slot has a set of storage entries.
 pub(crate) type SlotStores = Arc<RwLock<HashMap<usize, Arc<AccountStorageEntry>>>>;
@@ -4079,7 +4079,7 @@ impl AccountsDb {
 
     /// Scan through all the account storage in parallel
     fn scan_account_storage_no_bank<F, B>(
-        snapshot_storages: &[SnapshotStorage],
+        snapshot_storages: &[SnapshotStorageEntry],
         stats: &mut crate::accounts_hash::HashStats,
         scan_func: F,
     ) -> Vec<B>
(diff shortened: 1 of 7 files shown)


runtime/src/accounts_db.rs
```diff
@@ -4979,6 +4978,7 @@ impl AccountsDb {
                     .collect()
             })
             .filter(|snapshot_storage: &SnapshotStorage| !snapshot_storage.is_empty())
+            .flatten()
```
brooksprumo, line 4981: Here's the main change, just adding a `.flatten()`.

runtime/src/serde_snapshot/future.rs
```diff
@@ -243,14 +243,13 @@ impl<'a> TypeContext<'a> for Context {
...
         let entries =
             serialize_iter_as_map(serializable_db.account_storage_entries.iter().map(|x| {
-                *entry_count.borrow_mut() += x.len();
                 (
-                    x.first().unwrap().slot(),
+                    x.slot(),
                     serialize_iter_as_seq(
-                        x.iter()
+                        std::iter::once(x)
                             .map(|x| Self::SerializableAccountStorageEntry::from(x.as_ref())),
```
brooksprumo, line 253: And here's what causes the snapshot format to break/change.

carllin:
@brooksprumo this could be more hassle than it's worth, but if you wanted to pursue this I think you would have to

Release an update with
1) A new `SnapshotVersion` and upgrade everyone to have the logic to decode that new snapshot version `X`. 
2) Gate creating the new snapshot version `X`
3) Once everyone's updated, activate the feature.

brooksprumo:
> @brooksprumo this could be more hassle than it's worth, but if you wanted to pursue this I think you would have to:
> 
> Release an update with
> 
> 1. A new `SnapshotVersion` and upgrade everyone to have the logic to decode that new snapshot version `X`.
> 2. Feature gate creating the new snapshot version `X`
> 3. Once everyone's updated, activate the feature.

@carllin OK yeah, that sounds like quite a hassle for the given changes... I think for now I won't pursue this further. If I end up needing to bump the snapshot version for Incremental Snapshots (#17088), I may pull this in then.

Issue #16505, opened 2021-04-13
Integrate RBPF CLI in Solana SDK


dmakarov:
#### Problem

RBPF CLI is not very accessible to developers

#### Proposed Solution

Move the CLI source code into solana repository and make it part of the SDK.

dmakarov:
cc: @jackcmay @Lichtso @mvines: following up #16454

mvines:
Instead of moving the code, we could alternatively just install the latest published version fo the RBPF CLI into the full Solana cli release, like we do for `spl-token` tool: https://github.com/solana-labs/solana/blob/70f3f7e679baa6a66738acb57d5d813e00c7195a/scripts/cargo-install-all.sh#L129

jackcmay:
To align the cli tool with what is happening in the solana runtime we might
need to finally bring rbpf into the monorepo.  We might also want to
augment program-test so that it can produce the same outputs as the cli
tool or somehow run program-tests within the cli tool.

Lichtso:
> Instead of moving the code, we could alternatively just install the latest published version fo the RBPF CLI into the full Solana cli release, like we do for `spl-token` tool.

The problem is not only that it is hard to find in the first place, but also that it is not integrated. It is stand alone and the user has to manually extract program and data and then feed them separately into the CLI tool.


> To align the cli tool with what is happening in the solana runtime we might need to finally bring rbpf into the monorepo. We might also want to augment program-test so that it can produce the same outputs as the cli tool or somehow run program-tests within the cli tool.

I hope the RBPF repo can stay separate for now. I would move the CLI bin code into the lib part of the crate and then call it from within the monorepo.

Lichtso:
I [moved the most of the CLI bin code into the lib](https://github.com/solana-labs/rbpf/pull/159), as planned. So with the next release of RBPF we can implement the CLI in the mono repo as well.

dmakarov:
I'd like to get this issue resolved.  I'm not sure -- have we reached an agreement about moving the CLI tool into solana repository?  If yes, where should I move it?  Does `sdk/rbpf-cli/` sound ok?

Lichtso:
How about placing it at the root of the repo, the `solana-cli` and `ledger-tool` are similar and also at the root.

dmakarov:
> How about placing it at the root of the repo, the `solana-cli` and `ledger-tool` are similar and also at the root.

[...]

PR #17378, opened 2021-05-21
Add rbpf-cli tool

dmakarov:
#### Problem

Developers need a tool to run and debug their BPF programs outside the heavy machinery of the blockchain execution environment.

#### Summary of Changes

Adding a command line tool that executes BPF compiled modules on an RBPF instance.

Fixes #

This is an initial PR to get started with #16505 and solicit comments. Currently the tool depends on the top of the `main` branch of solana-labs/rbpf repository.  When the recent changes in RBPF are propagated to a released version, the dependency will be adjusted to a proper release version.

cc: @mvines @jackcmay @Lichtso


rbpf-cli/src/main.rs
```diff
@@ -0,0 +1,241 @@
...
+        println!(
+            "Syscall {}: {:#x}, {:#x}, {:#x}, {:#x}, {:#x}",
+            self.name, arg1, arg2, arg3, arg4, arg5,
+        );
+        *result = Result::Ok(0);
+    }
+}
+
+fn main() {
+    let matches = App::new("Solana RBPF CLI")
```
jackcmay, line 56: Why include the "R" in "RBPF"?  We should probably remove it internally too but at least publicly we should

rbpf-cli/src/main.rs
```diff
@@ -0,0 +1,241 @@
...
+                .long("trace"),
+        )
+        .arg(
+            Arg::new("profile")
+                .about("Display profile using tracing instrumentation")
+                .short('p')
+                .long("prof"),
+        )
+        .arg(
+            Arg::new("verify")
```
jackcmay, line 127: If this is the verifier in the rbpf crate then it won't match the verifier we use on-chain.  I'm going to consolidate these

jackcmay:
@dmakarov @Lichtso Looks like cli inputs could use some work.  If we move it to the mono-repo we should do that work before releasing it in the solana tools.

dmakarov:
> @dmakarov @Lichtso Looks like cli inputs could use some work. If we move it to the mono-repo we should do that work before releasing it in the solana tools.

I changed the input handling to read instruction data from a json file and serialize parameters along with the data.

rbpf-cli/src/main.rs
```diff
@@ -0,0 +1,226 @@
...
+}
+
+fn main() {
+    solana_logger::setup();
+    let matches = App::new("Solana BPF CLI")
+        .version("0.2.8")
+        .author("Solana Maintainers <maintainers@solana.foundation>")
+        .about("CLI to test and analyze eBPF programs")
+        .arg(
+            Arg::new("assembler")
```
jackcmay, line 48: Can we auto-detect this by looking for elf magic?

rbpf-cli/src/main.rs
```diff
@@ -0,0 +1,226 @@
...
+            Arg::new("elf")
+                .about("Load ELF as eBPF executable")
+                .short('e')
+                .long("elf")
+                .value_name("FILE")
+                .takes_value(true)
+                .required_unless_present("assembler"),
+        )
+        .arg(
+            Arg::new("input")
```
jackcmay, line 66: What entrypoint does this assume?

rbpf-cli/src/main.rs
```diff
@@ -0,0 +1,226 @@
...
+            Arg::new("memory")
+                .about("Heap memory for the program to run on")
+                .short('m')
+                .long("mem")
+                .value_name("BYTES")
+                .takes_value(true)
+                .default_value("0"),
+        )
+        .arg(
+            Arg::new("use")
```
jackcmay, line 84: Some description of these would be nice, for example, what is "cfg"?

rbpf-cli/src/main.rs
```diff
@@ -0,0 +1,226 @@
...
+        .arg(
+            Arg::new("instruction limit")
+                .about("Limit the number of instructions to execute")
+                .short('l')
+                .long("lim")
+                .takes_value(true)
+                .value_name("COUNT")
+                .default_value(&std::i64::MAX.to_string()),
+        )
+        .arg(
```
jackcmay, line 101: Can we direct the trace directly to a file?

rbpf-cli/src/main.rs
```diff
@@ -0,0 +1,226 @@
...
+            Arg::new("trace")
+                .about("Display trace using tracing instrumentation")
+                .short('t')
+                .long("trace"),
+        )
+        .arg(
```
jackcmay, line 107: Can we direct the profile output directly to a file?

jackcmay:
Folks are immediately going to try and run their solana programs here but the cli doesn't support syscalls, CPIs, etc...  At the very least the cli help page should spell out exactly the capabilities and limitations of this tool.

rbpf-cli/src/main.rs
dmakarov: It is directed to a file `profile.dot`.

rbpf-cli/src/main.rs
dmakarov: Yes, good point.

rbpf-cli/src/main.rs
dmakarov: I'll add descriptions.

rbpf-cli/src/main.rs
dmakarov: I guess the default `entrypoint`, what else?..

rbpf-cli/src/main.rs
dmakarov: Sounds like a good idea.

dmakarov:
> Folks are immediately going to try and run their solana programs here but the cli doesn't support syscalls, CPIs, etc... At the very least the cli help page should spell out exactly the capabilities and limitations of this tool.

Syscalls actually routed to Solana syscall handlers. What else other than CPI is not supported? I'll add this to the help page.

rbpf-cli/src/main.rs
jackcmay: I mean, what is the signature of the entrypoint?  Is it program id, account info, data or just a byte array?

jackcmay:
Not all the syscalls are supported, for example the sysvars syacalls probably are not?  If so, where does that data come from?

rbpf-cli/src/main.rs
dmakarov: What happens is that given a json file like this

```
{
    "accounts": [
        {
            "lamports": 0,
            "data": [0, 0, 0, 1],
            "owner": [
                0, 0, 0, 0, 0, 0, 0, 0,
                0, 0, 0, 0, 0, 0, 0, 0,
                0, 0, 0, 0, 0, 0, 0, 0,
                0, 0, 0, 0, 0, 0, 0, 0
            ],
            "executable": false,
            "rent_epoch": 0
        }
    ],
    "insndata": []
}
```
a vector of KeyedAccount objects is created.  For this specific example, KeyedAccount will wrap around a AccountSharedData object, with key generated by a call to `solana_sdk::pubkey::new_rand()`. Then the vector is passed to `solana_bpf_loader_program::serialization::serialize_parameters`, and the generated bytearray is passed to a function that creates an RBPF VM instance as `parameter_bytes` parameter.

dmakarov:
> Not all the syscalls are supported, for example the sysvars syacalls probably are not? If so, where does that data come from?

Right, if run-time needs some information from a blockchain or something like that, it is not available, of course. So what `sysvars` syscall will do in this case, I don't know.

dmakarov:
Make several changes. Hopefully, addressed all comments.

rbpf-cli/src/main.rs
jackcmay: https://github.com/solana-labs/rbpf/pull/177

jackcmay:
> > Not all the syscalls are supported, for example the sysvars syacalls probably are not? If so, where does that data come from?
> 
> Right, if run-time needs some information from a blockchain or something like that, it is not available, of course. So what `sysvars` syscall will do in this case, I don't know.

Be cool if we could send our bpf test programs through the cli to track that it's behavior is inline with the regular runtime.  Maybe the cli should run the regular runtime...  First question is why can't I run my program that uses sysvars and cpi?

rbpf-cli/src/main.rs
dmakarov: > > > Not all the syscalls are supported, for example the sysvars syacalls probably are not? If so, where does that data come from?
> > 
> > 
> > Right, if run-time needs some information from a blockchain or something like that, it is not available, of course. So what `sysvars` syscall will do in this case, I don't know.
> 
> Be cool if we could send our bpf test programs through the cli to track that it's behavior is inline with the regular runtime. Maybe the cli should run the regular runtime... First question is why can't I run my program that uses sysvars and cpi?

Can we iterate with that in later PRs? It would be good (it seems) to start with something and improve it over time. As is now the tool doesn’t degrade the existing functionality of the cli tool in solana_rbpf repository, but makes it a bit more accessible to users. There will be issues and questions, of course...

jackcmay (approved):
Approved to move it into the monorepo, but needs more work.  Lets not release it as part of the solana sdk tools until we have further review

PR #17382, opened 2021-05-21
Remove const qualifier from syscall out-parameters

machacekch:
#### Problem
In the C SDK header, syscalls' out-parameters are currently declared const, even though (being out-parameters) they are being written to by the function.

#### Summary of Changes
Removed const.


jackcmay (approved):

jackcmay:
@mvines Should the compiler less restrictive, any objections to throwing this into v1.6?

mvines:
> @mvines Should the compiler less restrictive, any objections to throwing this into v1.6?

no objections from me

PR #17391, opened 2021-05-21
account-decoder: don't use strings to convert between Pubkey types

polachok:
#### Problem

It doesn't make sense to allocate a string for this operation

#### Summary of Changes

Use arrays instead


CriesofCarrots (commented):
Thank you, this is better!
There are a couple other places we use this pattern. Would you mind updating those as well?
https://github.com/solana-labs/solana/blob/ff0e623d302baa0348f6d3158cfb08a978a282bc/transaction-status/src/parse_associated_token.rs#L11
https://github.com/solana-labs/solana/blob/ff0e623d302baa0348f6d3158cfb08a978a282bc/transaction-status/src/parse_instruction.rs#L24-L25

polachok:
Sure, done

CriesofCarrots (approved):
Thank you!

PR #17397, opened 2021-05-21
Lijun issue #17174 snapshot download enhancement

### LINKED ISSUE #17174: Validator could score snapshot download to optimize speed
#### Problem

Validators choose the newest snasphot by slot from the trusted validator set without any regard for how quickly it might download that snapshot. Sometimes this results in very slow download speeds when the validator tries to download a snapshot from across the world.

#### Proposed Solution

The validator could perform some kind of ping test or short download test of a couple different hosts to rate them on how fast it would be able to download a snapshot. Looking at some stats from the initial gossip connection is an idea.

lijunwangs:
#### Problem
Validators choose the newest snasphot by slot from the trusted validator set without any regard for how quickly it might download that snapshot. Sometimes this results in very slow download speeds when the validator tries to download a snapshot from across the world.
#### Summary of Changes
1. Allow the validator bootstrap code to specify the minimal snapshot download speed. If the snapshot download speed is detected below that, a different RPC can be retried. The default is 10MB/sec.
2. To prevent spinning on a number of sub-optimal choices and not making progress, the abort/retry logic is implemented with the following safe guards: 2.1 at maximum we do this retry for 5 times. 2.2 if the download in one notification round (5 second) is more than 2%, do not do retry -- it is not as bad anyway. 2.3 if the remaining estimate time is less than 1 minutes, do not abort retry as it will be done quickly anyway. 2.4 We do this abort/retry logic only at the first notification to avoid wasting download efforts -- the reasoning is being opportunistic and too greedy may not achieve overall shorter download time.
3. The download_snapshot and download_file is modified with the option allowing caller to notified of download progress via a callback. This allows the business logic of retrying to the place it belongs.

#### Tests
1. Cargo test
2. Manually start the validator against the mainnet with different minimal-snapshot-download-speed speed to observe the retry and non-retry situation and the maximum retry. Multiple validator restart also at different time can also observe the different snapshot sizes and see the download_size influencing the logic.
3. Still looking into enhancing unit test to improve coverage 
Fixes #17174

### DIFF
--- a/validator/src/main.rs
+++ b/validator/src/main.rs
@@ -85,7 +85,7 @@ enum Operation {
 
 const EXCLUDE_KEY: &str = "account-index-exclude-key";
 const INCLUDE_KEY: &str = "account-index-include-key";
-const MIN_SNAPSHOT_DOWNLOAD_SPEED: f32 = 10485760_f32;
+const DEFAULT_MIN_SNAPSHOT_DOWNLOAD_SPEED: f32 = 10485760_f32;
 const MAX_DOWNLOAD_ABORT: u32 = 5;
 
 fn monitor_validator(ledger_path: &Path) {
@@ -814,6 +814,7 @@ fn rpc_bootstrap(
     maximum_local_snapshot_age: Slot,
     should_check_duplicate_instance: bool,
     start_progress: &Arc<RwLock<ValidatorStartProgress>>,
+    minimal_snapshot_download_speed: f32,
 ) {
     if !no_port_check {
         let mut order: Vec<_> = (0..cluster_entrypoints.len()).collect();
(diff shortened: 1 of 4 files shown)


validator/src/main.rs
```diff
@@ -83,6 +85,8 @@ enum Operation {
 
 const EXCLUDE_KEY: &str = "account-index-exclude-key";
 const INCLUDE_KEY: &str = "account-index-include-key";
+const DEFAULT_MIN_SNAPSHOT_DOWNLOAD_SPEED: f32 = 10485760_f32;
```
mvines, line 88: Can we avoid a float here?  Also please document the units used

validator/src/main.rs
```diff
@@ -958,7 +964,31 @@ fn rpc_bootstrap(
...
+                                       && download_progress.notification_count <= 1
+                                       && download_progress.percentage_done <= 2_f32
+                                       && download_progress.estimated_remaining_time > 60_f32
+                                       && download_abort_count < MAX_DOWNLOAD_ABORT {
+                                        if let Some(ref trusted_validators) = validator_config.trusted_validators {
+                                            if trusted_validators.contains(&rpc_contact_info.id) && trusted_validators.len() == 1 {
+                                                return true; // Do not abort download from the one-and-only trusted validator
+                                            }
+                                        }
+                                        warn!("The snapshot download is too slow, throughput: {} < min speed {} bytes/sec, try a different node. Abort count: {}, Progress detail: {:?}",
```
mvines, line 980: I think this message is more accurate right?
```suggestion
                                        warn!("The snapshot download is too slow, throughput: {} < min speed {} bytes/sec, will abort and try a different node. Abort count: {}, Progress detail: {:?}",
```

validator/src/main.rs
```diff
@@ -83,6 +85,8 @@ enum Operation {
 
 const EXCLUDE_KEY: &str = "account-index-exclude-key";
 const INCLUDE_KEY: &str = "account-index-include-key";
+const DEFAULT_MIN_SNAPSHOT_DOWNLOAD_SPEED: f32 = 10485760_f32;
+const MAX_DOWNLOAD_ABORT: u32 = 5;
```
mvines, line 89: Let's make the max abort count a command-line option as well

validator/src/main.rs
lijunwangs: Ack

validator/src/main.rs
lijunwangs: Sure

validator/src/main.rs
lijunwangs: Will do

lijunwangs:
How can I reopen the PR?

PR #17399, opened 2021-05-21
rpc: add context toggle to getProgramAccounts

00nktk:
#### Problem
There is no way in current implementation to get `getProgramAccounts` response with context. 

#### Summary of Changes
Added an optional flag (`with_context`) in `getProgramAccounts` config to set the format of the response. The implementation is simple: if it is present and is `true`, the response will be wrapped in RpcResponse, otherwise it defaults to array. While it feels like returning RpcResponse should be the default behavior, there is a big chance that someone relies on the current format of the API. This might seem like not the most elegant solution, but it's backwards compatible.

### DIFF
--- a/cli/src/cluster_query.rs
+++ b/cli/src/cluster_query.rs
@@ -1676,11 +1676,11 @@ pub fn process_show_stakes(
     progress_bar.set_message("Fetching stake accounts...");
 
     let mut program_accounts_config = RpcProgramAccountsConfig {
-        filters: None,
         account_config: RpcAccountInfoConfig {
             encoding: Some(solana_account_decoder::UiAccountEncoding::Base64),
             ..RpcAccountInfoConfig::default()
         },
+        ..RpcProgramAccountsConfig::default()
     };
 
     if let Some(vote_account_pubkeys) = vote_account_pubkeys {
--- a/core/src/rpc.rs
+++ b/core/src/rpc.rs
@@ -10,6 +10,7 @@ use crate::{
 use bincode::{config::Options, serialize};
 use jsonrpc_core::{types::error, Error, Metadata, Result};
 use jsonrpc_derive::rpc;
+use serde::{Deserialize, Serialize};
 use solana_account_decoder::{
     parse_token::{spl_token_id_v2_0, token_amount_to_ui_amount, UiTokenAmount},
     UiAccount, UiAccountEncoding, UiDataSliceConfig,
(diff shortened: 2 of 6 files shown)


mvines (commented):
This is a nice improvement, thank you!

cli/src/cluster_query.rs
```diff
@@ -1681,6 +1681,7 @@ pub fn process_show_stakes(
             encoding: Some(solana_account_decoder::UiAccountEncoding::Base64),
             ..RpcAccountInfoConfig::default()
         },
+        with_context: None,
```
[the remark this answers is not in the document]
00nktk: Sure. Fixed

CriesofCarrots:
Nice, I bet there are a few other rpc methods that could take advantage of `OptionalContext`

core/src/rpc.rs
```diff
@@ -350,7 +361,8 @@ impl JsonRpcRequestProcessor {
         program_id: &Pubkey,
         config: Option<RpcAccountInfoConfig>,
         filters: Vec<RpcFilterType>,
-    ) -> Result<Vec<RpcKeyedAccount>> {
+        with_context: bool, // TODO: consider RpcProgramsConfig instead of separate arg
```
mvines, line 364:
```suggestion
        with_context: bool,
```

docs/src/developing/clients/jsonrpc-api.md
```diff
@@ -1780,7 +1780,7 @@ Returns all accounts owned by the provided program Pubkey
     "jsonParsed" encoding attempts to use program-specific state parsers to return more human-readable and explicit account state data. If "jsonParsed" is requested but a parser cannot be found, the field falls back to "base64" encoding, detectable when the `data` field is type `<string>`.
   - (optional) `dataSlice: <object>` - limit the returned account data using the provided `offset: <usize>` and `length: <usize>` fields; only available for "base58", "base64" or "base64+zstd" encodings.
   - (optional) `filters: <array>` - filter results using various [filter objects](jsonrpc-api.md#filters); account must meet all filter criteria to be included in results
-
+  - (optional) `withContext: bool` - wrap the returned array into RpcResponse JSON object.
```
mvines, line 1783:
```suggestion
  - (optional) `withContext: bool` - wrap the result in a RpcResponse JSON object.
```

mvines (approved):
lgtm with my documentation nits picked

00nktk:
> Nice, I bet there are a few other rpc methods that could take advantage of `OptionalContext`

Thanks! Glad to help!

mvines:
Downstream projects build fix landing via https://github.com/solana-labs/solana-program-library/pull/1766/files, once it does I'll retrigger the failed build here

PR #17405, opened 2021-05-22
Startup optimization in shrink - don't shrink non-shrinkable slots

sakridge:
#### Problem

Startup shrink and clean operations are less efficient than they could be

#### Summary of Changes

* Skip shrinking slots which do not save any space or mappings.
* Only add keys to dirty set when there is more than 1 version of the key in the index and thus has a chance to be cleaned. Detected in `generate_index`

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -1079,6 +1079,8 @@ struct ShrinkStats {
     recycle_stores_write_elapsed: AtomicU64,
     accounts_removed: AtomicUsize,
     bytes_removed: AtomicU64,
+    bytes_written: AtomicU64,
+    skipped_shrink: AtomicU64,
 }
 
 impl ShrinkStats {
@@ -1168,6 +1170,16 @@ impl ShrinkStats {
                 (
                     "bytes_removed",
                     self.bytes_removed.swap(0, Ordering::Relaxed) as i64,
+                    i64
+                ),
+                (
+                    "bytes_written",
+                    self.bytes_written.swap(0, Ordering::Relaxed) as i64,
+                    i64
+                ),
+                (
+                    "skipped_shrink",
+                    self.skipped_shrink.swap(0, Ordering::Relaxed) as i64,
                     i64
                 ),
             );
@@ -1970,6 +1982,7 @@ impl AccountsDb {
         debug!("do_shrink_slot_stores: slot: {}", slot);
         let mut stored_accounts: HashMap<Pubkey, FoundStoredAccount> = HashMap::new();
         let mut original_bytes = 0;
+        let mut num_stores = 0;
         for store in stores {
             let mut start = 0;
             original_bytes += store.total_bytes();
@@ -1993,10 +2006,12 @@ impl AccountsDb {
                 }
                 start = next;
             }
+            num_stores += 1;
         }
 
         let mut index_read_elapsed = Measure::start("index_read_elapsed");
         let mut alive_total = 0;
+
         let accounts_index_map_lock = if is_startup {
             // at startup, there is nobody else to contend with the accounts_index read lock, so it is more efficient for us to keep it held
             Some(self.accounts_index.get_account_maps_read_lock())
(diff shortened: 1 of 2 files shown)


sakridge:
This PR:
`datapoint: verify_snapshot_bank clean_us=9408005i shrink_all_slots_us=10999445i`
clean > 9.4s, shrink > 10.9s
master
` datapoint: verify_snapshot_bank clean_us=104287378i shrink_all_slots_us=55167563i`
clean > 104s, shrink > 55s

runtime/src/accounts_db.rs
```diff
@@ -8626,7 +8432,7 @@ pub mod tests {
     }
 
     #[test]
-    fn test_shrink_candidate_slots() {
+    fn test_shrink_candidate_slots1() {
```
jeffwashington, line 8435: was this accidentally put in the pr or am I missing something?

runtime/src/accounts_db.rs
```diff
@@ -1946,10 +1947,14 @@ impl AccountsDb {
                 }
                 start = next;
             }
+            num_stores += 1;
         }
 
         let mut index_read_elapsed = Measure::start("index_read_elapsed");
         let mut alive_total = 0;
+        let mut alive_accounts = Vec::with_capacity(stored_accounts.len() / 2);
```
jeffwashington, line 1955: Why is this len() / 2? Do we expect 1/2 the accounts to be alive?

runtime/src/accounts_db.rs
```diff
@@ -5262,257 +5278,47 @@ impl AccountsDb {
     // Requires all stores in the slot to be re-written otherwise the accounts_index
     // store ref count could become incorrect.
     fn do_shrink_slot_v1(&self, slot: Slot, forced: bool) -> usize {
```
jeffwashington, line 5280: Without going through this line by line is the idea here that this function is mostly the same code  as `do_shrink_slot_stores `?
@carllin had mentioned he thought it was time for us to get rid of the non cache code path and he was planning to do that at some point.

runtime/src/accounts_db.rs
```diff
@@ -1958,51 +1963,56 @@ impl AccountsDb {
...
                 }
-            })
-            .collect();
+            }
+        }
         drop(accounts_index_map_lock);
         index_read_elapsed.stop();
         let aligned_total: u64 = self.page_align(alive_total);
 
+        // If not saving maps or more than 1 page of data, then skip the shrink
```
jeffwashington, line 1996: What does 'maps' mean here? And just clarifying this should be 'maps or' vs 'maps of'?

runtime/src/accounts_index.rs
```diff
@@ -1205,7 +1209,7 @@ impl<T: 'static + Clone + IsCached + ZeroLamport> AccountsIndex<T> {
         &self,
         slot: Slot,
         items: Vec<(&Pubkey, T)>,
-    ) {
+    ) -> Vec<bool> {
```
jeffwashington, line 1212: This is a good idea. I've been working towards better leveraging the repeated scans over all storages at startup. I had imagined creating the index or calculating the hash could return by-products of useful info for clean or shrink.

runtime/src/accounts_db.rs
```diff
@@ -8940,7 +8746,7 @@ pub mod tests {
         accounts.add_root(1);
         accounts.clean_accounts(None, false);
         accounts.shrink_all_slots(false);
-        accounts.print_accounts_stats("post-shrink");
+        //accounts.print_accounts_stats("post-shrink");
```
jeffwashington, line 8749: accidental?

sakridge: Yea. removed it.

runtime/src/accounts_db.rs
sakridge: Well I was using it debugging. It's easier to select this test now, because there are a few tests with the prefix `test_shrink_candidate_slots`

runtime/src/accounts_db.rs
sakridge: memory maps

runtime/src/accounts_db.rs
sakridge: Yea, should be the same code.

runtime/src/accounts_db.rs
sakridge: Seemed like a decent guess for it

runtime/src/accounts_index.rs
```diff
@@ -1231,8 +1235,12 @@ impl<T: 'static + Clone + IsCached + ZeroLamport> AccountsIndex<T> {
                 }
                 if let Some(mut w_account_entry) = account_entry {
                     w_account_entry.update(slot, account_info, &mut _reclaims);
+                    true
+                } else {
+                    false
```
carllin, line 1240: Just to make sure I'm understanding the reasoning correctly, for this:
```
if let Some(mut w_account_entry) = account_entry {
     true
}
```

Combined with:
```
 dirty_pubkeys.retain(|_k| *dirty_pubkey_mask_iter.next().unwrap());
```
means if there was a previous entry for this pubkey we add it to the dirty pubkey list, but if there was no previous entry, we skip adding it to the dirty pubkey list. 

Is the reasoning that if this pubkey is only updated once, then it doesn't have a later update which renders the first update obsolete, which means it doesn't need to be cleaned? 

The one case where we might still need this is if the account is zero lamports so that it can be added to the zero-lamport list here: https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_db.rs#L1654-L1658

carllin: Can we also add a note/issue/todo to add a test for the `generate_index() -> clean_accounts()` flow to test some of these corner cases? This logic is getting complicated enough where I think it's warranted

runtime/src/accounts_db.rs
```diff
@@ -1958,51 +1963,56 @@ impl AccountsDb {
...
-            })
-            .collect();
+            }
+        }
         drop(accounts_index_map_lock);
         index_read_elapsed.stop();
         let aligned_total: u64 = self.page_align(alive_total);
 
+        // If not saving memory maps or more than 1 page of data, then skip the shrink
+        if num_stores == 1 && aligned_total + PAGE_SIZE >= original_bytes {
```
carllin, line 1997: I think it's also a good idea to extract this into a reusable `is_bytes_saved_sufficient_for_shrink()` that can be added to the check here where we add to the `shrink_candidates`: https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_db.rs#L4592-L4607.

Otherwise, we'll continually add to the shrink candidates as we delete accounts from the same slot, even if they continually fail this new `aligned_total + PAGE_SIZE >= original_bytes` check

carllin: Also, just out of curiosity, do we know the snapshot sizes before and after?

Theoretically if every slot had an extra `PAGE_SIZE = 4096` bytes, we might get an extra `432000 slots * 496 bytes ~= 2GB` raw extra size before compression, but wonder if the results reflect this.

runtime/src/accounts_db.rs
```diff
@@ -1958,51 +1963,56 @@ impl AccountsDb {
...
+            for pubkey in unrefed_pubkeys {
+                if let Some(locked_entry) = self.accounts_index.get_account_read_entry(pubkey) {
+                    locked_entry.addref();
```
carllin, line 2000: I wonder if we could avoid this `addref()` by either:

1) Removing the earlier `unref()`, and instead just  iterate through the `unrefed_pubkeys` and call `unref()` after we've passed the `aligned_total + PAGE_SIZE >= original_bytes` check above. Downside is an extra pass through the `unrefed_pubkeys` keys

2) We have a `store.alive_bytes()` and `store.total_bytes()`. Maybe using these we could move the `aligned_total + PAGE_SIZE >= original_bytes` check earlier in the function before the scan, and exit there if the check fails (would also avoid the overhead of the additional scan through the storages).  Something like:

`if num_stores == 1 && self.page_align(store.alive_bytes() as u64) + PAGE_SIZE >= store.total_bytes() { return 0;}`

runtime/src/accounts_index.rs
sakridge: > Is the reasoning that if this pubkey is only updated once, then it doesn't have a later update which renders the first update obsolete, which means it doesn't need to be cleaned?

Yes, exactly. If there is just 1 update for a given key, there is no cleaning that could be done unless that key is 0-lamport and that is happening through a different path.

runtime/src/accounts_db.rs
sakridge: I made a separate PR for that one: https://github.com/solana-labs/solana/pull/17385

Once that is merged I'll rebase to remove the diff. If you want to just look at the top 2 commits here, that should be indicative of the optimization changes.

runtime/src/accounts_db.rs
sakridge: Right for 1, I am thinking this is the not-common case since we queued it up to shrink and we have other signals that tell us that shrinking is likely productive like alive_bytes/written_bytes. So we should optimize for the case where we are actually doing the shrink.

Clean does run before this shrink. I was thinking those may not be accurate, hence the location of this check, but I think you might be right, they might actually be somewhat accurate.

jeffwashington (dismissed):
lgtm

runtime/src/accounts_db.rs
sakridge: I ran this for a while on mainnet. snapshot sizes didn't seem to be affected:

```
Every 2.0s: ls -alh mainnet-beta/validator-ledger/snapshot-*

-rw-rw-r-- 1 stephen_solana_com stephen_solana_com 6.3G May  4 15:00 mainnet-beta/validator-ledger/snapshot-76776676-EcV4U9gqqdXuhqgP69WaLhWmoyDQcJYxM1ZE5YuSspdF.tar.zst
-rw-rw-r-- 1 stephen_solana_com stephen_solana_com 6.6G May 27 15:13 mainnet-beta/validator-ledger/snapshot-80249526-9NwTfazvvgmdb4Wr8VnAhC1kEWqnU9APZjfpmrktjpRS.tar.zst
-rw-rw-r-- 1 stephen_solana_com stephen_solana_com 6.6G May 27 15:18 mainnet-beta/validator-ledger/snapshot-80249917-2WYkAHaKCYFEyCtA8gBF4Qxh6iNbrcxRbXUrqUQSvPai.tar.zst
```

sakridge: Left side is this change, right side master starting with that huge spike at startup.

bytes-written

runtime/src/accounts_db.rs
```diff
@@ -4592,9 +4639,7 @@ impl AccountsDb {
                 if count == 0 {
                     dead_slots.insert(*slot);
                 } else if self.caching_enabled
-                    && (self.page_align(store.alive_bytes() as u64) as f64
-                        / store.total_bytes() as f64)
-                        < SHRINK_RATIO
```
carllin, line 4597: Do we still want to keep this `SHRINK_RATIO` logic, or do we want to shrink every time we save at least a page of memory?

sakridge: Yea. I was debating this as well. The account store seems to stay tighter with this. It was around 28gb vs. 34gb for master. I was looking at the metrics for increased write IO. I think it is a bit higher, but it wasn't super noticeable in the graph. Ideally one could adjust it dynamically. https://github.com/solana-labs/solana/issues/17544

sakridge:
@carllin any other concerns on this?

carllin (commented):
Did one final pass, looks good!

PR #17411, opened 2021-05-22
sort storages by slot before scan

jeffwashington:
#### Problem
2 different use cases have emerged where it is useful to have storages sorted by slot.

1. merging write cache data with store data so we can use this scan in more places. Write cache data is organized by slot.
2. perf improvements for scanning accounts. If slots are in order, we don't have to keep track of a slot (or a write version) per slot, which is a large data savings.

#### Summary of Changes
Sort storages by slot.

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -30,6 +30,7 @@ use crate::{
     append_vec::{AppendVec, StoredAccountMeta, StoredMeta, StoredMetaWriteVersion},
     contains::Contains,
     read_only_accounts_cache::ReadOnlyAccountsCache,
+    sorted_storages::SortedStorages,
 };
 use blake3::traits::digest::Digest;
 use crossbeam_channel::{unbounded, Receiver, Sender};
@@ -4207,7 +4208,7 @@ impl AccountsDb {
 
     /// Scan through all the account storage in parallel
     fn scan_account_storage_no_bank<F, B>(
-        snapshot_storages: &[SnapshotStorage],
+        snapshot_storages: &SortedStorages,
         scan_func: F,
     ) -> Vec<B>
     where
@@ -4216,19 +4217,24 @@ impl AccountsDb {
     {
         // Without chunks, we end up with 1 output vec for each outer snapshot storage.
         // This results in too many vectors to be efficient.
-        const MAX_ITEMS_PER_CHUNK: usize = 5_000;
-        snapshot_storages
-            .par_chunks(MAX_ITEMS_PER_CHUNK)
-            .map(|storages: &[Vec<Arc<AccountStorageEntry>>]| {
+        const MAX_ITEMS_PER_CHUNK: Slot = 5_000;
+        let chunks = 1 + (snapshot_storages.range_width() as Slot / MAX_ITEMS_PER_CHUNK);
+        (0..chunks)
+            .into_par_iter()
+            .map(|chunk| {
                 let mut retval = B::default();
-
-                for sub_storages in storages {
-                    for storage in sub_storages {
-                        let slot = storage.slot();
-                        let accounts = storage.accounts.accounts(0);
-                        accounts.into_iter().for_each(|stored_account| {
-                            scan_func(LoadedAccount::Stored(stored_account), &mut retval, slot)
-                        });
+                let start = snapshot_storages.range().start + chunk * MAX_ITEMS_PER_CHUNK;
+                let end = std::cmp::min(start + MAX_ITEMS_PER_CHUNK, snapshot_storages.range().end);
+                for slot in start..end {
+                    let sub_storages = snapshot_storages.get(slot);
+                    if let Some(sub_storages) = sub_storages {
+                        for storage in sub_storages {
+                            let slot = storage.slot();
+                            let accounts = storage.accounts.accounts(0);
+                            accounts.into_iter().for_each(|stored_account| {
+                                scan_func(LoadedAccount::Stored(stored_account), &mut retval, slot)
+                            });
+                        }
                     }
                 }
                 retval
@@ -4294,7 +4300,7 @@ impl AccountsDb {
     }
 
     fn scan_snapshot_stores(
-        storage: &[SnapshotStorage],
+        storage: &SortedStorages,
         mut stats: &mut crate::accounts_hash::HashStats,
         bins: usize,
         bin_range: &Range<usize>,
@@ -4308,7 +4314,7 @@ impl AccountsDb {
         let mismatch_found = AtomicU64::new(0);
 
         let result: Vec<Vec<Vec<CalculateHashIntermediate>>> = Self::scan_account_storage_no_bank(
-            &storage,
+            storage,
             |loaded_account: LoadedAccount,
              accum: &mut Vec<Vec<CalculateHashIntermediate>>,
              slot: Slot| {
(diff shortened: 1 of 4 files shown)


runtime/src/accounts_db.rs
```diff
@@ -4230,6 +4228,19 @@ impl AccountsDb {
         result
     }
 
+    fn sort_storages(storages: &[SnapshotStorage]) -> Vec<&Arc<AccountStorageEntry>> {
+        let mut m1 = Measure::start("");
+        let mut result = storages
+            .iter()
+            .flatten()
+            .map(|item| (item.slot(), item))
```
brooksprumo, line 4236: Can this be combined to [`.flat_map()`](https://doc.rust-lang.org/beta/std/iter/trait.Iterator.html#method.flat_map)?

brooksprumo (commented):
I'm assuming the sorting helps speed up the scan? Also assuming the speed up to the scan is greater than the time to sort, yes?

runtime/src/accounts_db.rs
jeffwashington: flat map says:
```
You can think of flat_map(f) as the semantic equivalent of mapping, and then flattening as in map(f).flatten().
```
In this case, I have flatten().map(). Maybe there is an iterator way to do this.

jeffwashington: I just put this in a pr to see if anything would break in the ci and to discuss with you. The current code doesn't take advantage of the fact that the storages would be sorted yet.

I think I may rework this more anyway. I need to guarantee later that if I chunk up an array of storages, that all storages for a slot are in a single chunk. This means there can never be 2 successive elements with the same slot.

runtime/src/accounts_db.rs
```diff
@@ -4230,6 +4230,23 @@ impl AccountsDb {
...
+            .iter()
+            .map(|item| {
+                (
+                    item.first().map(|item| item.slot()).unwrap_or_default(),
+                    item,
+                )
+            })
+            .collect::<Vec<_>>();
+        m1.stop();
+        error!("storage_sort_us: {}", m1.as_us());
```
jeffwashington, line 4245: I'll get rid of this error and the timing once I see if builds have passed.

runtime/src/accounts_db.rs
```diff
@@ -4230,6 +4230,23 @@ impl AccountsDb {
...
+        result.par_sort_unstable_by(|a, b| a.0.cmp(&b.0));
+        result.into_iter().map(|(_, item)| item).collect()
```
brooksprumo, line 4247: I do not know which is faster: this version here (A), or something like this (B):

```rust
let mut result = storages.clone();
result.par_sort_unstable_by(|a, b| a.first().unwrap_or_default().slot().cmp(&b.first().unwrap_or_default().slot()));
```

I could see sorting in (A) being helped by already having the slot pulled out front, which I think could remove an indirection in the `sort_by` in (B)? But maybe the prefetcher would be able to hide that anyway? I'm not sure without profiling/looking at the IR.

Is this something that even makes a difference in time? Is it something you already checked?

runtime/src/accounts_db.rs
```diff
@@ -4103,7 +4103,7 @@ impl AccountsDb {
 
     /// Scan through all the account storage in parallel
     fn scan_account_storage_no_bank<F, B>(
-        snapshot_storages: &[SnapshotStorage],
+        snapshot_storages: &[&Vec<Arc<AccountStorageEntry>>],
```
brooksprumo, line 4106: For all these similar changes, is there a reason to not use `&[&SnapshotStorage]` instead?

runtime/src/accounts_db.rs
```diff
@@ -4230,6 +4230,23 @@ impl AccountsDb {
         result
     }
 
+    fn sort_storages(storages: &[SnapshotStorage]) -> Vec<&Vec<Arc<AccountStorageEntry>>> {
```
brooksprumo, line 4233: [Naming] Since Rust sorts in place, I think this function is surprising at first glance since it does not sort in place. In C++, many of the algorithms that make a copy are appended with `_copy`. Rust has `into` and `from` as a common thing I've seen, but I dunno if they always make copies...

Anyway, maybe something like `sort_storages_copy()`?

brooksprumo, line 4233: Lately I've been seeing Carl change function parameter types from slices/containers to iterators. [Here's an example](https://github.com/solana-labs/solana/pull/17319/files#diff-1090394420d51617f3233275c2b65ed706b35b53b115fe65f82c682af8134a6fR1504). I dunno if that is either more performant and/or more idiomatic, but figured I'd point that out here for `storages`.

brooksprumo (commented):
My comments are mostly questions :)

runtime/src/accounts_db.rs
jeffwashington: I didn't check this specifically. I did try a few iterations of changes to see what is fastest. I learned refs were better as the payload than a cloned arc. fwiw, I just completely refactored all this and it is back in progress again because of new requirements - namely that I leave a space for slots with no stores, but that are in the write cache and could be ancestors or roots.

runtime/src/accounts_db.rs
jeffwashington: arguing with clippy that wants [] instead of vec

runtime/src/sorted_storages.rs
```diff
@@ -0,0 +1,117 @@
...
+    pub fn get(&self, slot: Slot) -> Option<&SnapshotStorage> {
+        if !self.range.contains(&slot) {
+            None
+        } else {
+            let index = (slot - self.range.start) as usize;
+            self.storages[index]
+        }
+    }
+
+    pub fn range_width(&self) -> u64 {
```
brooksprumo, line 23: nit: Should this return value be `Slot` instead? I don't have strong feeling about this, just noticed it was different.

runtime/src/sorted_storages.rs
```diff
@@ -0,0 +1,117 @@
...
+        let mut min = Slot::MAX;
+        let mut max = Slot::MIN;
+        let mut count = 0;
+        let mut time = Measure::start("get slot");
+        let slots = source
+            .iter()
+            .map(|storages| {
+                count += storages.len();
+                if !storages.is_empty() {
+                    storages.first().map(|storage| {
```
brooksprumo, line 49: Is it correct that this code assumes there's only one `AccountsStorageEntry` per `storages` here? I think that's fine with the AccountsCache, yes, but does this code (need to) get used where AccountsCache is disabled? Maybe this assumption can be commented with `///` on the function and an assert?

brooksprumo (commented):
Looks good! Just the two questions/comments.

runtime/src/sorted_storages.rs
jeffwashington: No. This code assumes that everything in 'storages' (line 46) has the same answer for .slot() on line 50. first() is like storages[0] but returns None if storages is empty.

runtime/src/sorted_storages.rs
jeffwashington: done

runtime/src/sorted_storages.rs
brooksprumo: Oh duh, this is only getting the slots, not what's in each of these storages. Thanks!

brooksprumo (approved):

brooksprumo (approved):
Re-approving since mergify removed it.

PR #17412, opened 2021-05-22
add ledger-tool verify option to verify hash calculation

jeffwashington:
#### Problem
ledger-tool verify is useful for testing loading from a snapshot. Hash calculation is an important algorithm in accounts bg service. It would be nice for the non-index hash calculation to run during ledger-tool verify.
#### Summary of Changes
add a ledger-tool verify option to run the non-index hash calculation code.
Fixes #

### DIFF
--- a/ledger-tool/src/main.rs
+++ b/ledger-tool/src/main.rs
@@ -762,6 +762,9 @@ fn main() {
         .value_name("PATHS")
         .takes_value(true)
         .help("Comma separated persistent accounts location");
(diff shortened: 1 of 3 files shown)


brooksprumo (approved):
This looks good to me. Does anyone else need to approve adding new command line arguments too?

jeffwashington:
> This looks good to me. Does anyone else need to approve adding new command line arguments too?

great question. In this case, these are args to ledger-tool, which I don't think rises to the rigor of validator, for example. In this case, it is the same argument name I previously added to validator, so it seems to fall into the acceptable bucket, IMHO.

PR #17415, opened 2021-05-22
snapshot download enhancement

### LINKED ISSUE #17174: Validator could score snapshot download to optimize speed

lijunwangs:
### Problem

Validators choose the newest snasphot by slot from the trusted validator set without any regard for how quickly it might download that snapshot. Sometimes this results in very slow download speeds when the validator tries to download a snapshot from across the world.

### Summary of Changes

1. Allow the validator bootstrap code to specify the minimal snapshot download speed. If the snapshot download speed is detected below that, a different RPC can be retried. The default is 10MB/sec.
2. To prevent spinning on a number of sub-optimal choices and not making progress, the abort/retry logic is implemented with the following safe guards: 
    2.1 at maximum we do this retry for 5 times -- this number is configurable with default 5. 
    2.2 if the download in one notification round (5 second) is more than 2%, do not do retry -- it is not as bad anyway. 
    2.3 if the remaining estimate time is less than 1 minutes, do not abort retry as it will be done quickly anyway. 
    2.4 We do this abort/retry logic only at the first notification to avoid wasting download efforts -- the reasoning is being opportunistic and too greedy may not achieve overall shorter download time.
3. The download_snapshot and download_file is modified with the option allowing caller to notified of download progress via a callback. This allows the business logic of retrying to the place it belongs.

### Tests

Cargo test
1. Manually start the validator against the mainnet with different minimal-snapshot-download-speed speed to observe the retry and non-retry situation and the maximum retry. 
2. Multiple validator restarts at different time can also observe the different snapshot sizes and see the download_size influencing the logic.
3. Still looking into enhancing unit test to improve coverage

Fixes #17174

### DIFF
--- a/download-utils/src/lib.rs
+++ b/download-utils/src/lib.rs
@@ -4,7 +4,6 @@ use indicatif::{ProgressBar, ProgressStyle};
 use log::*;
 use solana_runtime::{bank_forks::ArchiveFormat, snapshot_utils};
 use solana_sdk::{clock::Slot, genesis_config::DEFAULT_GENESIS_ARCHIVE, hash::Hash};
-use std::fmt;
 use std::fs::{self, File};
 use std::io;
 use std::io::Read;
@@ -25,6 +24,7 @@ fn new_spinner_progress_bar() -> ProgressBar {
 }
 
 /// Structure modeling information about download progress
+#[derive(Debug)]
 pub struct DownloadProgressRecord {
     // Duration since the beginning of the download
     pub elapsed_time: Duration,
@@ -46,27 +46,9 @@ pub struct DownloadProgressRecord {
     pub notification_count: u64,
 }
 
-impl fmt::Debug for DownloadProgressRecord {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.debug_struct("DownloadProgress")
-            .field("elapsed_time", &self.elapsed_time)
-            .field("last_elapsed_time", &self.last_elapsed_time)
-            .field("last_throughput", &self.last_throughput)
-            .field("total_throughput", &self.total_throughput)
-            .field("last_elapsed_time", &self.last_elapsed_time)
-            .field("total_bytes", &self.total_bytes)
-            .field("current_bytes", &self.current_bytes)
-            .field("percentage_done", &self.percentage_done)
-            .field("estimated_remaining_time", &self.estimated_remaining_time)
-            .field("notification_count", &self.notification_count)
-            .finish()
-    }
-}
-
 /// This callback allows the caller to get notified of the download progress modelled by DownloadProgressRecord
 /// Return "true" to continue the download
 /// Return "false" to abort the download
-
 pub fn download_file<F>(
     url: &str,
     destination_file: &Path,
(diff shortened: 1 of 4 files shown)


download-utils/src/lib.rs
```diff
@@ -24,11 +25,58 @@ fn new_spinner_progress_bar() -> ProgressBar {
...
+    pub current_bytes: usize,
+    // percentage downloaded
+    pub percentage_done: f32,
+    // Estimated remaining time (in seconds) to finish the download if it keeps at the the last download speed
+    pub estimated_remaining_time: f32,
+    // The times of the progress is being notified, it starts from 1 and increments by 1 each time
+    pub notification_count: u64,
+}
+
+impl fmt::Debug for DownloadProgressRecord {
```
jeffwashington, line 50: maybe:
#[derive(Debug)]
Doesn't look like your debug is too special.

download-utils/src/lib.rs
```diff
@@ -84,38 +132,83 @@ pub fn download_file(
...
+            match self.response.read(buf) {
+                Ok(n) => {
                     self.current_bytes += n;
-                    if self.last_print.elapsed().as_secs() > 5 {
-                        let total_bytes_f32 = self.current_bytes as f32;
-                        let diff_bytes_f32 = (self.current_bytes - self.last_print_bytes) as f32;
+                    let total_bytes_f32 = self.current_bytes as f32;
+                    let diff_bytes_f32 = (self.current_bytes - self.last_print_bytes) as f32;
+                    let last_throughput = diff_bytes_f32 / self.last_print.elapsed().as_secs_f32();
+                    let mut estimated_remaining_time: f32 = f32::MAX;
```
jeffwashington, line 162: I think this is more rusty:
```
        let estimated_remaining_time = if last_throughput > 0_f32 {
            (self.download_size - self.current_bytes as f32) / last_throughput
        } else {
            f32::MAX
        };
```

download-utils/src/lib.rs
```diff
@@ -84,38 +132,83 @@ pub fn download_file(
...
+    impl<R: Read, F> Read for DownloadProgress<R, F>
+    where
+        F: Fn(&DownloadProgressRecord) -> bool,
+    {
         fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
-            self.response.read(buf).map(|n| {
-                if self.use_progress_bar {
-                    self.progress_bar.inc(n as u64);
-                } else {
+            match self.response.read(buf) {
```
jeffwashington, line 156: consider this:

https://github.com/jeffwashington/solana/blob/e4861e0435dc604e305e0822b0915bd4dee845a2/download-utils/src/lib.rs#L156-L210

lijunwangs: Yes. Looks cleaner!

download-utils/src/lib.rs
lijunwangs: Will do

download-utils/src/lib.rs
lijunwangs: Okay, will see how it looks in the logs.

download-utils/src/lib.rs
lijunwangs: Addressed

download-utils/src/lib.rs
lijunwangs: Addressed

download-utils/src/lib.rs
```diff
@@ -23,11 +23,41 @@ fn new_spinner_progress_bar() -> ProgressBar {
...
+/// Return "false" to abort the download
+pub fn download_file<F>(
     url: &str,
     destination_file: &Path,
     use_progress_bar: bool,
-) -> Result<(), String> {
+    progress_notify_callback: &Option<F>,
+) -> Result<(), String>
+where
+    F: Fn(&DownloadProgressRecord) -> bool,
```
mvines, line 59: nit: `Fn(&DownloadProgressRecord) -> bool` is repeated multiple times.  You could wrap that in a type and then document the true/value return behaviour at the type definition

validator/src/main.rs
```diff
@@ -891,7 +898,33 @@ fn rpc_bootstrap(
...
+                                    debug!("Download progress: {:?}", download_progress);
+
+                                    if download_progress.last_throughput <  minimal_snapshot_download_speed
+                                       && download_progress.notification_count <= 1
+                                       && download_progress.percentage_done <= 2_f32
+                                       && download_progress.estimated_remaining_time > 60_f32
+                                       && download_abort_count < maximum_snapshot_download_abort {
+                                        if let Some(ref trusted_validators) = validator_config.trusted_validators {
+                                            if trusted_validators.contains(&rpc_contact_info.id) && trusted_validators.len() == 1 {
+                                                return true; // Do not abort download from the one-and-only trusted validator
```
mvines, line 911: In this case, let's still warn about a slow download but also inform the user that we won't abort because there's no other options

validator/src/main.rs
```diff
@@ -891,7 +898,33 @@ fn rpc_bootstrap(
...
+                                &Some(|download_progress: &DownloadProgressRecord| {
+                                    debug!("Download progress: {:?}", download_progress);
+
+                                    if download_progress.last_throughput <  minimal_snapshot_download_speed
+                                       && download_progress.notification_count <= 1
+                                       && download_progress.percentage_done <= 2_f32
+                                       && download_progress.estimated_remaining_time > 60_f32
+                                       && download_abort_count < maximum_snapshot_download_abort {
+                                        if let Some(ref trusted_validators) = validator_config.trusted_validators {
+                                            if trusted_validators.contains(&rpc_contact_info.id) && trusted_validators.len() == 1 {
```
mvines, line 910: I think there's a missing `no_untrusted_rpc` check here.  It's ok to abort there's only a single trusted validator and the user hasn't restricted RPC to trusted validators only

mvines (dismissed):
lgtm, just some minor nits!

validator/src/main.rs
lijunwangs: Will do

validator/src/main.rs
lijunwangs: Good point. Will address

download-utils/src/lib.rs
lijunwangs: Let me see to that

validator/src/main.rs
lijunwangs: Fixed

validator/src/main.rs
lijunwangs: Fixed

sakridge (approved):
lgtm

download-utils/src/lib.rs
lijunwangs: Actually I attempted a type and I have run into all sort of issues with the type alias. Have not found a way to appease putting it into the Option, and passing around to the DownloadProgress struct. I am still struggling a bit with rust... I keep looking but I'd like to address that in a separate PR to reduce churns  with review and handling merge conflicts.

PR #17416, opened 2021-05-22
rework hash calculation to eliminate flatten, reduce data held in memory, for 1B accounts

jeffwashington:
#### Summary of Changes

Fixes #


jeffwashington:
reworked again in a separate pr

PR #17417, opened 2021-05-22
add --limit_load_slot_count_from_snapshot to ledger-tool

jeffwashington:
#### Problem
It takes a long time and uses a lot of memory to load snapshots with 360M accounts. For testing algorithms, it is nice to be able to use a smaller set of a large snapshot during testing cycles.

#### Summary of Changes
Add ledger-tool verify option to artificially limit how many slots are loaded from a snapshot.

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -5123,12 +5123,15 @@ impl AccountsDb {
     }
 
     #[allow(clippy::needless_collect)]
-    pub fn generate_index(&self) {
+    pub fn generate_index(&self, limit_load_slot_count_from_snapshot: Option<usize>) {
         type AccountsMap<'a> =
             HashMap<Pubkey, (StoredMetaWriteVersion, AppendVecId, StoredAccountMeta<'a>)>;
         let mut slots = self.storage.all_slots();
         #[allow(clippy::stable_sort_primitive)]
         slots.sort();
+        if let Some(limit) = limit_load_slot_count_from_snapshot {
+            slots.truncate(limit); // get rid of the newer slots and keep just the older
+        }
         let total_processed_slots_across_all_threads = AtomicU64::new(0);
         let outer_slots_len = slots.len();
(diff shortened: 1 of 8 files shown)


runtime/src/accounts_db.rs
```diff
@@ -4986,12 +4986,15 @@ impl AccountsDb {
     }
 
     #[allow(clippy::needless_collect)]
-    pub fn generate_index(&self) {
+    pub fn generate_index(&self, limit_load_slot_count_from_snapshot: Option<usize>) {
```
brooksprumo, line 4989: [Naming] From what I can see, the snapshot--as a concept--has already completed and isn't a part of this code (other than in a comment for chunk size). With that, I was thinking it could be better named to align with its use within this function.

My best suggestion is `limit_slots_used_to_generate_index`. If that feels too verbose, I would also be happy with naming the parameter `limit_slots` _and_ adding doc comments to the function that document this parameter.

Additionally, and thinking out loud, maybe this is the parameter name to use further up as well? Maybe the CLI parameter itself? I dunno about this, so happy to ignore that. I do acknowledge that keeping variable/parameter names the same all the way down is valuable, which is what prompted this follow-up thought.

brooksprumo (commented):
Similar to the other PR, does anyone else need to look at this to approve adding new arguments to the CLI?

EDIT: I saw your response on the other PR, so I'll assume that's the same for here as well.

brooksprumo:
@jeffwashington While looking through ledger-tool's `main.rs`, I saw it has parameters for "starting slot" and "ending slot". Maybe your "limit slots" could use "ending slot"? I.e. if someone sets an ending slot, you could limit the snapshot to that number as well?

I'm not sure all the pieces that "ending slot" interacts with, but figured I'd mention it in case it could be useful.

sakridge (approved):
seems reasonable to me

jeffwashington:
Thanks for the feedback, @brooksprumo. I think we can modify this cmd line arg in the future if needed. I like the # in terms of # of slots. I think we'll have to make more modifications to make working with 1B account snapshots later.

Issue #16992, opened 2021-05-02
Error when compiling from source on an M1 Macbook Air


RohanKapurDEV:
#### Problem
I am trying to compile this repo from source on an M1 MacBook air and I get the following message after i clone the repo, cd into it and try to run `cargo build`:
```
...
   Compiling pin-project-internal v1.0.1
   Compiling futures-macro v0.3.8
   Compiling tokio-macros v1.0.0
The following warnings were emitted during compilation:

warning: clang: error: the clang compiler does not support '-march=native'

error: failed to run custom build command for `reed-solomon-erasure v4.0.2`

Caused by:
  process didn't exit successfully: `/Users/rohan/Desktop/solana-bin/solana/target/debug/build/reed-solomon-erasure-f0dba0a3d69aae40/build-script-build` (exit code: 1)
  --- stdout
  TARGET = Some("aarch64-apple-darwin")
  HOST = Some("aarch64-apple-darwin")
  CC_aarch64-apple-darwin = None
  CC_aarch64_apple_darwin = None
  HOST_CC = None
  CC = None
  CFLAGS_aarch64-apple-darwin = None
  CFLAGS_aarch64_apple_darwin = None
  HOST_CFLAGS = None
  CFLAGS = None
  CRATE_CC_NO_DEFAULTS = None
  DEBUG = Some("true")
  CARGO_CFG_TARGET_FEATURE = None
  running: "cc" "-O3" "-ffunction-sections" "-fdata-sections" "-fPIC" "-g" "-fno-omit-frame-pointer" "-arch" "arm64" "-Wall" "-Wextra" "-march=native" "-std=c11" "-o" "/Users/rohan/Desktop/solana-bin/solana/target/debug/build/reed-solomon-erasure-d956d3b7629edcf3/out/simd_c/reedsolomon.o" "-c" "simd_c/reedsolomon.c"
  cargo:warning=clang: error: the clang compiler does not support '-march=native'
  exit code: 1

  --- stderr


[...]

PR #17440, opened 2021-05-24
Make sure cargo build succeeds on Apple M1 with Rosetta

felipecsl:
#### Problem

Even when building with Rosetta, Solana still fails to build on Apple M1 hardware.
I ran into these 2 blockers:

* First one was fixed with this workaround https://github.com/solana-labs/solana/issues/16992#issuecomment-838056642
* Second one, with rocksdb, was fixed by bumping it to version 0.16.0

#### Summary of Changes

* Bump rocksdb version to 0.16.0
* Remove feature `simd-accel` from `reed-solomon-erasure`

Fixes #

With this change,`cargo build` finishes successfully on Apple M1 hardware running with Rosetta

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -261,26 +261,21 @@ dependencies = [
 
 [[package]]
 name = "bindgen"
-version = "0.54.0"
+version = "0.57.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "66c0bb6167449588ff70803f4127f0684f9063097eca5016f37eb52b92c2cf36"
+checksum = "fd4865004a46a0aafb2a0a5eb19d3c9fc46ee5f063a6cfc605c69ac9ecf5263d"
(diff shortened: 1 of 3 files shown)


mvines:
We can't remove `simd-accel` generically like this, it's an essential performance feature.  We'd need a way turn it off when building on an M1

felipecsl:
hmm maybe we could have a platform bootstrap script that autogenerates the `ledger/Cargo.toml` file based on which platform you're running on?

mvines:
> hmm maybe we could have a platform bootstrap script that autogenerates the `ledger/Cargo.toml` file based on which architecture you're running on?

I think it'd be better to work upstream with the "reed-solomon-erasure" crate to add M1 support

felipecsl:
Seems like this is related to https://github.com/darrenldl/reed-solomon-erasure/issues/84

PR #17443, opened 2021-05-24
calculate_capitalization uses hash calculation

jeffwashington:
#### Problem
#17095

#### Summary of Changes
rework calculate_capitalization to use the store scan instead of index scan. Call update_accounts_hash_with_index_option with index=false.
Fixes #

### DIFF
--- a/runtime/src/accounts.rs
+++ b/runtime/src/accounts.rs
@@ -627,20 +627,24 @@ impl Accounts {
             .collect()
     }
 
-    pub fn calculate_capitalization(&self, ancestors: &Ancestors) -> u64 {
-        self.accounts_db.unchecked_scan_accounts(
-            "calculate_capitalization_scan_elapsed",
-            ancestors,
-            |total_capitalization: &mut u64, (_pubkey, loaded_account, _slot)| {
-                let lamports = loaded_account.lamports();
-                if Self::is_loadable(lamports) {
-                    *total_capitalization = AccountsDb::checked_iterative_sum_for_capitalization(
-                        *total_capitalization,
-                        lamports,
-                    );
-                }
-            },
-        )
+    pub fn calculate_capitalization(
+        &self,
+        ancestors: &Ancestors,
+        slot: Slot,
+        can_cached_slot_be_unflushed: bool,
+        debug_verify: bool,
+    ) -> u64 {
+        let use_index = false;
+        self.accounts_db
+            .update_accounts_hash_with_index_option(
+                use_index,
+                debug_verify,
+                slot,
+                ancestors,
+                None,
+                can_cached_slot_be_unflushed,
+            )
+            .1
     }
 
     #[must_use]
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -4451,11 +4451,11 @@ impl AccountsDb {
     }
 
     pub fn update_accounts_hash(&self, slot: Slot, ancestors: &Ancestors) -> (Hash, u64) {
-        self.update_accounts_hash_with_index_option(true, false, slot, ancestors, None)
+        self.update_accounts_hash_with_index_option(true, false, slot, ancestors, None, false)
     }
 
     pub fn update_accounts_hash_test(&self, slot: Slot, ancestors: &Ancestors) -> (Hash, u64) {
-        self.update_accounts_hash_with_index_option(true, true, slot, ancestors, None)
+        self.update_accounts_hash_with_index_option(true, true, slot, ancestors, None, false)
     }
 
     fn scan_multiple_account_storages_one_slot<F, B>(
@@ -4512,6 +4512,11 @@ impl AccountsDb {
 
     /// Scan through all the account storage in parallel
     fn scan_account_storage_no_bank<F, B>(
+        accounts_cache_and_ancestors: Option<(
+            &AccountsCache,
+            &Ancestors,
+            &AccountInfoAccountsIndex,
+        )>,
         snapshot_storages: &SortedStorages,
         scan_func: F,
     ) -> Vec<B>
@@ -4531,13 +4536,35 @@ impl AccountsDb {
                 let end = std::cmp::min(start + MAX_ITEMS_PER_CHUNK, snapshot_storages.range().end);
                 for slot in start..end {
                     let sub_storages = snapshot_storages.get(slot);
+                    let mut valid_slot = false;
                     if let Some(sub_storages) = sub_storages {
+                        valid_slot = true;
                         Self::scan_multiple_account_storages_one_slot(
                             sub_storages,
                             &scan_func,
                             slot,
                             &mut retval,
                         );
+                    }
+                    if let Some((cache, ancestors, accounts_index)) = accounts_cache_and_ancestors {
+                        if let Some(slot_cache) = cache.slot_cache(slot) {
+                            if valid_slot
+                                || ancestors.contains_key(&slot)
+                                || accounts_index.is_root(slot)
+                            {
+                                let keys = slot_cache.get_all_pubkeys();
+                                for key in keys {
+                                    if let Some(cached_account) = slot_cache.get_cloned(&key) {
+                                        let mut accessor = LoadedAccountAccessor::Cached(Some((
+                                            key,
+                                            Cow::Owned(cached_account),
+                                        )));
+                                        let account = accessor.get_loaded_account().unwrap();
+                                        scan_func(account, &mut retval, slot);
+                                    };
+                                }
+                            }
+                        }
                     }
                 }
                 retval
@@ -4551,14 +4578,26 @@ impl AccountsDb {
         slot: Slot,
         ancestors: &Ancestors,
         check_hash: bool,
+        can_cached_slot_be_unflushed: bool,
     ) -> Result<(Hash, u64), BankHashVerificationError> {
         if !use_index {
+            let accounts_cache_and_ancestors = if can_cached_slot_be_unflushed {
+                Some((&self.accounts_cache, ancestors, &self.accounts_index))
+            } else {
+                None
+            };
+
             let mut collect_time = Measure::start("collect");
             let (combined_maps, slots) = self.get_snapshot_storages(slot, Some(ancestors));
             collect_time.stop();
 
             let mut sort_time = Measure::start("sort_storages");
-            let storages = SortedStorages::new_with_slots(combined_maps.iter().zip(slots.iter()));
+            let min_root = self.accounts_index.min_root();
+            let storages = SortedStorages::new_with_slots(
+                combined_maps.iter().zip(slots.iter()),
+                min_root,
+                Some(slot),
+            );
             sort_time.stop();
 
             let timings = HashStats {
@@ -4572,6 +4611,7 @@ impl AccountsDb {
                 Some(&self.thread_pool_clean),
                 timings,
                 check_hash,
+                accounts_cache_and_ancestors,
             )
         } else {
             self.calculate_accounts_hash(slot, ancestors, check_hash)
@@ -4585,15 +4625,28 @@ impl AccountsDb {
         slot: Slot,
         ancestors: &Ancestors,
         expected_capitalization: Option<u64>,
+        can_cached_slot_be_unflushed: bool,
     ) -> (Hash, u64) {
         let check_hash = false;
         let (hash, total_lamports) = self
-            .calculate_accounts_hash_helper(use_index, slot, ancestors, check_hash)
+            .calculate_accounts_hash_helper(
+                use_index,
+                slot,
+                ancestors,
+                check_hash,
+                can_cached_slot_be_unflushed,
+            )
             .unwrap(); // unwrap here will never fail since check_hash = false
         if debug_verify {
             // calculate the other way (store or non-store) and verify results match.
             let (hash_other, total_lamports_other) = self
-                .calculate_accounts_hash_helper(!use_index, slot, ancestors, check_hash)
+                .calculate_accounts_hash_helper(
+                    !use_index,
+                    slot,
+                    ancestors,
+                    check_hash,
+                    can_cached_slot_be_unflushed,
+                )
                 .unwrap(); // unwrap here will never fail since check_hash = false
 
             let success = hash == hash_other
@@ -4607,12 +4660,17 @@ impl AccountsDb {
         (hash, total_lamports)
     }
 
-    fn scan_snapshot_stores(
+    fn scan_snapshot_stores_with_cache(
         storage: &SortedStorages,
         mut stats: &mut crate::accounts_hash::HashStats,
         bins: usize,
         bin_range: &Range<usize>,
         check_hash: bool,
+        accounts_cache_and_ancestors: Option<(
+            &AccountsCache,
+            &Ancestors,
+            &AccountInfoAccountsIndex,
+        )>,
     ) -> Result<Vec<Vec<Vec<CalculateHashIntermediate>>>, BankHashVerificationError> {
         let bin_calculator = PubkeyBinCalculator16::new(bins);
         assert!(bin_range.start < bins && bin_range.end <= bins && bin_range.start < bin_range.end);
@@ -4621,6 +4679,7 @@ impl AccountsDb {
         let mismatch_found = AtomicU64::new(0);
 
         let result: Vec<Vec<Vec<CalculateHashIntermediate>>> = Self::scan_account_storage_no_bank(
+            accounts_cache_and_ancestors,
             storage,
             |loaded_account: LoadedAccount,
              accum: &mut Vec<Vec<CalculateHashIntermediate>>,
@@ -4685,6 +4744,11 @@ impl AccountsDb {
         thread_pool: Option<&ThreadPool>,
         mut stats: HashStats,
         check_hash: bool,
+        accounts_cache_and_ancestors: Option<(
+            &AccountsCache,
+            &Ancestors,
+            &AccountInfoAccountsIndex,
+        )>,
     ) -> Result<(Hash, u64), BankHashVerificationError> {
         let mut scan_and_hash = move || {
             // When calculating hashes, it is helpful to break the pubkeys found into bins based on the pubkey value.
@@ -4711,12 +4775,13 @@ impl AccountsDb {
                     end: (pass + 1) * bins_per_pass,
                 };
 
-                let result = Self::scan_snapshot_stores(
+                let result = Self::scan_snapshot_stores_with_cache(
                     &storages,
                     &mut stats,
                     PUBKEY_BINS_FOR_CALCULATING_HASHES,
                     &bounds,
                     check_hash,
+                    accounts_cache_and_ancestors,
                 )?;
 
                 let (hash, lamports, for_next_pass) = AccountsHash::rest_of_hash_calculation(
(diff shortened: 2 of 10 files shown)


runtime/src/accounts_db.rs
```diff
@@ -4963,13 +4964,25 @@ impl AccountsDb {
         }
     }
 
-    pub fn get_snapshot_storages(&self, snapshot_slot: Slot) -> SnapshotStorages {
+    pub fn get_snapshot_storages(
+        &self,
+        snapshot_slot: Slot,
+        ancestors: Option<&Ancestors>,
```
jeffwashington, line 4970: @sakridge, it looks like we have to know ancestors here for getting storages, at least for tests. Stores are in slots that aren't in the roots set. I haven't thought this all the way through yet, but I figured I'd get feedback since it just passed all tests for the first time. This is related to the startup time issue of checking cap.

sakridge: Yes. This seems right. It does the capitalization check on the newest bank after replaying all of what is in blockstore which may or may not be rooted.

jeffwashington:
@sakridge replacing an index scan with a storage scan will not work if there is account data in the write cache.
The nominal test fails in [stable](https://buildkite.com/solana-labs/solana/builds/47104#68f7bae8-c970-4882-86ff-06e7559d92b6).

Here's where we get where we have cached accounts loaded:
https://github.com/solana-labs/solana/blob/9a5330b7eb617251ad1843b2aa5bfa4f1a8b3f7f/runtime/src/accounts.rs#L623-L628

Here's the call that gets us to calculate_capitalization.
https://github.com/solana-labs/solana/blob/179856c13afcfcde6dadd8c829b7155d3337f951/ledger/src/blockstore_processor.rs#L558
```
   1: solana_runtime::accounts::Accounts::calculate_capitalization
             at /Users/jeffreywashington/dev/s4/runtime/src/accounts.rs:617:9
   2: solana_runtime::bank::Bank::calculate_capitalization
             at /Users/jeffreywashington/dev/s4/runtime/src/bank.rs:4466:9
   3: solana_runtime::bank::Bank::calculate_and_verify_capitalization
             at /Users/jeffreywashington/dev/s4/runtime/src/bank.rs:4472:26
   4: solana_ledger::blockstore_processor::do_process_blockstore_from_root
             at /Users/jeffreywashington/dev/s4/ledger/src/blockstore_processor.rs:557:9
   5: solana_ledger::blockstore_processor::process_blockstore
             at /Users/jeffreywashington/dev/s4/ledger/src/blockstore_processor.rs:410:5
   ```
   
We would presumably need to flush the cache before this call, or we detect there are items in the cache and we fallback to the account scan as before. Or, we do something more complicated.

If we're just worried about the cap scan at startup, there will not be anything in the cache. We could just change that code path.

**Comment by jeffwashington** [comment]
This code path now supports appending the write cache data per slot to the store data per slot.

File: runtime/src/accounts_db.rs [hunk 2]
```diff
@@ -4357,13 +4358,29 @@ impl AccountsDb {
...
                     if let Some(sub_storages) = sub_storages {
                         for storage in sub_storages {
-                            let slot = storage.slot();
                             let accounts = storage.accounts.accounts(0);
                             accounts.into_iter().for_each(|stored_account| {
                                 scan_func(LoadedAccount::Stored(stored_account), &mut retval, slot)
                             });
                         }
                     }
+                    if let Some(cache) = accounts_cache {
```
jeffwashington, line 4367: @carllin - this is interesting to you. We are using the read cache to augment the data in stores in order to calculate capitalization on banks where we haven't flushed the cache yet.

runtime/src/accounts_db.rs
```diff
@@ -4364,6 +4365,23 @@ impl AccountsDb {
...
+                        if let Some(slot_cache) = cache.slot_cache(slot) {
+                            let keys = slot_cache.get_all_pubkeys();
```
brooksprumo, line 4370: Would it be beneficial/faster to have `get_all_pubkeys()` return `Vec<&Pubkey>` instead? It looks like sometimes the loop below might not need the copy. (And then it'd be simple to just `*key` in the `let mut accessor = ...` call.)

jeffwashington: get_all_pubkeys reads keys out of a dashmap. The key values are behind a lock structure. I do not imagine it is straightforward or even possible to get a reference without holding a lock open. There are other ways to do this as well, adding an iterate function perhaps to the SlotCache. This one seemed safe and straightforward, but I'm open to alternative suggestions.

sakridge: It's not possible. To hold a reference to the keys you would have to hold the lock open.

runtime/src/accounts.rs
```diff
@@ -619,8 +619,13 @@ impl Accounts {
...
     }
 
-    pub fn calculate_capitalization(&self, ancestors: &Ancestors) -> u64 {
-        self.accounts_db.unchecked_scan_accounts(
+    pub fn calculate_capitalization(&self, ancestors: &Ancestors, slot: Slot) -> u64 {
+        let cap = self
+            .accounts_db
+            .update_accounts_hash_with_index_option(false, false, slot, ancestors, None, true)
+            .1;
+        // debug code for the moment
```
jeffwashington, line 627: this old code path remains for now to make sure we are getting the right results in all cases. I'll run this against mnb for a while. I am not yet familiar with all the ways calculate_capitalization is called. I'll look into it.

jeffwashington: so, to be clear I will not complete this pr with the old code path and the assert present.

runtime/src/accounts.rs
```diff
@@ -619,8 +619,13 @@ impl Accounts {
             .collect()
     }
 
-    pub fn calculate_capitalization(&self, ancestors: &Ancestors) -> u64 {
-        self.accounts_db.unchecked_scan_accounts(
+    pub fn calculate_capitalization(&self, ancestors: &Ancestors, slot: Slot) -> u64 {
+        let cap = self
+            .accounts_db
+            .update_accounts_hash_with_index_option(false, false, slot, ancestors, None, true)
+            .1;
```
jeffwashington, line 626: we could also pass a parameter here to prevent actually calculating a hash value since we throw it away.

jeffwashington: Also, I need to change this to call a level deeper and not update the hash - but just calculate the hash to maintain the behavior of the function I'm attempting to replace.

runtime/src/accounts_db.rs
brooksprumo: Gotcha. Thanks for the info!

runtime/src/accounts_db.rs
```diff
@@ -4410,15 +4430,27 @@ impl AccountsDb {
         slot: Slot,
         ancestors: &Ancestors,
         expected_capitalization: Option<u64>,
+        use_write_cache: bool,
```
carllin, line 4433: nit: how about renaming `use_write_cache` -> `can_cached_slot_be_unflushed`, to signal this boolean is not an optimization as much as a correctness guarantee

runtime/src/accounts_db.rs
```diff
@@ -4410,15 +4430,27 @@ impl AccountsDb {
...
     ) -> (Hash, u64) {
         let check_hash = false;
+        let accounts_cache = if use_write_cache {
+            Some(&self.accounts_cache)
+        } else {
+            None
+        };
```
carllin, line 4440: Can we move this check into `calculate_accounts_hash_helper` which still has a reference to `self`?

Then `calculate_accounts_hash_helper` can also take the `use_write_cache/can_cached_slot_be_unflushed` boolean

runtime/src/accounts.rs
```diff
@@ -619,8 +619,13 @@ impl Accounts {
             .collect()
     }
 
-    pub fn calculate_capitalization(&self, ancestors: &Ancestors) -> u64 {
-        self.accounts_db.unchecked_scan_accounts(
+    pub fn calculate_capitalization(&self, ancestors: &Ancestors, slot: Slot) -> u64 {
```
carllin, line 622: should we also pass the `use_write_cache/can_cached_slot_be_unflushed` boolean here so callers have to be cognizant of whether they have left slots unflushed in the cache?

runtime/src/accounts_db.rs
carllin: Ah I see, this is because on startup when we run `if !bank_forks.root_bank().calculate_and_verify_capitalization() {` from `do_process_blockstore_from_root()` the cache may not have been flushed yet. This might be a good comment to add to the `calculate_capitalization()` function for why we pass the `use_cache` flag. If we do as this comment  suggests https://github.com/solana-labs/solana/pull/17443/files#r645215712, then that might also be good documentation for that function

runtime/src/sorted_storages.rs
```diff
@@ -45,21 +45,25 @@ impl<'a> SortedStorages<'a> {
...
 
     // source[i] is in slot slots[i]
     // assumptions:
     // 1. slots vector contains unique slot #s.
     // 2. slots and source are the same len
-    pub fn new_with_slots(source: &'a [SnapshotStorage], slots: &[Slot]) -> Self {
+    pub fn new_with_slots(
+        source: &'a [SnapshotStorage],
+        slots: &[Slot],
+        slot: Option<Slot>,
```
carllin, line 58: Could we add some documentation for what this argument is and what `Some()` vs `None` means for this particular argument?

If this `slot` the maximum slot in the `slots`slice, does it seem appropriate to rename `slot` to `max_slot`?

runtime/src/sorted_storages.rs
```diff
@@ -45,21 +45,25 @@ impl<'a> SortedStorages<'a> {
...
-        Self::new_with_slots(source, &slots)
+        Self::new_with_slots(source, &slots, None)
     }
 
     // source[i] is in slot slots[i]
     // assumptions:
     // 1. slots vector contains unique slot #s.
     // 2. slots and source are the same len
-    pub fn new_with_slots(source: &'a [SnapshotStorage], slots: &[Slot]) -> Self {
+    pub fn new_with_slots(
```
carllin, line 55: nit: maybe rename to `new_from_storage_entries`?

runtime/src/sorted_storages.rs
```diff
@@ -45,21 +45,25 @@ impl<'a> SortedStorages<'a> {
...
+        source: &'a [SnapshotStorage],
+        slots: &[Slot],
```
carllin, line 57: Would it be cleaner to pass an `impl Iter<Item=(SnapshotStorage, Slot)>` here, i.e. pass in a zipped iterator?

runtime/src/accounts_db.rs
```diff
@@ -4364,6 +4365,23 @@ impl AccountsDb {
                             });
                         }
                     }
+                    if let Some(cache) = accounts_cache {
+                        if let Some(slot_cache) = cache.slot_cache(slot) {
```
carllin, line 4369: Can there be slots in the cache earlier than the `let start = snapshot_storages.range().start` we start checking from? https://github.com/solana-labs/solana/pull/17443/files#diff-1090394420d51617f3233275c2b65ed706b35b53b115fe65f82c682af8134a6fR4356

runtime/src/sorted_storages.rs
jeffwashington: some callers have only &[SnapshotStorage]. We have to figure out the corresponding slots. Restoring from a snapshot is such a code path. Those callers call 'new'.
Other callers can easily generate a corresponding Slot per storage. Those callers call 'new_with_slots'.
We can rename either of them. This is the brief history of the naming.

runtime/src/sorted_storages.rs
jeffwashington: Well, first, we iterate the slots twice to find min and max. I have considered calculating those when we generate the slots vector upstream as well to simplify this code.
Second, the data itself is currently created as 2 corresponding vectors to prevent other callers who need just a vector of SnapshotStorage to not have to split out the slot values they don't need.
I think it makes things more complicated somewhere.

runtime/src/accounts_db.rs
jeffwashington: Yes, it could break. This gets to another aspect of a 'min' slot concept I talked to you the other day about. An assumption now is that we have a storage for the 'min' slot we care about - it is either a root or now possibly an ancestor. For this pr's new usage of scanning stores, we also have to introduce the concept of a 'max_inclusive' slot to make sure we check the write cache for all slots we know about up to and including 'max_inclusive' because the slot range defined by the storages is not sufficient for all the slots we care about - specifically on the max side. The min side is the same issue. I guess if we used min(root, ancestor), we could bracket the other end. This specifically then gets to the topic of roots not knowing the min root in its current implementation (because of the potential for excess). We can fix this - there just wasn't a use case yet. Maybe this pr is a use case for the min. Basically, to scan stores and the cache, I need to know every slot in the range because there could be something in the write cache for that.  Well, actually, this is wrong as it is because I don't want EVERYTHING in the write cache - I only want slots that are a root or an ancestor. So, already this code is wrong... Ugh. The value of prs...  I imagined there was at least 1 thing this code wasn't solving yet...

runtime/src/accounts_db.rs
jeffwashington: I don't understand what you linked to. Do you mean `can_cached_slot_be_unflushed`?

runtime/src/accounts_db.rs
jeffwashington: ok, I pushed a commit that pairs accounts cache with ancestors so we won't be adding account data for slots which were not our ancestors. I am assuming that a slot that is a root does not have data in the write cache? Is that true? @carllin

runtime/src/accounts_db.rs
carllin: arg! yeah sorry the links to the comments failed, good thing your intuition is good 😃. Yeah I meant this comment: https://github.com/solana-labs/solana/pull/17443#discussion_r645212955 regarding `can_cached_slot_be_unflushed`.

runtime/src/sorted_storages.rs
carllin: Yeah, so `get_snapshot_storages()`outputs 2 vectors, but we can zip those in `calculate_accounts_hash_helper()` before passing them here, so that shouldn't affect other callers of `get_snapshot_storages`. This seems to work: https://github.com/carllin/solana/commit/bb9ac0849a93f22e5aa526316b4a1fb8beb4a4d3

runtime/src/accounts_db.rs
carllin: > I am assuming that a slot that is a root does not have data in the write cache? Is that true?

Unfortunately this isn't true, the write cache can store many roots between flushes. On startup processing the cache is flushed every 10 seconds here https://github.com/solana-labs/solana/blob/bea7ce717c7bf68625cb9d5218d4460982cb3ef6/ledger/src/blockstore_processor.rs#L1025-L1031, which means there may be many roots present in the cache since the last flush

carllin: Yeah it seems like the safest thing to do would either: 

1) Force flush of the cache before computing the capitalization to avoid this issue altogether
2) Start iteration from the smallest item in the cache to the first stored slot. Because at startup the cache only keeps the latest 10 seconds of roots around, the *earliest* root in the cache is  probably much greater than the *earliest* root tracked in the `RootsTracker`. Maybe we could track all the slots in the cache with something similar to a small `BTreeSet` and just iterate over that set to find any cache items less than the first stored slot?

runtime/src/sorted_storages.rs
jeffwashington: ok. I cherry-picked your change.

runtime/src/accounts_db.rs
jeffwashington: @carllin , please take another look. I did #2. I put in a few separate prs to take care of some simple code movement.

runtime/src/sorted_storages.rs
```diff
@@ -54,9 +54,25 @@ impl<'a> SortedStorages<'a> {
...
+        // none, either, or both of min/max could be specified
+        if let Some(slot) = min_slot {
+            min = std::cmp::min(slot, min);
+            max = std::cmp::max(slot + 1, max);
+        }
+        if let Some(slot) = max_slot_inclusive {
+            min = std::cmp::min(slot, min);
+            max = std::cmp::max(slot + 1, max);
+        }
+
```
carllin, line 75: if we assert here:

```
if let Some(max_slot) = max_slot_inclusive {
        assert!(max_slot >= min_slot)
}
```

Then written this way, this block is easier for me to reason about all the cases:
```
let (mut min, mut max) = match (min_slot, max_slot_inclusive) {
            (Some(min_slot), Some(max_slot_inclusive)) => (min_slot, max_slot_inclusive),
            (Some(min_slot), None) => (min_slot, Slot::MAX),
            (None, Some(max_slot_inclusive)) => (Slot::MIN, max_slot_inclusive + 1),
            (None, None) => (Slot::MIN, Slot::MAX),
        };
```

runtime/src/sorted_storages.rs
```diff
@@ -54,9 +54,25 @@ impl<'a> SortedStorages<'a> {
     // 2. slots and source are the same len
     pub fn new_with_slots<'b>(
         source: impl Iterator<Item = (&'a SnapshotStorage, &'b Slot)> + Clone,
+        min_slot: Option<Slot>,
+        // highest valid slot. Only matters if slots array does not contain a slot >= max_slot_inclusive.
+        // An example is a slot that has accounts in the write cache at slots <= 'max_slot_inclusive' but no storages at those slots.
+        // None => self.range.end = slots.max() + 1
+        // Some(slot) => self.range.end = std::cmp::max(slot, slots.max())
```
carllin, line 61: nit: Rename both references to `slots` here to  `source`

runtime/src/sorted_storages.rs
```diff
@@ -54,9 +54,25 @@ impl<'a> SortedStorages<'a> {
     // 2. slots and source are the same len
     pub fn new_with_slots<'b>(
         source: impl Iterator<Item = (&'a SnapshotStorage, &'b Slot)> + Clone,
+        min_slot: Option<Slot>,
```
carllin, line 57: Nit: Add small comment `A slot used as a lower bound, but potentially smaller than the smallest slot in the given 'source' iterator`

runtime/src/accounts_db.rs
```diff
@@ -4430,14 +4457,26 @@ impl AccountsDb {
...
             let mut collect_time = Measure::start("collect");
             let (combined_maps, slots) = self.get_snapshot_storages(slot, Some(ancestors));
             collect_time.stop();
 
             let mut sort_time = Measure::start("sort_storages");
-            let storages = SortedStorages::new_with_slots(combined_maps.iter().zip(slots.iter()));
+            let min_root = self.accounts_index.min_root();
+            let storages = SortedStorages::new_with_slots(
+                combined_maps.iter().zip(slots.iter()),
+                min_root,
```
carllin, line 4477: Not a big deal, but one thing to note is there can potentially be a large gap between the min root and the first slot in the cache (we keep around 425,000 roots, but we only keep around the latest 10 seconds of roots at startup in the cache). If we ever need to repeatedly call this function, might be more efficient to either:

1) Have a separate min for the cache
2) Cache the starting point for the cache search after performing it the first time.

carllin (approved):
A few nits, but otherwise lgtm!

runtime/src/sorted_storages.rs
jeffwashington: I'd prefer not to add an assert. Also, it seems possible that calculations of min could end up > max given that we have several sources we're merging: global roots, ancestors, current slot of bank?, and the global write cache. We are really looking for the max range that we have to iterate over to check stores or write cache for accounts that we should include in our calculation. 

I did rework the function to make it much more clear what is going on. min_slot and max_slot_inclusive are just 2 more 'slot' values run through the min/max calculation, along with all the slots in the iter.

PR #17444, opened 2021-05-24
result from hash calculation

jeffwashington:
#### Problem
Prepare to replace index scan at startup with store (ie. non-index) scan. Index scan has 'check_hash' option to verify the hash value of each account.
#### Summary of Changes
add 'check_hash' to non-index scan.
Fixes #


jeffwashington:
replaced by #17558

PR #17446, opened 2021-05-24
ancestors.large_range_slots becomes HashSet

jeffwashington:
#### Problem
migrating ancestors to be a hashset instead of hashmap. API is already changed.
#### Summary of Changes
move Ancestors.large_range_slots to be HashSet
Fixes #


sakridge (approved):
lgtm

PR #17449, opened 2021-05-24
feat: add getBlock and getTransaction apis

jstarry:
#### Problem
The confirmed api's obscure the original account key ordering of confirmed transactions which makes metadata lookup unreliable.

#### Summary of Changes
- Added `Connection.getBlock` and `Connection.getTransaction` to match the new RPC API's but they still call the confirmed endpoints underneath.

Fixes #


CriesofCarrots (approved):
Cool, thx

PR #17471, opened 2021-05-25
Handle removing slots during account scans

carllin:
#### Problem
Built on top of https://github.com/solana-labs/solana/pull/17269, addressing the TODO: here:https://github.com/solana-labs/solana/blob/3d0cacc1a617b4344ea44afdb0b51e5627bd2761/runtime/src/accounts_db.rs#L3411

While a scan on some fork at bank `B` is happening, the slots on that fork may be removed by `remove_unrooted_slot()` if:
1) There was a duplicate slot on that fork
2) We detect another confirmed version of that slot

This means the scan results will be potentially inconsistent/missing accounts

#### Summary of Changes
If we detect a fork has been dumped during a scan, the scan result will be aborted. Important changes are summarized in comments below starting with https://github.com/solana-labs/solana/pull/17471#discussion_r645938525

Fixes #

### DIFF
--- a/client/src/rpc_custom_error.rs
+++ b/client/src/rpc_custom_error.rs
@@ -1,5 +1,7 @@
 //! Implementation defined RPC server errors
 
+use solana_runtime::accounts_index::ScanError;
+use thiserror::Error;
 use {
     crate::rpc_response::RpcSimulateTransactionResult,
     jsonrpc_core::{Error, ErrorCode},
@@ -17,35 +19,40 @@ pub const JSON_RPC_SERVER_ERROR_NO_SNAPSHOT: i64 = -32008;
 pub const JSON_RPC_SERVER_ERROR_LONG_TERM_STORAGE_SLOT_SKIPPED: i64 = -32009;
 pub const JSON_RPC_SERVER_ERROR_KEY_EXCLUDED_FROM_SECONDARY_INDEX: i64 = -32010;
 pub const JSON_RPC_SERVER_ERROR_TRANSACTION_HISTORY_NOT_AVAILABLE: i64 = -32011;
+pub const JSON_RPC_REMOVED_SLOT: i64 = -32012;
 
+#[derive(Error, Debug)]
 pub enum RpcCustomError {
+    #[error("BlockCleanedUp")]
     BlockCleanedUp {
         slot: Slot,
         first_available_block: Slot,
     },
+    #[error("SendTransactionPreflightFailure")]
     SendTransactionPreflightFailure {
         message: String,
         result: RpcSimulateTransactionResult,
     },
+    #[error("TransactionSignatureVerificationFailure")]
     TransactionSignatureVerificationFailure,
-    BlockNotAvailable {
-        slot: Slot,
-    },
-    NodeUnhealthy {
-        num_slots_behind: Option<Slot>,
-    },
+    #[error("BlockNotAvailable")]
+    BlockNotAvailable { slot: Slot },
+    #[error("NodeUnhealthy")]
+    NodeUnhealthy { num_slots_behind: Option<Slot> },
+    #[error("TransactionPrecompileVerificationFailure")]
     TransactionPrecompileVerificationFailure(solana_sdk::transaction::TransactionError),
-    SlotSkipped {
-        slot: Slot,
-    },
+    #[error("SlotSkipped")]
+    SlotSkipped { slot: Slot },
+    #[error("NoSnapshot")]
     NoSnapshot,
-    LongTermStorageSlotSkipped {
-        slot: Slot,
-    },
-    KeyExcludedFromSecondaryIndex {
-        index_key: String,
-    },
+    #[error("LongTermStorageSlotSkipped")]
+    LongTermStorageSlotSkipped { slot: Slot },
+    #[error("KeyExcludedFromSecondaryIndex")]
+    KeyExcludedFromSecondaryIndex { index_key: String },
+    #[error("TransactionHistoryNotAvailable")]
     TransactionHistoryNotAvailable,
+    #[error("ScanError")]
+    ScanError(#[from] ScanError),
 }
 
 #[derive(Debug, Serialize, Deserialize)]
@@ -139,6 +146,11 @@ impl From<RpcCustomError> for Error {
                     JSON_RPC_SERVER_ERROR_TRANSACTION_HISTORY_NOT_AVAILABLE,
                 ),
                 message: "Transaction history is not available from this node".to_string(),
+                data: None,
+            },
+            RpcCustomError::ScanError(scan_err) => Self {
+                code: ErrorCode::ServerError(JSON_RPC_REMOVED_SLOT),
+                message: scan_err.to_string(),
                 data: None,
             },
         }
--- a/client/src/rpc_custom_error.rs
+++ b/client/src/rpc_custom_error.rs
@@ -1,6 +1,4 @@
 //! Implementation defined RPC server errors
-
-use solana_runtime::accounts_index::ScanError;
 use thiserror::Error;
 use {
     crate::rpc_response::RpcSimulateTransactionResult,
@@ -52,7 +50,7 @@ pub enum RpcCustomError {
     #[error("TransactionHistoryNotAvailable")]
     TransactionHistoryNotAvailable,
     #[error("ScanError")]
-    ScanError(#[from] ScanError),
+    ScanError { message: String },
 }
 
 #[derive(Debug, Serialize, Deserialize)]
@@ -148,9 +146,9 @@ impl From<RpcCustomError> for Error {
                 message: "Transaction history is not available from this node".to_string(),
                 data: None,
             },
-            RpcCustomError::ScanError(scan_err) => Self {
+            RpcCustomError::ScanError { message } => Self {
                 code: ErrorCode::ServerError(JSON_RPC_SCAN_ERROR),
-                message: scan_err.to_string(),
+                message,
                 data: None,
             },
         }
--- a/client/src/rpc_custom_error.rs
+++ b/client/src/rpc_custom_error.rs
@@ -19,7 +19,7 @@ pub const JSON_RPC_SERVER_ERROR_NO_SNAPSHOT: i64 = -32008;
 pub const JSON_RPC_SERVER_ERROR_LONG_TERM_STORAGE_SLOT_SKIPPED: i64 = -32009;
 pub const JSON_RPC_SERVER_ERROR_KEY_EXCLUDED_FROM_SECONDARY_INDEX: i64 = -32010;
 pub const JSON_RPC_SERVER_ERROR_TRANSACTION_HISTORY_NOT_AVAILABLE: i64 = -32011;
-pub const JSON_RPC_REMOVED_SLOT: i64 = -32012;
+pub const JSON_RPC_SCAN_ERROR: i64 = -32012;
 
 #[derive(Error, Debug)]
 pub enum RpcCustomError {
@@ -149,7 +149,7 @@ impl From<RpcCustomError> for Error {
                 data: None,
             },
             RpcCustomError::ScanError(scan_err) => Self {
-                code: ErrorCode::ServerError(JSON_RPC_REMOVED_SLOT),
+                code: ErrorCode::ServerError(JSON_RPC_SCAN_ERROR),
                 message: scan_err.to_string(),
                 data: None,
             },
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -24,7 +24,7 @@ use crate::{
     accounts_hash::{AccountsHash, CalculateHashIntermediate, HashStats, PreviousPass},
     accounts_index::{
         AccountIndexGetResult, AccountSecondaryIndexes, AccountsIndex, AccountsIndexRootsStats,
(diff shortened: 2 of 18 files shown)


client/src/rpc_custom_error.rs
```diff
@@ -16,34 +18,38 @@ pub const JSON_RPC_SERVER_ERROR_SLOT_SKIPPED: i64 = -32007;
...
-    },
-    KeyExcludedFromSecondaryIndex {
-        index_key: String,
-    },
+    #[error("LongTermStorageSlotSkipped")]
+    LongTermStorageSlotSkipped { slot: Slot },
+    #[error("KeyExcludedFromSecondaryIndex")]
+    KeyExcludedFromSecondaryIndex { index_key: String },
+    #[error("ScanError")]
+    ScanError(#[from] ScanError),
```
carllin, line 52: @CriesofCarrots Would like our opinion on this proposed way of propagating the error up to the calling client

runtime/src/accounts_index.rs
```diff
@@ -736,6 +775,28 @@ impl<T: 'static + Clone + IsCached + ZeroLamport> AccountsIndex<T> {
...
+                w_tip_of_scanned_forks.remove(&scan_slot_id);
+            }
+            is_scan_aborted
+        };
+
+        if is_scan_aborted {
+            Err(ScanError::SlotRemoved {
+                slot: ancestors.max(),
+                slot_id: scan_slot_id,
+            })
```
carllin, line 796: @CriesofCarrots this is where the new error is first created

runtime/src/non_circulating_supply.rs
```diff
@@ -41,7 +41,10 @@ pub fn calculate_non_circulating_supply(bank: &Arc<Bank>) -> NonCirculatingSuppl
         )
     } else {
         bank.get_program_accounts(&solana_stake_program::id())
-    };
+    }
+    // TODO: Figure out how to properly handle errors here
+    .unwrap_or_default();
```
carllin, line 46: @CriesofCarrots not 100% sure the best way to handle this yet, specifically where this method is called from `process_rest()` in `impl RequestMiddleware for RpcRequestMiddleware`, is there a best way to plumb up an error?

carllin: Oh looks like that path in `process_rest()`  always uses a `root_bank`, and only unrooted forks can be removed, so maybe it's safe to just unwrap there

CriesofCarrots: Yeah, looks right to me.

client/src/rpc_custom_error.rs
CriesofCarrots: I guess the question is, do these details need to be exposed to the calling client? Or would a generic `internal_error` be sufficient (along with logging the details on the server side)?

runtime/src/accounts_db.rs
```diff
@@ -3507,13 +3533,26 @@ impl AccountsDb {
             }
         }
 
+        // Mark down these slots are about to be purged so that new attempts to scan these
+        // banks fail, and any ongoing scans over these slots will detect that they should abort
+        // their results
+        {
+            let mut locked_removed_slot_ids = self.accounts_index.removed_slot_ids.lock().unwrap();
```
carllin, line 3540: When we remove the state for a particular bank for slot `S`, we mark down the unique bank id of this particular `Bank` so that any ongoing scan over this bank will know their state may have been corrupted

runtime/src/accounts_index.rs
```diff
@@ -570,6 +597,17 @@ pub struct AccountsIndex<T> {
...
+    // is iterating over. The unique id of that slot `S` is recorded here (note we don't use
+    // `S` as the id because there can be more than one version of a slot `S`). If a fork
+    // is abandoned, all of the slots on that fork up to `S` will be removed via
+    // `AccountsDb::remove_unrooted_slots()`. When the scan finishes, it'll realize that the
+    // results of the scan may have been corrupted by `remove_unrooted_slots` and abort its results.
+    //
+    // `removed_slot_ids` tracks all the slot ids that were removed via `remove_unrooted_slots()` so any attempted scans
+    // on any of these slots fails. This is safe to purge once the associated Bank is dropped and
+    // scanning the fork with that Bank at the tip is no longer possible.
+    pub removed_slot_ids: Mutex<HashSet<SlotId>>,
```
carllin, line 610: Added tracker here of bank id's that have been removed by `remove_unrooted_slots`. Ongoing scans check against this when they're finished to see if the results have been corrupted

runtime/src/bank.rs
```diff
@@ -12131,6 +12226,242 @@ pub(crate) mod tests {
...
+
+                        // Remove the rest of the slots before the next iteration
+                        for (slot, slot_id) in slots_on_fork {
+                            bank_at_fork_tip.remove_unrooted_slots(&[(slot, slot_id)]);
+                        }
+                    }
+                },
+                None,
+                // Test removing the slot before the scan starts, should error every time
+                AcceptableScanResults::Both,
```
carllin, line 12464: These 3 tests show that safety of scan/remove_unrooted slots is upheld in 3 different interleavings of events:

1. Remove the slot before scan of the slot starts
2. Remove one version of the slot, recreate a second version of the slot, start a scan on the first version of the slot
2. Removing slot in the middle of an ongoing scan

runtime/src/bank.rs
```diff
@@ -5241,18 +5266,16 @@ impl Bank {
 
 impl Drop for Bank {
     fn drop(&mut self) {
-        if self.skip_drop.load(Relaxed) {
-            return;
-        }
-
```
carllin, line 5247: We remove this because now even after a call to `remove_unrooted_slot()` on a bank, we still want to run `purge_slot` to remove the entry for the bank from the newly introduced `AccountsIndex::removed_slot_ids`

runtime/src/accounts_db.rs
```diff
@@ -3213,14 +3239,20 @@ impl AccountsDb {
             // This function is only currently safe with respect to `flush_slot_cache()` because
             // both functions run serially in AccountsBackgroundService.
             let mut remove_cache_elapsed = Measure::start("remove_cache_elapsed");
-            if let Some(slot_cache) = self.accounts_cache.remove_slot(*remove_slot) {
+            // Note: we cannot remove this slot from the slot cache until we've removed its
+            // entries from the accounts index first. This is because `scan_accounts()` relies on
+            // holding the index lock, finding the index entry, and then looking up the entry
+            // in the cache. If it fails to find that entry, it will panic in `get_loaded_account()`
+            if let Some(slot_cache) = self.accounts_cache.slot_cache(*remove_slot) {
```
carllin, line 3246: Note the issue described here was not an issue before because scan would never run concurrently with `purge_slots_from_cache_and_store()` because `purge_slots_from_cache_and_store` only ran via `Bank::drop()`, and a scan on a bank means it had a reference to said bank, and thus the `drop()` must not have run yet.

 Now, however, because `remove_unrooted_slots()` can call into `purge_slots_from_cache_and_store()`,  which can happen while a scan is running, we need to account for this  case.

runtime/src/bank.rs
```diff
@@ -416,6 +416,8 @@ pub struct BankRc {
 
     /// Current slot
     pub(crate) slot: Slot,
+
+    pub(crate) slot_id_generator: Arc<AtomicU64>,
```
carllin, line 420: Used to distinguish between banks with the same slot, see description here: https://github.com/solana-labs/solana/pull/17471/files#r645937617

client/src/rpc_custom_error.rs
carllin: @CriesofCarrots  yeah I didn't see any downside in exposing the exact details to the client, it's just another error anyways. This is a pretty exceptional case, so if a client complains about this condition, we know what the details are exactly without having to dig through logs.

runtime/src/bank.rs
```diff
@@ -416,6 +416,8 @@ pub struct BankRc {
 
     /// Current slot
     pub(crate) slot: Slot,
+
+    pub(crate) bank_id_generator: Arc<AtomicU64>,
```
carllin, line 420: Used to distinguish between banks with the same slot, see description here: https://github.com/solana-labs/solana/pull/17471/files#r645937617

runtime/src/bank.rs
```diff
@@ -12131,6 +12226,242 @@ pub(crate) mod tests {
...
+
+                        // Remove the rest of the slots before the next iteration
+                        for (slot, bank_id) in slots_on_fork {
+                            bank_at_fork_tip.remove_unrooted_slots(&[(slot, bank_id)]);
+                        }
+                    }
+                },
+                None,
+                // Test removing the slot before the scan starts, should error every time
+                AcceptableScanResults::Both,
```
carllin, line 12464: These 3 tests show that safety of scan/remove_unrooted slots is upheld in 3 different interleavings of events:

1. Remove the slot before scan of the slot starts
2. Remove one version of the slot, recreate a second version of the slot, start a scan on the first version of the slot
3. Removing slot in the middle of an ongoing scan

Also `Bank::drop()` runs at the end of every iteration of the scans in the scan thread, so these also test interleaving that cleanup logic with ongoing scans and removes

runtime/src/accounts_index.rs
```diff
@@ -570,6 +597,17 @@ pub struct AccountsIndex<T> {
...
+    // is iterating over. The unique id of that slot `S` is recorded here (note we don't use
+    // `S` as the id because there can be more than one version of a slot `S`). If a fork
+    // is abandoned, all of the slots on that fork up to `S` will be removed via
+    // `AccountsDb::remove_unrooted_slots()`. When the scan finishes, it'll realize that the
+    // results of the scan may have been corrupted by `remove_unrooted_slots` and abort its results.
+    //
+    // `removed_bank_ids` tracks all the slot ids that were removed via `remove_unrooted_slots()` so any attempted scans
+    // on any of these slots fails. This is safe to purge once the associated Bank is dropped and
+    // scanning the fork with that Bank at the tip is no longer possible.
+    pub removed_bank_ids: Mutex<HashSet<BankId>>,
```
carllin, line 610: When we remove the state for for slot S, we mark down the unique bank id of this particular Bank so that any ongoing scan over this bank will know their state may have been corrupted.

Note we mark down the unique bank id and not the slot number, because we may start replaying another version of the same slot S where the slot number is the same, but the bank id will distinguish between two versions of the same slot S

runtime/src/accounts_db.rs
```diff
@@ -3507,13 +3533,26 @@ impl AccountsDb {
...
 
+        // Mark down these slots are about to be purged so that new attempts to scan these
+        // banks fail, and any ongoing scans over these slots will detect that they should abort
+        // their results
+        {
+            let mut locked_removed_bank_ids = self.accounts_index.removed_bank_ids.lock().unwrap();
+            for (_slot, remove_bank_id) in remove_slots.iter() {
+                locked_removed_bank_ids.insert(*remove_bank_id);
+            }
+        }
```
carllin, line 3544: Marking down these bank's account states have been removed

client/src/rpc_custom_error.rs
CriesofCarrots: Okay, then. If you're sure it's worth the churn, your error/result implementation looks a-okay!

client/src/rpc_custom_error.rs
```diff
@@ -141,6 +148,11 @@ impl From<RpcCustomError> for Error {
                 message: "Transaction history is not available from this node".to_string(),
                 data: None,
             },
+            RpcCustomError::ScanError(scan_err) => Self {
+                code: ErrorCode::ServerError(JSON_RPC_REMOVED_SLOT),
```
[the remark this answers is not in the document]
carllin: Renamed `JSON_RPC_REMOVED_SLOT` -> `JSON_RPC_SCAN_ERROR` to cover all possible scan errors.

runtime/src/accounts_index.rs
```diff
@@ -787,6 +838,23 @@ impl<T: 'static + Clone + IsCached + ZeroLamport> AccountsIndex<T> {
                 ongoing_scan_roots.remove(&max_root);
             }
         }
+
+        // If the fork with tip at bank `scan_bank_id` was removed durin our scan, then the scan
```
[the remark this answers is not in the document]
carllin: The check before the scan operation is to avoid incurring the cost of the scan if we already know that the slot has been dumped

The check after the scan operation is to account for slot dumping that occurred during the scan. Unfortunately, it's as you said, if the dump happens after the scan was finished and before this check is made, then we chalk this up to bad luck and still abort the results. I don't know of a great way to avoid this without blocking the remove during an ongoing scan.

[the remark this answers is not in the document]
lijunwangs: If the latter case can happen, which is after the scan and before the check, then it is also possible it can happen after we did the check and receive the negative answer and then it happens right after that while we already made decision to return a success. Would that have any correctness concern?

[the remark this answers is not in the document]
carllin: @lijunwangs correct that can happen as well. That should be ok because the results returned in that case are guaranteed to not have been corrupted by the slot dumping, which upholds our guarantee that scans should be consistent up to slot *boundaries*

lijunwangs (commented):
lgtm

lijunwangs (approved):

PR #17473, opened 2021-05-25
Only add keys with multiple index entries to the clean set

sakridge:
#### Problem

Dirty pubkey set will consist of all pubkeys in the snapshot which means the clean operation following a snapshot ingest is very slow when there are a lot of pubkeys in the snapshot.

#### Summary of Changes

Only add keys which have multiple entries to the dirty set. Pubkeys with only 1 entry cannot be cleaned unless they are 0-lamport updates. Those are already handled through the accounts_index::zero_lamport_pubkeys set:
https://github.com/solana-labs/solana/blob/1108cb32b524fedf7ae85bf694160033be7e60af/runtime/src/accounts_index.rs#L1229-L1231

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -5087,7 +5087,7 @@ impl AccountsDb {
                         let len = accounts_map.len();
 
                         let mut items = Vec::with_capacity(len);
-                        let dirty_keys = accounts_map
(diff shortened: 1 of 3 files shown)


sakridge:
Moved from https://github.com/solana-labs/solana/pull/17405

jeffwashington (approved):

PR #17482, opened 2021-05-25
implement ancestors as rolling bit field

jeffwashington:
#### Problem
ancestors has gone from HashMap<Slot, usize> to HashSet<Slot>
#### Summary of Changes
implement Ancestors in a more performant and straightforward manner using the RollingBitField we use for roots.
Fixes #


sakridge (approved):
lgtm

PR #17485, opened 2021-05-25
Add a flag to simulateTransaction to use most recent blockhash

jstarry:
#### Problem
RPC consistency issues makes it hard to fetch a recent blockhash for tx simulation and then use that blockhash for simulation. It also shouldn't be necessary to require clients to first fetch a recent blockhash for simulation.

#### Summary of Changes
- Add `replaceRecentBlockhash` flag to replace the recent blockhash used by a simulated transaction with the last blockhash

Fixes #

### DIFF
--- a/client/src/rpc_config.rs
+++ b/client/src/rpc_config.rs
@@ -28,6 +28,8 @@ pub struct RpcSendTransactionConfig {
 pub struct RpcSimulateTransactionConfig {
     #[serde(default)]
     pub sig_verify: bool,
+    #[serde(default)]
+    pub use_most_recent_blockhash: bool,
     #[serde(flatten)]
     pub commitment: Option<CommitmentConfig>,
     pub encoding: Option<UiTransactionEncoding>,
--- a/client/src/rpc_config.rs
+++ b/client/src/rpc_config.rs
@@ -29,7 +29,7 @@ pub struct RpcSimulateTransactionConfig {
     #[serde(default)]
     pub sig_verify: bool,
     #[serde(default)]
-    pub use_most_recent_blockhash: bool,
+    pub replace_recent_blockhash: bool,
     #[serde(flatten)]
     pub commitment: Option<CommitmentConfig>,
     pub encoding: Option<UiTransactionEncoding>,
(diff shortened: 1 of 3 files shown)


docs/src/developing/clients/jsonrpc-api.md
```diff
@@ -3206,6 +3206,7 @@ Simulate sending a transaction
   - `sigVerify: <bool>` - if true the transaction signatures will be verified (default: false)
   - `commitment: <string>` - (optional) [Commitment](jsonrpc-api.md#configuring-state-commitment) level to simulate the transaction at (default: `"finalized"`).
   - `encoding: <string>` - (optional) Encoding used for the transaction data. Either `"base58"` (*slow*, **DEPRECATED**), or `"base64"`. (default: `"base58"`).
+  - `useMostRecentBlockhash: <bool>` - (optional) if true the transaction recent blockhash will be ignored and overridden with the most recent blockhash. (default: false)
```
mvines, line 3209: We probably should error if `sigVerify` is true when this flag is set

client/src/rpc_config.rs
```diff
@@ -28,6 +28,8 @@ pub struct RpcSendTransactionConfig {
 pub struct RpcSimulateTransactionConfig {
     #[serde(default)]
     pub sig_verify: bool,
+    #[serde(default)]
+    pub use_most_recent_blockhash: bool,
```
mvines, line 32: Maybe this?
```suggestion
    pub replace_recent_blockhash: bool,
```

"most recent blockhash" reads a rough in my brain :)

docs/src/developing/clients/jsonrpc-api.md
jstarry: I added the recent_blockhash override after sig verification to avoid conflicts there. Due to RPC consistency issues I think it's still reasonable for a client to want enable both options. I don't think it does any harm to allow both but it is a little odd so if you feel strongly enough we can prohibit it

client/src/rpc_config.rs
jstarry: works for me

docs/src/developing/clients/jsonrpc-api.md
mvines: Yeah, I think we should prohibit it for now. It doesn't sit right to me that we first verify a signature, then promptly modify the message that was just verified even though it doesn't really matter in this case.

jstarry: Fixed and added tests

mvines (approved):

Issue #17451, opened 2021-05-25
Explorer: Block program list is incorrect


jstarry:
#### Problem
The programs called from inner instructions are listed on the block details page but are not always correct. The program id index is used incorrectly.

Example: [This block](https://explorer.solana.com/block/79856899/programs) incorrectly lists a [non executable account](https://explorer.solana.com/address/3Xg9Q4VtZhD4bVYJbTfgGWFV5zjE3U7ztSHa938zizte) as a program.

#### Proposed Solution
- Retrieve the original message account keys ordering from web3.js to determine which program was invoked in an inner instruction

Blocked on: https://github.com/solana-labs/solana/pull/17449

PR #17489, opened 2021-05-25
explorer: Display correct list of block programs

jstarry:
#### Problem
The programs called from inner instructions are listed on the block details page but are not always correct. The program id index is used incorrectly.

#### Summary of Changes
Use new `getBlock` web3 api to ensure that message keys are accessed in the correct order

Fixes https://github.com/solana-labs/solana/issues/17451


oJshua (approved):
Looks good to me!

PR #17494, opened 2021-05-25
Add custom error for tx-history queries when node does not support

### LINKED ISSUE #9316: RPC does not expose `enable_rpc_transaction_history` config setting
#### Problem
There is no way a client can determine whether a node has `enable_rpc_transaction_history` set to false, or if confirmed blocks / statuses have been purged.

#### Proposed Solution
* Return an error from `getConfirmedBlock` when `enable_rpc_transaction_history == false`
* Return an error from  `getSignatureStatuses` when `enable_rpc_transaction_history == false` and `searchTransactionHistory == true`
* Add a `getConfig` endpoint that returns config settings for a node

CriesofCarrots:
#### Problem
Clients making RPC queries for historical data cannot distinguish between a node that has `enable_rpc_transaction_history` set to false and one that simply no longer has the data retained ledger.

#### Summary of Changes
Add custom rpc error to return when a historical query is received (`getBlock`, `getTransaction`, or `getSignatureStatuses` with `searchTransactionHistory: true`) and `!config.enable_rpc_transaction_history`

Fixes #9316

### DIFF
--- a/client/src/rpc_custom_error.rs
+++ b/client/src/rpc_custom_error.rs
@@ -16,6 +16,7 @@ pub const JSON_RPC_SERVER_ERROR_SLOT_SKIPPED: i64 = -32007;
 pub const JSON_RPC_SERVER_ERROR_NO_SNAPSHOT: i64 = -32008;
 pub const JSON_RPC_SERVER_ERROR_LONG_TERM_STORAGE_SLOT_SKIPPED: i64 = -32009;
 pub const JSON_RPC_SERVER_ERROR_KEY_EXCLUDED_FROM_SECONDARY_INDEX: i64 = -32010;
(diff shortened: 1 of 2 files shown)


jstarry:
looks good, should we add it for `get_confirmed_signatures_for_address` too or was that intentionally excluded?

CriesofCarrots:
> looks good, should we add it for `get_confirmed_signatures_for_address` too or was that intentionally excluded?

Yes, I did intentionally exclude it since that method is deprecated. Happy to reconsider, if you feel strongly about it @jstarry .

jstarry:
> > looks good, should we add it for `get_confirmed_signatures_for_address` too or was that intentionally excluded?
> 
> Yes, I did intentionally exclude it since that method is deprecated. Happy to reconsider, if you feel strongly about it @jstarry .
> Merging now; can make follow-up if needed.

Nope that's fine!

PR #17506, opened 2021-05-26
Add last valid block height to rpc Fees

CriesofCarrots:
#### Problem
The `lastValidSlot` returned from `getFees` rpc has been deprecated, but nothing has been added to replace it.

#### Summary of Changes
- Add and populate `lastValidBlockHeight`
- Add `getBlockHeight` rpc to make drop-in replacement for `lastValidSlot`/`getSlot`

I plan to update solana client tooling to use `lastValidaBlockHeight` (and probably add details to exchange integration docs) in a separate PR for v1.7 only. Then we can backport to v1.6 in a future release after the clusters have upgraded to the new struct.


mvines (approved):
`getBlockHeight` - interesting that it took us this long to add such an RPC method!

PR lgtm

PR #17515, opened 2021-05-26
docs: Add `find_program_address` and example

joncinque:
#### Problem

There were some questions about using `create_program_address`, and typically `find_program_address` is a part of that development flow, but the docs don't have any mention or examples using `find_program_address`.

#### Summary of Changes

Add the `find_program_address` implementation and a similar escrow example using `find_program_address` instead of `create_program_address`.

Fixes #


mvines (approved):

jackcmay:
Looks great, thanks!

PR #17519, opened 2021-05-26
hoist slot() call in accounts hash scan

jeffwashington:
#### Problem
Move local out of loop for performance.
#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -4227,13 +4227,10 @@ impl AccountsDb {
 
                 for sub_storages in storages {
                     for storage in sub_storages {
+                        let slot = storage.slot();
                         let accounts = storage.accounts.accounts(0);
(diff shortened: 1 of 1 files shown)


jeffwashington:
```
with mnb snapshot
without this change:
calculate_accounts_hash_without_index accounts_scan=3463315i ... total=4161005i

with this change:
calculate_accounts_hash_without_index accounts_scan=3310614i ... total=3971208i

```

brooksprumo (approved):
Wow, quite a performance improvement! I'm actually surprised the compiler didn't optimize this already.

jbiseda (approved):

mvines:
> Wow, quite a performance improvement! I'm actually surprised the compiler didn't optimize this already.

Ah, I guess because `.slot()` is backed an `AtomicU64` so the compiler can't assume the value wouldn't change between iterations.

Nice find though!

PR #17521, opened 2021-05-26
Do not keep oldest snapshot

brooksprumo:
`snapshot_utils::purge_old_snapshot_archives()` kept around the oldest snapshot. After talking with @mvines, turns out that is no longer necessary, so I'm removing it.

### DIFF
--- a/runtime/src/snapshot_utils.rs
+++ b/runtime/src/snapshot_utils.rs
@@ -737,9 +737,7 @@ pub fn purge_old_snapshot_archives<P: AsRef<Path>>(
         maximum_snapshots_to_retain
     );
     let mut archives = get_snapshot_archives(snapshot_output_dir);
-    // Keep the oldest snapshot so we can always play the ledger from it.
-    archives.pop();
-    let max_snaps = max(1, maximum_snapshots_to_retain);
+    let max_snaps = max(1, maximum_snapshots_to_retain); // Always keep at least one snapshot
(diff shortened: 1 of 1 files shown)


sakridge:
Why not?

sakridge:
It's nice to have an older snapshot if you need to replay up to a target slot.

brooksprumo:
> It's nice to have an older snapshot if you need to replay up to a target slot.

Fair point.

Personally I was surprised when running `bootstrap_validator` multiple times and one really old snapshot never got cleaned up. I'll close this PR as the current behavior is intended. Maybe this can be addressed with some later PR that keeps around older snapshots with some exponential-like backoff, similar to how many file backup services work. Or some other command line argument(s)?

mvines:
After a couple days of running the older snapshot is totally useless on a real mainnet or testnet node, it's just eating disk space

sakridge:
> After a couple days of running the older snapshot is totally useless on a real mainnet or testnet node, it's just eating disk space

True, it can only give information about the accounts state at that slot. How about a flag to keep it which is off by default?

sakridge:
The exponential backoff seems like a decent idea, it would be nice to have a snapshot which allows one to play from whatever slots the ledger has or maybe within a few hundred slots of the start.

mvines:
> The exponential backoff seems like a decent idea

Yeah I like that idea too

PR #17523, opened 2021-05-26
Add block height to ConfirmedBlock structs

CriesofCarrots:
#### Problem
As of https://github.com/solana-labs/solana/pull/17506, `last_valid_block_height` is available to assess recent-blockhash expiration. But it is only possible to get the block height of the current Bank (via `getBlockHeight` or `getEpochInfo` rpcs).

#### Summary of Changes
- Add block_height to ConfirmedBlock structs
- Create BlockHeight column in Blockstore
- Cache block height along with block time (and rename service to be more general, now that it's doing two things -- this is most of the churn)
- Return block_height with Blockstore block requests, which means it will also be stored in BigTable blocks

### DIFF
--- a/cli-output/src/cli_output.rs
+++ b/cli-output/src/cli_output.rs
@@ -2126,6 +2126,9 @@ impl fmt::Display for CliBlock {
         if let Some(block_time) = self.encoded_confirmed_block.block_time {
             writeln!(f, "Block Time: {:?}", Local.timestamp(block_time, 0))?;
         }
+        if let Some(block_height) = self.encoded_confirmed_block.block_height {
+            writeln!(f, "Block Height: {:?}", block_height)?;
+        }
         if !self.encoded_confirmed_block.rewards.is_empty() {
             let mut rewards = self.encoded_confirmed_block.rewards.clone();
             rewards.sort_by(|a, b| a.pubkey.cmp(&b.pubkey));
--- a/core/src/cache_block_meta_service.rs
+++ b/core/src/cache_block_meta_service.rs
@@ -1,4 +1,4 @@
-pub use solana_ledger::blockstore_processor::CacheBlockTimeSender;
+pub use solana_ledger::blockstore_processor::CacheBlockMetaSender;
 use {
     crossbeam_channel::{Receiver, RecvTimeoutError},
     solana_ledger::blockstore::Blockstore,
@@ -14,18 +14,18 @@ use {
     },
 };
 
-pub type CacheBlockTimeReceiver = Receiver<Arc<Bank>>;
+pub type CacheBlockMetaReceiver = Receiver<Arc<Bank>>;
 
-pub struct CacheBlockTimeService {
+pub struct CacheBlockMetaService {
     thread_hdl: JoinHandle<()>,
 }
 
 const CACHE_BLOCK_TIME_WARNING_MS: u64 = 150;
 
-impl CacheBlockTimeService {
+impl CacheBlockMetaService {
     #[allow(clippy::new_ret_no_self)]
     pub fn new(
-        cache_block_time_receiver: CacheBlockTimeReceiver,
+        cache_block_meta_receiver: CacheBlockMetaReceiver,
         blockstore: Arc<Blockstore>,
         exit: &Arc<AtomicBool>,
     ) -> Self {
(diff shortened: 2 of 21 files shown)


core/src/rpc.rs
```diff
@@ -953,7 +953,20 @@ impl JsonRpcRequestProcessor {
...
+                        {
+                            let r_bank_forks = self.bank_forks.read().unwrap();
+                            if let Some(bank) = r_bank_forks.get(slot) {
+                                if confirmed_block.block_time.is_none() {
+                                    confirmed_block.block_time = Some(bank.clock().unix_timestamp);
+                                }
+                                if confirmed_block.block_height.is_none() {
+                                    confirmed_block.block_height = Some(bank.block_height());
+                                }
+                            }
```
CriesofCarrots, line 968: @carllin , we actually weren't doing the BankForks fallback in `get_block()`, even though we were in `get_transaction()` and `get_block_time()`. Added now!

cli-output/src/cli_output.rs
```diff
@@ -2126,6 +2126,9 @@ impl fmt::Display for CliBlock {
         if let Some(block_time) = self.encoded_confirmed_block.block_time {
             writeln!(f, "Block Time: {:?}", Local.timestamp(block_time, 0))?;
         }
+        if let Some(block_height) = self.encoded_confirmed_block.block_height {
```
mvines, line 2129: heh, I foresee support issues when people now start running `solana block <BLOCK>` instead of `solana block <SLOT>`

mvines (approved):

mvines (commented):
There's a JSON RPC doc update missing here right?

core/src/cache_block_meta_service.rs
```diff
@@ -59,10 +59,13 @@ impl CacheBlockTimeService {
         Self { thread_hdl }
     }
 
-    fn cache_block_time(bank: Arc<Bank>, blockstore: &Arc<Blockstore>) {
+    fn cache_block_meta(bank: Arc<Bank>, blockstore: &Arc<Blockstore>) {
         if let Err(e) = blockstore.cache_block_time(bank.slot(), bank.clock().unix_timestamp) {
             error!("cache_block_time failed: slot {:?} {:?}", bank.slot(), e);
         }
+        if let Err(e) = blockstore.cache_block_height(bank.slot(), bank.block_height()) {
```
carllin, line 66: @CriesofCarrots just summarizing our discussion, because this is async with replay, there's the possibility that somebody queries a block and the status/height is not yet written, in which case we'll return `None` for those fields.

Should we document this behavior?

ledger/src/blockstore.rs
```diff
@@ -1764,6 +1767,28 @@ impl Blockstore {
...
 
+    pub fn get_block_height(&self, slot: Slot) -> Result<Option<u64>> {
+        datapoint_info!(
+            "blockstore-rpc-api",
+            ("method", "get_block_height".to_string(), String)
+        );
+        let lowest_cleanup_slot = self.lowest_cleanup_slot.read().unwrap();
+        // lowest_cleanup_slot is the last slot that was not cleaned up by
+        // LedgerCleanupService
+        if *lowest_cleanup_slot > 0 && *lowest_cleanup_slot >= slot {
```
carllin, line 1778: lol beginning to realize this should probably be called `highest_cleanup_slot`

core/src/rpc.rs
```diff
@@ -953,7 +953,20 @@ impl JsonRpcRequestProcessor {
                             .load(Ordering::SeqCst)
                 {
                     let result = self.blockstore.get_complete_block(slot, true);
-                    return Ok(result.ok().map(|confirmed_block| {
+                    return Ok(result.ok().map(|mut confirmed_block| {
+                        if confirmed_block.block_time.is_none()
+                            || confirmed_block.block_height.is_none()
+                        {
+                            let r_bank_forks = self.bank_forks.read().unwrap();
```
carllin, line 960: nit usually like to clone out the `Arc<Bank>` to hold the `bank_forks` lock as little as possible

core/src/rpc.rs
carllin: Got it, so this should be fine as long as the statuses are written to before the `BankForks` new root purges these slots, ~16 seconds

ledger/src/blockstore.rs
```diff
@@ -1764,6 +1767,28 @@ impl Blockstore {
...
+            return Err(BlockstoreError::SlotCleanedUp);
+        }
+        self.block_height_cf.get(slot)
+    }
+
+    pub fn cache_block_height(&self, slot: Slot, block_height: u64) -> Result<()> {
+        if self.get_block_height(slot).unwrap_or_default().is_none() {
```
carllin, line 1785: I think it's ok to forgo this check here, if we're going to overwrite this it should be the same value, unless we have need to purge a slot, in which case the newer value should take precedent anyways

cli-output/src/cli_output.rs
CriesofCarrots: Totally. `solana block-by-height` in our future?

ledger/src/blockstore.rs
CriesofCarrots: Oh geez, 💯 
A fix for another PR, though.

core/src/cache_block_meta_service.rs
CriesofCarrots: Good call. I'll add a comment in `Blockstore::get_complete_block` that speaks to block time and block height.

core/src/rpc.rs
CriesofCarrots: Exactly

cli-output/src/cli_output.rs
mvines: `solana-ledger-tool slot ...` got it right though!  oh well :)

mvines (approved):

PR #17525, opened 2021-05-26
RollingBitField: bug fixes and add tests

jeffwashington:
#### Problem
RollingBitField will soon be the basis of Ancestors. Using this for Ancestors, which has different access patterns, has exposed some bugs and missing traits.

#### Summary of Changes
Improve RollingBitField.
Fixes #

### DIFF
--- a/runtime/src/accounts_index.rs
+++ b/runtime/src/accounts_index.rs
@@ -311,7 +311,7 @@ impl RollingBitField {
                 self.bits.set(address, false);
                 self.purge(&key);
 
-                if self.count == self.excess.len() {
+                if self.all_items_in_excess() {
                     // if we moved the last existing item to excess, then we are ready to insert the new item in the bits
                     bits_empty = true;
                     break;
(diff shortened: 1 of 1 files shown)


runtime/src/accounts_index.rs
```diff
@@ -1963,6 +2082,13 @@ pub mod tests {
         }
     }
 
+    fn compare(hashset: &HashSet<u64>, bitfield: &RollingBitField) {
+        compare_internal(hashset, bitfield);
+        let clone = bitfield.clone();
+        compare_internal(hashset, &clone);
+        assert!(clone.eq(&bitfield));
```
sakridge, line 2089: `assert_eq!(clone, bitfield)`?

runtime/src/accounts_index.rs
```diff
@@ -293,6 +310,12 @@ impl RollingBitField {
                 let address = self.get_address(&key);
                 self.bits.set(address, false);
                 self.purge(&key);
+
+                if self.count == self.excess.len() {
```
sakridge, line 314: `if self.all_items_in_excess() {` here?

sakridge (approved):

PR #17527, opened 2021-05-26
add metric for collecting storages

jeffwashington:
#### Problem
we don't know how long it takes to collect storages for hash calculation.
#### Summary of Changes
add a metric.
Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -4244,11 +4244,19 @@ impl AccountsDb {
         check_hash: bool,
     ) -> Result<(Hash, u64), BankHashVerificationError> {
         if !use_index {
+            let mut time = Measure::start("collect");
             let combined_maps = self.get_snapshot_storages(slot);
+            time.stop();
+
+            let timings = HashStats {
+                collect_snapshots_us: time.as_us(),
+                ..HashStats::default()
+            };
 
             Self::calculate_accounts_hash_without_index(
                 &combined_maps,
                 Some(&self.thread_pool_clean),
+                timings,
                 check_hash,
             )
         } else {
@@ -4365,10 +4373,10 @@ impl AccountsDb {
     pub fn calculate_accounts_hash_without_index(
         storages: &[SnapshotStorage],
         thread_pool: Option<&ThreadPool>,
+        mut stats: HashStats,
         check_hash: bool,
     ) -> Result<(Hash, u64), BankHashVerificationError> {
-        let scan_and_hash = || {
-            let mut stats = HashStats::default();
+        let mut scan_and_hash = move || {
             // When calculating hashes, it is helpful to break the pubkeys found into bins based on the pubkey value.
             // More bins means smaller vectors to sort, copy, etc.
             const PUBKEY_BINS_FOR_CALCULATING_HASHES: usize = 64;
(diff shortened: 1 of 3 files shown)


jeffwashington:
datapoint: calculate_accounts_hash_without_index accounts_scan=2308835i eliminate_zeros=404897i hash=250392i sort=1838954i hash_total=61837974i flatten=1571118i unreduced_entries=62270664i **collect_snapshots_us=231946i** num_snapshot_storage=372887i total=6606142i

runtime/src/accounts_db.rs
```diff
@@ -4339,9 +4347,9 @@ impl AccountsDb {
     pub fn calculate_accounts_hash_without_index(
         storages: &[SnapshotStorage],
         thread_pool: Option<&ThreadPool>,
+        mut stats: HashStats,
```
lijunwangs, line 4350: Where is this used?

jeffwashington: this function passes it to AccountsHash::rest_of_hash_calculation.
rest_of_hash_calculation eventually logs the datapoint after the last pass:
```
calculate_accounts_hash_without_index accounts_scan=3310614i eliminate_zeros=403638i hash=242073i sort=0i hash_total=61837974i flatten=0i storage_sort_us=14883i unreduced_entries=0i num_snapshot_storage=372887i total=3971208i
```

lijunwangs: I see, you replace the stats which was defined locally -- I mistook it was totally new. Any reason why it is not passed in by reference &mut HashStats which I thought is more efficient?

runtime/src/accounts_db.rs
```diff
@@ -4246,11 +4246,19 @@ impl AccountsDb {
         ancestors: &Ancestors,
     ) -> (Hash, u64) {
         if !use_index {
+            let mut time = Measure::start("collect");
```
lijunwangs, line 4249: maybe "snapshot_collect" instead of "collect" which is too generic?

jeffwashington: this string doesn't get used anywhere. If you print 'time', it will use the string, but I want numbers out, so I call as_us().

runtime/src/accounts_db.rs
jeffwashington: I want to transfer ownership in this case, so the function takes the stats by value. I imagine this is passing a reference in generated code anyway. You could verify this with the rust code generator webpage. Also, this is called once per many minutes. Another alternative was for me to pass just this u64 time into the function for it to add to its stats. I have found passing the stats struct around is the preferred method. It allows for measurements to be added more easily and reduces # of parameters to functions.
Also, struct HashStats does not implement Copy, so no copy can be made or there would be a compile error.
Feel free to ask followups on any of these assertions I've made. ;-)

runtime/src/accounts_db.rs
lijunwangs: okay

runtime/src/accounts_db.rs
lijunwangs: sounds good. So the compiler can do optimization on this for you?

lijunwangs (dismissed):
lgtm

jeffwashington:
here is a dashboard of these metrics.
https://metrics.solana.com:8888/sources/3/chronograf/data-explorer?query=SELECT%20mean%28%22flatten%22%29%20AS%20%22mean_flatten%22%20FROM%20%22mainnet-beta%22.%22autogen%22.%22calculate_accounts_hash_without_index%22%20WHERE%20time%20%3E%20now%28%29%20-%2030d%20GROUP%20BY%20time%28%3Ainterval%3A%29%20FILL%28null%29

runtime/src/accounts_db.rs
jeffwashington: lijun and I resolved this offline and with godbolt.

PR #17528, opened 2021-05-26
Plumb transaction-level rewards (aka "rent debits") into the `getTransaction` RPC method

mvines:
Rent debits that occur during Transaction execution are not nicely surfaced in RPC.  Hang a `rewards` object off transaction metadata that's analogous to the block-level `rewards` object

### DIFF
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -484,6 +484,7 @@ pub struct TransactionResults {
     pub fee_collection_results: Vec<Result<()>>,
     pub execution_results: Vec<TransactionExecutionResult>,
     pub overwritten_vote_accounts: Vec<OverwrittenVoteAccount>,
+    pub rent_debits: Vec<RentDebits>,
 }
 pub struct TransactionBalancesSet {
     pub pre_balances: TransactionBalances,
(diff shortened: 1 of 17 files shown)


runtime/src/bank.rs
```diff
@@ -3580,13 +3583,11 @@ impl Bank {
             let loaded_transaction = raccs.as_mut().unwrap();
 
             collected_rent += loaded_transaction.rent;
-            self.rewards
-                .write()
-                .unwrap()
-                .append(&mut loaded_transaction.rent_debits.0);
+            rent_debits.push(mem::take(&mut loaded_transaction.rent_debits));
```
mvines, line 3586: The story starts here building off what @t-nelson added yesterday: instead of appending transaction-level rent debits to the block rewards, they instead now get returned to the caller and routed though transaction status service

CriesofCarrots (dismissed):
Lgtm. I focused especially on the send-tx-svc through storage aspects.
Out of curiosity, did you test querying blocks pre- and post-change with a bigtable emulator at all?

mvines:
> Out of curiosity, did you test querying blocks pre- and post-change with a bigtable emulator at all?

Great idea, I'll run some tests later tonight before merging

t-nelson (commented):
Post-merge lgtm!  Thanks for picking up my slack!

PR #17540, opened 2021-05-27
Add reference to text-encoder module in keypair.ts

ctapang:
#### Problem
When calling Keypair.fromSecretKey, getting the error "ReferenceError: TextEncoder is not defined"

#### Summary of Changes
Add npm module text-encoder and add reference to it in keypair.ts

Fixes #


ctapang:
I ran out of time to get this through the checks.

PR #17550, opened 2021-05-27
Add create-stake command to solana-tokens CLI

garious:
#### Problem

When we created the `solana-tokens` CLI, we only had locked up tokens and so the `distribute-stake` command requires a source stake account to split up and a lockup authority signer to change the target accounts' lockup dates. Now that we have unlocked tokens, it is tedious to manually move tokens into stake accounts just so `solana-tokens` can split them. Instead, it can skip the split and just create the stake accounts directly.

#### Summary of Changes

Add `create-stake` command to `solana-tokens` CLI.  It uses the `--from` System account to both send the `recipient` SOL for fees and to fund the new stake accounts. Unlike the `distribute-stake` stake command, it does not require a source stake account or a lockup authority signer.

### DIFF
--- a/tokens/src/arg_parser.rs
+++ b/tokens/src/arg_parser.rs
@@ -149,6 +149,14 @@ where
                         .help("Keypair to fund accounts"),
                 )
                 .arg(
+                    Arg::with_name("unlocked_sol")
+                        .default_value("1.0")
+                        .long("unlocked-sol")
+                        .takes_value(true)
+                        .value_name("SOL_AMOUNT")
+                        .help("Amount of SOL to put in system account to pay for fees"),
+                )
+                .arg(
                     Arg::with_name("lockup_authority")
                         .long("lockup-authority")
                         .takes_value(true)
--- a/tokens/src/arg_parser.rs
+++ b/tokens/src/arg_parser.rs
@@ -1,5 +1,6 @@
 use crate::args::{
-    Args, BalancesArgs, Command, DistributeTokensArgs, SplTokenArgs, StakeArgs, TransactionLogArgs,
+    Args, BalancesArgs, Command, DistributeTokensArgs, SenderStakeArgs, SplTokenArgs, StakeArgs,
+    TransactionLogArgs,
 };
 use clap::{
     crate_description, crate_name, value_t, value_t_or_exit, App, Arg, ArgMatches, SubCommand,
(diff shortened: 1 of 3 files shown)


tokens/src/commands.rs
```diff
@@ -705,16 +759,17 @@ fn check_payer_balances(
...
-        undistributed_tokens -= total_unlocked_sol;
-        (
-            stake_args.stake_account_address,
-            Some((args.sender_keypair.pubkey(), total_unlocked_sol)),
-        )
-    } else {
-        (args.sender_keypair.pubkey(), None)
-    };
+    let (distribution_source, unlocked_sol_source) =
+        if let StakeArgs::SplitStake(split_stake_args) = &args.stake_args {
```
garious, line 763: Note to self: add the `CreateStake` case (and a test!)

garious:
Hold off on reviewing. The implementation can be simpler and I'll play with integrating this functionality into one of the existing commands.

CriesofCarrots:
> Hold off on reviewing. The implementation can be simpler and I'll play with integrating this functionality into one of the existing commands.

ack

garious:
@CriesofCarrots @danpaul000, ready for review.  Something to consider: we could merge the `distribute-tokens` and `create-stake` commands by making the rule that if a lockup date is in the input CSV file, then it should create a stake account.  That'd also mean that the output CSV file would contain a blank column of `new_stake_account_address` when that feature is not used, which might seem weird.  For now, I'm thinking we should just let separate things be separate.

garious:
@CriesofCarrots @danpaul000, wdyt?

tokens/src/arg_parser.rs
```diff
@@ -363,6 +427,58 @@ fn parse_distribute_tokens_args(
...
+                &signer_matches,
+                &path,
+                "lockup authority",
+                &mut wallet_manager,
+            )
+        })
+        .transpose()?;
+
+    let stake_args = StakeArgs {
+        unlocked_sol: sol_to_lamports(value_t_or_exit!(matches, "unlocked_sol", f64)),
```
CriesofCarrots, line 465: I don't see an `Arg::with_name("unlocked_sol")` on `create-stake`

CriesofCarrots (commented):
Generally looks good, but one missing cli Arg.

> For now, I'm thinking we should just let separate things be separate.

Agreed, I like the commands being explicit, although I do think `distribute-stake` is a little ambiguous now. Fine for now, though.

PR #17554, opened 2021-05-27
Per-program id timings

sakridge:
#### Problem

Hard to tell which program is taking the longest in the network.

#### Summary of Changes

Add per-program timings configurable with a validator flag.

Fixes #

### DIFF
--- a/core/src/progress_map.rs
+++ b/core/src/progress_map.rs
@@ -104,6 +104,27 @@ impl ReplaySlotStats {
                 i64
             ),
         );
+
+        let mut per_pubkey_timings: Vec<_> = self
+            .execute_timings
+            .details
+            .per_program_timings
+            .iter()
+            .collect();
+        per_pubkey_timings.sort_by(|a, b| b.1 .0.cmp(&a.1 .0));
+        let total: u64 = per_pubkey_timings.iter().map(|a| a.1 .0).sum();
+        for (pubkey, time) in per_pubkey_timings.iter().take(5) {
+            datapoint_info!(
+                "per_program_timings",
+                ("pubkey", pubkey.to_string(), String),
+                ("execute_us", time.0, i64)
+            );
+        }
+        datapoint_info!(
+            "per_program_timings",
+            ("pubkey", "all", String),
+            ("execute_us", total, i64)
+        );
     }
 }
 
--- a/runtime/src/message_processor.rs
+++ b/runtime/src/message_processor.rs
@@ -4,6 +4,7 @@ use crate::{
 };
 use log::*;
 use serde::{Deserialize, Serialize};
+use solana_measure::measure::Measure;
 use solana_sdk::{
     account::{AccountSharedData, ReadableAccount, WritableAccount},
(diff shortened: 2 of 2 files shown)


core/src/progress_map.rs
```diff
@@ -104,6 +104,27 @@ impl ReplaySlotStats {
...
+
+        let mut per_pubkey_timings: Vec<_> = self
+            .execute_timings
+            .details
+            .per_program_timings
+            .iter()
+            .collect();
+        per_pubkey_timings.sort_by(|a, b| b.1.cmp(&a.1));
+        let total: u64 = per_pubkey_timings.iter().map(|a| a.1).sum();
+        for (pubkey, time) in per_pubkey_timings.iter().take(5) {
```
taozhu-chicago, line 116: Perhaps to remove `take(5)` when utilize this change to feed cost model?

core/src/validator.rs
```diff
@@ -193,6 +194,7 @@ impl Default for ValidatorConfig {
             tpu_coalesce_ms: DEFAULT_TPU_COALESCE_MS,
             validator_exit: Arc::new(RwLock::new(ValidatorExit::default())),
             no_wait_for_vote_to_start_leader: true,
+            collect_per_program_timings: false,
```
taozhu-chicago, line 197: default it to `true`?

core/src/progress_map.rs
sakridge: Well this is just to report metrics for the top 5 costly programs. So I think we would leave this in. There would just be another code-path to update the cost model with these values.

taozhu-chicago: right, makes sense, this function just reports stats. Will add code at replay_stage to read execute_timings for cost model

runtime/src/message_processor.rs
```diff
@@ -1222,28 +1228,36 @@ impl MessageProcessor {
...
+                    timings,
+                    demote_sysvar_write_locks,
+                    account_db.clone(),
+                    ancestors,
+                )
+                .map_err(|err| TransactionError::InstructionError(instruction_index as u8, err));
+            time.stop();
+            if timings.collect_per_program_timings {
+                let program_id = instruction.program_id(&message.account_keys);
+                *timings.per_program_timings.entry(*program_id).or_insert(0) += time.as_us();
```
taozhu-chicago, line 1258: In the case of same program being executed multiple times in a transaction, do we want to see the accumulated execution time in stats report, or perhaps averaged time? 
Cost model interested in later, if reporting accumulated time is desired, it is always possible to take average at the time of updating cost model.

sakridge: yes. I think we'll have to have a count stat here as well. I'll add that.

core/src/validator.rs
sakridge: I'll just remove it, if we need it for the cost model then we need to do it all the time anyway.

taozhu-chicago (approved):
looks good

PR #8339, opened 2020-02-20
Correct missing entry handling to avoid bad warns

ryoqun:
#### Problem

Bogus warnings is introduced by #7281. Also this is needed for #8337 because compacting at `generate_index` causes a lot of missing storage entries.

#### Summary of Changes

~Put empty dummy files for missing storage entry's AppendVec in the snapshot file.~

=> Changed the design strategy to change `AccountStorageSerialize` as discussed on discord.

~*This changes ABI for snapshots.*~

=> this is not true anymore; this shouldn't introduce any ABI changes.

PR #17559, opened 2021-05-27
remove unnecessary clone

jeffwashington:
#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -4537,10 +4537,7 @@ impl Bank {
     }
 
     pub fn get_snapshot_storages(&self) -> SnapshotStorages {
-        self.rc
-            .get_snapshot_storages(self.slot())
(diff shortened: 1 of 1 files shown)


runtime/src/bank.rs
```diff
@@ -4537,10 +4537,7 @@ impl Bank {
     }
 
     pub fn get_snapshot_storages(&self) -> SnapshotStorages {
-        self.rc
-            .get_snapshot_storages(self.slot())
-            .into_iter()
-            .collect()
+        self.rc.get_snapshot_storages(self.slot())
```
ryoqun, line 4540: so this was unnecessary in the first place.. https://github.com/solana-labs/solana/pull/8339/files#diff-ed47b4a0198313377e091bb3957bbbc63d937805426d1b2b6de39d0a50d32a0cR1870

PR #17560, opened 2021-05-27
refactor SnapshotStorage helpers to prepare for later changes

jeffwashington:
#### Problem
Storages are likely to be changed soon.
#### Summary of Changes
Make some helper test methods for common test values to remove noise from more interesting PRs.
Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -5599,13 +5599,18 @@ pub mod tests {
         ancestors
     }
 
+    fn empty_storages<'a>() -> &'a [SnapshotStorage] {
+        &[]
+    }
+
     #[test]
     #[should_panic(expected = "assertion failed: bins <= max_plus_1 && bins > 0")]
     fn test_accountsdb_scan_snapshot_stores_illegal_bins2() {
         let mut stats = HashStats::default();
         let bounds = Range { start: 0, end: 0 };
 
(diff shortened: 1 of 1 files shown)


runtime/src/accounts_db.rs
```diff
@@ -5781,7 +5793,7 @@ pub mod tests {
             },
         );
         assert_eq!(result.len(), 2); // 2 chunks
-        assert_eq!(result[0].len(), 0); // nothing found in first slots
+        assert_eq!(result[0].len(), 1); // nothing found in first slots TODO
```
brooksprumo, line 5796: Is there anything left todo here? Or, why did this assert value change?

runtime/src/accounts_db.rs
```diff
@@ -5877,7 +5892,7 @@ pub mod tests {
             },
         );
         assert_eq!(result.len(), 2); // 2 chunks
-        assert_eq!(result[0].len(), 0); // nothing found in first slots
+        assert_eq!(result[0].len(), 256); // nothing found in first slots TODO
```
brooksprumo, line 5895: Same question here about the TODO and changed value.

brooksprumo (commented):
The changes look very minimal, but I wasn't sure why the tests values changed though.

runtime/src/accounts_db.rs
jeffwashington: ah, because I left in the splice changes above. They will be gone soon in a new push.

brooksprumo (approved):

PR #17561, opened 2021-05-27
add missing error conversions

2501babe:
#### Problem
presently if you return `ProgramError::AccountNotRentExempt` from a program it resolves to `InstructionError::InvalidError`

#### Summary of Changes
i added the rent error conversions where i found them to be missing. i also added the invalidseeds conversion for completeness, since the maxseedlengthexceeded error is there, but im not sure if this is necessary or accomplishes anything


jackcmay:
Hi @2501babe , thanks for the PR!

The changes here are similar to an existing PR that is hung up around lost consensus and featuring.  Check out the discussion:  https://github.com/solana-labs/solana/pull/16231

Michael has proposed a pretty good solution to the consensus issue, just need to put it in place.

jackcmay:
Closing because these changes have been incorporated into #16231

PR #17563, opened 2021-05-27
support bin divisions up to 65536

jeffwashington:
#### Problem
Initial implementation of binning account pubkeys for scans was limited to a max of 256 bins. With the growth in accounts and more applications of storage scanning, more bins produce better performance.
#### Summary of Changes
Increase the max # of bins to 65536. The # of bins will be tuned as we go. There are several more changes coming due to having to scan ancestors, interleave with the write cache, scan in parallel, sort storages, rework the hash and lamport calculation to use less memory and to take advantage of storages sorted by slot.
Fixes #


sakridge (approved):
nice, looks good

PR #17569, opened 2021-05-28
Always bail if program modifies a ro account

jackcmay:
#### Problem

Performance optimizations resulted in the runtime ignoring changes by programs that attempted to write to read-only accounts.

#### Summary of Changes

- Always bail if a program writes a read-only account
- Add and update tests to validate this

Fixes #

### DIFF
--- a/programs/bpf/Cargo.lock
+++ b/programs/bpf/Cargo.lock
@@ -2997,6 +2997,13 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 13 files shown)


mvines:
(v1.6 as well I think, unless you want to delay activating the feature on mainnet for another couple months?)

jackcmay:
I'm planning on never activating these two features, so they can sit on mainnet as is and we can let this change seep on testnet/devnet

PR #17583, opened 2021-05-28
Disable read-only optimization features

jackcmay:
#### Problem

To RO optimizations discard written read-only data instead of halting the program

#### Summary of Changes

Disable these features

Fixes #


jstarry:
@jackcmay was there a v1.7 backport as well? Looks like those features are enabled on testnet so we'll want to be careful about removing them to avoid forking like we just saw on devnet: https://github.com/solana-labs/solana/issues/17715

PR #17589, opened 2021-05-28
parallel get_snapshot_storages

jeffwashington:
#### Problem
Getting snapshots is serially in line in accounts bg svc before clean, flush cache, shrink, hash calc, etc. Speeding it up would be useful. It will also be useful to have slots already pulled out because getting slot #s is relatively expensive. We are also replacing other calls that currently use index scans to use store scans. Getting storages will be called in more places. All the places that call to get storages immediately spin up threads to act on the storages in parallel.

#### Summary of Changes
get snapshots more quickly and get slot#s at the same time. The slot #s will be used in subsequent prs.
Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -4251,7 +4251,7 @@ impl AccountsDb {
     ) -> Result<(Hash, u64), BankHashVerificationError> {
         if !use_index {
             let mut time = Measure::start("collect");
-            let combined_maps = self.get_snapshot_storages(slot, Some(ancestors));
+            let combined_maps = self.get_snapshot_storages(slot, Some(ancestors)).0;
             time.stop();
 
             let timings = HashStats {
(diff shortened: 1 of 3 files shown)


jeffwashington:
stats (us)
get keys: 11328, m2: 40995, m3: 13523

without these changes: ~200ms to get storages.

runtime/src/accounts_db.rs
```diff
@@ -5125,26 +5125,70 @@ impl AccountsDb {
         }
     }
 
-    pub fn get_snapshot_storages(&self, snapshot_slot: Slot) -> SnapshotStorages {
-        self.storage
+    pub fn get_snapshot_storages(&self, snapshot_slot: Slot) -> (SnapshotStorages, Vec<Slot>) {
+        let mut m = Measure::start("get keys");
+        let keys = self
```
sakridge, line 5130: I'm confused a bit by the name `keys`. Aren't these slots?

jeffwashington: We're iterating a dashmap<Slot, ...> to get the keys, which are also slots. Algorithmically and implementation-wise, this is the cost to get the list of keys (which, again, are slots). I want to replace this get keys step in a later pass, to instead get a range of possible keys from min_slot..=snapshot_slot. min_slot isn't available reliably right now, I don't think. So, I broke out the cost of getting the keys for the moment. This is only used in logging anyway (and the "get keys" in Measure::start("get keys") isn't used at all - it is just there for clarity, ironically.

jeffwashington: ah, the variable keys. I attached to "get keys" when seeing the github snippet. Ok, yes, again - similar idea. These are the keys we'll lookup in the DashMap later, but they are also conceptually slots. So, ok, sure. 'slots' is fine, too.

sakridge: Yea. I think of hashmap keys and pubkeys. I think using hashmap keys as a name is probably not the best name since it's so generic unless the function is very trivial.

jeffwashington: I renamed it to slots.

sakridge (approved):

PR #17590, opened 2021-05-28
Add BPF rustc option to reduce the optimizations to safer level

dmakarov:
#### Summary of Changes

Fixes #


solana-grimes:

PR #17592, opened 2021-05-28
rework test for flexibility

jeffwashington:
#### Problem
These tests are about to be hit by changes to storages, such as sorting and moving slot numbers. The changes are noisy.
#### Summary of Changes
Make the noisy test change separately.
Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -5693,20 +5693,26 @@ pub mod tests {
             },
         ];
 
-        let raw_accounts = vec![
-            AccountSharedData::new(
-                    raw_expected[0].lamports,
-                    1,
-                    AccountSharedData::default().owner(),
-                ),
-            AccountSharedData::new(128, 1, AccountSharedData::default().owner()),
-            AccountSharedData::new(129, 1, AccountSharedData::default().owner()),
-            AccountSharedData::new(256, 1, AccountSharedData::default().owner()),
+        let expected_hashes = vec![
+            Hash::from_str("5K3NW73xFHwgTWVe4LyCg4QfQda8f88uZj2ypDx2kmmH").unwrap(),
+            Hash::from_str("84ozw83MZ8oeSF4hRAg7SeW1Tqs9LMXagX1BrDRjtZEx").unwrap(),
+            Hash::from_str("5XqtnEJ41CG2JWNp7MAg9nxkRUAnyjLxfsKsdrLxQUbC").unwrap(),
+            Hash::from_str("DpvwJcznzwULYh19Zu5CuAA4AT6WTBe4H6n15prATmqj").unwrap(),
         ];
 
+        let mut raw_accounts = Vec::default();
+
         for i in 0..raw_expected.len() {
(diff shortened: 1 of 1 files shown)


runtime/src/accounts_db.rs
```diff
@@ -5653,84 +5653,79 @@ pub mod tests {
...
-        );
-        accounts.add_root(SLOT);
+            AccountSharedData::new(128, 1, AccountSharedData::default().owner()),
+            AccountSharedData::new(129, 1, AccountSharedData::default().owner()),
+            AccountSharedData::new(256, 1, AccountSharedData::default().owner()),
+        ];
+
+        for i in 0..raw_expected.len() {
+            raw_expected[i].hash =
+                AccountsDb::hash_account(slot, &raw_accounts[i], &raw_expected[i].pubkey);
```
brooksprumo, line 5709: Just to confirm, do you have any concerns about replacing the hardcoded Hash values in the tests here?

runtime/src/accounts_db.rs
```diff
@@ -5653,84 +5653,79 @@ pub mod tests {
...
-            )],
-        );
-        accounts.add_root(SLOT);
+            AccountSharedData::new(128, 1, AccountSharedData::default().owner()),
+            AccountSharedData::new(129, 1, AccountSharedData::default().owner()),
+            AccountSharedData::new(256, 1, AccountSharedData::default().owner()),
+        ];
+
+        for i in 0..raw_expected.len() {
+            raw_expected[i].hash =
```
brooksprumo, line 5708: Would this work too?
```rust
for x in &mut raw_expected {
    x.hash =
```

runtime/src/accounts_db.rs
```diff
@@ -5653,84 +5653,79 @@ pub mod tests {
...
+                hash: Hash::default(),
                 lamports: 129,
-                slot: SLOT,
+                slot,
                 pubkey: pubkey128,
             },
             CalculateHashIntermediate {
                 version: 3,
-                hash: Hash::from_str("DpvwJcznzwULYh19Zu5CuAA4AT6WTBe4H6n15prATmqj").unwrap(),
+                hash: Hash::default(),
```
brooksprumo, line 5689: Just to confirm, do you have any reservations about replacing the hardcoded hash values with computed values below?

runtime/src/accounts_db.rs
jeffwashington: I need the 'i' for the other loops. I'm sure there are other ways to do this, but this is just test code.

brooksprumo (commented):
Looks good, just some simple questions I wanted to double check.

runtime/src/accounts_db.rs
jeffwashington: no. I added those constants a few months ago. constants/hash values are a function of slot#.

runtime/src/accounts_db.rs
brooksprumo: Fair, no worries.

brooksprumo (approved):

PR #17601, opened 2021-05-30
Keep track of dirty stores on remove accounts to clean

sakridge:
#### Problem

zero_lamport_keys set grows and slows down clean_accounts which then prevents flushing for many seconds. On mainnet, 11m zero lamport keys are can be present taking 50 seconds to perform the clean operation.

#### Summary of Changes

On `remove_accounts` keep track of the store which is touched and add that to a dirty set. Then in `clean_accounts` use that dirty set to go and look at all keys which are present in the dirty stores and if they can be cleaned.

Reduces the clean_accounts operation down to about 4-5 seconds.

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -9241,6 +9241,9 @@ pub mod tests {
         // Do clean
         accounts.clean_accounts(None, false);
 
+        // 2nd clean needed to clean-up pubkey1
+        accounts.clean_accounts(None, false);
+
         // Ensure pubkey2 is cleaned from the index finally
         assert_not_load_account(&accounts, current_slot, pubkey1);
         assert_load_account(&accounts, current_slot, pubkey2, old_lamport);
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -443,9 +443,9 @@ struct CleanKeyTimings {
     collect_delta_keys_us: u64,
     delta_insert_us: u64,
     hashset_to_vec_us: u64,
-    zero_lamport_key_clone_us: u64,
+    dirty_store_processing_us: u64,
     delta_key_count: u64,
-    zero_lamport_count: u64,
+    dirty_pubkeys_count: u64,
 }
 
 /// Persistent storage structure holding the accounts
@@ -876,6 +876,11 @@ pub struct AccountsDb {
     remove_unrooted_slots_synchronization: RemoveUnrootedSlotsSynchronization,
 
     shrink_ratio: AccountShrinkThreshold,
+
+    /// Set of stores which are recently rooted or had accounts removed
+    /// such that potentially a 0-lamport account update could be present which
+    /// means we can remove the account from the index entirely.
+    dirty_stores: DashMap<(Slot, AppendVecId), Arc<AccountStorageEntry>>,
 }
 
 #[derive(Debug, Default)]
@@ -1320,6 +1325,7 @@ impl Default for AccountsDb {
             is_bank_drop_callback_enabled: AtomicBool::default(),
             remove_unrooted_slots_synchronization: RemoveUnrootedSlotsSynchronization::default(),
             shrink_ratio: AccountShrinkThreshold::default(),
+            dirty_stores: DashMap::default(),
         }
     }
 }
@@ -1643,20 +1649,40 @@ impl AccountsDb {
 
     // Construct a vec of pubkeys for cleaning from:
     //   uncleaned_pubkeys - the delta set of updated pubkeys in rooted slots from the last clean
-    //   zero_lamport_pubkeys - set of all alive pubkeys containing 0-lamport updates
+    //   dirty_stores - set of stores which had accounts removed or recently rooted
     fn construct_candidate_clean_keys(
         &self,
         max_clean_root: Option<Slot>,
         timings: &mut CleanKeyTimings,
     ) -> Vec<Pubkey> {
-        let mut zero_lamport_key_clone = Measure::start("zero_lamport_key");
-        let pubkeys = self.accounts_index.zero_lamport_pubkeys().clone();
-        timings.zero_lamport_count = pubkeys.len() as u64;
-        zero_lamport_key_clone.stop();
-        timings.zero_lamport_key_clone_us += zero_lamport_key_clone.as_us();
+        let mut dirty_store_processing_time = Measure::start("dirty_store_processing");
+        let max_slot = max_clean_root.unwrap_or_else(|| self.accounts_index.max_root());
+        let mut dirty_stores = Vec::with_capacity(self.dirty_stores.len());
+        self.dirty_stores.retain(|(slot, _store_id), store| {
+            if *slot > max_slot {
+                true
+            } else {
+                dirty_stores.push((*slot, store.clone()));
+                false
+            }
+        });
+        let dirty_stores_len = dirty_stores.len();
+        let pubkeys = DashSet::new();
+        for (_slot, store) in dirty_stores {
+            for account in store.accounts.accounts(0) {
+                pubkeys.insert(account.meta.pubkey);
+            }
+        }
+        trace!(
+            "dirty_stores.len: {} pubkeys.len: {}",
+            dirty_stores_len,
+            pubkeys.len()
+        );
+        timings.dirty_pubkeys_count = pubkeys.len() as u64;
+        dirty_store_processing_time.stop();
+        timings.dirty_store_processing_us += dirty_store_processing_time.as_us();
 
         let mut collect_delta_keys = Measure::start("key_create");
-        let max_slot = max_clean_root.unwrap_or_else(|| self.accounts_index.max_root());
         let delta_keys = self.remove_uncleaned_slots_and_collect_pubkeys_up_to_slot(max_slot);
         collect_delta_keys.stop();
         timings.collect_delta_keys_us += collect_delta_keys.as_us();
@@ -1719,16 +1745,6 @@ impl AccountsDb {
                                             self.accounts_index
                                                 .roots_and_ref_count(&locked_entry, max_clean_root),
                                         );
-                                    } else {
-                                        // prune zero_lamport_pubkey set which should contain all 0-lamport
-                                        // keys whether rooted or not. A 0-lamport update may become rooted
-                                        // in the future.
-                                        if !slot_list
-                                            .iter()
-                                            .any(|(_slot, account_info)| account_info.lamports == 0)
-                                        {
-                                            self.accounts_index.remove_zero_lamport_key(pubkey);
-                                        }
                                     }
                                     // Release the lock
                                     let slot = *slot;
@@ -1752,11 +1768,7 @@ impl AccountsDb {
                                     // touched in must be unrooted.
                                     purges_old_accounts.push(*pubkey);
                                 }
-                                AccountIndexGetResult::Missing(lock) => {
-                                    // pubkey is missing from index, so remove from zero_lamports_list
-                                    self.accounts_index.remove_zero_lamport_key(pubkey);
-                                    drop(lock);
-                                }
+                                AccountIndexGetResult::Missing(_lock) => {}
                             };
                         }
                         (purges_zero_lamports, purges_old_accounts)
@@ -1898,8 +1910,8 @@ impl AccountsDb {
                 i64
             ),
             (
-                "zero_lamport_key_clone_us",
-                key_timings.zero_lamport_key_clone_us,
+                "dirty_store_processing_us",
+                key_timings.dirty_store_processing_us,
                 i64
             ),
             ("accounts_scan", accounts_scan.as_us() as i64, i64),
@@ -1909,7 +1921,7 @@ impl AccountsDb {
             ("calc_deps", calc_deps_time.as_us() as i64, i64),
             ("reclaims", reclaims_time.as_us() as i64, i64),
             ("delta_key_count", key_timings.delta_key_count, i64),
-            ("zero_lamport_count", key_timings.zero_lamport_count, i64),
+            ("dirty_pubkeys_count", key_timings.dirty_pubkeys_count, i64),
             ("total_keys_count", total_keys_count, i64),
         );
     }
@@ -5118,6 +5130,8 @@ impl AccountsDb {
                     store.slot(), *slot
                 );
                 let count = store.remove_account(account_info.stored_size, reset_accounts);
+                self.dirty_stores
+                    .insert((*slot, store.append_vec_id()), store.clone());
                 if count == 0 {
                     dead_slots.insert(*slot);
                 } else if self.caching_enabled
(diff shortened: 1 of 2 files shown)


sakridge:
This allowed me to run a validator on `merckx` with 64gb of ram. Otherwise the clean prevents flush and the accounts cache runs the machine out of memory. Overall size of accounts doesn't seem too affected:

I ran it for a couple days:

```
Every 5.0s: du --max-depth 1 -h validator-ledger/accounts/

15G     validator-ledger/accounts/accounts
24G     validator-ledger/accounts/
```

snapshots seem similar:
```
-rw-rw-r--  1 sakridge sakridge 6.7G Jun  3 07:17 snapshot-81235135-DYpphBwbG6Qdw2yBTDxNaB4nCK8JUN7if1oE8doUVeqH.tar.zst
-rw-rw-r--  1 sakridge sakridge 6.7G Jun  3 07:21 snapshot-81235513-HicJFEFAptrrcSbo6FFzTg92yFAJiZyn3RKKACJe6tCX.tar.zst
```

Total memory use around 48GB.

runtime/src/accounts_db.rs
```diff
@@ -1609,20 +1612,44 @@ impl AccountsDb {
...
+            .filter_map(|entry| {
+                if entry.key().0 <= max_slot {
+                    Some((entry.key().0, entry.value().clone()))
+                } else {
+                    None
+                }
+            })
+            .collect();
+        self.dirty_stores
+            .retain(|(slot, _store_id), _store_ref| *slot > max_slot);
```
carllin, line 1635: Can we avoid the `filter_map` iteration above if we add to the vector in the `retain` here?

runtime/src/accounts_db.rs
```diff
@@ -1609,20 +1612,44 @@ impl AccountsDb {
...
     // Construct a vec of pubkeys for cleaning from:
     //   uncleaned_pubkeys - the delta set of updated pubkeys in rooted slots from the last clean
-    //   zero_lamport_pubkeys - set of all alive pubkeys containing 0-lamport updates
+    //   dirty_stores - set of stores which had accounts removed
     fn construct_candidate_clean_keys(
         &self,
         max_clean_root: Option<Slot>,
         timings: &mut CleanKeyTimings,
     ) -> Vec<Pubkey> {
         let mut zero_lamport_key_clone = Measure::start("zero_lamport_key");
```
carllin, line 1621: nit: `zero_lamport_key_clone` -> `dirty_stores_processing`, `zero_lamport_key->dirty_stores_processing`

runtime/src/accounts_db.rs
```diff
@@ -4799,6 +4812,8 @@ impl AccountsDb {
                     store.slot(), *slot
                 );
                 let count = store.remove_account(account_info.stored_size, reset_accounts);
+                self.dirty_stores
+                    .insert((*slot, store.append_vec_id()), store.clone());
```
carllin, line 4816: hmmm it's not immediately obvious to me why removing an account from a storage entry should add the entire store to the `dirty_stores` list.  Maybe a comment would help?

I think the reasoning here is if for example:
1) the storage entry for slot `S` contains some pubkey `P`
2) Pubkey `P` has zero lamports in some slot `S' > S`.

 Then when the store for slot `S` hits `count ==0`, then this could potentially unblock the clean of the zero-lamport `P` in `S'`?

In this case, should we only insert into the dirty stores if the `count == 0` below?

runtime/src/accounts_db.rs
```diff
@@ -1609,20 +1612,44 @@ impl AccountsDb {
 
     // Construct a vec of pubkeys for cleaning from:
     //   uncleaned_pubkeys - the delta set of updated pubkeys in rooted slots from the last clean
-    //   zero_lamport_pubkeys - set of all alive pubkeys containing 0-lamport updates
+    //   dirty_stores - set of stores which had accounts removed
```
carllin, line 1615: Nit: maybe this comment would also be helpful here: https://github.com/solana-labs/solana/pull/17601/files#diff-1090394420d51617f3233275c2b65ed706b35b53b115fe65f82c682af8134a6fR851 when the `dirty_stores` field is declared in the struct.

Also, `dirty_stores` is added to in `add_root`, when accounts haven't technically been "removed" yet: https://github.com/solana-labs/solana/pull/17601/files#diff-1090394420d51617f3233275c2b65ed706b35b53b115fe65f82c682af8134a6fR5340. Maybe a more apt description is something along the lines of

 "set of stores which contain updates to accounts that:

1) Outdate earlier updates to earlier stores
2) Were set to zero-lamports in later stores, and are now potentially free for cleaning"

runtime/src/accounts_db.rs
```diff
@@ -1609,20 +1612,44 @@ impl AccountsDb {
...
     fn construct_candidate_clean_keys(
         &self,
         max_clean_root: Option<Slot>,
         timings: &mut CleanKeyTimings,
     ) -> Vec<Pubkey> {
         let mut zero_lamport_key_clone = Measure::start("zero_lamport_key");
-        let pubkeys = self.accounts_index.zero_lamport_pubkeys().clone();
```
carllin, line 1619: Ha interesting, so most of the time in clean was spent cloning out the 11m zero-lamport keys? I would have thought most of the time was spent in the par scan: https://github.com/solana-labs/solana/blob/9388aaca1539a518c52c7c6d9cd9b99fd044ef21/runtime/src/accounts_db.rs#L1672-L1675 and `clean_accounts_older_than_root` call: https://github.com/solana-labs/solana/blob/9388aaca1539a518c52c7c6d9cd9b99fd044ef21/runtime/src/accounts_db.rs#L1751. Good find!

sakridge: No, it's not the cost of the clone itself. This change reduces the number of keys that we look at to something much smaller than 11 million.

The scan and then the dependency checks where I think where each about 50% of the time, around 10-20 seconds each.

The structure of this change is to reduce the number of keys that is input into this function.

carllin: Ah yup, that makes sense, fewer inputs to the clean.

runtime/src/accounts_db.rs
sakridge: Well it's not just count==0, a shrink which removes an entry could also unblock it. Maybe the logic should be then to add if count==0 and then in shrink add the unref'ed key set there as well. I don't *think* there are any other cases.

sakridge: What if you have a 0-lamport account which is the newest entry for that pubkey in slot `S` for `X`. You also have a non-zero lamport update there for `Y`. Then in slot `S+1` you update `Y` to 0. Then the clean will deref `Y` from `S` and the 0-lamport update for `X` will keep the count=1, although now this store can be removed as long as `X` reference count in storage is 1 or if any other existing updates for it can be also removed.

runtime/src/accounts_db.rs
sakridge: changed

runtime/src/accounts_db.rs
sakridge: yes. updated!

runtime/src/accounts_db.rs
sakridge: Ok. I think that's right. I changed to only inserting when count==0 such that the store is removed. That will free up those keys potentially in newer slots.

carllin: > What if you have a 0-lamport account which is the newest entry for that pubkey in slot `S` for `X`. You also have a non-zero lamport update there for `Y`. Then in slot `S+1` you update `Y` to 0. Then the clean will deref `Y` from `S` and the 0-lamport update for `X` will keep the count=1, although now this store can be removed as long as `X` reference count in storage is 1 or if any other existing updates for it can be also removed.

Hmmm, is this above situation resolved?

So we had this situation:
```
<Slot S>                   <Slot S+1>         
X=0, Y=1                        Y=0
```

1) Slot `S` gets rooted, incurs a clean on `X` and `Y`. Account `X` cannot be removed because the `count` of slot `S`  starts at `2`, gets decremented by 1 for the zero lamport account `X`: https://github.com/solana-labs/solana/blob/43b09c4a24979b0d13f89002534797f815c56891/runtime/src/accounts_db.rs#L1834, but is ultimately not not zero in `calc_delete_dependencies`: https://github.com/solana-labs/solana/blob/43b09c4a24979b0d13f89002534797f815c56891/runtime/src/accounts_db.rs#L1506

2)`S+1` gets rooted and added to the `dirty_stores`, which incurs a clean on account `Y`. dropping the count of slot `S` to 1.  However, `Y` won't be deleted because its ref count is still 2 (due to the update in slot `S`), the ref count is only decremented when either slot `S` is dead or shrunk.

3) In shrink, there's two issues:<\br>
a) We may not shrink `S` because it doesn't meet the page size shrink criteria, even though not performing this shrink prevents the ref count of zero lamport slots from hitting zero, and thus is blocking a lot of zero lamport cleans in later slots.<\br>
b) We shrink slot `S`, but the store still has `count == 1` (the update to X is the latest update to 0), so we don't hit this case: https://github.com/solana-labs/solana/blob/43b09c4a24979b0d13f89002534797f815c56891/runtime/src/accounts_db.rs#L2205-L2207 and we *still* don't add `X` to the clean list.

runtime/src/accounts_db.rs
```diff
@@ -5119,6 +5133,8 @@ impl AccountsDb {
                 );
                 let count = store.remove_account(account_info.stored_size, reset_accounts);
                 if count == 0 {
+                    self.dirty_stores
+                        .insert((*slot, store.append_vec_id()), store.clone());
```
carllin, line 5137: Another annoying case to think about is that right now unrooted slots can also prevent zero lamport cleanup due to them counting toward the `ref_count_from_storage`, but only rooted slots are included in `account_infos.len()` here:  https://github.com/solana-labs/solana/blob/43b09c4a24979b0d13f89002534797f815c56891/runtime/src/accounts_db.rs#L1483. So for example if we had:

```
<Slot S, Rooted>                   <Slot S+1, Unrooted>               <Slot S+2, Rooted>
Y=1                                               Y=2, Z=1                                         Y=0
```

The `ref_count_from_storage` of `Y` is `2`: https://github.com/solana-labs/solana/blob/43b09c4a24979b0d13f89002534797f815c56891/runtime/src/accounts_db.rs#L1812, because even though *rooted* slot `S` is dead, *unrooted* slot `S+1` is alive and adds to the ref count for `Y`. Meanwhile `account_infos` only contains the *rooted* update in slot `S+2`, so this check fails https://github.com/solana-labs/solana/blob/43b09c4a24979b0d13f89002534797f815c56891/runtime/src/accounts_db.rs#L1483.

Luckily, later, when the bank for slot `S+1` is dropped, cleanup happens in `purge_slots()` and calls into `purge_slot_storage()`: https://github.com/solana-labs/solana/blob/43b09c4a24979b0d13f89002534797f815c56891/runtime/src/accounts_db.rs#L3394, which should should ad slot `S+1` to the dirty stores: https://github.com/solana-labs/solana/blob/43b09c4a24979b0d13f89002534797f815c56891/runtime/src/accounts_db.rs#L5136-L5138. 

I think it would be good to add a test here. I think you can copy and paste this one: https://github.com/carllin/solana/commit/c2683cdc983a7a574047d45dbba183b8a4f60009#diff-1090394420d51617f3233275c2b65ed706b35b53b115fe65f82c682af8134a6fR11025-R11077, which tests exactly this case 😃 .

runtime/src/accounts_db.rs
carllin: I wonder if the fix here is to say if in this check here: https://github.com/solana-labs/solana/blob/43b09c4a24979b0d13f89002534797f815c56891/runtime/src/accounts_db.rs#L1483, if `if account_infos.len() as u64 == *ref_count_from_storage == 1`, then we don't need to make the additional check in this else block: https://github.com/solana-labs/solana/blob/43b09c4a24979b0d13f89002534797f815c56891/runtime/src/accounts_db.rs#L1496-L1510 since we know there's no other slots that we are shielding

A test might be good to illustrate this situation here. I think there are similar ones, but the key difference here is we're cleaning after setting a root on S, and then cleaning after setting a root on S+1, instead of cleaning only once after rooting both S and S + 1

runtime/src/accounts_db.rs
sakridge: This test fails on master as well. I think we can do another change to handle this case. It doesn't seem to be related to this change specifically.

runtime/src/accounts_db.rs
carllin: > Doesn't this test do that?

Yeah, this test tests exactly this case, thanks!

>I don't think this is the case. When we shrink, the old store(s) should always have count==0 because all live accounts would be updated in it.

Ah yeah, I misread that bit of logic and thought it would only execute if the newly shrunk store had `count==0`, not the old unshrunken store, which will always have `count == 0` . So this will essentially add all the accounts in the old store for that slot to the candidate clean set.

Is it true that if we shrink unshrunken storage entry `U` to a shrunken storage `S`, we only need to add the zero-lamport pubkeys in `S` to the candidate clean set if the *only* remaining alive accounts in `S` are zero lamport themselves?

i.e. if we declared a 
```
let mut only_zero_lamport_accounts_alive = true;
```

Had a check here: 
https://github.com/solana-labs/solana/blob/43b09c4a24979b0d13f89002534797f815c56891/runtime/src/accounts_db.rs#L2107

```
if stored_account.account.account_meta.lamports != 0 {
     only_zero_lamport_accounts_alive = false;
}
```

And then add to the dirty stores here: https://github.com/solana-labs/solana/blob/43b09c4a24979b0d13f89002534797f815c56891/runtime/src/accounts_db.rs#L2205
```
// Add only the zero lamport accounts to be cleaned
if only_zero_lamport_accounts_alive && store.count() > 0 {
         self.dirty_stores
                            .insert((slot, store.append_vec_id()), store.clone());
}
```

Reasoning being:
1) If the post-shrunk store `S` has any non-zero lamport keys `K`, then when `K` is updated in later slots and `S` becomes dead, these keys will be added to the clean list (so don't need to add them in shrink).
2) If the post-shrunk store `S` *only* has zero lamport keys, then we can run into the edge case so we should add them to clean.

runtime/src/accounts_db.rs
carllin: > This test fails on master as well

Shoot, that's because the test was wrong! 

> It doesn't seem to be related to this change specifically

Yeah it's not directly related, but the removal of the zero-lamport accounts in the candidate key generation had a chance of making the clean fail here: https://github.com/carllin/solana/blob/5bb5284438f0c5a545a4db53924f28e5759a3591/runtime/src/accounts_db.rs#L11670-L11673 if the slot storage wasn't being correctly added to the `dirty_stores` list in the `purge_slots()` path, so it was just an additional corner case to consider.

I extracted the test into a separate PR, which passes now on master and with these changes, so not a blocker here 😃

sakridge: Awesome, thanks!

runtime/src/accounts_db.rs
sakridge: Yea. This could be an optimization on the reducing the clean pubkey set, right? I think that could be a follow-up change. This already has a huge impact on the clean time, and a few extra keys are not that expensive to deal with.

PR #17620, opened 2021-05-31
feat: add decodeUnchecked to borsh-schema.ts

lieuzhenghong:
https://github.com/solana-labs/solana-program-library/pull/1656#discussion_r642598440

Add function `decodeUnchecked` in `borsh-schema.ts` which calls `deserializeUnchecked` under the hood, for use in decoding `ValidatorList` objects.

@joncinque


joncinque:
Thanks for your contribution!  Looks like this is missing a couple of things:

* you'll need to update the commit message to conform to the commit lint https://github.com/conventional-changelog/commitlint/#what-is-commitlint, so something like "feat: add decodeUnchecked to borsh"
* CI is failing because `deserializeUnchecked` isn't actually imported

After that, we should be good to go!

PR #17634, opened 2021-06-01
Purge expired BlockHeight data from blockstore

ryoqun:
#### Problem

block_height cf's data is never removed.

Fortunately, storage impact should be very small.

#### Summary of Changes

Purge it correctly and add reminder comment to avoid future re-occurrence.

follow up to #17523

### DIFF
--- a/ledger/src/blockstore/blockstore_purge.rs
+++ b/ledger/src/blockstore/blockstore_purge.rs
@@ -291,6 +291,10 @@ impl Blockstore {
             && self
                 .perf_samples_cf
                 .compact_range(from_slot, to_slot)
+                .unwrap_or(false)
+            && self
+                .block_height_cf
+                .compact_range(from_slot, to_slot)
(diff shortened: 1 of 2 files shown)


CriesofCarrots (approved):
Oof, thank you!

CriesofCarrots:
Actually, we probably need to add it to `compact_storage()` as well for v1.6, don't we?

ryoqun:
> Actually, we probably need to add it to `compact_storage()` as well for v1.6, don't we?

yeah. lol. I was rushing too. xD

ryoqun:
@CriesofCarrots (I'm going to bed..) could you monitor this while bots back-port to v1.6 & v1.7?

CriesofCarrots:
> @CriesofCarrots (I'm going to bed..) could you monitor this while bots back-port to v1.6 & v1.7?

💯 I'll handle @ryoqun

PR #17635, opened 2021-06-01
skips values which are within push-window when generating pull-responses

behzadnouri:
#### Problem
Gossip pull-response traffic shows a lot of relatively new values
(mostly votes) which by the time they are processed at the receiving
end, they are already available in its crds table (from push messages).

#### Summary of Changes
There is currently some existing logic to skip new values in pull
responses:
https://github.com/solana-labs/solana/blob/a0d721c6b/gossip/src/crds_gossip_pull.rs#L506-L509

This commit expands above logic by skipping all values which are recent
enough that are still pushed between nodes, except for:
* contact-infos: since when starting a validator, it does not receive
  any push-messages until it adopts entrypoint shred version.
* node-instances: since it is best to broadcast duplicate instances
  through all possible channels as soon as possible.

### DIFF
--- a/local-cluster/tests/local_cluster.rs
+++ b/local-cluster/tests/local_cluster.rs
@@ -2544,7 +2544,7 @@ fn do_test_optimistic_confirmation_violation_with_or_without_tower(with_tower: b
     cluster.restart_node(&validator_c_pubkey, validator_c_info);
 
     let mut votes_on_c_fork = std::collections::BTreeSet::new(); // S4 and S5
-    for _ in 0..100 {
+    for _ in 0..900 {
         sleep(Duration::from_millis(100));
 
         if let Some((last_vote, _)) = last_vote_in_tower(&val_c_ledger_path, &validator_c_pubkey) {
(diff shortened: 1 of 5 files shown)


local-cluster/tests/local_cluster.rs
```diff
@@ -2544,7 +2544,7 @@ fn do_test_optimistic_confirmation_violation_with_or_without_tower(with_tower: b
     cluster.restart_node(&validator_c_pubkey, validator_c_info);
 
     let mut votes_on_c_fork = std::collections::BTreeSet::new(); // S4 and S5
-    for _ in 0..100 {
+    for _ in 0..900 {
```
carllin, line 2547: ah shucks, are the `test_no_optimistic_confirmation_violation_with_tower/test_no_optimistic_confirmation_violation_without_tower` still flaky without this change?

carllin (approved):

local-cluster/tests/local_cluster.rs
behzadnouri: not flaky, they just fail. which doesn't seem good (i.e. these shouldn't fail even if votes are not propagated through pull responses at all).

PR #17638, opened 2021-06-01
use slots returned from get_snapshot_storages to sort

jeffwashington:
#### Problem
get_snapshot_storages now returns an array of corresponding slots along with SnapshotStorage items.
#### Summary of Changes
Use the slots vector to facilitate sorting.
Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -4378,17 +4378,22 @@ impl AccountsDb {
         check_hash: bool,
     ) -> Result<(Hash, u64), BankHashVerificationError> {
         if !use_index {
-            let mut time = Measure::start("collect");
-            let combined_maps = self.get_snapshot_storages(slot, Some(ancestors)).0;
-            time.stop();
+            let mut collect_time = Measure::start("collect");
+            let (combined_maps, slots) = self.get_snapshot_storages(slot, Some(ancestors));
+            collect_time.stop();
+
+            let mut sort_time = Measure::start("sort_storages");
+            let storages = SortedStorages::new_with_slots(&combined_maps, &slots);
+            sort_time.stop();
 
             let timings = HashStats {
-                collect_snapshots_us: time.as_us(),
+                collect_snapshots_us: collect_time.as_us(),
+                storage_sort_us: sort_time.as_us(),
                 ..HashStats::default()
             };
 
             Self::calculate_accounts_hash_without_index(
-                &combined_maps,
+                &storages,
                 Some(&self.thread_pool_clean),
                 timings,
                 check_hash,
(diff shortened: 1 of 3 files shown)


jeffwashington:
```
this change:
calculate_accounts_hash_without_index accounts_scan=2047108i eliminate_zeros=450910i hash=245630i sort=1577508i hash_total=61837974i flatten=1608918i storage_sort_us=  870i unreduced_entries=62270664i collect_snapshots_us=47779i num_snapshot_storage=372887i total=5978723i
calculate_accounts_hash_without_index accounts_scan=2169028i eliminate_zeros=442086i hash=233665i sort=1467708i hash_total=61837974i flatten=1852799i storage_sort_us=12971i unreduced_entries=62270664i collect_snapshots_us=64245i num_snapshot_storage=372887i total=6242502i
master^
```

runtime/src/accounts_db.rs
```diff
@@ -4378,17 +4378,22 @@ impl AccountsDb {
...
-                collect_snapshots_us: time.as_us(),
+                collect_snapshots_us: collect_time.as_us(),
+                storage_sort_us: sort_time.as_us(),
                 ..HashStats::default()
             };
 
             Self::calculate_accounts_hash_without_index(
-                &combined_maps,
+                &storages,
                 Some(&self.thread_pool_clean),
```
brooksprumo, line 4397: I know this is unrelated to this PR; I meant to ask about it the other day. What's the guidance for choosing to use `thread_pool_clean` vs `thread_pool`?

runtime/src/snapshot_utils.rs
```diff
@@ -999,8 +1000,9 @@ pub fn process_accounts_package_pre(
 
     let hash = accounts_package.hash; // temporarily remaining here
     if let Some(expected_hash) = accounts_package.hash_for_testing {
+        let sorted_storages = SortedStorages::new(&accounts_package.storages);
```
brooksprumo, line 1003: Is `process_accounts_package_pre()` only called when a testing CLI option is enabled too? Asking because this fn calls regular `SortedStorages::new()`, which has this comment:

https://github.com/solana-labs/solana/blob/cf20c7990d8fdbbf7dcf2d0ebb09e5e23ad1bf00/runtime/src/sorted_storages.rs#L35-L39

runtime/src/sorted_storages.rs
```diff
@@ -7,7 +7,7 @@ use std::ops::Range;
 pub struct SortedStorages<'a> {
     range: Range<Slot>,
     storages: Vec<Option<&'a SnapshotStorage>>,
-    count: usize,
+    slot_count: usize,
```
brooksprumo, line 10: [Amount I care: 1/5] Wondering what prompted the rename to `slot_count`? Since the name of the struct is about storages, the count intuitively felt like a count of storages to me.

I know the number of slots and storages are the same, so no question in that regard.

brooksprumo (commented):
All looks good; just a few questions

runtime/src/accounts_db.rs
jeffwashington: the clean thread pool was designed to be used/shared by 'clean' type accounts bg operations, afaik.

runtime/src/snapshot_utils.rs
jeffwashington: No - the function is now being used in a place where the comment doesn't apply. I need to fix that.

runtime/src/sorted_storages.rs
jeffwashington: this is only used for returning a metric. The metric used to count the # of append vecs across all slots. Assuming we have only 1, we now only count the # of slots with non-empty values, which will return an equivalent # practically, but means something different.

brooksprumo (approved):

PR #17647, opened 2021-06-01
parallelizes gossip packets receiver with processing of requests

behzadnouri:
#### Problem
Gossip packet processing is composed of two stages:
  * The first is consuming packets from the socket, deserializing,
    sanitizing and verifying them:
    https://github.com/solana-labs/solana/blob/7f0349b29/gossip/src/cluster_info.rs#L2510-L2521
  * The second is actually processing the requests/messages:
    https://github.com/solana-labs/solana/blob/7f0349b29/gossip/src/cluster_info.rs#L2585-L2605

The former does not acquire any locks and so can be parallelized with
the later, allowing better pipelineing properties and smaller latency in
responding to gossip requests or propagating messages.

#### Summary of Changes
Added a new stage to gossip consuming packets received from the socket,
deserializing, sanitizing and verifying them and then sending them down the
channel for the actual handling of requests/messages.

### DIFF
--- a/gossip/src/cluster_info.rs
+++ b/gossip/src/cluster_info.rs
@@ -78,6 +78,7 @@ use {
         result::Result,
         sync::{
             atomic::{AtomicBool, Ordering},
+            mpsc::{Receiver, RecvTimeoutError, Sender},
             {Arc, Mutex, RwLock, RwLockReadGuard, RwLockWriteGuard},
         },
         thread::{sleep, Builder, JoinHandle},
@@ -235,7 +236,7 @@ impl Default for ClusterInfo {
 }
 
 #[derive(Clone, Debug, Default, Deserialize, Serialize, AbiExample)]
-struct PruneData {
+pub(crate) struct PruneData {
     /// Pubkey of the node that sent this prune data
     pubkey: Pubkey,
     /// Pubkeys of nodes that should be pruned
@@ -329,7 +330,7 @@ pub(crate) type Ping = ping_pong::Ping<[u8; GOSSIP_PING_TOKEN_SIZE]>;
 #[frozen_abi(digest = "GANv3KVkTYF84kmg1bAuWEZd9MaiYzPquuu13hup3379")]
 #[derive(Serialize, Deserialize, Debug, AbiEnumVisitor, AbiExample)]
 #[allow(clippy::large_enum_variant)]
-enum Protocol {
+pub(crate) enum Protocol {
     /// Gossip protocol messages
     PullRequest(CrdsFilter, CrdsValue),
     PullResponse(Pubkey, Vec<CrdsValue>),
@@ -2494,7 +2495,7 @@ impl ClusterInfo {
 
     fn process_packets(
         &self,
-        packets: VecDeque<Packet>,
+        packets: VecDeque<(/*from:*/ SocketAddr, Protocol)>,
         thread_pool: &ThreadPool,
         recycler: &PacketsRecycler,
         response_sender: &PacketSender,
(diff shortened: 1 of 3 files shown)


gossip/src/cluster_info.rs
```diff
@@ -2659,10 +2683,35 @@ impl ClusterInfo {
         Ok(())
     }
 
-    pub fn listen(
+    pub(crate) fn consume(
```
carllin, line 2686: nit: functions in the code base that start threads generally start with `start_*_thread`, how about `start_socket_consume_thread()`

gossip/src/cluster_info.rs
```diff
@@ -2606,23 +2589,64 @@ impl ClusterInfo {
         Ok(())
     }
 
+    // Consumes packets received from the socket, deserializing, sanitizing and
+    // verifying them and then sending them down the channel for the actual
+    // handling of requests/messages.
+    fn run_consume(
```
carllin, line 2595: nit: rename to `run_socket_consume` to be more specific

gossip/src/cluster_info.rs
```diff
@@ -2659,10 +2683,35 @@ impl ClusterInfo {
...
+            while !exit.load(Ordering::Relaxed) {
+                match self.run_consume(&receiver, &sender, &thread_pool) {
+                    Err(GossipError::RecvTimeoutError(RecvTimeoutError::Disconnected)) => break,
+                    Err(GossipError::RecvTimeoutError(RecvTimeoutError::Timeout)) => (),
+                    Err(err) => error!("gossip consume: {}", err),
+                    Ok(()) => (),
+                }
+            }
+        };
+        let thread_name = String::from("gossip-consume");
```
carllin, line 2707: nit: "gossip-consume" -> "gossip-socket-consume"

gossip/src/gossip_service.rs
```diff
@@ -56,11 +56,15 @@ impl GossipService {
             1,
         );
         let (response_sender, response_receiver) = channel();
-        let t_responder = streamer::responder("gossip", gossip_socket, response_receiver);
+        let (consume_sender, listen_receiver) = channel();
+        let t_consume =
```
carllin, line 60: nit `t_consume -> t_socket_consume`

gossip/src/cluster_info.rs
```diff
@@ -2659,10 +2683,35 @@ impl ClusterInfo {
...
+            .num_threads(get_thread_count().min(8))
+            .thread_name(|i| format!("gossip-consume-{}", i))
+            .build()
+            .unwrap();
+        let run_consume = move || {
+            while !exit.load(Ordering::Relaxed) {
+                match self.run_consume(&receiver, &sender, &thread_pool) {
+                    Err(GossipError::RecvTimeoutError(RecvTimeoutError::Disconnected)) => break,
+                    Err(GossipError::RecvTimeoutError(RecvTimeoutError::Timeout)) => (),
+                    Err(err) => error!("gossip consume: {}", err),
```
carllin, line 2702: Nice, the only other potential error other than the Timeout is `Ok(sender.send(packets)?)`, https://github.com/solana-labs/solana/pull/17647/files#diff-b07406ad1c913d21d39a4163cb34f48c55cef6d472c814419a732f3937ffceddR2631. Hopefully this one doesn't have a chance of being too spammy?

carllin (dismissed):

gossip/src/cluster_info.rs
behzadnouri: done

gossip/src/cluster_info.rs
behzadnouri: done

gossip/src/cluster_info.rs
behzadnouri: yeah, I guess should be fine.

gossip/src/cluster_info.rs
behzadnouri: thread names in linux are limited to 15 characters, and everything after the 15th character is trimmed when inspecting running processes from shell, which will make it confusing if I use a longer name here.

gossip/src/gossip_service.rs
behzadnouri: done

PR #17661, opened 2021-06-02
Fix getBlock API reference

oncecsc:
#### Problem
In solana-core v1.6 not containt getBlock method.
#### Summary of Changes

change  
`NEW: This method is only available in solana-core v1.7 or newer. Please use
[getBlock](jsonrpc-api.md#getblock) for solana-core v1.6 `
to 
`NEW: This method is only available in solana-core v1.7 or newer. Please use
[getConfirmedBlock](jsonrpc-api.md#getconfirmedblock) for solana-core v1.6`


mvines:
Thanks!

PR #17685, opened 2021-06-02
rework hash calculation to not keep slot and write version

jeffwashington:
#### Problem
It is constant time work to sort 400k storages by slot as opposed to n*lg(n)? [n=# accounts] work to sort account data by slot later.
Also, dragging around Slot and write_version per account can get expensive.
#### Summary of Changes
Take advantage of the fact that storages are sorted by slot to reduce the data we keep and eventually speed up hash and lamport calculation overall. Note that this isn't faster overall yet. So, this pr likely won't be actually submitted until the next phases and supporting pieces are ready. But, this is a logical unit of work. And, it isn't a serious regression and uses less memory, so there are some tradeoffs. With different bin tuning, this algorithm can be as fast or faster as-is.
We eliminate a write version and a slot for each account entry we find (keep in mind there are duplicates).

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -4335,6 +4335,80 @@ impl AccountsDb {
         self.update_accounts_hash_with_index_option(true, true, slot, ancestors, None)
     }
 
+    fn scan_multiple_account_storages_one_slot<F, B>(
+        storages: &[Arc<AccountStorageEntry>],
+        scan_func: &F,
+        slot: Slot,
+        retval: &mut B,
+    ) where
+        F: Fn(LoadedAccount, &mut B, Slot) + Send + Sync,
+        B: Send + Default,
+    {
+        // we have to call the scan_func in order of write_version within a slot if there are multiple storages per slot
+        let mut len = storages.len();
+        let mut progress = Vec::with_capacity(len);
+        let mut current = Vec::with_capacity(len);
+        for storage in storages {
+            let accounts = storage.accounts.accounts(0);
+            let mut iterator: std::vec::IntoIter<StoredAccountMeta<'_>> = accounts.into_iter();
+            if let Some(item) = iterator
+                .next()
+                .map(|stored_account| (stored_account.meta.write_version, Some(stored_account)))
+            {
+                current.push(item);
+                progress.push(iterator);
+            }
+        }
+        while !progress.is_empty() {
+            let mut min = current[0].0;
+            let mut min_index = 0;
+            for (i, (item, _)) in current.iter().enumerate().take(len).skip(1) {
+                if item < &min {
+                    min_index = i;
+                    min = *item;
+                }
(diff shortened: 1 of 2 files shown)


jeffwashington:
```
this change:
calculate_accounts_hash_without_index accounts_scan=1984282i eliminate_zeros=400528i hash=231545i sort=2685521i hash_total=61837974i flatten=1174477i storage_sort_us=20538i unreduced_entries=62270664i collect_snapshots_us=214111i num_snapshot_storage=372887i total=6711002i
calculate_accounts_hash_without_index accounts_scan=2209913i eliminate_zeros=443076i hash=233774i sort=1473023i hash_total=61837974i flatten=1490250i storage_sort_us=20576i unreduced_entries=62270664i collect_snapshots_us=228110i num_snapshot_storage=372887i total=6098722i
master^
```

runtime/src/accounts_hash.rs
```diff
@@ -667,29 +668,43 @@ impl AccountsHash {
...
-            let mut look_for_first_key = !is_first_slice;
             'outer: loop {
                 // at start of loop, item at 'i' is the first entry for a given pubkey - unless look_for_first
-                let now = &slice[i];
-                let last = now.pubkey;
-                if !look_for_first_key && now.lamports != ZERO_RAW_LAMPORTS_SENTINEL {
-                    // first entry for this key that starts in our slice
-                    result.push(now.hash);
-                    sum += now.lamports as u128;
+                let mut now = &slice[i];
```
jeffwashington, line 679: The main change here is that we need to find the last entry for each pubkey. The previous code wanted the first item. Since we read directly from appendvecs and write_versions increase as we read, this is the natural sort order.

runtime/src/accounts_db.rs
```diff
@@ -4357,11 +4357,63 @@ impl AccountsDb {
...
+                                accounts.into_iter().for_each(|stored_account| {
+                                    scan_func(
+                                        LoadedAccount::Stored(stored_account),
+                                        &mut retval,
+                                        slot,
+                                    )
+                                });
+                            }
+                        } else {
+                            // we have to call the scan_func in order of write_version within a slot if there are multiple storages per slot
```
jeffwashington, line 4373: this messy code is required if we have > 1 append vec per slot. We need to write_version to be in increasing order in the final accumulated results.

jeffwashington:
Building off this pr, the next step gets us from 6s (master) to 3.6s by reworking flatten and sort.

sakridge (dismissed):
looks good

jeffwashington:
it turns out there is no discernible penalty for using the multiple storages/slot code path all the time, so that makes things simpler.
Also, there is enough noise in the measurements that it is approximately the same speed for this version of hash calculation vs keeping slot and write version. Perhaps it because the # of bins have increased in master.

PR #17690, opened 2021-06-02
assert that we only have 1 storage per slot

jeffwashington:
#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -4356,6 +4356,7 @@ impl AccountsDb {
                 for slot in start..end {
                     let sub_storages = snapshot_storages.get(slot);
                     if let Some(sub_storages) = sub_storages {
+                        assert!(sub_storages.len() == 1);
                         for storage in sub_storages {
                             let slot = storage.slot();
                             let accounts = storage.accounts.accounts(0);


jeffwashington:
@sakridge , @carllin , @brooksprumo 
This is where I have to make a decision whether to bifurcate and implement 2 mechanisms for iterating storages or whether I can assert that we only have 1 storage per slot. Carl and I discussed this a while on slack. Here, it is important that I return items from all the storages in this slot in write_version order. If there is only 1 storage, that works fine. If there are 2, then I have to interleave them to return items in write_version order. Maybe that isn't too bad and I should just do that. I don't want to chase assert failures. Note this specific code in master doesn't care if write_versions are out of order. Code that I'm working on does require that assumption.

brooksprumo:
> This is where I have to make a decision whether to bifurcate and implement 2 mechanisms for iterating storages or whether I can assert that we only have 1 storage per slot. Carl and I discussed this a while on slack. Here, it is important that I return items from all the storages in this slot in write_version order. If there is only 1 storage, that works fine. If there are 2, then I have to interleave them to return items in write_version order. Maybe that isn't too bad and I should just do that. I don't want to chase assert failures. Note this specific code in master doesn't care if write_versions are out of order. Code that I'm working on does require that assumption.

If the AccountsCache is enabled, does that effectively mean there will only be one storage here? Then does that mean if the AccountsCache is disabled that there are one-or-more storages?

jeffwashington:
Here is my first pass at the path to use when we have > 1 append vec per slot:
https://github.com/solana-labs/solana/blob/26856594ef5a0aa619a08695e203189285e8f81c/runtime/src/accounts_db.rs#L4372-L4417

PR #17691, opened 2021-06-02
remove slot lookup

jeffwashington:
#### Problem
no reason to ask storage for slot when we know it. getting slot is expensive.
#### Summary of Changes
remove the call.
Fixes #


brooksprumo (approved):
Nice find

PR #17698, opened 2021-06-03
Create solana-poh and move remaining rpc modules to solana-rpc

CriesofCarrots:
#### Problem
Some rpc modules still live in core, separate from `solana-rpc`. A few dependencies still prevent detangling these modules.

#### Summary of Changes
- Create `solana-poh` and move PoH-related modules and benches to it
- Move several services plus ValidatorExit into `solana-rpc`
- Move remaining rpc modules out of core


CriesofCarrots:
Ready for a look.
@mvines it's tempting to just move the poh stuff into `solana-ledger`, since there is a `poh` module and a `Poh` struct entangled there.

mvines:
> it's tempting to just move the poh stuff into `solana-ledger`, since there is a `poh` module and a `Poh` struct entangled there.

I don't think I'd feel good about putting the PohService in ledger/

`ValidatorExit` definitely don't sit right in `rpc/` now that I'm looking at it.  Not the worst for now I guess. `ValidatorExit` is really just a general `CoordinatedExit` facility, not validator specific.  We could drop Validator from its name and then shove it in, maybe runtime/ or even sdk/src/.  It'll need to exist even once we can build a validator without `rpc/` included.

bigtable_upload_service.rs doesn't feel great in `rpc/` either.  `ledger/` actually seems like a pretty good fit!

CriesofCarrots:
> `ValidatorExit` definitely don't sit right in `rpc/` now that I'm looking at it. Not the worst for now I guess. `ValidatorExit` is really just a general `CoordinatedExit` facility, not validator specific. We could drop Validator from its name and then shove it in, maybe runtime/ or even sdk/src/. It'll need to exist even once we can build a validator without `rpc/` included.

Oh, excellent. I was eyeballing sdk for it, but the `Validator` aspect held me back. I'll PR that separately and rebase this on top, since it will be a bunch of churn.

> bigtable_upload_service.rs doesn't feel great in `rpc/` either. `ledger/` actually seems like a pretty good fit!

Yep, I agree. Good suggestion!

CriesofCarrots:
Rebased & updated

mvines (approved):
lgtm

PR #17700, opened 2021-06-03
Bump compute budget for neon evm

jackcmay:
#### Problem

The Neon EVM requires a larger compute budget than the default

#### Summary of Changes

Bump the max compute units and the heap size for the Neon EVM program.  These are stop-gap changes that eventually will tie into tx-wide compute caps and optional fee-based compute budget bumps.

Fixes #


jackcmay:
@s-medvedev do you still need the heap bumped?

s-medvedev:
> @s-medvedev do you still need the heap bumped?

Yes, we need to increase the heap size for the contract to work. Uniswap tests show that 256 KB will be sufficient.

PR #17713, opened 2021-06-03
Combine repeated logic for checking if slot has been cleaned up

steviez:
#### Problem
The logic to check if a slot had been cleaned up was repeated several times. Additionally, a first glance at the code may not have registered that `lowest_cleanup_slot` is a read lock

#### Summary of Changes
Combine repeated code into a function (no functional changes)

### DIFF
--- a/ledger/src/blockstore.rs
+++ b/ledger/src/blockstore.rs
@@ -1523,12 +1523,7 @@ impl Blockstore {
         to_index: u64,
         buffer: &mut [u8],
     ) -> Result<(u64, usize)> {
-        // lowest_cleanup_slot is the last slot that was not cleaned up by
-        // LedgerCleanupService
-        let lowest_cleanup_slot = self.lowest_cleanup_slot.read().unwrap();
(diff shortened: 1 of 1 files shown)


jeffwashington (approved):
love it. lgtm.

ledger/src/blockstore.rs
```diff
@@ -2056,6 +2037,17 @@ impl Blockstore {
         Ok(())
     }
 
+    fn check_lowest_cleanup_slot(&self, slot: Slot) -> Result<std::sync::RwLockReadGuard<Slot>> {
+        // lowest_cleanup_slot is the last slot that was not cleaned up by LedgerCleanupService
+        let lowest_cleanup_slot = self.lowest_cleanup_slot.read().unwrap();
+        if *lowest_cleanup_slot > 0 && *lowest_cleanup_slot >= slot {
+            return Err(BlockstoreError::SlotCleanedUp);
+        }
+        // Make caller hold this lock properly; otherwise LedgerCleanupService can purge/compact
```
jeffwashington, line 2046: good catch

PR #17714, opened 2021-06-03
adds fallback logic if retransmit multicast fails

behzadnouri:
#### Problem
In retransmit-stage, based on the `packet.meta.seed` and resulting
children/neighbors, each packet is sent to a different set of peers:
https://github.com/solana-labs/solana/blob/708bbcb00/core/src/retransmit_stage.rs#L421-L457

However, current code errors out as soon as a multicast call fails,
which will skip all the remaining packets:
https://github.com/solana-labs/solana/blob/708bbcb00/core/src/retransmit_stage.rs#L467-L470

This can exacerbate packets loss in turbine.

#### Summary of Changes
This commit:
  * keeps iterating over retransmit packets for loop even if some
    intermediate sends fail.
  * adds a fallback to `UdpSocket::send_to` if `multicast` fails.

Recent discord chat:
https://discord.com/channels/428295358100013066/689412830075551748/849530845052403733

### DIFF
--- a/gossip/src/cluster_info.rs
+++ b/gossip/src/cluster_info.rs
@@ -1382,12 +1382,7 @@ impl ClusterInfo {
     /// retransmit messages to a list of nodes
     /// # Remarks
     /// We need to avoid having obj locked while doing a io, such as the `send_to`
-    pub fn retransmit_to(
-        peers: &[&ContactInfo],
-        packet: &Packet,
-        s: &UdpSocket,
-        forwarded: bool,
-    ) -> Result<(), GossipError> {
+    pub fn retransmit_to(peers: &[&ContactInfo], packet: &Packet, s: &UdpSocket, forwarded: bool) {
         trace!("retransmit orders {}", peers.len());
         let dests: Vec<_> = if forwarded {
             peers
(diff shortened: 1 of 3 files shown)


gossip/src/cluster_info.rs
```diff
@@ -1398,22 +1393,28 @@ impl ClusterInfo {
...
-                        1
-                    );
-                    error!("retransmit result {:?}", e);
-                    return Err(GossipError::Io(e));
+        let mut dests = &dests[..];
+        let data = &packet.data[..packet.meta.size];
+        while !dests.is_empty() {
+            match multicast(s, data, dests) {
+                Ok(n) => dests = &dests[n..],
+                Err(err) => {
```
carllin, line 1401: Out of curiosity, what types of errors was this multicast failing with?

gossip/src/cluster_info.rs
```diff
@@ -1398,22 +1393,28 @@ impl ClusterInfo {
...
+                    inc_new_counter_error!("cluster_info-retransmit-send_to_error", dests.len(), 1);
```
carllin, line 1402: nit: maybe this error should be called `cluster_info-retransmit-multicast_error`, and the `cluster_info-retransmit-error` error below should be called `"cluster_info-retransmit-send_to_error"?

carllin (approved):

gossip/src/cluster_info.rs
behzadnouri: I linked to the discord chat in the description:

It says: 
https://discord.com/channels/428295358100013066/689412830075551748/849530845052403733
```
ERROR solana_core::cluster_info] retransmit result Os { code: 1, kind: PermissionDenied, message: "Operation not permitted" }
```

gossip/src/cluster_info.rs
behzadnouri: I agree that the error names are misleading, but I intentionally didn't want to change them so that in the period of time the cluster is upgrading to the new code we can track the error rates here and compare it with before.
For example, if half of the cluster upgrades to this new code, and the other half doesn't, then changes in the metric would be very confusing if the metric name has changed as well.

I would rather update the metrics name once the cluster has already updated to this new code and the code seems to be working fine.

Issue #6672, opened 2019-11-01
Validators should not make decisions based on which ports data was received


sagar-solana:
#### Problem

Validators make many shred decisions based on which port it was received on. 
This is not secure at all. This issue is related to the regular DoS that is possible on any of a validator's ports. However it leaves the window open for an amplification attack. 

#### Proposed Solution

- For turbine Shreds, validators should do a reverse lookup to see if the Shred _should_ have arrived from where it did (i.e paired node in the layer above). If it did not match up then it shouldn't be sent through Turbine 
- For Repair Shreds, validators should only retransmit them if they came from the layer above. This can be coupled with the regular turbine shred described above. 
- For Repairman, validators should never send these shreds through turbine. No way to detect this right now. Needs a solution. 
- For all shreds received on the window, they should only be retransmitted if they have not already been inserted/recv'd

@aeyakovenko

aeyakovenko:
Why isn’t the rule simply:

* retransmit once and only once if shred is from layer above.

Regardless how it arrived?  Where the leader is the layer above per node per shred.  It shouldn't matter how it arrived.

sagar-solana:
Yeah that applies to everything except for Repairman. So Turbine doesn't care about where (in terms of slot) your peers are. Repairman is sending really old shreds (from the epoch before) to you and you should definitely not be sending them through turbine. Because your peers might not be as far behind as you are.

sagar-solana:
So we need a solution for repairman. Something that isn't port dependent. 

Also We also send shreds received from neighbors to the children in layers below. So the rules should really be:
- Retransmit each shred just once. No matter where you received it from. This will involve a blocktree lookup. Layer above or below doesn't matter if you look at blocktree first. But this will have a performance impact (might need a cache)
- In addition to that, the scope of the retransmit will depend on whether the shred was destined to arrive from the layer above. If it was supposed to come from the layer above (doesn't matter if you repaired it) you should retransmit it neighbors and children. If it wasn't supposed to be delivered from the layer above, only send it to children.

aeyakovenko:
@sagar-solana isn’t repairman request/response?

sagar-solana:
@aeyakovenko no its not. It's automatic. Infact I just realized it repairs anyone more than 50 slots behind. Not an epoch anymore. So repairman looks at others' gossip and catches them up. No request. Directly response.

aeyakovenko:
[...]

PR #17716, opened 2021-06-03
removes port-based forwarding logic from turbine retransmit

behzadnouri:
#### Problem
Turbine retransmit logic is based on which socket it received the packet
from (i.e `packet.meta.forward`):
https://github.com/solana-labs/solana/blob/708bbcb00/core/src/retransmit_stage.rs#L467-L470

This can leave the cluster vulnerable to spoofing and selective
propagation of packets; see
https://github.com/solana-labs/solana/issues/6672
https://github.com/solana-labs/solana/pull/7774

#### Summary of Changes
This commit identifies if the node is on the "critical path" based on
its index in the shuffled cluster. If so, it forwards the packet to both
neighbors and children; otherwise, the packet is only forwarded to the
children.

The metrics added in
https://github.com/solana-labs/solana/pull/17351
shows that the number of times the index does not match the port is very
rare, and therefore this change should be safe.

### DIFF
--- a/core/src/retransmit_stage.rs
+++ b/core/src/retransmit_stage.rs
@@ -428,7 +428,9 @@ fn retransmit(
         // neighborhood), then we expect that the packet arrives at tvu socket
         // as opposed to tvu-forwards. If this is not the case, then the
         // turbine broadcast/retransmit tree is mismatched across nodes.
-        if packet.meta.forward == (my_index % DATA_PLANE_FANOUT == 0) {
+        let anchor_node = my_index % DATA_PLANE_FANOUT == 0;
+        if packet.meta.forward == anchor_node {
+            // TODO: Consider forwarding the packet to the root node here.
             retransmit_tree_mismatch += 1;
         }
         peers_len = cmp::max(peers_len, shuffled_stakes_and_index.len());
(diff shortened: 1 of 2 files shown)


behzadnouri:
`retransmit-stage::retransmit_tree_mismatch`:
https://github.com/solana-labs/solana/blob/7b58dcac1/core/src/retransmit_stage.rs#L427-L433

TDS:
tds-mismatch

MB:
mb-mismatch

carllin:
cc @jbiseda, here we are moving to computing whether we should retransmit based on the shred shuffle instead of the port

core/src/retransmit_stage.rs
```diff
@@ -428,7 +428,9 @@ fn retransmit(
         // neighborhood), then we expect that the packet arrives at tvu socket
         // as opposed to tvu-forwards. If this is not the case, then the
         // turbine broadcast/retransmit tree is mismatched across nodes.
-        if packet.meta.forward == (my_index % DATA_PLANE_FANOUT == 0) {
+        let anchor_node = my_index % DATA_PLANE_FANOUT == 0;
```
carllin, line 431: nit: `anchor_node` -> `is_on_critical_path`

core/src/retransmit_stage.rs
```diff
@@ -464,10 +466,19 @@ fn retransmit(
...
+        if anchor_node {
+            ClusterInfo::retransmit_to(&neighbors, packet, sock, /*forward socket=*/ true);
         }
-        ClusterInfo::retransmit_to(&children, packet, sock, packet.meta.forward);
+        ClusterInfo::retransmit_to(
+            &children,
+            packet,
+            sock,
+            !anchor_node, // send to forward socket!
+        );
```
carllin, line 481: nit: For clarity, how about extracting out `should_send_to_forward_socket = !anchor_node`, and using that for both calls to `retransmit_to`

carllin (approved):

core/src/retransmit_stage.rs
behzadnouri: I am using the same terminology as in here:
https://github.com/solana-labs/solana/blob/c2191d885/gossip/src/cluster_info.rs#L2857

core/src/retransmit_stage.rs
behzadnouri: The part which decides to send to forward socket or tvu socket is effectively redundant going forward. It only impacts that `retransmit_tree_mismatch` metric, otherwise is not anywhere in retransmit logic.

So will probably be deleted soon, and I am leaning on brevity here. I think the comments already describe the logic.

PR #17718, opened 2021-06-03
replace all index hash scans with non-index hash scans

jeffwashington:
#### Problem
Simplify code paths, reduce contention for account index, speed up startup.
#### Summary of Changes
Replace all index hash scans with non-index hash scans.
Fixes #


jeffwashington:
replaced by another

Issue #17711, opened 2021-06-03
Later updates to EpochSlots can overwrite earlier updates


carllin:
#### Problem
We currently track 255 EpochSlots bitvectors that are updated via `push_epoch_slots(slots)` where we:

1) Read out all the individually indexed EpochSlots here: https://github.com/solana-labs/solana/blob/0b0b9d91a836ca0c1054d760fe64dd1a73a91e2b/gossip/src/cluster_info.rs#L907
2) Update the appropriate EpochSlots, then push them to the pending push queue https://github.com/solana-labs/solana/blob/0b0b9d91a836ca0c1054d760fe64dd1a73a91e2b/gossip/src/cluster_info.rs#L948

An issue arises when there are two updates to the same indexed EpochSlots structure (say index 0), before the asynchronous flush of the pending flush queue happens in the gossip thread, so the updated value is not yet observable in the `Crds` table. 

In this case, the second update will read out a stale `EpochSlots` from the `Crds` table in step 1) above that doesn't include the first update, updates that value with the second update, then pushes that item to the pending flush queue. Then when the push happens later, the later update doesn't have the first update, thereby overwriting the first update when both values are pushed to the network.

We could probably simulate the problem in this test if we got rid of the calls to `flush_push_queue()` in between the updates https://github.com/solana-labs/solana/blob/0b0b9d91a836ca0c1054d760fe64dd1a73a91e2b/gossip/src/cluster_info.rs#L4136-L4137.

#### Proposed Solution
Should we also check the push queue for more recent versions of the same message label in `crds.get()`?

PR #17719, opened 2021-06-03
writes epoch-slots to crds table synchronously

behzadnouri:
#### Problem
epoch-slots may be overwritten before they are written to crds table:
https://github.com/solana-labs/solana/issues/17711

#### Summary of Changes
This commit writes new epoch-slots to crds table synchronously with
push_epoch_slots. The functions is still not thread-safe as commented in
the code, however currently only one threads is invoking this code.

### DIFF
--- a/gossip/src/cluster_info.rs
+++ b/gossip/src/cluster_info.rs
@@ -3781,7 +3781,6 @@ mod tests {
         let slots = cluster_info.get_epoch_slots(&mut Cursor::default());
         assert!(slots.is_empty());
         cluster_info.push_epoch_slots(&[0]);
-        cluster_info.flush_push_queue();
 
         let mut cursor = Cursor::default();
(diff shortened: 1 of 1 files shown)


carllin:
Awesome, thanks, this unblocks me 😃 

Can we also remove the flushes in this test here: https://github.com/solana-labs/solana/blob/master/gossip/src/cluster_info.rs#L4137-L4139 to enforce this guarantee that sequential writes don't overwrite each other even without an intermediary flush?

carllin (dismissed):

behzadnouri:
> Can we also remove the flushes in this test here: https://github.com/solana-labs/solana/blob/master/gossip/src/cluster_info.rs#L4137-L4139 (the test passes now without those flushes) to enforce this guarantee that sequential writes don't overwrite each other even without an intermediary flush?

good point, done

PR #17720, opened 2021-06-03
Add ecrecover syscall

s-medvedev:
#### Problem
The EVM implementation requires an `ecrecover` function, but its computation in the BPF code requires too many instructions (6 million).

#### Summary of Changes
Implement an `ecrecover` system call.

Fixes #


s-medvedev:
@jackcmay Please review `ecrecover` system call implementation. The value of `ecrecover_base_cost` will be specified soon.

programs/bpf_loader/src/syscalls.rs
```diff
@@ -1329,6 +1343,79 @@ impl<'a> SyscallObject<BpfError> for SyscallMemset<'a> {
...
+        );
+        let ecrecover_result = question_mark!(
+            translate_slice_mut::<u8>(memory_mapping, result_addr, 64u64, self.loader_id, true,),
+            result
+        );
+
+        let message = match libsecp256k1::Message::parse_slice(hash) {
+            Ok(msg) => msg,
+            Err(_) => {
+                *result = Ok(1);
```
jackcmay, line 1387: I'm assuming that "1" here translates into something like "Parse error".  Can we define error types in program space for these for both Rust and C?

jackcmay:
How about program space binding for both C and Rust?

Tests?

jackcmay:
For CI it looks like not all the `Cargo.lock` file changes were commited

seanyoung:
I'm not sure it should be called ecrecover, since that means something specific in ethereum. This is a general function for recovering the public key from a secp256k1 signature, which might be useful in other situations than ethereum compatibility. How about something like `sol_secp256k1_recover`?

programs/bpf_loader/src/syscalls.rs
```diff
@@ -1329,6 +1344,79 @@ impl<'a> SyscallObject<BpfError> for SyscallMemset<'a> {
...
+                memory_mapping,
+                hash_addr,
+                keccak::HASH_BYTES as u64,
+                self.loader_id,
+                true,
+            ),
+            result
+        );
+        let signature = question_mark!(
+            translate_slice::<u8>(memory_mapping, signature_addr, 64u64, self.loader_id, true,),
```
jackcmay, line 1377: How about const defs for these magic numbers "64u64", I see them already defined in C, probably just have to define the same ones in rust and use them here

programs/bpf_loader/src/syscalls.rs
```diff
@@ -1329,6 +1344,79 @@ impl<'a> SyscallObject<BpfError> for SyscallMemset<'a> {
...
+        );
+        let secp256k1_recover_result = question_mark!(
+            translate_slice_mut::<u8>(memory_mapping, result_addr, 64u64, self.loader_id, true,),
+            result
+        );
+
+        let message = match libsecp256k1::Message::parse_slice(hash) {
+            Ok(msg) => msg,
+            Err(_) => {
+                *result = Ok(1); // invalid digest
```
jackcmay, line 1388: More magic numbers, what is "1"

sdk/program/src/secp256k1_recover.rs
```diff
@@ -0,0 +1,92 @@
...
+    Clone,
+    Copy,
+    Eq,
+    PartialEq,
+    Ord,
+    PartialOrd,
+    Hash,
+    AbiExample,
+)]
+pub struct Secp256k1Pubkey(pub [u8; 64]);
```
jackcmay, line 29: more magic numbers, looks like this 64 is used multiple places, be good to define it

jackcmay, line 29: Just checking, are we recovering a public key or an address?

sdk/program/src/secp256k1_recover.rs
```diff
@@ -0,0 +1,92 @@
...
+            sol_secp256k1_recover(
+                digest.as_ptr(),
+                recovery_id as u64,
+                signature.as_ptr(),
+                pubkey_buffer.as_mut_ptr(),
+            )
+        };
+
+        match result {
+            0 => Ok(Secp256k1Pubkey::new(&pubkey_buffer)),
```
jackcmay, line 71: Could have a from/into implementation for these that does the conversion for you here and also in syscall.rs where you return a result

sdk/program/src/secp256k1_recover.rs
```diff
@@ -0,0 +1,92 @@
...
+            1 => Err(Secp256k1RecoverError::InvalidDigest),
+            2 => Err(Secp256k1RecoverError::InvalidRecoveryId),
+            3 => Err(Secp256k1RecoverError::InvalidSignature),
+            _ => panic!("Unsupported Secp256k1RecoverError"),
```
jackcmay, line 75: Why panic?

programs/bpf_loader/src/syscalls.rs
```diff
@@ -317,6 +322,16 @@ pub fn bind_syscall_context_objects<'a>(
         }),
     );
 
+    bind_feature_gated_syscall_context_object!(
+        vm,
+        invoke_context.is_feature_active(&secp256k1_recover_syscall_enabled::id()),
+        Box::new(SyscallSecp256k1Recover {
+            base_cost: bpf_compute_budget.secp256k1_recover_base_cost,
```
jackcmay, line 329: just "cost", there is not per-byte cost associated with this so no need to distinguish this as a "base" cost

sdk/bpf/c/inc/solana_sdk.h
```diff
@@ -457,6 +457,37 @@ uint64_t sol_keccak256(
...
+ * Recover public key from a signed message.
+ *
+ * @param digest Hashed message
+ * @param recovery_id Tag used for public key recovery from signatures. Can be 0 or 1
+ * @param signature An ECDSA signature.
+ * @param result 64 byte array to hold the result
+ * @return 0 if executed successfully
+ */
+uint64_t sol_secp256k1_recover(
+    const uint8_t *digest,
```
jackcmay, line 485: s/digest/hash/?

sdk/bpf/c/inc/solana_sdk.h
```diff
@@ -457,6 +457,37 @@ uint64_t sol_keccak256(
...
+#define SECP256K1_RECOVER_ERROR_INVALID_DIGEST 1
+/** The recovery_id provided to a sol_secp256k1_recover is invalid */
+#define SECP256K1_RECOVER_ERROR_INVALID_RECOVERY_ID 2
+/** The signature provided to a sol_secp256k1_recover is invalid */
+#define SECP256K1_RECOVER_ERROR_INVALID_SIGNATURE 3
+
+/**
+ * Recover public key from a signed message.
+ *
+ * @param digest Hashed message
```
jackcmay, line 478: Might want to specify what mechanism must be sued to hash this message

sdk/bpf/c/inc/solana_sdk.h
```diff
@@ -457,6 +457,37 @@ uint64_t sol_keccak256(
...
+ * @param recovery_id Tag used for public key recovery from signatures. Can be 0 or 1
+ * @param signature An ECDSA signature.
+ * @param result 64 byte array to hold the result
```
jackcmay, line 481: Describe what exactly the result is?

sdk/program/src/secp256k1_recover.rs
anton-lisanin: This should never happen

sdk/program/src/secp256k1_recover.rs
anton-lisanin: a public key

sdk/bpf/c/inc/solana_sdk.h
anton-lisanin: Ethereum uses Keccak256. But in general, any other mechanism will work too.

programs/bpf_loader/src/syscalls.rs
anton-lisanin: Removed all magic numbers

programs/bpf_loader/src/syscalls.rs
anton-lisanin: Fixed

sdk/bpf/c/inc/solana_sdk.h
anton-lisanin: Renamed

sdk/src/feature_set.rs
```diff
@@ -139,6 +139,10 @@ pub mod memory_ops_syscalls {
     solana_sdk::declare_id!("ENQi37wsVhTvFz2gUiZAAbqFEWGN2jwFsqdEDTE8A4MU");
 }
 
+pub mod secp256k1_recover_syscall_enabled {
+    solana_sdk::declare_id!("Cj9ydNGWLePKRztuE3m3zT1uvj2We517k55vq2e65jtP");
```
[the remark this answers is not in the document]
anton-lisanin: Done

sdk/program/src/secp256k1_recover.rs
jackcmay: Any reason not to return `Secp256k1RecoverError::Unknown`?  Avoids specific panics in the sdk

anton-lisanin: Panic already used in the sdk for similar code. For example in sdk/program/src/pubkey.rs:38.
This can happen only if there is an error in the code. And in this case I prefer to be yelled with panic instead of silent Unknown.

jackcmay:
Looks good, we need to decide on compute cost.

jon-chuang:
@s-medvedev are you able to provide benchmarks for the bpf and the native cases? (probably one step of what you need to estimate the compute cost)

Separately, as far as I know [libsecp256k1](https://github.com/paritytech/libsecp256k1) doesn't use assembly. This could mean it's 1.2-2x slower than an optimised implementation. If this is a concern.

s-medvedev:
> @s-medvedev are you able to provide benchmarks for the bpf and the native cases? (probably one step of what you need to estimate the compute cost)

We compare bpf implementation and native implementation via syscall. The both implementation use libsecp256k1 library.
BPF implementation takes 3'000'000 instructions for computations. It takes approximately 0.014237809 seconds on our CI node.
```
[2021-06-14T07:41:41.929294200Z DEBUG solana_runtime::message_processor] Program HkpHRyDL129CgrhkdvqR36XMu5pQKGt2oCybomsk8CDs invoke [1]
[2021-06-14T07:41:41.943518531Z DEBUG solana_runtime::message_processor] Program log: Total memory occupied: 34800
[2021-06-14T07:41:41.943532009Z DEBUG solana_runtime::message_processor] Program HkpHRyDL129CgrhkdvqR36XMu5pQKGt2oCybomsk8CDs consumed 3266457 of 200000000 compute units

```
> 
> Separately, as far as I know [libsecp256k1](https://github.com/paritytech/libsecp256k1) doesn't use assembly. This could mean it's 1.2-2x slower than an optimised implementation. If this is a concern.

@jon-chuang Yes, you are right. libsecp256k1 performs this operation very slowly. We compare alternative libraries and get next results:

Library | Time (sec)
----- | ----
libsecp256k1 | 0.0001925
k256 | 0.0001561
secp256k1 | 0.00004772

_Note: Core Intel i7-7700HQ used for measurements_

To determine the cost regardless of computer performance, we decided to compare the time to compute `ecrecover` with the computation of a Keccak hash of a buffer with a certain length.
Based on this, measurements show that the **libsecp256k1** library is comparable to calculating the hash from a buffer with a length of 50,000 bytes, **secp256k1** - from a buffer with a length of 13,000 bytes.

You can see benchmark code in neonlabsorg/Solana-EVM#73

sdk/program/src/secp256k1_recover.rs
```diff
@@ -0,0 +1,113 @@
...
+        }
+    }
+}
+
+impl From<Secp256k1RecoverError> for u64 {
+    fn from(v: Secp256k1RecoverError) -> u64 {
+        match v {
+            Secp256k1RecoverError::InvalidHash => 1,
+            Secp256k1RecoverError::InvalidRecoveryId => 1,
```
ryoqun, line 30: this doesn't match to `From<u64> for Secp256k1RecoverError `. is it intentional?

anton-lisanin: It is not intentional. Fixed.

ryoqun:
Alternatively, we can introduce another special native program like `secp256k1 program` so that the operation could be easily pipeline-able?

This assumes the given signature is deterministic at the time of transaction (I think so?) and there needs some way to pass the recovered public key from the `ecrecover program` to the interested program.  `secp256k1 program` was easy because it just needs pass/fail as outcome...

jackcmay:
> Alternatively, we can introduce another special native program like `secp256k1 program` so that the operation could be easily pipeline-able?
> 
> This assumes the given signature is deterministic at the time of transaction (I think so?) and there needs some way to pass the recovered public key from the `ecrecover program` to the interested program. `secp256k1 program` was easy because it just needs pass/fail as outcome...

My understanding is that in this use case `ecrecover` can be called by a Solidty program and that information is unknown at the transaction level.  @s-medvedev can you confirm?

jackcmay:
> > @s-medvedev are you able to provide benchmarks for the bpf and the native cases? (probably one step of what you need to estimate the compute cost)
> 
> We compare bpf implementation and native implementation via syscall. The both implementation use libsecp256k1 library.
> BPF implementation takes 3'000'000 instructions for computations. It takes approximately 0.014237809 seconds on our CI node.
> 
> ```
> [2021-06-14T07:41:41.929294200Z DEBUG solana_runtime::message_processor] Program HkpHRyDL129CgrhkdvqR36XMu5pQKGt2oCybomsk8CDs invoke [1]
> [2021-06-14T07:41:41.943518531Z DEBUG solana_runtime::message_processor] Program log: Total memory occupied: 34800
> [2021-06-14T07:41:41.943532009Z DEBUG solana_runtime::message_processor] Program HkpHRyDL129CgrhkdvqR36XMu5pQKGt2oCybomsk8CDs consumed 3266457 of 200000000 compute units
> ```
> 
> > Separately, as far as I know [libsecp256k1](https://github.com/paritytech/libsecp256k1) doesn't use assembly. This could mean it's 1.2-2x slower than an optimised implementation. If this is a concern.
> 
> @jon-chuang Yes, you are right. libsecp256k1 performs this operation very slowly. We compare alternative libraries and get next results:
> 
> Library	Time (sec)
> libsecp256k1	0.0001925
> k256	0.0001561
> secp256k1	0.00004772
> _Note: Core Intel i7-7700HQ used for measurements_
> 
> To determine the cost regardless of computer performance, we decided to compare the time to compute `ecrecover` with the computation of a Keccak hash of a buffer with a certain length.
> Based on this, measurements show that the **libsecp256k1** library is comparable to calculating the hash from a buffer with a length of 50,000 bytes, **secp256k1** - from a buffer with a length of 13,000 bytes.
> 
> You can see benchmark code in [neonlabsorg/Solana-EVM#73](https://github.com/neonlabsorg/Solana-EVM/issues/73)

@s-medvedev Does switching from libsecp256k1 to secp256k1 require any api changes that might prevent us from switching the underlying implementation between the two later?

s-medvedev:
@jackcmay Yes that's right. Solidity contracts can use the ecrecover function and pass any information to it. Uniswap contracts use this functionality.

s-medvedev:
@jackcmay These two libraries differ slightly in interface, but nothing critical. The ecrecover system call can be implemented using these libraries without changing the interface.
See: https://github.com/neonlabsorg/ecrecover-bench/blob/1fffb9fad35371270d46d4dbba38393022866a38/src/ecrecover.rs#L165 & https://github.com/neonlabsorg/ecrecover-bench/blob/1fffb9fad35371270d46d4dbba38393022866a38/src/ecrecover.rs#L317

jackcmay:
@s-medvedev Want to resolve the conflicts and get us to all green for merge?

jon-chuang:
@s-medvedev , separately, would you mind sharing with me the bpf secp crate? I'd like to get the wallclock results when `bpf_jit` is set to `true`, as for `ProgramTest` it is in fact set to false by default.

Btw @jackcmay , if we're talking about the performance of bytecode, it may be instructive to consider that a WASM secp256k1 in this [bitcoin lib ](https://github.com/bitauth/libauth) is shown to [achieve](https://blog.bitjson.com/just-released-webassembly-version-of-secp256k1-10x-faster-than-javascript/) performance comparable to the native rust libsecp256k1:
0.00025195263s / op

s-medvedev:
> @s-medvedev , separately, would you mind sharing with me the bpf secp crate? I'd like to get the wallclock results when `bpf_jit` is set to `true`, as for `ProgramTest` it is in fact set to false by default.

We used the `libsecp256k1` library in BPF and got 3,000,000,000 instructions to recover one key. We then ported this as a system call to the Solana kernel level. The measurement results can be viewed above.
We could not use the `secp256k1` library as a BPF. When we compile it in our code and try to deploy program into Solana, we get an error:
```
$ solana program deploy evm_loader/program/target/deploy/evm_loader.so
Error: ELF error: ELF error: Relocation failed, unknown type 1
```

jon-chuang:
I see, thanks @s-medvedev . I am guessing it was most likely run with JIT, as without JIT, for an ed25519 scalar mul, I was getting 1s and above.

jackcmay:
> > @s-medvedev , separately, would you mind sharing with me the bpf secp crate? I'd like to get the wallclock results when `bpf_jit` is set to `true`, as for `ProgramTest` it is in fact set to false by default.
> 
> We used the `libsecp256k1` library in BPF and got 3,000,000,000 instructions to recover one key. We then ported this as a system call to the Solana kernel level. The measurement results can be viewed above.
> We could not use the `secp256k1` library as a BPF. When we compile it in our code and try to deploy program into Solana, we get an error:
> 
> ```
> $ solana program deploy evm_loader/program/target/deploy/evm_loader.so
> Error: ELF error: ELF error: Relocation failed, unknown type 1
> ```

I suspect this was due to rw data in the crate

PR #17721, opened 2021-06-03
add hash_time_pre_us metric to disambiguate

jeffwashington:
#### Problem
add hash calculation metric to disambiguate two areas timed by the same metric.
With high bin counts (65k), 'hash' takes huge amounts of time (~1s normal to ~8s).
#### Summary of Changes

Fixes #


lijunwangs (approved):
lgtm

PR #17727, opened 2021-06-03
add info for vm.max_map_count incorrectly set.

jeffwashington:
#### Problem
validator or other tools will fail with misleading message when vm.max_map_count is incorrectly set.
#### Summary of Changes
provide 'info' guiding weary travelers through this bog
Fixes #


t-nelson (approved):
:medal_sports:

CriesofCarrots:
Neat! TIL mergify won't automerge draft PRs

Issue #17696, opened 2021-06-02
Replay Stage feeds realtime instruction execution cost to Cost Model


taozhu-chicago:
#### Problem
Cost Model estimates transaction cost by summing up it's instructions execution costs and accounts access cost (see issue #16641), therefore Cost Model needs a way to be updated with execution cost for new and existing on-chain instructions, somewhat in realtime. 

#### Proposed Solution 
Replay Stage can collect instruction execution cost when replaying blockstore to bank, send collected data to Cost Model that is shared with Banking Stage via `RwLock`. 

- The collected data would be `Vec<(program_id: as PubKey, cost: as u64)>`
- Above data are sent to Cost Model at end of each bank's replay
- For new on-chain instructions, it's cost will be simply inserted to `cost_model.instruction_execution_cost` table
- For known instructions, I propose to average the new cost into existing one, so the cost table stay current. (eg `new_cost = AVG(collected_cost, existing_cost)`)

#### Implementation notes
The instruction is executed at `message_processor.execute_instruction(...)`, the path to it from Replay Stage is `replay_stage -> blockstore_process -> bank -> message_processor`; perhaps can utilize `replay_stats.execute_timings` by adding `Vec<(program_id: as PubKey, cost: as u64)>` to pass collected data back to Replay Stage

sakridge:
This one is adding per-program timings already:
https://github.com/solana-labs/solana/pull/17554

taozhu-chicago:
Great! This does all the heavy lifting for me. Left few comments on PR. Can we get it merged?

taozhu-chicago:
#28281 replaced cost estimation with user requested CU. This part of code (replay collects program execution cost, share with cost model) had been removed from codebase. Issue can be closed

PR #17731, opened 2021-06-04
replay stage feed back program cost

taozhu-chicago:
#### Problem
issue #17696

#### Summary of Changes
- shared CostModel between banking_stage and replay_stage with RwLock
- replay_stage writes collected pre-program execution cost to CostModel

Fixes #

### DIFF
--- a/core/src/cost_model.rs
+++ b/core/src/cost_model.rs
@@ -8,32 +8,20 @@
 //! The main function is `calculate_cost` which returns a TransactionCost struct.
 //!
 use log::*;
-use solana_sdk::{
-    bpf_loader, bpf_loader_deprecated, bpf_loader_upgradeable, feature, incinerator,
-    message::Message, native_loader, pubkey::Pubkey, secp256k1_program, system_program,
-    transaction::Transaction,
-};
+use solana_sdk::{message::Message, pubkey::Pubkey, transaction::Transaction};
 use std::collections::HashMap;
 
-// from mainnet-beta data, taking `vote program` as 1 COST_UNIT to load and execute
-// amount all type programs, the costs are:
-// min: 0.9 COST_UNIT
-// max: 110 COST UNIT
-// Median: 12 COST_UNIT
-// Average: 19 COST_UNIT
-const COST_UNIT: u32 = 1;
-const DEFAULT_PROGRAM_COST: u32 = COST_UNIT * 100;
-// re-adjust these numbers if needed
-const SIGNED_WRITABLE_ACCOUNT_ACCESS_COST: u32 = COST_UNIT * 10;
-const SIGNED_READONLY_ACCOUNT_ACCESS_COST: u32 = COST_UNIT * 2;
-const NON_SIGNED_WRITABLE_ACCOUNT_ACCESS_COST: u32 = COST_UNIT * 5;
-const NON_SIGNED_READONLY_ACCOUNT_ACCESS_COST: u32 = COST_UNIT;
-// running 'ledger-tool compute-cost' over mainnet ledger, the largest block cost
-// is 575_687, and the largest chain cost (eg account cost) is 559_000
-// Configuring cost model to have larger block limit and smaller account limit
-// to encourage packing parallelizable transactions in block.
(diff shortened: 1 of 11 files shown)


core/src/replay_stage.rs
```diff
@@ -1697,6 +1701,41 @@ impl ReplayStage {
...
                 );
+                debug!(
+                    "replayed into bank slot: {:?} / {:?}, result: {:?}, the execute_timings stats are: {:?}",
+                    bank.slot(),
+                    bank_slot,
+                    replay_result,
+                    bank_progress.replay_stats.execute_timings
+                );
+                let mut cost_model_mutable = cost_model.write().unwrap();
+                for (program_id, stats) in &bank_progress
```
sakridge, line 1712: Does the cost model get pruned for programs which have not been executed in a long time?

core/src/replay_stage.rs
```diff
@@ -1697,6 +1701,41 @@ impl ReplayStage {
...
                     verify_recyclers,
                 );
+                debug!(
+                    "replayed into bank slot: {:?} / {:?}, result: {:?}, the execute_timings stats are: {:?}",
+                    bank.slot(),
+                    bank_slot,
+                    replay_result,
+                    bank_progress.replay_stats.execute_timings
+                );
+                let mut cost_model_mutable = cost_model.write().unwrap();
```
sakridge, line 1711: Can we put this block into it's own function, like `update_cost_model`

taozhu-chicago: good idea ... will move it to function

core/src/replay_stage.rs
taozhu-chicago: Good point, if one intended to attack by repeatedly sending transactions with random program keys, it is a security risk. 

One solution is to set cost table capacity upfront, once its capacity is reached, it will have to push out oldest record to make room to new one. With this, there is no need to add additional logic to persisting/restoring table.

taozhu-chicago: Add ExecuteCostTable with fixed capacity; when capacity is reached, adding new program data will trigger pruning. Pruning removes program that are old and less frequently executed.

core/src/execute_cost_table.rs
```diff
@@ -0,0 +1,274 @@
...
+            return;
+        }
+
+        let now = SystemTime::now();
+        let mut sorted_by_weighted_age: Vec<_> = self
+            .occurrences
+            .iter()
+            .map(|(key, (count, timestamp))| {
+                let weighted_age: f64 =
+                    *count as f64 / (now.duration_since(*timestamp).unwrap().as_micros()) as f64;
```
sakridge, line 124: potential to divide by 0 here?

taozhu-chicago: yea, dont like division anyway, just pushed a commit changing formula to `=count * WEIGHT + (-1)*age`

core/src/cost_model.rs
```diff
@@ -7,33 +7,21 @@
...
-const SIGNED_READONLY_ACCOUNT_ACCESS_COST: u32 = COST_UNIT * 2;
-const NON_SIGNED_WRITABLE_ACCOUNT_ACCESS_COST: u32 = COST_UNIT * 5;
-const NON_SIGNED_READONLY_ACCOUNT_ACCESS_COST: u32 = COST_UNIT;
-// running 'ledger-tool compute-cost' over mainnet ledger, the largest block cost
-// is 575_687, and the largest chain cost (eg account cost) is 559_000
-// Configuring cost model to have larger block limit and smaller account limit
-// to encourage packing parallelizable transactions in block.
-pub const ACCOUNT_MAX_COST: u32 = COST_UNIT * 10_000;
-pub const BLOCK_MAX_COST: u32 = COST_UNIT * 10_000_000;
+// Guestimated from mainnet-beta data, sigver averages 1us, read averages 7us and write avergae 25us
```
sakridge, line 15: typo here: "write average 24us"

sakridge (approved):
lgtm

core/src/cost_model.rs
taozhu-chicago: thanks! will fix typo in separate PR.

PR #17754, opened 2021-06-04
Document ProgramTest::new and fix ProgramTest::add_program

xuoe:
This is a docs-only change for `ProgramTest::new` that should hopefully help others avoid misnaming their `ProgramTest`s and run into unexpected behavior during tests (e.g., tests hanging if the processor panics), which is caused by the test runner not being able to find the BPF shared object. 

@jstarry, you helped me figure this out earlier on Discord, thanks! Hope the doc-comment is clear enough.

### DIFF
--- a/program-test/src/lib.rs
+++ b/program-test/src/lib.rs
@@ -502,14 +502,12 @@ impl Default for ProgramTest {
 }
 
 impl ProgramTest {
-    /// Create a `ProgramTest` with the given arguments, where `program_name` *must* match the
-    /// name of the BPF shared object file (located at `<crate>/target/deploy/<program>.so` by
-    /// default), which is derived from the name of the crate. For example, if the program is
-    /// housed in a crate named `my-program`, then `program_name` should be provided as
-    /// `"my_program"`.
+    /// Create a `ProgramTest`.
     ///
-    /// See also [`add_program`], which is called by `new` internally.
+    /// This is a wrapper around [`default`] and [`add_program`]. See their documentation for more
+    /// details.
     ///
+    /// [`default`]: #method.default
     /// [`add_program`]: #method.add_program
     pub fn new(
         program_name: &str,
--- a/program-test/src/lib.rs
+++ b/program-test/src/lib.rs
@@ -502,6 +502,15 @@ impl Default for ProgramTest {
 }
 
 impl ProgramTest {
+    /// Create a `ProgramTest` with the given arguments, where `program_name` *must* match the
+    /// name of the BPF shared object file (located at `<crate>/target/deploy/<program>.so` by
+    /// default), which is derived from the name of the crate. For example, if the program is
+    /// housed in a crate named `my-program`, then `program_name` should be provided as
+    /// `"my_program"`.
+    ///
+    /// See also [`add_program`], which is called by `new` internally.
+    ///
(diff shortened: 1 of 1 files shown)


program-test/src/lib.rs
```diff
@@ -502,6 +502,15 @@ impl Default for ProgramTest {
 }
 
 impl ProgramTest {
+    /// Create a `ProgramTest` with the given arguments, where `program_name` *must* match the
```
jstarry, line 505: Looks like the default impl of `ProgramTest` already has a pretty good description: https://docs.rs/solana-program-test/1.7.0/solana_program_test/struct.ProgramTest.html#impl-Default. Do you mind copying that here and then also adding in the snippet from `add_program` which explains how `program_name` will be used to locate the shared object?

xuoe: Sure, I can do that, though I think it would be better to point to their documentation instead of duplicating the comments under `new`? After all, it's just a wrapper around `default()` and `add_program()`. What do you think?

xuoe:
Looking at this closely, I think the underlying issue stems from the following lines: https://github.com/solana-labs/solana/blob/ae0bea1ad3cb905ca33c766d7c397045f9fef46f/program-test/src/lib.rs#L600-L605

If we're invoked with `test-bpf` and no matching shared object is found, then we jump to the `else` branch and attempt to load the program natively. Shouldn't it always fail instead, regardless of whether a `process_instruction` is provided to `add_program()`/`new()`?

My thinking is: if I invoke it with `test` or `test-bpf`, then that's my intent, regardless of the parameters provided to `add_program()`. If those parameters happen to be incompatible with the invocation, then the program should panic with a descriptive message. Unless I'm misunderstanding and the `program-test` crate has other uses than simply being a bridge between `cargo` and a Solana program, then the behavior I describe above seems correct.

xuoe:
I've addressed the issue in my previous reply in https://github.com/solana-labs/solana/pull/17754/commits/211fa97fb3bd989111dcc6d5087f81bd63cc81d2 by unrolling the conditionals into a `match` expression. 

There's still a lingering issue that was difficult to spot before this commit: if a `ProgramTest` is invoked as native code and it panics, the test runner hangs. I mentioned this yesterday, but at that time I was under the impression that if the `program_name` was correct, the test would not hang anymore. However, that was only true because `ProgramTest::add_program` would always prefer BPF data over native code if the program name was valid, even for `test` invocations. So, a test that would otherwise hang when run as native code, would instead be run as BPF code, panic, and exit quickly, as expected, giving the impression that the program name was the culprit. Not sure if that was a bug, or a feature. :upside_down_face:

This issue can be reproduced on [v1.6.11](https://github.com/solana-labs/solana/tree/v1.6.11) by invoking `test` with an invalid program name and a panicking processor.

jstarry (approved):
Excellent PR, thank you for your contribution! Looks like there are some unrelated CI failures, I'll keep an eye on it.

jstarry:
> Shouldn't it always fail instead, regardless of whether a process_instruction is provided to add_program()/new()?

I've also felt the same way, thanks for making that change!

program-test/src/lib.rs
jstarry: Yup, that works, thanks!

xuoe:
Glad I could help!

PR #17769, opened 2021-06-06
explorer: Add transaction inspector

jstarry:
#### Problem
No easy way for devs to share, inspect, and simulate transaction messages. This is often a problem when transactions fail simulation and devs aren't sure why the transaction is failing.

#### Summary of Changes
This adds a new "Transaction Inspector" to the explorer. The inspector provides a convenient ui for decoding transaction messages for debugging purposes. Developer tooling can now emit explorer inspector links with url encoded message data for debugging!

- New inspector route: `/tx/inspector?message=<BASE64 ENCODED MESSAGE>`
- Added simulator card with pretty logging
- Display context for every account (balance, program owner, executable)

##### Empty state:

Screen Shot 2021-06-06 at 1 28 50 AM

##### Transaction inspection:

Screen Shot 2021-06-06 at 12 57 21 AM

##### Pretty logging

Screen Shot 2021-06-06 at 1 24 59 AM

Fixes #

### DIFF
--- a/explorer/src/pages/inspector/RawInputCard.tsx
+++ b/explorer/src/pages/inspector/RawInputCard.tsx
@@ -3,16 +3,60 @@ import { Message } from "@solana/web3.js";
 import type { TransactionData } from "./InspectorPage";
 import { useQuery } from "utils/url";
 import { useHistory, useLocation } from "react-router";
+import base58 from "bs58";
+
+function deserializeTransaction(bytes: Uint8Array): {
+  message: Message;
+  signatures: string[];
+} | null {
+  const SIGNATURE_LENGTH = 64;
+  const signatures = [];
+  try {
+    const signaturesLen = bytes[0];
+    bytes = bytes.slice(1);
+    for (let i = 0; i < signaturesLen; i++) {
+      const rawSignature = bytes.slice(0, SIGNATURE_LENGTH);
+      bytes = bytes.slice(SIGNATURE_LENGTH);
+      signatures.push(base58.encode(rawSignature));
+    }
+
+    const requiredSignatures = bytes[0];
+    if (requiredSignatures !== signaturesLen) {
+      throw new Error("Signature length mismatch");
+    }
+  } catch (err) {
+    // Errors above indicate that the bytes do not encode a transaction.
+    return null;
+  }
+
+  const message = Message.from(bytes);
+  return { message, signatures };
+}
+
+const MIN_MESSAGE_LENGTH =
+  3 + // header
+  1 + // accounts length
+  32 + // accounts, must have at least one address for fees
+  32 + // recent blockhash
+  1; // instructions length
+
+const MIN_TRANSACTION_LENGTH =
+  1 + // signatures length
+  64 + // signatures, must have at least one for fees
+  MIN_MESSAGE_LENGTH;
+
+const MAX_TRANSACTION_SIGNATURES =
+  Math.floor((1232 - MIN_TRANSACTION_LENGTH) / (64 + 32)) + 1;
 
 export function RawInput({
   value,
-  setMessage,
+  setTransactionData,
 }: {
   value?: string;
-  setMessage: (param: TransactionData | undefined) => void;
+  setTransactionData: (param: TransactionData | undefined) => void;
 }) {
   const rawTransactionInput = React.useRef<HTMLTextAreaElement>(null);
-  const [error, setError] = React.useState();
+  const [error, setError] = React.useState<string>();
   const [rows, setRows] = React.useState(3);
   const query = useQuery();
   const history = useHistory();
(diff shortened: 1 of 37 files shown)


jstarry:
@mvines fyi, does this seem useful in its current form? I'd like to display account changes in the simulation ui next.

mvines:
> @mvines fyi, does this seem useful in its current form? I'd like to display account changes in the simulation ui next.

This looks really great!  My immediate feedback is that it would be nice to support a copy-paste of a full Transaction in addition to just a Message.  Maybe try to deserialize both?  I feel like a non-zero amount of people will try to paste the full Transaction and then ping for support on Discord when it doesn't work.

jon-chuang:
Would you be able to show some screenshots for when there are errors? Both for system/native program errors, SPL library errors, and user defined program errors.

Currently, on-chain program errors tend to show up as a non human-readable enum value. It would be nice if somehow, custom program errors could return a description of the error?

explorer/src/pages/inspector/SimulatorCard.tsx
```diff
@@ -0,0 +1,249 @@
...
+          )
+        );
+
+        // Simulate without signers to skip signer verification
+        const resp = await connection.simulateTransaction(tx);
+
+        let depth = 0;
+        let instructionLogs: InstructionLogs[] = [];
+        const prefixBuilder = (depth: number) => {
+          const prefix = new Array(depth - 1).fill("\u00A0\u00A0").join();
```
oJshua, line 143:
```suggestion
          const prefix = new Array(depth - 1).fill("\u00A0\u00A0").join("");
```

explorer/src/pages/inspector/RawInputCard.tsx
```diff
@@ -0,0 +1,117 @@
...
+      }
+
+      setRows(Math.max(3, Math.min(10, Math.round(base64.length / 150))));
+
+      let buffer;
+      try {
+        buffer = Uint8Array.from(atob(base64), (c) => c.charCodeAt(0));
+      } catch (err) {
+        console.error(err);
+        setError(err);
```
oJshua, line 36: nit: The exception thrown by atob isn't very user friendly: "Failed to execute 'atob' on 'Window': The string to be decoded is not correctly encoded."

oJshua (dismissed):

jstarry:
> Would you be able to show some screenshots for when there are errors? Both for system/native program errors, SPL library errors, and user defined program errors.
> 
> Currently, on-chain program errors tend to show up as a non human-readable enum value. It would be nice if somehow, transaction inspector could return a description of the error?

@jon-chuang Here's the screenshot. Improving error messages would be great, are you concerned about having standardized error reporting for user deployed programs or the solana native programs?
Screen Shot 2021-06-21 at 3 51 35 PM

explorer/src/pages/inspector/SimulatorCard.tsx
jstarry: Thanks!

explorer/src/pages/inspector/RawInputCard.tsx
jstarry: Thanks, changed to `"Input must be base64 encoded"`

jstarry:
> > @mvines fyi, does this seem useful in its current form? I'd like to display account changes in the simulation ui next.
> 
> This looks really great! My immediate feedback is that it would be nice to support a copy-paste of a full Transaction in addition to just a Message. Maybe try to deserialize both? I feel like a non-zero amount of people will try to paste the full Transaction and then ping for support on Discord when it doesn't work.

Full transaction support added!

jon-chuang:
> @jon-chuang Here's the screenshot. Improving error messages would be great, are you concerned about having standardized error reporting for user deployed programs or the solana native programs?
> Screen Shot 2021-06-21 at 3 51 35 PM

Awesome! I think it's probably fine that a non-native program shows up as a `Custom` error as it does on the current explorer. 

Though, I'm not sure if the team is open to maintaining a mapping between the `Custom` enum value and an error string, provided by the bpf program authors. Perhaps, this could be opt-in and on-chain, similar to the way that validators can register some information about themselves like their name and website into an on-chain registry (`solana validator-info`)? Just an idea for future work.

jstarry:
> Though, I'm not sure it might be nice to maintain a mapping between the Custom enum value and an error string, provided by the bpf program authors. Perhaps, this could be opt-in and on-chain, similar to the way that validators can register some information about themselves like their name and website into an on-chain registry (solana validator-info)? Just an idea for future work.

Yeah, I think this should be standardized by the anchor IDL

> For now, native program custom error mapping could possibly be baked in to the explorer via derive(Debug) or otherwise, though. I could work on it in a separate PR if you'd like. This could possibly be handled generically.

Feel free, that would be awesome

jon-chuang:
Ok. As a note to self for the approach: 
Modify snippet below by parsing a `Custom(u32)`, define a different component with the `Pubkey, u32 -> Err string` mapping,  and print as 
`Custom(0x0): YourErrorString`
```
 text: `Program returned error: ${log.slice(
    log.indexOf(": ") + 2
  )}`,
```

jstarry:
> Possibly, we could even include querying error strings as an RPC call... This is trivial for native programs... Not sure how to make this generic over bpf programs... Anchor's solution would rely on a per-program IDL.

Sounds like a good idea, and yeah Anchor supports defining interfaces per program

PR #17772, opened 2021-06-06
Fix handling of stale links to bpf-tools in cargo-build-bpf

dmakarov:
#### Problem

When `sdk/bpf/dependencies/bpf-tools` is linked to an older version of bpf-tools and the link target doesn't exist, the tool fails to relink bpf-tools to correct target.

#### Summary of Changes

Update the logic of link checking and relinking.


solana-grimes:

jstarry:
Any reason not to port this to v1.7?

jstarry:
Any reason not to port this to v1.7?

dmakarov:
It can be backported. I should have done it.

jstarry:
Awesome, thanks for confirming

Issue #17765, opened 2021-06-06
AccountsDb Async Fetching Service


jon-chuang:
#### Problem
Currently, MMap access can trigger page fault. Typical usage is sequential `load_with_fixed_root`.

#### Proposed Solution
The solution is an `AccountsDb` async worker pool that context switches to interleave page faults.

The design is as follows: `AccountsDb` or `Accounts` exposes some methods that rely on async workers instead of sequentially performing IO. 

TO DO: MORE DETAILS

#### Choices and Evidence
According to https://github.com/jimblandy/context-switch, a tokio context switch (similar to go green threads), is lower than kernel thread context switch/thread creation. (although apparently not for blocking IO :cry: )

An alternative is [may](https://github.com/Xudong-Huang/may#the-io-heavy-bound-examples), which implements something similar to coroutines: [example use](https://github.com/oyundev/du-rs/blob/main/src/main.rs)

While asking a threadpool to perform IO would actually increase the IO time in the non-page fauting case, this time is actually very small and doesn't add much to the critical path. But, it can drastically reduce the time waiting around for page faults, possible even on Optane-like low latency SSDs.

Issue #17761, opened 2021-06-05
Handle Page Faults with timeouts and retries


jon-chuang:
1. If page fault to disk is forced, the current default 4-thread parallelism in the banking stage won’t save you from the sequential read latency.
This is 250us on NVME, v.s. about 20 ns from RAM - a 10,000x difference...! Realistically, given the deserialization overhead, a load operation would be in the 100ns-1us range, so let’s call that a 1000x factor. 
I think this is especially important from both a performance and security standpoint since we are considering increasing the account access list capacity to about 10-20x of current (I still don’t know if the cost model that taozhu is designing is based on the average or worst case). A single page fault can destroy throughput.

2. If txs bounce due to other accounts being locked, forwarding the transactions rather than retrying to process them immediately seems a bit of an odd choice.

Solutions:
**Timeouts and Retries, Scheduling in the Unprocessed Buffer**
When deserializing the transactions from packets, one should perhaps immediately submit the account data to a background service that handles the async IO. 

The bank loading is designed to timeout on slow loading accounts, and the banking stage retries them later on in the block. 

During the replay stage the prefetch service should be run on all accounts in the block. Hopefully, this can give time for slow-loading accounts to have time to be fetched into memory during the 400ms or so block construction window. 

In general, I think a leader is incentivised to follow this behaviour as having their blocks have a slower replay would mean a lower chance of their block being included and thus losing (currently immaterial) tx fees. We assume there is a correlation amongst validators which accounts are slow-loading and which are not. 

The disk-load timeout and retry policy can be specifically timed to agree with a 250us (or other validator specifiable, e.g. 20us for Optane or Z-SSD) window.

Weakness of eager preloading: crafting large accounts can OOM the node. Solution: make use of TX account size metadata and account-size aware cost model to prevent OOM.

cc @sakridge

jon-chuang:
The thinking was that this may not be very urgent for current cluster TPS of ~1K. So it's more of future-proofing, similar to the tx and block cost limits that are based on historical data.

Thinking a bit further, there are some nice "cooincidences":
1. With 4 threads, one can reach 100K-250K TPS on a normal machine. This means, 40-100K TXs per block, or 6-16 TXs every 250us per thread. So if one pipelines account loading to be about 25 txs, or even the 128 tx batch lock size, one can handle page-faulting completely adequately for these TPS (as long as those disk fetches are handled in parallel).

[...]

PR #17774, opened 2021-06-07
[draft] accounts: parallel load across txs via core-pinned threadpool

jon-chuang:
#### Problem
Currently, `load_and_execute_transactions` loads sequentially within bank. This means the worst case latency for multiple sequential page faults can stall throughput.

#### Summary of Changes
Parallelise loads across TXs using an over-provisioned thread pool pinned to a fixed number of logical cores. The thread-pool is over-provisioned (by default by 8x) to deal with OS blocking threads when page faulting.

Fixes: https://github.com/solana-labs/solana/issues/17765 https://github.com/solana-labs/solana/issues/17761

CC (for early feedback): @sakridge @carllin 

TODO: 
1. Check no perf regression when running bench TPS
2. Check for perf improvement when triggering OS page evictions, thus forcing page faults.

Related:
1. TODO: Pipeline load->execution in banking and replay stage to hide latency.

### DIFF
--- a/runtime/src/accounts.rs
+++ b/runtime/src/accounts.rs
@@ -9,6 +9,7 @@ use crate::{
         TransactionExecutionResult,
     },
     blockhash_queue::BlockhashQueue,
+    hashed_transaction::HashedTransaction,
     rent_collector::RentCollector,
     system_instruction_processor::{get_system_account_kind, SystemAccountKind},
 };
@@ -18,6 +19,7 @@ use dashmap::{
 };
 use log::*;
 use rand::{thread_rng, Rng};
+use rayon::prelude::*;
 use solana_sdk::{
     account::{Account, AccountSharedData, ReadableAccount, WritableAccount},
     account_utils::StateMut,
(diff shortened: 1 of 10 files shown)


runtime/src/pinned_threads.rs
```diff
@@ -0,0 +1,74 @@
...
+        let mut b = thread::Builder::new();
+        if let Some(name) = thread.name() {
+            b = b.name(name.to_owned());
+        }
+        if let Some(stack_size) = thread.stack_size() {
+            b = b.stack_size(stack_size);
+        }
+        let id_for_spawn = self.cores[self.core_id_pointer];
+        b.spawn(move || {
+            core_affinity::set_for_current(id_for_spawn);
```
carllin, line 54: Could you expand on in a comment what the problem we're trying to solve with pinning the threads/benefit of pinning these threads to particular cores would be? Also perhaps a new bench in `runtime/benches/accounts.rs` to demonstrate this difference?

> Check no perf regression when running bench TPS
Might also be a good idea to try running this on a full validator to see if there are any unexpected drawdowns in other components of the validator.

runtime/src/accounts.rs
```diff
@@ -467,6 +469,87 @@ impl Accounts {
             .collect()
     }
 
+    pub fn load_accounts_parallel<'a>(
```
carllin, line 472: Hmmm, this is not a bad idea, but I'm wondering if we can get even better perf if we scheduled the entire load + execute in pathway in parallel. Might be worth exploring 😃

jon-chuang: Hmm, the reason why I think that's not necessarily a good idea is that the execute portion is CPU-bound.

The purpose of making the load portion parallel or async was to reduce the tail-end critical path of sequentially loading 128 TXs with up to 20-256 accounts each.

In general tx parallelism can already be handled by spawning more banking or replay threads.

jon-chuang: Example API:
`get_accounts_async_batch(keys: &[Pubkey], to_load: &[Res<()>]) -> Vec<LoadedAccount>`

runtime/src/pinned_threads.rs
jon-chuang: Hey @carllin 

My apologies that the thoughts are a little jumbled and still evolving on the issue. Will eventually summarise investigations in a report of sorts - there is a larger question of avoiding mmap, as do most modern SSD-oriented DBs. And it ought to at least be spelled out what the potential benefits are. The biggest benefit may be to reduce memory requirement while improving the worst-case critical path (this is related to some security, performance, and pricing issues, such as the tx cost model).

---

Pinning the threads in particular was just an idea to reduce resource contention, so that the loading threads don’t eat up the resources of the entire CPU. 

However, I’ve found by measuring the time to perform `load_accounts_parallel` when running bench-tps that pinning more than 2 threads per core makes things worse. But, pinning 2 threads to each CPU rather than to a subset of the CPUs is actually the best option for the thread pool (250-500us for 128 tx batch). One can avoid core 0 for PoH.

Note that in the replay stage, a different scenario arises. One can no longer “cheat” by parallelising across threads. That’s because the default behaviour is to already saturate the threads with `load_and_execute`. What this means is that essentially, every job

Results (approximate from eyeballing):
| Description | Time for 128 tx `load_accounts_parallel` (banking stage) | (replay stage) |
|---|---|---|
| Pin 2 threads to all logical cores |250-500us| avg: 12_000us tail: 25_000us|
| Pin 1 thread to all logical cores |250-500us| avg: 25_000us tail: 50_000us|
| Pin 4 threads to all logical cores | 800-1800us | avg: 25_000us tail 50_000us|
| Pin 1 thread to 1/4 of logical cores | 800-1800us | NIL |
| Ordinary thread pool 1 thread per logical core | 800-1500us | avg: 25_000us tail 50_000us|
| Ordinary thread pool 2 threads per logical core | 800-1500us | avg: 12_000us tail 25_000us|
| Single thread | 1800-5000us | 1800-5000us |

jon-chuang: As it is difficult to manually trigger page faults, I decided to simulate a single 250us page faults on every account load. 
```rust
pub fn get_account<'a>(&'a self, offset: usize) -> Option<(StoredAccountMeta<'a>, usize)> {
        std::thread::sleep(std::time::Duration::from_micros(250));
```
Here are my results (32 IO threads on 8C16T machine):
This PR: 10_000-25_000us (banking). Master 350_000-500_000us. 

I honestly don't have an explanation for why a factor of 40 increase in page loading is observed...

PR #17778, opened 2021-06-07
Make account shrink configurable #17544

### LINKED ISSUE #17544: Accounts shrink could be configurable
#### Problem

Different hardware is used for storing the accounts data with different properties. For instance an all-RAM setup might tolerate high write-amplification but would like to have the smallest possible size. A disk setup which cares about write bandwidth might be willing to sacrifice some space to reduce write IO.

#### Proposed Solution

The shrink heuristics could be controllable via a validator argument which tunes how aggressive the accounts shrink policy is. Very aggressive would re-write often and shrink the smallest size even if it isn't saving many bytes per store. Less aggressive could allow for stores to be somewhat unused and not rewrite them until they are very sparse.

A target allowance on space might be an interesting way to put it. `--accounts-extra-space 5` would try to shrink the most sparse stores until it got to the 5% over the ideal target.

lijunwangs:
#### Problem
Different hardware is used for storing the accounts data with different properties. For instance an all-RAM setup might tolerate high write-amplification but would like to have the smallest possible size. A disk setup which cares about write bandwidth might be willing to sacrifice some space to reduce write IO.
#### Summary of Changes
1. Added both options for measuring space usage using total accounts usage and for individual store shrink ratio using an enum. Validator CLI options: --accounts-shrink-optimize-total-space and --accounts-shrink-ratio
2. Added code for selecting candidates based on total usage in a separate function select_candidates_by_total_usage
3. Added unit tests for the new functions added
4. The default implementations is kept at 0.8 shrink ratio on individual storage entries
Fixes #17544

#### Tests
Do integration testing running validator against MNB using both options. And observing the metrics captured. Did not find any material CPU usage over when accounts-shrink-optimize-total-space is true. In term of actual shrinking, I have measured the performance over several hours with both approaches. 
With AccountShrinkThreshold::IndividualStore way, we spend about 20% of the elapse time doing shrinking
With AccountShrinkThreshold::TotalSpace optimization: we spend about 10% of the time doing shrinking
Cargo test.

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -114,23 +114,31 @@ lazy_static! {
     pub static ref FROZEN_ACCOUNT_PANIC: Arc<AtomicBool> = Arc::new(AtomicBool::new(false));
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone, Copy)]
 pub enum AccountShrinkThreshold {
     /// Measure the total space sparseness across all candididates
     /// And select the candidiates by using the top sparse accounts to shrink
     /// The value is the overall shrink threshold measured as ratio of the total live bytes
     /// over the total bytes.
-    TotalSpace {ratio: f64},
+    TotalSpace { ratio: f64 },
     /// Use the following option to shrink all accounts whose alive ratio is below
     /// the specified threshold. All accounts with alive usage ratio below this thresholds
-    /// will be shrank. 
-    IndividalAccount {ratio: f64},
+    /// will be shrank.
+    IndividalAccount { ratio: f64 },
 }
-
+pub const DEFAULT_ACCOUNTS_SHRINK_OPTIMIZE_TOTAL_SPACE: bool = false;
 pub const DEFAULT_ACCOUNTS_SHRINK_RATIO: f64 = 0.80;
 // The default extra account space in percentage from the ideal target
-pub const DEFAULT_ACCOUNTS_SHRINK_THRESHOLD_OPTION: AccountShrinkThreshold = AccountShrinkThreshold::IndividalAccount {ratio: DEFAULT_ACCOUNTS_SHRINK_RATIO};
+pub const DEFAULT_ACCOUNTS_SHRINK_THRESHOLD_OPTION: AccountShrinkThreshold =
+    AccountShrinkThreshold::IndividalAccount {
+        ratio: DEFAULT_ACCOUNTS_SHRINK_RATIO,
+    };
 
+impl Default for AccountShrinkThreshold {
+    fn default() -> AccountShrinkThreshold {
+        DEFAULT_ACCOUNTS_SHRINK_THRESHOLD_OPTION
+    }
+}
 
 pub enum ScanStorageResult<R, B> {
     Cached(Vec<R>),
@@ -2370,7 +2378,7 @@ impl AccountsDb {
         let shrink_candidates_slots =
             std::mem::take(&mut *self.shrink_candidate_slots.lock().unwrap());
         let shrink_slots = {
-            if let AccountShrinkThreshold::TotalSpace {ratio} = self.shrink_ratio {
+            if let AccountShrinkThreshold::TotalSpace { ratio } = self.shrink_ratio {
                 self.select_candidates_by_total_usage(&shrink_candidates_slots, ratio)
             } else {
                 shrink_candidates_slots
@@ -4888,12 +4896,11 @@ impl AccountsDb {
 
     fn is_store_candidate_for_shrink(&self, store: Arc<AccountStorageEntry>) -> bool {
         match self.shrink_ratio {
-            AccountShrinkThreshold::TotalSpace {ratio: _} => { 
+            AccountShrinkThreshold::TotalSpace { ratio: _ } => {
                 Self::page_align(store.alive_bytes() as u64) < store.total_bytes()
-            },
-            AccountShrinkThreshold::IndividalAccount {ratio} => {
-                (Self::page_align(store.alive_bytes() as u64) as f64
-                    / store.total_bytes() as f64)
+            }
+            AccountShrinkThreshold::IndividalAccount { ratio } => {
+                (Self::page_align(store.alive_bytes() as u64) as f64 / store.total_bytes() as f64)
                     < ratio
             }
         }
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -124,7 +124,7 @@ pub enum AccountShrinkThreshold {
     /// the specified threshold.
     IndividalStore { shrink_ratio: f64 },
 }
-pub const DEFAULT_ACCOUNTS_SHRINK_OPTIMIZE_TOTAL_SPACE: bool = false;
+pub const DEFAULT_ACCOUNTS_SHRINK_OPTIMIZE_TOTAL_SPACE: bool = true;
 pub const DEFAULT_ACCOUNTS_SHRINK_RATIO: f64 = 0.80;
 // The default extra account space in percentage from the ideal target
 const DEFAULT_ACCOUNTS_SHRINK_THRESHOLD_OPTION: AccountShrinkThreshold =
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -57,6 +57,7 @@ use solana_vote_program::vote_state::MAX_LOCKOUT_HISTORY;
 use std::{
     borrow::{Borrow, Cow},
     boxed::Box,
+    cmp,
     collections::{hash_map::Entry, BTreeSet, HashMap, HashSet},
     convert::TryFrom,
     io::{Error as IoError, Result as IoResult},
@@ -82,7 +83,11 @@ const SCAN_SLOT_PAR_ITER_THRESHOLD: usize = 4000;
 pub const DEFAULT_FILE_SIZE: u64 = PAGE_SIZE * 1024;
 pub const DEFAULT_NUM_THREADS: u32 = 8;
 pub const DEFAULT_NUM_DIRS: u32 = 4;
-pub const SHRINK_RATIO: f64 = 0.80;
+
+// The default extra account space in percentage from the ideal target
+pub const DEFAULT_ACCOUNTS_SHRINK_OPTIMIZE_TOTAL_SPACE: bool = false;
+
+pub const DEFAULT_ACCOUNTS_SHRINK_RATIO: f64 = 0.80;
 
 // A specially reserved storage id just for entries in the cache, so that
 // operations that take a storage entry can maintain a common interface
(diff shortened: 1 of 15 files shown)


validator/src/main.rs
```diff
@@ -1777,6 +1784,28 @@ pub fn main() {
...
+                       the specified accounts-shrink-ratio, the shrink will stop and \
+                       it will skip all other less sparse accounts."),
+        )
+        .arg(
+            Arg::with_name("accounts_shrink_ratio")
+                .long("accounts-shrink-ratio")
+                .takes_value(true)
+                .value_name("RATIO")
+                .default_value(default_accounts_shrink_ratio)
+                .help("Specifies the shrink ratio for the accounts to be shrank. \
```
jeffwashington, line 1804: ...to be **_shrunk_**. I think.

lijunwangs: Will fix

validator/src/main.rs
```diff
@@ -1777,6 +1784,28 @@ pub fn main() {
...
+                       The shrink ratio is defined as the ratio of the bytes alive over the  \
+                       total bytes used. If the account's shrink ratio is less than this ratio \
+                       it becomes a candidate for shrinking. The value must between 0. - 1.0."),
```
jeffwashington, line 1807: nits: replace 'between 0. - 1.0.' with 'between 0. and 1.0, inclusive'

lijunwangs: Will fix

validator/src/main.rs
```diff
@@ -2075,6 +2104,23 @@ pub fn main() {
     let account_indexes = process_account_indexes(&matches);
 
     let restricted_repair_only_mode = matches.is_present("restricted_repair_only_mode");
+    let accounts_shrink_optimize_total_space =
+        value_t_or_exit!(matches, "accounts_shrink_optimize_total_space", bool);
+    let shrink_ratio = value_t_or_exit!(matches, "accounts_shrink_ratio", f64);
+    if !(0.0..=1.0).contains(&shrink_ratio) {
```
jeffwashington, line 2110: Can you set arg value limits that automatically check bounds for you? .takes_value, etc. I think there are range-limiting functions. I could be wrong.

lijunwangs: I wondered about the same and I did not find anything like that. I think the Arg does not have a mechanism to set the types of the argument value and that is probably we use value_t_or_exit! later to check the types.

validator/src/main.rs
lijunwangs: Fixed

runtime/src/accounts_db.rs
```diff
@@ -4767,6 +4891,18 @@ impl AccountsDb {
         true
     }
 
+    fn is_candidate_for_shrink(&self, store: Arc<AccountStorageEntry>) -> bool {
```
sakridge, line 4894: Should be able to borrow here to not clone the Arc. `&Arc<AccountStorageEntry>`

lijunwangs: Will fix

lijunwangs: Fixed

runtime/src/accounts_db.rs
```diff
@@ -2265,15 +2296,108 @@ impl AccountsDb {
         self.accounts_index.all_roots()
     }
 
+    fn select_candidates_by_total_usage(
```
carllin, line 2299: Nit: Maybe a comment on what this function does/returns?

lijunwangs: Will do

runtime/src/accounts_db.rs
```diff
@@ -2265,15 +2296,108 @@ impl AccountsDb {
...
+                .insert(store.append_vec_id(), store.clone());
+            let alive_ratio = (total_alive_bytes as f64) / (total_bytes as f64);
+            if alive_ratio > shrink_ratio {
+                // we have reached our goal, stop
+                debug!(
+                    "Shrinking goal can be achieved at slot {:?}, total_alive_bytes: {:?} \
+                    total_bytes: {:?}, alive_ratio: {:}, shrink_ratio: {:?}",
+                    usage.slot, total_alive_bytes, total_bytes, alive_ratio, shrink_ratio
+                );
+                break;
```
carllin, line 2362: Hmm can this section be simplified to not use the StoreUsageInfo. all_total_bytes calculation?

For instance if we:

1) Calculated the total number of bytes `old_total_bytes` here by summing up the bytes in this loop here: https://github.com/solana-labs/solana/blob/2dfdf1ee74bd130eba8b71f7a7c78e4514713dd8/runtime/src/accounts_db.rs#L2314-L2327

2) Sort from smallest alive ratio to largest alive ratio as is already done here: https://github.com/solana-labs/solana/blob/2dfdf1ee74bd130eba8b71f7a7c78e4514713dd8/runtime/src/accounts_db.rs#L2328-L2332

3) Then iterate through the sorted `store_usage` list from 2),  finding storages until we get to the ratio we want:

```
   let old_total_bytes = XXXX; // Calculated above
   let mut new_total_bytes = old_total_bytes;
  for usage in &store_usage {
         let store = &usage.store;     
        let current_store_size = store.total_bytes();
         let after_shrink_size = Self::page_align(store.alive_bytes() as u64);
        let bytes_saved = current_store_size.saturating_sub(after_shrink_size);
      new_total_bytes -= bytes_saved;
     shrink_slots                
      .entry(usage.slot)                
      .or_default()                
      .insert(store.append_vec_id(), store.clone());
     let new_to_old_ratio = (new_total_bytes as f64) / (old_total_bytes as f64);
     if alive_ratio < shrink_ratio {
           break;
     }
     
  }
```

runtime/src/accounts_db.rs
```diff
@@ -2265,15 +2296,108 @@ impl AccountsDb {
...
+        for usage in &store_usage {
+            let store = &usage.store;
+            shrunk_total += Self::page_align(store.alive_bytes() as u64);
+            let unshrunk_total = usage.all_total_bytes - store.total_bytes(); // the unshrunk total after this
+            let total_bytes = shrunk_total + unshrunk_total;
+            shrink_slots
+                .entry(usage.slot)
+                .or_default()
+                .insert(store.append_vec_id(), store.clone());
+            let alive_ratio = (total_alive_bytes as f64) / (total_bytes as f64);
```
carllin, line 2354: Is `total_alive_bytes` supposed to be updated in this loop?

runtime/src/accounts_db.rs
```diff
@@ -2265,15 +2296,108 @@ impl AccountsDb {
...
+            if alive_ratio > shrink_ratio {
```
carllin, line 2355: Is this condition flipped?

runtime/src/accounts_db.rs
```diff
@@ -2265,15 +2296,108 @@ impl AccountsDb {
...
+            measure.as_ms() as usize
+        );
+        inc_new_counter_info!("select_top_sparse_storage_entries-seeds", candidates_count);
+        shrink_slots
+    }
+
     pub fn shrink_candidate_slots(&self) -> usize {
-        let shrink_slots = std::mem::take(&mut *self.shrink_candidate_slots.lock().unwrap());
+        let shrink_candidates_slots =
+            std::mem::take(&mut *self.shrink_candidate_slots.lock().unwrap());
```
sakridge, line 2376: I was thinking that we would not remove the candidates from the queue in the case that we are optimizing for total usage. Then when we hit the usage limit then we would have a list to pick from.

runtime/src/accounts_db.rs
lijunwangs: Let me examine your idea.

runtime/src/accounts_db.rs
lijunwangs: total_alive_bytes does not change in my algorithm regardless if the candidate is selected or not.

runtime/src/accounts_db.rs
lijunwangs: I think it can work -- saving the loop of handling all_total_bytes. Thanks @carllin

runtime/src/accounts_db.rs
```diff
@@ -2265,15 +2296,105 @@ impl AccountsDb {
         self.accounts_index.all_roots()
     }
 
+    /// Given the input `ShrinkCandidates`, this function sorts the stores by their alive ratio
+    /// in increasing order with the most sparse entries in the front. It will then simulate the
+    /// shrinking by working on the most sparse entries first and if the overal alive ratio is
```
carllin, line 2301: nit: overal -> overall

runtime/src/accounts_db.rs
carllin: Yup, makes sense, thanks!

runtime/src/accounts_db.rs
lijunwangs: Fixed

sakridge (dismissed):
looks good, nice improvement.

runtime/src/accounts_db.rs
```diff
@@ -8973,6 +9105,99 @@ pub mod tests {
         );
     }
 
+    #[test]
+    fn test_select_candidates_by_total_usage() {
+        solana_logger::setup();
+
+        // case 1: no canidates
```
carllin, line 9112: nit: canidates -> candidates

runtime/src/accounts_db.rs
```diff
@@ -10933,4 +11168,23 @@ pub mod tests {
...
     }
+
+    #[test]
+    fn test_is_candidate_for_shrink() {
+        solana_logger::setup();
+
+        let mut accounts = AccountsDb::new_single();
+        let dummy_path = Path::new("");
+        let dummy_size = 2 * PAGE_SIZE;
+        let entry = Arc::new(AccountStorageEntry::new(&dummy_path, 0, 1, dummy_size));
```
carllin, line 11179: For clarity, maybe an assert, or an explicit assignment to make clear what the default assumed `accounts.shrink_ratio` is there?

runtime/src/accounts_db.rs
```diff
@@ -8973,6 +9105,99 @@ pub mod tests {
...
+
+        let output_candidates =
+            accounts.select_candidates_by_total_usage(&candidates, DEFAULT_ACCOUNTS_SHRINK_RATIO);
+        assert_eq!(2, output_candidates.len());
+        assert_eq!(1, output_candidates[&dummy_slot].len());
+        assert_eq!(1, output_candidates[&dummy_slot2].len());
+
+        assert!(output_candidates[&dummy_slot].contains(&entry1.append_vec_id()));
+        assert!(output_candidates[&dummy_slot2].contains(&entry2.append_vec_id()));
+    }
```
carllin, line 9199: Nice, thanks for the test coverage 😃

carllin (approved):

runtime/src/accounts_db.rs
lijunwangs: Yes. Will do

runtime/src/accounts_db.rs
lijunwangs: Fixed

PR #17779, opened 2021-06-07
Port unconfirmed duplicate tracking logic from ProgressMap to ForkChoice

carllin:
#### Problem
Keeping the duplicate/duplicate confirmed tracking logic in ProgressMap is unwieldy for a number of reasons:

1) Requires a separate update in `cluster_slot_state_verifier` to update the ProgressMap, and then separately mark certain forks valid/invalid in ForkChoice: https://github.com/solana-labs/solana/blob/18ec6756e49507c3945ab3d13d34dacbe897843b/core/src/cluster_slot_state_verifier.rs#L212-L217

2) Updating the ProgressMap for duplicate/duplicate confirmed tracking requires passing in ancestors/descendants: https://github.com/solana-labs/solana/blob/18ec6756e49507c3945ab3d13d34dacbe897843b/core/src/progress_map.rs#L477 because it doesn't know about fork structure. This means ancestors/descendants have to be passed along to `cluster_slot_state_verifier::check_slot_agrees_with_cluster()`  https://github.com/solana-labs/solana/blob/18ec6756e49507c3945ab3d13d34dacbe897843b/core/src/cluster_slot_state_verifier.rs#L242-L243 and a variety of ReplayStage functions that call down to that function

Furthermore, there are correctness issues that can arise because we track all the duplicate ancestor information in the ProgressMap and not ForkChoice. These stem from the fact that the fork choice only knows that a specific block was marked valid/invalid, but doesn't know if descendants of that block are valid/invalid. For example, consider the following order of events:

Given a fork `0 -> 1 -> 2 -> 3`
1) Fork 1 is marked duplicate. Currently, we set the `is_candidate` flag to false https://github.com/solana-labs/solana/blob/18ec6756e49507c3945ab3d13d34dacbe897843b/core/src/heaviest_subtree_fork_choice.rs#L819 and then removes the entire subtree from fork choice.
2) Fork 2 is marked duplicate, and removes its entire subtree from fork choice, similar to 1) above.
3) Fork 1 is marked valid. At this point, without searching all of its ancestors, fork 3 has no idea whether it is part of fork choice because its own `is_candidate` flag is not set.

Now 3) is not an issue when considering the *overall* best slot returned by fork choice: https://github.com/solana-labs/solana/blob/18ec6756e49507c3945ab3d13d34dacbe897843b/core/src/heaviest_subtree_fork_choice.rs#L175-L177, because that looks at the best slot field at the root node, and the root node contains aggregate information from the invalid fork `2`. However, functions like `heaviest_slot_on_same_voted_fork()`: https://github.com/solana-labs/solana/blob/18ec6756e49507c3945ab3d13d34dacbe897843b/core/src/heaviest_subtree_fork_choice.rs#L688, look at the best slot field of a node that is potentially in the subtree of `2`, and didn't contain the aggregate information that `2` was marked invalid.

#### Summary of Changes
Port duplicate fork/duplicate confirmed status tracking and tests to fork choice

Fixes #

### DIFF
--- a/core/src/heaviest_subtree_fork_choice.rs
+++ b/core/src/heaviest_subtree_fork_choice.rs
@@ -143,7 +143,7 @@ impl ForkInfo {
         newly_invalid_ancestor: Slot,
     ) {
         // Should not be marking a duplicate confirmed slot as invalid
-        assert!(!fork_info.is_duplicate_confirmed);
+        assert!(!self.is_duplicate_confirmed);
         if self
             .latest_invalid_ancestor
             .map(|latest_invalid_ancestor| newly_invalid_ancestor > latest_invalid_ancestor)
@@ -3222,7 +3222,6 @@ mod test {
         let last_duplicate_confirmed_key = last_duplicate_confirmed_slot.slot_hash();
         heaviest_subtree_fork_choice.mark_fork_valid_candidate(&last_duplicate_confirmed_key);
         for slot_hash_key in heaviest_subtree_fork_choice.fork_infos.keys() {
-            let slot = slot_hash_key.0;
             assert!(heaviest_subtree_fork_choice
                 .is_duplicate_confirmed(&slot_hash_key)
                 .unwrap());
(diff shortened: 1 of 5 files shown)


core/src/heaviest_subtree_fork_choice.rs
```diff
@@ -80,9 +94,64 @@ struct ForkInfo {
...
+            }
+        }
+    }
+
+    fn update_with_newly_invalid_ancestor(
+        &mut self,
+        my_key: &SlotHashKey,
+        newly_invalid_ancestor: Slot,
+    ) {
+        if self
```
jstarry, line 143: I think this should also check if the slot is already duplicate confirmed

carllin: I think we can do one better and add an assert here that `assert!(!fork_info.is_duplicate_confirmed);`: https://github.com/solana-labs/solana/pull/17779/commits/3ad6e3736d8598d7a9d72a4ad5ae990bff07205c#diff-f1f89ae82910cdfebacc27d6ef29a9a55dc89eea426a8e4fcb1586f4ca8da55cR146. 

This is because we should never be marking a duplicate confirmed slot as invalid. There's a test case `test_state_ancestor_confirmed_descendant_duplicate()`https://github.com/solana-labs/solana/blob/7a62cc1e977a6369d9f382ed47c1952e9dbe9546/core/src/cluster_slot_state_verifier.rs#L751 that covers this case

jstarry: Awesome, that was exactly the case I was thinking about

jstarry (approved):
lgtm

Issue #17762, opened 2021-06-05
Bad actors can abuse the privilege extension feature for Cross-Program Invocations via system_instruction::transfer, spl_token::instruction::approve, spl_token::instruction::transfer 


max-block:
#### Problem

A bad actor can create on-chain Solana programs which can steal SOL and SPL tokens. Solana wallets don't inform users about it.

Here is a simple program for example:
```rust
/// Accounts expected:
/// 0. `[signer, writable]` Debit lamports from this account
/// 1. `[writable]` Credit lamports to this account
/// 2. `[]` System program
pub fn process_instruction(
    _program_id: &Pubkey,
    accounts: &[AccountInfo],
    input: &[u8],
) -> ProgramResult {
    let acc_iter = &mut accounts.iter();
    let alice_info = next_account_info(acc_iter)?;
    let bob_info = next_account_info(acc_iter)?;
    // the third account is SystemProgram. Don't forget it in a client

    let amount = input
        .get(..8)
        .and_then(|slice| slice.try_into().ok())
        .map(u64::from_le_bytes)
        .ok_or(ProgramError::InvalidInstructionData)?;

    invoke(
        &system_instruction::transfer(alice_info.key, bob_info.key, amount),
        &[alice_info.clone(), bob_info.clone()],
    )?;
    msg!(
        "transfer {} lamports from {:?} to {:?}: done",
        amount,
        alice_info.key,
        bob_info.key
    );
    Ok(())
}
``` 

A bad actor can create such a program and to lie users about its behavior. For example, he can say that it a simple game, just for fun. 

The problem that Solana wallets don't inform users at all that their SOL or SPL tokens can be transferred.
I've made such a program and here are screenshots of Phantom and Sollet wallets, how they ask users to approve this dangerous tx:

**Phantom**
<img width="437" alt="phantom" src="https://user-images.githubusercontent.com/40041609/120899533-db05f100-c638-11eb-8e3f-1741452239b0.png">
Phantom wallet doesn't give any chance to a user to realize that his SOL can be stolen.

**Sollet**
<img width="495" alt="sollet" src="https://user-images.githubusercontent.com/40041609/120899539-e3f6c280-c638-11eb-96ed-49ff01779e8c.png">
Sollet wallet noticed that 11111111111111111111111111111111 account is used, but nothing about that lamports can be transferred.

Here is a repo with this demo: https://github.com/max-block/spl-examples/tree/v0.1.0/02__transfer-lamports
[...]

PR #17796, opened 2021-06-07
Propose traced transaction construction

ryoqun:
(boring meta-joke due to imminent sleep time.)

#### Problem

This is half-done!

#### Summary of Changes

Push more commitz and squaz

context (original idea is written when the itoken reviewing: https://github.com/solana-labs/solana/pull/15927#discussion_r595842064)

### DIFF
--- a/docs/src/proposals/traced-transaction-construction.md
+++ b/docs/src/proposals/traced-transaction-construction.md
@@ -241,7 +241,7 @@ breaks compatibility?
 unavoidable latency.  (and scheduling difficulty).
 The proposed implementation doesn't have such a downside
 
-# Future possible expansions
+## Future possible expansions
 
-- Wallet account balance delta check?
+- Wallet account balance delta check: append VerifyTokenBalances ix at last and pass the expected balances of spl-tokens or hash of it (also with delegation status).
 - Browser-side bpf program simualtion for server-less architecture with online wasm translation
--- a/docs/src/proposals/traced-transaction-construction.md
+++ b/docs/src/proposals/traced-transaction-construction.md
@@ -34,6 +34,10 @@ transaction execution in the simulation.
 Then, client signs the final transaction after appending these additional
 account addresses and submits it to the cluster via `sendTransaction`.
 
+To be specific, wallet should be able to manage these transaction finalization,
+agnostic to individual defi programs. This extra simulation process results in
+no-op if the executed program aren't aware of this new functionality.
+
 For that end, introduce a new syscall to trace account accesses, so that
 programs can dynamically hint their _internal_ or _derivable_ account addresses
 implicitly via , attaning the classic implementation-detail abstraction.
--- a/docs/src/proposals/traced-transaction-construction.md
+++ b/docs/src/proposals/traced-transaction-construction.md
@@ -1,4 +1,4 @@
-# Traced transaction construction via RPC's transaction simulation
+# Traced transaction construction via RPC's `simulateTransaction`
 
 ## Problem
 
--- a/docs/src/proposals/traced-transaction-construction.md
+++ b/docs/src/proposals/traced-transaction-construction.md
@@ -2,18 +2,18 @@
 
 ## Problem
 
-With more composition comes, it will be harder and harder for the higher-level
-consumer to correctly construct high-level transactions by precisely specifying
-each and every accessed accounts with the `read_only` or `read_write` proper
-flag.
+With more composition for Solana's program comes, it will be harder and harder
+for the higher-level consuming program to correctly construct lower-level
+program's instructions by precisely specifying each and every accessed accounts
+with the `read_only` or `read_write` flag.
 
 Also, there will be no way to signal the need of additional accounts for the
 arbitrarily customized behavior from the lower-level program to the higher-level
 one under the standardized instruction API like spl-token, even though we're
-about to explore the flexibility of allowing custom programs conforming to a
-standard like itoken.
+about to explore the flexibility of allowing customzed programs conforming to a
+standard like the itoken.
 
-The preceding runtime limitation are applicable both to on-chain (instruction
+These preceding runtime limitations are applicable both to on-chain (instruction
 construction for CPI) or off-chain (transaction construction by end-user
 interfaces) boundaries likewise.
 
@@ -141,7 +141,6 @@ pub fn process_transfer(
 }
 ```
 
-
 ## example: sharding
 
 
@@ -160,6 +159,23 @@ pub fn process_place_order(
     );
 }
 ```
+
+### example: instruction versioning by traced access after redeploy
+
+```
+pub fn check_instruction_version_and_load_data() {
+   // this is newly added after the redeploy
+   if access_as_unsigned(declare!("OldVersion111111111111111111111111111111111111"), READ).is_ok() {
+     return Err("we changed internal data loading implementation. to avoid unexpected malfunction retry");
+   }
+
+   // this is changed from OldVersion1111
+   access_as_unsigned(declare!("NewVersion111111111111111111111111111111111111"), READ)?
+
+   access_as_unsigned(...);
+}
+```
+
 
 ## RPC changes
 
--- a/docs/src/proposals/traced-transaction-construction.md
+++ b/docs/src/proposals/traced-transaction-construction.md
@@ -34,13 +34,15 @@ transaction execution in the simulation.
 Then, client signs the final transaction after appending these additional
 account addresses and submits it to the cluster via `sendTransaction`.
 
-To be specific, wallet should be able to manage these transaction finalization,
-agnostic to individual defi programs. This extra simulation process results in
-no-op if the executed program aren't aware of this new functionality.
-
 For that end, introduce a new syscall to trace account accesses, so that
 programs can dynamically hint their _internal_ or _derivable_ account addresses
-implicitly via , attaning the classic implementation-detail abstraction.
+implicitly, attaning the classic implementation-detail abstraction in the
+context Soalna's program ecosystem.
+
+In general, wallets should be able to manage these transaction finalization,
+agnostic to individual defi programs. Also, this extra simulation step results
+in no-op and harmless if all of the executed programs aren't aware of this new
+functionality.
 
 ## Traced syscalls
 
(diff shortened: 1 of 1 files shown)


docs/src/proposals/traced-transaction-construction.md
```diff
@@ -0,0 +1,247 @@
+# Traced transaction construction via RPC's `simulateTransaction`
+
+## Problem
+
+With more composition comes in the Solana's program ecosystem, it will be harder
```
ryoqun, line 5: this doc is still in the draft phase. but I think the first half of this is written decently. ;)

docs/src/proposals/traced-transaction-construction.md
```diff
@@ -0,0 +1,247 @@
...
+breaks compatibility?
+
+### Run these before replaying stage on validator
+
+unavoidable latency.  (and scheduling difficulty).
+The proposed implementation doesn't have such a downside
+
+## Future possible expansions
+
+- Wallet account balance delta check: append VerifyTokenBalances ix at last and pass the expected balances of spl-tokens or hash of it (also with delegation status).
```
ryoqun, line 246: this is my (hopefully nice) solution to this: https://github.com/solana-labs/solana/issues/17762

docs/src/proposals/traced-transaction-construction.md
```diff
@@ -0,0 +1,247 @@
...
+- Browser-side bpf program simualtion for server-less architecture with online wasm translation
```
ryoqun, line 247: this is only possible because solana tx execution doesn't touch the whole account state in the chain. :)

docs/src/proposals/traced-transaction-construction.md
ryoqun: also, dynamic ix append as directed traced simulation might also be useful.

docs/src/proposals/traced-transaction-construction.md
```diff
@@ -0,0 +1,247 @@
...
++    let associated_token_account_info = access_as_unsigned(associated_token_address, READ_WRITE)?;
+ 
+     let associated_token_account_signer_seeds: &[&[_]] = &[
+         &wallet_account_info.key.to_bytes(),
+```
+
+This reduces the number of required accounts for CPI and clients from 7 to 3, reducing by 4.
+(Obviously, we can't actually deploy these changes because this will break
+compatibility. This is only for illustration of effective applicability to
+moderately-complex program construct.)
```
ryoqun, line 122: also, maybe worth to mention that this doesn't reduce the resultant tx binary size, we need still this badly for more composability: https://github.com/solana-labs/solana/pull/15927 cc: @jstarry 

The good thing is that this nor the big tx block each other. :)

docs/src/proposals/traced-transaction-construction.md
```diff
@@ -0,0 +1,247 @@
...
+
+race condition around program deploy (= possible account loading change) => just retry also seek to the ix versioning for perfection
+
+(traced program just load dummy addresses for tx versionsing for completeness)
+
+## careful consideration
+
+prevent program  detect to be simulated or really-executed
+
+
```
ryoqun, line 232: also, write about the strip of is_signer and no use case limited by this stripping.

docs/src/proposals/traced-transaction-construction.md
```diff
@@ -0,0 +1,247 @@
...
+construction for CPI) or off-chain (transaction construction by end-user
+interfaces) boundaries likewise.
+
+Lastly, it's practically impossible to change those accessed account addresses
+once deployed while retaining compatibility, given the fact the consumers are
+expected to hard-code the account addresses for consumption, even if the
+deployed programs can be now upgradeable. This severely hampers ongoing
+consumed-program's implementation refinement in the areas of data migration and
+improved concurrency, by preventing transparent account switching and
+sharding with vectorized accounts, respectively.
```
jstarry, line 26: what is meant by "vectorized accounts"?

ryoqun: @jstarry don't be bothered with it.. ;) well, I used this fancy adjective, remotely referencing the usage like `vectorization with SSE4/AVX2`. Maybe `sharding` by itself should be enough.

`Sharding` here means avoiding write lock to the same internal account by using different accounts to write by deriving from the first byte of other account addresses like order_id for DEX. So, program can avoid write lock contention easily. Also, see the sharding example for code.

I want to say, hiding internal account addresses in this way makes these optimizations can be introduced by drop-in basis.

jstarry: Ah got it, I like the idea of abstracting away implementation details like this, thanks for the explanation!

docs/src/proposals/traced-transaction-construction.md
```diff
@@ -0,0 +1,247 @@
...
+in no-op and harmless if all of the executed programs aren't aware of this new
+functionality.
+
+## Traced syscalls
+
+The runtime exposes new syscall to provide new account access mechanism which
+can be traced:
+
+```rust
+fn access_as_unsigned(address: &Pubkey, flags = enum { READ, WRITE}):
```
ryoqun, line 53: Considering this https://github.com/solana-labs/solana/pull/17901,

Maybe we should augment this by returning Enum<AccountInfo, AccountRef> with new flag ACCESS_AS_REF. so that we can avoid any serializing into program execution environment.

AccountRef can only be used for passing to child cpi.

docs/src/proposals/traced-transaction-construction.md
```diff
@@ -0,0 +1,247 @@
...
+) -> ProgramResult {
+    let counter_address = Pubkey::find_program_address(
+        &[  
+            &program_id.to_bytes(),
+            "counter",
+        ],
+        program_id,
+    );
+   let transfer_counter = access_as_unsigned(counter_address, READ_WRITE)?;
+   let clock_sysvar = access_as_unsigned(sysvars::clock::id(), READ_WRITE)?;
```
jackcmay, line 141: Why is clock rw?

docs/src/proposals/traced-transaction-construction.md
```diff
@@ -0,0 +1,247 @@
...
+    expected_decimals: Option<u8>,    
+) -> ProgramResult {
+    let counter_address = Pubkey::find_program_address(
+        &[  
+            &program_id.to_bytes(),
+            "counter",
+        ],
+        program_id,
+    );
+   let transfer_counter = access_as_unsigned(counter_address, READ_WRITE)?;
```
jackcmay, line 140: What type is `transfer_counter`?

jackcmay:
This is interesting.  I like that it reduces and simplifies a program's calling interface by hiding internal-only accounts that the caller doesn't need to be or shouldn't be aware of.

jackcmay:
In addition to a syscall used to discover dependent accounts, if any cpi is invoked during simulation that specifies new accounts, those new accounts should also be added to the list.  During non-simulation those dependent accounts would be listed in the tx, loaded along with the rest of the accounts and referenced as needed.

docs/src/proposals/traced-transaction-construction.md
```diff
@@ -0,0 +1,247 @@
...
+
+```patch
+diff --git a/associated-token-account/program/src/processor.rs b/associated-token-account/program/src/processor.rs
+index 98eb08b..fa8ae50 100644
+--- a/associated-token-account/program/src/processor.rs
++++ b/associated-token-account/program/src/processor.rs
+@@ -22,13 +22,12 @@ pub fn process_instruction(
+     let account_info_iter = &mut accounts.iter();
+ 
+     let funder_info = next_account_info(account_info_iter)?;
```
ryoqun, line 91: related to #18588

docs/src/proposals/traced-transaction-construction.md
```diff
@@ -0,0 +1,247 @@
+# Traced transaction construction via RPC's `simulateTransaction`
```
ryoqun, line 1: traced => instrumented?

ruuda:
Tracing can only reveal references in code that is actually executed. Consider something like

```rust
let clock = Clock::get()?;
if clock.slot % 2 == 0 {
    reference_account_a();
} else {
    reference_account_b();
}
```

If the RPC traces the program execution and says it referenced account A, then it is likely that by the time it executes for real, the program will try to reference account B.

This is similar to how tests can only reveal bugs in code that is executed; to get an exhaustive answer, you need some kind of (static) program analysis or abstract interpretation.

ryoqun:
> Tracing can only reveal references in code that is actually executed. Consider something like
> 
> ```rust
> let clock = Clock::get()?;
> if clock.slot % 2 == 0 {
>     reference_account_a();
> } else {
>     reference_account_b();
> }
> ```
> 
> If the RPC traces the program execution and says it referenced account A, then it is likely that by the time it executes for real, the program will try to reference account B.
> 
> This is similar to how tests can only reveal bugs in code that is executed; to get an exhaustive answer, you need some kind of (static) program analysis or abstract interpretation.

@ruuda thanks for chiming in! you're first penguin from community. I really appreciate that.

Yeah, I understand the limitation. You have a point. But I still think this proposal could be reasonable solution for the problems described, at the cost of the strict requirement of rpc's simulation when issuing transactions to any advanced programs.

Maybe I should have added a prose like this?:

```
# Limitations (or Compromises)

This facility must not be abused and used wisely, needing proper developer
guidelines/communication. Firstly, dynamic account selection logic shouldn't
be timing sensitive or not change frequently. Because there is unavoidable
timing delays between simulation and actual execution on chain, transactions
could fail unexpectedly if implemented badly.

In other words, while dynamic account loading could give greater flexibility, the
logic should rather be static over time and for individual user sessions. The dynamic loading
isn't goal par itself. Rather this proposal tries to reduce and simplify a program's
calling interface by hiding internal-only accounts that the caller doesn't need to be or
shouldn't be aware of.

# Alternatives

- Some (static) IDL/DSL for account loading logic: hard to develop/design to cover any conceivable use cases. extra complexity
- provide no programmatic/systemic way to describe account loading (status quo): hard to compose defi programs. 
```

How do you like?

docs/src/proposals/traced-transaction-construction.md
```diff
@@ -0,0 +1,247 @@
...
+
+
+
+also, page-indexed account can be incorporated to find compact combination of existing account indexes to match the required account list
+
+## downside
+
+race condition around program deploy (= possible account loading change) => just retry also seek to the ix versioning for perfection
+
+(traced program just load dummy addresses for tx versionsing for completeness)
```
ryoqun, line 226: also, I have to embed this pr comment into doc: https://github.com/solana-labs/solana/pull/17796#issuecomment-885060430

ruuda:
> How do you like?

I don’t have a strong opinion on it, I just wanted to point out the limitation. Personally I think that with the way the Solana runtime is exposed to Rust at the moment, Rust is not a great language for writing Solana programs, and in the long run I expect some DSL to take over. A DSL could take care of those details internally, with a compiler figuring out which accounts are referenced.

PR #17806, opened 2021-06-07
add min to roots_tracker

jeffwashington:
#### Problem
We need to know the minimum root so we can start scanning the write cache at that root.
#### Summary of Changes
Implement min() on RollingBitField and up the call chain.
Fixes #


jeffwashington:
in the 100%* case, we will be able to just return the min.

carllin (approved):

PR #17808, opened 2021-06-07
Restart test without supermajority

AshwinSekar:
#### Problem
Simple test that restarts a 5 node network with equal stake and tries to confirm 5 transactions
Checking this in for ease of running in buildkite

#### Summary of Changes

Fixes #

### DIFF
--- a/system-test/restart-testcases/restart_test_automation.sh
+++ b/system-test/restart-testcases/restart_test_automation.sh
@@ -8,6 +8,10 @@ source "$(dirname "$0")"/../automation_utils.sh
 
 RESULT_FILE="$1"
 
+if [[ -z $CONSENSUS_TIMEOUT ]]; then
+  CONSENSUS_TIMEOUT=180
+fi
+
 startGpuMode="off"
 if [[ -z $ENABLE_GPU ]]; then
   ENABLE_GPU=false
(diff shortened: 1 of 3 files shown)


system-test/restart-testcases/restart_test_automation.sh
```diff
@@ -21,20 +21,16 @@ if [[ "$ASYNC_NODE_INIT" = "true" ]]; then
...
 sleep 2
 
 # shellcheck disable=SC2086
 "$REPO_ROOT"/net/net.sh start --skip-setup --no-snapshot-fetch --no-deploy \
-  --gpu-mode $startGpuMode $maybeAsyncNodeInit $maybeExtraPrimordialStakes
+  --gpu-mode $startGpuMode $maybeAsyncNodeInit
 
-# TODO add the test here
+# Basic test, check 5 transactions
+check_transaction_confirmations >> "$RESULT_FILE"
```
mvines, line 34: This sanity test may not be resilient enough once we start testing restarts with a larger set of nodes.  I imagine it could take a network several minutes to sort out consensus if they've been running forked for several hours.

Instead reading the total active stake from `solana validators --output=json | jq .totalCurrentStake` might be a better way than running `solana ping`.  Basically poll `totalCurrentStake` once every 5-10 seconds, and then timeout after N seconds (not sure what N is, maybe 60 for now).

system-test/restart-testcases/restart_test_automation.sh
```diff
@@ -21,20 +25,27 @@ if [[ "$ASYNC_NODE_INIT" = "true" ]]; then
...
-# TODO add the test here
-
-echo "Restart Test Succeeded" >>"$RESULT_FILE"
+  --gpu-mode $startGpuMode $maybeAsyncNodeInit
+
+# wait until consensus
+start=$SECONDS
+activeStake=get_active_stake
+while [[ $((SECONDS - start)) -lt $CONSENSUS_TIMEOUT ]]
+do
```
mvines, line 41: nit :)
```suggestion
while [[ $((SECONDS - start)) -lt $CONSENSUS_TIMEOUT ]]; do
```

mvines (approved):

PR #17810, opened 2021-06-07
feat: add convenience methods to EpochSchedule

Arrowana:
#### Problem
`EpochSchedule` in rust has methods that allow easy navigation from epoch to slot.
https://github.com/solana-labs/solana/blob/18ec6756e49507c3945ab3d13d34dacbe897843b/sdk/program/src/epoch_schedule.rs#L151-L162

web3js consumer currently will have to replicate this logic client side, possibly being wrong if they don't know the cluster specificity (like I did, thinking `slotsPerEpoch` was a constant throughout valid for any epoch).

#### Summary of Changes
Change `EpochSchedule` to be a class and instantiate it before returning from `getEpochSchedule`. It provides the relevant methods, it could potentially have more but I added the one relevant to my use case.

Fixes #

### DIFF
--- a/web3.js/src/epoch-schedule.ts
+++ b/web3.js/src/epoch-schedule.ts
@@ -3,7 +3,7 @@ const MINIMUM_SLOT_PER_EPOCH = 32;
 // Returns the number of trailing zeros in the binary representation of self.
 function trailingZeros(n: number) {
   let trailingZeros = 0;
-  while((n & 1) == 0) {
+  while ((n & 1) == 0) {
     n /= 2;
     trailingZeros++;
   }
@@ -26,7 +26,13 @@ export class EpochSchedule {
   /** The first slot of `firstNormalEpoch` */
   public firstNormalSlot: number;
 
-  constructor(slotsPerEpoch: number, leaderScheduleSlotOffset: number, warmup: boolean, firstNormalEpoch: number, firstNormalSlot: number) {
+  constructor(
+    slotsPerEpoch: number,
+    leaderScheduleSlotOffset: number,
+    warmup: boolean,
+    firstNormalEpoch: number,
+    firstNormalSlot: number,
+  ) {
     this.slotsPerEpoch = slotsPerEpoch;
     this.leaderScheduleSlotOffset = leaderScheduleSlotOffset;
     this.warmup = warmup;
@@ -35,11 +41,13 @@ export class EpochSchedule {
   }
 
   getFirstSlotInEpoch(epoch: number): number {
-    if(epoch <= this.firstNormalEpoch) {
+    if (epoch <= this.firstNormalEpoch) {
       return (Math.pow(2, epoch) - 1) * MINIMUM_SLOT_PER_EPOCH;
-    }
-    else {
-      return (epoch - this.firstNormalEpoch) * this.slotsPerEpoch + this.firstNormalSlot;
+    } else {
+      return (
+        (epoch - this.firstNormalEpoch) * this.slotsPerEpoch +
+        this.firstNormalSlot
+      );
     }
   }
 
@@ -48,11 +56,10 @@ export class EpochSchedule {
   }
 
   getSlotsInEpoch(epoch: number) {
-    if(epoch < this.firstNormalEpoch) {
+    if (epoch < this.firstNormalEpoch) {
       return Math.pow(2, epoch + trailingZeros(MINIMUM_SLOT_PER_EPOCH));
-    }
-    else {
+    } else {
       return this.slotsPerEpoch;
     }
   }
-};
+}
--- a/web3.js/src/epoch-schedule.ts
+++ b/web3.js/src/epoch-schedule.ts
@@ -1,5 +1,15 @@
 const MINIMUM_SLOT_PER_EPOCH = 32;
 
+// Returns the number of trailing zeros in the binary representation of self.
+function trailingZeros(n: number) {
+  let trailingZeros = 0;
+  while((n & 1) == 0) {
+    n /= 2;
+    trailingZeros++;
+  }
+  return trailingZeros;
+}
+
 /**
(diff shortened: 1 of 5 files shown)


web3.js/test/epoch-schedule.test.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import {EpochSchedule} from '../src';
+
+describe('EpochSchedule', () => {
+  it('slot methods work', () => {
+    const firstNormalEpoch = 8;
+    const firstNormalSlot = 8160;
+    const leaderScheduleSlotOffset = 8192;
+    const slotsPerEpoch = 8192;
+    const warmup = false;
```
jstarry, line 11: I think it'd be better to use the actual defaults here to ensure that this implementation is correct
```suggestion
    const firstNormalEpoch = 14;
    const firstNormalSlot = 524_256;
    const leaderScheduleSlotOffset = 432_000;
    const slotsPerEpoch = 432_000;
    const warmup = true;
```

web3.js/src/epoch-schedule.ts
```diff
@@ -0,0 +1,65 @@
+const MINIMUM_SLOT_PER_EPOCH = 32;
+
+// Returns the number of trailing zeros in the binary representation of self.
+function trailingZeros(n: number) {
+  let trailingZeros = 0;
+  while ((n & 1) == 0) {
```
jstarry, line 6: This will not stop looping if n is 0
```suggestion
  while (n > 1) {
```

web3.js/src/epoch-schedule.ts
```diff
@@ -0,0 +1,65 @@
...
+    trailingZeros++;
+  }
+  return trailingZeros;
+}
+
+/**
+ * Epoch schedule
+ * (see https://docs.solana.com/terminology#epoch)
+ */
+export class EpochSchedule {
```
jstarry, line 17: It would be awesome to have `getEpoch(slot: number)` and `getEpochAndSlotIndex(slot: number)` too :)

jstarry (commented):
Looks good @Arrowana! Just a few small changes requested + bonus points if you have time to add support for `getEpochAndSlotIndex` too

jstarry:
Feel free to ignore the commit linter errors, I'll squash the PR when it's ready to merge

web3.js/src/epoch-schedule.ts
```diff
@@ -3,16 +3,29 @@ const MINIMUM_SLOT_PER_EPOCH = 32;
...
+  n |= n >> 4;
+  n |= n >> 8;
+  n |= n >> 16;
+  return n + 1;
+}
+
 /**
  * Epoch schedule
  * (see https://docs.solana.com/terminology#epoch)
+ * Can be retreived with {@link connection.getEpochSchedule} method
```
jstarry, line 28:
```suggestion
 * Can be retrieved with the {@link Connection.getEpochSchedule} method
```

web3.js/src/epoch-schedule.ts
```diff
@@ -3,16 +3,29 @@ const MINIMUM_SLOT_PER_EPOCH = 32;
...
 
+// Returns the smallest power of two greater than or equal to n
+function nextPowerOfTwo(n: number) {
+  if (n === 0) return 1;
+  n--;
+  n |= n >> 1;
+  n |= n >> 2;
+  n |= n >> 4;
+  n |= n >> 8;
+  n |= n >> 16;
```
jstarry, line 21: I think you need one more iteration because `n` can be greater than `2^32`

web3.js/src/epoch-schedule.ts
```diff
@@ -40,6 +53,29 @@ export class EpochSchedule {
...
+        trailingZeros(nextPowerOfTwo(slot + MINIMUM_SLOT_PER_EPOCH + 1)) -
+        trailingZeros(MINIMUM_SLOT_PER_EPOCH) -
+        1;
+
+      const epochLen = this.getSlotsInEpoch(epoch);
+      const slotIndex = slot - (epochLen - MINIMUM_SLOT_PER_EPOCH);
+      return [epoch, slotIndex];
+    } else {
+      const normalSlotIndex = slot - this.firstNormalSlot;
+      const normalEpochIndex = (normalSlotIndex / this.slotsPerEpoch) | 0;
```
jstarry, line 72: Nice, I didn't know about this truncation method! I would prefer if an explicit `Math.floor` is used though since `number | 0` only works for 32 bit integers (source: https://stackoverflow.com/a/7488075)

web3.js/src/epoch-schedule.ts
```diff
@@ -40,6 +53,29 @@ export class EpochSchedule {
...
+      const epoch = this.firstNormalEpoch + normalEpochIndex;
+      const slotIndex = normalSlotIndex % this.slotsPerEpoch | 0;
```
jstarry, line 74: nit: no truncation needed here

web3.js/src/epoch-schedule.ts
Arrowana: correct, i blindly followed a solution i found and forgot we are "dealing" with u64 here

web3.js/src/epoch-schedule.ts
Arrowana: There is one in rust:
https://github.com/solana-labs/solana/blob/18ec6756e49507c3945ab3d13d34dacbe897843b/sdk/program/src/epoch_schedule.rs#L144-L146
For the edge case where we rem with 0...

jstarry (approved):
Awesome, thank you!

web3.js/src/epoch-schedule.ts
jstarry: Yeah good point, either way is fine then. I'm fine with expecting `slotsPerEpoch` to never be 0

PR #17816, opened 2021-06-07
Remove budget program

CriesofCarrots:
#### Problem
Tidying up. While no longer blocking Account data changes, the budget program is still unnecessary as per #16960. We know how to get it back if we need it.

#### Summary of Changes
Remove budget program and use other programs/instructions in tests

Closes #16960


ledger/src/entry.rs
```diff
@@ -819,23 +795,6 @@ mod tests {
         assert!(e0.verify(&zero));
     }
 
-    #[test]
-    fn test_witness_reorder_attack() {
```
CriesofCarrots, line 823: @garious , I didn't understand this unit test to be demonstrating anything different than `test_transaction_reorder_attack()` (ie. swapping two transactions that are identical other than instruction data). Am I missing something?

jeffwashington (approved):
I slashed code, and you fixed the test ;-) I think you know more than me about what the tests were doing and how to replace with equivalent functionality.

ledger/src/entry.rs
garious: Yep, looks redundant

garious (approved):

PR #17824, opened 2021-06-08
Add separate rules to generate dependencies files for C BPF build

dmakarov:
#### Problem

When bpf-tools updates clang version, the generated dependencies files are becoming invalid and need to be rebuilt.
The current bps.mk file doesn't handle such situations.

#### Summary of Changes

The generated dependencies files made dependent on changes to bpf.mk and regenerated separately when necessary.

### DIFF
--- a/sdk/bpf/c/bpf.mk
+++ b/sdk/bpf/c/bpf.mk
@@ -147,14 +147,28 @@ define C_RULE
 $1: $2
 	@echo "[cc] $1 ($2)"
 	$(_@)mkdir -p $(dir $1)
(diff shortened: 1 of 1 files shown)


t-nelson (approved):
lgtm!  Thanks again for taking this on!

FWIW, I think there's a way to do it without the `sed` stuff, but I'm not fluent in Makefile macros. I'd typically do a direct target with something like
```
%.d: %.c $(LOCAL_PATH)/bpf.mk
    ...
```
But if this works, let's go with it :slightly_smiling_face:

Issue #17304, opened 2021-05-18
Solana Validator 1.6.7 is failing with CUDA enabled


mabalaru:
Validator configuration:
```
solana-validator 1.6.7 (src:ebb5fc12; feat:3458834192)

4x Nvidia 2080 TI
256GB RAM
1TB PMEM drive

Ubuntu 20.04.2 LTS
Nvidia: Driver Version: 465.27       CUDA Version: 11.3
```
#### Problem
Validator is failing when CUDA is enabled with the following error :
```
May 18 11:43:29 m2-solana01 solana-validator[4867]: thread 'solana-replay-stage' panicked at 'cudaHostRegister error: 2 ptr: 0x7f0d491a1000 bytes: 2560', /var/lib/buildkite-agent/builds/bananas-1/solana-labs/solana-secondary/perf/src/cuda_runtime.rs:33:17
May 18 11:43:29 m2-solana01 solana-validator[4867]: stack backtrace:
May 18 11:43:29 m2-solana01 solana-validator[4867]: thread 'solana-receiver' panicked at 'cudaHostRegister error: 2 ptr: 0x7eff6ed7d700 bytes: 167936', perf/src/cuda_runtime.rs:33:17
May 18 11:43:29 m2-solana01 solana-validator[4867]: [2021-05-18T08:43:29.491517951Z INFO  solana_metrics::metrics] datapoint: retransmit-first-shred slot=78930907i
May 18 11:43:29 m2-solana01 solana-validator[4867]:    0: rust_begin_unwind
May 18 11:43:29 m2-solana01 solana-validator[4867]:              at ./rustc/2fd73fabe469357a12c2c974c140f67e7cdd76d0/library/std/src/panicking.rs:493:5
May 18 11:43:29 m2-solana01 solana-validator[4867]:    1: std::panicking::begin_panic_fmt
May 18 11:43:29 m2-solana01 solana-validator[4867]:              at ./rustc/2fd73fabe469357a12c2c974c140f67e7cdd76d0/library/std/src/panicking.rs:435:5
May 18 11:43:29 m2-solana01 solana-validator[4867]:    2: solana_perf::cuda_runtime::PinnedVec<T>::resize
May 18 11:43:29 m2-solana01 solana-validator[4867]:    3: <[solana_ledger::entry::Entry] [2021-05-18T08:43:29.502453020Z INFO  solana_metrics::metrics] datapoint: shred_insert_is_full total_time_ms=524i slot=78930906i last_index=328i
[...]

PR #17832, opened 2021-06-08
Don't use pinned memory for gossip and repair responses

### LINKED ISSUE #17775: when use cuda , validator node error
#### Problem
datapoint: panic program="validator" thread="solana-receiver" one=1i message="panicked at 'cudaHostRegister error: 2 ptr: 0x7f9d881ced40 bytes: 167936', perf/src/cuda_runtime.rs:32:13" location="perf/src/cuda_runtime.rs:32:13"

#### Proposed Solution
How to solved it

sakridge:
#### Problem

Reports of excessive GPU memory usage and errors from cudaHostRegister

#### Summary of Changes

Don't use pinned memory when unnecessary.

Fixes #17775 #17304

### DIFF
--- a/gossip/src/gossip_service.rs
+++ b/gossip/src/gossip_service.rs
@@ -54,6 +54,7 @@ impl GossipService {
             Recycler::default(),
             "gossip_receiver",
             1,
+            false,
         );
         let (response_sender, response_receiver) = channel();
         let t_responder = streamer::responder("gossip", gossip_socket, response_receiver);
--- a/streamer/src/streamer.rs
+++ b/streamer/src/streamer.rs
@@ -36,13 +36,18 @@ fn recv_loop(
     recycler: &PacketsRecycler,
     name: &'static str,
     coalesce_ms: u64,
+    use_pinned_memory: bool,
 ) -> Result<()> {
     let mut recv_count = 0;
     let mut call_count = 0;
     let mut now = Instant::now();
     let mut num_max_received = 0; // Number of times maximum packets were received
     loop {
-        let mut msgs = Packets::new_with_recycler(recycler.clone(), PACKETS_PER_BATCH, name);
+        let mut msgs = if use_pinned_memory {
+            Packets::new_with_recycler(recycler.clone(), PACKETS_PER_BATCH, name)
+        } else {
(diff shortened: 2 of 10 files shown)


gossip/src/gossip_service.rs
```diff
@@ -54,6 +54,7 @@ impl GossipService {
             Recycler::default(),
             "gossip_receiver",
             1,
+            false,
```
behzadnouri, line 57: I thought that we need this to be pinned for signature verification using gpu, no?

sakridge: GPU doesn't support the gossip path today. We can enable it if we add it.

mvines:
(nice, definitely want a 1.7 backport.  Maybe we don't want 1.6 just yet...)

streamer/src/streamer.rs
```diff
@@ -36,13 +36,18 @@ fn recv_loop(
...
     let mut recv_count = 0;
     let mut call_count = 0;
     let mut now = Instant::now();
     let mut num_max_received = 0; // Number of times maximum packets were received
     loop {
-        let mut msgs = Packets::new_with_recycler(recycler.clone(), PACKETS_PER_BATCH, name);
+        let mut msgs = if use_pinned_memory {
+            Packets::new_with_recycler(recycler.clone(), PACKETS_PER_BATCH, name)
+        } else {
+            Packets::new_unpinned_with_recycler(recycler.clone(), PACKETS_PER_BATCH, name)
```
behzadnouri, line 49: does the unpinned one need a recycler?

also, I think using the same recylcer for both pinned and unpinned ones will hurt performance when an unpinned vector is returned from the recycler but we need a pinned vector.

I think, the unpinned one just doesn't need a recycler.

sakridge: yea, it doesn't need it.

sakridge: I think in general we shouldn't mix pinned/unpinned recyclers. As far as I know, there wouldn't be any mixing with this PR. A `receiver` is created either with `use_pinned_memory=true` or `=false`, and then for the lifetime of that receiver it would always use either pinned or unpinned.

PR #17841, opened 2021-06-08
Add more info about how to safely change the identity of a staked validator

mvines:
Document the hidden gotcha that you need to run a second validator instance for two epochs after changing your identity


t-nelson (approved):
lgtm.  Maybe we should consider adding a flag that swaps the identity at an epoch boundary?

mvines:
> lgtm. Maybe we should consider adding a flag that swaps the identity at an epoch boundary?

Yep I've been poking around and thinking about how we can make this easier, it's really bad right now.  Plumbing an identity change affects most of core/ in lots of ways so it's nowhere near as easy as an authorized voter swap

PR #17851, opened 2021-06-08
Add more checks to verify_index_integrity test

steviez:
#### Problem
As is, `verify_index_integrity()` wasn't actually checking shreds that could be pulled out against the slot's index entry (index of all shreds that have been received for a given slot)

#### Summary of Changes
Add the checks in


carllin (approved):
Ha, this one made me chuckle. Good catch!

PR #17864, opened 2021-06-09
Warn about InstructionError meta

jackcmay:
#### Problem

Dangerous to add errors to InstructionError that contain meta from 3rd party crates and there is no warning developers to be careful of this

#### Summary of Changes

Warn and explain with comments

Fixes #

### DIFF
--- a/sdk/program/src/instruction.rs
+++ b/sdk/program/src/instruction.rs
@@ -206,12 +206,12 @@ pub enum InstructionError {
 
     /// Failed to serialize or deserialize account data
     ///
-    /// Warning: This error should never be emmited by the runtime.
+    /// Warning: This error should never be emitted by the runtime.
     ///
     /// This error includes strings from the underlying 3rd party Borsh crate
     /// which can be dangerous beause the error strings could change across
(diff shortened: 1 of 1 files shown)


sdk/program/src/instruction.rs
```diff
@@ -197,6 +205,14 @@ pub enum InstructionError {
     IncorrectAuthority,
 
     /// Failed to serialize or deserialize account data
+    ///
+    /// Warning: This error should never be emmited by the runtime.
```
CriesofCarrots, line 209:
```suggestion
    /// Warning: This error should never be emitted by the runtime.
```

sdk/program/src/instruction.rs
```diff
@@ -197,6 +205,14 @@ pub enum InstructionError {
...
+    ///
+    /// This error includes strings from the underlying 3rd party Borsh crate
+    /// which can be dangerous beause the error strings could change across
+    /// Borsh versions. Only programs can this error because they are consistent
```
CriesofCarrots, line 213: Only programs can "what" this error?

PR #17866, opened 2021-06-10
Account for duplicate before a bank is frozen or replayed

carllin:
#### Problem
Duplicate signals from before a bank is frozen are currently ignored

#### Summary of Changes
Add a `DuplicateSlotsTracker` for holding duplicate signals
Fixes #

### DIFF
--- a/core/src/replay_stage.rs
+++ b/core/src/replay_stage.rs
@@ -990,22 +990,23 @@ impl ReplayStage {
         progress: &mut ProgressMap,
         fork_choice: &mut HeaviestSubtreeForkChoice,
     ) {
-        let duplicate_slots: Vec<Slot> = duplicate_slots_receiver.try_iter().collect();
-
+        let new_duplicate_slots: Vec<Slot> = duplicate_slots_receiver
+            .try_iter()
+            .filter(|duplicate_slot|
+                // If this is the first time seeing this duplicate, run the processing
+                !duplicate_slots_tracker.insert(*duplicate_slot))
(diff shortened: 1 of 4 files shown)


core/src/replay_stage.rs
```diff
@@ -974,31 +982,38 @@ impl ReplayStage {
...
         descendants: &HashMap<Slot, HashSet<Slot>>,
         progress: &mut ProgressMap,
         fork_choice: &mut HeaviestSubtreeForkChoice,
     ) {
-        let duplicate_slots: Vec<Slot> = duplicate_slots_receiver.try_iter().collect();
+        let new_duplicate_slots: Vec<Slot> = duplicate_slots_receiver
+            .try_iter()
+            .filter(|duplicate_slot|
+                // If this is the first time seeing this duplicate, run the processing
+                !duplicate_slots_tracker.insert(*duplicate_slot))
```
jstarry, line 997: This filtering seems unnecessary since duplicate slots are sent if they were already detected in blockstore, right? Also, I think it would be cleaner to do the duplicate_slot insert inside the `check_slot_agrees_with_cluster` method (before the bank hash check)

jstarry (approved):
Looks good, great catch!

jstarry (dismissed):

core/src/replay_stage.rs
carllin: Updated!

jstarry (approved):

PR #17875, opened 2021-06-10
Rudimentary working implementation for incremental snapshots

brooksprumo:
This is a draft PR to get some eyes on my initial implementation for Incremental Snapshots (#17088).

So far this is only running through tests in `snapshot_utils.rs` (see the `roundtrip` tests at the bottom of the file). There are still many things to do for ISS:

- figure out cleaning (there will be some comments below)
- incorporate into the validator fully, so ISS are created while running, and ISS is used at startup
- add CLI args to validator to control ISS
- add ISS to RPC
- add ISS to gossip

In this PR you'll see a lot of `debug!()` prints. Those are just for me while developing and will eventually be removed once the impl is finalized.

The code also likely will not pass CI, since it is not meant to be a full/complete PR/implementation.

Additionally, many of the snapshot functions have been duplicated into sibling "incremental" versions. I'd like to remove this duplication and intend to do so by creating functions to handle the real impls with `Option<>` around the incremental snapshot parts. Unfortunately this doesn't work everywhere, in particular the deserialization, which I will call out below too.

### DIFF
--- a/runtime/src/serde_snapshot.rs
+++ b/runtime/src/serde_snapshot.rs
@@ -14,6 +14,7 @@ use {
         message_processor::MessageProcessor,
         rent_collector::RentCollector,
         serde_snapshot::future::SerializableStorage,
+        snapshot_info::SyncSnapshotInfo,
         stakes::Stakes,
     },
     bincode,
@@ -139,6 +140,7 @@ pub(crate) fn bank_from_stream<R>(
     caching_enabled: bool,
     limit_load_slot_count_from_snapshot: Option<usize>,
     shrink_ratio: AccountShrinkThreshold,
+    snapshot_info: Option<SyncSnapshotInfo>,
 ) -> std::result::Result<Bank, Error>
 where
     R: Read,
@@ -160,6 +162,7 @@ where
                 caching_enabled,
                 limit_load_slot_count_from_snapshot,
                 shrink_ratio,
+                snapshot_info,
             )?;
             Ok(bank)
         }};
@@ -176,8 +179,8 @@ where
 #[allow(clippy::too_many_arguments)]
 pub(crate) fn bank_from_stream_incremental<R>(
     serde_style: SerdeStyle,
-    fss_stream: &mut BufReader<R>,
-    iss_stream: &mut BufReader<R>,
+    full_snapshot_stream: &mut BufReader<R>,
+    incremental_snapshot_stream: &mut BufReader<R>,
     account_paths: &[PathBuf],
     unpacked_append_vec_map: UnpackedAppendVecMap,
     genesis_config: &GenesisConfig,
@@ -188,19 +191,22 @@ pub(crate) fn bank_from_stream_incremental<R>(
     caching_enabled: bool,
     limit_load_slot_count_from_snapshot: Option<usize>,
     shrink_ratio: AccountShrinkThreshold,
+    snapshot_info: Option<SyncSnapshotInfo>,
 ) -> std::result::Result<Bank, Error>
 where
     R: Read,
 {
     macro_rules! INTO {
         ($x:ident) => {{
-            let (_, fss_accounts_db_fields) = $x::deserialize_bank_fields(fss_stream)?;
-            let (bank_fields, iss_accounts_db_fields) = $x::deserialize_bank_fields(iss_stream)?;
+            let (_, full_snapshot_accounts_db_fields) =
+                $x::deserialize_bank_fields(full_snapshot_stream)?;
+            let (bank_fields, incremental_snapshot_accounts_db_fields) =
+                $x::deserialize_bank_fields(incremental_snapshot_stream)?;
 
             let bank = reconstruct_bank_from_fields_incremental(
                 bank_fields,
-                fss_accounts_db_fields,
-                iss_accounts_db_fields,
+                full_snapshot_accounts_db_fields,
+                incremental_snapshot_accounts_db_fields,
                 genesis_config,
                 frozen_account_pubkeys,
                 account_paths,
@@ -211,6 +217,7 @@ where
                 caching_enabled,
                 limit_load_slot_count_from_snapshot,
                 shrink_ratio,
+                snapshot_info,
             )?;
             Ok(bank)
         }};
@@ -307,6 +314,7 @@ fn reconstruct_bank_from_fields<E>(
     caching_enabled: bool,
     limit_load_slot_count_from_snapshot: Option<usize>,
     shrink_ratio: AccountShrinkThreshold,
+    snapshot_info: Option<SyncSnapshotInfo>,
 ) -> Result<Bank, Error>
 where
     E: SerializableStorage,
@@ -320,6 +328,7 @@ where
         caching_enabled,
         limit_load_slot_count_from_snapshot,
         shrink_ratio,
+        snapshot_info,
     )?;
     accounts_db.freeze_accounts(
         &Ancestors::from(&bank_fields.ancestors),
@@ -341,8 +350,8 @@ where
 #[allow(clippy::too_many_arguments)]
 fn reconstruct_bank_from_fields_incremental<E>(
     bank_fields: BankFieldsToDeserialize,
-    fss_accounts_db_fields: AccountsDbFields<E>,
-    iss_accounts_db_fields: AccountsDbFields<E>,
+    full_snapshot_accounts_db_fields: AccountsDbFields<E>,
+    incremental_snapshot_accounts_db_fields: AccountsDbFields<E>,
     genesis_config: &GenesisConfig,
     frozen_account_pubkeys: &[Pubkey],
     account_paths: &[PathBuf],
@@ -353,13 +362,14 @@ fn reconstruct_bank_from_fields_incremental<E>(
     caching_enabled: bool,
     limit_load_slot_count_from_snapshot: Option<usize>,
     shrink_ratio: AccountShrinkThreshold,
+    snapshot_info: Option<SyncSnapshotInfo>,
 ) -> Result<Bank, Error>
 where
     E: SerializableStorage,
 {
     let mut accounts_db = reconstruct_accountsdb_from_fields_incremental(
-        fss_accounts_db_fields,
-        iss_accounts_db_fields,
+        full_snapshot_accounts_db_fields,
+        incremental_snapshot_accounts_db_fields,
         account_paths,
         unpacked_append_vec_map,
         &genesis_config.cluster_type,
@@ -367,6 +377,7 @@ where
         caching_enabled,
         limit_load_slot_count_from_snapshot,
         shrink_ratio,
+        snapshot_info,
     )?;
     accounts_db.freeze_accounts(
         &Ancestors::from(&bank_fields.ancestors),
@@ -394,6 +405,7 @@ fn reconstruct_accountsdb_from_fields<E>(
     caching_enabled: bool,
     limit_load_slot_count_from_snapshot: Option<usize>,
     shrink_ratio: AccountShrinkThreshold,
+    snapshot_info: Option<SyncSnapshotInfo>,
 ) -> Result<AccountsDb, Error>
 where
     E: SerializableStorage,
@@ -404,9 +416,14 @@ where
         account_secondary_indexes,
         caching_enabled,
         shrink_ratio,
+        snapshot_info,
     );
     let AccountsDbFields(storage, version, slot, bank_hash_info) = accounts_db_fields;
-    // bprumo TODO: accounts_db.set_last_full_snapshot_slot(slot);
+
+    // bprumo TODO: fix this below
+    //accounts_db.set_last_full_snapshot_slot(slot);
+    //warn!("bprumo DEBUG: reconstruct_accountsdb_from_fields(), &accounts_db: {:p}, setting last full snapshot slot, slot: {}, accounts_db.last_full_snapshot_slot(): {:?}", &accounts_db, slot, accounts_db.last_full_snapshot_slot());
+    //snapshot_info::set_last_full_snapshot_slot(&snapshot_info, slot);
 
     // Ensure all account paths exist
     for path in &accounts_db.paths {
@@ -494,9 +511,10 @@ where
     Ok(accounts_db)
 }
 
+#[allow(clippy::too_many_arguments)]
 fn reconstruct_accountsdb_from_fields_incremental<E>(
-    fss_accounts_db_fields: AccountsDbFields<E>,
-    iss_accounts_db_fields: AccountsDbFields<E>,
+    full_snapshot_accounts_db_fields: AccountsDbFields<E>,
+    incremental_snapshot_accounts_db_fields: AccountsDbFields<E>,
     account_paths: &[PathBuf],
     unpacked_append_vec_map: UnpackedAppendVecMap,
     cluster_type: &ClusterType,
@@ -504,6 +522,7 @@ fn reconstruct_accountsdb_from_fields_incremental<E>(
     caching_enabled: bool,
     limit_load_slot_count_from_snapshot: Option<usize>,
     shrink_ratio: AccountShrinkThreshold,
+    snapshot_info: Option<SyncSnapshotInfo>,
 ) -> Result<AccountsDb, Error>
 where
     E: SerializableStorage,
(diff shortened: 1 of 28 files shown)


runtime/src/snapshot_utils.rs
```diff
@@ -509,6 +610,62 @@ where
     Ok(ret)
 }
 
+fn deserialize_incremental_snapshot_data_file_capped<F, T>(
```
brooksprumo, line 613: And here's what calls the deserializer above.

runtime/src/snapshot_utils.rs
```diff
@@ -809,25 +1243,133 @@ fn rebuild_bank_from_snapshots(
...
-                account_paths,
-                unpacked_append_vec_map,
+
+    debug!(
+        "bprumo DEBUG: rebuild_bank_from_incremental_snapshots(), Loading bank from {} and {}",
+        &fss_root_paths.snapshot_file_path.display(),
+        &iss_root_paths.snapshot_file_path.display()
+    );
+
+    let bank = deserialize_incremental_snapshot_data_file(
```
brooksprumo, line 1277: Here's the top that calls into the deserialization that spits out the bank.

runtime/src/snapshot_utils.rs
```diff
@@ -996,6 +1540,75 @@ pub fn bank_to_snapshot_archive<P: AsRef<Path>, Q: AsRef<Path>>(
...
+    bank.force_flush_accounts_cache();
+
+    // bprumo TODO: Cleaning still needs to be figured out.  My understanding is that we cannot
+    // clean past the last full snapshot slot here, but more testing needs to be done.  For now,
+    // I'm not cleaning at all when creating an incremental snapshot, since a full snapshot must
+    // have been created first, which would have performed a clean then.  Below are the different
+    // `clean_accounts()` options:
+    //      bank.clean_accounts(true, false);
+    //      bank.clean_accounts_up_to_slot(Some(full_snapshot_slot), false);
+    //      bank.clean_accounts_up_to_slot(Some(full_snapshot_slot.saturating_sub(1)), false);
```
brooksprumo, line 1575: Note that these three are all effectively the same, if `AccountsDb::last_full_snapshot_slot` is correct.

runtime/src/snapshot_utils.rs
```diff
@@ -1259,4 +2197,245 @@ mod tests {
...
+    /// this version, build up a few slots and take a full snapshot.  Continue on a few more slots
+    /// and take an incremental snapshot.  Rebuild the bank from both the incremental snapshot and
+    /// full snapshot.
+    ///
+    /// For the full snapshot, touch all the accounts, but only one for the incremental snapshot.
+    /// This is intended to mimic the real behavior of transactions, where only a small number of
+    /// accounts are modified often, which are captured by the incremental snapshot.  The majority
+    /// of the accounts are not modified often, and are captured by the full snapshot.
+    #[test]
+    fn test_roundtrip_bank_to_incremental_snapshot_to_bank() {
```
brooksprumo, line 2345: This is the main test I'm using the ensure this impl works. Does it look OK? Any flaws?

On my TODO list is to also add in more complete tests in `core/tests/snapshots.rs`.

brooksprumo (commented):
OK there's a lot in here! Thanks in advance for looking this over. I've tried to call out some spots that I thought would be helpful.

runtime/src/snapshot_utils.rs
```diff
@@ -39,15 +40,28 @@ use {
     thiserror::Error,
 };
 
+/// Information about a snapshot archive: its path, slot, hash, and archive format
+pub type SnapshotArchiveInfo = (PathBuf, (Slot, Hash, ArchiveFormat));
+
+/// Information about an incremental snapshot archive: its path, full snapshot base slot, incremental snapshot slot, hash, and archive format
+pub type IncrementalSnapshotArchiveInfo = (PathBuf, (Slot, Slot, Hash, ArchiveFormat));
```
carllin, line 47: I think this should be a struct with named fields

runtime/src/serde_snapshot.rs
```diff
@@ -387,3 +494,117 @@ where
     accounts_db.generate_index(limit_load_slot_count_from_snapshot);
     Ok(accounts_db)
 }
+
+fn reconstruct_accountsdb_from_fields_incremental<E>(
+    fss_accounts_db_fields: AccountsDbFields<E>,
```
carllin, line 499: nit: out of curiosity, what's the second `s`? full snapshot _ / incremental snapshot _?

runtime/src/serde_snapshot.rs
```diff
@@ -173,6 +173,57 @@ where
...
+    serde_style: SerdeStyle,
+    fss_stream: &mut BufReader<R>,
+    iss_stream: &mut BufReader<R>,
+    account_paths: &[PathBuf],
+    unpacked_append_vec_map: UnpackedAppendVecMap,
+    genesis_config: &GenesisConfig,
+    frozen_account_pubkeys: &[Pubkey],
+    debug_keys: Option<Arc<HashSet<Pubkey>>>,
+    additional_builtins: Option<&Builtins>,
+    account_indexes: AccountSecondaryIndexes,
```
carllin, line 187: nit: might also be a good time to rename `account_indexes` -> `account_secondary_indexes`

runtime/src/serde_snapshot.rs
```diff
@@ -173,6 +173,57 @@ where
     })
 }
 
+#[allow(clippy::too_many_arguments)]
+pub(crate) fn bank_from_stream_incremental<R>(
```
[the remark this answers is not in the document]
carllin: Yeah I think an option works.

Alternatively we could wrap them into a series of objects that convert into the next step so we don't have to pass as many objects around.

For instance maybe starting with:

1) `SerializedStream {
        fss_stream,
        iss_stream,
}`

2) Converts via ```let (_, fss_accounts_db_fields) = $x::deserialize_bank_fields(fss_stream)?;
            let (bank_fields, iss_accounts_db_fields) = $x::deserialize_bank_fields(iss_stream)?;```

to `DeserializedSnapshotFields {
      fss_accounts_db_fields,
      iss_accounts_db_fields,
       bank_fields
}`

and these can all expose getters that return `Option` if an incremental snapshot doesn't exist

runtime/src/serde_snapshot.rs
```diff
@@ -387,3 +494,117 @@ where
...
+                (slot, Arc::new(RwLock::new(slot_storage_entry)))
+            }),
+        );
+    }
+
+    if max_id > AppendVecId::MAX / 2 {
+        panic!("Storage id {} larger than allowed max", max_id);
+    }
+
+    accounts_db.next_id.store(max_id + 1, Ordering::Relaxed);
```
carllin, line 604: heh, actually not sure what we use the `next_id` here in the accounts code for 😝 . But one thing that seems possible if we are getting the incremental snapshot and full snapshot from different nodes is that the `next_id` is not unique between storage entries in the incremental and full snapshot, i.e. they could potentially overlap if the two validators generating those snapshots started from snapshots with different `next_ids`. 

We might need to re-id storage entries with the same id.

runtime/src/snapshot_utils.rs
brooksprumo: I agree. Will do!

runtime/src/serde_snapshot.rs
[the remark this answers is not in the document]
brooksprumo: I like it.

runtime/src/serde_snapshot.rs
brooksprumo: Hah, yeah, so this was for two reasons. One, is sometimes I wanted to talk about snapshots generally with an abbreviation, and a single "S" didn't work for that, so I was doing "SS", i.e. **s**nap**s**hot. Two, I found myself writing variables/comments/debug prints for things like "incremental snapshot slot" or "incremental snapshot hash", etc. With "IS", this ended up as `is_slot`, so it could appear to be a boolean.

I'm also fine fully spelling this vars out as well.

runtime/src/serde_snapshot.rs
brooksprumo: I agree. Will do.

runtime/src/serde_snapshot.rs
```diff
@@ -387,3 +494,117 @@ where
...
+        std::fs::create_dir_all(path)
+            .unwrap_or_else(|err| panic!("Failed to create directory {}: {}", path.display(), err));
+    }
+
+    let mut last_log_update = Instant::now();
+    let mut remaining_slots_to_process = fss_storage.len() + iss_storage.len();
+
+    // Remap the deserialized AppendVec paths to point to correct local paths
+    let mut storage = fss_storage.into_iter().chain(iss_storage.into_iter())
+        .map(|(slot, mut slot_storage)| {
```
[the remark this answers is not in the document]
brooksprumo: Sounds like a good idea to me. Will do.

runtime/src/serde_snapshot.rs
brooksprumo: Oh interesting point! I'll look into how to do that.

runtime/src/serde_snapshot.rs
mvines: > I'm also fine fully spelling this vars out as well.

Yes please fully spell it everything, for the benefit of all future code readers that'll assume ISS means a piece of space junk orbiting the earth, etc.

brooksprumo: > Yes please fully spell it everything [...]

Will do!

runtime/src/snapshot_utils.rs
```diff
@@ -809,25 +1243,133 @@ fn rebuild_bank_from_snapshots(
...
+        info!(
+            "Rebuilding status cache from {}",
+            status_cache_path.display()
+        );
+        let slot_deltas: Vec<BankSlotDelta> = bincode::options()
+            .with_limit(MAX_SNAPSHOT_DATA_FILE_SIZE)
+            .with_fixint_encoding()
+            .allow_trailing_bytes()
+            .deserialize_from(stream)?;
+        Ok(slot_deltas)
```
carllin, line 1312: maybe a comment here about why we take the status cache from the incremental snapshot and not the full snapshot

runtime/src/snapshot_utils.rs
```diff
@@ -996,6 +1540,75 @@ pub fn bank_to_snapshot_archive<P: AsRef<Path>, Q: AsRef<Path>>(
...
+    thread_pool: Option<&ThreadPool>,
+    maximum_snapshots_to_retain: usize,
+) -> Result<PathBuf>
+where
+    P: fmt::Debug,
+{
+    assert!(bank.is_complete());
+    assert!(bank.slot() > full_snapshot_slot);
+
+    bank.squash(); // Bank may not be a root
```
carllin, line 1565: Bank should already be rooted since AccountsBackgroundService only gets requests to snapshot roots?

runtime/src/snapshot_utils.rs
```diff
@@ -996,6 +1540,75 @@ pub fn bank_to_snapshot_archive<P: AsRef<Path>, Q: AsRef<Path>>(
...
+    // clean past the last full snapshot slot here, but more testing needs to be done.  For now,
+    // I'm not cleaning at all when creating an incremental snapshot, since a full snapshot must
+    // have been created first, which would have performed a clean then.  Below are the different
+    // `clean_accounts()` options:
+    //      bank.clean_accounts(true, false);
+    //      bank.clean_accounts_up_to_slot(Some(full_snapshot_slot), false);
+    //      bank.clean_accounts_up_to_slot(Some(full_snapshot_slot.saturating_sub(1)), false);
+
+    bank.update_accounts_hash();
+    bank.rehash(); // Bank accounts may have been manually modified by the caller
```
carllin, line 1578: Whether these are necessary will be determined in relation to where this function is called with the flush + clean + shrink logic in AccountsBackgroundService

runtime/src/snapshot_utils.rs
```diff
@@ -996,6 +1540,75 @@ pub fn bank_to_snapshot_archive<P: AsRef<Path>, Q: AsRef<Path>>(
...
+
+    bank.squash(); // Bank may not be a root
+    bank.force_flush_accounts_cache();
+
+    // bprumo TODO: Cleaning still needs to be figured out.  My understanding is that we cannot
+    // clean past the last full snapshot slot here, but more testing needs to be done.  For now,
+    // I'm not cleaning at all when creating an incremental snapshot, since a full snapshot must
+    // have been created first, which would have performed a clean then.  Below are the different
+    // `clean_accounts()` options:
+    //      bank.clean_accounts(true, false);
```
carllin, line 1573: > My understanding is that we cannot clean past the last full snapshot slot here, but more testing needs to be done.  

We've already cleaned up to the last full snapshot slot `F` when the full snapshot was taken no? When taking the incremental snapshot at `S > F` we have to make sure:

1) We don't clean any updates in slots `<= S` based on updates in roots `> S`

2) Any "cleanable" zero-lamport updates in root `Z`, where `F < Z < S` are not removed, where they typically might be (recall zero-lamport keys are cleanable when the refcount of that account is equal to the number of updates).  This is because if these zero-lamport accounts are missing from the AppendVecs, future incremental snapshots will be missing these updates. <\br>
I think this is the biggest divergence from current behavior where we might have cleaned up these zero lamport accounts when running clean up to slot `Z`, we can't do that anymore unless we introduce some other mechanism of tracking certain accounts have become zero (maybe some other structure for tracking zero lamport accounts needs to be added to the snapshot so that we can do this clean from the AppendVecs)?

runtime/src/snapshot_utils.rs
brooksprumo: I was wondering the same thing myself. The first and last lines are copied from the non-incremental version, and I just left them. Maybe these lines were written before ABS existed? Should either/both of these line be removed/replaced with something else instead?

runtime/src/snapshot_utils.rs
brooksprumo: Good point, will do!

runtime/src/snapshot_utils.rs

runtime/src/serde_snapshot.rs
brooksprumo: Done!

runtime/src/serde_snapshot.rs
brooksprumo: Done!

runtime/src/snapshot_utils.rs
brooksprumo: This function (well, it's FSS version, `bank_to_snapshot_archive()`) is called in two places: (1) In `Validator::new()` before `AccountsBackgroundService` is created, and (2) in `ledger-tool` with the `create-snapshot` option. So neither have ABS.

In the `Validator::new()` path, the bank _does_ get rooted before calling `bank_to_snapshot_archive()`. In the `ledger-tool` path, the bank does _not_ get rooted before calling `bank_to_snapshot_archive()`. So I think the `bank.squash()` line exists because of `ledger-tool`. Do you think this is OK?

runtime/src/snapshot_utils.rs
brooksprumo: Yep yep. So I just pushed another commit that has incremental snapshots working, but relies on preventing cleaning past the last full snapshot slot. After chatting elsewhere, I put in a test to ensure we don't lose zero-lamport accounts that went to zero _after_ the last full snapshot slot. So that test is still good, but now I need to change the clean logic.

Instead of preventing cleaning past the last full snapshot slot, I'll be keeping track of all zero-lamport accounts that went to zero after the last full snapshot slot. So that when an Incremental Snapshot is taken, we'll always pull these back in, so there's no zombie resurrections in subsequent incremental snapshots.

runtime/src/serde_snapshot.rs
brooksprumo: I was thinking maybe a node could get incremental snapshots from the same node it gets the full snapshot from. Would that resolve this?

runtime/src/serde_snapshot.rs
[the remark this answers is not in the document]
brooksprumo: Done!

brooksprumo:
This last commit adds a lot of code, but there is working incremental snapshots now, yay!

After talking with Carl, I need to re-do how clean_accounts() works for ISS though. So, still work to be done on that front. The core/tests/incremental_snapshots.rs has two useful tests for ISS that I think will be good moving forward.

Also, I'll be splitting on smaller pieces of the incremental snapshots code into other PRs to make review easier. Two are up: #18199 and #18207.

Issue #17789, opened 2021-06-07
ClusterSlots uses too much memory


sakridge:
#### Problem

```
const CLUSTER_SLOTS_TRIM_SIZE: usize = 524_288; // 512K
pub type SlotPubkeys = HashMap<Pubkey, u64>;

#[derive(Default)]
pub struct ClusterSlots {
    cluster_slots: RwLock<BTreeMap<Slot, Arc<RwLock<SlotPubkeys>>>>,
```

```
hashbrown::raw::RawTableInner$LT$A$GT$::prepare_resize::hd9de4d7705dc4632
  in /home/stephen_solana_com/solana/target/release/solana-validator
5.89GB consumed over 1807948 calls from:
    0x55b0a435cb61
      in /home/stephen_solana_com/solana/target/release/solana-validator
    _$LT$hashbrown..map..HashMap$LT$K$C$V$C$S$C$A$GT$$u20$as$u20$core..iter..traits..collect..Extend$LT$$LP$K$C$V$RP$$GT$$GT$::extend::hc3281369a9ad10d2
      in /home/stephen_solana_com/solana/target/release/solana-validator
    solana_core::cluster_slots::ClusterSlots::update::h8fcd014c2713807e
      in /home/stephen_solana_com/solana/target/release/solana-validator
    solana_core::cluster_slots_service::ClusterSlotsService::run::h4f4d6ebf13410164
      in /home/stephen_solana_com/solana/target/release/solana-validator
    std::sys_common::backtrace::__rust_begin_short_backtrace::h253287d3ac949d1d
      in /home/stephen_solana_com/solana/target/release/solana-validator
    core::ops::function::FnOnce::call_once$u7b$$u7b$vtable.shim$u7d$$u7d$::h5d585c3e7664364f
      in /home/stephen_solana_com/solana/target/release/solana-validator
    _$LT$alloc..boxed..Box$LT$F$C$A$GT$$u20$as$u20$core..ops..function..FnOnce$LT$Args$GT$$GT$::call_once::hc444a77f8dd8d825
      at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/alloc/src/boxed.rs:1546
      in /home/stephen_solana_com/solana/target/release/solana-validator
    _$LT$alloc..boxed..Box$LT$F$C$A$GT$$u20$as$u20$core..ops..function..FnOnce$LT$Args$GT$$GT$::call_once::h8b68a0a9a2093dfc
[...]

PR #17899, opened 2021-06-11
excludes epoch-slots from nodes with unknown or different shred version

behzadnouri:
#### Problem
Inspecting TDS gossip table shows that crds values of nodes with
different shred-versions are creeping in. Their epoch-slots are
accumulated in ClusterSlots causing bogus slots very far from current
root which are not purged and so cause ClusterSlots keep consuming more
memory:
https://github.com/solana-labs/solana/issues/17789
https://github.com/solana-labs/solana/issues/14366#issuecomment-769896036
https://github.com/solana-labs/solana/issues/14366#issuecomment-832754654

#### Summary of Changes
This commit updates ClusterInfo::get_epoch_slots, and discards entries
from nodes with unknown or different shred-version.

Follow up commits will patch gossip not to waste bandwidth and memory
over crds values of nodes with different shred-version.

### DIFF
--- a/gossip/src/cluster_info.rs
+++ b/gossip/src/cluster_info.rs
@@ -1170,10 +1170,19 @@ impl ClusterInfo {
             .map(map)
     }
 
+    /// Returns epoch-slots inserted since the given cursor.
+    /// Excludes entries from nodes with unkown or different shred version.
     pub fn get_epoch_slots(&self, cursor: &mut Cursor) -> Vec<EpochSlots> {
+        let self_shred_version = self.my_shred_version();
         let gossip = self.gossip.read().unwrap();
         let entries = gossip.crds.get_epoch_slots(cursor);
         entries
+            .filter(
+                |entry| match gossip.crds.get_contact_info(entry.value.pubkey()) {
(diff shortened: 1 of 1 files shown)


carllin:
Ah nice catch! I believe @ryoqun suspected this as well here: https://github.com/solana-labs/solana/issues/14366#issuecomment-769896036, but as @t-nelson pointed out, https://github.com/solana-labs/solana/issues/14366#issuecomment-769915411, we thought shred versions should be keeping networks apart.

Are the push/pull shred version filter in gossip missing something?

behzadnouri:
> Ah nice catch! I believe @ryoqun suspected this as well here: [#14366 (comment)](https://github.com/solana-labs/solana/issues/14366#issuecomment-769896036), but as @t-nelson pointed out, [#14366 (comment)](https://github.com/solana-labs/solana/issues/14366#issuecomment-769915411), we thought shred versions should be keeping networks apart.
> 
> Are the push/pull shred version filter in gossip missing something?

yeah, there are some but generally the shred version checks are not very consistent.
I am working on some changes to purge/drop values if shred versions do not match.

carllin (approved):

PR #17901, opened 2021-06-11
Amend transactions v2 proposal with dependent accounts

jstarry:
#### Problem

One of the short comings of the original tx format was that it wasn't able to explicitly describe CPI dependencies.  The result is the need to pass ALL the accounts and programs needed by an instruction execution chain to the top level program.  Doing so creates a lot more overhead since all the accounts needed below have to serialized/mapped into the program space of each caller.  Most accounts probably don't have to be passed in in their entirety or at least their data does not.  If the tx included a dependent account field then the runtime could keep those accounts in the wings and reference them when needed without requiring the programs to ferry them all along.  We should consider this now that we are bumping the tx format.

_Originally posted by @jackcmay in https://github.com/solana-labs/solana/pull/17103#issuecomment-859776324_

#### Summary of Changes

Add `inner_programs` and `inner_program_accounts` to transaction instruction structure to separate dependent accounts from account inputs needed by the outer instruction

Fixes #


docs/src/proposals/transactions-v2.md
```diff
@@ -158,15 +157,51 @@ pub struct AddressMap {
...
+
+  /// Only programs in this list may be invoked by the outer instruction
+  #[serde(with = "short_vec")]
+  pub inner_programs: Vec<u8>,
+
+  /// Dependent accounts do not need to be mapped to the memory of the outer
+  /// instruction's execution context. They are available for passing to inner
+  /// instructions
+  #[serde(with = "short_vec")]
+  pub inner_program_accounts: Vec<u8>,
```
ryoqun, line 178: I think we can avoid this additional churn to solve the problem described in this pr's description by providing yet another pathway (#17796 ) to access to tx's accounts in addition to the entrypoint one.

docs/src/proposals/transactions-v2.md
```diff
@@ -158,15 +157,51 @@ pub struct AddressMap {
...
+#### Dependent accounts
+
+The new transaction format supports lists of programs and accounts that a
+transaction instruction may use when invoking inner instructions. These programs
+and accounts are separate from transaction instruction account inputs so that
+they do not need to incur the cost of serializing into the outer instruction's
+memory which does need to read or write the account.
+
+For extra user protection, only the programs listed in `inner_programs` may be
+invoked by the outer transaction instruction.
```
ryoqun, line 195: is this strict requirement? is there any concrete thread in mind? if that's the case, I'll try to address this at #17796 as well.

docs/src/proposals/transactions-v2.md
jstarry: This can serve two purposes. One is avoiding the serialization overhead. The other (not mentioned in the PR description) is a stronger framework for isolating the scope of what an instruction can do. If we allow an instruction to dynamically load any account from the transaction, a malicious program might have enough resources for more damage. Imagine for instance, a program dynamically loads the fee payer signer account.

docs/src/proposals/transactions-v2.md
jstarry: I think I already outlined a general threat here: https://github.com/solana-labs/solana/pull/17901#discussion_r650412745 but more concretely: If a user inadvertently calls a compromised or otherwise malicious program in their transaction, sandboxing can help limit what damage may be done. Transactions can restrict which programs are called and which signers can be used in an instruction to avoid a program being able to invoke a system or token program CPI with a signer account.

PR #17913, opened 2021-06-12
add metrics for startup

jeffwashington:
#### Problem
A goal is to speedup startup time of validator. Would be nice to have metrics over more parts.
#### Summary of Changes
Add metric for the major parts of preparing to start. This at least excludes snapshot download, gossip work, and initial replay.
Fixes #

### DIFF
--- a/core/tests/snapshots.rs
+++ b/core/tests/snapshots.rs
@@ -146,7 +146,7 @@ mod tests {
 
         let old_last_bank = old_bank_forks.get(old_last_slot).unwrap();
 
-        let deserialized_bank = snapshot_utils::bank_from_archive(
+        let (deserialized_bank, _timing) = snapshot_utils::bank_from_archive(
             &account_paths,
             &[],
             &old_bank_forks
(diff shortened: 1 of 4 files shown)


jeffwashington:
```
datapoint: bank_from_archive rebuild_bank_from_snapshots_us=60867700i untar_us=35856477i verify_snapshot_bank_us=42710792i total=139434969i
```

jeffwashington:
Looks like I should move this higher still. Startup metrics are likely different than steady state metrics.

ledger-tool/src/main.rs
```diff
@@ -1872,6 +1873,10 @@ fn main() {
                 working_bank.print_accounts_stats();
             }
             println!("Ok");
+            let mut drop_time = Measure::start("drop_time");
+            drop(bank_forks);
+            drop_time.stop();
+            info!("drop(bank_forks): {}us", drop_time.as_us());
```
jeffwashington, line 1879: This drop coincides with a long delay after ledger-tool verify is finished. Reporting this will help the next person verify that this portion of the validator code path is capturing time correctly. Without the explicit drop, the time is when bank_forks goes out of scope and is dropped.

sakridge (approved):
lgtm

PR #17937, opened 2021-06-14
add metrics to handle_snapshot_requests

jeffwashington:
#### Problem
As # accounts grows, handle_snapshot_requests code path execution time grows very large. Having metrics of the total processing time and idle time will help find and verify improvements. For example, if clean and hash calculation can be performed in parallel, then overall time could improve while individual times (hash & clean) could get worse.
#### Summary of Changes
add metrics
Fixes #

### DIFF
--- a/runtime/src/accounts_background_service.rs
+++ b/runtime/src/accounts_background_service.rs
@@ -89,11 +89,13 @@ impl SnapshotRequestHandler {
         accounts_db_caching_enabled: bool,
         test_hash_calculation: bool,
         use_index_hash_calculation: bool,
+        non_snapshot_time_us: u128,
     ) -> Option<u64> {
         self.snapshot_request_receiver
             .try_iter()
             .last()
             .map(|snapshot_request| {
+                let mut total_time = Measure::start("wallclock time elapsed");
                 let SnapshotRequest {
                     snapshot_root_bank,
                     status_cache_slot_deltas,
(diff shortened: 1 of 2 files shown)


jeffwashington:
lemond, against current mnb snapshot
```
datapoint: handle_snapshot_requests-timing hash_time=3370750i flush_accounts_cache_time=583487i shrink_time=39835i clean_time=1110224i snapshot_time=314121i purge_old_snapshots_time=6235i total_us=5424656i idle_time_us=0i
```

runtime/src/accounts_background_service.rs
```diff
@@ -89,11 +89,13 @@ impl SnapshotRequestHandler {
         accounts_db_caching_enabled: bool,
         test_hash_calculation: bool,
         use_index_hash_calculation: bool,
+        non_snapshot_time_us: u128,
     ) -> Option<u64> {
         self.snapshot_request_receiver
             .try_iter()
             .last()
             .map(|snapshot_request| {
+                let mut total_time = Measure::start("wallclock time elapsed");
```
sakridge, line 98: would be nice if this mentions something about snapshots. With nvtx labels "wallclock time elapsed" won't be very descriptive.

sakridge (approved):
small nit about Measure text, but looks good

runtime/src/accounts_background_service.rs
jeffwashington: will handle in separate pr

PR #17939, opened 2021-06-14
verify bank hash on startup with ledger tool option

jeffwashington:
#### Problem
non-index hash calculation is being used in more places with more options, including replacing index scan versions.
#### Summary of Changes
Plumb through more verification compared to index-scan so tests and optional args to validator or ledger-tool verify can check hash calculations.
Fixes #

### DIFF
--- a/ledger/src/bank_forks_utils.rs
+++ b/ledger/src/bank_forks_utils.rs
@@ -143,6 +143,7 @@ fn load_from_snapshot(
         process_options.accounts_db_caching_enabled,
         process_options.limit_load_slot_count_from_snapshot,
         process_options.shrink_ratio,
+        process_options.accounts_db_test_hash_calculation,
     )
     .expect("Load from snapshot failed");
     if let Some(shrink_paths) = shrink_paths {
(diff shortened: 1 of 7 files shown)


ledger/src/bank_forks_utils.rs
```diff
@@ -143,6 +143,7 @@ fn load_from_snapshot(
         process_options.accounts_db_caching_enabled,
         process_options.limit_load_slot_count_from_snapshot,
         process_options.shrink_ratio,
+        process_options.accounts_db_test_hash_calculation,
```
jeffwashington, line 146: here's where we pass true to verify hash calculations further down.

runtime/src/accounts_db.rs
```diff
@@ -4827,19 +4827,23 @@ impl AccountsDb {
...
         let check_hash = true;
         let can_cached_slot_be_unflushed = false;
-        let (calculated_hash, calculated_lamports) = self.calculate_accounts_hash_helper(
-            use_index,
-            slot,
-            ancestors,
-            check_hash,
-            can_cached_slot_be_unflushed,
-        )?;
+        let (calculated_hash, calculated_lamports) = self
```
jeffwashington, line 4837: switch to a lower level function that lets us verify the hash calculation.

sakridge (approved):

PR #17940, opened 2021-06-14
adds mapping from nodes pubkeys to their shred-version

behzadnouri:
#### Problem
Crds values of nodes with different shred versions are creeping into
gossip table resulting in runtime issues as the one addressed in:
https://github.com/solana-labs/solana/pull/17899

#### Summary of Changes
This commit works towards enforcing more checks and filtering based on
shred version by adding necessary mapping and api to gossip table.
Once populated, pubkey->shred-version mapping persists as long as there
are any values associated with the pubkey.

### DIFF
--- a/gossip/Cargo.toml
+++ b/gossip/Cargo.toml
@@ -18,6 +18,7 @@ indexmap = { version = "1.6", features = ["rayon"] }
 itertools = "0.9.0"
 log = "0.4.11"
 lru = "0.6.1"
+matches = "0.1.8"
 num-traits = "0.2"
 rand = "0.7.0"
 rand_chacha = "0.2.2"
--- a/gossip/src/cluster_info.rs
+++ b/gossip/src/cluster_info.rs
@@ -1173,16 +1173,13 @@ impl ClusterInfo {
     /// Returns epoch-slots inserted since the given cursor.
     /// Excludes entries from nodes with unkown or different shred version.
     pub fn get_epoch_slots(&self, cursor: &mut Cursor) -> Vec<EpochSlots> {
-        let self_shred_version = self.my_shred_version();
+        let self_shred_version = Some(self.my_shred_version());
         let gossip = self.gossip.read().unwrap();
         let entries = gossip.crds.get_epoch_slots(cursor);
         entries
-            .filter(
-                |entry| match gossip.crds.get_contact_info(entry.value.pubkey()) {
-                    Some(node) => node.shred_version == self_shred_version,
-                    None => false,
-                },
-            )
+            .filter(|entry| {
+                gossip.crds.get_shred_version(&entry.value.pubkey()) == self_shred_version
+            })
             .map(|entry| match &entry.value.data {
                 CrdsData::EpochSlots(_, slots) => slots.clone(),
                 _ => panic!("this should not happen!"),
@@ -2213,9 +2210,10 @@ impl ClusterInfo {
     ) -> (usize, usize, usize) {
         let len = crds_values.len();
         trace!("PullResponse me: {} from: {} len={}", self.id, from, len);
-        let shred_version = self
-            .lookup_contact_info(from, |ci| ci.shred_version)
-            .unwrap_or(0);
+        let shred_version = {
+            let gossip = self.gossip.read().unwrap();
+            gossip.crds.get_shred_version(from).unwrap_or_default()
+        };
         Self::filter_by_shred_version(
             from,
             &mut crds_values,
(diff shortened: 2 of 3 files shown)


carllin:
Out of curiosity, do we know how these other shred versions are creeping into the gossip table.

It seems like:

1) In pull, we should only making pull requests to other validators with the same shred version or 0: https://github.com/solana-labs/solana/blob/9a5330b7eb617251ad1843b2aa5bfa4f1a8b3f7f/gossip/src/crds_gossip_pull.rs#L298. Could the validators with shred version 0 be cross-pollinating between the clusters?

2) In push, we should only be pushing to other validators with the same shred version: https://github.com/solana-labs/solana/blob/9a5330b7eb617251ad1843b2aa5bfa4f1a8b3f7f/gossip/src/crds_gossip_push.rs#L351

behzadnouri:
> Could the validators with shred version 0 be cross-pollinating between the clusters?

Might be. Apparently it has been intentional to gossip with nodes with shred-version 0:
https://github.com/solana-labs/solana/blob/161838655/gossip/src/cluster_info.rs#L1248

carllin:
Hmm yeah what's more confusing is we already have this `filter_by_shred_version()` function that runs on both pushes: https://github.com/solana-labs/solana/blob/9a5330b7eb617251ad1843b2aa5bfa4f1a8b3f7f/gossip/src/cluster_info.rs#L2365 and pulls: https://github.com/solana-labs/solana/blob/9a5330b7eb617251ad1843b2aa5bfa4f1a8b3f7f/gossip/src/cluster_info.rs#L2203

Inside that function we deliberately don't filter if `shred_version == 0`, and we want to support nodes changing their shred versions: https://github.com/solana-labs/solana/blob/9a5330b7eb617251ad1843b2aa5bfa4f1a8b3f7f/gossip/src/cluster_info.rs#L2269, which seems contradictory to this PR. Should probably figure out if there's some use case for this, and why we make an exception for `shred_version == 0`

carllin:
Also, for some context, it seems we agreed that we could kill the special `0` handling in my PR https://github.com/solana-labs/solana/pull/11620#issuecomment-675163066.

 But because I have the memory of a goldfish 🐟 ,  I'm trying to refresh my memory around the context of this issue. Here was also a brief discord discussion: https://discord.com/channels/428295358100013066/478692221441409024/743945363211288616

behzadnouri:
> but these should be dropped pretty quickly after we set our shred version (unless those validators are using the same keys on both Tds and MB and are staked on both?)

That is one possibility.
I actually think shred-version was added to support validators using the same keys across clusters.

> and we want to support nodes changing their shred versions: which seems contradictory to this PR.

why is it contradictory to this PR?

> Also, for some context, it seems we agreed that we could kill the special 0 handling in my PR #11620 (comment) to try and remove this very same cross-pollination issue.

What happens then when starting a node and not yet have a shred-version?

behzadnouri:
> Hmm yeah what's more confusing is we already have this `filter_by_shred_version()` function that runs on both pushes:

We may receive a crds-value with pubkey `x` from node `a`. `filter_by_shred_version` does not check the shred-version of the owner of the crds-value (i.e the node with pubkey `x`). It only checks the shred-version of the node which is relaying the value (i.e node `a`):
https://github.com/solana-labs/solana/blob/5cc073420/gossip/src/cluster_info.rs#L2274-L2289

So crds-values with different shred versions can still pass through this function as long as they are relayed by a node with matching shred function. I do not know how that first node obtained the crds-value though!

carllin:
>> why is it contradictory to this PR?

Ah I misread the change, I thought once the Pubkey -> shred version mapping was made, we couldn't change it, but it can be modified on insert of a new ContactInfo: 
https://github.com/solana-labs/solana/pull/17940/files#diff-5b67451eccdd9787fc6bfecc40a9d00211eda78ac5adb01c46d4fa61598c8268R204-R208

> >  I do not know how that first node obtained the crds-value though

hmmm yeah this is the confusing bit. And it seems from the EpochSlots crds values must be relatively recent or they would've been purged right? So they must be constantly pushed/refreshed to be alive, which implies some node is a bridge.

>>We may receive a crds-value with pubkey x from node a. filter_by_shred_version does not check the >> shred-version of the owner of the crds-value (i.e the node with pubkey x)

Is it reasonable to add a check for the origin pubkey here to filter them out?

behzadnouri:
> And it seems from the EpochSlots crds values must be relatively recent or they would've been purged right?

I did check them on TDS, and some are very recent, like with wallclock less than 30 seconds ago.

> So they must be constantly pushed/refreshed to be alive, which implies some node is a bridge,

This is true.

> Is it reasonable to add a check for the origin pubkey in filter_by_shred_version to filter those bad apples out?

Yeah, that was the plan to follow up with adding more checks earlier on packets processing using this mapping. This PR is just adding the mapping.

carllin (approved):

t-nelson (approved):

PR #17942, opened 2021-06-14
name arguments to help with confusion

jeffwashington:
#### Problem
Multiple usage patterns result in true, false, etc. in function calls.
#### Summary of Changes
Make a caller that will likely go through 2 changes shortly more explicit.
Fixes #


jbiseda (approved):
LGTM

PR #17943, opened 2021-06-14
Explorer: display message when block rewards not found for interval

oJshua:
#### Problem
There is currently no user feedback for situations where no block rewards are found for a given interval.

#### Summary of Changes
Rather than render an empty table, display a message saying rewards were not found for the fetched interval.

Explorer-Solana (6)

Explorer-Solana (7)

### DIFF
--- a/explorer/src/providers/accounts/rewards.tsx
+++ b/explorer/src/providers/accounts/rewards.tsx
@@ -7,7 +7,7 @@ import { FetchStatus } from "providers/cache";
 import { reportError } from "utils/sentry";
 
 const REWARDS_AVAILABLE_EPOCH = new Map<Cluster, number>([
-  [Cluster.MainnetBeta, 133],
+  [Cluster.MainnetBeta, 132],
   [Cluster.Testnet, 43],
 ]);
 
@@ -95,10 +95,6 @@ async function fetchRewards(
   const lowestAvailableEpoch = REWARDS_AVAILABLE_EPOCH.get(cluster) || 0;
   const connection = new Connection(url);
 
-  if (!fromEpoch && highestEpoch) {
-    fromEpoch = highestEpoch;
-  }
-
   if (!fromEpoch) {
     try {
       const epochInfo = await connection.getEpochInfo();
@@ -115,6 +111,10 @@ async function fetchRewards(
         url,
       });
     }
+
+    if (highestEpoch && highestEpoch < fromEpoch) {
+      fromEpoch = highestEpoch;
+    }
   }
 
   const getInflationReward = async (epoch: number) => {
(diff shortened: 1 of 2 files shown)


explorer/src/providers/accounts/rewards.tsx
```diff
@@ -111,13 +115,20 @@ async function fetchRewards(
...
       const result = await connection.getInflationReward([pubkey], epoch);
-      return result[0];
+      if (result[0] !== null) {
+        return result[0];
+      }
     } catch (error) {
-      if (cluster !== Cluster.Custom) {
+      if (
+        cluster !== Cluster.Custom &&
+        !error.message.match(/Block not available/)
```
jstarry, line 124: Is this an actual response from this rpc method?

explorer/src/providers/accounts/rewards.tsx
```diff
@@ -111,13 +115,20 @@ async function fetchRewards(
...
+      ) {
         reportError(error, { url });
       }
     }
-    return null;
+    return {
+      epoch,
+    };
```
jstarry, line 131: Shouldn't the return value be different between a null response and a failed request?

explorer/src/providers/accounts/rewards.tsx
```diff
@@ -8,15 +8,19 @@ import { reportError } from "utils/sentry";
 
 const PAGE_SIZE = 15;
 
+export type NoReward = {
+  epoch: number;
```
jstarry, line 12: What's the purpose of this over using `null`? Doesn't look like `epoch` is used anywhere

jstarry (commented):
@oJshua I imagine there's a known epoch on each cluster where we know rewards aren't available, can we hardcode that to disable the "Load More" button at a certain point? 

On a separate note, looks like we're thrashing with each other on the `package-lock.json`. You probably want to update your `npm` to avoid these big diffs.

explorer/src/providers/accounts/rewards.tsx
oJshua: This was used in the reconciler: https://github.com/solana-labs/solana/blob/master/explorer/src/providers/accounts/rewards.tsx#L42-L43 but I think I'll go back to using null instead.

oJshua:
Ah, I was on a different machine than normal. Thrashing resolved!

I'll add the hardcoded epochs for when rewards began.

explorer/src/providers/accounts/rewards.tsx
oJshua: The full response is something like 'Failed to fetch inflation reward: Block not available for...', which should no longer happen on my latest commit.

jstarry (approved):

PR #17949, opened 2021-06-14
remove unused parameters from CalculateHashIntermediate

jeffwashington:
#### Problem
CalculateHashIntermediate has been refactored to not use slot or write version. The old constructor was left in place so tests didn't noisly change yet.
#### Summary of Changes
Get rid of unused parameters on the tests.
Fixes #


brooksprumo (approved):
Looks good to me

PR #17952, opened 2021-06-15
Update init-metrics.sh

axleiro:
#### Problem
In the new InfluxDB architecture  port 8086 is the metrics server that acts as a load balancer, /write will go through influx relay  and then to both influxdb intances, so we have the same metrics on both, however a /query will go to just one of them. That script only uses /query so it'll just go to one instance. To avoid that, I've forwarded 8087 directly to our influx1 instance, you can call this one our old influxdb instance, so it'll work as before

#### Summary of Changes
Change port 8086 to 8087

### DIFF
--- a/net/init-metrics.sh
+++ b/net/init-metrics.sh
@@ -29,7 +29,7 @@ EOF
 useEnv=false
 delete=false
 createWithoutConfig=false
-host="https://metrics.solana.com:8086"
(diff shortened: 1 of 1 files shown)


joeaba (approved):
Looks good, lets check if something else uses 8086 and /query to change that too

PR #17974, opened 2021-06-15
ledger tool limit_load_slot_count_from_snapshot avoids assert failures

jeffwashington:
#### Problem
It is useful to load snapshots with large #s of accounts and set limits on how many slots are loaded to gather metrics. The ledger-tool option limit_load_slot_count_from_snapshot enables this. For example, load 200k slots from a snapshot with 360M accounts allows loading of just ~180M accounts. However, this produces various asserts early in the load process. These are annoying and require manually commenting stuff out to get metrics.
#### Summary of Changes
Pass a parameter that allows lobotomizing a few asserts to allow progress even with 'invalid' system state.
Fixes #

### DIFF
--- a/ledger/src/bank_forks_utils.rs
+++ b/ledger/src/bank_forks_utils.rs
@@ -150,10 +150,6 @@ fn load_from_snapshot(
         deserialized_bank.set_shrink_paths(shrink_paths);
     }
 
-    if process_options.accounts_db_test_hash_calculation {
-        deserialized_bank.update_accounts_hash_with_index_option(false, true);
-    }
-
     let deserialized_bank_slot_and_hash = (
         deserialized_bank.slot(),
         deserialized_bank.get_accounts_hash(),
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -1015,6 +1015,7 @@ impl Bank {
             AccountSecondaryIndexes::default(),
             false,
             AccountShrinkThreshold::default(),
+            false,
         )
     }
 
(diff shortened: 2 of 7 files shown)


ledger/src/bank_forks_utils.rs
```diff
@@ -149,10 +149,6 @@ fn load_from_snapshot(
         deserialized_bank.set_shrink_paths(shrink_paths);
     }
 
-    if process_options.accounts_db_test_hash_calculation {
```
jeffwashington, line 152: this is no longer necessary here since we now use the non-index scan and support validation of index and non-index scans.

runtime/src/bank.rs
```diff
@@ -4262,12 +4276,14 @@ impl Bank {
                 .feature_builtins
                 .extend_from_slice(&additional_builtins.feature_builtins);
         }
-        for builtin in builtins.genesis_builtins {
-            self.add_builtin(
-                &builtin.name,
-                builtin.id,
-                builtin.process_instruction_with_context,
-            );
+        if !debug_do_not_add_builtins {
```
jeffwashington, line 4279: this is what we need to lobotomize so we don't try to add builtin accounts that normally would already exist in a full account set load.

runtime/src/serde_snapshot.rs
```diff
@@ -271,12 +271,16 @@ where
...
+
+    // if limit_load_slot_count_from_snapshot is set, then we need to side-step some correctness checks beneath this call
+    let debug_do_not_add_builtins = limit_load_slot_count_from_snapshot.is_some();
     let bank = Bank::new_from_fields(
         bank_rc,
         genesis_config,
         bank_fields,
         debug_keys,
         additional_builtins,
+        debug_do_not_add_builtins,
```
jeffwashington, line 283: here's where we pass 'true' if the cli was told to limit slot count.

sakridge (approved):

PR #17980, opened 2021-06-15
break out generate index reporting

jeffwashington:
#### Problem
We are about to refactor generate_index to pipeline it with other startup functionality. The logging code is large, complicated, and obfuscates the real work going on in the function.
#### Summary of Changes
Separate out the logging code.
Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -189,6 +189,52 @@ impl ZeroLamport for AccountInfo {
     }
 }
 
+struct MultiThreadProgress<'a> {
+    last_update: Instant,
+    my_last_report_count: u64,
+    total_count: &'a AtomicU64,
+    report_delay_secs: u64,
+    first_caller: bool,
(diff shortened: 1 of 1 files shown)


runtime/src/accounts_db.rs
```diff
@@ -189,6 +189,52 @@ impl ZeroLamport for AccountInfo {
     }
 }
 
+struct MultiThreadProgress<'a> {
+    last_update: Instant,
+    my_last_report_count: u64,
+    total_count: &'a AtomicU64,
+    report_delay_secs: u64,
```
steviez, line 196: Nitpick - what you do you think about `report_interval_secs` ?

steviez (approved):
Looks good to me!

runtime/src/accounts_db.rs
jeffwashington: good suggestion. Hopefully I fix that when this code next changes.

PR #17982, opened 2021-06-15
refactor generate_index process_storage_slot

jeffwashington:
#### Problem
We are about to refactor generate_index to pipeline it with other startup tasks.
#### Summary of Changes
Refactor first half.
Fixes #


brooksprumo (approved):
lgtm

PR #17983, opened 2021-06-15
generate_index timings

jeffwashington:
#### Problem
We are about to rework and pipeline generate_index with other stages of startup.
#### Summary of Changes
separate timings for generate_index into its own struct, like other places we capture timings.
Fixes #


jbiseda (approved):
Would it be useful to add a trait for report() calls used by similar structures? Perhaps as a separate issue.

PR #17986, opened 2021-06-15
move all sort to sort_unstable

jeffwashington:
#### Problem
clippy disable is required to use old 'sort' vs 'sort_unstable' in all cases where we call 'sort'. It is acceptable to call 'sort_unstable' in all these cases. 'sort_unstable' is recommended and faster.
#### Summary of Changes
Convert all remaining 'sort' calls to 'sort_unstable' and remove clippy work-around.
Fixes #


jeffwashington:
@mvines prefers site clippy suppressions and 'sort' in case the code is copied and applied to non-primitive types. So, closing.

PR #17988, opened 2021-06-15
refactor rebuild_bank_from_snapshots

jeffwashington:
#### Problem
We are about to rework snapshot->generate_index at startup. The validation and initial code in rebuild_bank_from_snapshots is simple and used one time and has value as a stand alone function.
#### Summary of Changes
Factor it out.
Fixes #


brooksprumo (approved):
lgtm

PR #18016, opened 2021-06-16
Remove "synthetic" from sysvar definition

garious:
#### Problem

sysvar definition may be confusing to those familiar with synthetic tokens

#### Summary of Changes

Remove "synthetic" from the sysvar definition


ronielex:
Please join me

PR #18021, opened 2021-06-16
introduce LockMapType for accounts_index

jeffwashington:
#### Problem
Create type for common complicated generic type that is also likely to change over time.
#### Summary of Changes

Fixes #


taozhu-chicago (approved):
made code cleaner, looks good to me

PR #18022, opened 2021-06-16
move zero lamport list out of per-account for loop

jeffwashington:
#### Problem
Account index insertion is the bottleneck on startup time.
#### Summary of Changes
Move code out of the per-account loop.
Fixes #


jeffwashington:
measurements show a 1% improvement. Not huge, but this code path is the limiting factor on startup time. Makes sense to move operations out of the critical path per-account loop as # accounts grows.

sakridge:
This is removed by https://github.com/solana-labs/solana/pull/17601 which is close to be ready to merged.

PR #18030, opened 2021-06-17
Removed pub from  some functions which are actually private to improve encapsulation

lijunwangs:
#### Problem
A few functions in bank.rs are marked as pub but -- not used anywhere else outside of the module. Remove the pub marker to improve encapsulation.

#### Summary of Changes
Remove the pub marker to improve encapsulation. Readability improvement only, no functional impact.
Fixes #


sakridge (approved):
lgtm

PR #18037, opened 2021-06-17
Websocket are you reopening or not? When no subscription at first

Arrowana:
#### Problem

Witchcraft exists in dapp-scaffold with no clear relevance
https://github.com/solana-labs/dapp-scaffold/blob/9352c93fe34fb061eed33030dc406074295a69b0/src/contexts/connection.tsx#L129-L161

"The websocket library solana/web3.js closes its websocket connection when the subscription list is empty after opening its first time, preventing subsequent subscriptions from receiving responses."

#### Summary of Changes
Let's add a test to guarantee the behavior is actually well taken care of in web3.js. We can ditch the "hack" in consumer code

Fixes #

### DIFF
--- a/web3.js/test/websocket.test.ts
+++ b/web3.js/test/websocket.test.ts
@@ -79,5 +79,22 @@ if (process.env.TEST_LIVE) {
 
       await connection.removeSignatureListener(id);
     });
+
+    it('reconnects once subscription are created', async () => {
+      // wait for websocket to disconnect
+      await sleep(1100);
+      expect(connection._rpcWebSocketConnected).to.be.false;
+      expect(connection._rpcWebSocketIdleTimeout).to.eq(null);
+
+      const testSignature = bs58.encode(Buffer.alloc(64));
+      const id = connection.onSignature(testSignature, () => {});
+      
+      // wait for websocket to connect
+      await sleep(100);
+      expect(connection._rpcWebSocketConnected).to.be.true;
+      expect(connection._rpcWebSocketIdleTimeout).to.eq(null);
+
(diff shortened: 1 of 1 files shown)


web3.js/test/websocket.test.ts
```diff
@@ -79,5 +79,24 @@ if (process.env.TEST_LIVE) {
 
       await connection.removeSignatureListener(id);
     });
+
+    it('reconnects once subscription are created', async () => {
+      const newConnection = new Connection(url);
+
+      // wait for websocket to disconnect
```
jstarry, line 86: websocket will start disconnected and will only connect when a subscription is added

web3.js/test/websocket.test.ts
```diff
@@ -79,5 +79,24 @@ if (process.env.TEST_LIVE) {
...
+
+      const testSignature = bs58.encode(Buffer.alloc(64));
+      const id = newConnection.onSignature(testSignature, () => {});
+
+      // wait for websocket to connect
+      await sleep(100);
+      expect(newConnection._rpcWebSocketConnected).to.be.true;
+      expect(newConnection._rpcWebSocketIdleTimeout).to.eq(null);
+
+      await newConnection.removeSignatureListener(id);
```
jstarry, line 99: Can you please also add a test that we can reconnect again after disconnection?

jstarry (commented):
Thanks for adding this!

jstarry:
@Arrowana do you mind taking another look at this?

Arrowana:
> @Arrowana do you mind taking another look at this?

Will do, sorry i completely missed that you added some comments

PR #18055, opened 2021-06-18
stake_program: Fix build issues downstream due to stake state shuffle

mvines:


programs/stake/src/stake_state.rs
```diff
@@ -18,7 +18,6 @@ use {
             config::Config,
             instruction::{LockupArgs, StakeError},
             program::id,
-            state::{Authorized, Delegation, Lockup, Meta, Stake, StakeAuthorize, StakeState},
```
mvines, line 21: Apparently this will override `pub use solana_sdk::stake::state::*;` line below

mvines:
FYI @joncinque, we'd run into this problem in SPL if we upgrade to 1.7.2 then try to rip out the stake state duplicated definitions in the stake-pool program.  So we'll need to wait until 1.7.3 ships with this PR before attempting that

joncinque:
@mvines  sorry about that, how did you catch this?

PR #18065, opened 2021-06-18
Fix Nightly Clippy Warnings

Lichtso:
#### Problem
In the future automatic dereferencing on a function call will become a warning.

#### Summary of Changes
Let clippy and fmt fix all of these.

Fixes #


mvines:
Thanks

mvines:
(attempting a v1.7 backport since this as a low risk, wide reaching change and likely to cause future 1.7 merge conflicts)

PR #18066, opened 2021-06-18
adds shred-version to ip-echo-server response

behzadnouri:
#### Problem
When starting a validator, the node initially joins gossip with
shred_verison = 0, until it adopts the entrypoint's shred-version:
https://github.com/solana-labs/solana/blob/9b182f408/validator/src/main.rs#L417

Depending on the load on the entrypoint, this adopting entrypoint
shred-version through gossip sometimes becomes very slow, and causes
several problems in gossip because we have to partially support
shred_version == 0 which is a source of leaking crds values from one
cluster to another. e.g. see
https://github.com/solana-labs/solana/pull/17899
and the other linked issues there.

#### Summary of Changes
In order to remove shred_version == 0 from gossip, this commit adds
shred-version to ip-echo-server response. Once the entrypoints are
updated, on validator start-up, if --expected_shred_version is not
specified we will obtain shred-version from the entrypoint using
ip-echo-server.

### DIFF
--- a/net-utils/src/bin/ip_address_server.rs
+++ b/net-utils/src/bin/ip_address_server.rs
@@ -19,7 +19,7 @@ fn main() {
         .unwrap_or_else(|_| panic!("Unable to parse {}", port));
     let bind_addr = SocketAddr::from(([0, 0, 0, 0], port));
     let tcp_listener = TcpListener::bind(bind_addr).expect("unable to start tcp listener");
-    let _runtime = solana_net_utils::ip_echo_server(tcp_listener);
+    let _runtime = solana_net_utils::ip_echo_server(tcp_listener, /*shred_version=*/ None);
     loop {
         std::thread::park();
     }
--- a/net-utils/src/ip_echo_server.rs
+++ b/net-utils/src/ip_echo_server.rs
@@ -1,8 +1,13 @@
 use {
-    crate::{ip_echo_server_reply_length, HEADER_LENGTH},
+    crate::{HEADER_LENGTH, IP_ECHO_SERVER_RESPONSE_LENGTH},
     log::*,
     serde_derive::{Deserialize, Serialize},
-    std::{io, net::SocketAddr, time::Duration},
+    solana_sdk::deserialize_utils::default_on_eof,
+    std::{
+        io,
+        net::{IpAddr, SocketAddr},
+        time::Duration,
+    },
     tokio::{
         io::{AsyncReadExt, AsyncWriteExt},
         net::{TcpListener, TcpStream},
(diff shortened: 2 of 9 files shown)


net-utils/src/ip_echo_server.rs
```diff
@@ -23,6 +27,14 @@ pub(crate) struct IpEchoServerMessage {
     udp_ports: [u16; MAX_PORT_COUNT_PER_MESSAGE], // Fixed size list of ports to avoid vec serde
 }
 
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
+pub struct IpEchoServerResponse {
+    // Public IP address of request echoed back to the node.
+    pub(crate) address: IpAddr,
+    // Cluster shred-version of the node running the server.
+    pub(crate) shred_version: Option<u16>,
```
mvines, line 35: Oh I think we might be able to use `#[serde(deserialize_with = "default_on_eof")]` here so that the new `IpEchoServerResponse` retains compatibility with the current endpoints that will only return  `address: IpAddr`

behzadnouri: added, thanks

net-utils/src/bin/ip_address_server.rs
```diff
@@ -19,7 +19,7 @@ fn main() {
         .unwrap_or_else(|_| panic!("Unable to parse {}", port));
     let bind_addr = SocketAddr::from(([0, 0, 0, 0], port));
     let tcp_listener = TcpListener::bind(bind_addr).expect("unable to start tcp listener");
-    let _runtime = solana_net_utils::ip_echo_server(tcp_listener);
+    let _runtime = solana_net_utils::ip_echo_server(tcp_listener, /*shred_version=*/ None);
```
mvines, line 22: I think it'd be ok if hard coded a shred_version here, say "0" or "123", which I think would allow the removal of the `Option` from `ip_echo_server()`

net-utils/src/lib.rs
```diff
@@ -102,7 +99,14 @@ fn ip_echo_server_request(
 /// Determine the public IP address of this machine by asking an ip_echo_server at the given
 /// address
 pub fn get_public_ip_addr(ip_echo_server_addr: &SocketAddr) -> Result<IpAddr, String> {
-    ip_echo_server_request(ip_echo_server_addr, IpEchoServerMessage::default())
+    let resp = ip_echo_server_request(ip_echo_server_addr, IpEchoServerMessage::default())?;
+    Ok(resp.address)
+}
+
+pub fn get_cluster_shred_version(ip_echo_server_addr: &SocketAddr) -> Result<u16, String> {
```
mvines, line 106: I think I'd rather roll this into `get_public_ip_addr()` instead of a separate public `get_cluster_shred_version()` function that somebody will 100% use after `get_public_ip_addr()`, causing a 2x load on the entrypoint

validator/src/main.rs
```diff
@@ -2145,6 +2145,8 @@ pub fn main() {
         expected_bank_hash: matches
             .value_of("expected_bank_hash")
             .map(|s| Hash::from_str(s).unwrap()),
+        // TODO: Once entrypoints are updated to return shred-version, obtain
+        // cluster shred version from ip-addr-server if this is None.
```
mvines, line 2149: Couldn't we plumb this code in this PR as well to avoid adding this TODO?  The current entrypoints would return `None` and `expected_shred_version` then remains None

mvines (approved):
lgtm!

PR #18072, opened 2021-06-18
filters crds values obtained through gossip by their shred version

behzadnouri:
#### Problem
filter_by_shred_version does not check the shred-version of the owner of
the crds-value. It only checks the shred-version of the node which is
relaying the value:
https://github.com/solana-labs/solana/blob/5cc073420/gossip/src/cluster_info.rs#L2274-L2289

So crds-values with different shred versions can still pass through this
function as long as they are relayed by a node with matching shred
version; and so, a single node can bridge different shred values
through-out the cluster.

see comments in https://github.com/solana-labs/solana/pull/17940

#### Summary of Changes
* Check the shred-version on the values in addition to node which sends the crds value.
* Aggregate all shred-version checks in one place and very early in gossip handling of packets.

### DIFF
--- a/gossip/src/cluster_info.rs
+++ b/gossip/src/cluster_info.rs
@@ -16,7 +16,7 @@ use {
     crate::{
         cluster_info_metrics::{submit_gossip_stats, Counter, GossipStats, ScopedTimer},
         contact_info::ContactInfo,
-        crds::Cursor,
+        crds::{Crds, Cursor},
         crds_gossip::CrdsGossip,
         crds_gossip_error::CrdsGossipError,
         crds_gossip_pull::{CrdsFilter, ProcessPullStats, CRDS_GOSSIP_PULL_CRDS_TIMEOUT_MS},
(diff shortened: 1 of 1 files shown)


gossip/src/cluster_info.rs
```diff
@@ -3102,6 +3047,70 @@ pub fn stake_weight_peers(
...
 }
 
+// Filters out values from nodes with different shred-version.
+fn check_shred_version(
+    mut msg: Protocol,
+    self_shred_version: u16,
+    crds: &Crds,
+    stats: &GossipStats,
+) -> Option<Protocol> {
+    let filter_values = |from: &Pubkey, values: &mut Vec<CrdsValue>, skipped_counter: &Counter| {
```
carllin, line 3057: nit: filter_values -> filter_on_shred_version

gossip/src/cluster_info.rs
```diff
@@ -3102,6 +3047,70 @@ pub fn stake_weight_peers(
     ClusterInfo::sorted_stakes_with_index(peers, stakes)
 }
 
+// Filters out values from nodes with different shred-version.
+fn check_shred_version(
```
carllin, line 3051: nit: check_shred_version -> filter_on_shred_version

carllin (dismissed):
Looks great, thanks!

gossip/src/cluster_info.rs
behzadnouri: done

PR #18079, opened 2021-06-18
Bump bpf-tools version to 1.11

### LINKED ISSUE #17758: ELF files contain local paths
#### Problem

ELF files generated with the bpf-tools contain local non-relative paths and this can not only reveal user information but also makes it more difficult to compare ELFs across machines.  (though the latter is less of a concern since we strip the ELFs before deployment)

#### Proposed Solution

Remove local absolute paths from the find ELF files

dmakarov:
Avoid including local paths in compiler generated ELF files

#### Summary of Changes

Fixes #17758


solana-grimes:

PR #18082, opened 2021-06-19
Add repair and recovered metrics per slot

sakridge:
#### Problem

No repair metrics per-slot makes it hard to see if a block was slow because of repair.

#### Summary of Changes

Add repaired and recovered metrics per-slot.

Fixes #

### DIFF
--- a/ledger/src/blockstore.rs
+++ b/ledger/src/blockstore.rs
@@ -45,7 +45,7 @@ use std::{
     borrow::Cow,
     cell::RefCell,
     cmp,
-    collections::{HashMap, HashSet},
+    collections::{BTreeMap, HashMap, HashSet},
     convert::TryInto,
     fs,
     io::{Error as IoError, ErrorKind},
@@ -56,6 +56,7 @@ use std::{
         mpsc::{sync_channel, Receiver, SyncSender, TrySendError},
         Arc, Mutex, RwLock, RwLockWriteGuard,
     },
+    time::Instant,
 };
 use thiserror::Error;
 use trees::{Tree, TreeWalk};
@@ -148,6 +149,27 @@ pub struct Blockstore {
     pub completed_slots_senders: Vec<CompletedSlotsSender>,
     pub lowest_cleanup_slot: Arc<RwLock<Slot>>,
     no_compaction: bool,
+    slots_stats: Arc<Mutex<SlotsStats>>,
+}
+
+struct SlotsStats {
+    last_cleanup_ts: Instant,
+    stats: BTreeMap<Slot, SlotStats>,
+}
+
+impl Default for SlotsStats {
+    fn default() -> Self {
+        SlotsStats {
+            last_cleanup_ts: Instant::now(),
+            stats: BTreeMap::new(),
+        }
+    }
+}
+
+#[derive(Default)]
+struct SlotStats {
+    num_repaired: usize,
+    num_recovered: usize,
 }
 
 pub struct IndexMetaWorkingSetEntry {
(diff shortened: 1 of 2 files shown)


ledger/src/blockstore.rs
```diff
@@ -148,6 +148,13 @@ pub struct Blockstore {
     pub completed_slots_senders: Vec<CompletedSlotsSender>,
     pub lowest_cleanup_slot: Arc<RwLock<Slot>>,
     no_compaction: bool,
+    slot_stats: Arc<Mutex<(u64, HashMap<Slot, SlotStats>)>>,
```
behzadnouri, line 151: can you please add a comment what `u64` is here?
It seems like it is the `timestamp` of last purge. Might also be better to use `std::time::Instant`.

ledger/src/blockstore.rs
```diff
@@ -1426,6 +1459,8 @@ impl Blockstore {
         data_index: &mut ShredIndex,
         shred: &Shred,
         write_batch: &mut WriteBatch,
+        is_recovered: bool,
+        is_repaired: bool,
```
behzadnouri, line 1463: Two `bool` arguments next to each other lacks type-safety and might be bug prone.
Also, I am guessing these are mutually exclusive. If so might be good to define some enum here like:
```
enum ShredSource {
   UNSPECIFIED,
   TURBINE,
   RECOVERED,
   REPAIRED,
   ....
}
```

ledger/src/blockstore.rs
```diff
@@ -1476,7 +1511,29 @@ impl Blockstore {
...
+                e.num_recovered += 1;
+            }
+        }
         if slot_meta.is_full() {
+            let (num_repaired, num_recovered) = {
+                let mut slot_stats = self.slot_stats.lock().unwrap();
+                if let Some(e) = slot_stats.1.remove(&slot_meta.slot) {
+                    if solana_sdk::timing::timestamp().saturating_sub(slot_stats.0) > 30 * 1000 {
+                        let root = self.last_root();
+                        slot_stats.1.retain(|slot, _stats| *slot > root);
```
behzadnouri, line 1529: I do not know how big we expect this hash-map to be. Might be better to use `BTreeMap::split_off` instead.
Maybe, instead of cleaning up every 30 seconds, also condition this purging logic on the size of the `HashMap`.

behzadnouri (dismissed):
some comments, but LGTM

ledger/src/blockstore.rs
sakridge: It could yea, it's just annoying `Instant` needs a custom Default impl.

ryoqun:
well, I might noticed that this introduced overhead probably (~ 100ms) due to maybe lock contention? (could be other prs.). haven't looked closely.

The highlighted node (Arvv3) is running v1.7.10 against mainnet-beta among the vast majority of v1.6.x nodes.

`insert_shred_is_full` (circle markers) looks like delayed from others.

image

ryoqun:
and this is testnet cluster.

`F3S8X...` and `FIEHv` are v1.6.x nodes in testnet. and it looks like they're generally faster then the rest (v1.7.x) in this example.

image

sakridge:
Wow, 100ms would be really incredibly slow and surprising for just these metrics.

ryoqun:
> Wow, 100ms would be really incredibly slow and surprising for just these metrics.

I'm looking into this now.

ryoqun:
(well, I'm looking into this because I'm about to add another metrics to precisely monitor turbine packet drops)

ryoqun:
breaking news: reverting this pr doesn't indeed fix the issue.. umm, there should be regression somewhere else...

ryoqun:
@sakridge how about backporting this to v1.6? I think this is pretty safe, considering it's been awhile on v1.7. I just want these metrics for the rainbow tool. :)

PR #18096, opened 2021-06-21
Add execution batch stats

sakridge:
#### Problem

Batch execution size not very visible from metrics.

#### Summary of Changes

Add the number of executed batches and the total batch size

Fixes #


jeffwashington (approved):
lgtm

PR #18101, opened 2021-06-21
Use newly stabilized Duration methods/consts

CriesofCarrots:
#### Problem
We added a hand-rolled trait to support saturating arithmetic methods and the MAX const on Durations; but these methods are now available in stable Rust.

#### Summary of Changes
Use Duration::MAX, Duration::saturating_mul, and Duration::saturating_sub from Rust core, and remove arithmetic module.


CriesofCarrots:
Blocked on https://github.com/solana-labs/solana/pull/18106

CriesofCarrots:
Blocked on [bpf-tools](https://github.com/solana-labs/bpf-tools) update to rust v1.53.0

PR #18103, opened 2021-06-21
Added `localhost` to the list of clusters

imprfekt:
#### Problem
When running a local node for development, `clusterApiUrl()` cannot be used, because it doesn't support `localhost` as a parameter. Network endpoint needs to be hardcoded instead.
#### Summary of Changes
Added `localhost` to the list of clusters.


mvines:
The custom URL feature already serves this purpose actually

imprfekt:
> The custom URL feature already serves this purpose actually

Correct, and this PR allows for something like this:
```
const cluster = env.get("CLUSTER"); //env variable => 'devnet' | 'testnet' | 'mainnet-beta' | 'localhost'
const explorerURL = "https://explorer.solana.com?custom&customUrl=" + clusterApiUrl(cluster)
```
Without implementing a custom logic for local, devnet, testnet, mainnet environments.

imprfekt:
@mvines if you don't think this adds any value, feel free to close it, I have no problem with that.

jstarry:
I think it adds some value, but adding a new value to the `Cluster` pseudo-enum is kind of a breaking change and `'localhost'` wouldn't work well out of the box for `clusterApiUrl` because the default is to use `https`. I think it's better to keep it as is, thanks anyways @imprfekt!

imprfekt:
Cheers!

Issue #17178, opened 2021-05-11
Writing account state on every slot inefficient and reduces incremental snapshot efficiency


sakridge:
#### Problem

Some percentage of the accounts is rewritten on each slot.
https://github.com/solana-labs/solana/blob/d4ffd909a4f06318387eb74bf21ff58f5c596284/runtime/src/bank.rs#L3574

This limits the number of append-vecs, but it creates more write IO and makes incremental snapshots larger since there are updates to accounts which are not necessary.

#### Proposed Solution

Don't rewrite on each slot. Be able to combine cross-slot accounts into a single append-vec.

sakridge:
Related to https://github.com/solana-labs/solana/issues/17088

ryoqun:
> it creates more write IO and makes incremental snapshots larger since there are updates to accounts which are not necessary.

yeah, this is kind of waste. but it was easy to address infinite append vec issue and introduce stale acocunt's rent collection and periodic on-chain whole-state check at the time...

I think some slot-based reflink mechanism should alleviate the situation while combining older appendvecs. So, we keep to be apparently rewriting the accounts for inclusion to bank hash for on-chain verification. The trick is that the previous updated slot for given account is deterministic across the cluster. So, we could even embed the slot pointer into delta snapshots.

Or rather, we just can give up this on-chain state check altogether and completely relying on background accounts hash check? I rather avoid it for the very blockchain existence.

sakridge:
> > it creates more write IO and makes incremental snapshots larger since there are updates to accounts which are not necessary.
> 
> yeah, this is kind of waste. but it was easy to address infinite append vec issue and introduce stale acocunt's rent collection and periodic on-chain whole-state check at the time...
> 
> I think some slot-based reflink mechanism should alleviate the situation while combining older appendvecs. So, we keep to be apparently rewriting the accounts for inclusion to bank hash for on-chain verification. The trick is that the previous updated slot for given account is deterministic across the cluster. So, we could even embed the slot pointer into delta snapshots.
> 
> Or rather, we just can give up this on-chain state check altogether and completely relying on background accounts hash check? I rather avoid it for the very blockchain existence.

Yea. At the time it was the right thing. I think we are getting to the point where we need to re-think the compaction strategy though because of the accounts growth and need for incremental snapshots.

brooksprumo:
[...]

PR #18111, opened 2021-06-21
bin accounts index

jeffwashington:
#### Problem
Startup time is slow. Account index generation is slow and single threaded. Account index is an area of much focus now and going forward towards 1B accounts.

Performance of btree degrades with # of entries. Reducing the # of entries/btree improves performance. Btree insertion is also single-threaded and exclusive, so inserting into the btree can be very expensive. Creating many individual btrees can allow insertions to be sorted and batch inserted into an individual btree in a thread while other items can be inserted in other bins simultaneously.
Lock contention can be reduced, startup time can drastically improve, lookup time can improve by 2x.
#### Summary of Changes
Internally, rework the account index to be:
```
Vec<RwLock<btree>>
```
instead of:
```
RwLock<btree>
```

A pubkey's value is used to determine which btree it belongs in. Using byte 0 of the pubkey's value retains sort order for ordered scan.
Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -2097,7 +2097,7 @@ impl AccountsDb {
         );
     }
 
-    fn do_shrink_slot_stores<'a, I>(&'a self, slot: Slot, stores: I, is_startup: bool) -> usize
+    fn do_shrink_slot_stores<'a, I>(&'a self, slot: Slot, stores: I, _is_startup: bool) -> usize
     where
         I: Iterator<Item = &'a Arc<AccountStorageEntry>>,
     {
@@ -2139,23 +2139,10 @@ impl AccountsDb {
         let mut index_read_elapsed = Measure::start("index_read_elapsed");
         let mut alive_total = 0;
 
-        let accounts_index_map_lock = if is_startup {
-            // at startup, there is nobody else to contend with the accounts_index read lock, so it is more efficient for us to keep it held
-            Some(self.accounts_index.get_account_maps_read_lock())
-        } else {
-            None
-        };
-        let accounts_index_map_lock_ref = accounts_index_map_lock.as_ref();
-
         let mut alive_accounts: Vec<_> = Vec::with_capacity(stored_accounts.len());
         let mut unrefed_pubkeys = vec![];
         for (pubkey, stored_account) in &stored_accounts {
-            let lookup = if is_startup {
-                self.accounts_index
-                    .get_account_read_entry_with_lock(pubkey, accounts_index_map_lock_ref.unwrap())
-            } else {
-                self.accounts_index.get_account_read_entry(pubkey)
-            };
+            let lookup = self.accounts_index.get_account_read_entry(pubkey);
             if let Some(locked_entry) = lookup {
                 let is_alive = locked_entry.slot_list().iter().any(|(_slot, i)| {
                     i.store_id == stored_account.store_id
@@ -2175,7 +2162,6 @@ impl AccountsDb {
             }
         }
 
-        drop(accounts_index_map_lock);
         index_read_elapsed.stop();
         let aligned_total: u64 = Self::page_align(alive_total as u64);
 
@@ -4458,10 +4444,9 @@ impl AccountsDb {
         let keys: Vec<_> = self
             .accounts_index
             .account_maps
-            .read()
-            .unwrap()
-            .keys()
-            .cloned()
+            .iter()
+            .map(|btree| btree.read().unwrap().keys().cloned().collect::<Vec<_>>())
+            .flatten()
             .collect();
         collect.stop();
 
@@ -5973,7 +5958,13 @@ impl AccountsDb {
         }
 
         let mut stored_sizes_and_counts = HashMap::new();
-        for account_entry in self.accounts_index.account_maps.read().unwrap().values() {
+        for account_entry in self
+            .accounts_index
+            .account_maps
+            .iter()
+            .map(|i| i.read().unwrap().values().cloned().collect::<Vec<_>>())
+            .flatten()
+        {
             for (_slot, account_entry) in account_entry.slot_list.read().unwrap().iter() {
                 let storage_entry_meta = stored_sizes_and_counts
                     .entry(account_entry.store_id)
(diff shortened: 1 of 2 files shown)


jeffwashington:
some metrics:
https://docs.google.com/spreadsheets/d/17CBXDmYs2lCkBftYTxkyFYUFTODhm0uWlp8uqGhGflI/edit?usp=sharing

jeffwashington:
Some thinking adapted from other discussions. Specifically related to attributes of dashmap and btree.

1. We know and control the sharding/binning in our data set and can group items by shard/bin prior to getting a lock. Thus, we can insert into each shard in a separate thread. This greatly improves startup time. Last I checked, sharding in dashmap was a feature which has to be enabled and is more internal, but perhaps I am misunderstanding the conventions. We also control the locks per shard in our case, meaning we can optimize lock contention.
2. The performance of btree degrades as # of elements increases. Search, for example, is lg(n). This is compared to a hash table, which doesn’t degrade the same way. So, reducing the # of items in each btree results in faster operations overall.
3. As you know, dashmap isn’t ordered, but yes, there are attributes of dashmap (sharding) and ordering (btree) in the solution I’m prototyping.
4. Ultimately, I want to get rid of the ordering requirement on the accounts index, freeing us to use other data structures, like (h/d)ashmap. I believe rent is the main holdover. Do we need to scan in order for rpc or something? Anatoly says rent can be rewritten.
5. The basic idea is to use existing implementations (btree) and requirements (ordered by pubkey) but increase performance because we understand the nature, distribution, and size of our keys and data.

jeffwashington:
An issue is whether we have a single lock for all accounts, as we do today, or whether we have 1 lock per pubkey bin. parallel inserts are important for startup. parallel lookups are always possible at runtime. Since Stephen got rid of remove_zero_lamport_key,  which used to have an implicit relationship to the single account index lock, this account index refactoring has gotten more clear.
One thing we can do, which I was thinking of prototyping was breaking account index structure into 2 phases: startup/population uses 1 lock per bin. Runtime uses 1 lock overall. After startup is complete (after generate_index), we extract all the btrees from their rwlocks and change from:
Vec<RwLock<btree>>
to
RwLock<Vec<btree>>

Today we have
RwLock<btree>

If we retain the single lock for the entire account index, then we have no unknown synchronization issues since we aren’t changing anything. But, we get vast startup performance and overall lookup performance (due to less items per btree).
Note that scan performance is worse in all cases. If we bin on byte 0 of the pubkey, then scan performance won’t be very bad. If we bin on a different byte, then an ordered scan across bins will require a merge sort of iterated values, which can be awful.

ryoqun:
>     4\. Do we need to scan in order for rpc or something?

@jeffwashington 

oh, this pr sounds nice stopgap until the accounts index spills over the ram with 1B entries? I had basically similar idea in mind for awhile.

As a context, _Ordered_ AccountsIndex is only needed for eager rent collection and its actual requirement isn't fully-ordered index. actually, entries just needs to be bucketed to one of 432000 bins (= # of slots in a epoch) using the first 8 bytes of pubkey.

There is no ordering requirement within each bucket as I commented here for possible parallel processing (i.e. unordered processing within each individual bucket): https://github.com/solana-labs/solana/blob/f2a25812596bdee6cfa170e7651f6b247c40383e/runtime/src/bank.rs#L3700-L3715

>  Anatoly says rent can be rewritten.

Moreover, the eager rent  collection itself could be reconsidered as well. ref: https://github.com/solana-labs/solana/issues/17178

jeffwashington:
```
this pr uses 16 bins:

64 core machine:
62M accounts:
master:  generate_index total_us=29750671i scan_stores_us=13178400i insertion_time_us=27230375i
16 bins: generate_index total_us=18576816i scan_stores_us=17977215i insertion_time_us=40063915i
32 bins: generate_index total_us=19276187i scan_stores_us=18152838i insertion_time_us=41399161i

360M accounts:
master:  generate_index total_us=110135794i scan_stores_us= 89552054i insertion_time_us=107437073i
16 bins: generate_index total_us= 57389495i scan_stores_us=148079315i insertion_time_us=172730514i
32 bins: generate_index total_us= 52960112i scan_stores_us=102114308i insertion_time_us=186820319i

16 core machine, 62M accounts:
master:  generate_index total_us=47246785i scan_stores_us=11533808i insertion_time_us=44789311i
16 bins: generate_index total_us=20266819i scan_stores_us=20278127i insertion_time_us=57765455i

```

jeffwashington:
> oh, this pr sounds nice stopgap until the accounts index spills over the ram with 1B entries?
I think the concept may still have value when we spill out of memory. It is reduce, based on values, the set of data that has to be searched.

runtime/src/accounts_db.rs
```diff
@@ -2097,7 +2097,7 @@ impl AccountsDb {
         );
     }
 
-    fn do_shrink_slot_stores<'a, I>(&'a self, slot: Slot, stores: I, is_startup: bool) -> usize
+    fn do_shrink_slot_stores<'a, I>(&'a self, slot: Slot, stores: I, _is_startup: bool) -> usize
```
sakridge, line 2100: remove `_is_startup` arg?

sakridge:
> 4. Ultimately, I want to get rid of the ordering requirement on the accounts index, freeing us to use other data structures, like (h/d)ashmap. I believe rent is the main holdover. Do we need to scan in order for rpc or something? Anatoly says rent can be rewritten.

Yes. RPC doesn't need any order at least today and we don't use the primary index for the RPC scans, we use the order of the secondary indexes which is a DashMap today. However, if we think about paging the data, then we may need to switch to an ordered data structure which can give a consistent set of accounts based on an index or prefix of a pubkey.

I think sharding the locks is a great idea, been meaning to experiment with that for a while. It should help reduce the lock contention with RPC scans preventing the validator from moving forward with transaction processing.

Should we do a bench-tps sanity run on this as well?

jeffwashington:
bench tps:
```
this pr:
    Average max TPS: 77647.80, 0 nodes had 0 TPS
    Highest TPS: 77647.80 sampling period 1s max transactions: 4145207 clients: 1 drop rate: 0.83
Average TPS: 45521.793

master:
    Average max TPS: 80821.48, 0 nodes had 0 TPS
    Highest TPS: 80821.48 sampling period 1s max transactions: 4160889 clients: 1 drop rate: 0.83
Average TPS: 45687.973
```

runtime/src/accounts_db.rs
jeffwashington: To avoid cluttering up this pr, I didn't add that change here. A by product of many locks is that it is harder to hold all of them open the same way we were before at startup. Meanwhile, shrink being called at startup is also tentatively on the chopping block separately, as you and I have been discussing. I was planning to avoid unnecessary code disturbance for the moment. And, maybe I'll add back in a lock held open version of shrink...

jeffwashington:
> > oh, this pr sounds nice stopgap until the accounts index spills over the ram with 1B entries?
> 
> I think the concept may still have value when we spill out of memory. It is useful to reduce, based on cheap to obtain values, the set of data that has to be searched or manipulated.

@aeyakovenko

runtime/src/accounts_index.rs
```diff
@@ -589,8 +595,14 @@ pub trait ZeroLamport {
     fn is_zero_lamport(&self) -> bool;
 }
 
+fn get_bin_pubkey(pubkey: &Pubkey) -> usize {
+    let byte_of_pubkey_to_bin = 0; // TODO: this should not be 0. For now it needs to be due to requests for in-order pubkeys
+    (pubkey.as_ref()[byte_of_pubkey_to_bin] as usize) * BINS / ((u8::MAX as usize) + 1)
```
sakridge, line 600: Why not just mask here?

sakridge (approved):
nice lgtm. Seems like a small regression on the peak perf but the average is quite close so seems similar.

runtime/src/accounts_index.rs
jeffwashington: Resolved in slack. We want high bits, so `byte >> (some function of BINS)`

PR #18118, opened 2021-06-21
capture sort time in hash calculation

jeffwashington:
#### Problem
Sort was reworked in latest hash calculation work.

#### Summary of Changes
plumb through metrics for sort of hash calculation

Fixes #


jbiseda (commented):
LGTM

PR #18141, opened 2021-06-22
Compute weighted_best() using WeightedIndex (#18143)

### LINKED ISSUE #18143: Compute weighted_best() using WeightedIndex
Update weighted_best() to use rand::distributions::WeightedIndex

jbiseda:
Update weighted_best() to use rand::distributions::WeightedIndex

Fixes #18143

### DIFF
--- a/gossip/src/weighted_shuffle.rs
+++ b/gossip/src/weighted_shuffle.rs
@@ -3,6 +3,7 @@
 use {
     itertools::Itertools,
     num_traits::{FromPrimitive, ToPrimitive},
+    rand::{distributions::WeightedIndex, prelude::Distribution},
     rand::{Rng, SeedableRng},
     rand_chacha::ChaChaRng,
     std::{iter, ops::Div},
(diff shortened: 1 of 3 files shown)


gossip/src/weighted_shuffle.rs
```diff
@@ -36,32 +37,17 @@ where
         .collect()
 }
 
-/// Returns the highest index after computing a weighted shuffle.
-/// Saves doing any sorting for O(n) max calculation.
-// TODO: Remove in favor of rand::distributions::WeightedIndex.
```
behzadnouri, line 41: So my comment here is actually misleading with respect to turbine usage of this function. The comment was added in https://github.com/solana-labs/solana/pull/13919 if you want to see the background context.

For all use cases which use `weighted_best` just to randomly select one index given some weights, we can replace `weighted_best` with `WeightedIndex` (and generally `WeightedIndex` has better performance unless the weights are extremely skewed).

But for turbine, we need both `broadcast` and `retransmit` shuffle nodes in the same way when they setup the fan-out tree. retransmit uses the `weighted_shuffle` function, which implements a heuristic and is not probabilistically equivalent with `WeightedIndex` (or how you would implement shuffle using `WeightedIndex`):
https://github.com/solana-labs/solana/blob/e808f34b0/gossip/src/weighted_shuffle.rs#L11-L37

Nevertheless `broadcast` has to use the same function (i.e `weighted_shuffle`), otherwise they will mismatch in how they arrange nodes on the fan-out tree which results in some nodes not getting packets.

Now, `broadcast` only cares about the root of the this tree (i.e. the very first node in `weighted_shuffle`), and `weighted_best` implements an optimization to obtain that very first node with less computation than `weighted_shuffle`. That optimization and the entire implementation of `weighted_best` is very specific and tied to `weighted_shuffle` heuristic algorithm, and so `weighted_best` cannot be replaced with `WeightedIndex` if we want to be consistent with `weighted_shuffle`.

PR #18144, opened 2021-06-22
correctly calculate hash_total for hash/lamport calculation

jeffwashington:
#### Problem
hash_total was double-counting some, resulting in incorrect stats.
#### Summary of Changes
don't double-count
Fixes #

### DIFF
--- a/runtime/src/accounts_hash.rs
+++ b/runtime/src/accounts_hash.rs
@@ -520,6 +520,7 @@ impl AccountsHash {
         let mut zeros = Measure::start("eliminate zeros");
         let overall_sum = Mutex::new(0u64);
         let unreduced_entries = AtomicUsize::new(0);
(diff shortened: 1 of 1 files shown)


runtime/src/accounts_hash.rs
```diff
@@ -688,7 +691,6 @@ impl AccountsHash {
         let mut next_pass = PreviousPass::default();
         let cumulative = CumulativeOffsets::from_raw(&hashes);
         let mut hash_total = cumulative.total_count;
-        stats.hash_total += hash_total;
```
jeffwashington, line 691: It isn't obvious, but a few lines above we re-add unprocessed hashes from the previous pass. This results in double-counts of those hashes.
```
hashes.insert(0...```

brooksprumo (approved):

PR #18160, opened 2021-06-23
Introduce slot dumping to ReplayStage

carllin:
#### Problem
Once the `cluster_slot_state_verifier()` determines 
1) a slot was DuplicateConfirmed 
2) We have the wrong version of that slot

We need to remove our version of the slot, which is not currently supported in ReplayStage

#### Summary of Changes
1) In replay, add `duplicate_slots_to_repair` which tracks slots for which we have the wrong version and need to dump and repair
2) Introduce function `dump_then_repair_correct_slots()` to dump all the state associated with a slot
3) Handle potential panics in switch threshold checks and other state checks in `consensus.rs` due to having potentially rolled back/dumped a slot we'd voted on.

Fixes #


core/src/replay_stage.rs
```diff
@@ -783,37 +799,93 @@ impl ReplayStage {
...
+                        descendants,
+                        progress,
+                        &root_bank,
+                        bank_forks,
+                        blockstore,
+                    );
+                    true
+                // TODO: Send signal to repair to repair the correct version of
+                // `duplicate_slot` with hash == `correct_hash`
+                } else {
```
carllin, line 854: We don't actually signal repair to go actively repair this duplicate slot, so right now repair will go through the passive path. This means we could potentially repair from the bad set of validators again and get the wrong version again.

core/src/replay_stage.rs
```diff
@@ -783,37 +799,93 @@ impl ReplayStage {
...
+            // should have been cleared as well. However, if there is some violation of this guarantee,
+            // then log here
+            let is_poh_building_on_duplicate_fork = poh_bank_slot
+                .map(|poh_bank_slot| {
+                    ancestors
+                        .get(&poh_bank_slot)
+                        .expect("Poh bank should exist in BankForks and thus in ancestors map")
+                        .contains(&duplicate_slot)
+                })
+                .unwrap_or(false);
```
jstarry, line 838: Is it possible that `poh_bank_slot == duplicate_slot` because we repaired the duplicate already? (not sure if this case is covered by your TODO above)

core/src/replay_stage.rs
```diff
@@ -827,18 +899,51 @@ impl ReplayStage {
...
 
-        for d in slot_descendants
-            .iter()
-            .chain(std::iter::once(&duplicate_slot))
-        {
-            // Clear the progress map of these forks
-            let _ = progress.remove(d);
+        // Grab the Slot and BankId's of the banks we need to purge, then clear the banks
+        // from BankForks
+        let (slots_to_purge, removed_banks): (Vec<(Slot, BankId)>, Vec<Arc<Bank>>) = {
```
jstarry, line 904: After purging from `BankForks` could we just get new `ancestors` and `descendants` rather than trying to keep them consistent above with `purge_ancestors_descendants`? I think that would be safer and could avoid future bugs

core/src/replay_stage.rs
```diff
@@ -783,37 +799,93 @@ impl ReplayStage {
...
 
-    #[allow(dead_code)]
     fn purge_unconfirmed_duplicate_slot(
         duplicate_slot: Slot,
         ancestors: &mut HashMap<Slot, HashSet<Slot>>,
         descendants: &mut HashMap<Slot, HashSet<Slot>>,
         progress: &mut ProgressMap,
+        root_bank: &Bank,
         bank_forks: &RwLock<BankForks>,
+        blockstore: &Blockstore,
```
jstarry, line 876: Should `duplicate_slots_tracker` be passed here too to have `duplicate_slot` removed from it?

core/src/cluster_slot_state_verifier.rs
```diff
@@ -3,9 +3,10 @@ use crate::{
     progress_map::ProgressMap,
 };
 use solana_sdk::{clock::Slot, hash::Hash};
-use std::collections::{BTreeMap, BTreeSet};
+use std::collections::{BTreeMap, BTreeSet, HashSet};
 
 pub(crate) type DuplicateSlotsTracker = BTreeSet<Slot>;
+pub(crate) type DuplicateSlotsToRepair = HashSet<(Slot, Hash)>;
```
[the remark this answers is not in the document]
carllin: I would if there weren't impending changes built on top of it that I want to check in after this, will cause more conflicts 🤕

core/src/consensus.rs
```diff
@@ -642,6 +642,22 @@ impl Tower {
...
+                // being a duplicate because `ancestors`/`descendants`/`progress` structurs may be missing this slot due
+                // to duplicate purging. This would cause many of the `unwrap()` checks below to fail.
+                if let Some(latest_duplicate_ancestor) = heaviest_subtree_fork_choice.latest_invalid_ancestor(&(last_voted_slot, last_voted_hash)) {
+                    // We're rolling back because one of the ancestors of the last vote was a duplicate. In this
+                    // case, it's acceptable if the switch candidate is one of ancestors of the previous vote,
+                    // just fail the switch check because there's no point in voting on an ancestor. ReplayStage
+                    // should then have a special case continue building an alternate fork from this ancestor, NOT
+                    // the `last_voted_slot`. This is in contrast to usual SwitchFailure where ReplayStage continues to build blocks
+                    // on latest vote. See `select_vote_and_reset_forks()` for more details.
+                    if heaviest_subtree_fork_choice.is_ancestor_slot(switch_slot, &(last_voted_slot, last_voted_hash)) {
```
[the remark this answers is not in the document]
carllin: Yeah this is a subtle point.

It is possible for ReplayStage to call `check_switch_threshold()` on a smaller slot than your last voted slot if it's trying to switch to a different fork. In these cases the switch threshold succeeds, but then the lockout check fails. The subtle difference here is that even though the lockout check fails in both cases, if the switch threshold succeeds we build our next block on the other fork, if it fails, we continue building blocks on our fork. This may not be a critical distinction now that we ingest votes for switching threshold from gossip (before you needed somebody to build blocks to see switching votes), but hesitant to make that change here.

So for this change:

What should never happen is you try to switch to smaller slot than your last voted slot AND that smaller slot is an ancestor of your last vote. This correctness check is enforced by this panic https://github.com/solana-labs/solana/blob/b2d071a9ea01c622b90cbece49f313b5bb886afb/core/src/consensus.rs#L690-L695. The only exception to this rule is the rollback based on dupe slots, which is why there's a special callout for that case and *only* this case here.

core/src/consensus.rs
```diff
@@ -642,6 +642,22 @@ impl Tower {
...
+                        return rollback_due_to_to_to_duplicate_ancestor(latest_duplicate_ancestor);
+                    }
```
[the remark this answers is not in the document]
carllin: Yeah you're talking about a case like this right?

```
                     0
            /            \
         1                3 (correct dup version that was duplicate confirmed)
      /                       \
   3 (our last vote)  6 (< Switch threshold have voted on this)
 /
5
```

We marked  our fork `1->3`duplicate, dumped it, and repaired `3->6`. We try to switch to slot `6` and this check `switch_slot_ancestors.contains(&last_voted_slot)` here https://github.com/solana-labs/solana/blob/b2d071a9ea01c622b90cbece49f313b5bb886afb/core/src/consensus.rs#L680  returns true, even though this fork did not reach switching threshold.

When thinking about this situation, there are a few cases to consider:

1) Our last vote was on our dumped version of`3`
2) Our last vote was on a descendant of the dumped version of `3`.

We know this new version of  slot `3` must be duplicate confirmed so it has at least 52% of the votes on it, which should be enough to generate a switching proof if your last vote was on your version of `3`. In case 2 however, it's possible your last vote was on `5` in the diagram above, in which case the lockouts on those observed votes for the duplicate threshold do not extend past `5`. In these cases I think we have either have to:

1) Distinguish between the old version of `3` and the new version of `3` using the fork choice (which is why we don't purge the fork choice, it keeps the entire history around).  However, then what happens if we restart and lose the fork choice tree? 
2) Wait for everybody to vote past slot `5` on the new fork at which time we know *for sure* the lockout is past 3. 

In both cases we won't halt waiting because at least 52% of the network has voted on the correct version of 3, so they should be on the right fork


Basically my approach right now is to ignore these cases and allow a "freebie" vote that may violate switching proofs in the case of duplicates. Later we'll need to patch all these edge cases up.

core/src/replay_stage.rs
carllin: I don't think so, once a slot is put into `duplicate_slots_tracker`, it's to stay there forever until the root moves past that slot. The reason is if the goal of the tracker is to inform the `cluster_slot_state_verifier` state machine that this slot is duplicate and needs special handling. Just because we dumped one version of the slot, doesn't make this slot no longer a duplicate in the cluster.

core/src/replay_stage.rs
carllin: `poh_bank_slot` is the slot of a leader block we are currently generating, so the only way `poh_bank_slot == duplicate_slot` is if we are the ones that generated the duplicate block.

core/src/replay_stage.rs
carllin: Yeah I thought about this as well, but was concerned that if we had a large tree with a bunch of slots on the same branch being deleted, it would cause a bunch of repetitive `n^2` operations to regenerate the ancestors and descendants.

For instance if you had a branch

```
1 -> 2 -> 3 -> 4 -> 5
   | -------BIG BRANCH
```

and 2, 3 ,4, 5 were all being dumped, then you would either:

1) Regenerate the ancestors/descendants after every single slot was dumped.
2) Only regenerate the ancestors/descendants at the end when all the dumping was done. But this means you have an inconsistency between the data structures in earlier vs later slots, i.e. `BankForks/AccountsDb/Blockstore` could already be missing some of the slots that had been cleared earlier, so you would have to handle those cases instead of hard asserting that they're consistent as we do right now.

core/src/replay_stage.rs
jstarry: Got it, thanks

core/src/replay_stage.rs
jstarry: Makes sense, thanks

core/src/replay_stage.rs
jstarry: Got it, yeah this is more efficient by not recreating ancestors, but at least descendants is updated with each removal from bank forks. Not sure if the efficiency savings are worth it if it means a bug is introduced down the road. Where is the consistency check you're referring to?

core/src/consensus.rs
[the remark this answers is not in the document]
jstarry: I don't think that situation 2) is a concern because it wouldn't pass this check `switch_slot_ancestors.contains(&last_voted_slot)` to return `SameFork`. I was concerned about situation 1) and bypassing lockout, wasn't thinking about switch threshold but that's a good point too. Curious what you think about the lockout, but for the switch threshold it would be good to document the freebie vote issue

core/src/consensus.rs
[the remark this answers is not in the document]
jstarry: Got it, thanks for pointing at that panic, sounds good

core/src/consensus.rs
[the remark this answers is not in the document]
carllin: Yup I made a special case for that freebie now, and added some tests for these edge cases: https://github.com/solana-labs/solana/pull/18160/files#diff-6d8458bb2e53158ac472a9ad4709e6a0a52b75d930c019013298e8acda133828R4844-R5184. 

I found an additional bug in the rollback + lockout scenario illustrated by this ignored test: https://github.com/solana-labs/solana/pull/18160/files#diff-6d8458bb2e53158ac472a9ad4709e6a0a52b75d930c019013298e8acda133828R5166-R5167

The issue stems from the fact that the locked out check can't distinguish between two ancestors with the same slot but different hashes. This `contains` check can pass even though we are on a different fork: https://github.com/solana-labs/solana/blob/master/core/src/consensus.rs#L538

core/src/consensus.rs
```diff
@@ -642,6 +642,34 @@ impl Tower {
                     SwitchForkDecision::FailedSwitchDuplicateRollback(latest_duplicate_ancestor)
                 };
 
+                // `heaviest_subtree_fork_choice` entries are not cleaned by duplicate block purging/rollback logic,
+                // so this is safe to check here. We return here if the last voted slot was rolled back/purged due to
+                // being a duplicate because `ancestors`/`descendants`/`progress` structurs may be missing this slot due
+                // to duplicate purging. This would cause many of the `unwrap()` checks below to fail.
+                //
+                // TODO: Handle if the last vote is on a dupe, and then we restart. The dupe won't be in
```
jstarry, line 650: This todo feels pretty important, do you plan to follow up in another PR soon to address this?

jstarry (commented):
Fixes look good!

core/src/consensus.rs
```diff
@@ -571,7 +571,7 @@ impl Tower {
     #[allow(clippy::too_many_arguments)]
     fn make_check_switch_threshold_decision(
         &self,
-        switch_slot: u64,
+        switch_slot: Slot,
```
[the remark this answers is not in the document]
carllin: ha, got lazy and didn't want to plumb it through all the tests. Reusing the progress map seemed convenient, but you're right, it should be passed in.

core/src/consensus.rs
carllin: Yeah I think this and the lockout issue illustrated by this problem here: https://github.com/solana-labs/solana/pull/18160/files#diff-6d8458bb2e53158ac472a9ad4709e6a0a52b75d930c019013298e8acda133828R5166-R5167 require that the tower keep track of all the hashes of all the votes in the tower, and not just the slot number

This means making an ABI change to tower and a migration, it's kind of a PITA but needs to be done 🤕

core/src/replay_stage.rs
carllin: > Only regenerate the ancestors/descendants at the end when all the dumping was done. But this means you have an inconsistency between the data structures in earlier vs later slots, i.e. `BankForks/AccountsDb/Blockstore` could already be missing some of the slots that had been cleared earlier, so you would have to handle those cases instead of hard asserting that they're consistent as we do right now.

I meant if you followed this then in places like here: https://github.com/solana-labs/solana/blob/b69afd9d88a1a13de6437d6e54bd974f3728e106/core/src/replay_stage.rs#L946-L954.  If you 

1) Cleared a later descendant and didn't update the  the `ancestors/descendants`  
2) Then cleared an ancestor, the `slot_descendants` of that ancestor will not be updated and the assertion panics

carllin: > Got it, yeah this is more efficient by not recreating ancestors, but at least descendants is updated with each removal from bank forks. Not sure if the efficiency savings are worth it if it means a bug is introduced down the road. Where is the consistency check you're referring to?

It's an n^2 operation to recreate all the ancestors and descendants. If there's a large number of banks, let's say 500 from some duplicate forking, that could potentially be a lot of scans

carllin: https://github.com/solana-labs/solana/pull/18160/commits/1dac1a8933029b0bb628a5c4e3de4333a5c4ca83, seems to be taking multiple seconds

core/src/replay_stage.rs
```diff
@@ -807,37 +823,120 @@ impl ReplayStage {
...
-                descendants,
-                progress,
-                bank_forks,
-            );
+        if duplicate_slots_to_repair.is_empty() {
+            return;
         }
+
+        let root_bank = bank_forks.read().unwrap().root_bank();
+        // TODO: handle if alternate version of descendant also got confirmed after ancestor was
```
AshwinSekar, line 840: Does this mean if we have a fork of multiple duplicate blocks that we've confirmed we will try to repair slots that were already dumped?

core/src/replay_stage.rs
AshwinSekar: So we retain the duplicate slots for which the PoH bank is building on, and then later if we reset PoH to a different slot we will go ahead and repair. Does this order of events cause issues?

core/src/replay_stage.rs
```diff
@@ -851,18 +950,51 @@ impl ReplayStage {
...
+        let (slots_to_purge, removed_banks): (Vec<(Slot, BankId)>, Vec<Arc<Bank>>) = {
+            let mut w_bank_forks = bank_forks.write().unwrap();
+            slot_descendants
+                .iter()
+                .chain(std::iter::once(&duplicate_slot))
+                .map(|slot| {
+                    // Clear the duplicate banks from BankForks
+                    let bank = w_bank_forks
+                        .remove(*slot)
+                        .expect("BankForks should not have been purged yet");
```
AshwinSekar, line 964: Could we have a situation where the descendant was also a duplicate and purged first due to the ordering of `duplicate_slots_to_repair`?

core/src/replay_stage.rs
carllin: I think what I was imagining was something like:

```
1 -> 2 -> 3  -> 4
         | ---3' -> 4' -> 5
```

I was imagining we got duplicate confirmed for `3`, then `4`. So `3` and `4` would both get added to the `duplicate_slots_to_repair` list.

Then here: https://github.com/solana-labs/solana/blob/b74ff2362aaf4fe4a14a0ce2cbc30f5c05dce33e/core/src/replay_stage.rs#L843, either of `3` or `4` could be purged first. This function should not panic under both orderings

Test added `test_dump_then_repair_correct_slots()` here: https://github.com/solana-labs/solana/pull/18160/commits/33fc32f09eb517d4028f75593d1b829c0cc064bc

core/src/replay_stage.rs
carllin: ha! refer to comment: https://github.com/solana-labs/solana/pull/18160#discussion_r665851025. 

I think this check https://github.com/solana-labs/solana/blob/b74ff2362aaf4fe4a14a0ce2cbc30f5c05dce33e/core/src/replay_stage.rs#L886-L890 should cover that case. Maybe it shouldn't be a warn?

carllin:
>So we retain the duplicate slots for which the PoH bank is building on, and then later if we reset PoH to a different slot we will go ahead and repair. Does this order of events cause issues?

The PoH bank is being modified by a different thread running `BankingStage`. So as long as the ancestors of the PoH bank are not being concurrently modified by ReplayStage, this should be fine. 

Thus if PoH is reset to a different slot on a different fork/same fork, as long as we uphold. that we are not modifying/purging its ancestors, we should be ok.

Does that make sense?

AshwinSekar:
> > So we retain the duplicate slots for which the PoH bank is building on, and then later if we reset PoH to a different slot we will go ahead and repair. Does this order of events cause issues?
> 
> The PoH bank is being modified by a different thread running `BankingStage`. So as long as the ancestors of the PoH bank are not being concurrently modified by ReplayStage, this should be fine.
> 
> Thus if PoH is reset to a different slot on a different fork/same fork, as long as we uphold. that we are not modifying/purging its ancestors, we should be ok.
> 
> Does that make sense?

Makes sense!

AshwinSekar (approved):

PR #18178, opened 2021-06-23
explorer: retain inspector signatures and show simulation error

jstarry:
#### Problem
- If a full transaction is pasted into the inspector, the signatures are not encoded in the url.
- When simulation fails, no error message is displayed


jstarry:
Screen Shot 2021-06-23 at 10 39 49 AM

oJshua (approved):
Very cool. Works great!

Issue #18091, opened 2021-06-21
Explore sled as a replacement to both rocksdb and accountsdb


jon-chuang:
#### Problem
Many separate problems:
1. rocksdb has been reported to have problems. Maybe sled is an answer.
2. AccountsDB: One shouldn't rely on mmap/kernel as a page cache. Sled offers a [page cache](https://github.com/spacejam/sled/tree/main/src/pagecache). 
3. AccountsDB: BTreeMap is slow. Sled uses a lock-free (via optimistic concurrency control), optimised bw-tree.

#### Proposed Solution
[Sled](https://github.com/spacejam/sled) offers 1B ops in < 60s for 95% read 5% write workload in in-memory case (17.5+MOps). 
Sled has support for io_uring (writes). 

#### Open Questions/Issues
1. Is sled index-larger-than-memory friendly? A: yes.
2. LLAMA-based materialized page fragments, while decreasing write amp, increases read amp... Can this be tuned/is this significant in the first place?
3. `AccountsDb` as a whole is very tighly integrated with `AppendVecId`s. But since index needs to be reworked for larger than memory case anyway...
4. Does sled offer io_uring/batched support for reads? A: not yet. But possibly soon.

jeffwashington:
I hope to build this into the current accounts index prototype within the next day or so. I almost have rocksdb 'working'. I have anatoly's buckets working already.

PR #18179, opened 2021-06-23
WIP: MMAP backed bucket map that is suitable as a KV store for the accounts index

aeyakovenko:
#### Problem

Accounts index takes up way to much RAM

##### Design
1. create a large set of buckets, 32k or more up front
2. pubkeys based on their first 16 bits are stored in each bucket
3. each bucket has a mmap for the index entry and N mmaps for variable sized [SlotInfo] data
4. each data mmap stores a 2^n SlotInfo slices

Each bucket  index or bucket data can double independently.  

##### Locking

Each bucket has its own RWLock.  We should test performance with spin::RwLock vs sync::RwLock.  The DataBucket allocator can function with atomics, so it's possible to redesign this such that the write lock is only necessary when reallocating.

##### In Order Scan

The buckets are in order, while keys in each bucket are going to be out of order.  To scan the entire index in order, read each bucket in order and use a rayon parallel sort.

Fixes #

tag @carllin @jeffwashington @sakridge @ryoqun


jeffwashington:
I have done several passes of test code that could be helpful here to gather metrics across # accounts, insert, lookup, memory usage for hashmap, dashmap, btree, binned btree, skip-list and binned skip-list. So far, I've been focusing on the various in-memory representations.

jon-chuang:
I want to advise against further use of MMAP, and, instead of tuning a new index datastructure, look into the possibility of an existing implementation (a optimistic concurrency control based, lock-free bw-tree, that also reduces write amp via LLAMA) which already has good performance claims:
https://github.com/solana-labs/solana/issues/18091

jon-chuang (commented):
Just doing some basic sanity checking

aeyakovenko:
Some benchmarks.  So single insert isn't super optimized yet, but its still pretty good.

```
test bucket_map_bench_hashmap_baseline       ... bench:         269 ns/iter (+/- 134)
test bucket_map_bench_insert_1               ... bench:         770 ns/iter (+/- 116)
test bucket_map_bench_insert_16x512          ... bench:  15,533,122 ns/iter (+/- 44,959,341)
test bucket_map_bench_insert_16x512_baseline ... bench:  23,809,525 ns/iter (+/- 4,480,271)
```

it seems to start outperform the `RwLock<HashMap>` with 16 threads and 256+ items per thread

aeyakovenko:
4096 buckets on Lemond with each thread adding 4096 items:

```
running 1 test
time: 471ns per item
dir_size: "/tmp/bucket_map_test_mt" 2712141824
dir_size: "/mnt/data/aeyakovenko/bucket_map_test_mt" 2774007808
```

Overhead per account: (2774007808 + 2712141824)/(4096 * 4096) = 327 bytes 
(that's pretty decent I think)

runtime/src/data_bucket.rs
```diff
@@ -0,0 +1,232 @@
...
+        self.lock.load(Ordering::Relaxed)
+    }
+}
+
+pub struct DataBucket {
+    drives: Arc<Vec<PathBuf>>,
+    path: PathBuf,
+    mmap: MmapMut,
+    pub cell_size: u64,
+    pub capacity: u64,
```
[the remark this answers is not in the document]
aeyakovenko: Yea. num_cells would be better :)

runtime/src/data_bucket.rs
```diff
@@ -0,0 +1,232 @@
...
+            std::slice::from_raw_parts_mut(item, len as usize)
+        }
+    }
+
+    fn new_map(drives: &[PathBuf], cell_size: usize, capacity: u64) -> (MmapMut, PathBuf) {
+        let r = thread_rng().gen_range(0, drives.len());
+        let drive = &drives[r];
+        let pos = format!("{}", thread_rng().gen_range(0, u128::MAX),);
+        let file = drive.join(pos);
+        let mut data = OpenOptions::new()
```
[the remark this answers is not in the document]
aeyakovenko: Yea. V2

runtime/src/accounts_db.rs
```diff
@@ -4231,7 +4231,15 @@ impl AccountsDb {
         let chunks = crate::accounts_hash::MERKLE_FANOUT.pow(4);
         let total_lamports = Mutex::<u64>::new(0);
         let get_hashes = || {
-            keys.par_chunks(chunks)
+            let mut hashes = vec![];
+            for ix in 0..self.accounts_index.account_maps.num_buckets() {
+                let keys = self.accounts_index.account_maps.keys(ix);
```
aeyakovenko, line 4236: @ryoqun @sakridge @jeffwashington 

wdyt, would this work here?  the initial design had a snapshot slot that would keep the data behind that slot form being cleaned up as the index is updated.  So we shouldn't need a global lock on all the keys to do the snapshot.

carllin: @brooksprumo is good to reference here, as he's been working on snapshotting related logic

jeffwashington:
I can check more at my computer but I think this code path is already dead except for debug validation. We don’t want to use account index for hash calculation.

runtime/src/accounts_db.rs
jeffwashington: I think this code path is already dead except for debug validation. We don’t want to use account index for hash calculation. Even so, does bucket[x] have keys that are all < the keys in bucket[x+1]? This algorithm is written to assume the keys vector is in sorted order. If not, we have to save key and hash from this scan, sort by key, then hash.

brooksprumo: I can confirm what @jeffwashington said that on the snapshot side this is never called/used. The snapshot code that calls `AccountsDb::calculate_accounts_hash_without_index()` is inside an `if` that is always false.

There are hooks for testing, but none of the tests actually exercise this code path. I assume in the past this was different, but as of now `AccountsDb::calculate_accounts_hash_without_index()` is never called in the snapshot code (nor anything that calls these snapshot functions).

Relevant snapshot functions:
`snapshot_utils::snapshot_bank()`
`snapshot_utils::package_snapshot()`

brooksprumo: I can confirm what @jeffwashington said that on the snapshot side this is never called/used. The snapshot code that calls `AccountsDb::calculate_accounts_hash_without_index()` is inside an `if` that is always false.

There are hooks for testing, but none of the tests actually exercise this code path. I assume in the past this was different, but as of now `AccountsDb::calculate_accounts_hash_without_index()` is never called in the snapshot code (nor anything that calls these snapshot functions).

Relevant snapshot functions:
`snapshot_utils::snapshot_bank()`
`snapshot_utils::package_snapshot()`

aeyakovenko: @jeffwashington 
> Even so, does bucket[x] have keys that are all < the keys in bucket[x+1]? 

Yep.  that is guaranteed. https://github.com/solana-labs/solana/pull/18179/files#diff-d90ce8cbcea9459dbca60678b48f6d6d2816b8d4de8b8ebfd5c434b819f153b3R97

jeffwashington:
fyi, I am working on this

jeffwashington:
@aeyakovenko , @sakridge 
fyi that pubkeys are almost perfectly uniform [currently](https://github.com/solana-labs/solana/pull/18338#issuecomment-871717801) on mnb:

jeffwashington:
```
ledger-tool verify test

wiggins, mnb, 61M accounts, 71M accounts with duplicates, single normal folder for files
master:
generate_index total_us= 16595216i scan_stores_us= 16823797i
generate_index total_us=825187084i scan_stores_us=276917451i
new accounts_index^


    327808 Jun 30 13:41 107904300402519040087491652047439004753
   2637824 Jun 30 13:36 109345486004409218025167728360945078848
   2637824 Jun 30 13:36 111703861789759256386814238270430935733
   1311744 Jun 30 13:41 116046545619161449364167284873912648143
  92274688 Jun 30 13:41 117683797302450908233447399583968109837
    327808 Jun 30 13:36 122057823506595257348099622663023862920
   5308416 Jun 30 13:41 13285813583199267676624996780009823364
3221225472 Jun 30 13:41 136992253147470056637757453963745784918
  92274688 Jun 30 13:36 144942986957396586694162914976464644780
1474836480 Jun 30 13:41 151948783710700931791705349917011340835
  10747904 Jun 30 13:36 165875409675962637757766775146792395370
   2625536 Jun 30 13:36 170931019416804702255098208828420343482
   2637824 Jun 30 13:41 173609210299367268293759929740876385704
1474836480 Jun 30 13:36 177774730398300555057755705150624094737
  92274688 Jun 30 13:36 178235986421029890768303950224301994789
   1311744 Jun 30 13:36 178708986363673407147625341314098709410
   1311744 Jun 30 13:36 184250407062993388498508125326056013397
3221225472 Jun 30 13:36 185491846533121536344599546326624101735
5368709120 Jun 30 13:41 191445498694755269669143011082615013672
   2637824 Jun 30 13:41 2042179688534498252169527402618460346
    327808 Jun 30 13:41 210767942703357350555851792023338475983
  22020096 Jun 30 13:41 216348693308995961494877040157519890257
   5308416 Jun 30 13:36 221044763825635906762449545102898613129
5368709120 Jun 30 13:36 245095890302948737710208538026124458693
   2625536 Jun 30 13:41 249525702686317488026247834407761494879
   1314816 Jun 30 13:36 253590354277014389242162150298080574666
  22020096 Jun 30 13:36 259921755917702521907108000656051434326
   2625536 Jun 30 13:36 266808801864297656587189452489456853566
    327808 Jun 30 13:36 267739506204975156762614520273657710490
6442450944 Jun 30 13:41 268772086480765926121775674360524325760
5368709120 Jun 30 13:41 270393137229078290044361969634464408776
  10747904 Jun 30 13:41 275115113425023436410260381737802059819
   5308416 Jun 30 13:36 29220792411945194733181704118491382268
    393216 Jun 30 13:38 292361426992516571813833999311023787886
   1310720 Jun 30 13:38 293035005090054264009782127557246873956
  22020096 Jun 30 13:36 29864249271048726823797681567135834882
   2625536 Jun 30 13:41 302545613006025209996030690761277733597
  10747904 Jun 30 13:41 318789413178262821609329697529453301106
   2629632 Jun 30 13:41 323904516372193207169021063878690731045
   2629632 Jun 30 13:36 325762219668990833818585440670038542421
   5308416 Jun 30 13:41 331541511519991026047134937733136964747
   2629632 Jun 30 13:41 46688159280637054803068807706077037190
5368709120 Jun 30 13:36 51191087325416087376239576951208600927
   1310720 Jun 30 13:38 57859127627816537105434335467753332727
  92274688 Jun 30 13:41 67651739661754000165984139124278391585
  22020096 Jun 30 13:41 694773002099287328588246018998073976
   1311744 Jun 30 13:41 71725465245843727890467210651900348351
    786432 Jun 30 13:38 78886221147038265722916895303761205940
  10747904 Jun 30 13:36 98859644244956868752802778284410325312
3221225472 Jun 30 13:36 99917048349695894341172296625530923069

du accounts_index_buckets/
41366724	accounts_index_buckets/
```

jeffwashington:
just a comment here since its stale. This is still actively in prototyping.

PR #18182, opened 2021-06-23
Add ability to change the validator identity at runtime

mvines:
```
$ solana-validator set-identity --help
Set the validator identity

USAGE:
    solana-validator --ledger <DIR> set-identity [KEYPAIR]

FLAGS:
    -h, --help    Prints help information

ARGS:
    <KEYPAIR>    Validator identity keypair

Note: the new identity only applies to the currently running validator instance
```

### DIFF
--- a/core/src/banking_stage.rs
+++ b/core/src/banking_stage.rs
@@ -331,7 +331,6 @@ impl BankingStage {
         // Single thread to generate entries from many banks.
         // This thread talks to poh_service and broadcasts the entries once they have been recorded.
         // Once an entry has been recorded, its blockhash is registered with the bank.
-        let my_pubkey = cluster_info.id();
         let duplicates = Arc::new(Mutex::new((
             LruCache::new(DEFAULT_LRU_SIZE),
             PacketHasher::default(),
@@ -358,7 +357,6 @@ impl BankingStage {
                     .name("solana-banking-stage-tx".to_string())
                     .spawn(move || {
                         Self::process_loop(
-                            my_pubkey,
                             &verified_receiver,
                             &poh_recorder,
                             &cluster_info,
@@ -699,7 +697,6 @@ impl BankingStage {
 
     #[allow(clippy::too_many_arguments)]
     fn process_loop(
-        my_pubkey: Pubkey,
         verified_receiver: &CrossbeamReceiver<Vec<Packets>>,
         poh_recorder: &Arc<Mutex<PohRecorder>>,
         cluster_info: &ClusterInfo,
@@ -718,6 +715,7 @@ impl BankingStage {
         let mut buffered_packets = VecDeque::with_capacity(batch_limit);
         let banking_stage_stats = BankingStageStats::new(id);
         loop {
+            let my_pubkey = cluster_info.id();
             while !buffered_packets.is_empty() {
                 let decision = Self::process_buffered_packets(
                     &my_pubkey,
--- a/core/src/broadcast_stage.rs
+++ b/core/src/broadcast_stage.rs
@@ -173,7 +173,7 @@ pub struct BroadcastStage {
 impl BroadcastStage {
     #[allow(clippy::too_many_arguments)]
     fn run(
-        keypair: &Keypair,
+        cluster_info: Arc<ClusterInfo>,
         blockstore: &Arc<Blockstore>,
         receiver: &Receiver<WorkingBankEntry>,
         socket_sender: &Sender<(TransmitShreds, Option<BroadcastShredBatchInfo>)>,
@@ -182,7 +182,7 @@ impl BroadcastStage {
     ) -> BroadcastStageReturnType {
         loop {
             let res = broadcast_stage_run.run(
-                keypair,
+                &cluster_info.keypair(),
                 blockstore,
                 receiver,
                 socket_sender,
@@ -248,21 +248,23 @@ impl BroadcastStage {
         let bs_run = broadcast_stage_run.clone();
 
         let socket_sender_ = socket_sender.clone();
-        let keypair = cluster_info.keypair.clone();
-        let thread_hdl = Builder::new()
-            .name("solana-broadcaster".to_string())
-            .spawn(move || {
-                let _finalizer = Finalizer::new(exit);
-                Self::run(
-                    &keypair,
-                    &btree,
-                    &receiver,
-                    &socket_sender_,
-                    &blockstore_sender,
-                    bs_run,
-                )
-            })
-            .unwrap();
+        let thread_hdl = {
+            let cluster_info = cluster_info.clone();
+            Builder::new()
+                .name("solana-broadcaster".to_string())
+                .spawn(move || {
+                    let _finalizer = Finalizer::new(exit);
+                    Self::run(
+                        cluster_info,
+                        &btree,
+                        &receiver,
+                        &socket_sender_,
+                        &blockstore_sender,
+                        bs_run,
+                    )
+                })
+                .unwrap()
+        };
         let mut thread_hdls = vec![thread_hdl];
         let socket_receiver = Arc::new(Mutex::new(socket_receiver));
         for sock in socks.into_iter() {
--- a/core/src/replay_stage.rs
+++ b/core/src/replay_stage.rs
@@ -341,7 +341,8 @@ impl ReplayStage {
             .spawn(move || {
                 let verify_recyclers = VerifyRecyclers::default();
                 let _exit = Finalizer::new(exit.clone());
-                let my_pubkey = cluster_info.id();
+                let mut identity_keypair = cluster_info.keypair().clone();
+                let mut my_pubkey = identity_keypair.pubkey();
                 let (
                     mut progress,
                     mut heaviest_subtree_fork_choice,
@@ -517,6 +518,7 @@ impl ReplayStage {
                                                     heaviest_bank_on_same_voted_fork,
                                                     &poh_recorder, my_latest_landed_vote,
                                                     &vote_account,
+                                                    &identity_keypair,
                                                     &authorized_voter_keypairs.read().unwrap(),
                                                     &mut voted_signatures,
                                                     has_new_vote_been_rooted, &mut
@@ -584,6 +586,7 @@ impl ReplayStage {
                             &mut tower,
                             &mut progress,
                             &vote_account,
+                            &identity_keypair,
                             &authorized_voter_keypairs.read().unwrap(),
                             &cluster_info,
                             &blockstore,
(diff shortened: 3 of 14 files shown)


jon-chuang:
This will help immensely with safely transitioning to fallback machines.

mvines:
It's a start for sure.  The harder problem is next up: dynamically switching the vote account

jon-chuang:
Why is that desirable? 

Not sure if feasible - with associated assortment of stake accounts...

mvines:
> Why is that desirable?

For failover

mvines:
> It's a start for sure. The harder problem is next up: dynamically switching the vote account

Actually upon further reflection, we might not need to support dynamically switching the vote account at all.  Just a safe way to transition a non-voting node to a voting node while respecting the latest vote lockups

core/src/banking_stage.rs
```diff
@@ -648,6 +645,7 @@ impl BankingStage {
         let mut buffered_packets = VecDeque::with_capacity(batch_limit);
         let banking_stage_stats = BankingStageStats::new(id);
         loop {
+            let my_pubkey = cluster_info.id();
```
mvines, line 648: Allows BankingStage to pick up an id change at the top of its main loop

core/src/broadcast_stage.rs
```diff
@@ -182,7 +182,7 @@ impl BroadcastStage {
     ) -> BroadcastStageReturnType {
         loop {
             let res = broadcast_stage_run.run(
-                keypair,
+                &cluster_info.keypair(),
```
mvines, line 185: Allows BroadcastStage to pick up an id change at the top of its main loop

core/src/replay_stage.rs
```diff
@@ -613,6 +616,14 @@ impl ReplayStage {
                                     i64
                                 ),
                             );
+
+                            if my_pubkey != cluster_info.id() {
```
mvines, line 620: ReplayStage id change needs to be timed with a fork reset, to ensure PoH recorder will be reset the include the next leader slot for the new id.  This effectively means that an id reset while the validator is leader will be postponed until their time as leader is up

gossip/src/cluster_info.rs
```diff
@@ -213,7 +213,7 @@ pub struct ClusterInfo {
     /// The network
     pub gossip: RwLock<CrdsGossip>,
     /// set the keypair that will be used to sign crds values generated. It is unset only in tests.
-    pub keypair: Arc<Keypair>,
+    keypair: RwLock<Arc<Keypair>>,
```
mvines, line 216: A new RwLock here so keypair can be updated. cc: @behzadnouri

gossip/src/cluster_info.rs
```diff
@@ -224,7 +224,7 @@ pub struct ClusterInfo {
     local_message_pending_push_queue: Mutex<Vec<CrdsValue>>,
     contact_debug_interval: u64, // milliseconds, 0 = disabled
     contact_save_interval: u64,  // milliseconds, 0 = disabled
-    instance: NodeInstance,
+    instance: RwLock<NodeInstance>,
     contact_info_path: PathBuf,
 }
```
mvines, line 230: A new RwLock here so instance can be updated. cc: @behzadnouri

gossip/src/cluster_info.rs
```diff
@@ -680,6 +680,25 @@ impl ClusterInfo {
         self.my_contact_info.read().unwrap().id
     }
 
+    pub fn keypair(&self) -> RwLockReadGuard<Arc<Keypair>> {
+        self.keypair.read().unwrap()
+    }
+
+    pub fn set_keypair(&self, new_keypair: Arc<Keypair>) {
```
mvines, line 687: @behzadnouri, does `set_keypair()` look ok?  Anything else I might be missing to encourage gossip to jump over to the new keypair?

gossip/src/cluster_info.rs
```diff
@@ -2263,7 +2286,7 @@ impl ClusterInfo {
         let packets: Vec<_> = pings
             .into_iter()
             .filter_map(|(addr, ping)| {
-                let pong = Pong::new(&ping, &self.keypair).ok()?;
+                let pong = Pong::new(&ping, &self.keypair()).ok()?;
```
mvines, line 2289: Think it's worth pulling this lock out of the iterator?

gossip/src/gossip_service.rs
```diff
@@ -133,8 +133,10 @@ pub fn discover(
     Vec<ContactInfo>, // all gossip peers
     Vec<ContactInfo>, // tvu peers (validators)
 )> {
-    let keypair = keypair.unwrap_or_else(|| Arc::new(Keypair::new()));
-
+    let keypair = {
+        #[allow(clippy::redundant_closure)]
```
mvines, line 137: I was defeated by clippy here...

gossip/src/cluster_info.rs
behzadnouri: It looks ok to me but I am bit worried to claim that nothing is left out because, well, you never know.

This seems a very nuanced change, and feels to me a bit risky. Is there any way to avoid adding this capability?

mvines: > This seems a very nuanced change, and feels to me a bit risky. Is there any way to avoid adding this capability?

I don't think so. Basically the bigger picture here is to enable online rollover from one validator to another, for upgrades or failover purposes. One has a backup unstaked validator running with another identity, and then upon signal it needs to flip over to using the staked validator identity and start voting. The "start voting" part is not in this PR, that's next and will be a much larger change/complex change.

gossip/src/cluster_info.rs
behzadnouri: Why are these `RwLock<Arc<...>>` as opposed to `Arc<RwLock<...>>`?

mvines: Heh, good question:
1. One can't clone/copy a `Keypair`, so we need an `Arc<Keypair>` to be able to generally pass the keypair around.  Shoving a `Rwlock` between the `Arc` and `Keypair` would cause a ton of code churn and `read().unlock()` overhead on each access
2. There's no need for an `Arc` around the `RwLock` since everybody holds an `Arc` to `ClusterInfo` already

gossip/src/cluster_info.rs
behzadnouri: Would it make sense to wrap the key-pair in a struct, such as
```
struct ValidatorIdentity(Arc<RwLock<Keypair>>)
```
and only allow keypair to be borrowed, as in:
```
impl ValidatorIdentity {
   fn keyapir(&self) -> RwLockReadGuard<Keypair> {
     self.0.read().unwrap()
   }
}
```
I think this would prevent the `Keypair` or `Arc<Keypair>` leaking elsewhere and would enforce some type-safety. e.g. `ClusterInfo` would then have a `ValidatorIdentity` field.

core/src/validator.rs
```diff
@@ -827,6 +829,8 @@ impl Validator {
     }
 
     pub fn join(self) {
+        drop(self.cluster_info);
```
behzadnouri, line 832: Why do we need `drop` here?

gossip/src/cluster_info.rs
```diff
@@ -680,6 +680,25 @@ impl ClusterInfo {
...
+        self.keypair.read().unwrap()
+    }
+
+    pub fn set_keypair(&self, new_keypair: Arc<Keypair>) {
+        let id = new_keypair.pubkey();
+
+        self.gossip.write().unwrap().set_self(&id);
+        {
+            let mut instance = self.instance.write().unwrap();
+            *instance = instance.with_id(id);
```
behzadnouri, line 693: maybe this should update the `timestamp` as well:
https://github.com/solana-labs/solana/blob/7782a0855/gossip/src/crds_value.rs#L393

effective this instance with this keypair is starting now, and if the timestamp is not updated and the old machine had more recent starting timestamp, then this machine will abort.

mvines: Thanks, yep quite right.  Fixed

core/src/validator.rs
mvines: I'm just being defensive.  Some of the validator stages are sensitive to when drops occur.  For example of you hold a blockstore in `Validator` and try to `Validator::join()`, retransmit stage will never terminate.  So by dropping the new cluster_info reference here first, I'm 100% sure the stage shutdown path will remain unchanged.  Not worth it?

gossip/src/cluster_info.rs
mvines: 🤔 I'll give that a shot and see what the code looks like. I do like the idea as a way to prevent the "swappable" keypair from escaping.

mvines: This almost worked out.  The problem is Replay stage.  On an identity change, Replay needs to hold onto the previous identity until the next fork reset and not doing so would cause significant code churn in Replay.  

So for `ValidatorIdentity` to work out, I think it needs to look more like
```rust
struct GossipIdentity {
  keypair: Keypair,
  my_contact_info: ContactInfo,
  instance: NodeInstance,
}
```
and then `ClusterInfo` holds a `gossip_id: RwLock<Arc<GossipIdentity>>>`

This doesn't seem half bad as then:
1. `set_keypair()` becomes atomic, one just passes in a new `GossipIdentity`
2. Replay can hold a clone of `Arc<GossipIdentity>` and swap it out on a fork reset.
3. `GossipIdentity` internally holds the ugly logic of ensuring that its three fields hold the same pubkey (duplicated 3 times...)

It'll be a bunch of code churn to accomplish though though, wdyt?

core/src/validator.rs
behzadnouri: I see, thanks for the explanation.

mvines: > maybe it implies that there are some circular references

Yeah for sure.  In the blockstore case, it holds a channel that stages assume get dropped.  Without that drop their thread loops never terminate.

behzadnouri (approved):
`NodeInstance::with_wallclock_and_id` needs to be replaced with `NodeInstance::new`, so that `timestamp` is updated. Otherwise, LGTM.

Though again this seems a very nuanced change, so maybe have some more eyes take a look just in case.

gossip/src/cluster_info.rs
mvines: done

gossip/src/cluster_info.rs
```diff
@@ -680,6 +680,25 @@ impl ClusterInfo {
...
+        self.keypair.read().unwrap()
+    }
+
+    pub fn set_keypair(&self, new_keypair: Arc<Keypair>) {
+        let id = new_keypair.pubkey();
+
+        self.gossip.write().unwrap().set_self(&id);
+        {
+            let mut instance = self.instance.write().unwrap();
+            *instance = instance.with_wallclock_and_id(timestamp(), id);
```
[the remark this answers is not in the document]
mvines: Ah, yes.  Duh that's way better

gossip/src/cluster_info.rs
```diff
@@ -2466,8 +2489,9 @@ impl ClusterInfo {
         // this node with more recent timestamp.
```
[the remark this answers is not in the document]
mvines: Good point, done

gossip/src/cluster_info.rs
mvines: Yeah at this point I feel this might generate even more complexity.  How the stages that deal with the identity keypair are pretty simple, they can just refresh it at a time that's convenient for them.  There's no real need to make the swap fully atomic across the entire validator process

PR #18192, opened 2021-06-24
DOCS: Update leader-rotation.md

marceljay:
#### Problem
Confusing wording regarding epoch length, which is stated as 100.  ("For example" does not really mean "100 is fictional")

grafik

#### Summary of Changes
Made clear it's a fictive number


mvines:
Lol Google.  Thanks for the PR!

PR #18196, opened 2021-06-24
remove unnecessary copies from accounts index code paths

jeffwashington:
#### Problem
We copy account info unnecessarily through a few code paths of account info insert.
#### Summary of Changes
Rework code to preserve the already allocated data for re-use. This also allows us to switch to an iterator vs a vec and have less data that we have to collect into vectors during processing of many slots.
Fixes #


sakridge (approved):
lgtm

jeffwashington:
ignoring flaky tests

PR #18198, opened 2021-06-24
replace account index bulk insert with iterator

jeffwashington:
#### Problem
Account index creation is slow. Account index takes a lot of memory. During startup, there is contention for memory. Soon, the bulk accounts to be inserted will be sorted and binned inside accounts index. This work is preparation for binning account data.
#### Summary of Changes
During bulk creation of account index, accept an iterator instead of requiring a vector. This allows lazy evaluation and prevents having to create vectors of data that will be resorted anyway.
Fixes #


sakridge (approved):
lgtm

PR #18199, opened 2021-06-24
Add snapshot_runtime_info module

brooksprumo:
#### Problem

Cannot share runtime snapshot information between AccountsDb/Bank, snapshot_utils, and SnapshotPackagerService.

#### Summary of Changes

The `snapshot_runtime_info` module adds the `SnapshotRuntimeInfo` and `SyncSnapshotRuntimeInfo`
types to handle sharing runtime snapshot information between threads; in
particular: AccountsDb, Bank, SnapshotPackagerService, and
snapshot_utils.

This will be used in a follow-up PR (#18207) to plumb `SyncSnapshotRuntimeInfo` through all of the above modules/types.

Related to #17088 Incremental Snapshots, and broken out from PR #17875.


runtime/src/snapshot_info.rs
```diff
@@ -0,0 +1,173 @@
...
+///
+/// Information about the snapshots is needed in multiple threads at different times; for example,
+/// `AccountsDb` (and `Bank` via `AccountsDb`) need to ensure cleaning does not happen past the
+/// `last_full_snapshot_slot`.  Whereas `snapshot_utils`, `serde_snapshot`, `AccountsBackgroundService`, and
+/// `SnapshotPackagerService` need to be able to set these values when snapshots are taken.
+///
+/// Share this object between all interested parties to ensure a consistent view of the snapshot
+/// information.
+#[derive(Debug)]
+pub struct SnapshotInfo {
```
lijunwangs, line 33: SnapshotInfo can be easily to read as info about a particular snapshot. Maybe SnapshotRuntimeInfo or LastSnapshotInfo is more appropriate?

lijunwangs (dismissed):

runtime/src/snapshot_info.rs
```diff
@@ -0,0 +1,173 @@
...
+    /// The slot when the last full snapshot was taken.  Can be `None` if there are no full
+    /// snapshots yet.
+    last_full_snapshot_slot: Option<Slot>,
+
+    /// The slot when the last incremental snapshot was taken.  Can be `None` if there are no
+    /// incremental snapshots yet.  It is required that this incremental snapshot's base slot is
+    /// the same as the `last_full_snapshot_slot`.
+    last_incremental_snapshot_slot: Option<Slot>,
```
carllin, line 41: might be easier to just make these `pub` fields instead of providing getters and setters.  Some nuances explained here: https://users.rust-lang.org/t/public-getter-method-vs-pub-field/20147

carllin (dismissed):

runtime/src/snapshot_info.rs
brooksprumo: After I put up this PR, then added `SnapshotArchiveInfo` in a different PR, I had the same thought for this one. I'll rename it `SnapshotRuntimeInfo`.

runtime/src/snapshot_info.rs
brooksprumo: Yah, I agree. I wanted to make the "set" case a little better. Since the field is wrapped in an `Option` right now, if you're setting a "last slot" value, then it's never going to be `None` again, so clearing the Option/setting back to `None` should be illegal. Does that seem worth it in this case?

runtime/src/snapshot_info.rs
brooksprumo: Done!

runtime/src/snapshot_info.rs
carllin: Yeah if there's integrity checks, I think the getter/setter methods are worth it!

brooksprumo:
I've renamed the module and types to be `snapshot_runtime_info` and `SnapshotRuntimeInfo`. No functional changes outside of the naming.

Requesting re-review: @lijunwangs @carllin

lijunwangs (approved):
Looks good to me

PR #18200, opened 2021-06-24
rework dirty_pubkeys from insert_new_if_missing_into_primary_index

jeffwashington:
#### Problem
When we rework how we generate the initial accounts index, it becomes difficult to return a mask for finding dirty pubkeys.
#### Summary of Changes
Directly return a list of dirty pubkeys, which is what the calling function has to create anyway.
Fixes #


sakridge (approved):

jeffwashington:
measured performance improvement in the noise. Improvement wasn't the specific goal of this.
```
generate_index total_us=31731750i scan_stores_us=13000509i (this change)
generate_index total_us=32807486i scan_stores_us=12981153i (master)
```

PR #18203, opened 2021-06-24
stop doing clean and shrink at validator startup

jeffwashington:
#### Problem
Validator startup does not need to clean and shrink prior to starting processing transactions. Clean and shrink operate as part of the accounts background service in a running validator.
#### Summary of Changes
Remove clean and shrink from verify_snapshot_bank.
Fixes #

### DIFF
--- a/runtime/src/accounts_background_service.rs
+++ b/runtime/src/accounts_background_service.rs
@@ -77,6 +77,7 @@ impl SendDroppedBankCallback {
 pub struct SnapshotRequest {
     pub snapshot_root_bank: Arc<Bank>,
     pub status_cache_slot_deltas: Vec<BankSlotDelta>,
+    pub just_clean_and_shrink: bool,
 }
 
 pub struct SnapshotRequestHandler {
@@ -102,6 +103,7 @@ impl SnapshotRequestHandler {
                 let SnapshotRequest {
                     snapshot_root_bank,
                     status_cache_slot_deltas,
+                    just_clean_and_shrink,
                 } = snapshot_request;
 
                 let previous_hash = if test_hash_calculation {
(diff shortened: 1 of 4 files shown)


jeffwashington:
@sakridge , can/should we kick the abs service so that it starts cleaning and shrinking immediately on startup? I'm not sure practically how long a validator waits after startup until it requests to create a snapshot.

sakridge:
> @sakridge , can/should we kick the abs service so that it starts cleaning and shrinking immediately on startup? I'm not sure practically how long a validator waits after startup until it requests to create a snapshot.

Yes. Especially if we remove this, I think we need to start it as soon as possible.

Right now the startup is structured to replay the local ledger first before starting any network services and playing newer blocks. This local replay is done through the special blockstore_processor path which does an explicit clean/shrink every 10 seconds or so. Instead we could have the accounts background service handle the shrinking/cleaning just as it does when the validator is doing in the steady-state replay_stage playback of ledger.

It's somewhat an open question if the validator should be creating snapshots during the local blockstore replay stage, we don't do it today, so matching that would be fine. On the other hand it might actually be nice to do it, consider if the validator has many slots to replay and encounters some error mid-way through processing. If it created a newer snapshot, it wouldn't have to replay as many blocks when it starts up. On the other hand, it takes resources to create those snapshots.

jeffwashington:
Stopping this so we don't waste build resources. Working on starting up cleans early on.

jeffwashington:
@sakridge I implemented 'kicking' the abs service in tvu::new. I tested it on a validator and it seems to run as I expected. I'm not sure if we should start it earlier during local replay somehow.

runtime/src/accounts_background_service.rs
```diff
@@ -120,7 +122,7 @@ impl SnapshotRequestHandler {
                 shrink_time.stop();
 
                 let mut flush_accounts_cache_time = Measure::start("flush_accounts_cache_time");
-                if accounts_db_caching_enabled {
+                if accounts_db_caching_enabled && !just_clean_and_shrink {
                     // Forced cache flushing MUST flush all roots <= snapshot_root_bank.slot().
                     // That's because `snapshot_root_bank.slot()` must be root at this point,
                     // and contains relevant updates because each bank has at least 1 account update due
```
jeffwashington, line 128: @carllin line 134 below fails when I try to kick off a clean earlier. Clearly something isn't quite happy. I haven't had a chance to dig into this failure yet. So, for now, I disabled the cache flush. That may not be correct depending on if we've been doing local replay and there are contents in the cache.

sakridge:
This was my attempt a while ago for reference if that helps: https://github.com/solana-labs/solana/pull/17307

Issue #16986, opened 2021-05-01
AccountsDb replication for RPC


carllin:
#### Problem
Validators fall behind the network when bogged down by heavy RPC load.

Tag @aeyakovenko, @sakridge, @t-nelson, @ryoqun, @CriesofCarrots 

#### Proposed Solution: AccountsDb Streaming

Create replicas that stream per-slot accounts updates from a validator, and otherwise just manage AccountsDb and AccountsBackgroundService clean + shrink.

The protocol would look roughly like:

1) Validator maintains some TCP connections to backing replicas.

2) On replica starts up and establishes a connection to the validator, they do one of the following:
       a) Grab a snapshot for root `S` from the validator, along with all the rooted + *unrooted* `Storage(storage_entry)` for frozen slots `T >  S` up to the to the `last_cleaned_root` by the validator. After sending the storage entry for `last_cleaned_root`, the validator starts sending `Frozen(storage_entry)` or `Frozen(cache_entry)` for future slots `F`, which means the replica after applying `F` should be consistent up to slot `F` (because those have not been cleaned yet).

      b) In the future it may also be possible to stream the startup data from `a)` from other replicas to lessen the burden on the validator. 

3) After the startup process, the replica gets a constant stream of `(Frozen(cache_entry)` from the validator's replication background thread. The validator's replication background thread receives a clone of the `Arc<SlotCacheInner>` for each slot on bank freeze and forwards this info to the replica. In the future, we could also build a Turbine-like structure such that the validator does not have the burden of propagating all of these updates to every replica

The validator also streams roots + optimistic confirmation so that the replica's clean + shrink logic can prune outdated data, and also answer RPC queries about rooted + optimistically confirmed accounts.

I think if we want to support `Confirmation` levels outside of `root + optimistically confirmed` we may also need to stream parent slots so that the replica can know the fork structure on queries

Bandwidth estimate of streaming:
https://metrics.solana.com:8888/sources/3/chronograf/data-explorer?query=SELECT%20mean%28%22total_unique_writes_size%22%29%20AS%20%22mean_total_unique_writes_size%22%2C%20mean%28%22num_slots%22%29%20AS%20%22mean_num_slots%22%20FROM%20%22mainnet-beta%22.%22autogen%22.%22accounts_cache_size%22%20WHERE%20time%20%3E%20now%28%29%20-%201h%20GROUP%20BY%20time%28%3Ainterval%3A%29%20FILL%28null%29

[...]

PR #18206, opened 2021-06-24
Add design proposal for accountsDB replication

CriesofCarrots:
#### Problem

Validators fall behind the network when bogged down by heavy RPC load. https://github.com/solana-labs/solana/issues/16986 outlines some possible solutions; time to pick one and formalize our approach.


CriesofCarrots:
302 https://github.com/solana-labs/solana/pull/18651

PR #18207, opened 2021-06-24
Plumbing for `snapshot_info`

brooksprumo:
Do the plumbing for `snapshot_info` (PR #18199).

Additionally, set the last full snapshot slot in `SnapshotPackagerService` and `snapshot_utils`. This will be expanded when Incremental Snapshot support is added. As of now, AccountsDb does not use this information.

Related to Incremental Snapshots, #17088, and broken out from PR #17875.

### DIFF
--- a/core/src/validator.rs
+++ b/core/src/validator.rs
@@ -57,9 +57,10 @@ use solana_runtime::{
     accounts_db::AccountShrinkThreshold,
     accounts_index::AccountSecondaryIndexes,
     bank::Bank,
-    bank_forks::{BankForks, SnapshotConfig},
+    bank_forks::BankForks,
     commitment::BlockCommitmentCache,
     hardened_unpack::{open_genesis_config, MAX_GENESIS_ARCHIVE_UNPACKED_SIZE},
+    snapshot_config::SnapshotConfig,
     snapshot_runtime_info::SyncSnapshotRuntimeInfo,
 };
 use solana_sdk::{
--- a/core/src/validator.rs
+++ b/core/src/validator.rs
@@ -60,6 +60,7 @@ use solana_runtime::{
     bank_forks::{BankForks, SnapshotConfig},
     commitment::BlockCommitmentCache,
     hardened_unpack::{open_genesis_config, MAX_GENESIS_ARCHIVE_UNPACKED_SIZE},
+    snapshot_runtime_info::SyncSnapshotRuntimeInfo,
 };
 use solana_sdk::{
     clock::Slot,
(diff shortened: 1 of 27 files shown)


core/src/validator.rs
```diff
@@ -359,6 +360,7 @@ impl Validator {
                 .register_exit(Box::new(move || exit.store(true, Ordering::Relaxed)));
         }
 
+        let snapshot_info = SyncSnapshotInfo::default();
```
carllin, line 363: If we always have a `snapshot_info`, is there a reason to pass an `Option` around instead of just the `snapshot_info`

brooksprumo: Yeah, you're right, that's a good point. I was originally thinking that since snapshots can be disabled, that I should put an `Option` around the whole thing to indicate as such. The `default()` here is kind of a placeholder until Incremental Snapshots gets hooked up the the configuration params of the validator properly. 

I guess I probably only need *either* (1) `Option` around the whole `SnapshotInfo`, *or* (2) `Option` on each field of `SnapshotInfo`. Does that sound right to you? If that's right, then I'm leaning more towards (1) than (2).

(The one corner case that I'm not sure matters is if a validator enables full snapshots but disables incremental snapshots. In that case, per-field `Option` better reflects what's possible, config-wise. But is that likely to happen/matter?)

brooksprumo: And now that I'm reading this again, maybe (2) makes more sense...

brooksprumo: I remember again, so there's a possibility where snapshots are both enabled and no snapshots have been taken yet. In that case, the whole `snapshot_info` would be `Some`, but the individual fields would be `None`.

carllin: > I was originally thinking that since snapshots can be disabled, that I should put an Option around the whole thing to indicate as such.

We already do have a `snapshot_config: Option<&SnapshotConfig>` that's being passed around which is `None` when the snapshots are disabled. I think it would be cleaner then to also pass the `snapshot_info` inside `snapshot_config` since they are either both `Some` or both `None`

brooksprumo: OK, will do!

brooksprumo: @carllin For places that have a `SyncSnapshotRuntimeInfo` but do not currently have a `SnapshotConfig` (i.e. `Bank`, `Accounts`, `AccountsDb`, and `serde_snapshot`), would you like me to pipe a `SnapshotConfig` through all those instead of the `SyncSnapshotRuntimeInfo`?

From what I can tell also, the `SnapshotConfig` is copied/moved around everywhere now, so it's not shared, which is different than the usage for `SyncSnapshotRuntimeInfo`. Is there any issue with this mixed use?

lijunwangs (approved):
Looks good to me. Nice refactoring

PR #18234, opened 2021-06-25
Added notes to running validator documentation to ensure that 'exec' …

bji:
…is used from scripts when starting the validator to prevent logrotate signals from killing the validator.

#### Summary of Changes

Fixes #


mvines (approved):
Thank you ser

PR #18238, opened 2021-06-25
encapsulates turbine peers computations of broadcast & retransmit stages

behzadnouri:
#### Problem
Broadcast stage and retransmit stage should arrange nodes on turbine
broadcast tree in exactly same order. Additionally any changes to this
ordering (e.g. updating how unstaked nodes are handled) requires feature
gating to keep the cluster in sync.

Current implementation is scattered out over several public methods and
exposes too much of implementation details (e.g. usize indices into
peers vector) which makes code changes and checking for feature
activations more difficult.

#### Summary of Changes

This commit encapsulates turbine peer computations into a new struct,
and only exposes two public methods, get_broadcast_peer and
get_retransmit_peers, for call-sites.

### DIFF
--- a/core/src/broadcast_stage.rs
+++ b/core/src/broadcast_stage.rs
@@ -6,17 +6,15 @@ use self::{
     fail_entry_verification_broadcast_run::FailEntryVerificationBroadcastRun,
     standard_broadcast_run::StandardBroadcastRun,
 };
-use crate::result::{Error, Result};
+use crate::{
+    cluster_nodes::ClusterNodes,
+    result::{Error, Result},
+};
 use crossbeam_channel::{
     Receiver as CrossbeamReceiver, RecvTimeoutError as CrossbeamRecvTimeoutError,
     Sender as CrossbeamSender,
 };
-use solana_gossip::{
-    cluster_info::{self, ClusterInfo, ClusterInfoError},
-    contact_info::ContactInfo,
-    crds_gossip_pull::CRDS_GOSSIP_PULL_CRDS_TIMEOUT_MS,
-    weighted_shuffle::weighted_best,
-};
+use solana_gossip::cluster_info::{ClusterInfo, ClusterInfoError};
 use solana_ledger::{blockstore::Blockstore, shred::Shred};
 use solana_measure::measure::Measure;
 use solana_metrics::{inc_new_counter_error, inc_new_counter_info};
@@ -390,26 +388,16 @@ fn update_peer_stats(
     }
 }
 
-pub fn get_broadcast_peers(
-    cluster_info: &ClusterInfo,
-    stakes: Option<&HashMap<Pubkey, u64>>,
-) -> (Vec<ContactInfo>, Vec<(u64, usize)>) {
-    let mut peers = cluster_info.tvu_peers();
-    let peers_and_stakes = cluster_info::stake_weight_peers(&mut peers, stakes);
-    (peers, peers_and_stakes)
-}
-
 /// broadcast messages from the leader to layer 1 nodes
 /// # Remarks
 pub fn broadcast_shreds(
     s: &UdpSocket,
     shreds: &[Shred],
-    peers_and_stakes: &[(u64, usize)],
-    peers: &[ContactInfo],
+    cluster_nodes: &ClusterNodes<BroadcastStage>,
     last_datapoint_submit: &Arc<AtomicU64>,
     transmit_stats: &mut TransmitShredsStats,
 ) -> Result<()> {
-    let broadcast_len = peers_and_stakes.len();
+    let broadcast_len = cluster_nodes.num_peers();
     if broadcast_len == 0 {
         update_peer_stats(1, 1, last_datapoint_submit);
         return Ok(());
@@ -417,10 +405,9 @@ pub fn broadcast_shreds(
     let mut shred_select = Measure::start("shred_select");
     let packets: Vec<_> = shreds
         .iter()
-        .map(|shred| {
-            let broadcast_index = weighted_best(peers_and_stakes, shred.seed());
-
-            (&shred.payload, &peers[broadcast_index].tvu)
+        .filter_map(|shred| {
+            let node = cluster_nodes.get_broadcast_peer(shred.seed())?;
+            Some((&shred.payload, &node.tvu))
         })
         .collect();
     shred_select.stop();
@@ -439,32 +426,13 @@ pub fn broadcast_shreds(
     send_mmsg_time.stop();
     transmit_stats.send_mmsg_elapsed += send_mmsg_time.as_us();
 
-    let num_live_peers = num_live_peers(peers);
+    let num_live_peers = cluster_nodes.num_peers_live(timestamp()) as i64;
     update_peer_stats(
         num_live_peers,
         broadcast_len as i64 + 1,
         last_datapoint_submit,
     );
     Ok(())
-}
-
-fn distance(a: u64, b: u64) -> u64 {
-    if a > b {
-        a - b
-    } else {
-        b - a
-    }
-}
-
-fn num_live_peers(peers: &[ContactInfo]) -> i64 {
-    let mut num_live_peers = 1i64;
-    peers.iter().for_each(|p| {
-        // A peer is considered live if they generated their contact info recently
-        if distance(timestamp(), p.wallclock) <= CRDS_GOSSIP_PULL_CRDS_TIMEOUT_MS {
-            num_live_peers += 1;
-        }
-    });
-    num_live_peers
 }
 
 #[cfg(test)]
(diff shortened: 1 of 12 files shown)


core/src/broadcast_stage.rs
```diff
@@ -388,37 +386,26 @@ fn update_peer_stats(
...
     let mut shred_select = Measure::start("shred_select");
     let packets: Vec<_> = shreds
         .iter()
-        .map(|shred| {
-            let broadcast_index = weighted_best(peers_and_stakes, shred.seed());
-
-            (&shred.payload, &peers[broadcast_index].tvu)
+        .filter_map(|shred| {
+            let node = cluster_nodes.get_broadcast_peer(shred.seed())?;
+            Some((&shred.payload, &node.tvu))
```
carllin, line 408: This should always be `Some` because of the` broadcast_len() == 0` check above right?

core/src/cluster_nodes.rs
```diff
@@ -0,0 +1,441 @@
...
+    }
+}
+
+fn new_cluster_nodes<T: 'static>(
+    cluster_info: &ClusterInfo,
+    stakes: &HashMap<Pubkey, u64>,
+) -> ClusterNodes<T> {
+    let self_pubkey = cluster_info.id();
+    let nodes = get_nodes(cluster_info, stakes);
+    let broadcast = TypeId::of::<T>() == TypeId::of::<BroadcastStage>();
```
carllin, line 158: nit: `broadcast` -> `is_broadcast`

core/src/cluster_nodes.rs
```diff
@@ -0,0 +1,441 @@
...
+pub struct ClusterNodes<T> {
+    pubkey: Pubkey, // The local node itself.
+    // All staked nodes + other known tvu-peers + the node itself;
+    // sorted by (stake, pubkey) in descending order.
+    nodes: Vec<Node>,
+    // Weights and indices for sampling peers. weighted_{shuffle,best} expect
+    // weights >= 1. For backward compatibility we use max(1, stake) for
+    // weights and exclude nodes with no contact-info.
+    index: Vec<(/*weight:*/ u64, /*index:*/ usize)>,
+    _phantom: PhantomData<T>,
```
carllin, line 35: nice, first `PhantomData` in core 😃

core/src/cluster_nodes.rs
```diff
@@ -0,0 +1,441 @@
...
+    stakes: &HashMap<Pubkey, u64>,
+) -> ClusterNodes<T> {
+    let self_pubkey = cluster_info.id();
+    let nodes = get_nodes(cluster_info, stakes);
+    let broadcast = TypeId::of::<T>() == TypeId::of::<BroadcastStage>();
+    // For backward compatibility:
+    //   * nodes which do not have contact-info are excluded.
+    //   * stakes are floored at 1.
+    // The sorting key here should be equivalent to
+    // ClusterInfo::sorted_stakes_with_index.
```
carllin, line 163: nit: equivalent to `solana_gossip::deprecated::sorted_stakes_with_index`

core/src/cluster_nodes.rs
```diff
@@ -0,0 +1,441 @@
...
+            .iter()
+            .filter(|(_, stake)| **stake > 0)
+            .map(|(&pubkey, &stake)| Node {
+                node: NodeId::from(pubkey),
+                stake,
+            }),
+    )
+    .sorted_by_key(|node| Reverse((node.stake, node.pubkey())))
+    // Since sorted_by_key is stable, in case of duplicates, this
+    // will keep nodes with contact-info.
```
carllin, line 209: This is because in the enum, the `ContactInfo` variant is declared before the `Pubkey` variant?

core/src/cluster_nodes.rs
```diff
@@ -0,0 +1,441 @@
...
+        cluster_info::{compute_retransmit_peers, ClusterInfo},
+        contact_info::ContactInfo,
+        crds_gossip_pull::CRDS_GOSSIP_PULL_CRDS_TIMEOUT_MS,
+        weighted_shuffle::{weighted_best, weighted_shuffle},
+    },
+    solana_sdk::pubkey::Pubkey,
+    std::{any::TypeId, cmp::Reverse, collections::HashMap, marker::PhantomData},
+};
+
+enum NodeId {
```
carllin, line 14: How about `NodeGossipStatus`, with the variants being, `AliveNode(ContactInfo)`, and `DeadStakedNode(Pubkey)`

core/src/cluster_nodes.rs
```diff
@@ -0,0 +1,441 @@
...
+        }
+    }
+}
+
+impl ClusterNodes<RetransmitStage> {
+    pub fn new(cluster_info: &ClusterInfo, stakes: &HashMap<Pubkey, u64>) -> Self {
+        new_cluster_nodes(cluster_info, stakes)
+    }
+
+    pub fn get_retransmit_peers(
```
carllin, line 103: note to self: should be equivalent to the deleted code:https://github.com/solana-labs/solana/blob/master/core/src/retransmit_stage.rs#L282-L305

core/src/cluster_nodes.rs
```diff
@@ -0,0 +1,441 @@
...
+            None
+        } else {
+            let index = weighted_best(&self.index, shred_seed);
+            match &self.nodes[index].node {
+                NodeId::ContactInfo(node) => Some(node),
+                NodeId::Pubkey(_) => panic!("this should not happen!"),
+            }
+        }
+    }
+}
```
carllin, line 96: not to self: should be equivalent to deleted code: https://github.com/solana-labs/solana/blob/master/core/src/broadcast_stage.rs#L421-L423

core/src/cluster_nodes.rs
carllin: tag @jbiseda, would be awesome if you could double check this as well 😃

core/src/cluster_nodes.rs
carllin: tag @jbiseda, would be awesome if you could double check this as well!

core/src/cluster_nodes.rs
```diff
@@ -0,0 +1,441 @@
...
+        };
+        let index: Vec<_> = {
+            let shuffle = weighted_shuffle(&weights, shred_seed);
+            shuffle.into_iter().map(|i| index[i]).collect()
+        };
+        let self_index = index
+            .iter()
+            .position(|i| self.nodes[*i].pubkey() == self.pubkey)
+            .unwrap();
+        let (neighbors, children) = compute_retransmit_peers(fanout, self_index, &index);
```
carllin, line 135: Is this filter: https://github.com/solana-labs/solana/blob/e7b36c84841e7a3e8d635bec2e7c0ee18150f212/core/src/retransmit_stage.rs#L443-L447 for `self` no longer necessary on `the neighbors`?

core/src/broadcast_stage.rs
behzadnouri: yeah, but that is an internal implementation detail in `get_broadcast_peer` and here we do not want to rely on that not changing.

core/src/cluster_nodes.rs
behzadnouri: This struct is meant to include node identifying info. No need to tie it to gossip. Also, this is only a private type not exposed outside of this file, so leaning on brevity here.

core/src/cluster_nodes.rs
behzadnouri: yes, this is tested in `test_cluster_nodes_broadcast`

core/src/cluster_nodes.rs
behzadnouri: yes, tested in `test_cluster_nodes_retransmit`

core/src/cluster_nodes.rs
behzadnouri: That logic, like before, is in `retransmit_stage.rs`, where it says:
```
            // First neighbor is this node itself, so skip it.
            ClusterInfo::retransmit_to(
                &neighbors[1..],
```

core/src/cluster_nodes.rs
behzadnouri: done.

core/src/cluster_nodes.rs
behzadnouri: No, it is because in the above iterator `chain`, all `NodeId:ContactInfo` come before all `NodeId:Pubkey`

core/src/broadcast_stage.rs
carllin: yup, just confirming my understanding, thanks!

core/src/cluster_nodes.rs
carllin: fine by me!

core/src/cluster_nodes.rs
carllin: Ah, missed that bit, thanks!

carllin (approved):
Looks good, great stuff 😃

core/src/cluster_nodes.rs
carllin: Ah yup, make sense, thanks!

PR #18257, opened 2021-06-28
feat: add filters support to the connection.onProgramAccountChange method

### LINKED ISSUE #18256: Support RPC filters in web3.js
#### Problem
I'm looking to subscribe to a program via the web3.js [`Connection.onProgramAccountChange`](https://solana-labs.github.io/solana-web3.js/classes/connection.html#onprogramaccountchange), but I'd like to use the optional filter that comes with the underlying RPC [`programSubscribe`](https://docs.solana.com/developing/clients/jsonrpc-api#programsubscribe) request.

#### Proposed Solution
Update the build arguments of a subscription to include optional filters argument. As of writing the `_buildArgs` method seems to only take in a programId and commitment. It seems like it could be easily passed in to the _extra_ argument of the `buildArgs`.

1. Update [onProgramAccountChange](https://github.com/solana-labs/solana/blob/2fbedd834fa48f09d87c5d60a9a820412f513324/web3.js/src/connection.ts#L3841) to include optional filters in the subscription object
2. update program subscription builder ([here](https://github.com/solana-labs/solana/blob/2fbedd834fa48f09d87c5d60a9a820412f513324/web3.js/src/connection.ts#L3711)) to pass in the filters as the _extra_ arg

tomjohn1028:
#### Problem
Looking to subscribe to a program via the web3.js Connection.onProgramAccountChange, but I'd like to use the optional filter that comes with the underlying RPC programSubscribe request.

#### Summary of Changes
Adds option _filters_ argument to `onProgramAccountChange` which gets built into the RPC arguments.

Fixes #18256 

(original tests pass but I haven't validated this code with an actual RPC connection yet)

### DIFF
--- a/web3.js/src/connection.ts
+++ b/web3.js/src/connection.ts
@@ -1725,7 +1725,7 @@ type ProgramAccountSubscriptionInfo = {
   callback: ProgramAccountChangeCallback;
   commitment?: Commitment;
   subscriptionId: SubscriptionId | null; // null when there's no current server subscription id
-  filters?: GetProgramAccountsFilter[]
+  filters?: GetProgramAccountsFilter[];
 };
 
 /**
@@ -3709,7 +3709,9 @@ export class Connection {
       this._subscribe(
         sub,
         'programSubscribe',
-        this._buildArgs([sub.programId], sub.commitment, 'base64', {filters: sub.filters}),
+        this._buildArgs([sub.programId], sub.commitment, 'base64', {
+          filters: sub.filters,
+        }),
       );
     }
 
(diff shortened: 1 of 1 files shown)


mvines:
Hey looks like Travis CI is broken, it is incorrectly flagging this PR as abuse.  Common problem, they think blockchain is abuse.  I've contacted their support.

In the meantime, once Travis CI is back up it'll complain that the commit in this PR doesn't follow the Angular commit convention.  Try fixing that by squashing the two commit in here and giving the one commit the title:
"feat: add filters support to the connection.onProgramAccountChange method"

jstarry (approved):
looks good, thanks!

PR #18304, opened 2021-06-29
Fail simulation if transaction contains duplicate accounts

jstarry:
#### Problem
Simulation allows processing invalid transactions with duplicate accounts.


jstarry:
low risk change, merging!

PR #18205, opened 2021-06-24
during untar, read ahead, causing decompression to occur in the bg

jeffwashington:
#### Problem
The untar process is:

1. open file
2. open decompressor
3. open archive
4. start reading entries in archive

Reading the archive is:
1. archive reader reads something like a header or data
2. which results in a read to the decompressor
3. which results in the decompressor reading some bytes from the file and decompressing and returning
4. archive reader then processes and checks the bytes
5. and maybe writes some bytes to disk
6. and then, goto 1 until archive is over

while 4-6 are going on, no reading or decompression can be done on the original file. Thus, 4 is blocked by 1-3.

#### Summary of Changes
Read and decompress ahead in the background, so that the archive has immediate access to decompressed data. Later, we can also use the fact that the decompressed data is already in memory to read it directly (such as pulling accounts out of it). We can also write it asynchronously to disk so we aren't blocked on that. This change is a building block for other optimizations related to snapshots and startup operations. We can hold the current mnb decompressed snapshots entirely in memory while bringing up a validator, but we cannot hold the 360M account snapshots all in memory, so we'll have to work on the file in phases, governed by disk io and memory constraints in an async pipeline model at some point.
Fixes #

PR #18310, opened 2021-06-29
refactor untar_snapshot_in to push parallelism deeper for further refactoring

jeffwashington:
#### Problem
Startup takes a long time. Untar of snapshot takes a long time. There is time to be saved by improving parallelism and shared data across parallel untar paths. Related to #18205
#### Summary of Changes
Refactor to push paralleism deeper.
Fixes #


jeffwashington:
This change doesn't save time by itself. It reworks where we do parallelism so that Archive can be created with a shared, buffered reader from unzip or whatever format the snapshot is in.

brooksprumo (approved):
Looks good to me

PR #18314, opened 2021-06-30
Fixed an issue doing the set_roots repeatedly for the same set.

lijunwangs:
#### Problem
Iterating of chunks is being used but the code is doing the set_roots for the whole vector repeatedly.

#### Summary of Changes
set_roots per chunks.

Fixes #


CriesofCarrots:
Thanks! Might as well fixup stable and beta branches...

CriesofCarrots (approved):

PR #18328, opened 2021-06-30
Add sol_ed25519_sig_check syscall

seanyoung:
Signed-off-by: Sean Young <sean@mess.org>

#### Problem
For Solidity using Solang, a method is needed where an address (=account) can sign a message client side and then the signature can be verified at runtime in a smart contract. Since solana accounts are ed25519, this needs to be an ed25519 signature check.

#### Summary of Changes
Adds new system call and tests/sdk changes for rust and c.

The cost is just a place holder. There is a rust benchmark which shows the sig check takes 45 microseconds:

```
$ rustup run nightly cargo bench
    Finished bench [optimized] target(s) in 0.19s
     Running unittests (target/release/deps/bpf_loader-5434cd00f483d7aa)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running unittests (target/release/deps/ed25519_sig_check-f819e8f57f23532f)

running 1 test
test ed25519_sig_check ... bench:      44,061 ns/iter (+/- 2,183)

test result: ok. 0 passed; 0 failed; 0 ignored; 1 measured; 0 filtered out; finished in 0.76s
```


sdk/bpf/c/inc/solana_sdk.h
```diff
@@ -457,6 +457,34 @@ uint64_t sol_keccak256(
     uint8_t *result
 );
 
+/** The ed25519 public key is not a valid public key */
```
jackcmay, line 460: Can we move this into its own header?

PR #18330, opened 2021-06-30
`solana validators` output now includes average skip rate

mvines:
The average skip rate displayed on https://solanabeach.io/ is consistently wrong.  Fight misinformation by adding it to `solana validators` (including JSON output).

The true mainnet beta skiprate at this time:
```
Average Stake Weighted Skip Rate: 23.44%
Average Non-weighted Skip Rate:   18.99%
```
which is still demonstrates a fat whaley stake distribution where the little nodes are outperforming the whales on average, but a far cry from the 51% skip rate that solana beach currently reports

### DIFF
--- a/cli-output/src/cli_output.rs
+++ b/cli-output/src/cli_output.rs
@@ -347,6 +347,8 @@ pub struct CliValidators {
     pub total_current_stake: u64,
     pub total_delinquent_stake: u64,
     pub validators: Vec<CliValidator>,
+    pub average_skip_rate: f64,
+    pub average_stake_weighted_skip_rate: f64,
     #[serde(skip_serializing)]
     pub validators_sort_order: CliValidatorsSortOrder,
     #[serde(skip_serializing)]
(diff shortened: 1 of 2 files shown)


cli/src/cluster_query.rs
```diff
@@ -1892,14 +1892,40 @@ pub fn process_show_validators(
...
+        let mut skip_rate_weighted_sum = 0.;
+        for validator in validators.iter() {
+            if let Some(skip_rate) = validator.skip_rate {
+                skip_rate_sum += skip_rate;
+                skip_rate_len += 1;
+                skip_rate_weighted_sum += skip_rate * validator.activated_stake as f64;
+            }
+        }
+
+        if skip_rate_len > 0 {
```
CriesofCarrots, line 1912:
```suggestion
        if skip_rate_len > 0 && total_active_stake > 0 {
```
Since we're already checking for something we think is impossible, might as well prevent that NaN potentiality?

cli-output/src/cli_output.rs
```diff
@@ -510,6 +512,18 @@ impl fmt::Display for CliValidators {
             writeln!(f, "{}", header)?;
         }
 
+        writeln!(f)?;
+        writeln_name_value(
+            f,
+            "Average Stake Weighted Skip Rate:",
```
CriesofCarrots, line 518:
```suggestion
            "Average Stake-Weighted Skip Rate:",
```

cli-output/src/cli_output.rs
```diff
@@ -510,6 +512,18 @@ impl fmt::Display for CliValidators {
...
+            &format!("{:.2}%", self.average_stake_weighted_skip_rate,),
+        )?;
+        writeln_name_value(
+            f,
+            "Average Non-weighted Skip Rate:  ",
```
CriesofCarrots, line 523:
```suggestion
            "Average Unweighted Skip Rate:  ",
```

CriesofCarrots (dismissed):
Lgtm! Nits only

PR #18333, opened 2021-06-30
Consider all peers as potential candidates during pull-request in case of offline nodes

### LINKED ISSUE #18279: Re-enable optimistic-confirmation local-cluster tests
#### Problem
 #18278 ignores 2 local-cluster tests failing consistently in CI

#### Proposed Solution
Identify root cause of test failures, fix, and re-enable them

AshwinSekar:
#### Problem
When we receive contact info from an offline node in our list of peers, we assign it a much higher weight than other peers because we have not sent a pull request to it yet. This causes the sampling to always select only this peer to send pull requests to. However the ping check never allows us to send a pull request and update the timestamp, as the node is offline. Thus we never get a chance to send pull requests to other nodes.

#### Summary of Changes
After assigning the weights, instead of performing random sampling we use a weighted shuffle to ensure that if the higher weighted nodes fail the ping check we can still send pull requests to the other nodes. 
Also reenables `test_no_optimistic_confirmation_violation_with_tower` and `test_optimistic_confirmation_violation_without_tower`

Fixes #18279

### DIFF
--- a/gossip/src/crds_gossip_pull.rs
+++ b/gossip/src/crds_gossip_pull.rs
@@ -978,7 +978,7 @@ pub(crate) mod tests {
         node.mark_pull_request_creation_time(new.contact_info().unwrap().id, now);
         let offline = ContactInfo::new_localhost(&solana_sdk::pubkey::new_rand(), now);
         let offline = CrdsValue::new_unsigned(CrdsData::ContactInfo(offline));
-        crds.insert(offline.clone(), now).unwrap();
+        crds.insert(offline, now).unwrap();
         let req = node.new_pull_request(
             &thread_pool,
             &crds,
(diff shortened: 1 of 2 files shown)


sakridge:
All peers will be all 500 or 1000 nodes?

AshwinSekar:
Sorry about the wording we are still only sending 1 pull request. 

Instead of taking `2 * # nodes` samples of the peers and finding the first available to send a pull request to, we do a weighted shuffle amongst all the peers to find the first available. The idea is that if offline nodes have a large weight, all the samples could result in offline nodes which prevents us from sending any pull requests.

behzadnouri (dismissed):
LGTM, thanks for debugging the local-cluster tests

CriesofCarrots:
v1.7 backport?

PR #18338, opened 2021-06-30
report on min/max bin stats for accounts index

jeffwashington:
#### Problem
We now bin accounts index based on the first few bits of the pubkey value. It is important that we have visibility into the binning behavior.
#### Summary of Changes
add metrics
Fixes #


jeffwashington:
almost perfectly balanced bins. ie. uniform distribution of pubkeys
```
generate_index total_us=16595216 scan_stores_us=16823797 insertion_time_us=51218396 min_bin_size=3888540 max_bin_size=3896996 total_items=62272320
```

sakridge (approved):

PR #18343, opened 2021-06-30
implements an unbiased weighted shuffle using binary indexed tree

behzadnouri:
#### Problem
Current implementation of weighted_shuffle:
https://github.com/solana-labs/solana/blob/b08f8bd1b/gossip/src/weighted_shuffle.rs#L11-L37
uses a heuristic which results in biased samples.

For example, if the weights are [1, 10, 100], then the 3rd index should
come first 100 times more often than the 1st index. However,
weighted_shuffle is picking the 3rd index 200+ times more often than the
1st index, showing a disproportional bias in favor of higher weights.

#### Summary of Changes

This commit implements weighted shuffle using binary indexed tree to
maintain cumulative sum of weights while sampling. The resulting samples
are demonstrably unbiased and precisely proportional to the weights.

Additionally the iterator interface allows to skip computations when
not all indices are processed.

Of the use cases of weighted_shuffle, changing turbine code requires
feature-gating to keep the cluster in sync. That is not updated in
this commit, but can be done together with future updates to turbine.

### DIFF
--- a/gossip/src/crds_gossip_pull.rs
+++ b/gossip/src/crds_gossip_pull.rs
@@ -18,7 +18,7 @@ use {
         crds_gossip_error::CrdsGossipError,
         crds_value::CrdsValue,
         ping_pong::PingCache,
-        weighted_shuffle::weighted_shuffle,
+        weighted_shuffle::WeightedShuffle,
     },
     itertools::Itertools,
     lru::LruCache,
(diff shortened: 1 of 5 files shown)


gossip/src/crds_gossip_pull.rs
```diff
@@ -307,7 +304,9 @@ impl CrdsGossipPull {
                 let since = (now.saturating_sub(req_time).min(3600 * 1000) / 1024) as u32;
                 let stake = get_stake(&item.id, stakes);
                 let weight = get_weight(max_weight, since, stake);
-                (weight, item)
+                // Weights are bounded by max_weight defined above.
+                // So this type-cast should be safe.
+                ((weight * 100.0) as u64, item)
```
jbiseda, line 309: Add a function for this calculation to use here and in crds_gossip_push.rs?

behzadnouri: Yeah, I think might be good to refactor `push_options` and `pull_options` to share the common parts. But I would rather not to mix that refactoring with this pull-request.

jbiseda (approved):
looks good

PR #18353, opened 2021-07-01
Add storage-proto build.rs and readme

CriesofCarrots:
#### Problem
It is easy to edit rust protobuf-generated files manually, but these changes will be overwritten the next time someone builds them from the .proto.

#### Summary of Changes
~~Add a readme try to make it slightly obvious how updates should be done
Also fixup recent change~~

(edit) Add build.rs to build protobufs on demand

One note for future readers: this treatment was not applied to `storage-bigtable` because the generated googleapi rust files require several manual edits (to satisfy clippy and doc tests).


CriesofCarrots:
@t-nelson I'm tagging you for review just because you happen to be the most recent person to fall into this trap. Sorry!

I tried adding comments to the generated rust files, but thought they ended up being more confusing than helpful, since they have to appear in the protobuf files as well.

t-nelson:
Hmm.  Do the generated Rust files need to be under revision control?  Seems like a job for `build.rs`

CriesofCarrots:
> Hmm. Do the generated Rust files need to be under revision control? Seems like a job for `build.rs`

Oo yeah, maybe not. I'll play with build.rs

CriesofCarrots:
How does this look, @t-nelson ? (Ignore the last commit; it's purely cosmetic)

t-nelson:
> How does this look, @t-nelson ? 

Screenshot from 2021-07-09 12-10-23

Yes!

> (Ignore the last commit; it's purely cosmetic)

Thanks for breaking these out!  It's always a headache trying to review a commit that combines them with functional changes

t-nelson (approved):

PR #18231, opened 2021-06-25
Added formats to notifications in WebSocket RPC client.

tiago18c:
#### Problem
Missing documentation in WebSocket RPC client.

#### Summary of Changes
Added the formats to the notification objects in the WebSocket RPC client.

PR #18354, opened 2021-07-01
Docs: Update notification format info to be consistent

CriesofCarrots:
#### Summary of Changes
Follow-up to #18231 to update newly documented subscription and make consistent


solana-grimes:

PR #18357, opened 2021-07-01
Make set roots an iterator

carllin:
#### Problem
Blockstore `set_roots` could be more generic

#### Summary of Changes
Make `set_roots` take any iterator

Fixes #

### DIFF
--- a/ledger/src/blockstore.rs
+++ b/ledger/src/blockstore.rs
@@ -3095,7 +3095,7 @@ impl Blockstore {
                     return Ok(());
                 }
                 trace!("{:?}", chunk);
-                self.set_roots(roots_to_fix.iter())?;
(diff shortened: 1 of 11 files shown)


ledger/src/blockstore.rs
```diff
@@ -3093,7 +3095,7 @@ impl Blockstore {
                     return Ok(());
                 }
                 trace!("{:?}", chunk);
-                self.set_roots(chunk)?;
+                self.set_roots(roots_to_fix.iter())?;
```
jstarry, line 3098: Should be `chunk.iter()`

carllin: ick, great catch, thanks!

jstarry (approved):

PR #18369, opened 2021-07-01
chore: configurable num of retries for send and confirm

hrls:
#### Problem
`RpcClient::send_and_confirm_transaction` logic is hard to understand IMO.

#### Summary of Changes
Refactor method body.
Add support of const configuration for num of retries for sending and receiving confirmation.

### DIFF
--- a/client/src/rpc_client.rs
+++ b/client/src/rpc_client.rs
@@ -1012,49 +1012,55 @@ impl RpcClient {
         &self,
         transaction: &Transaction,
     ) -> ClientResult<Signature> {
-        let signature = self.send_transaction(transaction)?;
-        let recent_blockhash = if uses_durable_nonce(transaction).is_some() {
-            self.get_recent_blockhash_with_commitment(CommitmentConfig::processed())?
-                .value
-                .0
-        } else {
-            transaction.message.recent_blockhash
-        };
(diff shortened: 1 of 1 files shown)


client/src/rpc_client.rs
```diff
@@ -1012,49 +1012,55 @@ impl RpcClient {
         &self,
         transaction: &Transaction,
     ) -> ClientResult<Signature> {
```
mvines, line 1014: I'm a little uncomfortable about changing `send_and_confirm_transaction()` to infinitely retry sending the transaction.  At the very least I'd like to discuss that change in a PR that claims to just be performing a refactoring chore.  

If you'd like to proceed with the transaction retrying, I suggest that:
1. This PR be reduced to refactoring while preserving the existing send_and_confirm_transaction logic
2. A new PR be created to add retries

Thanks for the contribution though!

hrls: Oh, yeah, default value of `SEND_RETRIES` must be `1`, not `MAX`.

https://github.com/solana-labs/solana/pull/18389 this PR contains simple refactoring.

hrls:
https://github.com/solana-labs/solana/pull/18390

PR #18389, opened 2021-07-02
сhore: refactoring RPC client send and confirm

hrls:
#### Problem
`RpcClient::send_and_confirm_transaction` logic is hard to understand IMO.

#### Summary of Changes
Simple refactoring of method body.

PR #18390, opened 2021-07-02
refactor(rpc_client): const number of retries for send and confirm transaction

hrls:
#### Summary of Changes
`RpcClient::send_and_confirm_transaction` body was refactored for re-sending transaction after some number of failed attempts receiving previous transaction status. Currently number of retries is `1`, so function do the same as the current one.

Based on https://github.com/solana-labs/solana/pull/18389


mvines (approved):
Thanks!

I realized that the logic for expiring a transaction here is wrong (predates your change), and wrote up https://github.com/solana-labs/solana/issues/18426 about it. Would you be interested in fixing this up since you're already interested in improving this function? If not no big deal, I'll take care of it sometime

PR #18391, opened 2021-07-02
log warning when channel send fails

taozhu-chicago:
#### Problem
cost_update_sender fails to send intermittently during validator tests. 

#### Summary of Changes
Log warning when send fails instead of panic.

Fixes #


CriesofCarrots (approved):
lgtm, thanks!

Issue #18378, opened 2021-07-02
Rewards are gapped across "Load More" boundaries


ryoqun:
#### Problem

https://explorer.solana.com/address/ArneDNpRpcrFPckzj3kaVwdUS3Di1cYf4b627gqsYMQZ/rewards

BEFORE

![image](https://user-images.githubusercontent.com/117807/124216022-35a74580-db30-11eb-9d25-3034189b4a2b.png)

AFTER

![image](https://user-images.githubusercontent.com/117807/124216035-40fa7100-db30-11eb-840d-b880aef9a884.png)

Notice epoch 181 is missing...

PR #18392, opened 2021-07-02
Explorer: regression in rewards provider where lowestFetched is off by one

oJshua:
#### Problem
Pagination for rewards was off by one (gaps).

#### Summary of Changes
Make sure lowest epoch in provider is correct.

Fixes https://github.com/solana-labs/solana/issues/18378


ryoqun:
@oJshua thanks for super quick fix!

PR #18400, opened 2021-07-02
persists repair-peers cache across repair service loops

behzadnouri:
#### Problem
The repair-peers cache is reset each time repair service loop runs,
and so computed repeatedly for the same slots:
https://github.com/solana-labs/solana/blob/d2b07dca9/core/src/repair_service.rs#L275

#### Summary of Changes
This commit uses an LRU cache to persists repair-peers for each slot.
In addition to LRU eviction rules, in order to avoid re-using outdated
data, each entry also has 10 seconds TTL.

### DIFF
--- a/core/src/serve_repair.rs
+++ b/core/src/serve_repair.rs
@@ -6,7 +6,11 @@ use crate::{
     result::{Error, Result},
 };
 use bincode::serialize;
-use rand::distributions::{Distribution, WeightedIndex};
+use lru::LruCache;
+use rand::{
+    distributions::{Distribution, WeightedError, WeightedIndex},
+    Rng,
+};
 use solana_gossip::{
     cluster_info::{ClusterInfo, ClusterInfoError},
     contact_info::ContactInfo,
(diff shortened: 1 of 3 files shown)


behzadnouri:
Left chunk is this code compare to current master code on the right:

`cluster_info_states2::repair_peers`:
cluster_info_states2::repair_peers

`serve_repair-repair-timing::send-repairs-elapsed`:
serve_repair-repair-timing::send-repairs-elapsed

core/src/serve_repair.rs
```diff
@@ -99,7 +107,38 @@ pub struct ServeRepair {
     cluster_info: Arc<ClusterInfo>,
 }
 
-type RepairCache = HashMap<Slot, (Vec<ContactInfo>, WeightedIndex<u64>)>;
+// Cache entry for repair peers for a slot.
+pub(crate) struct RepairPeers {
+    asof: Instant,
```
carllin, line 112: nit: maybe something more explicit like `insertion_time`?

core/src/serve_repair.rs
```diff
@@ -378,37 +417,33 @@ impl ServeRepair {
...
-                if repair_peers.is_empty() {
-                    return Err(Error::from(ClusterInfoError::NoPeers));
-                }
                 let weights = cluster_slots.compute_weights(slot, &repair_peers);
-                debug_assert_eq!(weights.len(), repair_peers.len());
-                let weighted_index = WeightedIndex::new(weights)?;
-                entry.insert((repair_peers, weighted_index))
+                let repair_peers = RepairPeers::new(Instant::now(), &repair_peers, &weights)?;
+                peers_cache.put(slot, repair_peers);
+                peers_cache.get(&slot).unwrap()
```
carllin, line 440: heh, wish `LruCache` had an entry API

carllin (approved):

PR #18442, opened 2021-07-06
Improve account unlock performance

jstarry:
#### Problem
Allocations for writable / readable keys are made while the `AccountLocks` mutex is locked

#### Summary of Changes
Pre-build account key lists before taking lock

Fixes #


jstarry:
@sakridge this looked like an easy small win, let me know if you want to see a benchmark

sakridge (approved):
Looks good

PR #18445, opened 2021-07-06
Add NodeInstance program

mvines:
The NodeInstance program tracks a validator instance on-chain similarly to how `CrdsData::NodeInstance` does for gossip.

Votes may now be optionally submitted with a NodeInstance.  This'll be used to enable ReplayStage to make assumptions about inflight vote transactions that'll be useful to enable failover to a backup validator, without risk of a lockout violation even when the primary node's tower.bin file is inaccessible.


mvines:
This PR is probably NGMI

PR #18446, opened 2021-07-06
improves parallelism in window-service recv_window

behzadnouri:
#### Problem
sigverify-stage is breaking batches to single-item vectors before
sending them down the channel:
https://github.com/solana-labs/solana/blob/d451363dc/core/src/sigverify_stage.rs#L88-L92

In addition, window-service is processing inner packets.packets in
serial (only outer packets are parallelized):
https://github.com/solana-labs/solana/blob/d451363dc/core/src/window_service.rs#L244-L246

#### Summary of Changes
This commit:
  * sends the packets in batches from sigverify-stage.
  * uses `flat_map` to parallelize window-service processing of packets
    down to individual packets.
  * simplifies window-service code, reducing number of nested branches.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2317,6 +2317,25 @@ dependencies = [
 
 [[package]]
 name = "libsecp256k1"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bd1137239ab33b41aa9637a88a28249e5e70c40a42ccc92db7f12cc356c1fcd7"
+dependencies = [
+ "arrayref",
+ "base64 0.12.3",
+ "digest 0.9.0",
+ "hmac-drbg",
+ "libsecp256k1-core",
+ "libsecp256k1-gen-ecmult",
+ "libsecp256k1-gen-genmult",
+ "rand 0.7.3",
+ "serde",
+ "sha2 0.9.5",
+ "typenum",
+]
+
+[[package]]
+name = "libsecp256k1"
 version = "0.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c9d220bc1feda2ac231cb78c3d26f27676b8cf82c96971f7aeef3d0cf2797c73"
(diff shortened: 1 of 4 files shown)


sakridge:
Benchmark of this?

Having each packet be a work-item could increase synchronization overhead.

behzadnouri:
> Having each packet be a work-item could increase synchronization overhead.

The documentation says:
> Parallel iterators take care of deciding how to divide your data into tasks; it will dynamically adapt for maximum performance.
https://github.com/rayon-rs/rayon/blob/7c3b97b02/README.md#parallel-iterators-and-more

So this commit does not make each packet a single task, but leave it to rayon to split packets to tasks (There is also an [`IndexedParallelIterator::with_min_len`](https://docs.rs/rayon/1.5.1/rayon/iter/trait.IndexedParallelIterator.html#method.with_min_len) method to limit how small each task can be).

I did some measurement on TDS and it shows 3-4% improvement in elapsed-time per packet: [recv_win.csv](https://github.com/solana-labs/solana/files/6785947/recv_win.csv)
The restart is at line 2600.

behzadnouri:
The performance difference of using nested parallel iterators seems like within the noise and not very conclusive. So I reverted the `.flat_map` part to `.flat_map_iter`, which like before, each nested iterator is sequential and the work is parallelized only between them. 
Any objection to merging this now? I want to test some downstream turbine changes, and would rather this commit gets merged first.

https://docs.rs/rayon/1.5.1/rayon/iter/trait.ParallelIterator.html#flat_map_iter-versus-flat_map
```diff
--- a/core/src/window_service.rs
+++ b/core/src/window_service.rs
@@ -270,8 +270,7 @@ where
     let (shreds, repair_infos): (Vec<_>, Vec<_>) = thread_pool.install(|| {
         packets
             .par_iter_mut()
-            .flat_map(|packet| &mut packet.packets)
-            .filter_map(handle_packet)
+            .flat_map_iter(|packet| packet.packets.iter_mut().filter_map(handle_packet))
             .unzip()
     });
```

sakridge (approved):

PR #18451, opened 2021-07-06
Use tempfile to make unique temps in download utils

dmakarov:
download utils create temporary files to store downloaded data until
all data is downloaded.  These files should be uniquely named to avoid
being overwritten if multiple processes happen to download the same
file simultaneously.

### DIFF
--- a/download-utils/src/lib.rs
+++ b/download-utils/src/lib.rs
@@ -10,6 +10,7 @@ use std::io::Read;
 use std::net::SocketAddr;
 use std::path::{Path, PathBuf};
 use std::time::{Duration, Instant};
+use tempfile::NamedTempFile;
 
 static TRUCK: Emoji = Emoji("🚚 ", "");
 static SPARKLE: Emoji = Emoji("✨ ", "");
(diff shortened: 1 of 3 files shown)


download-utils/src/lib.rs
```diff
@@ -65,17 +66,9 @@ pub fn download_file<'a, 'b>(
...
-    temp_destination_file.set_file_name(format!(
-        "tmp-{}",
-        destination_file
-            .file_name()
-            .expect("file_name")
-            .to_str()
-            .expect("to_str")
-    ));
-
+    let temp_destination_file = NamedTempFile::new()
```
mvines, line 69: Can you append a random suffix to `temp_destination_file` instead.  We have snapshot cleanup code in particular that removes aborted tmp snapshots (`TMP_SNAPSHOT_PREFIX`), which will break with this change.

What's the actual issue you ran into BTW?  Seems weird to be trying to download the same file multiple times

dmakarov: Multiple tests that run cargo build-bpf concurrently when they need to download a new version of bpf-tools.  I thought about using file locks, but it doesn't seem very straightforward to work with in Rust.

dmakarov: I'll close this PR. I need to think how to avoid simultaneous downloads of the same file.

mvines: Maybe use the `#[serial]` attribute for those tests?

dmakarov: Yes, this is probably the best choice. Thank you for the suggestion. I was disappointed Rust test harness wouldn't allow to specify an order of execution of individual test cases or define dependencies between them, but didn't search 3rd party crates for this functionality.

PR #18452, opened 2021-07-06
wip: eliminate accounts index arc and rwlock

jeffwashington:
#### Problem
Prototyping moving to a disk-based accounts index. Working on 1B accounts. The disk-based accounts index model relies on higher level account index locks instead of locks per entry. Much more research to be done.
#### Summary of Changes
Eliminate arc and rwlock on slotlist. This results in more lock contention. But, recently we enabled binning of the accounts index. More bins means less lock contention.
Fixes #

### DIFF
--- a/runtime/src/accounts_index.rs
+++ b/runtime/src/accounts_index.rs
@@ -25,7 +25,7 @@ use std::{
     },
     sync::{
         atomic::{AtomicU64, Ordering},
-        Arc, Mutex, RwLock, RwLockReadGuard, RwLockWriteGuard,
+        Mutex, RwLock, RwLockReadGuard, RwLockWriteGuard,
     },
 };
 use thiserror::Error;
@@ -37,8 +37,6 @@ pub type SlotList<T> = Vec<(Slot, T)>;
 pub type SlotSlice<'s, T> = &'s [(Slot, T)];
 pub type RefCount = u64;
 pub type AccountMap<K, V> = BTreeMap<K, V>;
-
-type AccountMapEntry<T> = Arc<AccountMapEntryInner<T>>;
 
 pub trait IsCached {
     fn is_cached(&self) -> bool;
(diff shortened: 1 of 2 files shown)


jeffwashington:
total_us, insertion_time_us go down. scan_stores_us even goes down. Likely less thread over-scheduling.
```
lemond:
this change with serial insertion (other pr):
generate_index total_us= 7587545i scan_stores_us=10500839i insertion_time_us=19661952i min_bin_size=4565247i max_bin_size=4577150i total_items=73129210i
generate_index total_us=12805024i scan_stores_us=15308801i insertion_time_us=27794907i min_bin_size=4565247i max_bin_size=4577150i total_items=73129210i
master^

64 bins, serial insertion:
generate_index total_us= 5661059i scan_stores_us= 9342401i insertion_time_us=14970163i min_bin_size=1139650i max_bin_size=1146464i total_items=73129210i
```

jeffwashington:
index READ metrics for reading every item in all storages in parallel:
note that insertion_time_us is really 'get' time us. The code and metrics are overloaded for simplicity.
```
this pr:
generate_index total_us=3869973i scan_stores_us=7386426i insertion_time_us=17054601i min_bin_size=4565247i max_bin_size=4577150i total_items=73129210i
generate_index total_us=4916640i scan_stores_us=7522161i insertion_time_us=23235349i min_bin_size=4565247i max_bin_size=4577150i total_items=73129210i
master^

64 bins:
generate_index total_us=3787284i scan_stores_us=7533124i insertion_time_us=16517258i min_bin_size=1139650i max_bin_size=1146464i total_items=73129210i

```

runtime/src/accounts_index.rs
```diff
@@ -1455,15 +1564,11 @@ impl<T: 'static + Clone + IsCached + ZeroLamport + std::marker::Sync + std::mark
...
-            if let Some((mut w_account_entry, account_info)) =
-                self.get_account_write_entry_else_create(pubkey, slot, account_info)
-            {
-                w_account_entry.update(slot, account_info, reclaims);
-                false
-            } else {
-                true
-            }
-        };
+            let w_account_maps = self.get_account_maps_write_lock(pubkey);
```
carllin, line 1567: Primary concerns here are:

1) If a set of hotly updated pubkeys end up in the same bin, this could cause a performance degradation, especially if people are grinding for certain vanity pubkeys.

A good benchmark of this would be to generate a bunch of keys with the same first byte so they end up in the same bin, then run a set of transfers from those keys to another set of keys, and compare the perf to master.

2) RPC lookups (not even scans) for existing accounts like `get_balance` or `get_account` can no longer run concurrently with updates, so a lot of RPC lookups may starve progress here. This would be another good thing to try with an RPC node. I think Stephen has set up a bench for this before to test RPC scan + index contention, which we could repurpose for other RPC queries. 

The metric to look at for contention would be `accounts_db_store_timings::update_index`

PR #18453, opened 2021-07-06
Offload cost_tracking to its own service from banking_stage main

taozhu-chicago:
#### Problem
Currently banking_stage runs cost tracking logic inline in its main threads. This inline-ness may not be necessary, cost can be tracked on separate thread to minimize overhead. 

#### Summary of Changes
- add a `cost_tracking_service` module, running in its own thread;
- banking_stage sends a batch of processed transactions to `cost_tracking_serivce` immediately after transactions committed to bank;

Fixes #

### DIFF
--- a/core/src/banking_stage.rs
+++ b/core/src/banking_stage.rs
@@ -1,7 +1,10 @@
 //! The `banking_stage` processes Transaction messages. It is intended to be used
 //! to contruct a software pipeline. The stage uses all available CPU cores and
 //! can do its processing in parallel with signature verification on the GPU.
-use crate::{cost_tracker::CostTracker, packet_hasher::PacketHasher};
+use crate::{
+    cost_tracker::CostTracker, cost_tracking_service::CommittedTransactionBatch,
+    packet_hasher::PacketHasher,
+};
 use crossbeam_channel::{Receiver as CrossbeamReceiver, RecvTimeoutError};
 use itertools::Itertools;
 use lru::LruCache;
@@ -50,8 +53,11 @@ use std::{
     mem::size_of,
     net::UdpSocket,
     ops::DerefMut,
-    sync::atomic::{AtomicU64, AtomicUsize, Ordering},
-    sync::{Arc, Mutex, RwLock},
+    sync::{
+        atomic::{AtomicU64, AtomicUsize, Ordering},
+        mpsc::Sender,
+        Arc, Mutex, RwLock,
+    },
     thread::{self, Builder, JoinHandle},
     time::Duration,
     time::Instant,
@@ -100,8 +106,6 @@ pub struct BankingStageStats {
     packet_conversion_elapsed: AtomicU64,
     unprocessed_packet_conversion_elapsed: AtomicU64,
     transaction_processing_elapsed: AtomicU64,
-    cost_tracker_update_elapsed: AtomicU64,
-    cost_tracker_clone_elapsed: AtomicU64,
     cost_tracker_check_elapsed: AtomicU64,
 }
 
@@ -224,16 +228,6 @@ impl BankingStageStats {
                     i64
                 ),
                 (
-                    "cost_tracker_update_elapsed",
-                    self.cost_tracker_update_elapsed.swap(0, Ordering::Relaxed) as i64,
-                    i64
-                ),
-                (
-                    "cost_tracker_clone_elapsed",
-                    self.cost_tracker_clone_elapsed.swap(0, Ordering::Relaxed) as i64,
-                    i64
-                ),
-                (
                     "cost_tracker_check_elapsed",
                     self.cost_tracker_check_elapsed.swap(0, Ordering::Relaxed) as i64,
                     i64
@@ -267,6 +261,7 @@ impl BankingStage {
         transaction_status_sender: Option<TransactionStatusSender>,
         gossip_vote_sender: ReplayVoteSender,
         cost_tracker: Arc<RwLock<CostTracker>>,
+        cost_tracking_sender: Sender<CommittedTransactionBatch>,
     ) -> Self {
         Self::new_num_threads(
             cluster_info,
@@ -277,6 +272,7 @@ impl BankingStage {
             transaction_status_sender,
             gossip_vote_sender,
             cost_tracker,
+            cost_tracking_sender,
         )
     }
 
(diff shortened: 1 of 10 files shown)


core/src/cost_tracking_service.rs
```diff
@@ -0,0 +1,185 @@
...
+    fn update(&mut self, cost_tracker_update_count: u64, cost_tracker_update_elapsed: u64) {
+        self.cost_tracker_update_count += cost_tracker_update_count;
+        self.cost_tracker_update_elapsed += cost_tracker_update_elapsed;
+
+        let now = timestamp();
+        let elapsed_ms = now - self.last_print;
+        if elapsed_ms > 1000 {
+            datapoint_info!(
+                "cost-tracking-service-stats",
+                ("total_elapsed_us", elapsed_ms * 1000, i64),
```
carllin, line 38: Is this one necessary? Seems like this is measuring the time between datapoints which

1) Should update roughly every two seconds  based on the check `if elapsed_ms > 1000` ,
2)  The datapoints themselves are already timestamped on arrival to Influx

taozhu-chicago: Oops, a copy/paste from `banking_stage.rs`, will remove

core/src/banking_stage.rs
```diff
@@ -890,6 +897,21 @@ impl BankingStage {
                     tx_results.rent_debits,
                 );
             }
+
+            // track committed transactions' cost
+            let transactions: Vec<_> = batch.transactions_iter().cloned().collect();
+            let execution_results = results.to_vec();
```
carllin, line 903: Can this be offloaded to the separate thread as well?

core/src/cost_tracking_service.rs
```diff
@@ -0,0 +1,185 @@
...
+        // bank
+        for ((result, _), tx) in batch
+            .execution_results
+            .iter()
+            .zip(batch.transactions.iter())
+        {
+            if result.is_err() {
+                continue;
+            }
+            cost_tracker.write().unwrap().add_transaction_cost(tx);
```
carllin, line 132: Not entirely sure of the future flow here, but I'm assuming the BankingStage threads will eventually be checking the cost limit via `CostTracker::try_add()`.

Isn't there the danger that if the check for the block limit is ahead of these updates (say this thread is not scheduled for a while), those banking threads will have an outdated view of the current cost?

carllin: Also this seems prone to races with the cost tracking reset on bank boundaries: https://github.com/solana-labs/solana/blob/6188283ba61015eb1af33fe2e1d5b714f97d6926/core/src/banking_stage.rs#L430-L434

We could receive cost updates from the old bank, reset to a new bank, and then this service applies the old bank's updates on the new bank's cost model for instance. Seems like a identification like the slot of the correct cost model to update should be passed into the channel as well.

core/src/cost_tracking_service.rs
```diff
@@ -0,0 +1,185 @@
+//! this service receives transactions that have been successfully added
+//! to a bank from banking_stage, it therefore calculates and applies the
+//! costs of these transactions to cost_tracker.
+//! This process has some overhead, being in its own service thread is to
+//! minimize impact on main TPU threads.
```
carllin, line 5: hmmm the name of the existing `cost_update_service.rs` and this new `cost_tracking_service.rs` don't seem to distinguish between the functionalities very well.

How about something along the lines of `block_generation_cost_tracking_service`, and `cost_governing_service", taking a page out of the book from Rent and Fees (thanks @t-nelson)

core/src/cost_tracking_service.rs
taozhu-chicago: > Not entirely sure of the future flow here, but I'm assuming the BankingStage threads will eventually be checking the cost limit via `CostTracker::try_add()`.
> 
> Isn't there the danger that if the check for the block limit is ahead of these updates (say this thread is not scheduled for a while), those banking threads will have an outdated view of the current cost?

This is a good observation. I considered it as well when making this change. My thought was this:`add_transaction_cost` to be offline, resulting banking_stage getting out dated result when calling `would_transaction_fit`

core/src/banking_stage.rs
taozhu-chicago: yea, I'd love to but feel sketch of sending ref of `batch` and `results` via channel to another thread. I probably can move `results` to sender since it is the last one used, but not `batch`. Not sure. what is the best practice.

core/src/cost_tracking_service.rs
taozhu-chicago: sold! thanks @carllin and @t-nelson

core/src/cost_tracking_service.rs
taozhu-chicago: in the commit https://github.com/solana-labs/solana/pull/18453/commits/d979e63d96b6cd142918fff08386a7a954264320

carllin: >  however it does not fundamentally impact cost-model function, which is to estimate transaction cost in order to improve block parallelism

I was under the impression that the cost function also served the purpose of setting an upper limit on the cost of a block, and that limit could be severely undermined by bad scheduling of the cost updating service.

Would it make more sense to offload the cost calculation for the transactions to a separate stage before the transactions arrive to BankingStage? Then each transaction would have the cost information paired with it, and BankingStage just has to track the overall block cost with an atomic counter.

taozhu-chicago: This is a valid point, especially if the block limit needs to be adhered accurately. Perhaps trying to offload cost updating is a premature optimization. I am happy to discuss how that limit should be imposed, is it OK to allow a block to exceed the limit sometime so long it pays correct fee? If the block limit is to prevent DOS type attack, and preserve overall cluster health, it might be OK if a block exceeds it a bit?

Instruction's executing cost is updated in real time by replay_stage; Cost calculation is intentionally done close to when it needed to utilize the latest value.

carllin: >  is it OK to allow a block to exceed the limit sometime so long it pays correct fee 

Hmm any scalable fee would have to be based on an accurate representation of the current cost

>If the block limit is to prevent DOS type attack, and preserve overall cluster health, it might be OK if a block exceeds it a bit?

I'm worried "A bit" is unpredictable here, as there is no predictability in thread scheduling.

carllin: It does seem like a good idea to calculate the cost of a transaction up front before executing it though, if the idea is to save unnecessary compute

PR #18469, opened 2021-07-07
serial insertion of bins into accounts index

jeffwashington:
#### Problem
Accounts index generation is slow. Looks like we end up with too few items per bin to insert to be efficiently done in parallel. At 72M accounts, 400k slots, we end up with an average of ~11 accounts per bin per slot.
#### Summary of Changes
Rely on parallelism at storage level instead of within index generation. This can be tuned over time. We could collect multiple storages/slots together to reduce lock grabbing cost. We could chunk multiple bins into a thread and spin up something like 4 or 8 or 16 threads total instead of BINS (currently 16) threads per insert.
Fixes #

### DIFF
--- a/runtime/src/accounts_index.rs
+++ b/runtime/src/accounts_index.rs
@@ -7,7 +7,6 @@ use crate::{
 use bv::BitVec;
 use log::*;
 use ouroboros::self_referencing;
-use rayon::prelude::*;
 use solana_measure::measure::Measure;
 use solana_sdk::{
     clock::{BankId, Slot},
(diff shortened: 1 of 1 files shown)


jeffwashington:
```
lemond:
this pr:
generate_index total_us= 9739623i scan_stores_us=12983168i insertion_time_us=20758566i min_bin_size=4565247i max_bin_size=4577150i total_items=73129210i
generate_index total_us=12805024i scan_stores_us=15308801i insertion_time_us=27794907i min_bin_size=4565247i max_bin_size=4577150i total_items=73129210i
master^

wiggins:
this pr:
generate_index total_us=14764011i scan_stores_us=13823149i insertion_time_us=37207121i min_bin_size=3888540i max_bin_size=3896996i total_items=62272320i
generate_index total_us=16692284i scan_stores_us=17065313i insertion_time_us=51326983i min_bin_size=3888540i max_bin_size=3896996i total_items=62272320i
master^
```

sakridge (approved):

PR #18489, opened 2021-07-07
update ledger tool to restore cost table from blockstore

taozhu-chicago:
#### Problem
Instruction executing costs are persisted in blockstore as CF `cost_table`, ledger-tool's `compute-slot-cost` command could get more accurate results if restore that table from blockstore.

#### Summary of Changes
- initialize cost_model from blockstore::cost_table before computing slot cost. No negative impact if cost-table is empty or doesn't exist in blockstore.

Fixes #

### DIFF
--- a/core/src/cost_model.rs
+++ b/core/src/cost_model.rs
@@ -91,6 +91,29 @@ impl CostModel {
         self.block_cost_limit
     }
 
+    pub fn initialize_cost_table(&mut self, cost_table: &[(Pubkey, u64)]) {
+        for (program_id, cost) in cost_table {
+            match self.upsert_instruction_cost(program_id, cost) {
+                Ok(c) => {
+                    debug!(
+                        "initiating cost table, instruction {:?} has cost {}",
+                        program_id, c
+                    );
+                }
+                Err(err) => {
+                    debug!(
+                        "initiating cost table, failed for instruction {:?}, err: {}",
+                        program_id, err
+                    );
+                }
+            }
+        }
+        debug!(
+            "restored cost model instruction cost table from blockstore, current values: {:?}",
+            self.get_instruction_cost_table()
+        );
+    }
+
     pub fn calculate_cost(&mut self, transaction: &Transaction) -> &TransactionCost {
         self.transaction_cost.reset();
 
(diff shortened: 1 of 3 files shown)


steviez (commented):
Makes sense to do this; however, this code looks almost identical to this function:
https://github.com/solana-labs/solana/blob/master/core/src/validator.rs#L901-L925

Maybe this function could be moved to cost_model.rs, and then could be used by both validator and ledger-tool to avoid duplicate implementations?

[`CostUpdateService::update_cost_model()`](https://github.com/solana-labs/solana/blob/master/core/src/cost_update_service.rs#L136) is similar as well, but given that it takes in an `ExecuteTimings` and not a `Vec<(Pubkey, u64)>` slice, may not be doable to have this one piggyback on a common implementation

taozhu-chicago:
> Makes sense to do this; however, this code looks almost identical to this function:
> https://github.com/solana-labs/solana/blob/master/core/src/validator.rs#L901-L925
> 
> Maybe this function could be moved to cost_model.rs, and then could be used by both validator and ledger-tool to avoid duplicate implementations?
> 
Good point, I actually copied the code from `validator` to `ledger-tool`. Should have it shared at first place :). 
Pushed changes.

taozhu-chicago:
> [`CostUpdateService::update_cost_model()`](https://github.com/solana-labs/solana/blob/master/core/src/cost_update_service.rs#L136) is similar as well, but given that it takes in an `ExecuteTimings` and not a `Vec<(Pubkey, u64)>` slice, may not be doable to have this one piggyback on a common implementation

I'll leave `costUpdateService` alone for now since it is on the other end of process - it updates cost table with executeTiming. Also it is possible it will be updating other info such as account lock time etc.

core/src/cost_model.rs
```diff
@@ -450,4 +473,33 @@ mod tests {
...
+        let cost_model = Arc::new(RwLock::new(cost_model));
+
+        // verify
+        instruction_ids.iter().enumerate().for_each(|(index, id)| {
+            assert_eq!(
+                instruction_costs[index],
+                cost_model.read().unwrap().find_instruction_cost(id)
+            );
+        });
+    }
```
steviez, line 504: Nitpicks
- I think we could simplify the cost table construction, which would in turn simplify the verify step.
- I think we could use `cost_model` as-is without wrapping it into an `Arc<RwLock<CostModel>>`; it looks like the other tests do not wrap the `CostModel` object up either

```suggestion
    fn test_cost_model_init_cost_table() {
        // build cost table
        let cost_table = vec![
            (Pubkey::new_unique(), 10),
            (Pubkey::new_unique(), 20),
            (Pubkey::new_unique(), 30),
        ];
        
        // init cost model
        let mut cost_model = CostModel::default();
        cost_model.initialize_cost_table(&cost_table);

        // verify
        for (id, cost) in cost_table.iter() {
            assert_eq!(*cost, cost_model.find_instruction_cost(id));
        }
    }
```

steviez (dismissed):
Just the (non-functional) suggestions on the new test; otherwise, ship it!

core/src/cost_model.rs
taozhu-chicago: thanks @steviez ... like your suggestions! For #2 regarding wrapping cost_model object, my thought is to mimi actual use case in test, so I wrapped it as it is in client code. It does seem redundant by itself. But #1 makes the test so much cleaner. I'll update the test.

PR #18494, opened 2021-07-07
Fix example in docs for getInflationReward

maxogden:
#### Problem

The example in the docs returns the following error:

```
{"jsonrpc":"2.0","error":{"code":-32602,"message":"Invalid params: invalid type: integer `2`, expected struct RpcEpochConfig."},"id":1}
```

#### Summary of Changes

This simply fixes it to use a working example


mvines:
Thanks!

PR #18504, opened 2021-07-07
Add incremental snapshot utils

brooksprumo:
This commit adds high-level functions (`bank_to_incremental_snapshot()` and `bank_from_incremental_snapshot()`), and additionally
refactors snapshot_utils and serde_snapshot to use a common code path for
snapshots.  New tests were added to snapshot_utils for incremental snapshots.

This PR grabs bits from PR #17875, and incorporated feedback from that PR review. This PR is another in the pursuit of Incremental Snapshots (#17088), and this PR finally/actually adds the ability to both load-from and store-to an incremental snapshot. Yay!

This PR is quite big. A lot of the deserialization code was touched to support incremental snapshots and avoid code duplication. The diff may be easier to read when ignoring whitespace.

The next PR in this series will add new integration-style tests to `core/tests/`, and will address new issues those tests uncover (likely `clean_accounts()`).

#### Additional Bits

Care has been taken to rename (almost) everything around snapshots.
- Wherever specific, all variables and functions will be `full_` or `incremental_` unless they work for both.
- The term "snapshot" was overloaded in the code; sometimes called a bank snapshot, sometimes a slot snapshot, and sometimes referring to the snapshot package/snapshot archive. Outside of Bank, it is now a "bank snapshot".

## Outstanding Questions

- [x] **Naming: Should it be "snapshot" and "incremental snapshot" , or "full snapshot" and "incremental snapshot"** There's quite a bit of existing non-code that uses just "snapshot"; should that be updated too? [More discussion here](https://github.com/solana-labs/solana/pull/18504#discussion_r668915990)
    - resolved by renaming
- [x] **Assert vs error when storages overlap.** [Thread 1 here](https://github.com/solana-labs/solana/pull/18504#discussion_r668281871), [thread 2 here](https://github.com/solana-labs/solana/pull/18504#discussion_r668284384), and [comment 3 here](https://github.com/solana-labs/solana/pull/18504#issuecomment-883501258)
    - resolved by error-ing instead of panic-ing
- [x] **Should I remove `deserialize_snapshot_data_file_capped()`?** [Thread here](https://github.com/solana-labs/solana/pull/18504#discussion_r668267484)
- [x] **Should I create another struct for common parameters?** [Thread here](https://github.com/solana-labs/solana/pull/18504#discussion_r668294748)

## New Issues Created

- #18639
- #18641
- #18813 
- #18815 
- #18841

### DIFF
--- a/ledger/src/blockstore_processor.rs
+++ b/ledger/src/blockstore_processor.rs
@@ -567,7 +567,16 @@ fn do_process_blockstore_from_root(
         ("slot", bank_forks.root(), i64),
         ("forks", initial_forks.len(), i64),
         ("calculate_capitalization_us", time_cap.as_us(), i64),
-        ("untar_us", timings.untar_us, i64),
+        (
+            "full_snapshot_untar_us",
+            timings.full_snapshot_untar_us,
+            i64
+        ),
+        (
+            "incremental_snapshot_untar_us",
+            timings.incremental_snapshot_untar_us,
+            i64
+        ),
         (
             "rebuild_bank_from_snapshots_us",
             timings.rebuild_bank_from_snapshots_us,
--- a/runtime/src/serde_snapshot.rs
+++ b/runtime/src/serde_snapshot.rs
@@ -74,6 +74,62 @@ struct AccountsDbFields<T>(
     BankHashInfo,
 );
 
+/// Helper type to wrap BufReader streams when deserializing and reconstructing from either just a
+/// full snapshot, or both a full and incremental snapshot
+pub struct SnapshotStreams<'a, R> {
+    pub full_snapshot_stream: &'a mut BufReader<R>,
+    pub incremental_snapshot_stream: Option<&'a mut BufReader<R>>,
+}
+
+/// Helper type to wrap AccountsDbFields when reconstructing AccountsDb from either just a full
+/// snapshot, or both a full and incremental snapshot
+#[derive(Debug)]
+struct SnapshotAccountsDbFields<T> {
+    full_snapshot_accounts_db_fields: AccountsDbFields<T>,
+    incremental_snapshot_accounts_db_fields: Option<AccountsDbFields<T>>,
+}
+
+impl<T> SnapshotAccountsDbFields<T> {
+    /// Collapse the SnapshotAccountsDbFields into a single AccountsDbFields.  If there is no
+    /// incremental snapshot, this returns the AccountsDbFields from the full snapshot.  Otherwise
+    /// this uses the version, slot, and bank hash info from the incremental snapshot, then the
+    /// combination of the storages from both the full and incremental snapshots.
+    fn collapse_into(self) -> Result<AccountsDbFields<T>, Error> {
+        match self.incremental_snapshot_accounts_db_fields {
+            None => Ok(self.full_snapshot_accounts_db_fields),
+            Some(AccountsDbFields(
+                mut incremental_snapshot_storages,
+                incremental_snapshot_version,
+                incremental_snapshot_slot,
+                incremental_snapshot_bank_hash_info,
+            )) => {
+                let full_snapshot_storages = self.full_snapshot_accounts_db_fields.0;
+                let full_snapshot_slot = self.full_snapshot_accounts_db_fields.2;
+
+                // filter out incremental snapshot storages with slot <= full snapshot slot
+                incremental_snapshot_storages.retain(|slot, _| *slot > full_snapshot_slot);
+
+                // There must not be any overlap in the slots of storages between the full snapshot and the incremental snapshot
+                incremental_snapshot_storages
+                    .iter()
+                    .all(|storage_entry| !full_snapshot_storages.contains_key(storage_entry.0)).then(|| ()).ok_or_else(|| {
+                        io::Error::new(io::ErrorKind::InvalidData, "Snapshots are incompatible: There are storages for the same slot in both the full snapshot and the incremental snapshot!")
+                    })?;
+
+                let mut combined_storages = full_snapshot_storages;
+                combined_storages.extend(incremental_snapshot_storages.into_iter());
+
+                Ok(AccountsDbFields(
+                    combined_storages,
+                    incremental_snapshot_version,
+                    incremental_snapshot_slot,
+                    incremental_snapshot_bank_hash_info,
+                ))
+            }
+        }
+    }
+}
+
 trait TypeContext<'a> {
     type SerializableAccountStorageEntry: Serialize
         + DeserializeOwned
@@ -127,16 +183,16 @@ where
 }
 
 #[allow(clippy::too_many_arguments)]
-pub(crate) fn bank_from_stream<R>(
+pub(crate) fn bank_from_streams<R>(
     serde_style: SerdeStyle,
-    stream: &mut BufReader<R>,
+    snapshot_streams: &mut SnapshotStreams<R>,
     account_paths: &[PathBuf],
     unpacked_append_vec_map: UnpackedAppendVecMap,
     genesis_config: &GenesisConfig,
     frozen_account_pubkeys: &[Pubkey],
     debug_keys: Option<Arc<HashSet<Pubkey>>>,
     additional_builtins: Option<&Builtins>,
-    account_indexes: AccountSecondaryIndexes,
+    account_secondary_indexes: AccountSecondaryIndexes,
     caching_enabled: bool,
     limit_load_slot_count_from_snapshot: Option<usize>,
     shrink_ratio: AccountShrinkThreshold,
@@ -147,18 +203,33 @@ where
 {
     macro_rules! INTO {
         ($x:ident) => {{
-            let (bank_fields, accounts_db_fields) = $x::deserialize_bank_fields(stream)?;
+            let (full_snapshot_bank_fields, full_snapshot_accounts_db_fields) =
+                $x::deserialize_bank_fields(snapshot_streams.full_snapshot_stream)?;
+            let (incremental_snapshot_bank_fields, incremental_snapshot_accounts_db_fields) =
+                if let Some(ref mut incremental_snapshot_stream) =
+                    snapshot_streams.incremental_snapshot_stream
+                {
+                    let (bank_fields, accounts_db_fields) =
+                        $x::deserialize_bank_fields(incremental_snapshot_stream)?;
+                    (Some(bank_fields), Some(accounts_db_fields))
+                } else {
+                    (None, None)
+                };
 
+            let snapshot_accounts_db_fields = SnapshotAccountsDbFields {
+                full_snapshot_accounts_db_fields,
+                incremental_snapshot_accounts_db_fields,
+            };
             let bank = reconstruct_bank_from_fields(
-                bank_fields,
-                accounts_db_fields,
+                incremental_snapshot_bank_fields.unwrap_or(full_snapshot_bank_fields),
+                snapshot_accounts_db_fields,
                 genesis_config,
                 frozen_account_pubkeys,
                 account_paths,
                 unpacked_append_vec_map,
                 debug_keys,
                 additional_builtins,
-                account_indexes,
+                account_secondary_indexes,
                 caching_enabled,
                 limit_load_slot_count_from_snapshot,
                 shrink_ratio,
@@ -243,14 +314,14 @@ impl<'a, C> IgnoreAsHelper for SerializableAccountsDb<'a, C> {}
 #[allow(clippy::too_many_arguments)]
 fn reconstruct_bank_from_fields<E>(
     bank_fields: BankFieldsToDeserialize,
-    accounts_db_fields: AccountsDbFields<E>,
+    snapshot_accounts_db_fields: SnapshotAccountsDbFields<E>,
     genesis_config: &GenesisConfig,
     frozen_account_pubkeys: &[Pubkey],
     account_paths: &[PathBuf],
     unpacked_append_vec_map: UnpackedAppendVecMap,
     debug_keys: Option<Arc<HashSet<Pubkey>>>,
     additional_builtins: Option<&Builtins>,
-    account_indexes: AccountSecondaryIndexes,
+    account_secondary_indexes: AccountSecondaryIndexes,
     caching_enabled: bool,
     limit_load_slot_count_from_snapshot: Option<usize>,
     shrink_ratio: AccountShrinkThreshold,
@@ -260,11 +331,11 @@ where
     E: SerializableStorage + std::marker::Sync,
 {
     let mut accounts_db = reconstruct_accountsdb_from_fields(
-        accounts_db_fields,
+        snapshot_accounts_db_fields,
         account_paths,
         unpacked_append_vec_map,
         &genesis_config.cluster_type,
-        account_indexes,
+        account_secondary_indexes,
         caching_enabled,
         limit_load_slot_count_from_snapshot,
         shrink_ratio,
@@ -310,11 +381,11 @@ where
 }
 
 fn reconstruct_accountsdb_from_fields<E>(
-    accounts_db_fields: AccountsDbFields<E>,
+    snapshot_accounts_db_fields: SnapshotAccountsDbFields<E>,
     account_paths: &[PathBuf],
     unpacked_append_vec_map: UnpackedAppendVecMap,
     cluster_type: &ClusterType,
-    account_indexes: AccountSecondaryIndexes,
+    account_secondary_indexes: AccountSecondaryIndexes,
     caching_enabled: bool,
     limit_load_slot_count_from_snapshot: Option<usize>,
     shrink_ratio: AccountShrinkThreshold,
@@ -326,11 +397,19 @@ where
     let mut accounts_db = AccountsDb::new_with_config(
         account_paths.to_vec(),
         cluster_type,
-        account_indexes,
+        account_secondary_indexes,
         caching_enabled,
         shrink_ratio,
     );
-    let AccountsDbFields(storage, version, slot, bank_hash_info) = accounts_db_fields;
+
+    let AccountsDbFields(
+        snapshot_storages,
+        snapshot_version,
+        snapshot_slot,
+        snapshot_bank_hash_info,
+    ) = snapshot_accounts_db_fields.collapse_into()?;
+
+    let snapshot_storages = snapshot_storages.into_iter().collect::<Vec<_>>();
 
     // Ensure all account paths exist
     for path in &accounts_db.paths {
@@ -338,13 +417,11 @@ where
             .unwrap_or_else(|err| panic!("Failed to create directory {}: {}", path.display(), err));
     }
 
-    let storage = storage.into_iter().collect::<Vec<_>>();
-
     // Remap the deserialized AppendVec paths to point to correct local paths
-    let mut storage = (0..storage.len())
+    let mut storage = (0..snapshot_storages.len())
         .into_par_iter()
         .map(|i| {
-            let (slot, slot_storage) = &storage[i];
+            let (slot, slot_storage) = &snapshot_storages[i];
             let mut new_slot_storage = HashMap::new();
             for storage_entry in slot_storage {
                 let file_name = AppendVec::file_name(*slot, storage_entry.id());
@@ -376,7 +453,7 @@ where
         .bank_hashes
         .write()
         .unwrap()
-        .insert(slot, bank_hash_info);
+        .insert(snapshot_slot, snapshot_bank_hash_info);
 
     // Process deserialized data, set necessary fields in self
     let max_id: usize = *storage
@@ -400,7 +477,7 @@ where
     accounts_db.next_id.store(max_id + 1, Ordering::Relaxed);
     accounts_db
         .write_version
-        .fetch_add(version, Ordering::Relaxed);
+        .fetch_add(snapshot_version, Ordering::Relaxed);
     accounts_db.generate_index(limit_load_slot_count_from_snapshot, verify_index);
     Ok(accounts_db)
 }
--- a/runtime/src/snapshot_utils.rs
+++ b/runtime/src/snapshot_utils.rs
@@ -1130,8 +1130,12 @@ where
     info!("{}", measure_untar);
 
     let unpacked_version_file = unpack_dir.path().join("version");
-    let mut snapshot_version = String::new();
-    File::open(unpacked_version_file).and_then(|mut f| f.read_to_string(&mut snapshot_version))?;
+    let snapshot_version = {
+        let mut snapshot_version = String::new();
+        File::open(unpacked_version_file)
+            .and_then(|mut f| f.read_to_string(&mut snapshot_version))?;
+        snapshot_version.trim().to_string()
+    };
 
     Ok(UnarchivedSnapshot {
         unpack_dir,
@@ -1517,16 +1521,19 @@ fn untar_snapshot_in<P: AsRef<Path>>(
 fn verify_unpacked_snapshots_dir_and_version(
     unpacked_snapshots_dir_and_version: &UnpackedSnapshotsDirAndVersion,
 ) -> Result<(SnapshotVersion, BankSnapshotInfo)> {
-    let snapshot_version_str = unpacked_snapshots_dir_and_version.snapshot_version.trim();
-    info!("snapshot version: {}", snapshot_version_str);
+    info!(
+        "snapshot version: {}",
+        &unpacked_snapshots_dir_and_version.snapshot_version
+    );
 
     let snapshot_version =
-        SnapshotVersion::maybe_from_string(snapshot_version_str).ok_or_else(|| {
-            get_io_error(&format!(
-                "unsupported snapshot version: {}",
-                snapshot_version_str
-            ))
-        })?;
+        SnapshotVersion::maybe_from_string(&unpacked_snapshots_dir_and_version.snapshot_version)
+            .ok_or_else(|| {
+                get_io_error(&format!(
+                    "unsupported snapshot version: {}",
+                    &unpacked_snapshots_dir_and_version.snapshot_version,
+                ))
+            })?;
     let mut bank_snapshot_infos =
         get_bank_snapshots(&unpacked_snapshots_dir_and_version.unpacked_snapshots_dir);
     if bank_snapshot_infos.len() > 1 {
--- a/runtime/src/snapshot_utils.rs
+++ b/runtime/src/snapshot_utils.rs
@@ -5,7 +5,8 @@ use {
         bank::{Bank, BankSlotDelta, Builtins},
         hardened_unpack::{unpack_snapshot, ParallelSelector, UnpackError, UnpackedAppendVecMap},
         serde_snapshot::{
-            bank_from_stream, bank_to_stream, SerdeStyle, SnapshotStorage, SnapshotStorages,
+            bank_from_streams, bank_to_stream, SerdeStyle, SnapshotStorage, SnapshotStorages,
+            SnapshotStreams,
         },
         shared_buffer_reader::{SharedBuffer, SharedBufferReader},
         snapshot_package::{
@@ -16,54 +17,170 @@ use {
     bincode::{config::Options, serialize_into},
     bzip2::bufread::BzDecoder,
     flate2::read::GzDecoder,
+    lazy_static::lazy_static,
     log::*,
     rayon::{prelude::*, ThreadPool},
     regex::Regex,
     solana_measure::measure::Measure,
     solana_sdk::{clock::Slot, genesis_config::GenesisConfig, hash::Hash, pubkey::Pubkey},
     std::{
-        cmp::max,
-        cmp::Ordering,
+        cmp::{max, Ordering},
         collections::HashSet,
         fmt,
         fs::{self, File},
-        io::{
-            self, BufReader, BufWriter, Error as IoError, ErrorKind, Read, Seek, SeekFrom, Write,
-        },
+        io::{self, BufReader, BufWriter, Error as IoError, ErrorKind, Read, Seek, Write},
         path::{Path, PathBuf},
         process::{self, ExitStatus},
         str::FromStr,
         sync::Arc,
     },
     tar::Archive,
+    tempfile::TempDir,
     thiserror::Error,
 };
 
-/// Information about a snapshot archive: its path, slot, hash, and archive format
-pub struct SnapshotArchiveInfo {
+/// Common information about a snapshot archive
+#[derive(PartialEq, Eq, Debug)]
+struct SnapshotArchiveInfo {
     /// Path to the snapshot archive file
-    pub path: PathBuf,
+    path: PathBuf,
 
     /// Slot that the snapshot was made
-    pub slot: Slot,
+    slot: Slot,
 
     /// Hash of the accounts at this slot
-    pub hash: Hash,
+    hash: Hash,
 
     /// Archive format for the snapshot file
-    pub archive_format: ArchiveFormat,
+    archive_format: ArchiveFormat,
 }
+
+/// Information about a full snapshot archive: its path, slot, hash, and archive format
+#[derive(PartialEq, Eq, Debug)]
+pub struct FullSnapshotArchiveInfo(SnapshotArchiveInfo);
+
+impl FullSnapshotArchiveInfo {
+    /// Parse the path to a full snapshot archive and return a new `FullSnapshotArchiveInfo`
+    fn new_from_path(path: PathBuf) -> Result<Self> {
+        let filename = path_to_file_name_str(path.as_path())?;
+        let (slot, hash, archive_format) = parse_full_snapshot_archive_filename(filename)?;
+
+        Ok(Self(SnapshotArchiveInfo {
+            path,
+            slot,
+            hash,
+            archive_format,
+        }))
+    }
+
+    pub fn path(&self) -> &PathBuf {
+        &self.0.path
+    }
+
+    pub fn slot(&self) -> &Slot {
+        &self.0.slot
+    }
+
+    pub fn hash(&self) -> &Hash {
+        &self.0.hash
+    }
+
+    pub fn archive_format(&self) -> &ArchiveFormat {
+        &self.0.archive_format
+    }
+}
+
+impl PartialOrd for FullSnapshotArchiveInfo {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+// Order `FullSnapshotArchiveInfo` by slot (ascending), which practially is sorting chronologically
+impl Ord for FullSnapshotArchiveInfo {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.slot().cmp(other.slot())
+    }
+}
+
+/// Information about an incremental snapshot archive: its path, slot, base slot, hash, and archive format
+#[derive(PartialEq, Eq, Debug)]
+pub struct IncrementalSnapshotArchiveInfo {
+    /// The slot that the incremental snapshot was based from.  This is the same as the full
+    /// snapshot slot used when making the incremental snapshot.
+    base_slot: Slot,
+
+    /// Use the `SnapshotArchiveInfo` struct for the common fields: path, slot, hash, and
+    /// archive_format, but as they pertain to the incremental snapshot.
+    inner: SnapshotArchiveInfo,
+}
+
+impl IncrementalSnapshotArchiveInfo {
+    /// Parse the path to an incremental snapshot archive and return a new `IncrementalSnapshotArchiveInfo`
+    fn new_from_path(path: PathBuf) -> Result<Self> {
+        let filename = path_to_file_name_str(path.as_path())?;
+        let (base_slot, slot, hash, archive_format) =
+            parse_incremental_snapshot_archive_filename(filename)?;
+
+        Ok(Self {
+            base_slot,
+            inner: SnapshotArchiveInfo {
+                path,
+                slot,
+                hash,
+                archive_format,
+            },
+        })
+    }
+
+    fn path(&self) -> &PathBuf {
+        &self.inner.path
+    }
+
+    fn base_slot(&self) -> &Slot {
+        &self.base_slot
+    }
+
+    fn slot(&self) -> &Slot {
+        &self.inner.slot
+    }
+
+    fn _hash(&self) -> &Hash {
+        &self.inner.hash
+    }
+
+    fn _archive_format(&self) -> &ArchiveFormat {
+        &self.inner.archive_format
+    }
+}
+
+impl PartialOrd for IncrementalSnapshotArchiveInfo {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+// Order `IncrementalSnapshotArchiveInfo` by base slot (ascending), then slot (ascending), which
+// practially is sorting chronologically
+impl Ord for IncrementalSnapshotArchiveInfo {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.base_slot()
+            .cmp(other.base_slot())
+            .then(self.slot().cmp(other.slot()))
+    }
+}
+
 pub const SNAPSHOT_STATUS_CACHE_FILE_NAME: &str = "status_cache";
 
-pub const MAX_SNAPSHOTS: usize = 8; // Save some snapshots but not too many
+pub const MAX_BANK_SNAPSHOTS: usize = 8; // Save some snapshots but not too many
 const MAX_SNAPSHOT_DATA_FILE_SIZE: u64 = 32 * 1024 * 1024 * 1024; // 32 GiB
 const VERSION_STRING_V1_2_0: &str = "1.2.0";
 const DEFAULT_SNAPSHOT_VERSION: SnapshotVersion = SnapshotVersion::V1_2_0;
-const TMP_SNAPSHOT_PREFIX: &str = "tmp-snapshot-";
-pub const DEFAULT_MAX_SNAPSHOTS_TO_RETAIN: usize = 2;
-
-pub const SNAPSHOT_ARCHIVE_FILENAME_REGEX: &str =
-    r"^snapshot-(\d+)-([[:alnum:]]+)\.(tar|tar\.bz2|tar\.zst|tar\.gz)$";
+const TMP_FULL_SNAPSHOT_PREFIX: &str = "tmp-snapshot-";
+const TMP_INCREMENTAL_SNAPSHOT_PREFIX: &str = "tmp-incremental-snapshot-";
+pub const DEFAULT_MAX_FULL_SNAPSHOT_ARCHIVES_TO_RETAIN: usize = 2;
+pub const FULL_SNAPSHOT_ARCHIVE_FILENAME_REGEX: &str = r"^snapshot-(?P<slot>[[:digit:]]+)-(?P<hash>[[:alnum:]]+)\.(?P<ext>tar|tar\.bz2|tar\.zst|tar\.gz)$";
+pub const INCREMENTAL_SNAPSHOT_ARCHIVE_FILENAME_REGEX: &str = r"^incremental-snapshot-(?P<base>[[:digit:]]+)-(?P<slot>[[:digit:]]+)-(?P<hash>[[:alnum:]]+)\.(?P<ext>tar|tar\.bz2|tar\.zst|tar\.gz)$";
 
 #[derive(Copy, Clone, Eq, PartialEq, Debug)]
 pub enum SnapshotVersion {
@@ -129,10 +246,48 @@ pub enum ArchiveFormat {
     Tar,
 }
 
+/// A slot and the path to its bank snapshot
 #[derive(PartialEq, Eq, Debug)]
-pub struct SlotSnapshotPaths {
+pub struct BankSnapshotInfo {
     pub slot: Slot,
-    pub snapshot_file_path: PathBuf,
+    pub snapshot_path: PathBuf,
+}
+
+impl PartialOrd for BankSnapshotInfo {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+// Order BankSnapshotInfo by slot (ascending), which practially is sorting chronologically
+impl Ord for BankSnapshotInfo {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.slot.cmp(&other.slot)
+    }
+}
+
+/// Helper type when rebuilding from snapshots.  Designed to handle when rebuilding from just a
+/// full snapshot, or from both a full snapshot and an incremental snapshot.
+#[derive(Debug)]
+struct SnapshotRootPaths {
+    full_snapshot_root_file_path: PathBuf,
+    incremental_snapshot_root_file_path: Option<PathBuf>,
+}
+
+/// Helper type to bundle up the results from `unarchive_snapshot()`
+#[derive(Debug)]
+struct UnarchivedSnapshot {
+    unpack_dir: TempDir,
+    unpacked_append_vec_map: UnpackedAppendVecMap,
+    unpacked_snapshots_dir_and_version: UnpackedSnapshotsDirAndVersion,
+    measure_untar: Measure,
+}
+
+/// Helper type for passing around the unpacked snapshots dir and the snapshot version together
+#[derive(Debug)]
+struct UnpackedSnapshotsDirAndVersion {
+    unpacked_snapshots_dir: PathBuf,
+    snapshot_version: String,
 }
 
 #[derive(Error, Debug)]
@@ -157,53 +312,139 @@ pub enum SnapshotError {
 
     #[error("source({1}) - I/O error: {0}")]
     IoWithSource(std::io::Error, &'static str),
+
+    #[error("could not get file name from path: {}", .0.display())]
+    PathToFileNameError(PathBuf),
+
+    #[error("could not get str from file name: {}", .0.display())]
+    FileNameToStrError(PathBuf),
+
+    #[error("could not parse snapshot archive's file name: {0}")]
+    ParseSnapshotArchiveFileNameError(String),
+
+    #[error("snapshots are incompatible: full snapshot slot ({0}) and incremental snapshot base slot ({1}) do not match")]
+    MismatchedBaseSlot(Slot, Slot),
 }
 pub type Result<T> = std::result::Result<T, SnapshotError>;
 
-impl PartialOrd for SlotSnapshotPaths {
-    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
-        Some(self.slot.cmp(&other.slot))
-    }
-}
-
-impl Ord for SlotSnapshotPaths {
-    fn cmp(&self, other: &Self) -> Ordering {
-        self.slot.cmp(&other.slot)
-    }
-}
-
-pub fn package_snapshot<P: AsRef<Path>, Q: AsRef<Path>>(
+/// Package up bank snapshot files, snapshot storages, and slot deltas for a full snapshot.
+pub fn package_full_snapshot<P, Q>(
     bank: &Bank,
-    snapshot_files: &SlotSnapshotPaths,
-    snapshot_path: Q,
+    bank_snapshot_info: &BankSnapshotInfo,
+    snapshots_dir: P,
+    status_cache_slot_deltas: Vec<BankSlotDelta>,
+    snapshot_package_output_path: Q,
+    snapshot_storages: SnapshotStorages,
+    archive_format: ArchiveFormat,
+    snapshot_version: SnapshotVersion,
+    hash_for_testing: Option<Hash>,
+) -> Result<AccountsPackagePre>
+where
+    P: AsRef<Path>,
+    Q: AsRef<Path>,
+{
+    info!(
+        "Package full snapshot for bank: {} has {} account storage entries",
+        bank.slot(),
+        snapshot_storages.len()
+    );
+
+    let snapshot_tmpdir = tempfile::Builder::new()
+        .prefix(&format!("{}{}-", TMP_FULL_SNAPSHOT_PREFIX, bank.slot()))
+        .tempdir_in(snapshots_dir)?;
+
+    do_package_snapshot(
+        bank,
+        bank_snapshot_info,
+        status_cache_slot_deltas,
+        snapshot_package_output_path,
+        snapshot_storages,
+        archive_format,
+        snapshot_version,
+        hash_for_testing,
+        snapshot_tmpdir,
+    )
+}
+
+/// Package up bank snapshot files, snapshot storages, and slot deltas for an incremental snapshot.
+#[allow(clippy::too_many_arguments)]
+fn package_incremental_snapshot<P, Q>(
+    bank: &Bank,
+    incremental_snapshot_base_slot: Slot,
+    bank_snapshot_info: &BankSnapshotInfo,
+    snapshots_dir: P,
+    status_cache_slot_deltas: Vec<BankSlotDelta>,
+    snapshot_package_output_path: Q,
+    snapshot_storages: SnapshotStorages,
+    archive_format: ArchiveFormat,
+    snapshot_version: SnapshotVersion,
+    hash_for_testing: Option<Hash>,
+) -> Result<AccountsPackagePre>
+where
+    P: AsRef<Path>,
+    Q: AsRef<Path>,
+{
+    info!(
+        "Package incremental snapshot for bank {} (from base slot {}) has {} account storage entries",
+        bank.slot(),
+        incremental_snapshot_base_slot,
+        snapshot_storages.len()
+    );
+
+    assert!(
+        snapshot_storages.iter().all(|storage| storage
+            .iter()
+            .all(|entry| entry.slot() > incremental_snapshot_base_slot)),
+        "Incremental snapshot package must only contain storage entries where slot > incremental snapshot base slot (i.e. full snapshot slot)!"
+    );
+
+    let snapshot_tmpdir = tempfile::Builder::new()
+        .prefix(&format!(
+            "{}{}-{}-",
+            TMP_INCREMENTAL_SNAPSHOT_PREFIX,
+            incremental_snapshot_base_slot,
+            bank.slot()
+        ))
+        .tempdir_in(snapshots_dir)?;
+
+    do_package_snapshot(
+        bank,
+        bank_snapshot_info,
+        status_cache_slot_deltas,
+        snapshot_package_output_path,
+        snapshot_storages,
+        archive_format,
+        snapshot_version,
+        hash_for_testing,
+        snapshot_tmpdir,
+    )
+}
+
+fn do_package_snapshot<P>(
+    bank: &Bank,
+    bank_snapshot_info: &BankSnapshotInfo,
     status_cache_slot_deltas: Vec<BankSlotDelta>,
     snapshot_package_output_path: P,
     snapshot_storages: SnapshotStorages,
     archive_format: ArchiveFormat,
     snapshot_version: SnapshotVersion,
     hash_for_testing: Option<Hash>,
-) -> Result<AccountsPackagePre> {
-    // Hard link all the snapshots we need for this package
-    let snapshot_tmpdir = tempfile::Builder::new()
-        .prefix(&format!("{}{}-", TMP_SNAPSHOT_PREFIX, bank.slot()))
-        .tempdir_in(snapshot_path)?;
-
+    snapshot_tmpdir: TempDir,
+) -> Result<AccountsPackagePre>
+where
+    P: AsRef<Path>,
+{
     // Create a snapshot package
-    info!(
-        "Snapshot for bank: {} has {} account storage entries",
-        bank.slot(),
-        snapshot_storages.len()
-    );
 
     // Hard link the snapshot into a tmpdir, to ensure its not removed prior to packaging.
     {
         let snapshot_hardlink_dir = snapshot_tmpdir
             .as_ref()
-            .join(snapshot_files.slot.to_string());
+            .join(bank_snapshot_info.slot.to_string());
         fs::create_dir_all(&snapshot_hardlink_dir)?;
         fs::hard_link(
-            &snapshot_files.snapshot_file_path,
-            &snapshot_hardlink_dir.join(snapshot_files.slot.to_string()),
+            &bank_snapshot_info.snapshot_path,
+            &snapshot_hardlink_dir.join(bank_snapshot_info.slot.to_string()),
         )?;
     }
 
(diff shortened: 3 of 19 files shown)


runtime/src/serde_snapshot.rs
```diff
@@ -126,6 +141,7 @@ where
         .deserialize_from::<R, T>(reader)
 }
 
+#[cfg(test)]
 #[allow(clippy::too_many_arguments)]
 pub(crate) fn bank_from_stream<R>(
```
brooksprumo, line 146: This function is now only used in serde_snapshots/tests.rs. I could remove it and update the test, but figured it was fine to leave it. Let me know if it should be removed instead.

runtime/src/snapshot_utils.rs
```diff
@@ -814,23 +1387,66 @@ where
...
+        );
         fs::remove_file(old_archive.path)
-            .unwrap_or_else(|err| info!("Failed to remove old snapshot: {:}", err));
+            .unwrap_or_else(|err| info!("Failed to remove old snapshot: {}", err));
     }
+
+    // Only keep incremental snapshots for the latest full snapshot
+    // bprumo TODO: As an option to further reduce the number of incremental snapshots, only a
+    // subset of the incremental snapshots for the lastest full snapshot could be kept.  This could
+    // reuse maximum_snapshots_to_retain, or use a new field just for incremental snapshots.
```
brooksprumo, line 1437: I plan to address this TODO in a subsequent PR when adding integration tests for incremental snapshots.

runtime/src/serde_snapshot.rs
```diff
@@ -146,18 +200,33 @@ where
...
+            let (incremental_snapshot_bank_fields, incremental_snapshot_accounts_db_fields) =
+                if let Some(ref mut incremental_snapshot_stream) =
+                    snapshot_streams.incremental_snapshot_stream
+                {
+                    let (bank_fields, accounts_db_fields) =
+                        $x::deserialize_bank_fields(incremental_snapshot_stream)?;
+                    (Some(bank_fields), Some(accounts_db_fields))
+                } else {
+                    (None, None)
+                };
```
brooksprumo, line 214: I use this pattern a number of times to handle `Option<>` on incremental snapshot stuffs. While this way works, it feels very redundant and verbose. Is there a better/clearer way to do this?

runtime/src/serde_snapshot.rs
carllin: yeah looks like this is just a convenience function around passing a `SnapshotStreams` at this point. I think this can be removed and the functionality just moved inside the one test

runtime/src/serde_snapshot.rs
carllin: This seems to make sense to me, the `?` in `$x::deserialize_bank_fields(incremental_snapshot_stream)?;` makes it hard to use combinators

runtime/src/serde_snapshot.rs
```diff
@@ -321,19 +390,54 @@ where
...
+        .as_ref()
+        .map(|fields| fields.1);
+    let incremental_snapshot_slot = snapshot_accounts_db_fields
+        .incremental_snapshot_accounts_db_fields
+        .as_ref()
+        .map(|fields| fields.2);
+    let incremental_snapshot_bank_hash_info = snapshot_accounts_db_fields
+        .incremental_snapshot_accounts_db_fields
+        .as_ref()
+        .map(|fields| fields.3.clone());
```
carllin, line 415: How about 

```
let (snapshot_version, snapshot_slot, snapshot_bank_hash_info) = snapshot_accounts_db_fields
        .incremental_snapshot_accounts_db_fields
        .as_ref()
        .map(
            |AccountsDbFields(
                _,
                incremental_snapshot_version,
                incremental_snapshot_slot,
                incremental_snapshot_bank_hash_info,
            )| {
                (
                    *incremental_snapshot_version,
                    *incremental_snapshot_slot,
                    incremental_snapshot_bank_hash_info.clone(),
                )
            },
        )
        .unwrap_or((
            full_snapshot_version,
            full_snapshot_slot,
            full_snapshot_bank_hash_info,
        ));
```

and removing the `unwrap_or` for each of the references below

runtime/src/serde_snapshot.rs
```diff
@@ -367,11 +471,10 @@ where
...
-        .write()
-        .unwrap()
-        .insert(slot, bank_hash_info);
+    accounts_db.bank_hashes.write().unwrap().insert(
+        incremental_snapshot_slot.unwrap_or(full_snapshot_slot),
+        incremental_snapshot_bank_hash_info.unwrap_or(full_snapshot_bank_hash_info),
+    );
 
     // Process deserialized data, set necessary fields in self
     let max_id: usize = *storage
```
carllin, line 480: Is it guaranteed that between the full snapshot generated on one node, an the incremental snapshot generated/fetched from another node, that *none* the storage entries will have overlapping ids?

runtime/src/serde_snapshot.rs
brooksprumo: OK will do!

runtime/src/serde_snapshot.rs
brooksprumo: Yeah... I was hoping there was some other Rust magic out there :)

runtime/src/serde_snapshot.rs
brooksprumo: Right, yeah, I remember your comment on the other PR about the AppendVec IDs. I would assume there's no guarantee that different nodes will have the same IDs, therefore there may be overlapping IDs. I think there are a few options for this PR:

1. Re-map duplicate IDs before reconstructing the storage map.
2. Leave the implementation as-is, and then address the overlapping IDs in a later PR.
3. Prevent duplicate IDs by requiring incremental snapshots come from the same node that provides the full snapshot.

Originally I was thinking (3) would be the way to go. Maybe this is too restrictive though? How often does a node download a full snapshot? I would guess not very often, right? After a node has joined, it creates its own full snapshots and incremental snapshots for future restarts. And I think on the gossip/RPC side this is the plan; i.e. when a node asks for a snapshot and downloads it, it will be given a full snapshot and an incremental snapshot.

For (2), I think this is also my vote, regardless of (3). I tried to make this PR as small as possible, and just have the necessary bits to create and rebuild from incremental snapshots. So then (1) could be added later as necessary, when it becomes a requirement to enable getting full snapshots and incremental snapshots from different nodes.

Is that OK for this PR?

runtime/src/serde_snapshot.rs
brooksprumo: Love it! Done.

runtime/src/serde_snapshot.rs
brooksprumo: Done!

runtime/src/serde_snapshot.rs
carllin: > Maybe this is too restrictive though?

Agreed, I think this is too restrictive.

> For (2), I think this is also my vote, regardless of (3). I tried to make this PR as small as possible, and just have the necessary bits to create and rebuild from incremental snapshots. So then (1) could be added later as necessary, when it becomes a requirement to enable getting full snapshots and incremental snapshots from different nodes.

Yeah that makes sense, as long as that code makes it in before the integration with gossip it should be fine, maybe an issue to track it?

runtime/src/snapshot_utils.rs
```diff
@@ -53,18 +55,42 @@ pub struct SnapshotArchiveInfo {
...
+    pub base_slot: Slot,
+
+    /// Slot that the incremental snapshot was made
+    pub slot: Slot,
+
+    /// Hash of the accounts at this slot
+    pub hash: Hash,
+
+    /// Archive format for the incremental snapshot file
+    pub archive_format: ArchiveFormat,
```
carllin, line 75: does it make sense to just have `IncrementalSnapshotArchiveInfo` hold a `SnapshotArchiveInfo` and a `base_slot`

runtime/src/snapshot_utils.rs
```diff
@@ -649,64 +896,280 @@ pub fn bank_from_snapshot_archive<P>(
...
+
+/// Rebuild a bank from full and incremental snapshot archives
+#[cfg(test)]
+#[allow(clippy::too_many_arguments)]
+fn bank_from_incremental_snapshot_archive<P, Q>(
+    account_paths: &[PathBuf],
+    frozen_account_pubkeys: &[Pubkey],
+    snapshot_path: &Path,
+    full_snapshot_archive_path: P,
+    incremental_snapshot_archive_path: Q,
```
carllin, line 925: If we just passed this as an option to the existing `bank_from_snapshot_archive` could we avoid this function?

runtime/src/snapshot_utils.rs
```diff
@@ -536,6 +677,112 @@ where
...
+/// files were consumed correctly.
+fn check_deserialize_snapshot_data_file<P>(
+    file_size: u64,
+    file_path: P,
+    file_stream: &mut BufReader<File>,
+) -> Result<()>
+where
+    P: AsRef<Path>,
+{
+    let consumed_size = file_stream.seek(SeekFrom::Current(0))?;
```
carllin, line 772: I think this is equivalent to `file_stream.stream_position()`

runtime/src/snapshot_utils.rs
```diff
@@ -536,6 +677,112 @@ where
...
+
+    let snapshot_data_file = File::open(&snapshot_root_file_path)?;
+    let snapshot_data_file_stream = BufReader::new(snapshot_data_file);
+
+    Ok((snapshot_file_size, snapshot_data_file_stream))
+}
+
+/// After running the deserializer function, perform common checks to ensure the snapshot archive
+/// files were consumed correctly.
+fn check_deserialize_snapshot_data_file<P>(
```
carllin, line 764: nit: `check_deserialize_snapshot_data_file -> check_deserialize_file_consumed/check_file_consumed`

runtime/src/snapshot_utils.rs
```diff
@@ -536,6 +677,112 @@ where
     Ok(ret)
 }
 
+fn deserialize_snapshot_data_files_capped<F, T>(
```
carllin, line 680: Nice, can the previous version `deserialize_snapshot_data_file_capped` be deprecated/removed now in favor of this version?

runtime/src/snapshot_utils.rs
```diff
@@ -932,46 +1551,86 @@ fn verify_snapshot_version_and_folder(
...
 #[allow(clippy::too_many_arguments)]
-fn rebuild_bank_from_snapshots(
-    snapshot_version: &str,
+fn rebuild_bank_from_snapshots<P, Q>(
+    full_snapshot_version: &str,
+    incremental_snapshot_version: Option<&str>,
     frozen_account_pubkeys: &[Pubkey],
-    unpacked_snapshots_dir: &Path,
+    full_snapshot_unpacked_snapshots_dir: P,
+    incremental_snapshot_unpacked_snapshots_dir: Option<Q>,
```
carllin, line 1559: Might be a good idea to fold `incremental_snapshot_unpacked_snapshots_dir` and `incremental_snapshot_version` into a tuple under the same `Option` here since they are both Some or None, maybe a struct `UnpackedSnapshotDir`? (maybe could use this for the `full_snapshot_unpacked_snapshots_dir + full_snapshot_version` as well

runtime/src/snapshot_utils.rs
```diff
@@ -649,64 +896,280 @@ pub fn bank_from_snapshot_archive<P>(
...
+/// unpacked append vec map.
+fn do_snapshot_unarchive_preparation<P, Q>(
+    snapshot_path: P,
+    unpacked_snapshots_dir_prefix: &'static str,
+    snapshot_archive_path: Q,
+    measure_name: &'static str,
+    account_paths: &[PathBuf],
+    archive_format: ArchiveFormat,
+    parallel_divisions: usize,
+) -> Result<(TempDir, PathBuf, UnpackedAppendVecMap, String, Measure)>
```
carllin, line 1097: This is a rather large tuple that could deserve a struct. Might be good to also include the struct described here: https://github.com/solana-labs/solana/pull/18504/files#r668273623   as a sub-struct of that struct

runtime/src/snapshot_utils.rs
```diff
@@ -932,46 +1551,86 @@ fn verify_snapshot_version_and_folder(
...
+            let (version, paths) =
+                verify_snapshot_version_and_folder(snapshot_version, unpacked_snapshots_dir)?;
+            (Some(version), Some(paths))
+        }
+        _ => (None, None),
+    };
     info!(
-        "Loading bank from {}",
-        &root_paths.snapshot_file_path.display()
+        "Loading bank from {} and {:?}",
```
carllin, line 1591: nit: `Loading bank from full snapshot: {} and incremental snapshot: {:?}`

runtime/src/snapshot_utils.rs
```diff
@@ -476,6 +603,20 @@ where
     )
 }
 
+fn deserialize_snapshot_data_files<F, T>(
+    snapshot_root_paths: &SnapshotRootPaths,
+    deserializer: F,
```
carllin, line 608: nit: throughout the codebase we've been using `deserializer: impl FnOnce...` instead of the `where` clause below. Might be good to keep that consistent across all the functions here.

runtime/src/serde_snapshot.rs
```diff
@@ -321,19 +351,64 @@ where
...
+        .map_or_else(HashMap::new, |fields| {
+            fields
+                .0
+                .into_iter()
+                .filter(|(slot, _)| *slot > full_snapshot_slot)
+                .collect::<HashMap<_, _>>()
+        });
+
+    assert!(incremental_snapshot_storage.iter().all(|storage_entry| !full_snapshot_storage.contains_key(storage_entry.0)),
+    "There must not be any overlap in the slots of storages between the full snapshot and the incremental snapshot");
```
carllin, line 400: Should this be a hard assert or just another error result?

runtime/src/snapshot_utils.rs
```diff
@@ -649,64 +896,280 @@ pub fn bank_from_snapshot_archive<P>(
...
+}
+
+/// Check if an incremental snapshot is compatible with a full snapshot.  This function parses the
+/// paths to see if the incremental snapshot's base slot is the same as the full snapshot's slot.
+/// Return an error if they are incompatible (or if the paths cannot be parsed), otherwise return a
+/// tuple of the full snapshot slot and the incremental snapshot slot.
+fn check_are_snapshots_compatible<P, Q>(
+    full_snapshot_archive_path: P,
+    incremental_snapshot_archive_path: Q,
+) -> Result<(Slot, Slot)>
```
carllin, line 1138: Since it seems the tuple from this check is never actually used, we could either:

1) just return a bool and have the caller return the `SnapshotError::IncompatibleSnapshots`

2) Return a `Result<()>`

runtime/src/serde_snapshot.rs
```diff
@@ -321,19 +351,64 @@ where
...
+    // filter out incremental snapshot storages with slot <= full snapshot slot
+    let incremental_snapshot_storage = snapshot_accounts_db_fields
+        .incremental_snapshot_accounts_db_fields
+        .map_or_else(HashMap::new, |fields| {
+            fields
+                .0
+                .into_iter()
+                .filter(|(slot, _)| *slot > full_snapshot_slot)
+                .collect::<HashMap<_, _>>()
+        });
```
carllin, line 397: Should we filter here or just return an error? Since we checked earlier via `check_are_snapshots_compatible` that the incremental snapshot was based on this full snapshot, there should be no slots `<= full snapshot slot` right?

runtime/src/serde_snapshot.rs
carllin: For instance could a malicious validator make a full snapshot for `X` with slots `Y > X` such that those extra slots `Y` collide with slots in future incremental snapshots?

To prevent this, should we also 
1) verify that a full snapshot for `X` only contains slots `<=X`, 
2) and then if we know incremental snapshots based on `X` only contain slots `> X` from this earlier check here: https://github.com/solana-labs/solana/pull/18504/files#r668284384

Then this non-overlapping naturally falls out of the two guarantees

runtime/src/snapshot_utils.rs
```diff
@@ -417,9 +543,9 @@ pub fn archive_snapshot_package(
     Ok(())
 }
 
-pub fn get_snapshot_paths<P: AsRef<Path>>(snapshot_path: P) -> Vec<SlotSnapshotPaths>
+pub fn get_snapshot_paths<P>(snapshot_path: P) -> Vec<SlotSnapshotPaths>
```
carllin, line 546: nit: This is getting the storage entry paths right? How about `get_slot_storage_paths`

runtime/src/snapshot_utils.rs
```diff
@@ -169,7 +209,8 @@ impl Ord for SlotSnapshotPaths {
     }
 }
 
-pub fn package_snapshot<P: AsRef<Path>, Q: AsRef<Path>>(
+/// Package up slot snapshot files, snapshot storages, and slot deltas for a full snapshot.
+pub fn package_snapshot<P, Q>(
```
carllin, line 213: nit: package_snapshot -> package_full_snapshot

runtime/src/snapshot_utils.rs
```diff
@@ -179,19 +220,104 @@ pub fn package_snapshot<P: AsRef<Path>, Q: AsRef<Path>>(
...
     hash_for_testing: Option<Hash>,
-) -> Result<AccountsPackagePre> {
-    // Hard link all the snapshots we need for this package
+) -> Result<AccountsPackagePre>
+where
+    P: AsRef<Path>,
+    Q: AsRef<Path>,
+{
+    info!(
+        "Package snapshot for bank: {} has {} account storage entries",
```
carllin, line 229: nit: `Package snapshot -> Package full snapshot`

runtime/src/snapshot_utils.rs
```diff
@@ -179,19 +220,104 @@ pub fn package_snapshot<P: AsRef<Path>, Q: AsRef<Path>>(
...
+
+    do_package_snapshot(
+        bank,
+        snapshot_files,
+        status_cache_slot_deltas,
+        snapshot_package_output_path,
+        snapshot_storages,
+        archive_format,
+        snapshot_version,
+        hash_for_testing,
```
carllin, line 300: Since all of these arguments are shared between `do_package_snapshot`, `package_incremental_snapshot`, and `package_snapshot`, it might be cleaner to wrap these up in an argument struct

runtime/src/snapshot_utils.rs
carllin, line 546: nit: `get_snapshot_paths -> get_slot_snapshot_paths`, 

and a comment along the lines of, "returns a vector of paths to existing snapshot directories under `snapshot_path`"

runtime/src/snapshot_utils.rs
```diff
@@ -135,6 +161,14 @@ pub struct SlotSnapshotPaths {
     pub snapshot_file_path: PathBuf,
 }
```
carllin, line 162: really want to rename this to `SlotSnapshotPath`, the plural is confusing here when its only signifying one slot

runtime/src/snapshot_utils.rs
```diff
@@ -1131,9 +1790,82 @@ pub fn bank_to_snapshot_archive<P: AsRef<Path>, Q: AsRef<Path>>(
...
+    bank.squash(); // Bank may not be a root
+    bank.force_flush_accounts_cache();
+    bank.clean_accounts(true, false);
+    bank.update_accounts_hash();
+    bank.rehash(); // Bank accounts may have been manually modified by the caller
+
+    let temp_dir = tempfile::tempdir_in(snapshot_path)?;
+
+    let storages = bank.get_incremental_snapshot_storages(full_snapshot_slot);
+    let slot_snapshot_paths = add_snapshot(&temp_dir, bank, &storages, snapshot_version)?;
```
carllin, line 1822: nit: `slot_snapshot_paths -> slot_snapshot_path`, since it's only one path?

runtime/src/snapshot_utils.rs
```diff
@@ -179,19 +220,104 @@ pub fn package_snapshot<P: AsRef<Path>, Q: AsRef<Path>>(
...
+        snapshot_tmpdir,
+    )
+}
+
+/// Package up slot snapshot files, snapshot storages, and slot deltas for an incremental snapshot.
+#[allow(clippy::too_many_arguments)]
+fn package_incremental_snapshot<P, Q>(
+    bank: &Bank,
+    incremental_snapshot_base_slot: Slot,
+    snapshot_files: &SlotSnapshotPaths,
```
carllin, line 256: `snapshot_files -> slot_snapshot_path` since it's only one slot?

runtime/src/snapshot_utils.rs
lijunwangs: I second that -- it does look a lot of fields have the same names and have same meanings.

runtime/src/snapshot_utils.rs
lijunwangs: I think at interface level maximum_snapshots_to_retain is good enough. Snapshot is a snapshot. Incremental or full is implementation detail. for efficiency. The user may be advised that with incremental feature they can afford to keep more snapshots around for the same storage overhead.

runtime/src/snapshot_utils.rs
```diff
@@ -1131,9 +1790,82 @@ pub fn bank_to_snapshot_archive<P: AsRef<Path>, Q: AsRef<Path>>(
...
+    full_snapshot_slot: Slot,
+    snapshot_version: Option<SnapshotVersion>,
+    snapshot_package_output_path: Q,
+    archive_format: ArchiveFormat,
+    thread_pool: Option<&ThreadPool>,
+    maximum_snapshots_to_retain: usize,
+) -> Result<PathBuf> {
+    let snapshot_version = snapshot_version.unwrap_or_default();
+
+    assert!(bank.is_complete());
```
lijunwangs, line 1811: Maybe we can put this common code into a separate function to share with bank_to_snapshot_archive?

lijunwangs:
What is the impact for existing validators downloading snapshots from validators generating incremental snapshots? Will they work off the full snapshot available?

runtime/src/snapshot_utils.rs
brooksprumo: I've created a new issue to track this TODO, #18639. I think this may need some additional configuration options. Stephen has mentioned wanting to keep old (full) snapshots around longer for replay purposes, and I could see similar configurations requested for incremental snapshots.

runtime/src/serde_snapshot.rs
brooksprumo: OK, new issue has been created to track this: #18641.

runtime/src/snapshot_utils.rs
brooksprumo: Love it. Will do!

runtime/src/snapshot_utils.rs
brooksprumo: Yep, can have just the one top-level function and not add a new one. Will do!

runtime/src/snapshot_utils.rs
brooksprumo: Neat! Yeah, the actual code wasn't changed from the original implementation, just moved. I'll make this change though, since I agree it is simpler.

runtime/src/snapshot_utils.rs
brooksprumo: Yes, that can be done. I was wondering about that myself. This function is also called when deserializing the status cache[1], and that's just a single file, so I left this function here for that. I could change the status cache code to call this version, but I thought that looked clunky and didn't communicate intent well. Whatcha think?

[1] https://github.com/brooksprumo/solana/blob/4f45db66c833ab1828c668c84de05b55e26a83e8/runtime/src/snapshot_utils.rs#L1634

brooksprumo: Oh, sorry, I think I misunderstood ^^^

Yes, the `_capped()` versions can be collapsed into just one. That's the intent, yes?

runtime/src/snapshot_utils.rs
brooksprumo: I like it. Will do!

runtime/src/snapshot_utils.rs
brooksprumo: I agree! Will do.

runtime/src/snapshot_utils.rs
brooksprumo: Ah OK. I was following the existing impl of the file; I'll happily make these change though!

runtime/src/serde_snapshot.rs
brooksprumo: Ooo good points. I like this idea and will do.

runtime/src/snapshot_utils.rs
brooksprumo: Ah good catch! This was leftover when I used the return values to set the last full snapshot slot and last incremental snapshot slot. I'll remove those here.

I think I may need them again when doing the `clean_accounts()` work, but that'll be easy to re-add then.

runtime/src/snapshot_utils.rs
brooksprumo: Yah, I wasn't sure what to do; either rename all the existing things to `full_`, or not. Since currently a "snapshot" is a full snapshot, that's why I didn't add `full_` to all the functions/variables. Inside all the incremental snapshot functions, that's where everything is either `full_` or `incremental_`.

I also wasn't sure how far to take the renaming. Like should it cause gossip/RPC to change names too? I'm not sure on that part. What do you think @carllin ?

runtime/src/snapshot_utils.rs
brooksprumo: omg yes. I wanted to do this myself too. I initially tried to change as little as possible for this PR. Happy to make this change!

runtime/src/snapshot_utils.rs
brooksprumo: I was very confused when I first read through the code. Happy to change this :)

runtime/src/snapshot_utils.rs
```diff
@@ -649,64 +896,280 @@ pub fn bank_from_snapshot_archive<P>(
...
+        _full_snapshot_unpack_dir,
+        full_snapshot_unpacked_snapshots_dir,
+        full_snapshot_unpacked_append_vec_map,
+        full_snapshot_version,
+        full_snapshot_measure_untar,
+    ) = do_snapshot_unarchive_preparation(
+        snapshot_path,
+        TMP_SNAPSHOT_PREFIX,
+        &full_snapshot_archive_path,
+        "snapshot untar",
```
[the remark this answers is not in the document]
brooksprumo: Yeah, I wasn't sure how much to rename. See this comment: https://github.com/solana-labs/solana/pull/18504#discussion_r668915990

My thought for this PR was everything that didn't include "incremental" meant it was a full snapshot, which is the way it works today, since all snapshots are full snapshots. Then within the context of incremental snapshot code, I would always add "full" to differentiate. Whatcha think?

runtime/src/snapshot_utils.rs
```diff
@@ -649,64 +896,280 @@ pub fn bank_from_snapshot_archive<P>(
...
+    measure_verify.stop();
+
     let timings = BankFromArchiveTimings {
-        rebuild_bank_from_snapshots_us: measure.as_us(),
-        untar_us: untar.as_us(),
-        verify_snapshot_bank_us: verify.as_us(),
+        rebuild_bank_from_snapshots_us: measure_rebuild.as_us(),
+        untar_us: full_snapshot_measure_untar.as_us()
+            + incremental_snapshot_measure_untar.map_or(0, |measure| measure.as_us()),
+        verify_snapshot_bank_us: measure_verify.as_us(),
```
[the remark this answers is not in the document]
brooksprumo: Good idea, will do!

runtime/src/snapshot_utils.rs
```diff
@@ -649,64 +896,280 @@ pub fn bank_from_snapshot_archive<P>(
...
+    full_snapshot_archive_path: P,
+    incremental_snapshot_archive_path: Q,
+) -> Result<(Slot, Slot)>
+where
+    P: AsRef<Path>,
+    Q: AsRef<Path>,
+{
+    fn path_to_filename(path: &Path) -> Result<&str> {
+        path.file_name()
+            .ok_or(SnapshotError::PathParseError("Could not get file name!"))?
```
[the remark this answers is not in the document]
brooksprumo: Good point. I will fix this.

runtime/src/snapshot_utils.rs
```diff
@@ -796,6 +1296,69 @@ fn sort_snapshot_archives(snapshot_archives: &mut Vec<SnapshotArchiveInfo>) {
...
+/// Get a list of the incremental snapshot archives in a directory
+fn get_incremental_snapshot_archives<P>(
+    snapshot_output_dir: P,
+) -> Vec<IncrementalSnapshotArchiveInfo>
+where
+    P: AsRef<Path>,
+{
+    match fs::read_dir(&snapshot_output_dir) {
+        Err(err) => {
+            info!("Unable to read snapshot directory: {}", err);
```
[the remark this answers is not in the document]
brooksprumo: I believe so. I will fix this.

brooksprumo:
> What is the impact for existing validators downloading snapshots from validators generating incremental snapshots? Will they work off the full snapshot available?

Sorry I could reply to your comment directly, @lijunwangs. Firstly, in this PR there are no changes to the actual way validators function; no incremental snapshots are taken. Going forward, yes, validators that are running older versions without incremental snapshot support will continue to operate the same as today: asking for (and working off) full snapshots.

runtime/src/snapshot_utils.rs
brooksprumo: Done.

runtime/src/snapshot_utils.rs
brooksprumo: Done.

runtime/src/snapshot_utils.rs
brooksprumo: Done.

runtime/src/snapshot_utils.rs
brooksprumo: Done.

runtime/src/snapshot_utils.rs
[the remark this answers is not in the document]

runtime/src/snapshot_utils.rs

runtime/src/snapshot_utils.rs
brooksprumo: Same naming questions from here [1] on how to name snapshots.

[1] https://github.com/solana-labs/solana/pull/18504#discussion_r668915990

runtime/src/snapshot_utils.rs

runtime/src/snapshot_utils.rs
brooksprumo: Done!

runtime/src/snapshot_utils.rs

runtime/src/snapshot_utils.rs

runtime/src/snapshot_utils.rs
brooksprumo: Oh, and also the on-disk snapshots right now are also named "snapshot" and are full snapshots. To handle renaming the on-disk snapshots, I'd need to add code to support both `full_snapshot` and `snapshot` files/directories for a while, for backwards compatibility.

runtime/src/snapshot_utils.rs

runtime/src/snapshot_utils.rs
brooksprumo: Same naming questions from here [1] on how to name snapshots.

[1] https://github.com/solana-labs/solana/pull/18504#discussion_r668915990

runtime/src/snapshot_utils.rs

runtime/src/snapshot_utils.rs

runtime/src/snapshot_utils.rs

runtime/src/snapshot_utils.rs
```diff
@@ -649,64 +896,280 @@ pub fn bank_from_snapshot_archive<P>(
...
+            &incremental_snapshot_archive_path,
+        )?;
 
-    let mut snapshot_version = String::new();
-    File::open(unpacked_version_file).and_then(|mut f| f.read_to_string(&mut snapshot_version))?;
+        let (
+            incremental_snapshot_unpack_dir,
+            incremental_snapshot_unpacked_snapshots_dir,
+            incremental_snapshot_unpacked_append_vec_map,
+            incremental_snapshot_version,
```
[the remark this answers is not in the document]
brooksprumo: I wondered about this same thing. In a previous PR I had a TODO in the code and Carl answered the question:

https://github.com/solana-labs/solana/pull/17875#discussion_r652452533

So AFAIK the handling here of versions is sufficient. Is that OK? Would you like to see something else?

runtime/src/snapshot_utils.rs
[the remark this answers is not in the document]
brooksprumo: Done!

runtime/src/snapshot_utils.rs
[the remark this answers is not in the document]

runtime/src/serde_snapshot.rs
brooksprumo: Yeah... this one confused me too. So when AccountsHashVerifier is taking hashes, it's doing it from snapshots. Those snapshots (not the whole snapshot archive) need to have all the storages to compute the hash correctly.

When I was writing tests in `core/tests/snapshots.rs`, which calls functions a little differently to create snapshot archives, I was running into issues where rebuild would fail due to AppendVec ID mismatch for slots less than the full snapshot slot in the incremental snapshot. The accounts_db_fields.0 also has info for all the slots here, so this filtering needs to happen when rebuilding with an incremental snapshot.

Both the accounts_db_fields.0 _and_ the actual storages need to be filtered out. The storages can get filtered out when creating the incremental snapshot archive, and then the accounts_db_fields.0 needs to be filtered out when rebuilding. Here's some more info when I was asking about it on Discord [1].

[1] https://discord.com/channels/428295358100013066/439194979856809985/854856681082978304

runtime/src/serde_snapshot.rs
brooksprumo: I've changed the assert to return an error instead. Is that sufficient for now?

I originally put in the assert from your comment here: https://github.com/solana-labs/solana/pull/17875#discussion_r652475517

runtime/src/snapshot_utils.rs
brooksprumo: Yeah, I thought about doing that too. Unfortunately the only real common code are the first few lines of `bank.xxx()`, then afterwards they are different depending on full snapshot vs incremental snapshot. I decided to leave them separate for that reason, as the duplication wasn't much. Also, I am anticipating needing to change the impl in the incremental snapshot side w.r.t. clean_accounts(), so that's another reason I've left them separate. Is that OK for now?

runtime/src/snapshot_utils.rs
brooksprumo: While implementing this, I ended up feeling like I didn't actually help anything, just move around where I was writing all the parameters, and even adding a bit of work to create the struct and then destructure it later. How strongly would you like this change to go in?

brooksprumo:
@carllin @lijunwangs OK! I had fixed or replied to all the review comments. Please take another pass and let me know what you think. Thanks for all the help so far!

@ryoqun Sorry, I should've added you earlier as well!

ledger/src/blockstore_processor.rs
```diff
@@ -559,7 +559,16 @@ fn do_process_blockstore_from_root(
         ("slot", bank_forks.root(), i64),
         ("forks", initial_forks.len(), i64),
         ("calculate_capitalization_us", time_cap.as_us(), i64),
-        ("untar_us", timings.untar_us, i64),
```
ryoqun, line 562: unintentional remove due to merge conflict?

ryoqun: oops. never mind.. prefixed...

runtime/src/serde_snapshot.rs
```diff
@@ -326,19 +356,68 @@ where
...
+
+    // There must not be any overlap in the slots of storages between the full snapshot and the incremental snapshot
+    incremental_snapshot_storage
+        .iter()
+        .all(|storage_entry| !full_snapshot_storage.contains_key(storage_entry.0)).then(|| ()).ok_or_else(|| {
+            io::Error::new(io::ErrorKind::InvalidData, "Snapshots are incompatible: There are storages for the same slot in both the full snapshot and the incremental snapshot!")
+        })?;
```
ryoqun, line 409: how about extracting this code block to a newly-defined method at `SnapshotAccountsDbFields`?

runtime/src/serde_snapshot.rs
```diff
@@ -326,19 +356,68 @@ where
...
     for path in &accounts_db.paths {
         std::fs::create_dir_all(path)
             .unwrap_or_else(|err| panic!("Failed to create directory {}: {}", path.display(), err));
     }
 
-    let storage = storage.into_iter().collect::<Vec<_>>();
+    let storage = full_snapshot_storage
+        .into_iter()
+        .chain(incremental_snapshot_storage.into_iter())
+        .collect::<Vec<_>>();
```
ryoqun, line 420: this

runtime/src/serde_snapshot.rs
ryoqun: i think this incremental/full snapshot new logic can be well encapsulated, considering this following code (https://github.com/solana-labs/solana/pull/18504/files#r671230020) like this:

```rust

impl SnapshotAccountsDbFields {
    fn combined_storages(...) -> Vec<_> {
    }
}

```

runtime/src/snapshot_utils.rs
```diff
@@ -417,43 +543,56 @@ pub fn archive_snapshot_package(
...
-                        slot,
-                        snapshot_file_path: snapshot_path.join(get_snapshot_file_name(slot)),
-                    }
+    match fs::read_dir(&snapshots_dir) {
+        Ok(paths) => paths
+            .filter_map(|entry| {
+                entry.ok().and_then(|e| {
+                    e.path()
+                        .file_name()
+                        .and_then(|n| n.to_str().map(|s| s.parse::<u64>().ok()))
```
ryoqun, line 557: nits: `u64` => `Slot`

runtime/src/snapshot_utils.rs
```diff
@@ -489,7 +635,7 @@ where
     serializer(&mut data_file_stream)?;
     data_file_stream.flush()?;
 
-    let consumed_size = data_file_stream.seek(SeekFrom::Current(0))?;
+    let consumed_size = data_file_stream.stream_position()?;
```
ryoqun, line 638: TIL :)

runtime/src/snapshot_utils.rs
```diff
@@ -649,66 +901,188 @@ pub fn bank_from_snapshot_archive<P>(
 ) -> Result<(Bank, BankFromArchiveTimings)>
 where
     P: AsRef<Path> + std::marker::Sync,
+    Q: AsRef<Path> + std::marker::Sync,
```
ryoqun, line 904: dumb question: so, `full_snapshot_archive_path` and `incremental_snapshot_archive_path` could be different types for any given this fn invocation? if possible, I want to avoid another type param (Q in this case)...

runtime/src/snapshot_utils.rs
```diff
@@ -649,66 +901,188 @@ pub fn bank_from_snapshot_archive<P>(
...
-    let divisions = std::cmp::min(
+    let parallel_divisions = std::cmp::min(
         PARALLEL_UNTAR_READERS_DEFAULT,
         std::cmp::max(1, num_cpus::get() / 4),
     );
-    let unpacked_append_vec_map = untar_snapshot_in(
-        &snapshot_tar,
-        unpack_dir.as_ref(),
+
+    let full_snapshot_unarchive_preparation_result = do_snapshot_unarchive_preparation(
```
ryoqun, line 911: nits: rename to `prepared_snapshot_archive`? a bit mouthful variable name, imo... like `Option<_>`, `Result<_>` type variable doesn't need to be suffixed with `_result`.

runtime/src/snapshot_utils.rs
```diff
@@ -154,53 +194,139 @@ pub enum SnapshotError {
...
-        self.slot.cmp(&other.slot)
-    }
+    #[error("could not get str from file name: {}", .0.display())]
+    FileNameToStrError(PathBuf),
+
+    #[error("could not parse snapshot archive's file name: {0}")]
+    ParseSnapshotArchiveFileNameError(String),
+
+    #[error("snapshots are incompatible: full snapshot slot ({0}) and incremental snapshot base slot ({1}) do not match")]
+    IncompatibleSnapshots(Slot, Slot),
```
ryoqun, line 208: nits: `MismatchedBaseSlot`? `Incompatible` a bit sounds too general.

runtime/src/snapshot_utils.rs
```diff
@@ -750,14 +1143,36 @@ fn parse_snapshot_archive_filename(archive_filename: &str) -> Option<(Slot, Hash
...
 }
 
+/// Parse an incremental snapshot archive filename into its base Slot, actual Slot, Hash, and Archive Format
+fn parse_incremental_snapshot_archive_filename(
+    archive_filename: &str,
+) -> Option<(Slot, Slot, Hash, ArchiveFormat)> {
+    let regex = Regex::new(INCREMENTAL_SNAPSHOT_ARCHIVE_FILENAME_REGEX);
+
+    regex.ok()?.captures(archive_filename).and_then(|captures| {
+        let base_slot = captures.get(1).map(|x| x.as_str().parse::<Slot>())?.ok()?;
```
ryoqun, line 1153: maybe, there is no `captures.iter()` or something similar? I like multiple `capture_iter.next()`s than `.get(1)`/`.get(2)`/`.get(3)`.... (no magic numbers if possible..)

runtime/src/snapshot_utils.rs
```diff
@@ -798,6 +1213,75 @@ fn sort_snapshot_archives(snapshot_archives: &mut Vec<SnapshotArchiveInfo>) {
...
+    sort_incremental_snapshot_archives(&mut incremental_snapshot_archives);
+    incremental_snapshot_archives
+}
+
+/// Sort the list of incremental snapshot archives, first by base (full snapshot) slot in
+/// descending order, then by "regular" (incremental snapshot) slot in descending order
+fn sort_incremental_snapshot_archives(
+    incremental_snapshot_archives: &mut Vec<IncrementalSnapshotArchiveInfo>,
+) {
+    incremental_snapshot_archives.sort_unstable_by(|a, b| {
```
ryoqun, line 1278: how about defining `Ord` for `IncrementalSnapshotArchiveInfo`, so that we can just call `sort()`?

runtime/src/snapshot_utils.rs
```diff
@@ -536,18 +679,121 @@ where
...
+            incremental_snapshot_data_file_stream.as_mut().unwrap(),
+        )?;
+    }
+
+    Ok(ret)
+}
+
+/// Before running the deserializer function, perform common operations on the snapshot archive
+/// files, such as checking the file size and opening the file into a stream.
+fn prepare_deserialize_snapshot_data_file<P>(
```
ryoqun, line 737: nits: naming: `create_snapshot_data_file_stream`? `prepare_` is a bit too general. I know you picked `prepare_` because of the extra checks to encompass these functionalities. but as par defensive programming, doing routine checks are obvious unless we're too lazy. xD

runtime/src/snapshot_utils.rs
```diff
@@ -817,23 +1311,67 @@ where
...
-            .unwrap_or_else(|err| info!("Failed to remove old snapshot: {:}", err));
+            .unwrap_or_else(|err| info!("Failed to remove old snapshot: {}", err));
     }
+
+    // Only keep incremental snapshots for the latest full snapshot
+    // bprumo TODO issue #18639: As an option to further reduce the number of incremental
+    // snapshots, only a subset of the incremental snapshots for the lastest full snapshot could be
+    // kept.  This could reuse maximum_snapshots_to_retain, or use a new field just for incremental
+    // snapshots.
+    let last_full_snapshot_slot = get_highest_snapshot_archive_slot(snapshot_output_dir.as_ref());
```
ryoqun, line 1363: if last_full_snapshot_slot is None, purging interval snapshots should happen? I don't think `Some(archive_info.base_slot) < None` evaluates to `true`...

runtime/src/snapshot_utils.rs
```diff
@@ -798,6 +1213,75 @@ fn sort_snapshot_archives(snapshot_archives: &mut Vec<SnapshotArchiveInfo>) {
...
+                            return Some(IncrementalSnapshotArchiveInfo {
+                                inner: SnapshotArchiveInfo {
+                                    path,
+                                    slot,
+                                    hash,
+                                    archive_format,
+                                },
+                                base_slot,
+                            });
+                        }
```
ryoqun, line 1251: maybe extract to `IncrementalSnapshotArchiveInfo::new_from_path(path) -> Result<Self>`?

runtime/src/snapshot_utils.rs
```diff
@@ -1038,14 +1619,20 @@ pub fn verify_snapshot_archive<P, Q, R>(
     assert!(!dir_diff::is_different(&storages_to_verify, unpacked_accounts).unwrap());
 }
 
-pub fn purge_old_snapshots(snapshot_path: &Path) {
-    // Remove outdated snapshots
-    let slot_snapshot_paths = get_snapshot_paths(snapshot_path);
+/// Remove outdataed snapshots
```
ryoqun, line 1622: nits: `outdated`

ryoqun:
> @ryoqun Sorry, I should've added you earlier as well!

no worries. i was lurking. xD late for the party but i finally reviewed _the great of plumbing_ pr. well cooked and documented work. :)

runtime/src/snapshot_utils.rs
brooksprumo: In theory yes the types could be different, but in practice probably not. I forget which Rust resource I read that recommended this, but I did see it someone, so that one could be a `&Path` and the other could be a `PathBuf` and the function doesn't need to care about that. Without `Q`, then both FSS and ISS paths need to be the same type. Would you like me to remove `Q` here?

runtime/src/snapshot_utils.rs
brooksprumo: I agree; no magic numbers if possible! I'll take a look to see if there's another way (i.e. `.iter()`) to accomplish this.

runtime/src/snapshot_utils.rs
brooksprumo: I actually purposely didn't want to do that, since I didn't feel there was a _universal_ ordering of this type that applied in all situations. We know in the context of the _callers_ of this function that they want the sorting to be first by base slot in descending order, then by "regular" slot in descending order, but maybe that's not the case for the next caller to come along.

What do you think? I can definitely change this to use `Ord` instead, as we know all the callers of this function.

runtime/src/snapshot_utils.rs
brooksprumo: If `last_full_snapshot_slot` is `None`, then there are also no incremental snapshots to remove (since it is not possible to create an incremental snapshot until a full snapshot is created first). I think this should work OK then, yes?

runtime/src/snapshot_utils.rs

runtime/src/snapshot_utils.rs
brooksprumo: I like it. Done!

runtime/src/snapshot_utils.rs

runtime/src/snapshot_utils.rs
brooksprumo: Goo points. Done!

runtime/src/snapshot_utils.rs

runtime/src/snapshot_utils.rs

runtime/src/snapshot_utils.rs

runtime/src/serde_snapshot.rs
brooksprumo: Done!

brooksprumo:
@ryoqun OK! I think I've addressed all of your review comments. Most are fixed in the code now. I replied to the other comments with questions for clarification or how to proceed. I'll re-request the review as well. Thanks!

runtime/src/snapshot_utils.rs
```diff
@@ -417,43 +574,56 @@ pub fn archive_snapshot_package(
...
-                .collect::<Vec<SlotSnapshotPaths>>();
-
-            names.sort();
-            names
-        }
+            })
+            .map(|slot| {
+                let snapshot_file_name = get_snapshot_file_name(slot);
+                // Yep, you read that right, the same snapshot file name is joined twice in a row.
+                // "Why" you ask?  This I do not know.
```
ryoqun, line 595: i think this is just overlook from original implementation... so could you word like this?

```
// The redundant snapshot_file_name in path is unintentional and should be simplified
// kept just for compatibility.
```

the motto here is that no Qs into our own code... xD

runtime/src/snapshot_utils.rs
ryoqun: @brooksprumo yeah, in general, fewer type params are desired to make code simpler if our current code base allows.

> I read that recommended this, but I did see it somewhere... So that one could be a &Path and the other could be a PathBuf and the function doesn't need to care about that.

seems you have good memory. I think this is perfectly good practice for library context. Unless sdk or related, our codebase is mostly application. so, we don't need to be that _generic_ :)

runtime/src/snapshot_utils.rs
ryoqun: @brooksprumo heh, welcome to happy fs programming especially under byzantine (or untrained ops) environment... :) well, admins can manually remove all of full snapshots and then `solana-validator` will leave these outdated and bricked incremental snapshots indefinitely with this current code? I want to prevent the garbage from piling to avoid another mysterious support ticket.. lol

runtime/src/snapshot_utils.rs
ryoqun: ooh, `a` and `b` is swapped in the code... 

could you at least comment it. I think it's too easy to miss this...

or better, I rather strive to keep to avoid descending order except performance sensitive code.

in this case, I'd write like this:

```rust
impl Ord for IncrementalSnapshotArchiveInfo {
   ... // COMMENT: yea, (base slot asc, regular slot asc) should be kind of universal order, assuming it's aligning with chronological order in the  snapshot lifecycle.
}

...


sort_incremental_snapshot_archives(&mut incremental_snapshot_archives); // COMMENT: or just call straightly: incremental_snapshot_archives.sort_unstable();


// pick the highest (i.e. latest one) one
incremental_snapshot_archives.into_iter().rev().next()
```

runtime/src/snapshot_utils.rs

runtime/src/snapshot_utils.rs
```diff
@@ -649,66 +934,177 @@ pub fn bank_from_snapshot_archive<P>(
...
+/// unpacked append vec map.
+fn do_snapshot_unarchive_preparation<P, Q>(
+    snapshots_dir: P,
+    unpacked_snapshots_dir_prefix: &'static str,
+    snapshot_archive_path: Q,
+    measure_name: &'static str,
+    account_paths: &[PathBuf],
+    archive_format: ArchiveFormat,
+    parallel_divisions: usize,
+) -> Result<UnarchivePreparationResult>
```
ryoqun, line 1036: > `Result<UnarchivePreparationResult>`

I think the type sig of this fn is a bit mouthful and I want to avoid 2 nested `Result`s... could you reword this function like this?

```rust
struct UnarchivedSnapshot { // <= renamed from UnarchivePreparationResult
}


fn unarchive_snapshot(...) -> Result<UnarchivedSnapshot> { // renamed from do_snapshot_unarchive_preparation
}
```

better names are welcome. btw, I don't like too reliance of `MetaFooBar`, `FooBarInfo`, `FooBarContext` across our codebase, so I'm eager to remove them. xD

i think enginner's job is to name things properly with good vocabulary for future generation's code readability.. lol

runtime/src/serde_snapshot.rs
```diff
@@ -326,25 +397,31 @@ where
...
+
+    let snapshot_storages = snapshot_storages.into_iter().collect::<Vec<_>>();
 
     // Ensure all account paths exist
     for path in &accounts_db.paths {
         std::fs::create_dir_all(path)
             .unwrap_or_else(|err| panic!("Failed to create directory {}: {}", path.display(), err));
     }
 
-    let storage = storage.into_iter().collect::<Vec<_>>();
```
ryoqun, line 341: good bye bad singular name. :)

runtime/src/snapshot_utils.rs
```diff
@@ -500,14 +679,11 @@ where
     Ok(consumed_size)
 }
 
-fn deserialize_snapshot_data_file_capped<F, T>(
+fn deserialize_snapshot_data_file_capped<T: Sized>(
```
ryoqun, line 682: out of curious, why `T: Sized` trait bound is needed now?

runtime/src/snapshot_utils.rs
brooksprumo: I like this, done!

runtime/src/snapshot_utils.rs
brooksprumo: OK, I've put `Ord` back in and reworked the code to use `sort_unstable()` directly.

runtime/src/snapshot_utils.rs

runtime/src/snapshot_utils.rs
brooksprumo: Gotcha! I've made it so now if there are no full snapshots that all the incremental snapshots will be purged. Done!

brooksprumo:
@lijunwangs @ryoqun @carllin Checks have passed, and all review comments have been addressed. I've replied with answers/questions to any outstanding review comments. PR is ready for another round of reviewing. Thanks in advance!

runtime/src/snapshot_utils.rs
ryoqun: oh i see. yeah, `impl Trait` requires `Sized`, iirc. no problem then. :)

runtime/src/snapshot_utils.rs
ryoqun: I think existing identifiers should be prefixed with `full_` to avoid any ambiguity.

and file system and rpc / gossip should remain same.

in other words, `full_` rename should occur in all code where appropriate for humans, while keeping compatibility.

it's a bit daunting for me. but could you do that once for all for us, considering your pr's high quality of plumbing?

ryoqun:
> - [ ] **Assert vs error when storages overlap.** [Thread 1 here](https://github.com/solana-labs/solana/pull/18504#discussion_r668281871), and [thread 2 here](https://github.com/solana-labs/solana/pull/18504#discussion_r668284384)

@brooksprumo as a context, we want to support untrusted snapshots eventually in the (distant) future. so, panic! shouldn't be reachable when consuming (= untaring) any downloaded snapshots. #7167

That context might help to decide panic!/error?

Also, well-groomed pr description. :)

runtime/src/snapshot_utils.rs
brooksprumo: @ryoqun Yes, will do.

brooksprumo: OK, done!

brooksprumo:
> @brooksprumo as a context, we want to support untrusted snapshots eventually in the (distant) future. so, panic! shouldn't be reachable when consuming (= untaring) any downloaded snapshots. #7167
> 
> That context might help to decide panic!/error?

@ryoqun Yes, thank you. Makes sense!

> Also, well-groomed pr description. :)

Thanks :)

runtime/src/snapshot_utils.rs
carllin: > in other words, full_ rename should occur in all code where appropriate for humans, while keeping compatibility.

Sorry late to the party here, agreed, especially given we're introducing `incremental_snapshots`

runtime/src/serde_snapshot.rs
carllin: thanks!

runtime/src/serde_snapshot.rs
carllin: Man, what a flip flopper - can't trust this guy with anything 😛

runtime/src/snapshot_utils.rs
carllin: heh, awesome, thanks!

runtime/src/snapshot_utils.rs
carllin: nice, much more readable, thanks :)

runtime/src/snapshot_utils.rs
carllin: I think the "full_" vs "incremental_" is good for clarity, thanks!

runtime/src/snapshot_utils.rs
carllin: Heh, looking at it again, I think `package_incremental_snapshot` and  `package_snapshot` may just be better off as constructors for `AccountsPackagePre`, where logic like https://github.com/solana-labs/solana/pull/18504/files#diff-c957621ae265a12e03f7b9c5656e7aa74137223e603cc3cfcc2828dda1af5479R398-R405 could be moved into a constructor.

Also has the added benefit of offloading from the `snapshot_utils` module as it's growing rather large.

runtime/src/snapshot_utils.rs
carllin: > Yes, the _capped() versions can be collapsed into just one. That's the intent, yes?

yeah it seems like one function should be enough to handle either just a full snapshot or a full + incremental snapshot

runtime/src/serde_snapshot.rs
```diff
@@ -74,6 +74,62 @@ struct AccountsDbFields<T>(
...
+        match self.incremental_snapshot_accounts_db_fields {
+            None => Ok(self.full_snapshot_accounts_db_fields),
+            Some(AccountsDbFields(
+                mut incremental_snapshot_storages,
+                incremental_snapshot_version,
+                incremental_snapshot_slot,
+                incremental_snapshot_bank_hash_info,
+            )) => {
+                let full_snapshot_storages = self.full_snapshot_accounts_db_fields.0;
+                let full_snapshot_slot = self.full_snapshot_accounts_db_fields.2;
```
carllin, line 107: heh, really wish we'd gone with named fields for `AccountsDbFields` 😛

runtime/src/serde_snapshot.rs
```diff
@@ -74,6 +74,62 @@ struct AccountsDbFields<T>(
...
+                    combined_storages,
+                    incremental_snapshot_version,
+                    incremental_snapshot_slot,
+                    incremental_snapshot_bank_hash_info,
+                ))
+            }
+        }
+    }
+}
+
```
carllin, line 132: nice, like what you did here

runtime/src/serde_snapshot.rs
carllin: Little confused here, let's have a brief chat to go over it.

runtime/src/serde_snapshot.rs
brooksprumo: Me too. I had grabbed `.1` for `slot` and all my tests were failing until I realized...

runtime/src/snapshot_utils.rs
brooksprumo: Gotcha. Done!

lijunwangs (approved):
Looks good to me.

runtime/src/serde_snapshot.rs
brooksprumo: After talking with @carllin, I've created a new issue to track this: #18813. The gist is during serialization of accounts_db for an incremental snapshot, to filter _out_ the storage entries for all slots less-than-or-equal-to the full snapshot slot.

runtime/src/snapshot_utils.rs
brooksprumo: After talking with @carllin, these changes will not be part of this PR. I'll likely pull it into another change later. Basically just pulling the TempDir into constructors for AccountsPackagePre.

brooksprumo: New issue created to track this work: #18815

runtime/src/serde_snapshot.rs
carllin: Thanks!

carllin (approved):

brooksprumo:
@carllin @lijunwangs I had to fix a merge conflict, so your approvals were removed. Nothing else changed, but let me know if you'd like me to wait for you to re-review before I merge.

@ryoqun Would you like to re-review before I merge?

runtime/src/snapshot_utils.rs
```diff
@@ -750,105 +1224,220 @@ fn archive_format_from_str(archive_format: &str) -> Option<ArchiveFormat> {
...
+
+    // Only keep incremental snapshots for the latest full snapshot
+    // bprumo TODO issue #18639: As an option to further reduce the number of incremental
+    // snapshots, only a subset of the incremental snapshots for the lastest full snapshot could be
+    // kept.  This could reuse maximum_snapshots_to_retain, or use a new field just for incremental
+    // snapshots.
+    // In case there are incremental snapshots but no full snapshots, make sure all the incremental
+    // snapshots are purged.
+    let last_full_snapshot_slot =
+        get_highest_full_snapshot_archive_slot(&snapshot_archives_dir).unwrap_or(Slot::MAX);
```
ryoqun, line 1428: :eyes: good

runtime/src/snapshot_utils.rs
```diff
@@ -665,81 +1011,209 @@ pub fn bank_from_snapshot_archive<P>(
...
+        account_paths,
+        archive_format,
+        parallel_divisions,
+    )?;
+    measure_untar.stop();
+    info!("{}", measure_untar);
+
+    let unpacked_version_file = unpack_dir.path().join("version");
+    let mut snapshot_version = String::new();
+    File::open(unpacked_version_file).and_then(|mut f| f.read_to_string(&mut snapshot_version))?;
```
ryoqun, line 1134: here

runtime/src/snapshot_utils.rs
```diff
@@ -925,72 +1514,116 @@ fn untar_snapshot_in<P: AsRef<Path>>(
...
 
-fn verify_snapshot_version_and_folder(
-    snapshot_version: &str,
-    unpacked_snapshots_dir: &Path,
-) -> Result<(SnapshotVersion, SlotSnapshotPaths)> {
-    info!("snapshot version: {}", snapshot_version);
+fn verify_unpacked_snapshots_dir_and_version(
+    unpacked_snapshots_dir_and_version: &UnpackedSnapshotsDirAndVersion,
+) -> Result<(SnapshotVersion, BankSnapshotInfo)> {
+    let snapshot_version_str = unpacked_snapshots_dir_and_version.snapshot_version.trim();
```
ryoqun, line 1520: maybe `.trim()` could be moved here?: https://github.com/solana-labs/solana/pull/18504/files#r674890169

That's to group similar code together.

ryoqun: also, sanitizing untrusted input early in the codepath is always better practice. :)

runtime/src/snapshot_utils.rs
ryoqun: also, I just noticed we're better off checking file size of `snapshot_version`? for eventual untrusted snapshot thing?

Otherwise, it could be dos vector if crafted snapshot contains a 100G zeroed snapshot_version file because we read all to memory first....

unrelated to this pr off course. so could be done as a follow up pr...

ryoqun (approved):
I'd say a masterpiece of plumbing pr ever. xD

I only left some minor comments. merge anytime without needing another review round.

runtime/src/snapshot_utils.rs
ryoqun: ref: #7167

brooksprumo: New issue created: #18841

runtime/src/snapshot_utils.rs

PR #18506, opened 2021-07-08
Add updated duplicate broadcast test

carllin:
#### Problem
1) Old retransmit test no longer causes two complete yet valid versions of duplicate slots to propagate because retransmit now forwards based on the node's position in the shuffled turbine tree, not based on what is received over the TvuForward port, which means we can no longer use that port to guarantee shreds are not retransmitted. This means if we send two different versions of a slot, most likely nodes will see a mix of both slots and mark the slot as dead.

2) In the current test, the nodes vote on every slot so it's hard to simulate a threshold failure. In particular we want to test the case where a slot on the main fork is NOT duplicate confirmed, but some descendant of that slot IS duplicate confirmed. This looks like:

```
0 -> 1 -> 2 (duplicate slot) -> 3 ->  5 (duplicate confirmed)
```

In this situation the node with the bad version of `2` will not be able to determine that it should dump and repair `2` because it marks `3` as dead before it can replay `5` and determine the fork is duplicate confirmed. In these scenarios, we will need to invoke the new ancestor repair protocol.

#### Summary of Changes
1) Fixup test to only broadcast a different version of the last shred in the block (added an extra transaction at the end of the block), maximizing the chance that nodes get two complete yet valid versions of the block.

2) Prevent retransmit from the node we want to partition to the rest of the network by making the weight of the partitioned node very small. This prevents the node from being selected for the "critical path" in turbine. Thus this node will never retransmit the second version of the shred to other nodes, and only the intended node will be corrupted

3) Purposely spin up another node that we control the voting for. This lets us simulate the situation in the problem statement above where we don't vote on every slot, and only certain slots reach duplicate confirmed instead of every slot reaching duplicate confirmed

Fixes #

### DIFF
--- a/core/src/broadcast_stage.rs
+++ b/core/src/broadcast_stage.rs
@@ -1,8 +1,9 @@
 //! A stage to broadcast data from a leader node to validators
 #![allow(clippy::rc_buffer)]
 use self::{
-    broadcast_duplicates_run::BroadcastDuplicatesRun,
-    broadcast_fake_shreds_run::BroadcastFakeShredsRun, broadcast_metrics::*,
+    broadcast_duplicates_run::{BroadcastDuplicatesConfig, BroadcastDuplicatesRun},
+    broadcast_fake_shreds_run::BroadcastFakeShredsRun,
+    broadcast_metrics::*,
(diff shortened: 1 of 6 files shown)


jstarry:
@carllin is there any particular part of this test that you would like extra eyes on?

PR #18528, opened 2021-07-08
cost model nits

jackcmay:
#### Problem

No problem but some nits

#### Summary of Changes

- don't pass a reference to u64
- tuples are hard to follow if not used locally, use struct
- auto-drop

Fixes #

### DIFF
--- a/core/src/cost_model.rs
+++ b/core/src/cost_model.rs
@@ -92,7 +92,7 @@ impl CostModel {
 
     pub fn initialize_cost_table(&mut self, cost_table: &[(Pubkey, u64)]) {
         for (program_id, cost) in cost_table {
-            match self.upsert_instruction_cost(program_id, cost) {
+            match self.upsert_instruction_cost(program_id, *cost) {
                 Ok(c) => {
                     debug!(
                         "initiating cost table, instruction {:?} has cost {}",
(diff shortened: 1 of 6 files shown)


core/src/cost_model.rs
```diff
@@ -147,7 +147,7 @@ impl CostModel {
     pub fn upsert_instruction_cost(
         &mut self,
         program_key: &Pubkey,
-        cost: &u64,
+        cost: u64,
```
taozhu-chicago, line 150: Makes perfect sense - no need of ref on u64. Did that just for cosmetic reason - so all parameters are in same "fashion". Bit silly :) All for removing ref for better practice.

core/src/cost_update_service.rs
```diff
@@ -135,26 +135,27 @@ impl CostUpdateService {
...
+                    Err(err) => {
+                        debug!(
                         "after replayed into bank, instruction {:?} failed to update cost, err: {}",
                         program_id, err
                     );
+                    }
                 }
             }
         }
-        drop(cost_model_mutable);
```
taozhu-chicago, line 157: Yea, I heard it few times to avoid explicit drop call. New to rust, still carries a lot of C++ habits, where being explicit is often preferred. But I am happy to adapt new style.

runtime/src/message_processor.rs
```diff
@@ -56,6 +56,12 @@ impl Executors {
     }
 }
 
+#[derive(Default, Debug)]
+pub struct ProgramStats {
+    pub accumulated_us: u64,
+    pub num_invocations: u32,
+}
+
```
taozhu-chicago, line 64: totally agree - using tuple just another evidence of me lazy  🤣

taozhu-chicago (approved):
All valid points! thanks for pointing them out, will keep them in my "rust good practice" list

PR #18254, opened 2021-06-27
web3.js - more esm friendly

Niputi:
#### Problem

currently esm browser file includes node-fetch which requires multiple polyfills to use

#### Summary of Changes

changed `node-fetch` to `ky-universal`
changed `url` module usage to `new URL` native js as the used `url` functions were marked as deprecated

this change requires node.js version 14 according to ky-universal

#### Fixes 

vite.js issue which can be observed in https://github.com/NorbertBodziony/vite-test as it requires several polyfills to work, making bundle heavier.

PR #18531, opened 2021-07-09
fix: remove buggy node polyfill plugin

jstarry:
#### Problem
The rollup node polyfills plugin injects invalid JS code into the bundled release which causes issues in downstream app bundlers.

#### Summary of Changes
Builds off of https://github.com/solana-labs/solana/pull/18254
- changed `node-fetch` to `ky-universal`
- changed url module usage to new URL native js as the used url functions were marked as deprecated
- added dependency on `asset` package to avoid the need for rollup's polyfill
- removed rollup node polyfill plugin

Fixes #

### DIFF
--- a/web3.js/package-lock.json
+++ b/web3.js/package-lock.json
@@ -11,7 +11,6 @@
       "dependencies": {
         "@babel/runtime": "^7.12.5",
         "@solana/buffer-layout": "^3.0.0",
-        "assert": "^2.0.0",
         "bn.js": "^5.0.0",
         "borsh": "^0.4.0",
         "bs58": "^4.0.1",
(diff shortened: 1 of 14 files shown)


Niputi:
couldn't you replace `assert` with if statement with new error messages?
I have just tried with `cross-fetch` instead instead of `ky-universal` and I was able to run the tests and compared package.json files which have lead me to believe mocha/something else reads the package.json wrong.
problem with `cross-fetch` is that it looks like it's not usable with esm

jstarry:
> couldn't you replace assert with if statement with new error messages?

Yeah, that works.

> currently esm browser file includes node-fetch which requires multiple polyfills to use

I just removed `node-fetch` from browser esm rollup externals config. The following `package.json` params should ensure that client projects using esm should resolve the `index.browser.esm.js` module which won't depend on `node-fetch` anymore.

`package.json`
```json
  "browser": {
    "./lib/index.cjs.js": "./lib/index.browser.esm.js",
    "./lib/index.esm.js": "./lib/index.browser.esm.js"
  },
  "main": "lib/index.cjs.js",
  "module": "lib/index.esm.js",
```

`index.browser.esm.js`
```js
var global = getGlobal();

module.exports = exports = global.fetch;

// Needed for TypeScript and Webpack.
if (global.fetch) {
	exports.default = global.fetch.bind(global);
}
```

PR #18534, opened 2021-07-09
hopefully remove intermittency from test_waitable_condvar test

jeffwashington:
#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/waitable_condvar.rs
+++ b/runtime/src/waitable_condvar.rs
@@ -30,7 +30,7 @@ pub mod tests {
     use super::*;
     use std::{
         sync::{
-            atomic::{AtomicBool, Ordering},
+            atomic::{AtomicBool, AtomicUsize, Ordering},
             Arc,
         },
         thread::Builder,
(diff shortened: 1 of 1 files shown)


runtime/src/waitable_condvar.rs
```diff
@@ -44,29 +44,44 @@ pub mod tests {
...
+        let waiters_active_ = waiters_active.clone();
+        let waiters_active__ = waiters_active.clone();
         // several passes to check re-notification and drop one of the
         let passes = 3;
         let handle = Builder::new().spawn(move || {
             for _pass in 0..passes {
                 let mut notified = false;
+                let mut timed_out = false;
+                waiters_active_.fetch_add(1, Ordering::SeqCst);
                 while cv2_.wait_timeout(Duration::from_millis(1)) {
```
carllin, line 57: Can these condition variables wake up spuriously and return false for the `timed_out()` check? https://doc.rust-lang.org/nightly/std/sync/struct.Condvar.html#method.wait_timeout

PR #18537, opened 2021-07-09
Bump prost, prost-types, and tonic

CriesofCarrots:
#### Problem
`prost-types` security advisory is ignored due to googleapi build errors -> need to bump tonic_build and regenerate rust googleapis in unison.

#### Summary of Changes
- Bump prost, prost-types and tonic versions
- Regenerate google.bigtable.v2.rs
- Accommodate generated Client interface changes
- Unignore advisory
- Also, fixup transaction_by_addr.proto error list (replaces part of #18353)

Closes #18541
Closes #18545

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1854,12 +1854,13 @@ dependencies = [
 
 [[package]]
 name = "http-body"
-version = "0.4.0"
+version = "0.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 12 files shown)


CriesofCarrots:
@mvines , could you please see if you see anything concerning with the `BigtableConnection::client()` changes? You know that method much better.
This changeset works with both emulator and real BigTable.
...The new type signatures are beastly, but I tried to make them as readable as possible.

mvines (approved):

PR #18552, opened 2021-07-09
Add voting service

sakridge:
#### Problem

Voting is slow and blocks replay of forks.

#### Summary of Changes

Push voting to another thread

Fixes #

### DIFF
--- a/core/src/replay_stage.rs
+++ b/core/src/replay_stage.rs
@@ -20,6 +20,7 @@ use crate::{
     repair_service::DuplicateSlotsResetReceiver,
     rewards_recorder_service::RewardsRecorderSender,
     unfrozen_gossip_verified_vote_hashes::UnfrozenGossipVerifiedVoteHashes,
+    voting_service::VoteOp,
     window_service::DuplicateSlotReceiver,
 };
 use solana_client::rpc_response::SlotUpdate;
@@ -316,6 +317,7 @@ impl ReplayStage {
         gossip_verified_vote_hash_receiver: GossipVerifiedVoteHashReceiver,
         cluster_slots_update_sender: ClusterSlotsUpdateSender,
         cost_update_sender: Sender<ExecuteTimings>,
+        voting_sender: Sender<VoteOp>,
     ) -> Self {
         let ReplayStageConfig {
             vote_account,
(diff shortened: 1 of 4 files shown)


behzadnouri (commented):
There are 2 paths that are writing votes to crds table: 
1. `ClusterInfo::push_vote`:
https://github.com/solana-labs/solana/blob/761de8b1a/gossip/src/cluster_info.rs#L1047-L1092
2. `ClusterInfo::refresh_vote`:
https://github.com/solana-labs/solana/blob/761de8b1a/gossip/src/cluster_info.rs#L1094-L1126


Both are computing the `vote_index` under a gossip _read_ lock, releasing the lock and then writing the votes to crds table under a 2nd write lock. So technically they have a race condition with each other where both may compute the same vote-index and overwrite each other.

Last time I checked both paths are called from the same thread, and so to my knowledge we are not hitting that race condition. But this commit is adding a new thread, so making the race condition a runtime possibility.

One possibility is just to compute the vote-index under an exclusive lock and hold on to that lock until the vote is written to the crds table. There are only 32 lookups to compute vote-index, so maybe it would not be that bad.

But, can we please hold on this approach until I can finish some tests with gossip locking and see where those go? Adding a new thread adds its own subtle complexities like the race condition above. I would rather use this separate thread approach as a last resort.

sakridge:
> There are 2 paths that are writing votes to crds table:
> 
> 1. `ClusterInfo::push_vote`:
>    https://github.com/solana-labs/solana/blob/761de8b1a/gossip/src/cluster_info.rs#L1047-L1092
> 2. `ClusterInfo::refresh_vote`:
>    https://github.com/solana-labs/solana/blob/761de8b1a/gossip/src/cluster_info.rs#L1094-L1126
> 
> Both are computing the `vote_index` under a gossip _read_ lock, releasing the lock and then writing the votes to crds table under a 2nd write lock. So technically they have a race condition with each other where both may compute the same vote-index and overwrite each other.
> 
> Last time I checked both paths are called from the same thread, and so to my knowledge we are not hitting that race condition. But this commit is adding a new thread, so making the race condition a runtime possibility.
> 
> One possibility is just to compute the vote-index under an exclusive lock and hold on to that lock until the vote is written to the crds table. There are only 32 lookups to compute vote-index, so maybe it would not be that bad.
> 
> But, can we please hold on this approach until I can finish some tests with gossip locking and see where those go? Adding a new thread adds its own subtle complexities like the race condition above. I would rather use this separate thread approach as a last resort.

Yea, let’s see the results of the locking optimization.

couldn’t we just use the same thread to do both the writes? They would be ordered by the channel, right?

core/src/replay_stage.rs
```diff
@@ -81,6 +81,16 @@ pub enum HeaviestForkFailures {
...
+pub enum VoteOp {
+    PushVote(Vec<Slot>),
+    RefreshVote(Slot),
+}
+
+pub struct VoteInfo {
+    pub vote_tx: Transaction,
+    pub op: VoteOp,
+}
+
```
behzadnouri, line 93: I think a single enum would be simpler:
```
pub enum VoteOp {
    Push { vote: Transaction, tower: Vec<Slot> },
    Refresh { vote: Transaction, slot: Slot },
}
```
it will also be more explicit what `Vec<Slot>` for example is.

also, shouldn't these type definitions go in `core/src/voting_service.rs`?

core/src/voting_service.rs
```diff
@@ -0,0 +1,54 @@
...
+            .spawn(move || {
+                for vote_info in vote_receiver.iter() {
+                    Self::push_vote(&cluster_info, &poh_recorder, vote_info);
+                }
+            })
+            .unwrap();
+        Self { thread_hdl }
+    }
+
+    pub fn push_vote(
```
behzadnouri, line 30: This `push_vote` name is a bit confusing, because then there is another `PushVote` and also a `RefreshVote`. Maybe rename this `handle_vote` or something.

behzadnouri (approved):
This LGTM aside from the fact that it is adding a new thread, and I would prefer exhaust all optimaztions before adding new threads. Your call to merge this now or wait until gossip lock reworks are complete, and see if this is still needed.

PR #18560, opened 2021-07-09
remove race condition from SharedBuffer's bg reader

jeffwashington:
#### Summary of Changes

Fixes #


jeffwashington:
less clean solution than refactoring the problem away.

PR #18563, opened 2021-07-09
refactor SharedBuffer to eliminate Arc refcount shutdown issues

jeffwashington:
#### Problem
previous code allowed a bg thread to try to join its own handle during drop of an object behind an arc.
#### Summary of Changes
Rework code path so bg thread does not have a refcount to the object whose refcount is responsible for joining the thread.
Fixes #


jeffwashington:
This avoids a failure that can occur if the bg thread is the last one holding onto the arc reference and on Drop tries to join the bg thread from the bg thread. An intermittent test failure illustrates this.

sakridge (approved):

PR #18565, opened 2021-07-09
Add test_bank_forks_incremental_snapshot()

brooksprumo:
This commit builds on PR #18504 by adding a test to `core/tests/snapshot.rs` for Incremental Snapshots. The test adds banks to bank forks in a loop and takes both full snapshots and incremental snapshots at intervals, and validates they are rebuild-able.

For background info about Incremental Snapshots, see #17088.

### DIFF
--- a/core/tests/snapshots.rs
+++ b/core/tests/snapshots.rs
@@ -29,6 +29,11 @@ macro_rules! DEFINE_SNAPSHOT_VERSION_PARAMETERIZED_TEST_FUNCTIONS {
             fn test_slots_to_snapshot() {
                 run_test_slots_to_snapshot(SNAPSHOT_VERSION, CLUSTER_TYPE)
             }
+
+            #[test]
+            fn test_bank_forks_incremental_snapshot_n() {
+                run_test_bank_forks_incremental_snapshot_n(SNAPSHOT_VERSION, CLUSTER_TYPE)
+            }
         }
     };
 }
@@ -39,6 +44,7 @@ mod tests {
     use crossbeam_channel::unbounded;
     use fs_extra::dir::CopyOptions;
     use itertools::Itertools;
+    use log::{info, trace};
     use solana_core::snapshot_packager_service::{PendingSnapshotPackage, SnapshotPackagerService};
     use solana_gossip::{cluster_info::ClusterInfo, contact_info::ContactInfo};
     use solana_runtime::{
@@ -57,7 +63,7 @@ mod tests {
     use solana_sdk::{
         clock::Slot,
         genesis_config::{ClusterType, GenesisConfig},
-        hash::hashv,
+        hash::{hashv, Hash},
         pubkey::Pubkey,
         signature::{Keypair, Signer},
         system_transaction,
@@ -66,6 +72,7 @@ mod tests {
     use std::{
         collections::HashSet,
         fs,
+        io::{Error, ErrorKind},
         path::PathBuf,
         sync::{
             atomic::{AtomicBool, Ordering},
@@ -83,8 +90,8 @@ mod tests {
 
     struct SnapshotTestConfig {
         accounts_dir: TempDir,
-        snapshot_dir: TempDir,
-        _snapshot_output_path: TempDir,
+        bank_snapshots_dir: TempDir,
+        snapshot_archives_dir: TempDir,
         snapshot_config: SnapshotConfig,
         bank_forks: BankForks,
         genesis_config_info: GenesisConfigInfo,
@@ -94,11 +101,12 @@ mod tests {
         fn new(
             snapshot_version: SnapshotVersion,
             cluster_type: ClusterType,
-            snapshot_interval_slots: u64,
+            accounts_hash_interval_slots: Slot,
+            snapshot_interval_slots: Slot,
         ) -> SnapshotTestConfig {
             let accounts_dir = TempDir::new().unwrap();
-            let snapshot_dir = TempDir::new().unwrap();
-            let snapshot_output_path = TempDir::new().unwrap();
+            let bank_snapshots_dir = TempDir::new().unwrap();
+            let snapshot_archives_dir = TempDir::new().unwrap();
             let mut genesis_config_info = create_genesis_config(10_000);
             genesis_config_info.genesis_config.cluster_type = cluster_type;
             let bank0 = Bank::new_with_paths(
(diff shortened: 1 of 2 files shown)


brooksprumo:
@ryoqun @carllin @lijunwangs Here's the next PR for incremental snapshots. This one is just a test, and should be _much_ easier to review. Thanks in advance!

core/tests/snapshots.rs
```diff
@@ -610,4 +625,244 @@ mod tests {
...
+        let incremental_snapshot_package = snapshot_utils::process_accounts_package_pre(
+            incremental_snapshot_package,
+            Some(bank.get_thread_pool()),
+            Some(incremental_snapshot_base_slot),
+        );
+        snapshot_utils::archive_snapshot_package(
+            &incremental_snapshot_package,
+            snapshot_config.maximum_snapshots_to_retain,
+        )
+    }
```
carllin, line 795: hmm is there a reason this can't be encapsulated within a single call to the `bank_to_incremental_snapshot_archive()` function? Would prefer to call/test the top-level function instead of replicating that logic

core/tests/snapshots.rs
```diff
@@ -610,4 +625,244 @@ mod tests {
...
+            Some(incremental_snapshot_archive_info.path()),
+            snapshot_config.archive_format,
+            genesis_config,
+            None,
+            None,
+            AccountSecondaryIndexes::default(),
+            false,
+            None,
+            accounts_db::AccountShrinkThreshold::default(),
+            false,
```
carllin, line 855: This flag to not check the hash was introduced here: https://github.com/solana-labs/solana/pull/17939/files#diff-b5dead0a38e5f63447595bbeda1428d3aa851d33954343bb1bc38b4b00add5fdR149, not sure why it's not safe to check the hash here. 

Maybe the hash is verified elsewhere already? Would be good to check the accounts state is being verified to match the snapshot hash somewhere in the test.

carllin: heh, looks like in the normal path, the hash is verified *after*  the call to `bank_from_snapshot_archives()` https://github.com/solana-labs/solana/blob/86bda02fc4116660d779654971b1a83471be0bea/ledger/src/bank_forks_utils.rs#L149-L171

core/tests/snapshots.rs
brooksprumo: @carllin Yeah, I would prefer that too. I did it this way for a few reasons. Firstly, that's how the existing tests in this file are written w.r.t. full snapshots. Second, and more importantly, the `bank_to_xxx_snapshot_archive()` is _not_ the function that gets called during regular node operation. This test is sort of simulating the regular operation of going through AccountsBackgroundService and the calls it would make. And since this is an integration test, I like having that path tested too.

Some additional info, `bank_to_xxx_snapshot_archive()` is only called (1) when the node first starts up (during validator `new_banks_from_ledger()`, and (2) in ledger-tool when calling it with "create-snapshot". So the `bank_to_xxx_snapshot_archive()` function is not the commonly-used one for making snapshots (even though it's the nicer one to use).

More info: these functions getting called _are_ top-level functions (in the sense that they are `pub`), because of the Second reason above. The tests in snapshot_utils kind of test these functions, but more in an unit-test manor. The simulated AccountsBackgroundService is really the additional benefit here.

With that, are there still changes you'd like?

core/tests/snapshots.rs
brooksprumo: @carllin Sure thing. Can I break this out into a separate PR/issue instead of fixing inside this PR?

core/tests/snapshots.rs
carllin: Hmmm, from my understanding, `bank_to_xxx_snapshot_archive` reflects the path of the regular operation through `AccountsBackgroundService` (but I could be missing something).

The call stacks look rather similar:

```
make_incremental_snapshot_archive
get_bank_snapshots
package_incremental_snapshot
process_accounts_package_pre
archive_snapshot_package

bank_to_incremental_snapshot_archive
get_incremental_snapshot_storages
add_bank_snapshot
package_incremental_snapshot
process_accounts_package_pre
archive_snapshot_package
```
If these are close approximations of each other, I would prefer we coalesce this logic so that if we modify any of these intermediary steps in the snapshotting process, we only have to modify one function rather than multiple

core/tests/snapshots.rs
carllin: If extracting the above logic into a function and calling it from the test works, I think it makes sense to include in this PR, as testing the accounts hash seems important to the snapshot tests.

If somethings broken/hash doesn't verify, then a separate PR makes sense to me.

core/tests/snapshots.rs
brooksprumo: Regular node operation for the snapshots is as follows:

- `BankForks::set_root()` will call `AbsRequestSender::send_snapshot_request()` if the bank's slot is a multiple of the `accounts_hash_interval_slots`. This puts the (bank) snapshot request into the `SnapshotRequest` channel.
- `AccountsBackgroundService`'s main loop calls `AbsRequestHandler::handle_snapshot_request()`, which then calls `SnapshotRequestHandler::handle_snapshot_request()`
    - if there's a snapshot request in the channel, it grabs it, calls `Bank::clean_accounts()`, then calls `snapshot_utils::snapshot_bank()` (which creates a bank snapshot by serializing the bank to disk)
        - inside `snapshot_utils::snapshot_bank()`, it calls `add_bank_snapshot()` to do the bank serialization, and then it makes a snapshot package (`AccountsPackagePre`) and puts that in the channel for snapshot packages
- `AccountsHashVerifier`'s main loop reads from the snapshot packages channel, then
    - calls `snapshot_utils::process_accounts_package_pre()`
    - gets the hash from the processed package ^^ and checks it against the trusted validators
    - if the block height is a multiple of the `snapshot_interval_slots` then it puts this snapshot package into the `PendingSnapshotPackage` (`pub type PendingSnapshotPackage = Arc<Mutex<Option<AccountsPackage>>>`)
- Lastly, `SnapshotPackagerService`'s main loop takes the pending snapshot package (if there is one) and archives the snapshot package via `snapshot_utils::archive_snapshot_package()`

Since there are four separate threads handling the different parts of the snapshot process, it's not possible to just call `bank_to_xxx_snapshot()` during normal operation. So the `bank_to_xxx_snapshot()` functions are basically just wrappers around the other functions above, for the two times that making a snapshot archive can happen sequentially/blocking (1. at startup, 2. ledger-tool "create-snapshot").

I'm definitely in favor of coalescing the logic, I just am not sure how to do that here. I'm guessing I'm missing something though; I'd like to simplify this process if at all possible. Thank you in advance for the additional info!

brooksprumo: Carl and I talked. I've created a new issue to track this request: #18972

core/tests/snapshots.rs
brooksprumo: Carl and I talked. I've created a new issue to track this request: #18973.

core/tests/snapshots.rs
brooksprumo: OK, done!

brooksprumo:
@carllin I refactored the functions for package, process, and archive as per our call (and issue #18972). I'll handle the "verify" stuff (issue #18973) in a separate PR. So this PR should be good-to-go!

carllin (approved):
Thanks, looks much cleaner!

PR #18566, opened 2021-07-09
Eliminate doc warnings and fix some markdown

brson:
#### Problem

`cargo doc` emits warnings

#### Summary of Changes

This fixes a lot of docs that look like markdown links such that they are not interpreted as links. This happens especially in the "account references" sections of docs where the requirements of program accounts are listed. In these cases I've surrounded them with code spans, like "\`[write]\`", to match prevailing conventions in the spl.

It also turns bare URLs into "autolinks" by surrounding them with angle brackets.

I couldn't resist fixing some of the surrounding incorrect markdown in various ways so that it renders correctly, and follows rustdoc conventions more closely.

I also have a corresponding patch for the spl here: https://github.com/solana-labs/solana-program-library/pull/2058

### DIFF
--- a/sdk/program/src/loader_upgradeable_instruction.rs
+++ b/sdk/program/src/loader_upgradeable_instruction.rs
@@ -46,10 +46,12 @@ pub enum UpgradeableLoaderInstruction {
     /// The ProgramData address is derived from the Program account's address as
     /// follows:
     ///
-    /// `let (program_data_address, _) = Pubkey::find_program_address(
+    /// ```
+    /// let (program_data_address, _) = Pubkey::find_program_address(
     ///      &[program_address],
     ///      &bpf_loader_upgradeable::id()
-    ///  );`
+    ///  );
+    /// ```
     ///
     /// The `DeployWithMaxDataLen` instruction does not require the ProgramData
     /// account be a signer and therefore MUST be included within the same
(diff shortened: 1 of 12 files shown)


sdk/program/src/loader_upgradeable_instruction.rs
```diff
@@ -46,10 +46,12 @@ pub enum UpgradeableLoaderInstruction {
     /// The ProgramData address is derived from the Program account's address as
     /// follows:
     ///
-    /// `let (program_data_address, _) = Pubkey::find_program_address(
+    /// ```
```
jstarry, line 49:
```suggestion
    /// ```no_run
```
Looks like this PR will need a smattering of these until we invest time into nicer doc tests

jstarry: @brson would you like us to pick this change up from where you left off?

jstarry: Really appreciate the clean up you've done so far btw 😄

brson: @jstarry I pushed a commit that fixes the test so it works. Thanks for the review.

brson:
Rebased to fix merge conflicts.

brson:
Rebased to fix merge conflicts.

brson:
Ok, this is fixed up. @jstarry can you give this another look? And maybe https://github.com/solana-labs/solana-program-library/pull/2058 too?

jstarry (approved):
looks awesome, thank you so much!

PR #18575, opened 2021-07-10
Change ShredIndex to use BitVector for underlying storage

steviez:
#### Problem
The blockstore uses ShredIndex structures to track whether shreds have been received or not. The elements that we store are the indices (u64) themselves so we can use a BitVector instead of a BTreeSet.

Here are some results for running the benchmark (on my machine) that is included as part of this change; I ran with size of 500 and 5000 (these sizes would be the slot size in shreds). The reported execution times would be per slot.

**Current tip of master**
```
Results for 500
test bench_shred_index_is_present  ... bench:         743 ns/iter (+/- 62)
test bench_shred_index_set_present ... bench:      31,303 ns/iter (+/- 5,349)
Results for 5000
test bench_shred_index_is_present  ... bench:       7,282 ns/iter (+/- 612)
test bench_shred_index_set_present ... bench:     460,156 ns/iter (+/- 19,850)
```
**this branch**
```
Results for 500
test bench_shred_index_is_present  ... bench:         801 ns/iter (+/- 65)
test bench_shred_index_set_present ... bench:       2,860 ns/iter (+/- 184)
Results for 5000
test bench_shred_index_is_present  ... bench:       8,850 ns/iter (+/- 303)
test bench_shred_index_set_present ... bench:      28,112 ns/iter (+/- 4,920)
```
Fairly comparable on `is_present()`; 10x speedup for `set_present()` at slot size of 500 (and more pronounced at 5000), although it isn't a large chunk of time to start with. This should save us some space too though; with a a BTreeSet<u64> and assuming 500 shreds / slot, that would be 4k bytes / slot for each index, 8k total (data and coding). Hypothetically, should be some time savings in serialize/deserialize too.

### Upgrade / Downgrade
This is a breaking change since the ShredIndex gets serialized out of Rocks. Per some suggestions in this PR and conversation in Discord, I have implemented the following strategy:
- Introduce a second column family that uses BitVec, keeping the original BTreeSet column around
- Make a backport PR that stubs new column into older releases (this still creates a break point, but hopefully less painful)
   - Older validator binary would open DB with new column (RocksDB requires specifying all columns at open), but never use that column (except in purge maybe)

In the future once this change is reflected on a given network ...
- Strip out the original (BTreeSet) CF and only use the new (BitVec) one. Accomplish this with a feature switch (disabled state of feature writes both columns, enabled state of feature will only write new column)

### DIFF
--- a/ledger/src/blockstore_meta.rs
+++ b/ledger/src/blockstore_meta.rs
@@ -436,7 +436,7 @@ mod test {
         assert_eq!(index.num_shreds(), 0);
 
         // Check boundary conditions; start is inclusive, end is inclusive
-        index.set_many_present((5..10 as u64).zip(repeat(true)));
+        index.set_many_present((5..10_u64).zip(repeat(true)));
         assert_eq!(index.present_in_bounds(0..5), 0);
         assert_eq!(index.present_in_bounds(2..7), 2);
         assert_eq!(index.present_in_bounds(5..10), 5);
--- a/ledger/src/blockstore_meta.rs
+++ b/ledger/src/blockstore_meta.rs
@@ -227,8 +227,9 @@ impl ErasureMeta {
     pub fn status(&self, index: &Index) -> ErasureMetaStatus {
         use ErasureMetaStatus::*;
 
-        let coding_indices = self.set_index..self.set_index + self.config.num_coding() as u64;
-        let num_coding = index.coding().present_in_bounds(coding_indices);
+        let num_coding = index
+            .coding()
+            .present_in_bounds(self.set_index..self.set_index + self.config.num_coding() as u64);
         let num_data = index
             .data()
             .present_in_bounds(self.set_index..self.set_index + self.config.num_data() as u64);
@@ -290,6 +291,30 @@ mod test {
     use super::*;
     use rand::{seq::SliceRandom, thread_rng};
     use std::iter::repeat;
+
+    #[test]
+    fn test_shred_index() {
+        let mut index = ShredIndex::default();
+        assert!(!index.is_present(new_idx));
+        assert_eq!(index.num_shreds(), 0);
+
+        let new_idx = 2;
+        index.set_present(new_idx, true);
+        assert!(index.is_present(new_idx));
+        assert_eq!(index.num_shreds(), 1);
+
+        index.set_present(new_idx, false);
+        assert!(!index.is_present(new_idx));
+        assert_eq!(index.num_shreds(), 0);
+
+        // Check boundary conditions; start is inclusive, end is inclusive
+        index.set_many_present((5..10 as u64).zip(repeat(true)));
+        assert_eq!(index.present_in_bounds(0..5), 0);
+        assert_eq!(index.present_in_bounds(2..7), 2);
+        assert_eq!(index.present_in_bounds(5..10), 5);
+        assert_eq!(index.present_in_bounds(7..12), 3);
+        assert_eq!(index.present_in_bounds(10..15), 0);
+    }
 
     #[test]
     fn test_erasure_meta_status() {
--- a/ledger/src/blockstore_meta.rs
+++ b/ledger/src/blockstore_meta.rs
@@ -1,7 +1,8 @@
 use crate::erasure::ErasureConfig;
+use bv::BitVec;
 use serde::{Deserialize, Serialize};
 use solana_sdk::{clock::Slot, hash::Hash};
-use std::{collections::BTreeSet, ops::RangeBounds};
+use std::{cmp, collections::BTreeSet, ops::Range};
 
 #[derive(Clone, Debug, Default, Deserialize, Serialize, Eq, PartialEq)]
 // The Meta column family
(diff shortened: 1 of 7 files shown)


ledger/src/blockstore_meta.rs
```diff
@@ -125,10 +144,6 @@ impl ShredIndex {
             self.set_present(idx, present);
         }
     }
-
-    pub fn largest(&self) -> Option<u64> {
```
steviez, line 129: This function wasn't used anywhere so I deleted it. If we did need to keep this function around for some reason, BitVector implementation wouldn't be super pretty. We'd have to do something like this in in `set_present()`:
```
            ...
            self.index.set(index, presence);
            if presence {
                self.num_present += 1;
                // Compute new largest
                largest = cmp::max(largest, index);
            } else {
                self.num_present -= 1;
                // Scan for new largest, starting at index and running backwards
            }
```

ledger/src/blockstore_meta.rs
```diff
@@ -101,22 +104,38 @@ impl Index {
...
-        self.index.len()
+        self.num_present
     }
 
-    pub fn present_in_bounds(&self, bounds: impl RangeBounds<u64>) -> usize {
-        self.index.range(bounds).count()
+    pub fn present_in_bounds(&self, range: Range<u64>) -> usize {
+        let mut count = 0;
+        // We should only search in the overlap of the index's values and range
+        for idx in cmp::min(range.start, self.index.len())..cmp::min(range.end, self.index.len()) {
```
steviez, line 113: With the BitVector, we have to read each bit in the range. While this isn't ideal, I don't think it is a huge issue since practically speaking, we only call this function with [a range the same size of FEC sets](https://github.com/solana-labs/solana/blob/master/ledger/src/blockstore_meta.rs#L204-L207).

carllin:
> This would be a breaking change since ShredIndex gets serialized/deserialized into/out of Rocks; I'm unsure how we should go about rolling it out.

You could introduce a separate column family to hold this new index structure, and then we can deprecate and delete the old index once everyone's updated

steviez:
Thanks for the suggestions to kickstart the brainstorm @ryoqun & @carllin. Agreed that we have to have two separate cf's to avoid mixing data-structures within one cf.

> > > This would be a breaking change since ShredIndex gets serialized/deserialized into/out of Rocks; I'm unsure how we should go about rolling it out.
> > 
> > 
> > You could introduce a separate column family to hold this new index structure, and then we can deprecate and delete the old index once everyone's updated. If we need to support the downgrade path, we'll have to run a converter though
> 
> yeah, downgrade option would be nice to have...:
> 
> so alternatively,
> 
> * write to the old & new cfs (bite the 2x temporal write?...)
> 
> or
> 
> * introduce feature flag switching which to use?

Is the idea with a feature flag something like this:
```
if new_index_feature_enabled {
    // Write to new cf
} else {
    // Write to old cf
}
```
If so, I think this could get messy.
- For the upgrade case, we'd need to have logic to fallback on reading the old cf if an item isn't in new column family
- For the downgrade case where validator ran for `d` slots, all of the other cf's would be up to `slot X`, but the old index cf would only be up to `slot X-d` which would cause issues

So, I think Ryo's suggestion of writing both (and absorbing the hit) might be necessary. To sum it up, I think it'd look like this:
1) Add a new cf (something like `index2` or `shred_index`)
2) Add a feature flag to write to both the existing and the new index cf's
3) Only read from the new index cf
   - This means old cf only populated to support possible rollback situation
4) At some point in the future (once all networks running whatever release this first goes into), strip out the old cf & feature flag (only write to new cf)
5) As part of previous step, may want to check for existence of old cf; if it exists, drop it in order to remove all remnants of old cf from blockstore
6) Remove the check from previous step once change from previous step propagated through all networks

So, 1-3 would be part of this PR, 4-5 would be to clean that up, and then step 6. to cleanup the cleanup haha? Does this sound right / reasonable?

steviez:
@carllin @ryoqun - I updated the PR with logic to write both columns, not super pretty with some near-duplicate code. Some things:
- I named new column `Index2` with `ShredIndex2` sub-types; don't love it, but creative juices weren't flowing for anything better
- I commented out functions on original `Index` and `ShredIndex` objects; this made it easier to find places to update. I could uncomment these back out; I think leaving them commented out (or deleted) is the move to avoid someone unintentionally using the old CF
- One of the shortcomings of the new implementation was shown by unit test `test_should_insert_coding_shred()`. Part of this test attempts to insert a shred with `index = u32::MAX - 1`.
  - For a BTreeSet, this is fine as we just add another node in the tree. 
  - For a BitVec, this is trouble since the shred index is the same as the bit index in `BitVec`. In order to access `u32::MAX - 1`, that `BitVec` will need access to `2^32 bits == 2^29 bytes == 512 MB`. When I implemented this, I was using the below assumption as a bound. I can update the test but I'm wondering if we should more strictly enforce this in shred to avoid allowing the the `BitVec` from trying to claim all that memory
  https://github.com/solana-labs/solana/blob/b13caa14c95794aa44e2b7160731040accd87092/ledger/src/blockstore.rs#L87

PR #18579, opened 2021-07-10
Optimize `Message::is_non_loader_key` method

jstarry:
#### Problem
`Message::is_non_loader_key` is inefficient because it builds a vector of program id's which is used to check against a given public key. There is also a cache to avoid rebuilding the program id vector in the transaction processing path but the pubkey check is still done for each message key.

#### Summary of Changes
- Check keys by index rather than the key itself

Fixes #


jeffwashington:
Seems like an even better improvement than the cache I put in. ;-) I'll investigate Monday.

jeffwashington (approved):
LGTM. Good find!

PR #18581, opened 2021-07-10
makes CrdsGossipPush thread-safe

behzadnouri:
#### Problem
Working towards reducing lock contention in gossip.

#### Summary of Changes
This commit makes `CrdsGossipPush` thread safe by adding `RwLock` to inner fields. This is in preparation of making `CrdsGossip` thread-safe, and implementing more granular locking in gossip.

### DIFF
--- a/gossip/src/crds_gossip_push.rs
+++ b/gossip/src/crds_gossip_push.rs
@@ -20,6 +20,7 @@ use {
     },
     bincode::serialized_size,
     indexmap::map::IndexMap,
+    itertools::Itertools,
     lru::LruCache,
     rand::{seq::SliceRandom, Rng},
     solana_runtime::bloom::{AtomicBloom, Bloom},
@@ -27,7 +28,12 @@ use {
     std::{
         cmp,
         collections::{HashMap, HashSet},
-        ops::RangeBounds,
+        iter::repeat,
+        ops::{DerefMut, RangeBounds},
+        sync::{
+            atomic::{AtomicUsize, Ordering},
+            Mutex, RwLock,
+        },
     },
 };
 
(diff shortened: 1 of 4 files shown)


carllin (approved):

gossip/src/crds_gossip_push.rs
```diff
@@ -164,37 +200,59 @@ impl CrdsGossipPush {
...
+                let origin = value.pubkey();
+                match crds.insert(value, now) {
+                    Ok(()) => Ok(origin),
+                    Err(_) => {
+                        self.num_old.fetch_add(1, Ordering::Relaxed);
+                        Err(CrdsGossipError::PushMessageOldVersion)
+                    }
+                }
+            })
+            .collect()
```
carllin, line 244: might be missing something, but can this be folded into the above collect via something like:

```
                   peers
                        .entry(*from)
                        .and_modify(|(_pruned, timestamp)| *timestamp = now)
                        .or_insert((/*pruned:*/ false, now));
                    let origin = value.pubkey();
                    match crds.insert(value, now) {
                        Ok(()) => Ok(origin),
                        Err(_) => {
                            self.num_old.fetch_add(1, Ordering::Relaxed);
                            Err(CrdsGossipError::PushMessageOldVersion)
                        }
                    }
```

carllin, line 244: might be missing something, but can this be coalesced together with something like

```
peers
    .entry(*from)
    .and_modify(|(_pruned, timestamp)| *timestamp = now)
    .or_insert((/*pruned:*/ false, now));
let origin = value.pubkey();
match crds.insert(value, now) {
    Ok(()) => Ok(origin),
    Err(_) => {
        self.num_old.fetch_add(1, Ordering::Relaxed);
        Err(CrdsGossipError::PushMessageOldVersion)
    }
}
```

behzadnouri: I have a follow up change that wraps crds in a RwLock. So here  we need to write-lock crds first, and this is done so to avoid locking crds and received_cache at the same time.

PR #18582, opened 2021-07-11
Generate a script to rerun the failed sub-command in cargo-test-bpf

dmakarov:


dmakarov:
@mvines would it be ok to merge this?  I often find it useful for debugging to have a script with the same environment that cargo test-bpf executes tests in.

mvines:
> @mvines would it be ok to merge this? I often find it useful for debugging to have a script with the same environment that cargo test-bpf executes tests in.

How about adding a command-line option to enable the creation of this script?

dmakarov:
good idea. added.

sdk/cargo-test-bpf/src/main.rs
```diff
@@ -43,13 +47,13 @@ impl Default for Config {
...
 
-fn spawn<I, S>(program: &Path, args: I)
+fn spawn<I, S>(program: &Path, args: I, generate_rerun: bool)
 where
     I: IntoIterator<Item = S>,
     S: AsRef<OsStr>,
 {
     let args = args.into_iter().collect::<Vec<_>>();
-    print!("Running: {}", program.display());
+    print!("CARGO-TEST-BPF Running: {}", program.display());
```
mvines, line 56: Can this hunk be reverted?  It seems a little yelly

sdk/cargo-test-bpf/src/main.rs
```diff
@@ -65,6 +69,22 @@ where
 
     let exit_status = child.wait().expect("failed to wait on child");
     if !exit_status.success() {
+        eprintln!("CARGO-TEST-BPF exited on command execution failure.");
```
mvines, line 72: This also seem too yelly and verbose, wrap behind the verbose flag?

sdk/cargo-test-bpf/src/main.rs
```diff
@@ -65,6 +69,22 @@ where
...
+        if !generate_rerun {
+            exit(1);
+        }
+        let rerun = File::create("rerun-script.sh").unwrap();
```
mvines, line 76: "rerun" doesn't feel like the right nomenclature here. "spawn" perhaps, or "child"? 
Also we should namespace the script since it gets dumped into the cwd, so perhaps call it `cargo-test-bpf-child-script.sh`.  Even better maybe `format!("cargo-test-bpf-child-script-{}.sh", program.display())` (but strip the path) since `spawn()` invokes multiple programs.

There's also a duplicate `spawn()` in `cargo-build-bpf` that could benefit from this same feature in the future perhaps

sdk/cargo-test-bpf/src/main.rs
```diff
@@ -239,6 +263,12 @@ fn main() {
                 .takes_value(false)
                 .help("Run without accessing the network"),
         )
+        .arg(
+            Arg::with_name("generate_rerun")
+                .long("generate-rerun")
```
mvines, line 268: `--generate-child-script-on-failure` perhaps?

sdk/cargo-test-bpf/src/main.rs
dmakarov: I added this because cargo  and cargo-build-bpf output their own "Running: " message, and it's annoying to figure out which of the several child processes in a complicated build system is running the printed command, but ok, I'll revert it.

mvines: I see. I think the yelling put me off. Childhood flashbacks perhaps, how about this instead:
```rust
print!("cargo-test-bpf child: {}", program.display());
```

dmakarov: caps help me to quickly see in a stream of text on terminal the message i'm looking for, but I understand the sentiment :-) i'll lower the case

mvines: Welp that an arms race and everybody ends up yelling in the end :)

sdk/cargo-test-bpf/src/main.rs
dmakarov: as this is relevant mostly for the message about the rerun script, I'll output it only when the script is about to be generated too. with lower case.

dmakarov:
Made changes according to review comments.  Thank you for reviewing.

dmakarov:
Also added the same functionality to cargo-build-bpf.

mvines (approved):

PR #18585, opened 2021-07-11
Record vote account commission with voting/staking rewards and surface in RPC

mvines:
#### Problem
It's not trivial to detect if the validator you staked with is defrauding you by raising their commission immediately before an epoch boundary, and restoring it as soon as possible in the new epoch.

#### Summary of Changes
Record the vote account's commission in both voting and staking rewards, and surface the commission in all RPC methods that provide reward information.  Also include the _epoch_ commission in the output of `solana vote-account --with-rewards` and `solana stake-account --with-rewards`

Freebie: The Everstake vote account is no longer decoded more than forty two thousand times on an epoch boundary!

### DIFF
--- a/programs/stake/src/stake_state.rs
+++ b/programs/stake/src/stake_state.rs
@@ -1029,14 +1029,13 @@ pub fn redeem_rewards(
     rewarded_epoch: Epoch,
     stake_account: &mut AccountSharedData,
     vote_account: &mut AccountSharedData,
+    vote_state: &VoteState,
     point_value: &PointValue,
     stake_history: Option<&StakeHistory>,
     inflation_point_calc_tracer: &mut Option<impl FnMut(&InflationPointCalculationEvent)>,
     fix_stake_deactivate: bool,
 ) -> Result<(u64, u64), InstructionError> {
     if let StakeState::Stake(meta, mut stake) = stake_account.state()? {
-        let vote_state: VoteState =
-            StateMut::<VoteStateVersions>::state(vote_account)?.convert_to_current();
         if let Some(inflation_point_calc_tracer) = inflation_point_calc_tracer {
             inflation_point_calc_tracer(
                 &InflationPointCalculationEvent::EffectiveStakeAtRewardedEpoch(stake.stake(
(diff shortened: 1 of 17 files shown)


runtime/src/bank.rs
```diff
@@ -830,8 +835,9 @@ pub trait DropCallback: fmt::Debug {
 #[derive(Debug, PartialEq, Serialize, Deserialize, AbiExample, Clone, Copy)]
 pub struct RewardInfo {
     pub reward_type: RewardType,
-    pub lamports: i64,     // Reward amount
-    pub post_balance: u64, // Account balance in lamports after `lamports` was applied
+    pub lamports: i64,          // Reward amount
+    pub post_balance: u64,      // Account balance in lamports after `lamports` was applied
+    pub commission: Option<u8>, // Vote account commission when the reward was credited, only present for voting and staking rewards
```
mvines, line 840: Start from here!

mvines:
@CriesofCarrots - can you please look through this.  It's 99% plumbing plus a little optimization in the bank to decode the vote state only once per epoch boundary when computing rewards, so that the bank has access to the vote state commission without decoding the vote state **once more**

programs/stake/src/stake_state.rs
```diff
@@ -1029,14 +1029,13 @@ pub fn redeem_rewards(
...
     vote_account: &mut AccountSharedData,
+    vote_state: &VoteState,
     point_value: &PointValue,
     stake_history: Option<&StakeHistory>,
     inflation_point_calc_tracer: &mut Option<impl FnMut(&InflationPointCalculationEvent)>,
     fix_stake_deactivate: bool,
 ) -> Result<(u64, u64), InstructionError> {
     if let StakeState::Stake(meta, mut stake) = stake_account.state()? {
-        let vote_state: VoteState =
-            StateMut::<VoteStateVersions>::state(vote_account)?.convert_to_current();
```
CriesofCarrots, line 1039: 🎉

CriesofCarrots (approved):
Nice, lgtm! Sorry for delay; I wanted to sanity-check the Rewards/Blockstore/Bigtable backward compatibility.

PR #18586, opened 2021-07-11
Sysvar::to_account_info() is dead code, remove it

mvines:


mvines:
> error: could not compile `serum_dex`

Welp, it's used downstream.  ⚰️

PR #18590, opened 2021-07-11
Add get_clock to BanksClient

fanatid:
#### Problem

No simple way to get `Clock` from `BanksClient`.

#### Summary of Changes

Add `BanksClient::get_clock`


mvines:
LGTM, thanks!

CriesofCarrots:
@mvines , should we add this method to RpcClient?

mvines:
> @mvines , should we add this method to RpcClient?

Not opposed.  Also don't feel too strongly about it since it's not too much code to do it outside of RpcClient.  If somebody ever wants it, adding seems fine to me

PR #18592, opened 2021-07-11
Remove outdated program verification

jstarry:
#### Problem
The program account is now always loaded from the invoke context and therefore cannot be be modified by the caller or callee program unless passed as an instruction account input. Despite not being modifiable, its entire account state is checked for each cross program invocation.

#### Summary of Changes
- Remove unnecessary verification step for the program account

Fixes #


jstarry:
The verification was made obsolete from this change: https://github.com/solana-labs/solana/pull/14574 and it appears that native invocation has never needed program verification due to the implementation of this change: https://github.com/solana-labs/solana/pull/14193

jackcmay (approved):

PR #18604, opened 2021-07-12
Add method `id` to SysbarId trait

fanatid:
#### Problem

Yesterday I submitted https://github.com/solana-labs/solana/pull/18590 with the new method `BanksClient::get_clock` but today realize that instead of new method for each `Sysvar` it can be generic, like:
```rust
    pub fn get_sysvar<T: Sysvar>(&mut self) -> impl Future<Output = io::Result<T>> + '_ {
        self.get_account(T::id()).map(|result| {
            let sysvar = result?
                .ok_or_else(|| io::Error::new(io::ErrorKind::Other, "Sysvar not present"))?;
            from_account::<T, _>(&sysvar).ok_or_else(|| {
                io::Error::new(io::ErrorKind::Other, "Failed to deserialize sysvar")
            })
        })
    }
```

but `SysvarId` have only `check_id`, not `id`.

#### Summary of Changes

Add `SysvarId::id` method.

### DIFF
--- a/sdk/program/src/sysvar/mod.rs
+++ b/sdk/program/src/sysvar/mod.rs
@@ -32,6 +32,10 @@ macro_rules! declare_sysvar_id(
         $crate::declare_id!($name);
 
         impl $crate::sysvar::SysvarId for $type {
+            fn id() -> $crate::pubkey::Pubkey {
+                id()
+            }
+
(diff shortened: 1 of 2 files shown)


jstarry (approved):
lgtm, thanks!

Issue #18277, opened 2021-06-28
web3: return custom transaction error with logs


beautyfree:
It will be great to have logs of failed transactions for ability to showing true errors and instant notifications with errors

Some places that could be modificated:
https://github.com/solana-labs/solana/blob/1f1e54b9d8ee8639c9f55df5008b1a6f2e25d8df/web3.js/src/connection.ts#L3540
https://github.com/solana-labs/solana/blob/f228d113c199c8a1c896cc20c942e2fe83b1046c/web3.js/src/connection.ts#L3448

jon-chuang:
Related: https://github.com/solana-labs/solana/pull/17769#issuecomment-865372427

Maintaining custom errors is the key work. I envision it as dapp devs should submit a PR to include their custom errors in explorer/web3.js. interested too to see how the anchor version of custom error conversion will play out

beautyfree:
Yeah, but this issue about other mechanic.

I think we can make custom error in lines like this:
https://github.com/solana-labs/solana/blob/1f1e54b9d8ee8639c9f55df5008b1a6f2e25d8df/web3.js/src/connection.ts#L3540

Where we can return custom Error with logs something like: return TransactionError('failed to send transaction: ' + res.error.message, logs);
And TransactionError will contains property with logs, which we can parse after simulation fails and other transactions errors

jstarry:
I think this sounds useful, would you mind opening a PR with this change?

beautyfree:
#18606

yamen:
I hate to be commenting on a closed issue, but the related commit has a `console.error` call to output the logs, which creates noise in many application usages. Console logging is not used elsewhere in this library. I'm happy to remove this but am not a current contributor and can't open an issue either.

jstarry:
Feel free to make a PR!

PR #18606, opened 2021-07-12
feat(web3): add send transaction error with logs #18277

beautyfree:


web3.js/src/errors.ts
```diff
@@ -0,0 +1,9 @@
+export class SendTransactionError extends Error {
+  logs: string | undefined;
+
+  constructor(message: string, logs?: string) {
```
jstarry, line 4: logs will either be undefined or an array of strings. Can you please update the type of `logs` everywhere to indicate that it's an array?

beautyfree: oh yeah, of course

jstarry:
Could you please fix the prettier check too?

jstarry (approved):

jstarry:
Thanks for your contribution @beautyfree!

PR #18613, opened 2021-07-12
accounts index scan starts looking in bin where bound would exist

jeffwashington:
#### Problem
Scans with binned pubkeys can be smarter about which bins they look in for matching pubkeys.
#### Summary of Changes
start looking at the bin where the start bound fits.
Fixes #


sakridge (approved):
lgtm

PR #18615, opened 2021-07-12
make CrdsGossip thread-safe

behzadnouri:
#### Problem
Working towards reducing lock contention in gossip.

#### Summary of Changes
This commit makes `CrdsGossip` thread safe by wrapping inner crds field in `RwLock`.

### DIFF
--- a/core/src/cluster_nodes.rs
+++ b/core/src/cluster_nodes.rs
@@ -296,12 +296,12 @@ mod tests {
         let cluster_info = ClusterInfo::new_with_invalid_keypair(this_node);
         {
             let now = timestamp();
-            let mut gossip = cluster_info.gossip.write().unwrap();
+            let mut gossip_crds = cluster_info.gossip.crds.write().unwrap();
             // First node is pushed to crds table by ClusterInfo constructor.
             for node in nodes.iter().skip(1) {
                 let node = CrdsData::ContactInfo(node.clone());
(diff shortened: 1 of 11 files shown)


carllin (approved):

sakridge (approved):
Nice, did it help?

behzadnouri:
Not much. Very little on the `replay-loop-voting-stats.vote_push_us` metric running on TDS.

This plot below is the write lock contention on crds table in one of the call stacks when processing incoming gossip packets.
Apparently `std::sync::RwLock` on linux is reader-preferred (pthread default), which might contribute to write lock contentions. So, I am going to also try `parking_lot` which claims fair locking policy. But I am not expecting it will make a lot of difference. So no need to wait on merging in the voting-service commit.

lock-contention-process-gossip

carllin:
Yeah, seems like all the contention would still exist around the crds table, probably need to use a concurrent hashmap to see any dramatic difference

behzadnouri:
`replay-loop-voting-stats::vote_push_us`: Left side is `parking_lot`.
replay-loop-voting-stats vote_push_us_parking_lot_16s

I also implemented a writer-preferred RwLock on top of `std::sync::RwLock`, using an extra Mutex trick and similar result:  

replay-loop-voting-stats vote_push_us_rwlock_16s

PR #18616, opened 2021-07-12
Fix caller write privileges for native cpi

jstarry:
#### Problem
Caller write privileges are not created correctly for native cpi. The `caller_write_privileges` vector should match up with `message.account_keys` because they will be indexed the same.

#### Summary of Changes
- Remove incorrect write privilege value insertion
- Match up write privileges vector with the `accounts` vector built from `message.account_keys`

Fixes #

### DIFF
--- a/runtime/src/message_processor.rs
+++ b/runtime/src/message_processor.rs
@@ -773,27 +773,28 @@ impl MessageProcessor {
             let invoke_context = invoke_context.borrow();
 
             // Translate and verify caller's data
-            let keyed_accounts = invoke_context.get_keyed_accounts()?;
-            let keyed_accounts = keyed_account_indices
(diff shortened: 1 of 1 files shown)


jstarry:
Since `native_invoke` is only called for the bpf upgradeable program and that program doesn't modify caller accounts before invoking the native create account instruction, there shouldn't be any change of behavior here.

jackcmay:
Do you have a test that highlights the issue?

jstarry:
Not yet, just running through CI first to see what pops up. Should have opened as draft

jstarry:
Hmm looks like there aren't any tests for native cpi yet beyond bpf loader integration tests, I don't have the time to put a test suite together right now so I'll write up an issue instead.

jstarry:
Issue here: https://github.com/solana-labs/solana/issues/18629

PR #18619, opened 2021-07-12
remove unnecessary generic args to make refactoring easier

jeffwashington:
#### Problem
The specifics around AccountsIndex are in flux as we move to an on-disk implementation.
#### Summary of Changes
Make it less noisy to change the implementation by removing unnecessary specificity.
Fixes #


jbiseda (approved):
lgtm

PR #18620, opened 2021-07-12
Cli: expose last valid block height

CriesofCarrots:
#### Problem
The `lastValidSlot` field returned from the `getFees` rpc has been deprecated. However, solana-cli still reports this value exclusively in non-json responses:
```
$ solana fees

Blockhash: 5yC46JNB5XYNNdfZCjDuTvDjTb7ZZueEn95JFPEgPkFf
Lamports per signature: 5000
Last valid slot: 86798918
```

#### Summary of Changes
- Update solana-cli to return last valid block height:

```
$ solana fees

Blockhash: 7BvMD73ZFVuZHJUpZyGcNc4QvWM8TS35FHpAXNpDNAVQ
Lamports per signature: 5000
Last valid block height: 7794751

// JSON response retains deprecated field for backward compatibility
$ solana fees --output json

{
  "slot": 86798889,
  "blockhash": "9EMv6dBvaUnRVHrpxbkwkeBVucT9ujNn5Kj9DiyfoyYF",
  "lamportsPerSignature": 5000,
  "lastValidSlot": null,
  "lastValidBlockHeight": 77947583
}
```
- Adds new sdk struct and client methods -> Next steps are to deprecate `RpcClient::get_recent_blockhash` methods, and update monorepo to `RpcClient::get_fees` and last_valid_block_height logic.

### DIFF
--- a/cli-output/src/cli_output.rs
+++ b/cli-output/src/cli_output.rs
@@ -1720,6 +1720,7 @@ pub struct CliFeesInner {
     pub blockhash: String,
     pub lamports_per_signature: u64,
     pub last_valid_slot: Option<Slot>,
+    pub last_valid_block_height: Option<Slot>,
 }
 
 impl QuietDisplay for CliFeesInner {}
@@ -1733,11 +1734,11 @@ impl fmt::Display for CliFeesInner {
             "Lamports per signature:",
             &self.lamports_per_signature.to_string(),
         )?;
-        let last_valid_slot = self
-            .last_valid_slot
+        let last_valid_block_height = self
+            .last_valid_block_height
             .map(|s| s.to_string())
             .unwrap_or_default();
-        writeln_name_value(f, "Last valid slot:", &last_valid_slot)
+        writeln_name_value(f, "Last valid block height:", &last_valid_block_height)
     }
 }
 
@@ -1766,6 +1767,7 @@ impl CliFees {
         blockhash: Hash,
         lamports_per_signature: u64,
         last_valid_slot: Option<Slot>,
+        last_valid_block_height: Option<Slot>,
     ) -> Self {
         Self {
             inner: Some(CliFeesInner {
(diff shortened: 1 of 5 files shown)


client/src/rpc_client.rs
```diff
@@ -1247,6 +1248,34 @@ impl RpcClient {
         )
     }
 
+    pub fn get_fees(&self) -> ClientResult<Fees> {
```
CriesofCarrots, line 1251: I'm hoping there might be an opportunity to make these methods consistent with those in BanksClient at some point...

cli-output/src/cli_output.rs
```diff
@@ -1766,13 +1767,15 @@ impl CliFees {
         blockhash: Hash,
         lamports_per_signature: u64,
         last_valid_slot: Option<Slot>,
```
CriesofCarrots, line 1769: `last_valid_slot` will always return None as of this change. Does this seem too breaky?

mvines: 🤔 My gut is that this will be ok,  it only affects `solana fees--output json`, and I'd be surprised if anybody relies on `lastValidSlot` from there

sdk/src/lib.rs
```diff
@@ -23,6 +23,7 @@ pub mod epoch_info;
 pub mod exit;
 pub mod feature;
 pub mod feature_set;
+pub mod fees;
```
mvines, line 26: It's not clear me why this lives in the SDK?

CriesofCarrots: Only so I wouldn't have to move it when bringing the BanksClient methods in line with these RpcClient ones. I don't feel strongly, though, so will move to `/client` if you prefer.

mvines: I think I _slightly_ prefer keeping it in `/client`, given how general the mod name of "fees" is.  Or rather, out of `sdk/` for that same reason.  Perhaps it moves into `banks-interface/` eventually if it's only shared between `client/` and `banks-client/`.

CriesofCarrots: Moved.

mvines (approved):

Issue #12211, opened 2020-09-14
RPC: Add ability to view accounts at a given slot


mvines:
It's currently not possible to view account states from any slot older than the max confirmed slot.  Ideally `getMultipleAccounts`/`getAccountInfo` worked for slots that are days/weeks old.

This will require dumping all account data into some kind of storage like BigTable

PR #18624, opened 2021-07-12
Add Rpc AccountHistory service

### LINKED ISSUE #18197: Ability to query historic account state
#### Problem

For analytics/archival, a user often needs to capture the full history of account state over time. We specifically need this for Pyth data archival, but it's a common requirement for anyone doing in-depth analysis.

The current Pyth implementation uses account subscriptions over Websockets, however, on mainnet-beta, approximately 10% of account updates were dropped in a test with mainnet nodes running on default RPC settings. We had similar results when we tried to use subscriptions for Wormhole. As far as I can tell, websockets are best-effort and weren't designed for lossless capture:

- In case of queue overruns, messages are dropped silently without informing the consumer.
- Large account updates lead to head-of-line blocking.
- Streams cannot be resumed - this is because queues are managed per connection rather than using a log with consumer tracking (like Kafka) or persistent queues (like RabbitMQ). If the stream ends, data is lost till reconnection.

#### Proposed Solution

Solana can already store historic state locally for program logs and account balance changes. This could be extended to include the ability to optionally store account data for specific program-derived accounts as an opt-in feature, similar to `--enable-cpi-and-log-storage`. This would be ideal and allow us to backfill account state using load-balanced unary RPC calls backed by multiple nodes.

Related discussions:

- https://github.com/solana-labs/solana/issues/15461
- https://github.com/solana-labs/solana/issues/10194
- https://github.com/solana-labs/solana/issues/16986 (different use case, but same pull vs. consumer tracking discussion)
- https://github.com/solana-labs/solana/issues/16937
- https://github.com/solana-labs/solana/issues/11904
- https://github.com/solana-labs/solana/issues/9909
- https://github.com/solana-labs/solana/issues/15705
- https://github.com/solana-labs/solana/issues/18587

CriesofCarrots:
#### Problem
For analytics/archival, a user often needs to capture the full history of account state over time, but there is no way to query historical account state (and pubsub subscriptions aren't guaranteed to catch all updates).

#### Summary of Changes
Add an in-memory cache of historical account state per slot, for a configurable number of slots

Fixes #18197
Towards #12211

### DIFF
--- a/rpc/src/account_history.rs
+++ b/rpc/src/account_history.rs
@@ -1,0 +1,206 @@
+use {
+    crossbeam_channel::{Receiver, RecvTimeoutError},
+    solana_measure::measure::Measure,
+    solana_metrics::datapoint_info,
+    solana_runtime::bank::Bank,
+    solana_sdk::{account::Account, clock::Slot, pubkey::Pubkey},
+    std::{
+        collections::{BTreeMap, HashMap, HashSet},
+        sync::{
+            atomic::{AtomicBool, Ordering},
+            Arc, RwLock, RwLockWriteGuard,
+        },
+        thread::{self, Builder, JoinHandle},
+        time::Duration,
+    },
+};
+
+#[derive(Debug, Clone, Copy)]
+pub struct AccountHistoryConfig {
+    pub num_slots: usize,
+    pub updates_only: bool,
+}
+
+pub type AccountHistory = BTreeMap<Slot, HashMap<Pubkey, Account>>;
+pub type AccountKeys = HashSet<Pubkey>;
+
+pub struct RpcAccountHistoryService {
+    thread_hdl: JoinHandle<()>,
+}
+
+impl RpcAccountHistoryService {
+    pub fn new(
+        config: AccountHistoryConfig,
+        account_keys: Arc<RwLock<AccountKeys>>,
+        account_history: Arc<RwLock<AccountHistory>>,
+        account_history_receiver: Receiver<Arc<Bank>>,
+        exit: &Arc<AtomicBool>,
+    ) -> Self {
+        let exit = exit.clone();
+        let thread_hdl = Builder::new()
+            .name("solana-account-history".to_string())
+            .spawn(move || loop {
+                if exit.load(Ordering::Relaxed) {
+                    break;
+                }
+                if let Err(RecvTimeoutError::Disconnected) = Self::receive_bank(
+                    &config,
+                    &account_keys,
+                    &account_history,
+                    &account_history_receiver,
+                ) {
+                    break;
+                }
+            })
+            .unwrap();
+        Self { thread_hdl }
+    }
+
+    fn receive_bank(
+        config: &AccountHistoryConfig,
+        account_keys: &Arc<RwLock<AccountKeys>>,
+        account_history: &Arc<RwLock<AccountHistory>>,
+        account_history_receiver: &Receiver<Arc<Bank>>,
+    ) -> Result<(), RecvTimeoutError> {
+        let frozen_bank = account_history_receiver.recv_timeout(Duration::from_secs(1))?;
+        debug!(
+            "rpc account-history service received bank: {:?}",
+            frozen_bank.slot()
+        );
+
+        let r_account_keys = account_keys.read().unwrap();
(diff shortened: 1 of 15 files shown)


rpc/src/account_history.rs
```diff
@@ -0,0 +1,226 @@
...
+            ("write", measure_write.as_us(), i64),
+            ("prune", measure_prune.as_us(), i64),
+        );
+
+        Ok(())
+    }
+
+    fn collect_accounts(
+        frozen_bank: &Bank,
+        accounts: &HashSet<Pubkey>,
```
carllin, line 104: nit `accounts -> account_keys`

rpc/src/account_history.rs
```diff
@@ -0,0 +1,226 @@
...
+        let r_account_keys = account_keys.read().unwrap();
+        let mut measure_collect = Measure::start("collect-account-history");
+        let slot_accounts =
+            Self::collect_accounts(&frozen_bank, &r_account_keys, config.updates_only);
+        measure_collect.stop();
+        drop(r_account_keys);
+
+        let mut measure_write = Measure::start("write-account-history");
+        let mut w_account_history = account_history.write().unwrap();
+        w_account_history.insert(frozen_bank.slot(), slot_accounts);
```
carllin, line 85: Assuming this will eventually be persisted to something like the blockstore?

rpc/src/account_history.rs
```diff
@@ -0,0 +1,226 @@
...
+            if let Some(shared_account) = shared_account {
+                slot_accounts.insert(*address, shared_account.into());
+            }
+        }
+        slot_accounts
+    }
+
+    fn remove_old_slots(
+        mut w_account_history: RwLockWriteGuard<AccountHistory>,
+        num_slots: &usize,
```
carllin, line 126: nit: `num_slots -> max_slot_history`

rpc/src/account_history.rs
```diff
@@ -0,0 +1,226 @@
...
+            Arc, RwLock, RwLockWriteGuard,
+        },
+        thread::{self, Builder, JoinHandle},
+        time::Duration,
+    },
+};
+
+#[derive(Debug, Clone)]
+pub struct AccountHistoryConfig {
+    pub num_slots: usize,
```
carllin, line 23: nit: `num_slots -> max_slot_history`

rpc/src/account_history.rs
```diff
@@ -0,0 +1,226 @@
...
+        frozen_bank: &Bank,
+        accounts: &HashSet<Pubkey>,
+        updates_only: bool,
+    ) -> HashMap<Pubkey, Account> {
+        let mut slot_accounts = HashMap::new();
+        for address in accounts.iter() {
+            let shared_account = if updates_only {
+                frozen_bank
+                    .get_account_modified_slot(address)
+                    .filter(|(_shared_account, slot)| *slot == frozen_bank.slot())
```
carllin, line 112: hmmm if the intention is to capture *exactly* the update for this account, at `frozen_bank.slot()`, this may race with account clean if we make a root past this `frozen_bank.slot()` before this call to `get_account_modified_slot` runs.

rpc/src/account_history.rs
CriesofCarrots: I don't think that's needed at the moment. Since this service is intended to replace account subscriptions for the motivating use-case, I think it's expected that updates that occur when the node is offline will be unavailable. But yeah, blockstore backup is def an option if we need stronger persistence.

rpc/src/account_history.rs
CriesofCarrots: Makes sense. I'm a little reluctant to dangle bits into AccountsBackgroundService for this, unless some sort of minimum-root-to-keep tracker would be useful for other purposes?
Which approach do you prefer?

carllin: I think 2. is probably the easiest :), but also open to suggestions from @sakridge or @ryoqun

CriesofCarrots: Since we both lean toward solution 2, and having heard no feedback from those other two, I will start moving on that.

carllin: 👻

CriesofCarrots:
Will probably abandon this approach in favor of #20047

PR #18640, opened 2021-07-13
lazy allocate buffers for bg reader in untar

jeffwashington:
#### Problem
Startup is slow.

#### Summary of Changes
Allocate new buffers within our budget for bg reader during untar, as needed, in the bg reader thread. Buffers are currently filled with 0s on allocation. This new mechanism causes us to fill the next new buffer with 0s and start using it immediately in the bg reader, ideally keeping the buffer in the cache. Previously, we allocated and filled all the buffers up front, in the client thread, before we kicked off the bg reader. We can also now allocate fewer buffers than we otherwise would if the readers are keeping up with decompression progress.

Fixes #

### DIFF
--- a/runtime/src/shared_buffer_reader.rs
+++ b/runtime/src/shared_buffer_reader.rs
@@ -60,8 +60,10 @@ impl SharedBuffer {
         chunk_size: usize,
         reader: T,
     ) -> Self {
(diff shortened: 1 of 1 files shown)


sakridge (approved):
nice

PR #18646, opened 2021-07-13
Explicitly sanitize program id indexes before usage

### LINKED ISSUE #18638: Cost model calculations are performed on unsanitized transactions
#### Problem
https://github.com/solana-labs/solana/blob/6829d210d4f5c61ef4fe21abf8bf8757c08fee53/core/src/cost_model.rs#L183 relies on `verify_precompiles` to sanitize program id indexes. This implicit reliance is not very clear and could easily break in the future

tag @taozhu-chicago

taozhu-chicago:
#### Problem
for issue https://github.com/solana-labs/solana/issues/18638

#### Summary of Changes
1. check transaction has valid program_id before using it to avoid possible panic;
2. change calculate_cost function signature to return Result;
3. add CostModelError enum, update return type from Result<_, str> to Result<_, CostModelError>

Fixes #18638

### DIFF
--- a/core/src/banking_stage.rs
+++ b/core/src/banking_stage.rs
@@ -1217,7 +1217,14 @@ impl BankingStage {
                 cost_tracker
                     .write()
                     .unwrap()
-                    .add_transaction_cost(tx.transaction());
+                    .add_transaction_cost(tx.transaction())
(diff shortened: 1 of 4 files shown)


jstarry (approved):
lgtm, thanks!

PR #18648, opened 2021-07-13
Add frozen hashes and marking DuplicateConfirmed in blockstore to state machine

carllin:
#### Problem
Added frozen hash column to blockstore, but it's not being populated

#### Summary of Changes
Add state transition for adding frozen hashes and duplicate confirmed to blockstore

Fixes #

### DIFF
--- a/core/src/cluster_slot_state_verifier.rs
+++ b/core/src/cluster_slot_state_verifier.rs
@@ -65,19 +65,23 @@ pub struct DeadState {
 }
 
 impl DeadState {
-    pub fn new(
+    pub fn new_from_state(
         slot: Slot,
         duplicate_slots_tracker: &mut DuplicateSlotsTracker,
         gossip_duplicate_confirmed_slots: &GossipDuplicateConfirmedSlots,
         fork_choice: &mut HeaviestSubtreeForkChoice,
     ) -> Self {
-        let duplicate_confirmed_hash = get_cluster_duplicate_confirmed_hash_from_state(
+        let duplicate_confirmed_hash = get_duplicate_confirmed_hash_from_state(
             slot,
             gossip_duplicate_confirmed_slots,
             fork_choice,
-            Hash::default(),
+            Some(Hash::default()),
         );
         let is_slot_duplicate = duplicate_slots_tracker.contains(&slot);
+        Self::new(duplicate_confirmed_hash, is_slot_duplicate)
+    }
+
+    fn new(duplicate_confirmed_hash: Option<Hash>, is_slot_duplicate: bool) -> Self {
         Self {
             duplicate_confirmed_hash,
             is_slot_duplicate,
@@ -95,21 +99,29 @@ pub struct BankFrozenState {
 }
 
 impl BankFrozenState {
-    pub fn new(
+    pub fn new_from_state(
         slot: Slot,
         frozen_hash: Hash,
         duplicate_slots_tracker: &mut DuplicateSlotsTracker,
         gossip_duplicate_confirmed_slots: &GossipDuplicateConfirmedSlots,
         fork_choice: &mut HeaviestSubtreeForkChoice,
     ) -> Self {
-        assert!(frozen_hash != Hash::default());
-        let duplicate_confirmed_hash = get_cluster_duplicate_confirmed_hash_from_state(
+        let duplicate_confirmed_hash = get_duplicate_confirmed_hash_from_state(
             slot,
             gossip_duplicate_confirmed_slots,
             fork_choice,
-            frozen_hash,
+            Some(frozen_hash),
         );
         let is_slot_duplicate = duplicate_slots_tracker.contains(&slot);
+        Self::new(frozen_hash, duplicate_confirmed_hash, is_slot_duplicate)
+    }
+
+    fn new(
+        frozen_hash: Hash,
+        duplicate_confirmed_hash: Option<Hash>,
+        is_slot_duplicate: bool,
+    ) -> Self {
+        assert!(frozen_hash != Hash::default());
         Self {
             frozen_hash,
             duplicate_confirmed_hash,
@@ -122,18 +134,23 @@ impl BankFrozenState {
 pub struct DuplicateConfirmedState {
     // Keep fields private, forces construction
     // via constructor
-    bank_status: BankStatus,
     duplicate_confirmed_hash: Hash,
+    bank_status: BankStatus,
 }
 impl DuplicateConfirmedState {
-    pub fn new(
+    pub fn new_from_state(
         duplicate_confirmed_hash: Hash,
         is_dead: impl Fn() -> bool,
         get_hash: impl Fn() -> Option<Hash>,
     ) -> Self {
+        let bank_status = BankStatus::new(is_dead, get_hash);
+        Self::new(duplicate_confirmed_hash, bank_status)
+    }
+
+    fn new(duplicate_confirmed_hash: Hash, bank_status: BankStatus) -> Self {
         Self {
             duplicate_confirmed_hash,
-            bank_status: BankStatus::new(is_dead, get_hash),
+            bank_status,
         }
     }
 }
@@ -142,11 +159,11 @@ impl DuplicateConfirmedState {
 pub struct DuplicateState {
     // Keep fields private, forces construction
     // via constructor
-    bank_status: BankStatus,
     duplicate_confirmed_hash: Option<Hash>,
+    bank_status: BankStatus,
 }
 impl DuplicateState {
-    pub fn new(
+    pub fn new_from_state(
         slot: Slot,
         gossip_duplicate_confirmed_slots: &GossipDuplicateConfirmedSlots,
         fork_choice: &mut HeaviestSubtreeForkChoice,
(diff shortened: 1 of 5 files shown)


core/src/cluster_slot_state_verifier.rs
```diff
@@ -33,147 +45,185 @@ impl SlotStateUpdate {
...
 }
 
-fn on_dead_slot(
+fn check_duplicate_confirmed_hash_against_frozen_hash(
+    state_changes: &mut Vec<ResultingStateChange>,
     slot: Slot,
-    bank_frozen_hash: &Hash,
-    cluster_duplicate_confirmed_hash: Option<&Hash>,
-    _is_slot_duplicate: bool,
+    cluster_duplicate_confirmed_hash: Option<Hash>,
```
AshwinSekar, line 58: should we just pass the args in here? seems like it's all the fields anyway

core/src/cluster_slot_state_verifier.rs
```diff
@@ -33,147 +45,185 @@ impl SlotStateUpdate {
...
+            // of the slot duplicate_confirmed (i.e. block above did not execute), then
+            // remove the slot from fork choice until we get confirmation.
+
+            // If we get here, we either detected duplicate from
+            // 1) WindowService
+            // 2) A gossip duplicate_confirmed version that didn't match our frozen
+            // version.
+            // In both cases, mark the progress map for this slot as duplicate
+            return vec![ResultingStateChange::MarkSlotDuplicate(bank_frozen_hash)];
+        }
```
AshwinSekar, line 208: might be cleaner like this since both cases result in marking duplicate
```suggestion
    if let Some(cluster_duplicate_confirmed_hash) = cluster_duplicate_confirmed_hash  && cluster_duplicate_confirmed_hash == bank_frozen_hash {
       return vec![]
    }
    vec![ResultingStateChange::MarkSlotDuplicate(bank_frozen__hash)];
}
```

core/src/cluster_slot_state_verifier.rs
```diff
@@ -255,76 +332,546 @@ pub(crate) fn check_slot_agrees_with_cluster(
...
+                    cluster_duplicate_confirmed_hash,
+                    is_dead: true,
+                    is_slot_duplicate: true,
+                },
+                SlotStateUpdate::DuplicateConfirmed,
+                vec![
+                ResultingStateChange::DuplicateConfirmedSlotMatchesCluster(bank_frozen_hash)]
+            )
+        },
+        duplicate_confirmed_state_update_4: {
```
AshwinSekar, line 589: Do we want to check case 4 & 5 with the slot being dead as well? I don't think we filter out dead frozen slots on DuplicateConfirmed right?

core/src/cluster_slot_state_verifier.rs
```diff
@@ -255,76 +332,546 @@ pub(crate) fn check_slot_agrees_with_cluster(
...
+                },
+                SlotStateUpdate::Duplicate,
+                vec![
+                ResultingStateChange::MarkSlotDuplicate(bank_frozen_hash)],
+            )
+        },
+        duplicate_state_update_2: {
+            // Slot must be marked duplicate
+            let is_slot_duplicate = true;
+            // Slot is frozen, so mark the duplicate
```
AshwinSekar, line 773: nit: Slot is frozen but same hash so no action

core/src/cluster_slot_state_verifier.rs
```diff
@@ -2,24 +2,36 @@ use crate::{
...
+}
 
 #[derive(PartialEq, Debug)]
 pub enum SlotStateUpdate {
-    Frozen,
+    BankFrozen,
     DuplicateConfirmed,
     Dead,
     Duplicate,
 }
```
jstarry, line 29: I think it'd be awesome to add some extra type safety and remove the need for the assertions in this module. With these adjustments you could also remove some test cases that are nonsensical.

```rust
pub enum BankStatus {
  Frozen(Hash),
  Dead,
  Unprocessed,
}

pub enum SlotStateUpdate {
    BankFrozen(Hash),
    DuplicateConfirmed(BankStatus),
    Dead,
    Duplicate(BankStatus),
}
```

core/src/cluster_slot_state_verifier.rs
```diff
@@ -33,147 +45,185 @@ impl SlotStateUpdate {
...
-            SlotStateUpdate::Duplicate => on_cluster_update,
+            SlotStateUpdate::BankFrozen => on_frozen_slot,
+            SlotStateUpdate::DuplicateConfirmed => on_duplicate_confirmed,
+            SlotStateUpdate::Duplicate => on_duplicate,
         }
     }
 }
 
-fn on_dead_slot(
+fn check_duplicate_confirmed_hash_against_frozen_hash(
```
jstarry, line 55: This function does too much in my opinion. As a start, how about this is only called when we _do_ have a cluster duplicate confirmed hash?

core/src/cluster_slot_state_verifier.rs
carllin: ah good catch, bad copy-pasta :'(

core/src/cluster_slot_state_verifier.rs
carllin: yeah a lot better, thanks

core/src/cluster_slot_state_verifier.rs
carllin: Yeah, agreed, updated!

core/src/cluster_slot_state_verifier.rs
carllin: In cases 4 and 5 the `bank_frozen_hash` is not default, so the slot cannot be dead. `duplicate_confirmed_state_update_6` and `duplicate_confirmed_state_update_7` test when the bank .Hash is default and dead

core/src/cluster_slot_state_verifier.rs
carllin: Not after Justin's latest suggestion: https://github.com/solana-labs/solana/pull/18648#discussion_r670763448

core/src/replay_stage.rs
```diff
@@ -2436,6 +2457,7 @@ impl ReplayStage {
 
     fn mark_slots_confirmed(
         confirmed_forks: &[Slot],
```
jstarry, line 2459: Could this be slot hashes instead to avoid the extra bank forks locking and assertions below?

carllin: yup, much cleaner, updated here: https://github.com/solana-labs/solana/pull/18648/commits/b6803479a84fc66b5605396c5f52ad5f9400ec62

core/src/cluster_slot_state_verifier.rs
carllin: Updated!

jstarry (commented):
looks good, just had a few questions

core/src/cluster_slot_state_verifier.rs
```diff
@@ -30,151 +223,218 @@ pub enum ResultingStateChange {
 }
 
 impl SlotStateUpdate {
-    fn to_handler(&self) -> SlotStateHandler {
+    fn into_state_changes(self, slot: Slot) -> Vec<ResultingStateChange> {
+        let bank_frozen_hash = self.bank_hash();
+        if bank_frozen_hash == None {
+            // If the bank hasn't been frozen yet, then there's nothing to do
+            // replay of the slot hasn't finished yet.
```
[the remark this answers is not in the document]
carllin: yeah it should repair up until the slot is complete or dead.

core/src/cluster_slot_state_verifier.rs
```diff
@@ -30,151 +223,218 @@ pub enum ResultingStateChange {
...
-    // Bank should not have been frozen if the slot was marked dead
-    assert_eq!(*bank_frozen_hash, Hash::default());
-    if let Some(cluster_duplicate_confirmed_hash) = cluster_duplicate_confirmed_hash {
+) {
+    if duplicate_confirmed_hash != bank_frozen_hash {
+        if is_dead {
+            // If the cluster duplicate confirmed some version of this slot, then
+            // there's another version of our dead slot
+            warn!(
+                "Cluster duplicate_confirmed slot {} with hash {}, but we marked slot dead",
```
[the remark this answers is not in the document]
carllin: Currently `dump_then_repair_correct_slots` doesn't actively engage a specific repair protocol for duplicate slots here. There is a change that could trigger repair_service to engage the duplicate repair protocol, like so: https://github.com/carllin/solana/blob/FixReplayStage2/core/src/replay_stage.rs#L938-L942, but this hasn't been checked in yet.

core/src/cluster_slot_state_verifier.rs
```diff
@@ -30,151 +223,218 @@ pub enum ResultingStateChange {
...
+    if let Some(duplicate_confirmed_hash) = duplicate_confirmed_hash {
+        if duplicate_confirmed_hash == bank_hash {
+            return vec![];
+        }
     }
 
-    vec![]
+    // If we either:
+    // 1) Have not yet seen any version of the slot duplicate_confirmed,
+    // 2) Our version of the slot does not match the duplicate_confirmed version,
```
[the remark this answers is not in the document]
carllin: Yeah I was thinking about this too, because if the slot is duplicate confirmed and our frozen hash doesn't match, the duplicate flag should already have been set by either the `Frozen` or the `DuplicateConfirmed`  state transitions.

There is no correctness issue for called the mark duplicate multiple times on the same node, it just incurs this cost: https://github.com/solana-labs/solana/blob/40f207546f5fbdee0c5a7293f0b788a810da1e14/core/src/heaviest_subtree_fork_choice.rs#L983-L1008. I guess this can be avoided in this situation.

Updated!

jstarry (approved):
lgtm, nice cleanup!

PR #18652, opened 2021-07-14
Broadcast method test infrastructure

AshwinSekar:
#### Summary of Changes

Fixes #

### DIFF
--- a/validator/src/main.rs
+++ b/validator/src/main.rs
@@ -21,6 +21,8 @@ use {
         rpc_request::MAX_MULTIPLE_ACCOUNTS,
     },
     solana_core::{
+        broadcast_stage::broadcast_duplicates_run::BroadcastDuplicatesConfig,
+        broadcast_stage::BroadcastStageType,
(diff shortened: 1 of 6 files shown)


validator/src/main.rs
```diff
@@ -1478,6 +1480,32 @@ pub fn main() {
...
+            Arg::with_name("duplicate_stake_partition")
+                .long("duplicate-stake-partition")
+                .value_name("STAKE")
+                .takes_value(true)
+                .requires("broadcast_stage_type")
+                .default_value("50")
+                .help("Specifies the percentage of the stake to send duplicate shreds to when using
+                    the broadcast_duplicates stage type."
+                ),
+        )
```
mvines, line 1508: At minimum we want to hide these options from the --help.  We might even want to go further and not even land the fully plumbing for this until @carllin's slot dumping work completes and rolls out to mainnet

AshwinSekar: Fair point, definitely don't want to land this until slot dumping has been rolled out. I created this pull request so that we could test @carllin's prs on the gce cluster. I'll also remove the help info.

mvines: Got it, thank!

PR #18673, opened 2021-07-14
experiment: refcount is no longer an atomic

jeffwashington:
#### Problem
Some advantages:
less memory per accounts index entry.
simpler api to clients - write lock to modify, read lock to read
allows flexibility on the implementation
#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -2288,7 +2288,8 @@ impl AccountsDb {
                 .skipped_shrink
                 .fetch_add(1, Ordering::Relaxed);
             for pubkey in unrefed_pubkeys {
-                if let Some(locked_entry) = self.accounts_index.get_account_write_entry(pubkey) {
+                if let Some(mut locked_entry) = self.accounts_index.get_account_write_entry(pubkey)
+                {
                     locked_entry.addref();
                 }
             }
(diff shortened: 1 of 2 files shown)


jeffwashington:
lemond, mnb snapshot, store_us went from 
store_us (but in s) out of a 80s ledger verify. Time was always ~80s so no overall effect, but individual metric impact for sure. There are dimensions of tuning the locks, # of bins, etc.
no atomic refcount
115 (I think this also requires write locks in more places)
no arc on slotlist/refcount and no atomic refcount
108 (note this requires write locks in more places)
master
107

jeffwashington:
a previous version: f053a54

jeffwashington:
bench-tps
wiggins, threads=16
```
this branch:
    Highest TPS: 79094.26 sampling period 1s max transactions: 4300387 clients: 1 drop rate: 0.82
Average TPS: 47565.402
    Highest TPS: 83139.75 sampling period 1s max transactions: 4578605 clients: 1 drop rate: 0.81
Average TPS: 50815.383

master:
    Highest TPS: 81955.91 sampling period 1s max transactions: 4730492 clients: 1 drop rate: 0.81
Average TPS: 51941.133
    Highest TPS: 82443.72 sampling period 1s max transactions: 4550791 clients: 1 drop rate: 0.81
Average TPS: 49979.543
```

jeffwashington:
kin got about 100M accounts further with the same memory on the validator with this change and a few others. I don't have isolated results. That makes sense that this change would cause us to use less memory per account index entry.

PR #18678, opened 2021-07-14
use more threads to generate index at startup now that lock contentio…

jeffwashington:
…n is reduced with more bins

#### Summary of Changes

Fixes #


jeffwashington:
this is slower with this accounts index for some reason. back to the drawing board. It measured faster with other changes.

PR #18681, opened 2021-07-14
Add AncestorHashesRepair type

carllin:
#### Problem
There is no way to query peers for the correct ancestors of a slot to resolve duplicate failures

#### Summary of Changes
Add a repair protocol for this type of query

Fixes #

### DIFF
--- a/core/src/serve_repair.rs
+++ b/core/src/serve_repair.rs
@@ -41,7 +41,7 @@ pub const MAX_ORPHAN_REPAIR_RESPONSES: usize = 10;
 pub(crate) const REPAIR_PEERS_CACHE_CAPACITY: usize = 128;
 // Limit cache entries ttl in order to avoid re-using outdated data.
 const REPAIR_PEERS_CACHE_TTL: Duration = Duration::from_secs(10);
-pub const MAX_ANCESTOR_RESPONSES: usize = 20;
+pub const MAX_ANCESTOR_RESPONSES: usize = 25;
 
 #[derive(Serialize, Deserialize, Debug, Clone, Copy, Hash, PartialEq, Eq)]
 pub enum ShredRepairType {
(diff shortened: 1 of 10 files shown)


AshwinSekar (dismissed):
lgtm

core/src/serve_repair.rs
```diff
@@ -40,46 +41,83 @@ pub const MAX_ORPHAN_REPAIR_RESPONSES: usize = 10;
 pub(crate) const REPAIR_PEERS_CACHE_CAPACITY: usize = 128;
 // Limit cache entries ttl in order to avoid re-using outdated data.
 const REPAIR_PEERS_CACHE_TTL: Duration = Duration::from_secs(10);
+pub const MAX_ANCESTOR_RESPONSES: usize = 25;
```
jstarry, line 44: Should be able to bump to 30. `1232 - 4 (nonce) - 4 (response version enum discriminator) - 4 (slot_hash length) == 1220` and each slot hash is 40 bytes

jstarry: Can you also add a comment on how you came to the number 25 (or 30 if you change it)

jstarry (approved):
looks good!

core/src/serve_repair.rs
carllin: Made things a lot more explicit here: https://github.com/solana-labs/solana/pull/18681/commits/c9816c36e8cd6fa091b7c66feb98474c7d94ab30

core/src/serve_repair.rs
```diff
@@ -35,13 +37,21 @@ use std::{
...
 // Limit cache entries ttl in order to avoid re-using outdated data.
 const REPAIR_PEERS_CACHE_TTL: Duration = Duration::from_secs(10);
-pub const MAX_ANCESTOR_RESPONSES: usize = 25;
+pub const MAX_ANCESTOR_BYTES_IN_PACKET: usize =
+    PACKET_DATA_SIZE -
+    SIZE_OF_NONCE -
+    4 /*(response version enum discriminator)*/ -
+    4 /*slot_hash length*/;
+pub const MAX_ANCESTOR_RESPONSES: usize =
+    MAX_ANCESTOR_BYTES_IN_PACKET / std::mem::size_of::<SlotHash>();
```
jstarry, line 54: could you add in a:
```rust
#[cfg(test)]
static_assertions::const_assert_eq!(MAX_ANCESTOR_RESPONSES, 30);
```

jstarry: it just helps for readability

jstarry (approved):

PR #18682, opened 2021-07-14
Fix race condition in bg file reader file error

jeffwashington:
#### Problem
error is being set before all the read data is being recorded. The result is that a reader can detect the error before it returns all the data that had been read prior to the error being hit.
#### Summary of Changes
Set the error after the data has been transferred.
Fixes #


lijunwangs (approved):
Looks good

PR #18684, opened 2021-07-15
`solana-validator wait-for-restart-window` command now accepts an optional --identity argument

mvines:
This allows one to wait for the restart window of some other validator.

### DIFF
--- a/validator/src/main.rs
+++ b/validator/src/main.rs
@@ -105,6 +105,7 @@ fn monitor_validator(ledger_path: &Path) {
 
 fn wait_for_restart_window(
     ledger_path: &Path,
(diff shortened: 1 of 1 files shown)


mvines:
I wouldn't mind, it's super low risk.  Was there somebody asking for this feature on v1.8/v1.7?

steviez:
Nobody asked for it; rather, I somewhat stumbled across it while browsing through PR's and finding https://github.com/solana-labs/solana/pull/20381 (the cherry-pick failed because the change that #20381 backports was merged after the change in this PR; they touch the some of the same spots so the auto-resolve failed).

PR #18686, opened 2021-07-15
Add comments on how to add a new feature switch

steviez:
#### Problem
Steps to add a new feature switch may be unclear for someone unfamiliar with the steps (like me)

#### Summary of Changes
Document steps to add a new feature switch

### DIFF
--- a/sdk/program/src/feature.rs
+++ b/sdk/program/src/feature.rs
@@ -1,12 +1,16 @@
-/// Runtime features.
-///
-/// Feature activation is accomplished by:
-/// 1. Activation is requested by the feature authority, who issues a transaction to create the
-///    feature account.  The newly created feature account will have the value of
-///    `Feature::default()`
-/// 2. When the next epoch is entered the runtime will check for new activation requests and
-///    active them.  When this occurs, the activation slot is recorded in the feature account
-///
+//! Runtime features.
+//!
+//! Runtime features provide a mechanism for features to be simultaneously activated across the
+//! network. Since validators may choose when to upgrade, features must remain dormant until a
+//! sufficient majority of the network is running a version that would support a given feature.
+//!
+//! Feature activation is accomplished by:
+//! 1. Activation is requested by the feature authority, who issues a transaction to create the
+//!    feature account. The newly created feature account will have the value of
+//!    `Feature::default()`
+//! 2. When the next epoch is entered the runtime will check for new activation requests and
+//!    active them.  When this occurs, the activation slot is recorded in the feature account
+
 use crate::{
     account_info::AccountInfo, clock::Slot, instruction::Instruction, program_error::ProgramError,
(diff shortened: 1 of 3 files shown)


steviez:
@mvines - Anything to add here? Specifically, it isn't clear to me if we need to do anything special (upload, explicitly discard, etc) with the private key that is generated

mvines:
Thanks, this is a great start.  The feature keypair management is probably the most important part in all of this, as the holder of that keypair has the ability to take down a cluster if the feature is activated when between 33% and 66% of the staked validators are running the build with the new feature present.

Keypair management basics:
1. The keypair should only be held by a core contributor, and even then somebody who is _tuned in_ with the cluster upgrade cycles. If a non-core contributor sends a PR with a feature, we decide informally who will hold the keypair on their behalf.  That person then creates and holds the keypair and shares its public key for the PR.
2. Features are usually enabled on testnet first, then devnet and finally mainnet using `solana feature ...`
3. If you are the core contributor that holds a feature keypair, you personally are responsible for keeping the keypair secure, and ensuring the feature activation is smooth and that mainnet does not die due to a poorly timed activation.  This may involve testing on testnet/devnet before mainnet, and for all three clusters ensuring that all validators, exchanges and other partners have upgraded to a compatible release before flipping the switch.  `solana feature ...` does attempt to prevent you from doing dumb things though (but you can still do them if you persist). 

Another important nodes is that we only use this feature mechanism for technical improvements that are expected to be non-contentious and are clearly for the common good of Solana. Community governance of features that people would care about (inflation, fee structure changes) are done with https://spl.solana.com/feature-proposal#feature-proposal-life-cycle, which is built on top of this facility

steviez:
Thanks for the response @mvines; I'll fully process it in the morning. Another question I have at the moment is that the process you described covers activating a feature, but I'm curious about deprecation process. If we have a feature switch that we no longer want, we can just update the codebase to always take the path we want. Ie,
```
if my_feature_enabled() { do_new_method(); }
else { do_old_method(); }
```
might simplify down to just
```
do_new_method();
```
In this scenario, I believe we could/would also remove `my_feature` and its' associated id from the `FEATURE_NAMES` map in `feature_set` (based on [previous commits](https://github.com/solana-labs/solana/commits/master/sdk/src/feature_set.rs)). However, I don't see a "deactivate" sub command in `solana feature`; is ripping the constant out of the map enough? I guess maybe the account would live in perpetuity, but when we remove if from `feature_set`, we cease to care about it (ie it won't be in the list of features we query).

mvines:
Runtime feature removal is supposed to follow https://docs.solana.com/developing/backwards-compatibility. In practice, it's currently more like: a feature can be removed from the `master` once it's been active on the `stable` channel through a full minor upgrade.

mvines (approved):
Thanks for adding this documentation

PR #18702, opened 2021-07-15
add Debug trait to T in AccountsIndex

jeffwashington:
#### Problem
AccountsIndex is changing a lot. It will be convenient if we have the debug trait on the type in AccountsIndex.
#### Summary of Changes
add the trait
Fixes #


brooksprumo (approved):
Looks good to me!

PR #18703, opened 2021-07-15
featurize cost model

taozhu-chicago:
#### Problem
To featurizing cost model to allow no affect release to testnet

#### Summary of Changes
- added feature id for cost_model in sdk feature_set
- checking if feature is enabled from current bank before: applying/resetting cost limits (banking_stage), updating cost table (replay_stage)

Fixes #

### DIFF
--- a/core/src/banking_stage.rs
+++ b/core/src/banking_stage.rs
@@ -1220,17 +1220,7 @@ impl BankingStage {
         if cost_model_enabled {
             transactions.iter().enumerate().for_each(|(index, tx)| {
                 if unprocessed_tx_indexes.iter().all(|&i| i != index) {
-                    cost_tracker
-                        .write()
-                        .unwrap()
-                        .add_transaction_cost(tx)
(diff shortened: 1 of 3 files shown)


core/src/banking_stage.rs
```diff
@@ -380,10 +381,14 @@ impl BankingStage {
 
     fn reset_cost_tracker_if_new_bank(
         cost_tracker: &Arc<RwLock<CostTracker>>,
-        bank_slot: Slot,
+        bank: Arc<Bank>,
         banking_stage_stats: &BankingStageStats,
     ) {
-        cost_tracker.write().unwrap().reset_if_new_bank(bank_slot);
+        if !bank.feature_set.is_active(&feature_set::cost_model::id()) {
```
taozhu-chicago, line 387: assuming bank inherits feature_set from its ancestors, that it is safe to fetch feature_set from current bank rather than from `root_bank`. Can someone comment if this is the preferred practice?

PR #18711, opened 2021-07-16
Add --acquire-node-instance-then-vote validator flag

mvines:
This PR builds on #18445 to add a simple two validator instance failover method that allows for
* upgrades of your validator software with virtually no down time, and
* clean failover to the backup instance when your monitoring detects a problem with the primary instance
without any safety issues that would otherwise be associated with running two instances of your validator.

This coordination is accomplished by integrating knowledge of the `NodeInstance` program into ReplayStage to control when each instance votes and becomes leader.

When the `--acquire-node-instance-then-vote` arg is given the validator always votes with `VoteInstruction::VoteWithInstance`, which allows the vote program itself to reject votes when the validator no longer holds its `NodeInstance` lock.

The hand-off between validator instances is accomplished as follows.  Before starting to vote, the secondary instance:
1. Sends transactions to acquire the `NodeInstance` lock for its identity
2. Monitors until it sees a finalized block with its `NodeInstance` lock
3. Reads its vote account off-chain, and waits until all vote lockouts expire. This means it does not require the tower .bin file from the primary instance, as at this point any inflight votes from the primary instance cannot land. 
4. Once on-chain lockouts expire, resume the existing startup sequence of trying to root a vote before becoming leader.

Somewhere between 1 and 4, it's likely but not guaranteed that the primary instance will detect the secondary instance in gossip and self-terminate. However this is not required and even if the primary never exits then the secondary will still succeed in taking over voting and block production safely.

TODO:
* [ ] First pass design/concept review of this PR
* [ ] Remove #18445 from this PR, add tests and land it first
* [ ] Tests for this PR once the above two items are addressed

### DIFF
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -19,7 +19,7 @@ use solana_sdk::{
     slot_history::{Check, SlotHistory},
 };
 use solana_vote_program::{
-    vote_instruction,
+    node_instance, vote_instruction,
     vote_state::{BlockTimestamp, Lockout, Vote, VoteState, MAX_LOCKOUT_HISTORY},
 };
 use std::{
@@ -49,6 +49,7 @@ impl SwitchForkDecision {
         vote: Vote,
         vote_account_pubkey: &Pubkey,
         authorized_voter_pubkey: &Pubkey,
+        instance_id: Option<&(node_instance::InstanceId, Pubkey)>,
     ) -> Option<Instruction> {
         match self {
             SwitchForkDecision::FailedSwitchThreshold(_, total_stake) => {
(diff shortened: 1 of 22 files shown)


core/src/consensus.rs
```diff
@@ -1263,6 +1283,32 @@ impl Tower {
...
+        tower.vote_state.root_slot = Some(0);
+    }
+    if let Err(err) = tower.adjust_lockouts_with_slot_history(slot_history) {
+        warn!(
+            "adjust_lockouts_with_slot_history failed, assuming locked out: {}",
+            err
+        );
+        return true;
+    }
+    tower.vote_state.root_slot = None; // Side step `ancestors` assert! in `is_locked_out`
```
mvines, line 1308: Note to self: this just needs to be dealt with within `Tower::is_locked_out()`

mvines:
Fiddle de dee, this PR needs a major rework.

PR #18714, opened 2021-07-16
Accumulate consumed units

jackcmay:
#### Summary of Changes

Record compute units consumed in the timing metrics so that the information is available to transaction simulation and cost model

Fixes #


taozhu-chicago (approved):
looks good. So cost model sees the execute cost in both microsec and compute units.

PR #18725, opened 2021-07-16
Version transaction message and add new message format

jstarry:
#### Problem
Need a new message format for processing transactions that load accounts from an on-chain address map. Also need a way to differentiate this new message from the legacy transaction message format.

#### Summary of Changes
- Introduce `VersionedMessage` which has custom serialization to determine which message version to deserialize by inspecting the first byte.
- Add `v0::Message` which supports address maps and is serialized with version == 0

Fixes #


sdk/program/src/message/original.rs
```diff
@@ -163,28 +164,7 @@ fn get_program_ids(instructions: &[Instruction]) -> Vec<Pubkey> {
...
-    /// but are not permitted to credit or debit lamports or modify account data. Transactions
-    /// targeting the same read-write account are evaluated sequentially.
-    pub num_readonly_signed_accounts: u8,
-
-    /// The last num_readonly_unsigned_accounts of the unsigned keys are read-only accounts.
-    pub num_readonly_unsigned_accounts: u8,
-}
-
-#[frozen_abi(digest = "BPBJZxpRQ4JS7LGJtsgoyctg4BXyBbbY4uc7FjowtxLV")]
+#[frozen_abi(digest = "8Rf1mAQbCgnTLkS4CUK9Tx2i7puRp3aq1qw6AGZCrj1L")]
```
jstarry, line 167: This changed because the ABI macro uses the fully qualified struct name in its hash (`solana_program::message::Message` vs `solana_program::message::original::Message`)

jstarry: I have a change here that would prevent this: https://github.com/solana-labs/solana/pull/18734

sdk/program/src/message/original.rs
```diff
@@ -163,28 +164,7 @@ fn get_program_ids(instructions: &[Instruction]) -> Vec<Pubkey> {
         .collect()
 }
 
-pub const MESSAGE_HEADER_LENGTH: usize = 3;
-
-#[frozen_abi(digest = "BVC5RhetsNpheGipt5rUrkR6RDDUHtD5sCLK1UjymL4S")]
```
ryoqun, line 168: (this is confuting part, maybe I have just removed this....)

sdk/program/src/message/v0.rs
```diff
@@ -0,0 +1,385 @@
...
+    hash::Hash,
+    instruction::CompiledInstruction,
+    pubkey::Pubkey,
+    sanitize::{Sanitize, SanitizeError},
+    short_vec,
+};
+
+/// Alternative version of `Message` that supports succinct account loading
+/// through an on-chain address map.
+#[frozen_abi(digest = "FqrtpvgZZxJBCqcETJ7LYVdaLia2n4izZ813DxgyrW2u")]
```
ryoqun, line 14: this can be removed and should be covered with new `frozen_abi()` on `MessageVersions`: for https://github.com/solana-labs/solana/pull/18725/files#diff-d4c52a4504e9e02f6f13335ff68c21b5dadccc0e2ac44435e74c86dc2c1f8b30R30

sdk/program/src/message/v0.rs
```diff
@@ -0,0 +1,385 @@
...
+                }
+            }
+        }
+
+        Ok(())
+    }
+}
+
+/// Address map specifies which entries to load and
+#[frozen_abi(digest = "Bus4bNKecGtgSixqPn3vaaFmzyPVfnmLUwh6as9cZmQa")]
```
ryoqun, line 112: this frozen_abi too: https://github.com/solana-labs/solana/pull/18725/files#r672290250

sdk/program/src/message/original.rs
ryoqun: @jstarry it's no surprise you created #18734 because of lack of design philosophy of frozen_abi... I constantly fail to empty my work queue to document it properly.... xD

well, frozen_abi is designed to catch even structurally-equal rename like this, forcing devs to check their sanity.
so, I'd rather keep this noisy (i know...) behavior.

I'm aware it's bothersome and no fun to be suspected for possibly breaking abi for well-sanctioned just _rename_s. but frozen_abi is tasked for this noisy policing in that way.

Sympathizing human, frozen_abi wants to minimize the maintenance burden of each and every safe abi changes. So, frozen_abi is designed to be used sparingly across the code base, ideally each for the topmost type of some hierarchy with abi concerns.

To that end, frozen_abi tries _super_ hard to structurally digest type information even piercing `enum` variants separately into each arm (and recursively). (I guess you felt the die-hardness of the frozen_code..) This hassle is all for to make fewer and more wider frozen_abi possible for the purpose.

in other words, the versioned transaction should comprise the single coherent frozen_abi hash now as a whole from frozen_abi's semantic viewpoint. For example, adding yet another newer next-gen tx format surely would break this abi, so versioning must be part of it. Likewise, even single tiny change deepest in the type hierarchy should be caught by the sole and mighty frozen_abi residing at the top versioning `enum`. That's the motto of frozen_abi.

So to recap, `MessageHeader` is anti-pattern (sorry this was confusing to be begin with) and `Protocol` is good-pattern according to the philosophy. And I left comments on this pr according to this thinking.

sdk/program/src/message/versions.rs
```diff
@@ -0,0 +1,262 @@
...
+/// Message versions supported by the Solana runtime.
+///
+/// # Serialization
+///
+/// If the first bit is set, the remaining 7 bits will be used to determine
+/// which message version is serialized starting from version `0`. If the first
+/// is bit is not set, all bytes are used to encode the original `Message`
+/// format.
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub enum MessageVersions {
```
ryoqun, line 30: if you find it hard to place `frozen_abi` because of this byte-clamping versioning scheme for transactions. let me know. I can tweak frozen_abi to work, if possible.

jstarry: I forgot to add it! It works great, thanks :)

sdk/program/src/message/mod.rs
```diff
@@ -0,0 +1,28 @@
...
+
+mod original;
+mod v0;
+mod versions;
+
+pub use original::Message;
+
+pub const MESSAGE_HEADER_LENGTH: usize = 3;
+
+#[frozen_abi(digest = "BVC5RhetsNpheGipt5rUrkR6RDDUHtD5sCLK1UjymL4S")]
```
ryoqun, line 11: related to this (https://github.com/solana-labs/solana/pull/18725/files#r672288029), could you stop carrying over my bad code by removing this redundant `frozen_abi`?

jstarry: Removed!

sdk/program/src/message/versions.rs
```diff
@@ -0,0 +1,263 @@
...
+
+                match prefix {
+                    MessagePrefix::Original(num_required_signatures) => {
+                        // The remaining fields of the original Message struct after the first byte.
+                        #[derive(Serialize, Deserialize)]
+                        struct RemainingMessage {
+                            pub num_readonly_signed_accounts: u8,
+                            pub num_readonly_unsigned_accounts: u8,
+                            #[serde(with = "short_vec")]
+                            pub account_keys: Vec<Pubkey>,
```
sakridge, line 128: the duplication here is unfortunate

sdk/program/src/message/v0.rs
```diff
@@ -0,0 +1,383 @@
...
+                    return Err(SanitizeError::IndexOutOfBounds);
+                }
+            }
+        }
+
+        Ok(())
+    }
+}
+
+/// Address map specifies which entries to load and
```
sakridge, line 110: and?

sdk/program/src/message/versions.rs
jstarry: Agreed, I spent a fair bit of time trying to "peek" at the first element to avoid this but couldn't get to a good solution. I'll add a comment in the `Message` struct pointing to this code.

sdk/program/src/message/v0.rs
jstarry: Fixed, thanks

jstarry:
Any other concerns here before merging? I have changes ready that build on this

jstarry:
@mvines any high level feedback on this?

sdk/program/src/message/versions.rs
```diff
@@ -0,0 +1,269 @@
...
+/// # Serialization
+///
+/// If the first bit is set, the remaining 7 bits will be used to determine
+/// which message version is serialized starting from version `0`. If the first
+/// is bit is not set, all bytes are used to encode the original `Message`
+/// format.
+#[frozen_abi(digest = "5hAeXsHoMFfGREfh8RqJYmAjRcs88sT1YxrpJJq3Ne1f")]
+#[derive(Debug, PartialEq, Eq, Clone, AbiEnumVisitor, AbiExample)]
+pub enum VersionedMessage {
+    Original(Message),
```
mvines, line 32: I found the word "original" confusing as to me "v0" also implies original.  How about renaming "Original" to "Unversioned" or "Legacy"

mvines (commented):
lgtm, all I have is a nit about the usage of the term "original"

sdk/program/src/message/versions.rs
jstarry: Legacy is better, updated

mvines (dismissed):

PR #18727, opened 2021-07-16
Use rustup default profile in travis ci mac image

CriesofCarrots:
#### Problem
Mac release artifacts are failing to build, due to rustfmt requirement in `solana-storage-proto`. The Travis CI rust image uses the minimal `rustup` profile, which doesn't include rustfmt.

#### Summary of Changes
Switch to default `rustup` profile


t-nelson (approved):
:shipit:

Issue #18732, opened 2021-07-16
Transaction timing used unchecked math


jackcmay:
#### Problem

The calculation of transaction timings uses unchecked/unsaturated math

Here are a couple of examples:
https://github.com/solana-labs/solana/blob/2ec81f627d478186461cd0a7985170edd6d7b6a1/runtime/src/bank.rs#L173
https://github.com/solana-labs/solana/blob/2ec81f627d478186461cd0a7985170edd6d7b6a1/runtime/src/bank.rs#L3315
https://github.com/solana-labs/solana/blob/2ec81f627d478186461cd0a7985170edd6d7b6a1/runtime/src/message_processor.rs#L92

#### Proposed Solution

The microsecond timings are for debugging only so I'm not too concerned.  The compute budget counting will be used by the cost model so we may want to safeguard the math from weird behavior.

@CriesofCarrots , you raised the flag, thoughts?
@jeffwashington you use these metrics extensively, thoughts?

jeffwashington:
seems like saturation is the best thing to do. I can implement that if it satisfies your concerns.

jackcmay:
That will also result in bad values, maybe because we know these should never overflow, throw out any entries that do?

jeffwashington:
ah, bad values for your computations? I was assuming metrics that saturated would just be recognized as such and be inconvenient and lose information. I don't expect the metrics to overflow for the purpose of monitoring and perf work. What do you propose we do? Maybe I misunderstood your question.

jeffwashington:
'throw out' means ignore in your computations? So, if we saturated, we'd get a metric showing u64::max or whatever and we'd know it was 'big'. And you'd just ignore u64::max?

jackcmay:
Yeah, true, `u64::MAX` is easier to identify as long as we look at it pre `/ count`

CriesofCarrots:
While I don't feel I have a complete view of how these are used, saturation makes sense to me.

jackcmay:
Resolved via #18735

PR #18735, opened 2021-07-16
Use saturating math for timings

jackcmay:
#### Problem

https://github.com/solana-labs/solana/issues/18732

#### Summary of Changes

Use saturating math for transaction timings

Fixes #

### DIFF
--- a/runtime/src/message_processor.rs
+++ b/runtime/src/message_processor.rs
@@ -76,7 +76,6 @@ pub struct ExecuteDetailsTimings {
     pub data_size_changed: usize,
     pub per_program_timings: HashMap<Pubkey, ProgramTiming>,
 }
-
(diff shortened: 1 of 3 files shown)


runtime/src/message_processor.rs
```diff
@@ -90,10 +89,18 @@ impl ExecuteDetailsTimings {
         for (id, other) in &other.per_program_timings {
             let program_timing = self.per_program_timings.entry(*id).or_default();
             program_timing.accumulated_us += other.accumulated_us;
```
jeffwashington, line 91: why not this one?

jackcmay: Missed that one, thanks!

jeffwashington (approved):
lgtm

PR #18736, opened 2021-07-16
add getMultipleAccountsInfo which uses the getMultipleAccounts RPC method

tomjohn1028:
#### Problem
Web3.js is missing the [`getMultipleAccounts`](https://docs.solana.com/developing/clients/jsonrpc-api#getmultipleaccounts) RPC method

#### Summary of Changes
Add `getMultipleAccountsInfo` to the Connection class

Fixes #

### DIFF
--- a/web3.js/src/connection.ts
+++ b/web3.js/src/connection.ts
@@ -2408,6 +2408,28 @@ export class Connection {
   }
 
   /**
+   * Fetch all the account info for multiple accounts specified by an array of public keys
+   */
+  async getMultipleAccountsInfo(
+    publicKeys: PublicKey[],
+    commitment?: Commitment,
+  ): Promise<AccountInfo<Buffer>[] | null> {
+    const keys = publicKeys.map(key => key.toBase58());
+    const args = this._buildArgs([keys], commitment, 'base64');
+    const unsafeRes = await this._rpcRequest('getMultipleAccounts', args);
+    const res = create(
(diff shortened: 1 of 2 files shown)


web3.js/src/connection.ts
```diff
@@ -2407,6 +2407,32 @@ export class Connection {
...
 
+  /**
+   * Fetch all the account info for multiple accounts specified by an array of public keys
+   */
+  async getMultipleAccountsInfo(
+    publicKeys: PublicKey[],
+    commitment?: Commitment,
+  ): Promise<AccountInfo<Buffer>[] | null> {
+    const keys = publicKeys.map(key => key.toBase58());
+    try {
```
oJshua, line 2418: This outer try block isn't necessary. (based on convention established by other calls and redundant throw if 'error' exists in res)

oJshua (changes requested):
Thanks for adding this! Just one minor change and would you mind adding a quick test?

tomjohn1028:
@oJshua I made the requested changes, but it seems the `npm run test:live` command fails. I'm not up to speed on the different test commands/environments. Care to help me bring this home? Or provide some more context so I can understand what's missing?

oJshua:
Thanks @tomjohn1028 for your effort on this. I simply added a couple airdrops using the helpers.

oJshua (approved):

PR #18748, opened 2021-07-17
Add some docs for RpcClient and friends

brson:
#### Problem

Rust APIs are underdocumented.

#### Summary of Changes

I am aiming to fill out some of the Rust API docs, and am starting with RpcClient. This PR adds documentation and examples for the RpcClient constructors, partial documentation for the RpcClient type, some additional documentation for types related to RpcClient.

It also adds some examples for a few non-constructor RpcClient methods, for which I have not written corresponding docs yet, but will.

I hope to get feedback on how this is going, and continue with more patches like this.

I am learning Solana as I write, and some things are more obvious than others. In this patch, please pay attention to the description of the `confirm_transaction_initial_timeout` of `new_with_timeouts_and_commitment`. I have only read the code and attempted to interpret what it is doing, but am not confident about it.

### DIFF
--- a/client/src/rpc_client.rs
+++ b/client/src/rpc_client.rs
@@ -1,3 +1,12 @@
+//! Communication with a Solana node over RPC.
+//!
+//! Software that interacts with the Solana blockchain,
+//! whether querying its state or submitting transactions,
+//! communicates with a Solana node over [JSON-RPC],
+//! using the [`RpcClient`] type.
+//!
+//! [JSON-RPC]: https://www.jsonrpc.org/specification
+
 #[allow(deprecated)]
 use crate::rpc_deprecated_config::{
     RpcConfirmedBlockConfig, RpcConfirmedTransactionConfig,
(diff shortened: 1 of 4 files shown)


client/src/rpc_client.rs
```diff
@@ -1,3 +1,12 @@
+//! Communication with a Solana node over RPC.
+//!
+//! Software that interacts with the Solana blockchain,
+//! whether querying its state or submitting transactions,
+//! communicates with a Solana node over [JSON-RPC],
+//! using the [`RpcClient`] type.
```
mvines, line 6: It looks like your editor is set to line break around 60 characters or so?  In general I'd prefer an 80 character limit, for example:
```suggestion
//! Software that interacts with the Solana blockchain, whether querying its
//! state or submitting transactions, communicates with a Solana node over
//! [JSON-RPC], using the [`RpcClient`] type.
```
to reduce the number of comment lines.

mvines (commented):
Thank you, this is a wonderful start to better documenting the `solana-client` crate.  I only have a minor nit about comment line length

client/src/rpc_client.rs
brson: I'll reflow it.

It's not that I'm breaking at 60 chars, but that I'm breaking at single clauses within sentences. I find this easier to maintain and diff, but reading it in source is ugly. Just my habit.

brson: Done.

brson:
@mvines I've reflowed the paragraphs in the docs to 80 columns. Thanks for the review.

mvines (approved):

CriesofCarrots:
Let's v1.7 this; I want this on docs.rs :)

PR #18750, opened 2021-07-18
correct typo

yangli-io:
#### Problem
Docs Typo

#### Summary of Changes
* Change "be" to "the"
* Change "less then" to "less than"


mvines:
Thanks!

PR #18762, opened 2021-07-19
Update simulateTransaction rpc docs

jackcmay:
#### Summary of Changes

Add documentation about `unitsConsumed`

Fixes #

### DIFF
--- a/docs/src/developing/clients/jsonrpc-api.md
+++ b/docs/src/developing/clients/jsonrpc-api.md
@@ -3282,7 +3282,7 @@ The result will be an RpcResponse JSON object with `value` set to a JSON object 
     - `data: <[string, encoding]|object>`, data associated with the account, either as encoded binary data or JSON format `{<program>: <state>}`, depending on encoding parameter
     - `executable: <bool>`, boolean indicating if the account contains a program \(and is strictly read-only\)
     - `rentEpoch: <u64>`, the epoch at which this account will next owe rent, as u64
-- `unitsConsumed: <u64>`, The number of compute budget units consumed during the processing of this transaction
+- `unitsConsumed: <u64 | null>`, The number of compute budget units consumed during the processing of this transaction
 
 #### Example:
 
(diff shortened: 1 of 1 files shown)


docs/src/developing/clients/jsonrpc-api.md
```diff
@@ -3282,6 +3282,7 @@ The result will be an RpcResponse JSON object with `value` set to a JSON object
     - `data: <[string, encoding]|object>`, data associated with the account, either as encoded binary data or JSON format `{<program>: <state>}`, depending on encoding parameter
     - `executable: <bool>`, boolean indicating if the account contains a program \(and is strictly read-only\)
     - `rentEpoch: <u64>`, the epoch at which this account will next owe rent, as u64
+- `unitsConsumed: <u64>`, The number of compute budget units consumed during the processing of this transaction
```
CriesofCarrots, line 3285:
```suggestion
- `unitsConsumed: <u64 | null>`, The number of compute budget units consumed during the processing of this transaction
```

docs/src/developing/clients/jsonrpc-api.md
```diff
@@ -3282,6 +3282,7 @@ The result will be an RpcResponse JSON object with `value` set to a JSON object
     - `data: <[string, encoding]|object>`, data associated with the account, either as encoded binary data or JSON format `{<program>: <state>}`, depending on encoding parameter
     - `executable: <bool>`, boolean indicating if the account contains a program \(and is strictly read-only\)
     - `rentEpoch: <u64>`, the epoch at which this account will next owe rent, as u64
+- `unitsConsumed: <u64 | null>`, The number of compute budget units consumed during the processing of this transaction
```
mvines, line 3285: nit: "undefined" right?  (older RPC nodes just won't include this field at all, which is expressed by "undefined" instead of "null")

CriesofCarrots: Sorry, that's correct. Should be `undefined`

PR #18768, opened 2021-07-19
Rename BpfComputeBudget

jackcmay:
#### Problem

Compute budget is no longer BPF specific

#### Summary of Changes

Rename it

Fixes #


program-test/src/lib.rs
```diff
@@ -523,8 +522,15 @@ impl ProgramTest {
     }
 
     /// Override the BPF compute budget
```
mvines, line 524:
```suggestion
    /// Override the default maximum compute units
```

mvines (dismissed):
lgtm, just a comment nit

PR #18776, opened 2021-07-20
Broadcasting stage quick hacks v1.7

ryoqun:
#### Problem

image

#### Summary of Changes

Fixes #


core/src/broadcast_stage/standard_broadcast_run.rs
```diff
@@ -330,31 +330,49 @@ impl StandardBroadcastRun {
...
         trace!("Broadcasting {:?} shreds", shreds.len());
         // Get the list of peers to broadcast to
         let mut get_peers_time = Measure::start("broadcast::get_peers");
-        let now = timestamp();
-        let last = self.last_peer_update.load(Ordering::Relaxed);
         #[allow(deprecated)]
-        if now - last > BROADCAST_PEER_UPDATE_INTERVAL_MS
-            && self
-                .last_peer_update
-                .compare_and_swap(now, last, Ordering::Relaxed)
```
ryoqun, line 348: ugh: https://doc.rust-lang.org/std/sync/atomic/struct.AtomicUsize.html#method.compare_and_swap

sakridge:
Nice, yea getting the cluster_info access out of broadcast should really help the timings. Should we put this in a normal service? I was thinking you could have it update right before the leader slot too. Use poh_recorder.would_be_leader to detect it and update a slot or 2 beforehand.

core/src/broadcast_stage/standard_broadcast_run.rs
sakridge: https://github.com/solana-labs/solana/pull/18808

core/src/broadcast_stage/standard_broadcast_run.rs
```diff
@@ -330,31 +335,69 @@ impl StandardBroadcastRun {
...
+                        let (ci, ss) = recv.lock().unwrap().recv().unwrap();
+                        if now
+                            .map(|n: Instant| {
+                                n.elapsed().as_secs() > BROADCAST_PEER_UPDATE_INTERVAL_MS / 1000
+                            })
+                            .unwrap_or(true)
+                        {
+                            now = Some(Instant::now());
+
+                            let mut update_time = Measure::start("cluster-node-update");
```
ryoqun, line 378: here

PR #18780, opened 2021-07-20
feature: add new token program feature

joncinque:
#### Problem

We need to deploy a new token program.

#### Summary of Changes

Add a feature for it!  Since we already have a new token program id in the code, and that one was enabled at slot 66528004, and we're at 87877768 as of this writing, it felt ok to change this up.

On the other hand, I left the old feature in the feature set so it can still be queried.  Let me know if you prefer to remove that too.

After, this will reuse the hack to the account db to force in the new token program for one of our validators: https://github.com/solana-labs/solana/pull/15212#issuecomment-776141581 <--- this is more for me to remember

Fixes #


mvines (dismissed):
Removing/reusing the existing token upgrade infra makes sense to me

mvines:
@joncinque - when we have the final SPL Token binary, let's make a SPL Token release over in SPL-land as well.

Then replace `program-test/src/programs/spl_token-3.1.0.so` and update `fetch-spl.sh`.  These changes probably belong in this PR as well actually

CriesofCarrots:
> let's make a SPL Token release over in SPL-land as well

Yes, please! We need that for some ledger-app-solana changes I have cued. I'll probably also bump spl-token in `solana-transaction-status` to parse the new instruction.

joncinque:
Ah yes, great call! Should this next release be version 3.2.0 then?

joncinque:
@CriesofCarrots I added the transaction parser with the upgrade, could you take a look at it?

Otherwise, just double-checking, is there a potential problem with backporting this to 1.6?  spl-token v3.2.0 uses solana-program v1.7.4

CriesofCarrots:
> @CriesofCarrots I added the transaction parser with the upgrade, could you take a look at it?

Thank you! 😍 Looks great.

> Otherwise, just double-checking, is there a potential problem with backporting this to 1.6? spl-token v3.2.0 uses solana-program v1.7.4

Oof. Correct, this won't be backportable to v1.6 for the reason you noted. I guess this will need to head to MNB on the v1.7 train. (Rolling back SPL to v1.6 doesn't seem very feasible to me.)

joncinque:
> Oof. Correct, this won't be backportable to v1.6 for the reason you noted.

Ah, beans.  The program will soak on mainnet for a bit, so let's hope lines up with the timeline for 1.7.

CriesofCarrots:
@joncinque , actually... while I believe I've had troubles in the past with spl dependencies bringing in newer minor-version solana-program crates, in this case it looks like if we release a new version of `spl-associated-token-account` and bump that as well, the backport will work. v1.6 builds with that setup, anyway. Will have to check to ensure there are no crates.io packaging issues.

CriesofCarrots:
No harm in releasing a new `solana-associated-token-account` on spl-token v3.2 anyway, so let's give it a try!

joncinque:
All right, let's roll the dice!

joncinque:
That should be everything now :crossed_fingers:

CriesofCarrots:
@joncinque , alas, it does build fine with the new atoken, but doesn't please `cargo package`:
```
$ cargo package
   Packaging solana-account-decoder v1.6.19 (/solana/account-decoder)
   Verifying solana-account-decoder v1.6.19 (/solana/account-decoder)
    Updating crates.io index
    
error: failed to verify package tarball

Caused by:
  failed to select a version for `solana-program`.
      ... required by package `spl-token v3.2.0`
      ... which is depended on by `solana-account-decoder v1.6.19 (/Users/tyeraeulberg/Documents/Solana/solana/target/package/solana-account-decoder-1.6.19)`
  versions that meet the requirements `^1.7.4` are: 1.7.6, 1.7.5, 1.7.4

  all possible versions conflict with previously selected packages.

    previously selected package `solana-program v1.6.19`
      ... which is depended on by `solana-sdk v1.6.19`
      ... which is depended on by `solana-account-decoder v1.6.19 (/Users/tyeraeulberg/Documents/Solana/solana/target/package/solana-account-decoder-1.6.19)`

  failed to select a version for `solana-program` which could resolve this conflict
```

Sorry for the hassle

joncinque:
Ah bummer, oh well, it was worth a shot!

joncinque:
Merging on not-really-red, the removal of the v1.6 label causes this to look failed, when it isn't

PR #18651, opened 2021-07-14
Accounts db replication design proposal

lijunwangs:
#### Problem
Validators fall behind the network when bogged down by heavy RPC load. This
seems to be due to a combination of CPU load and lock contention caused by
serving RPC requests. The most expensive RPC requests involve account scans.

#### Summary of Changes

The AccountsDb replication design proposal is described.

Fixes #

PR #18782, opened 2021-07-20
Add new rpc modules and split apis

CriesofCarrots:
#### Problem
https://github.com/solana-labs/solana/pull/18651 describes a need for a new kind of node that serves accountsDB rpc requests only. However, the majority of our rpc APIs are tied up in one `rpc_full` module, which will prevent handing any of them off to another rpc service (eg the proposed `JsonRpcAccountsService`)

#### Summary of Changes
Split up `rpc_full` according to the data source required:
- `rpc_bank` implements the APIs that only require data from a Bank's direct fields. I envision that once #18651 is implemented, this module would be serviced by both accounts-replica and full API nodes.
- `rpc_accounts` implements the APIs that require AccountsDB data. This module would be serviced by only accounts-replica nodes.
- `rpc_full` implements the remaining exiting APIs, which use other data sources (including gossip, Blockstore, and Bigtable) or interact with or depend on other services (eg. SendTransactionService, MaxSlots tracking), which would be serviced by full API nodes

There is no Validator or JsonRpcService functionality change in this PR

Toward #16986

### DIFF
--- a/rpc/src/rpc.rs
+++ b/rpc/src/rpc.rs
@@ -2345,37 +2345,13 @@ pub mod rpc_minimal {
     }
 }
 
-// Full RPC interface that an API node is expected to provide
-// (rpc_minimal should also be provided by an API node)
-pub mod rpc_full {
+// RPC interface that only depends on immediate Bank data
+// Expected to be provided by both API nodes and (future) accounts replica nodes
+pub mod rpc_bank {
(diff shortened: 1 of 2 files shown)


mvines (approved):

lijunwangs (approved):
Looks good

PR #18784, opened 2021-07-20
implements fifo rwlock

behzadnouri:
#### Problem
Reader threads can run ahead of writer thread causing write-starvation.

#### Summary of Changes
This commit implements a fifo rwlock by adding an additional queue of
condition variables. Each thread:
  - adds its condition variable to this queue.
  - waits until it is head of the queue.
  - obtains the read or write lock on underlying std::sync::RwLock.
  - removes itself from the queue.
  - and, notifies the next in line that it is now head of the queue.

### DIFF
--- a/gossip/src/sync.rs
+++ b/gossip/src/sync.rs
@@ -1,0 +1,151 @@
+use std::{
+    collections::VecDeque,
+    sync::{
+        Arc, Condvar, LockResult, Mutex, PoisonError, RwLockReadGuard, RwLockWriteGuard,
+        TryLockError, TryLockResult,
+    },
+};
+
+/// FIFO reader-writer lock.
+#[derive(Default)]
+pub struct RwLock<T> {
+    inner: std::sync::RwLock<T>,
+    queue: Mutex<VecDeque<Arc<Condvar>>>,
+}
+
+impl<T> RwLock<T> {
+    pub fn new(t: T) -> Self {
+        Self {
+            inner: std::sync::RwLock::new(t),
+            queue: Mutex::default(),
+        }
+    }
+
+    pub fn read(&self) -> LockResult<RwLockReadGuard<'_, T>> {
+        use std::sync::RwLock;
+        self.lock(RwLock::read, RwLock::try_read)
+    }
+
+    pub fn write(&self) -> LockResult<RwLockWriteGuard<'_, T>> {
+        use std::sync::RwLock;
+        self.lock(RwLock::write, RwLock::try_write)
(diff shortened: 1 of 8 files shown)


behzadnouri:
This is `replay-loop-voting-stats::vote_push_us` (sum over 16 seconds). Left chunk is this code.
The metric was removed in https://github.com/solana-labs/solana/pull/18552 but here I ran the validator without that commitjust for testing the impact on that call-site.
https://github.com/solana-labs/solana/blob/e6f32f6ae/core/src/replay_stage.rs#L1754-L1756

replay-loop-voting-stats vote_push_us_fifo_16s

gossip/src/sync.rs
```diff
@@ -0,0 +1,151 @@
...
+            // Check if it is in fact in front of the queue;
+            // Otherwise this has been a spurious wakeup.
+        }
+        // Should drop queue lock here, so that other threads are not blocked
+        // and are queued in order of their arrival.
+        drop(queue);
+        let guard = lock(&self.inner);
+        // Pop self from the queue, and notify the next one in line that it is
+        // now head of the queue.
+        let cvar: Arc<Condvar> = match self.queue.lock() {
```
sakridge, line 76: if multiple readers are queued, would they then be evaluated serially?

sakridge:
Really cool, functionally it looks good to me. Any downsides? I guess the per-lock cost is probably a bit higher to deal with the queue/arc construction.

gossip/src/sync.rs
behzadnouri: If `r1, r2, w1, r3` are in the queue (with `r1` being the head of the queue), then `r1` and `r2` will share the lock but `r3` has to wait until `w1` is done because of fifo.
So readers will share the lock as long as there is no writer between them.

behzadnouri (commented):
> Any downsides?

The downside is that concurrency is reduced. If the rwlock is locked for read, and a writer is already waiting, then new readers have to wait until the writer is done first. Whereas previously new readers will run ahead of the waiting writer and share the lock with the existing readers. So potentially more readers were previously running concurrently, at the cost of writer-starvation.

Having tested this more, it seems like some of the metrics related to read-locks have actually become worse. Previously they were quite fast because apparently they were piggy backing on some other readers already having the lock. But now they have to wait if a writer is ahead of them and waiting.

So, overall, I am not sure if this commit is a right trade-off. Seems like we would rather lean towards more concurrency as long as writes can be done async or if writers can wait.

PR #18785, opened 2021-07-21
Fix race in remove_unrooted_race and flush_slot_cache

carllin:
#### Problem
`test_cache_flush_remove_unrooted_race` is flaky https://buildkite.com/solana-labs/solana/builds/50349#b8b9efb2-8ace-496e-b975-e40a2a2ca3f4, where this assertion fails: https://github.com/solana-labs/solana/blob/395ee4ec68043ecd5022cf9da1678cfeec57fdfa/runtime/src/accounts_db.rs#L4215. This is because between these two checks, https://github.com/solana-labs/solana/blob/395ee4ec68043ecd5022cf9da1678cfeec57fdfa/runtime/src/accounts_db.rs#L4232-L4233, `remove_unrooted_slots()` could have removed the slot from the cache

#### Summary of Changes
Fixup order of checking, introduce new test.

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -4229,22 +4229,32 @@ impl AccountsDb {
         slot: Slot,
         should_flush_f: Option<&mut impl FnMut(&Pubkey, &AccountSharedData) -> bool>,
     ) -> Option<FlushStats> {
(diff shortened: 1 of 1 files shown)


lijunwangs (approved):
Looks good to me

PR #18804, opened 2021-07-21
feature: add new token program feature (v1.6 backport of #18780)

joncinque:
#### Problem

We need the new token program feature in 1.6, to have it running on a hacked validator (at least).  And depending on the release cadence for 1.7, this gives us the possibility of using the new token program separately.

#### Summary of Changes

Add a (limited) manual backport of #18780 

Fixes #

### DIFF
--- a/runtime/src/inline_spl_token_v2_0.rs
+++ b/runtime/src/inline_spl_token_v2_0.rs
@@ -2,7 +2,7 @@
 solana_sdk::declare_id!("TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA");
 
 pub(crate) mod new_token_program {
-    solana_sdk::declare_id!("t31zsgDmRntje65uXV3LrnWaJtJJpMd4LyJxq2R2VrU");
+    solana_sdk::declare_id!("NToKV6K2hAz79S73NtC9vVWrAGn77mBduBn95xQFGSZ");
 }
 
 /*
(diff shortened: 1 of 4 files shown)


joncinque:
Since the crate bump was done in #18805 , you can focus on the first two commits, 3fbf219 and d3a9557, which were cherry-picked from #18780

runtime/src/inline_spl_token_v2_0.rs
```diff
@@ -2,7 +2,7 @@
 solana_sdk::declare_id!("TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA");
 
 pub(crate) mod new_token_program {
-    solana_sdk::declare_id!("t31zsgDmRntje65uXV3LrnWaJtJJpMd4LyJxq2R2VrU");
+    solana_sdk::declare_id!("NToKV6K2hAz79S73NtC9vVWrAGn77mBduBn95xQFGSZ");
```
CriesofCarrots, line 5: Ha, I just noticed your pubkeys ;)

CriesofCarrots (approved):
r+ with rebase

runtime/src/inline_spl_token_v2_0.rs
joncinque: hehe I'm glad! It's not clear whether "NToK" should be "Not Token" or "New Token", and I like that

PR #18805, opened 2021-07-21
1.6: Bump crates to 1.6.20

joncinque:
#### Problem

There's an error in #18804 due to crates not being bumped up to 1.6.20

#### Summary of Changes

Bump them!

Fixes #


CriesofCarrots (approved):
Thanks for picking up!

PR #18806, opened 2021-07-21
adds feature for allowing private addresses for ci tests

behzadnouri:
#### Problem
To address mainnet issue, https://github.com/solana-labs/solana/pull/18728 excludes private ip addresses.
As a side effect, `system-performance-tests` broke:
https://github.com/solana-labs/solana/pull/18728#issuecomment-883848858

#### Summary of Changes
* This commit adds a _development only_ build feature `allow_private_addr` to allow private ip address for development and ci tests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2317,6 +2317,25 @@ dependencies = [
 
 [[package]]
 name = "libsecp256k1"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bd1137239ab33b41aa9637a88a28249e5e70c40a42ccc92db7f12cc356c1fcd7"
+dependencies = [
+ "arrayref",
+ "base64 0.12.3",
+ "digest 0.9.0",
+ "hmac-drbg",
+ "libsecp256k1-core",
+ "libsecp256k1-gen-ecmult",
+ "libsecp256k1-gen-genmult",
+ "rand 0.7.3",
+ "serde",
+ "sha2 0.9.5",
+ "typenum",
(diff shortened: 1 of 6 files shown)


mvines (changes requested):
FTR: I'm strongly opposed to Cargo features like this. It was hard work to rid the tree of them earlier in the project.

behzadnouri:
> FTR: I'm strongly opposed to Cargo features like this. It was hard work to rid the tree of them earlier in the project.

It would be helpful if you can please clarify what the concerns are.

This feature will never be used for an actual prod build, but only for development work and ci tests. I am guessing this is unlike the previous features that have been removed. So maybe the concerns justifying removing those earlier features will not necessarily carry over to this use-case.
We can even make the feature name more explicit like `TEST_ONLY_allow_private_addr` or something like that.

The downsides with runtime flags and arguments is that:
* It will be a runtime "if" branch that has to be checked on "every" packet.
* It will create a lot of noise in the code for something which will never have a production utility but only used in development and ci tests.
* It will add room for bugs because of the added noise and because more places in the code now have to make sure to pass in the right value.

mvines:
> It would be helpful if you can please clarify what the concerns are.

Build complexity and build times are the main concern.  Now we have multiple binaries that we need to build and test, and worse cargo incremental builds do not handle features nicely cross crates (even our internal solana crates).  

Worse, we now need to rebuild depending on the network configuration of the cluster.  The way is runtime configuration.  Sucks that in this case there's plumbing that may be required to accomplish this but that's probably a refactoring opportunity regardless

PR #18807, opened 2021-07-21
Fixed broken links and added more context to getTokenAccountBalance docs

murlokito:
#### Problem

Some links in the `getConfirmedBlock` docs were broken

#### Summary of Changes

Fixed those links and added more context to the `getTokenAccountBalance` method docs which can come in handy when implementing clients where shared code can be achieved


mvines:
Thanks for the doc fixes!

PR #18808, opened 2021-07-21
Write helper for multithread update

sakridge:
#### Problem

Common pattern of using at atomic to report stats a single time for multiple thread workers.

#### Summary of Changes

Helper class reduces boilerplate for this pattern and makes it easier to update to best-practices.

Fixes #


jeffwashington (approved):
lgtm

PR #18812, opened 2021-07-21
Introduce AncestorHashesService

carllin:
#### Problem
Need a service to initiate ancestor hash repairs and process ancestor hash responses

#### Summary of Changes
Introduce such a service

Fixes #

### DIFF
--- a/core/src/ancestor_hashes_service.rs
+++ b/core/src/ancestor_hashes_service.rs
@@ -2,14 +2,14 @@ use crate::{
     cluster_slots::ClusterSlots,
     duplicate_repair_status::DeadSlotAncestorRequestStatus,
     outstanding_requests::OutstandingRequests,
-    repair_response,
+    repair_response::{self},
     repair_service::{DuplicateSlotsResetSender, RepairInfo, RepairStatsGroup},
     replay_stage::DUPLICATE_THRESHOLD,
     result::{Error, Result},
     serve_repair::{AncestorHashesRepairType, ServeRepair},
 };
-use dashmap::DashMap;
-use solana_ledger::blockstore::Blockstore;
+use dashmap::{mapref::entry::Entry::Occupied, DashMap};
+use solana_ledger::{blockstore::Blockstore, shred::SIZE_OF_NONCE};
 use solana_measure::measure::Measure;
 use solana_perf::{packet::limited_deserialize, recycler::Recycler};
 use solana_runtime::bank::Bank;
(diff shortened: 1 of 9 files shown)


core/src/ancestor_hashes_service.rs
```diff
@@ -0,0 +1,364 @@
...
+    ) -> Result<()> {
+        let timeout = Duration::new(1, 0);
+        let mut responses = vec![response_receiver.recv_timeout(timeout)?];
+        let mut total_packets = responses[0].packets.len();
+
+        let mut dropped_packets = 0;
+        while let Ok(more) = response_receiver.try_recv() {
+            total_packets += more.packets.len();
+            if total_packets < *max_packets {
+                // Drop the rest in the channel in case of DOS
```
AshwinSekar, line 221: Probably unlikely but if for whatever reason we have 50 ancestor slot repairs and we receive all 21 samples for each at the same time what would happen here? Do we have any case to ask for the missing samples again?

core/src/ancestor_hashes_service.rs
```diff
@@ -0,0 +1,364 @@
...
+                            // means the earliest mismatched ancestor has yet to be found.
+                            //
+                            // In the best case scenario, this means after ReplayStage dumps
+                            // the earliest known ancestor `A` here, and then repairs `A`,
+                            // because we may still have the incorrect version of some ancestor
+                            // of `A`, we will mark `A` as dead and then continue the search
+                            // protocol through another round of ancestor repairs.
+                            //
+                            // However this process is a bit slow, so in an ideal world, the
+                            // protocol could be extended to keep searching by making
```
AshwinSekar, line 310: what's stopping us from checking `ContinueSearch` here and sending another request?

core/src/ancestor_hashes_service.rs
```diff
@@ -0,0 +1,364 @@
...
+        _ancestor_hashes_request_socket: Arc<UdpSocket>,
+        _repair_info: RepairInfo,
+        _outstanding_requests: Arc<RwLock<OutstandingAncestorHashesRepairs>>,
+        exit: Arc<AtomicBool>,
+    ) -> JoinHandle<()> {
+        let mut repair_stats = AncestorRepairRequestsStats::default();
+
+        Builder::new()
+            .name("solana-find-repairable-dead-slots".to_string())
+            .spawn(move || loop {
```
AshwinSekar, line 355: this coming in the next pr i assume?

core/src/ancestor_hashes_service.rs
carllin: hmmmm I think the ancestor request in the pending queue would time out and and we would try again

core/src/ancestor_hashes_service.rs
carllin: yup!

core/src/ancestor_hashes_service.rs
carllin: I think there's a race somewhere, but I can't think of it off the top of my head xD

Also more complexity for v1 to avoid.

PR #18831, opened 2021-07-22
Create get_epoch_info.js

amilz:
Adding to example repository. Thanks to Chase for sharing on SolHack discord.

#### Problem
no problem. just adding examples for others to learn.

#### Summary of Changes
added get_epoch_info.js, an example of how to use getEpochInfo();

Fixes #

### DIFF
--- a/web3.js/examples/get_epoch_info.js
+++ b/web3.js/examples/get_epoch_info.js
@@ -1,0 +1,19 @@
+import * as solanaWeb3 from '@solana/web3.js';
+
+export async function logEpochInfo () {
+    let conn = new solanaWeb3.Connection(clusterApiUrl('mainnet-beta')); //mainnet-beta or devnet -- connects to the network
+    let epochinfo = await conn.getEpochInfo(); 
+    /*returns an object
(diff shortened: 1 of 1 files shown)


jstarry:
Hey @amilz, I don't think we need an example for this but your explanations of the return values are really helpful! Would you be willing to improve the docs we have here: https://solana-labs.github.io/solana-web3.js/modules.html#EpochInfo by adding comments here:
https://github.com/solana-labs/solana/blob/71bd434297ba060a4246199f01039d23a5e8ba1a/web3.js/src/connection.ts#L371-L381

amilz:
> Hey @amilz, I don't think we need an example for this but your explanations of the return values are really helpful! Would you be willing to improve the docs we have here: https://solana-labs.github.io/solana-web3.js/modules.html#EpochInfo by adding comments here:
> https://github.com/solana-labs/solana/blob/71bd434297ba060a4246199f01039d23a5e8ba1a/web3.js/src/connection.ts#L371-L381

Oh good. This is what I was looking for. Sorry about that!

PR #18849, opened 2021-07-22
Move tower save into the VotingService

mvines:
Building on #18847, the first commit in this PR moves tower saving into the VotingService to avoid blocking replay stage progress.  

This is not super important with the current `FileTowerStorage` implementation, but once `EtcdTowerStorage` is implemented then it'll be no good to be blocking replay stage with network activity with an etcd cluster


mvines:
@sakridge - wdyt?

PR #18850, opened 2021-07-22
adds validator flag to allow private ip addresses

behzadnouri:
#### Problem
To address mainnet issue, https://github.com/solana-labs/solana/pull/18728 excludes private ip addresses.
As a side effect, `system-performance-tests` broke:
https://github.com/solana-labs/solana/pull/18728#issuecomment-883848858

#### Summary of Changes
* added validator flag to allow private ip addresses.

### DIFF
--- a/validator/src/bin/solana-test-validator.rs
+++ b/validator/src/bin/solana-test-validator.rs
@@ -21,6 +21,7 @@ use {
         signature::{read_keypair_file, write_keypair_file, Keypair, Signer},
         system_program,
     },
+    solana_streamer::socket::SocketAddrSpace,
     solana_validator::{
         admin_rpc_service, dashboard::Dashboard, println_name_value, redirect_stderr_to_file,
         test_validator::*,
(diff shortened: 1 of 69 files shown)


validator/src/bin/solana-test-validator.rs
```diff
@@ -21,6 +21,7 @@ use {
         signature::{read_keypair_file, write_keypair_file, Keypair, Signer},
         system_program,
     },
+    solana_streamer::socket::SocketAddrSpace,
```
mvines, line 24: I think `solana-test-validator` should just allow private addresses by default (and it's probably fine to not have an option to only allow global this).

mvines (approved):
💦 💦 That touched a lot more files than I was expecting it would!

But thanks, that was annoying I'm sure but the lack of a build feature is worth it.  

r+ with `solana-test-validator` supporting private addresses by default

validator/src/bin/solana-test-validator.rs
behzadnouri: yeah, true. I am merging this now not to get merge conflicts, but will follow up.

PR #18851, opened 2021-07-22
add getFees with test

Arrowana:
#### Problem
No getFees

#### Summary of Changes
Add getFees

Fixes #

### DIFF
--- a/web3.js/src/connection.ts
+++ b/web3.js/src/connection.ts
@@ -708,6 +708,17 @@ export type PerfSample = {
   samplePeriodSecs: number;
 };
 
+export type Fees = {
+  /** Recent blockhash */
+  blockhash: Blockhash;
+  /** Calculator for transaction fees */
+  feeCalculator: FeeCalculator;
+  /** @deprecated this value is inaccurate and should not be relied upon */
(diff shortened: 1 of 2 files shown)


oJshua (approved):

oJshua:
@Arrowana This looks great. Could you rename your commit for https://github.com/solana-labs/solana/pull/18851/commits/7ffe40a7b46035270ccabad8f5e364e456c66af6 to `feat: add getFees with test` just so we can verify CI passes fully?

oJshua:
Thanks again for your contribution on this @Arrowana. I'm just learning that this endpoint is likely set to be deprecated, so we won't be bringing this PR in.

Some more details can be found here (https://github.com/solana-labs/solana/pull/18875)

Arrowana:
Ok the reason i need this is because:
https://discordapp.com/channels/428295358100013066/517163444747894795/864632440865030174
"For the blockhash that you used, save the associated lastValidBlockHeight.  Once the current block height is greater, the transaction is invalid if not yet confirmed"
I want better guarantees on the state of a transaction, where can i get `lastValidBlockHeight` then, no other endpoint offers it?

PR #18869, opened 2021-07-23
Bumps solana_rbpf to v0.2.14

### LINKED ISSUE #17520: Unnecessary BPF verifier restriction on `MUL64_IMM`
#### Problem

The BPF verifier enforces a non-zero imm for the `MUL64_IMM` instruction when it doesn't need to here: https://github.com/solana-labs/solana/blob/6b9d8d41a3fd32a069812bf0ec7d6a392414d874/programs/bpf_loader/src/bpf_verifier.rs#L228

#### Proposed Solution

Featurize and remove this restriction

### LINKED ISSUE #18683: BPF program consumes the entire budget on panic with jit
#### Problem

When running a local node with jit and executing the [panic](https://github.com/solana-labs/solana/tree/master/programs/bpf/rust/panic) bpf program the runtime reports that the entire compute budget was consumed when actually only a few thousand were.

This behavior has been observed on mainnet, its not clear there if the issue is only with panic or other program exits as well.

#### Proposed Solution

This appears to have something to do with how jitt'd programs account for compute budget on exit.

This can be recreated by running a local node with jit enabled, deploying the panic program, and sending it an instruction with no accounts and no instruction data.  This should also be recreatable in a unit test but I was not able to observe that.

Lichtso:
#### Summary of Changes
See [RBPF release](https://github.com/solana-labs/rbpf/releases/tag/v0.2.14)

Fixes #17520
Fixes #18683


Lichtso:
@jackcmay Can you see if the feature gate looks about right?

If so you can merge.

jackcmay:
Sounds good, reviewing and will commit if it looks good, thanks!

programs/bpf_loader/src/syscalls.rs
```diff
@@ -2515,18 +2520,22 @@ mod tests {
...
         let memory_mapping = MemoryMapping::new::<UserError>(
-            vec![MemoryRegion {
-                host_addr: addr,
-                vm_addr: 100,
-                len: std::mem::size_of::<Pubkey>() as u64,
-                vm_gap_shift: 63,
-                is_writable: false,
-            }],
+            vec![
+                MemoryRegion::default(),
```
jackcmay, line 2524: What is driving these changes to the tests?

jackcmay: The reason I ask is because the the memory mapping `map` call changed which could be a change in behavior.  Looks like the vm address's must now be in the known ranges that the vm uses and can no longer be arbitrary values.  This looks fine since the VM doesn't use these arbitrary vm addresss, but want to make sure.

programs/bpf_loader/src/lib.rs
```diff
@@ -83,6 +82,8 @@ pub fn create_executor(
         max_call_depth: compute_budget.max_call_depth,
         stack_frame_size: compute_budget.stack_frame_size,
         enable_instruction_tracing: log_enabled!(Trace),
+        verify_mul64_imm_nonzero: !invoke_context
+            .is_feature_active(&stop_verify_mul64_imm_nonzero::id()), // TODO: Feature gate and then remove me
         ..Config::default()
```
jackcmay, line 87: Are the stack frame gaps disabled by default now?  If so they should be featurized

jackcmay: Is the Config's `instruction_meter_checkpoint_distance` default the same as v0.2.14?  If not, also should be feature gated

Lichtso: Still enabled by default: `enable_stack_frame_gaps: true`
https://github.com/solana-labs/rbpf/blob/302de4d10888dfb76534693308f4ef8060b86fcd/src/vm.rs#L206
Should we disable it in the next release?

Lichtso: `instruction_meter_checkpoint_distance` should have no influence on what the program sees, the point where a programs is aborted and what error is throws has to be the same for interpreter and JIT anyway. The option only makes the detection of that condition faster (so it affects the performance only).

programs/bpf_loader/src/syscalls.rs
Lichtso: Yes, the layout of `MemoryRegion`s in the `MemoryMapping` is now stricter so the tests had to be adjusted. We did not take the memory layout from the ELF and instead enforced a static one. So I don't think anything changed for the programs and they are still running on my validator node.

programs/bpf_loader/src/lib.rs
jackcmay: Sounds good as long as there isn't an edge case that's falls in the middle.  Also, maybe check with Tao about cost model stuff to make sure this change doesn't result in a behavioral change there.

PR #18870, opened 2021-07-23
Handle cleaning of zero-lamport accounts w.r.t. Incremental Snapshots

### LINKED ISSUE #18825: [Incremental Snapshots] Handle cleaning zero-lamport accounts in slots higher than the last full snapshot slot
Just like the title says. Created this issue so its easier to track in the larger #17088 issue.

brooksprumo:
#### Problem

Zero-lamport accounts are improperly cleaned up in the context of incremental snapshots.

#### Summary of Changes

- Add `Option<Slot>` parameter to `AccountsDb::clean_accounts()` for the `last_full_snapshot_slot`. If `Some`, then `purges_zero_lamports` will be checked against this slot. Any purges for slots higher than the last full snapshot slot will be filtered out (i.e. _not_purged).
- Add test to AccountsDb for cleaning zero-lamport accounts with the `last_full_snapshot_slot` parameter
- Add test to snapshot_utils for cleaning zero-lamport accounts and doing incremental snapshots
- Update all uses of `clean_accounts()`. Most all callers pass `None` now.

Fixes #18825

### DIFF
--- a/runtime/src/accounts_background_service.rs
+++ b/runtime/src/accounts_background_service.rs
@@ -160,7 +160,7 @@ impl SnapshotRequestHandler {
                 // accounts that were included in the bank delta hash when the bank was frozen,
                 // and if we clean them here, the newly created snapshot's hash may not match
                 // the frozen hash.
-                snapshot_root_bank.clean_accounts(true, false);
+                snapshot_root_bank.clean_accounts(true, false, None);
                 clean_time.stop();
 
                 if accounts_db_caching_enabled {
@@ -399,7 +399,7 @@ impl AccountsBackgroundService {
                                 // slots >= bank.slot()
                                 bank.force_flush_accounts_cache();
                             }
-                            bank.clean_accounts(true, false);
+                            bank.clean_accounts(true, false, None);
                             last_cleaned_block_height = bank.block_height();
                         }
                     }
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -12275,4 +12275,74 @@ pub mod tests {
         accounts_db.clean_accounts(None, false, Some(slot3));
         assert_eq!(accounts_db.ref_count_for_pubkey(&pubkey), 0);
     }
+
+    #[test]
+    fn test_filter_zero_lamport_clean_for_incremental_snapshots() {
+        solana_logger::setup();
+        let accounts_db = AccountsDb::new_single_for_tests();
+        let pubkey = solana_sdk::pubkey::new_rand();
+        let slot = 10;
+        let account_info = AccountInfo {
+            store_id: 42,
+            offset: 123,
+            stored_size: 234,
+            lamports: 0,
+        };
+        let store_count = 0;
+        let mut key_set = HashSet::default();
+        key_set.insert(pubkey);
+        let mut store_counts = HashMap::default();
+        store_counts.insert(account_info.store_id, (store_count, key_set));
+
+        // scenario 1: zero lamport account GREATER THAN last full snapshot slot => remove from purges set
+        {
+            let max_clean_root = None;
+            let last_full_snapshot_slot = Some(slot - 1);
+            let mut purges_zero_lamports = HashMap::default();
+            purges_zero_lamports.insert(pubkey, (vec![(slot, account_info.clone())], 1));
+
+            accounts_db.filter_zero_lamport_clean_for_incremental_snapshots(
+                max_clean_root,
+                last_full_snapshot_slot,
+                &store_counts,
+                &mut purges_zero_lamports,
+            );
+
+            assert!(!purges_zero_lamports.contains_key(&pubkey));
+        }
+
+        // scenario 2: zero lamport account EQUAL TO last full snapshot slot => keep in purges set
+        {
+            let max_clean_root = None;
+            let last_full_snapshot_slot = Some(slot);
+            let mut purges_zero_lamports = HashMap::default();
+            purges_zero_lamports.insert(pubkey, (vec![(slot, account_info.clone())], 1));
+
+            accounts_db.filter_zero_lamport_clean_for_incremental_snapshots(
+                max_clean_root,
+                last_full_snapshot_slot,
+                &store_counts,
+                &mut purges_zero_lamports,
+            );
+
+            assert!(purges_zero_lamports.contains_key(&pubkey));
+        }
+
+        // scenario 3: zero lamport account LESS THAN full snapshot slot => keep in purges set
+        {
+            let max_clean_root = None;
+            let last_full_snapshot_slot = Some(slot + 1);
+            let mut purges_zero_lamports = HashMap::default();
+            purges_zero_lamports.insert(pubkey, (vec![(slot, account_info)], 1));
+
+            accounts_db.filter_zero_lamport_clean_for_incremental_snapshots(
+                max_clean_root,
+                last_full_snapshot_slot,
+                &store_counts,
+                &mut purges_zero_lamports,
+            );
+
+            assert!(purges_zero_lamports.contains_key(&pubkey));
+        }
+    }
 }
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -2113,16 +2113,18 @@ impl AccountsDb {
     /// When using incremental snapshots, do not purge zero-lamport accounts if the slot is higher
     /// than the last full snapshot slot.  This is to protect against the following scenario:
     ///
-    ///      A full snapshot is taken, and it contains an account with a non-zero balance.  Later,
-    ///      that account's  goes to zero.  Evntually cleaning runs, and before, this account would
-    ///      be cleaned up.  Finally, an incremental snapshot is taken.
+    ///   ```text
+    ///   A full snapshot is taken, and it contains an account with a non-zero balance.  Later,
+    ///   that account's  goes to zero.  Evntually cleaning runs, and before, this account would be
+    ///   cleaned up.  Finally, an incremental snapshot is taken.
     ///
-    ///      Later, the incremental (and full) snapshot is used to rebuild the bank and accounts
-    ///      database (e.x. if the node restarts).  The full snapshot _does_ contain the account
-    ///      (from above) and its balance is non-zero, however, since the account was cleaned up in
-    ///      a later slot, the incremental snapshot does not contain any info about this account,
-    ///      thus, the accounts database will contain the old info from this account, which has its
-    ///      old non-zero balance.  Very bad!
+    ///   Later, the incremental (and full) snapshot is used to rebuild the bank and accounts
+    ///   database (e.x. if the node restarts).  The full snapshot _does_ contain the account (from
+    ///   above) and its balance is non-zero, however, since the account was cleaned up in a later
+    ///   slot, the incremental snapshot does not contain any info about this account, thus, the
+    ///   accounts database will contain the old info from this account, which has its old non-zero
+    ///   balance.  Very bad!
+    ///   ```
     ///
     /// This filtering step can be skipped if there is no `last_full_snapshot_slot`, or if the
     /// `max_clean_root` is less-than-or-equal-to the `last_full_snapshot_slot`.
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -24,7 +24,8 @@ use crate::{
     accounts_hash::{AccountsHash, CalculateHashIntermediate, HashStats, PreviousPass},
     accounts_index::{
         AccountIndexGetResult, AccountSecondaryIndexes, AccountsIndex, AccountsIndexRootsStats,
-        IndexKey, IsCached, ScanResult, SlotList, SlotSlice, ZeroLamport, BINS_FOR_TESTING,
+        IndexKey, IsCached, RefCount, ScanResult, SlotList, SlotSlice, ZeroLamport,
+        BINS_FOR_TESTING,
     },
     ancestors::Ancestors,
     append_vec::{AppendVec, StoredAccountMeta, StoredMeta, StoredMetaWriteVersion},
@@ -977,6 +978,10 @@ pub struct AccountsDb {
     /// such that potentially a 0-lamport account update could be present which
     /// means we can remove the account from the index entirely.
     dirty_stores: DashMap<(Slot, AppendVecId), Arc<AccountStorageEntry>>,
+
+    /// Zero-lamport accounts that are *not* purged during clean because they need to stay alive
+    /// for incremental snapshot support.
+    zero_lamport_accounts_to_purge_after_full_snapshot: DashSet<(Slot, Pubkey)>,
 }
 
 #[derive(Debug, Default)]
@@ -1410,6 +1415,7 @@ impl AccountsDb {
             remove_unrooted_slots_synchronization: RemoveUnrootedSlotsSynchronization::default(),
             shrink_ratio: AccountShrinkThreshold::default(),
             dirty_stores: DashMap::default(),
+            zero_lamport_accounts_to_purge_after_full_snapshot: DashSet::default(),
         }
     }
 
@@ -1735,6 +1741,7 @@ impl AccountsDb {
     fn construct_candidate_clean_keys(
         &self,
         max_clean_root: Option<Slot>,
+        last_full_snapshot_slot: Option<Slot>,
         timings: &mut CleanKeyTimings,
     ) -> Vec<Pubkey> {
         let mut dirty_store_processing_time = Measure::start("dirty_store_processing");
@@ -1783,9 +1790,27 @@ impl AccountsDb {
         timings.delta_key_count = pubkeys.len() as u64;
 
         let mut hashset_to_vec = Measure::start("flat_map");
-        let pubkeys: Vec<Pubkey> = pubkeys.into_iter().collect();
+        let mut pubkeys: Vec<Pubkey> = pubkeys.into_iter().collect();
         hashset_to_vec.stop();
         timings.hashset_to_vec_us += hashset_to_vec.as_us();
+
+        // Check if we should purge any of the zero_lamport_accounts_to_purge_later, based on the
+        // last_full_snapshot_slot.
+        assert!(
+            last_full_snapshot_slot.is_some() || self.zero_lamport_accounts_to_purge_after_full_snapshot.is_empty(),
+            "if snapshots are disabled, then zero_lamport_accounts_to_purge_later should always be empty"
+        );
+        if let Some(last_full_snapshot_slot) = last_full_snapshot_slot {
+            self.zero_lamport_accounts_to_purge_after_full_snapshot
+                .retain(|(slot, pubkey)| {
+                    let is_candidate_for_clean =
+                        max_slot >= *slot && last_full_snapshot_slot >= *slot;
+                    if is_candidate_for_clean {
+                        pubkeys.push(*pubkey);
+                    }
+                    !is_candidate_for_clean
+                });
+        }
 
         pubkeys
     }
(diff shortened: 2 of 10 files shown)


brooksprumo:
@carllin @lijunwangs @ryoqun This PR is now ready for review, but it is built on top of PR #18565. So this PR currently contains extra changes that will go away once the other PR is merged in.

Feel free to review these changes, which are mostly in `AccountsDb::clean_accounts()`. If you need to pick just one, the other PR is a better use of time. Once that PR is in, and this one gets rebased, I'll flip it to Ready For Review and add you all as reviewers.

Thanks in advance!

runtime/src/accounts_db.rs
```diff
@@ -1934,6 +1939,46 @@ impl AccountsDb {
...
+                        // understand.
+                        let cannot_purge =
+                            account_info.is_zero_lamport() && *slot > last_full_snapshot_slot;
+                        !cannot_purge
+                    })
+                });
+            }
+        }
+        measure_purge_filter_zero_lamports.stop();
+
```
carllin, line 1981: let's extract this into a separate function behind a feature flag for incremental snapshots so that we can toggle off incremental snapshot cleaning in case we need to roll this back.

runtime/src/accounts_background_service.rs
```diff
@@ -160,7 +160,7 @@ impl SnapshotRequestHandler {
                 // accounts that were included in the bank delta hash when the bank was frozen,
                 // and if we clean them here, the newly created snapshot's hash may not match
                 // the frozen hash.
-                snapshot_root_bank.clean_accounts(true, false);
+                snapshot_root_bank.clean_accounts(true, false, None);
```
carllin, line 163: thinking we maybe want to be able to toggle this `last_full_snapshot_slot` based on a validator flag that we can use to toggle incremental snapshots on/off

runtime/src/accounts_db.rs
```diff
@@ -1934,6 +1939,46 @@ impl AccountsDb {
...
+                        // full snapshot slot.  Since we're `retain`ing the accounts-to-purge, I
+                        // felt creating the `cannot_purge` variable made this easier to
+                        // understand.
+                        let cannot_purge =
+                            account_info.is_zero_lamport() && *slot > last_full_snapshot_slot;
+                        !cannot_purge
+                    })
+                });
+            }
+        }
```
carllin, line 1979: I think we can fold this filter into the `retain` above: https://github.com/solana-labs/solana/pull/18870/files#diff-1090394420d51617f3233275c2b65ed706b35b53b115fe65f82c682af8134a6fR1932-R1939. I think a flag like `should_filter_for_incremental_snapshots = `if let Some(last_full_snapshot_slot) = last_full_snapshot_slot && max_clean_root.unwrap_or(Slot::MAX) > last_full_snapshot_slot {`, then toggling the rest of this logic based on that flag would work.

runtime/src/snapshot_utils.rs
```diff
@@ -2814,4 +2820,211 @@ mod tests {
...
+    /// Test that cleaning works well in the edge cases of zero-lamport accounts and snapshots.
+    /// Here's the scenario:
+    ///
+    /// slot 1:
+    ///     - send some lamports to Account1 (from Account2) to bring it to life
+    ///     - take a full snapshot
+    /// slot 2:
+    ///     - make Account1 have zero lamports (send back to Account2)
+    ///     - take an incremental snapshot
+    ///     - ensure deserializing from this snapshot is equal to this bank
```
carllin, line 2833: I think a separate good test case is to have some additional, non-zero lamport updates to `Account1`, and ensure that the non-zero lamport update is cleaned up by clean, even though the zero lamport account itself wasn't cleaned

runtime/src/snapshot_utils.rs
```diff
@@ -2814,4 +2820,211 @@ mod tests {
...
+        .unwrap();
+        assert_eq!(
+            deserialized_bank, *bank4,
+            "Ensure rebuilding from an incremental snapshot works",
+        );
+        assert_eq!(
+            deserialized_bank.get_account(&key1.pubkey()),
+            None,
+            "Ensure Account1 has not been brought back from the dead"
+        );
```
carllin, line 3028: Unfortunately, this check doesn't check that the backing index entry has been cleaned. This is because when you call `get_account()` through the bank, it funnels to this call here: https://github.com/solana-labs/solana/blob/0247c280ffb420c758142dfa6cea0af0581d0849/runtime/src/accounts.rs#L484, which filters out zero lamport accounts :)

carllin: If this test now fails, this may be because `key1` is not found in this set here: https://github.com/solana-labs/solana/blob/0247c280ffb420c758142dfa6cea0af0581d0849/runtime/src/accounts_db.rs#L1790, so it never makes it into the zero-lamport set. 

`construct_candidate_clean_keys` gets created from `uncleaned_pubkeys` or extracted from the stores in `dirty_stores`. `uncleaned_pubkeys` gets updated on bank freeze via https://github.com/solana-labs/solana/blob/0247c280ffb420c758142dfa6cea0af0581d0849/runtime/src/accounts_db.rs#L5119, and `dirty_stores` is updated whenever an entire slot is removed from the database (dead) or we add a new root for a slot. 

Unfortunately, because the zero-lamport account will not be updated again, it won't be caught in any of the above and it doesn't make it into the `uncleaned_pubkeys` set.

We used to keep track of all the zero-lamport accounts and cleaned them on every iteration. This proved to be expensive however, hence the the comment from this PR: https://github.com/solana-labs/solana/pull/17601#discussion_r645362108, and why ultimately this was removed https://github.com/solana-labs/solana/pull/17601/files#diff-db6927be895a2b4d140c54b84392a861b4535e91afc466c5e93c243c6bf2b28aL1387.

carllin: Maybe next to `uncleaned_pubkeys` we could introduce a `skipped_clean_zero_lamport_pubkeys` that we add to when we filter out the zero lamport update, and then consume it on the *first* clean after each full snapshot?

runtime/src/accounts_background_service.rs
brooksprumo: Sure, that works. Can you point me to code/documentation on how to do that?

Also, this change doesn't actually enable incremental snapshots; there's no functionality change. So if you'd like we can wait to add the toggle at the same time I add actual incremental snapshot support to ABS.

runtime/src/accounts_db.rs
brooksprumo: OK will do.

runtime/src/snapshot_utils.rs
brooksprumo: Sure thing! Can you point me to code/docs for checking if updates are cleaned up?

Also, maybe it makes sense to put this as a test in AccountsDb, because these checks don't sound related to snapshots to me, more the clean logic directly. What do you think?

runtime/src/accounts_background_service.rs
carllin: >Can you point me to code/documentation on how to do that

It would pretty much be another option added here: https://github.com/solana-labs/solana/blob/v1.7.9/core/src/validator.rs#L91, and then plumbed through the system

>Also, this change doesn't actually enable incremental snapshots; there's no functionality change. So if you'd like we can wait to add the toggle at the same time I add actual incremental snapshot support to ABS.

That's fine!

runtime/src/snapshot_utils.rs
carllin: > Can you point me to code/docs for checking if updates are cleaned up

Yeah I think this is close: https://github.com/solana-labs/solana/blob/v1.7.9/runtime/src/accounts_db.rs#L8138-L8155. You'll have to read out the account index and check the slot list to see the update is gone I think.

> Also, maybe it makes sense to put this as a test in AccountsDb, because these checks don't sound related to snapshots to me, more the clean logic directly. What do you think?

Agreed, they should go in `AccountsDb`

runtime/src/snapshot_utils.rs
```diff
@@ -2980,7 +2980,7 @@ mod tests {
         bank4.squash();
         bank4.clean_accounts(true, false, Some(full_snapshot_slot));
         assert_eq!(
-            bank4.get_account(&key1.pubkey()),
+            bank4.get_account_modified_slot(&key1.pubkey()),
```
carllin, line 2983: I think we can simplify to `assert!(bank4.get_account_modified_slot(&key1.pubkey()).is_none())`

runtime/src/snapshot_utils.rs
```diff
@@ -3022,7 +3022,7 @@ mod tests {
             "Ensure rebuilding from an incremental snapshot works",
         );
         assert_eq!(
-            deserialized_bank.get_account(&key1.pubkey()),
+            deserialized_bank.get_account_modified_slot(&key1.pubkey()),
```
carllin, line 3025: I think we can simplify to assert!(deserialized_bank.get_account_modified_slot(&key1.pubkey()).is_none())

runtime/src/accounts_background_service.rs
brooksprumo: Thanks! Yes, I'll be adding incremental snapshot config stuff to the SnapshotConfig struct, which'll then go in to the validator CLI args. That'll handle toggling incremental snapshots on and off. And I'll default to "off" to begin with.

runtime/src/snapshot_utils.rs
carllin: I think that's because the earlier call to `bank_to_incremental_snapshot_archive()` has already removed the zero lamport account via the `clean_accounts()` call in that function. I think this was verified here: https://github.com/solana-labs/solana/pull/18870/commits/411f9da0f99fe8a685c54a455fe78751048b0179#diff-c957621ae265a12e03f7b9c5656e7aa74137223e603cc3cfcc2828dda1af5479R2964-R2973

I also added a separate clean after the bank was recreated: https://github.com/solana-labs/solana/pull/18870/commits/411f9da0f99fe8a685c54a455fe78751048b0179#diff-c957621ae265a12e03f7b9c5656e7aa74137223e603cc3cfcc2828dda1af5479R2964-R2973, this should remove the pubkey from the dirty list, and make the later clean fail, highlighting the issue described in my comment above

runtime/src/snapshot_utils.rs
brooksprumo: Ah yes, good catch!

runtime/src/snapshot_utils.rs

runtime/src/snapshot_utils.rs

brooksprumo:
Moving this back to Draft as I iterate on it.

runtime/src/snapshot_utils.rs
brooksprumo: Done!

runtime/src/snapshot_utils.rs

brooksprumo:
@carllin OK, PR should be good-to-review again! I've attempted to put the correct tests/checks in place, but I may've missed some. Hopefully this is moving in the right direction though!

runtime/src/accounts_db.rs
```diff
@@ -970,6 +970,10 @@ pub struct AccountsDb {
     /// such that potentially a 0-lamport account update could be present which
     /// means we can remove the account from the index entirely.
     dirty_stores: DashMap<(Slot, AppendVecId), Arc<AccountStorageEntry>>,
+
+    /// Zero-lamport accounts that are *not* purged during clean because they need to stay alive
+    /// for incremental snapshot support.
+    zero_lamport_accounts_to_purge_later: DashSet<(Slot, Pubkey)>,
```
carllin, line 976: nit: `zero_lamport_accounts_to_purge_later -> zero_lamport_accounts_to_purge_after_full_snapshot`

runtime/src/accounts_db.rs
```diff
@@ -1770,18 +1776,41 @@ impl AccountsDb {
...
+        let mut pubkeys: Vec<Pubkey> = pubkeys.into_iter().collect();
         hashset_to_vec.stop();
         timings.hashset_to_vec_us += hashset_to_vec.as_us();
 
+        // Check if we should purge any of the zero_lamport_accounts_to_purge_later, based on the
+        // last_full_snapshot_slot.
+        assert!(
+            last_full_snapshot_slot.is_some() || self.zero_lamport_accounts_to_purge_later.is_empty(),
+            "if snapshots are disabled, then zero_lamport_accounts_to_purge_later should always be empty"
+        );
```
carllin, line 1788: 👍

runtime/src/accounts_db.rs
```diff
@@ -1925,16 +1958,63 @@ impl AccountsDb {
...
+                // not get purged here are added to a list so they be considered for purging later
+                // (i.e. after the next full snapshot).
+                let cannot_purge =
+                    account_info.is_zero_lamport() && *slot > last_full_snapshot_slot.unwrap();
+                if cannot_purge {
+                    self.zero_lamport_accounts_to_purge_later
+                        .insert((*slot, *pubkey));
+                }
+                !cannot_purge
+            })
```
carllin, line 2017: let's extract this into a separate function `filter_zero_lamport_clean_for_incremental_snapshots()`

runtime/src/accounts_db.rs
```diff
@@ -1925,16 +1958,63 @@ impl AccountsDb {
...
+                .max_by_key(|(slot, _account_info)| slot);
+
+            slot_account_info_at_highest_slot.map_or(true, |(slot, account_info)| {
+                // Do *not* purge zero-lamport accounts if the slot is greater than the last full
+                // snapshot slot.  Since we're `retain`ing the accounts-to-purge, I felt creating
+                // the `cannot_purge` variable made this easier to understand.  Accounts that do
+                // not get purged here are added to a list so they be considered for purging later
+                // (i.e. after the next full snapshot).
+                let cannot_purge =
+                    account_info.is_zero_lamport() && *slot > last_full_snapshot_slot.unwrap();
```
carllin, line 2011: From where the `purges_zero_lamports` was constructed: https://github.com/solana-labs/solana/blob/v1.7/runtime/src/accounts_db.rs#L1811-L1819, I think it must be true that `account_info.is_zero_lamport()`, so I think we should assert that here.

runtime/src/accounts_db.rs
```diff
@@ -1925,16 +1958,63 @@ impl AccountsDb {
...
+                if cannot_purge {
+                    self.zero_lamport_accounts_to_purge_later
+                        .insert((*slot, *pubkey));
```
carllin, line 2014: looks like the crux of it 😃

runtime/src/accounts_db.rs
```diff
@@ -1925,16 +1958,63 @@ impl AccountsDb {
...
+            //      A full snapshot is taken, and it contains an account with a non-zero balance.
+            //      Later, that account's  goes to zero.  Evntually cleaning runs, and before, this
+            //      account would be cleaned up.  Finally, an incremental snapshot is taken.
+            //
+            //      Later, the incremental (and full) snapshot is used to rebuild the bank and
+            //      accounts database (e.x. if the node restarts).  The full snapshot _does_
+            //      contain the account (from above) and its balance is non-zero, however, since
+            //      the account was cleaned up in a later slot, the incremental snapshot does not
+            //      contain any info about this account, thus, the accounts database will contain
+            //      the old info from this account, which has its old non-zero balance.  Very bad!
```
lijunwangs, line 1996: Is this issue only limited to zero lamports? What about for older roots with non-zero lamports which are cleaned up because they were superseded by updates in later slots?

lijunwangs: For example for an account it is updated at slot 1, slot 2 and slot 3 with non zero lamports. We take full snapshot at slot 1, accounts cleanup happens at slot 3 (slot2) can be cleaned up as changes superseded by update in slot 3. And later an incremental snapshot is taken at slot 2. Can this happen?

runtime/src/accounts_db.rs

runtime/src/accounts_db.rs
brooksprumo: Good call. Done!

runtime/src/accounts_db.rs
brooksprumo: Correct, the issue only applies to zero-lamport accounts. When an account has a non-zero balance, it will not be removed when clean is called, which is the crux of the issue.

This is a good question about snapshots that occur at slots lower than the max rooted slot. This is the same issue that could occur today without incremental snapshot support. Since clean only occurs in the background services which also handle snapshots, this does two things: (1) clean knows what slot it is snapshot-ing, and (2) what slot the bank/bank forks is currently on. These two can make sure that clean doesn't remove dead stores before snapshots complete.

Or at least that's my understanding of how this process works! I reserve the right to be wrong though :)

runtime/src/accounts_db.rs

runtime/src/accounts_db.rs
```diff
@@ -2077,6 +2107,76 @@ impl AccountsDb {
...
+    ///   database (e.x. if the node restarts).  The full snapshot _does_ contain the account (from
+    ///   above) and its balance is non-zero, however, since the account was cleaned up in a later
+    ///   slot, the incremental snapshot does not contain any info about this account, thus, the
+    ///   accounts database will contain the old info from this account, which has its old non-zero
+    ///   balance.  Very bad!
+    ///   ```
+    ///
+    /// This filtering step can be skipped if there is no `last_full_snapshot_slot`, or if the
+    /// `max_clean_root` is less-than-or-equal-to the `last_full_snapshot_slot`.
+    fn filter_zero_lamport_clean_for_incremental_snapshots(
```
carllin, line 2131: Let's add a unit test for some of the cases here, rest of the change looks good!

brooksprumo: Good call, will do!

brooksprumo: Done! Tests have been added. Let me know if there are other cases you'd like me to tests as well.

runtime/src/accounts_db.rs
lijunwangs: @carllin, @CriesofCarrots  What is your take? I agree that we may have the same issue with full snap. This is probably corrected by repair_stage later? But if the RPC do a scan at that slot, would not we give wrong information?

carllin: > For example for an account it is updated at slot 1, slot 2 and slot 3 with non zero lamports. We take full snapshot at slot 1, accounts cleanup happens at slot 3 (slot2) can be cleaned up as changes superseded by update in slot 3. And later an incremental snapshot is taken at slot 2. Can this happen

This can't currently happen because the clean is synchronous with the storage fetch in AccountsBackgroundService. This means we save off all storage entries necessary for snapshotting up to slot `X` before we clean past slot `X`. this guarantees the storage entry for slot 2 is available for the snapshot, and has not yet been removed.

runtime/src/accounts_db.rs
```diff
@@ -12275,4 +12275,74 @@ pub mod tests {
...
+
+            accounts_db.filter_zero_lamport_clean_for_incremental_snapshots(
+                max_clean_root,
+                last_full_snapshot_slot,
+                &store_counts,
+                &mut purges_zero_lamports,
+            );
+
+            assert!(purges_zero_lamports.contains_key(&pubkey));
+        }
```
carllin, line 12346: These three cases would be nice:

1) `max_clean_root` and `last_full_snapshot_slot` both `None`
2) `max_clean_root` and `last_full_snapshot_slot` both `Some`
3) `max_clean_root` `Some` and `last_full_snapshot_slot` `None`.

brooksprumo: Done! Should cover a lot more cases now.

runtime/src/accounts_db.rs
```diff
@@ -12117,11 +12217,196 @@ pub mod tests {
...
+            should_contain: bool,
+        }
+
+        let do_test = |test_params: TestParameters| {
+            let account_info = AccountInfo {
+                store_id: 42,
+                offset: 123,
+                stored_size: 234,
+                lamports: 0,
+            };
```
carllin, line 12296: nice, good abstraction

carllin (approved):

PR #18876, opened 2021-07-23
passes through --allow-private-addr to validators in system perf tests

behzadnouri:
#### Problem
`system-performance-tests` has been broken since https://github.com/solana-labs/solana/pull/18728

#### Summary of Changes
This commit passes through `--allow-private-addr` which was added in https://github.com/solana-labs/solana/pull/18850 down to validators in system perf tests.

Successful run of the buildkite job:
https://buildkite.com/solana-labs/system-performance-tests/builds/5129

### DIFF
--- a/gossip/src/main.rs
+++ b/gossip/src/main.rs
@@ -217,7 +217,7 @@ fn process_spy_results(
     }
 }
 
-fn process_spy(matches: &ArgMatches) -> std::io::Result<()> {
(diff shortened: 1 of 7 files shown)


behzadnouri:
@t-nelson @sakridge any objection to merging this?

t-nelson (approved):
lgtm!

PR #18893, opened 2021-07-26
drop outstanding_requests lock before sending repair requests

jbiseda:
- misc turbine metrics
- drop outstanding_requests lock before sending repair request packets


core/src/repair_service.rs
```diff
@@ -262,29 +268,51 @@ impl RepairService {
...
+                        &repair_info.repair_validators,
+                        &mut outstanding_requests,
+                    ) {
+                        Some((req, to))
+                    } else {
+                        None
+                    }
+                })
+                .collect();
+            drop(outstanding_requests);
```
behzadnouri, line 290: I think you can utilize `?` operator here. Also giving `outstanding_requests` a scope will avoid manual `drop`. So something like:
```rust
let batch: Vec<_> = {
  let mut outstanding_requests = outstanding_requests.write().unwrap();
  repairs
    .into_iter()
    .filter_map(|repair_request| {
        let (to, req) = serve_repair.repair_request(...).ok()?;
        Some((req, to))
    })
    .collect()
};
```

core/src/window_service.rs
```diff
@@ -39,6 +40,34 @@ use std::{
 pub type DuplicateSlotSender = CrossbeamSender<Slot>;
 pub type DuplicateSlotReceiver = CrossbeamReceiver<Slot>;
 
+#[derive(Default)]
+pub struct WindowServiceMetrics {
+    pub run_insert_count: u64,
+    pub num_shreds_received: u64,
+    pub shred_receiver_elapsed_us: u64,
+    pub prune_shreds_elapsed_us: u64,
+}
```
behzadnouri, line 49: none seem to need `pub`

ledger/src/blockstore.rs
```diff
@@ -156,7 +156,7 @@ pub struct Blockstore {
     slots_stats: Arc<Mutex<SlotsStats>>,
 }
 
-struct SlotsStats {
+pub struct SlotsStats {
```
behzadnouri, line 159: why `pub`?

behzadnouri (approved):
LGTM

sakridge (approved):
lgtm

carllin (dismissed):

core/src/window_service.rs
jbiseda: removed pub

ledger/src/blockstore.rs
jbiseda: removed

core/src/repair_service.rs
jbiseda: That's more succinct, changed.

PR #18895, opened 2021-07-26
remove unused return value from account index upsert

jeffwashington:
#### Problem
removing this return value simplifies the api, allows optimization as the backing storage changes, and removes potential errors and testing burden.
#### Summary of Changes
remove an unused return value
Fixes #


brooksprumo (approved):
lgtm

PR #18896, opened 2021-07-26
during range, stop scanning earlier than currently scanning

jeffwashington:
#### Problem
Account index exists in bins. When scanning, we start with the bin that can satisfy the request. We also need to stop once we pass the bin that could possibly contain the end bound. As # bins increase, this becomes much more important.
#### Summary of Changes
If range is given an upper bound, stop scanning when we pass the bin that could contain the upper bound.
Fixes #

### DIFF
--- a/runtime/src/accounts_index.rs
+++ b/runtime/src/accounts_index.rs
@@ -541,14 +541,21 @@ impl<'a, T> AccountsIndexIterator<'a, T> {
         }
     }
 
+    fn bin_from_bound(bound: &Bound<Pubkey>, unbounded_bin: usize) -> usize {
+        match bound {
+            Bound::Included(bound) | Bound::Excluded(bound) => get_bin_pubkey(bound),
+            Bound::Unbounded => unbounded_bin,
+        }
+    }
+
     fn start_bin(&self) -> usize {
         // start in bin where 'start_bound' would exist
-        match &self.start_bound {
-            Bound::Included(start_bound) | Bound::Excluded(start_bound) => {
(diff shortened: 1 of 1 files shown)


brooksprumo (approved):
Nicely done! lgtm

runtime/src/accounts_index.rs
```diff
@@ -541,14 +541,35 @@ impl<'a, T> AccountsIndexIterator<'a, T> {
...
+        let start_bin = self.start_bin();
+        // calculate the max range of bins to look in
+        let end_bin_inclusive = self.end_bin_inclusive();
+        let bin_range = if start_bin > end_bin_inclusive {
+            0 // empty range
+        } else {
+            // the range is end_inclusive + 1 - start
+            // end_inclusive could be usize::MAX already if no bound was specified
+            end_bin_inclusive.saturating_add(1) - start_bin
+        };
```
brooksprumo, line 571: This may be an edge case that doesn't matter: If `end_bin_inclusive` is `usize::MAX` already, is that a valid bin? If yes, then I think this `saturating_add()` may produce a `bin_range` that is short by one. At the same time, if `start_bin` is `0` and `end_bin_inclusive` is `usize::MAX`, the `bin_range` should be `usize::MAX + 1`, but that's not a valid `usize` value.

Again, not sure how much this case matters. If bin `usize::MAX` is invalid, maybe moving to an "end-bin-_exclusive_" style would be an easy fix.

brooksprumo (approved):
lgtm

PR #18724, opened 2021-07-16
Backport cost model to v1.7

taozhu-chicago:
#### Problem
To prepare merge cost model into v1.7

#### Summary of Changes
cherry-picked following commits from master:
- b000d490ce3b798bdf32410985cd121b4e768d8b (cost model)
   - fixed conflicts related to `Secp256` changes and `serde_jason` in toml
- ae27fcbcda2721ad39301749c0262175a5e8f8ce (replay stage feedback)
- 9d6f1ebef489210e408eb9a4cbb21cd9059e43cb (perf tuning)
- 5e424826ba52e643bbd8e761b7bee11f699eb46c (persist costs to blockstore)
- 7cd6224cafb637357ae7e28043e763ce37bfe031 (Senders don't panic)
- 0e039b4094c263211047ad15ae27f4bc1073a28b (aggregate cost model to cost tracker)
- b6dff1292378c65bef8bc0d01bd8071b7ecdbef3 (update ledger tool)
   - for some reason, had to manually add back a big chuck of code to compute_slot_cost subcommand
- f97ce2cd7e77742dc4a1598116142f0238559fcd (pre program execute timing)

and two more picked from PR brach: https://github.com/solana-labs/solana/pull/18703
- 4d00aa9960323befd54e88e0d47cbee14eb05cfa (featurizing cost mode)
- b126e97a5c7dbbc67e80bd84a919740004ae99e8 (fix merge conflict)

also stub out bank_hash_cf from master (PR #18533) for backward compat commit f2b012abdf8e42751829a41ccd0ef3defe306cb6)

Fixes #18729

PR #18840, opened 2021-07-22
Exclude stubbed ProgramCosts column from compaction

CriesofCarrots:
#### Problem
Booting a v1.6 node with an existing ledger fails on compaction:
```
thread '<unnamed>' panicked at 'not implemented', ledger/src/blockstore_db.rs:801:9
```
The code excluding the ProgramCosts column from compaction wasn't included in https://github.com/solana-labs/solana/pull/18743 (fyi @taozhu-chicago )

#### Summary of Changes
Exclude stubbed ProgramCosts column from compaction

cc @joncinque

Issue #18729, opened 2021-07-16
solana-validator ledger/ is incompatible between master and v1.7/1.6 due to `program_costs` column


mvines:
A ledger/ that's created by a master validator will panic a v1.7/v1.6 validator:
```
thread 'main' panicked at 'Failed to open ledger database: RocksDb(Error { message: "Invalid argument: Column families not opened: program_costs" })', core/src/validator.rs:1065:6
```

STR to produce the panic:
```
$ solana-install init edge
$ solana-test-validator --reset
^C after it starts running
$ solana-install init beta
$ solana-test-validator --log
```

taozhu-chicago:
Going to stub-out/backport `program_cost` column family and `bank_hash` column family (from PR #18533 )  to v1.7 and v1.6.

ryoqun:
today i learnt: rocksdb all of cfs must be explicitly opened: https://github.com/facebook/rocksdb/blob/84eef260de9c0dc0c7ace8e6f605a2b3efd71038/db/version_edit_handler.cc#L360-L371

CriesofCarrots:
v1.6, v1.7, and master compatible as of column-family stubs ^

PR #18897, opened 2021-07-26
v1.7: backport new column families from master

CriesofCarrots:
#### Problem
New columns added to master creates ledgers that are incompatible with v1.7, as described in issue #18729.

#### Summary of Changes
Backport `bank_hash` and `program_costs` column families from master to 1.7 for backward compatibility.
Also excludes the stubbed ProgramCosts column from compaction, an oversight that footgunned v1.6 (cc #18840 )

Closes #18724

### DIFF
--- a/ledger/src/blockstore.rs
+++ b/ledger/src/blockstore.rs
@@ -124,6 +124,10 @@ pub struct BlockstoreSignals {
 }
 
 // ledger window
+//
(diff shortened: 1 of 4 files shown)


taozhu-chicago (approved):
looks good to me, thanks @CriesofCarrots

PR #18899, opened 2021-07-26
move AccountsIndex upsert into static WriteAccountMapEntry

jeffwashington:
#### Problem
accounts_index.upsert is the primary mechanism at runtime to add and update account info for a pubkey.
There are 2 patterns:

1. key doesn't exist, add it and set the info
2. key does exist
a. if there is already an entry for the slot, replace the info for that slot
b. if there is not already an entry for the slot, append the info for this new slot

There is one return value from this function, and it is 'reclaims'. If this return value did not exist, it would not be necessary to determine entirely which case we fall into above prior to being able to return.
It turns out that when we are calling upsert on a slot that is cached, then reclaims is asserted to be empty. Thus, the caller does not need reclaims to be filled out. As a result, there are optimizations we can make later with this information.
Step 1 in realizing these gains is to refactor AccountsIndex to support upsert deeper. This pr.
Step 2 is #18900
Step 3 occurs when it becomes more expensive and unnecessary to lookup a missing account or the existing account in disk-backed storage. This work can be delayed from the critical path, allowing upsert to return more quickly and still be correct.
#### Summary of Changes
refactor AccountsIndex to support upsert.
Fixes #

### DIFF
--- a/runtime/src/accounts_index.rs
+++ b/runtime/src/accounts_index.rs
@@ -219,7 +219,7 @@ impl<T: 'static + Clone + IsCached> WriteAccountMapEntry<T> {
         match w_account_maps.entry(*pubkey) {
             Entry::Occupied(mut occupied) => {
                 let current = occupied.get_mut();
-                Self::upsert_item(current, new_value, reclaims);
+                Self::upsert_item(current, &new_value, reclaims);
             }
             Entry::Vacant(vacant) => {
                 vacant.insert(new_value);
@@ -227,25 +227,25 @@ impl<T: 'static + Clone + IsCached> WriteAccountMapEntry<T> {
         }
     }
 
-    // returns None if upsert was successful
-    // otherwise, the pubkey has to be added to the maps with a write lock. call upsert_new
+    // returns true if upsert was successful. new_value is modified in this case. new_value contains a RwLock
+    // otherwise, new_value has not been modified and the pubkey has to be added to the maps with a write lock. call upsert_new
     pub fn upsert_existing_key<'a>(
         r_account_maps: AccountMapsReadLock<'a, T>,
         pubkey: &Pubkey,
-        new_value: AccountMapEntry<T>,
+        new_value: &AccountMapEntry<T>,
         reclaims: &mut SlotList<T>,
-    ) -> Option<AccountMapEntry<T>> {
+    ) -> bool {
         if let Some(current) = r_account_maps.get(pubkey) {
             Self::upsert_item(current, new_value, reclaims);
-            None
+            true
         } else {
-            Some(new_value)
+            false
         }
     }
 
     fn upsert_item(
         current: &Arc<AccountMapEntryInner<T>>,
-        new_value: AccountMapEntry<T>,
+        new_value: &AccountMapEntry<T>,
         reclaims: &mut SlotList<T>,
     ) {
         let mut slot_list = current.slot_list.write().unwrap();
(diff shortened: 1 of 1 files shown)


runtime/src/accounts_index.rs
```diff
@@ -204,30 +204,94 @@ impl<T: 'static + Clone + IsCached> WriteAccountMapEntry<T> {
...
+    }
+
+    // returns None if upsert was successful
+    // otherwise, the pubkey has to be added to the maps with a write lock. call upsert_new
+    pub fn upsert_existing_key<'a>(
+        r_account_maps: AccountMapsReadLock<'a, T>,
+        pubkey: &Pubkey,
+        new_value: AccountMapEntry<T>,
+        reclaims: &mut SlotList<T>,
+    ) -> Option<AccountMapEntry<T>> {
```
carllin, line 235: It seems more intuitive if this returned a boolean, true for success, false if a new entry needs to be created. `new_value` will have to be passed by reference though

runtime/src/accounts_index.rs
```diff
@@ -204,30 +204,94 @@ impl<T: 'static + Clone + IsCached> WriteAccountMapEntry<T> {
...
+            }
+            Entry::Vacant(vacant) => {
+                vacant.insert(new_value);
+            }
+        }
+    }
+
+    // returns None if upsert was successful
+    // otherwise, the pubkey has to be added to the maps with a write lock. call upsert_new
+    pub fn upsert_existing_key<'a>(
```
carllin, line 230: nit: upsert implies creation of a new key if it doesn't exist, maybe rename to `update_if_exists()`

runtime/src/accounts_index.rs
```diff
@@ -204,30 +204,94 @@ impl<T: 'static + Clone + IsCached> WriteAccountMapEntry<T> {
         })
     }
 
+    fn addref(item: &AtomicU64) {
+        item.fetch_add(1, Ordering::Relaxed);
+    }
+
+    pub fn upsert_new_key<'a>(
```
carllin, line 211: Because this function performs exactly the definition of  "upsert" , maybe rename to just `upsert()`

runtime/src/accounts_index.rs
```diff
@@ -204,30 +204,94 @@ impl<T: 'static + Clone + IsCached> WriteAccountMapEntry<T> {
...
+    ) -> Option<AccountMapEntry<T>> {
+        if let Some(current) = r_account_maps.get(pubkey) {
+            Self::upsert_item(current, new_value, reclaims);
+            None
+        } else {
+            Some(new_value)
+        }
+    }
+
+    fn upsert_item(
```
carllin, line 244: nit: `upsert_item` -> `lock_and_update_slot_list()` to distinguish from `update_slot_list` below.

runtime/src/accounts_index.rs
```diff
@@ -204,30 +204,94 @@ impl<T: 'static + Clone + IsCached> WriteAccountMapEntry<T> {
...
+        let (slot, new_entry) = new_value.slot_list.write().unwrap().remove(0);
+        let addref = Self::update_static(&mut slot_list, slot, new_entry, reclaims);
+        if addref {
+            Self::addref(&current.ref_count);
+        }
+    }
+
+    // modifies slot_list
+    // returns true if caller should addref
+    pub fn update_static(
```
carllin, line 259: nit: update_static() -> update_slot_list()

runtime/src/accounts_index.rs
jeffwashington: It is clearer. Since this new_value is still a RwLock, I can do this with a reference without incurring an unnecessary copy. The jury is still out on whether we retain a RwLock or not. Master still has it, so it makes sense to make a simpler api here. This may have to be revisited later. Compile errors will result, yay rust!

carllin (approved):

PR #18900, opened 2021-07-26
signal to upsert whether reclaims are important

jeffwashington:
#### Problem
upsert can be an insert or an update. Update semantics currently return to caller whether a reclaim exists or not. Determining this requires that we go to the account index, possibly on disk, to determine whether there already exists an entry for this pubkey and slot. Some code paths assert that there are no reclaims upon completion.
#### Summary of Changes
Move the assert into the accounts index, allowing the upsert to be async (completed later) in this case. Basically, with more information about what the caller needs, the implementation can optimize the implementation.
Fixes #


runtime/src/accounts_index.rs
```diff
@@ -263,18 +277,24 @@ impl<T: 'static + Clone + IsCached> WriteAccountMapEntry<T> {
         slot: Slot,
         account_info: T,
         reclaims: &mut SlotList<T>,
+        reclaims_must_be_empty: bool,
```
carllin, line 280: hmmm based on the usage of this flag in this function, because:
1) We never check that `reclaims.is_empty()`, we just never add to the `reclaims` list
2) This really just checks the previous entry was cached

is it more explicit to rename `reclaims_must_be_empty` to `previous_slot_entry_was_cached`? 

We also never check that `reclaims.is_empty()`, we just never add to `reclaims`, so

jeffwashington: ok, renamed

carllin:
I think the `runtime/src/hybrid_btree_map.rs ` was an accidental add?

jeffwashington:
> I think the `runtime/src/hybrid_btree_map.rs ` was an accidental add?

yep. good catch.

carllin (approved):

PR #18906, opened 2021-07-26
Added missing method to index and updated token account methods docs.

murlokito:
#### Problem

Following up with a bug report in [Solnet](https://github.com/bmresearch/Solnet/issues/164) regarding the `getTokenAccountsBy...` methods I was faced with a situation where the current docs on the token accounts misled us into believing the `delegatedAmount` property would be something like a `u64` when in reality it is the same object as the `tokenAmount`.

The docs were already outdated regarding some other data in the resulting structure of these requests, i.e. `accountType` vs the current normal response which is only `type`.

#### Summary of Changes

Added a method that was missing in the rpc docs index
Changed the `getTokenAccountsBy...` methods resulting data to be up-to-date and have more context


CriesofCarrots (approved):
Thank you!

Issue #18929, opened 2021-07-27
accountSubscribe with encoding:base58 may lead to validator lags


Fedor-sa:
#### Problem
_Steps_
1. Take account that has large data field e.g. `58KcficuUqPDcMittSddhT8LzsPJoH46YP4uURoMo5EB`
2. Send 
`{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "accountSubscribe",
  "params": [
    "58KcficuUqPDcMittSddhT8LzsPJoH46YP4uURoMo5EB",
    {
      "encoding": "base58",
      "commitment": "finalized"
    }
  ]
}`
3. Wait some time (you may optionally monitor using base64/base64+zstd subscriptions, until several notifications )

_Observed_
Validator hangs and does not accept new WS connections

#### Explanation
Account `58KcficuUqPDcMittSddhT8LzsPJoH46YP4uURoMo5EB` has over 1Mb data. 
`Base58` encoder has quadratic complexity, therefore it stuck with such input for some minutes which leads to denial of service of whole notification process
https://github.com/mycorrhiza/bs58-rs/blob/pr%C4%ABmum/src/encode.rs#L335

#### Proposed Solution
As it does not look feasible to verify data field size when validator receives subscription request, we propose 2 possible solutions
1. Check account data field size at notification moment and return some flag, reflecting error
2. Do not allow to use `base58` in subscriptions or forcing it to base64 (as done with jsonParsed encoding for some accounts)

Issue #18670, opened 2021-07-14
websocket connection becomes idle after subscribing to multiple accounts


davidbrota:
#### Problem

Hi. I'm subscribing to multiple accounts using the "accountSubscribe" RPC method. Those addresses belong to Serum eventqueues, whose data field takes 1 MB. I'm also using a heartbeat to prevent idle timeouts in the connection. 

Problem is that after successfully subscribing to 50 different addresses and receiving a result response for each one of them, the server, at some point, stops sending accountNotification responses and replying to the pings. The connection just stays open and idle (the server never closes it).

This scenario happens on both endpoints of the [Mainnet Beta](https://docs.solana.com/cluster/rpc-endpoints#mainnet-beta). We're also making sure that we aren't hitting the rate limits described there.

When subscribing to Serum asks and bids addresses the connection works perfectly and we don't experience this issue at all.

Any ideas on what might be going on there?

polachok:
I think we might have the same problem with `programSubscribe`.
I made a python script to reproduce (sorry for code quality, reduced from a larger program):

```python
from websocket import create_connection
import json
import hashlib

programs = ['9xQeWvG816bUx9EPjHmaT23yvVM2ZWbrrpZb9PusVFin']
accounts = []

def subscribe(sock, id, key, is_program):
    msg = None
    if not is_program:
        msg = {
                "jsonrpc": "2.0",
                "id": id,
                "method": "accountSubscribe",
                "params": [ key, { "commitment": "recent", "encoding": "base64" } ],
        }
    else:
        msg = {
                "jsonrpc": "2.0",
                "id": id,
                "method": "programSubscribe",
                "params": [ key, { "commitment": "recent", "encoding": "base64" } ],
        }
    sock.send(json.dumps(msg))

#ws = create_connection("ws://$MYINTERNALVALIDATOR/")
ws = create_connection("ws://api.mainnet-beta.solana.com/")
for key in accounts:
    id = key
    subscribe(ws, id, key, False)
for key in programs:
    id = key
    subscribe(ws, id, key, True)

print("Sent")
print("Receiving...")

total_length = 0
while True:
    msg =  ws.recv()
    length = len(msg)
    total_length += length
    obj = json.loads(msg)
    # response
    if 'result' in obj:
        print(obj)

    # notification
    if 'method' in obj:
        print("received %s" % obj["method"])
    print("total bytes received: %s (%sMb)" % (total_length, total_length / 1024 / 1024))
```

[...]

PR #18943, opened 2021-07-28
Optimize RPC pubsub for multiple clients with the same subscription

Riateche:
#### Problem

Current implementation generates notifications separately for each client, even if they use identical subscriptions. Each client also has a separate outgoing data buffer. This results in suboptimal performance and high memory usage. It also significantly limits the number of connections that can be served at the same time.

#### Summary of Changes

Subscriptions are delivered to clients using a broadcast queue. Each notification for a subscription is only generated and serialized once for all clients.

Implementation notes:

- soketto is used as the WebSocket implementation, but `jsonrpc_core` is still used for parsing requests and generating responses.
- `SubscriptionsTracker` serves the same purpose as old `solana_rpc::rpc_subscriptions::Subscriptions` type, but is modified to make deduplication of subscriptions possible. It provides indexes for efficient selection of relevant subscriptions (by params, by signature, etc.) In addition, it ensures automatic unsubscription when the last subscription token is dropped.
- `notify_accounts_logs_programs_signatures` now does a single iteration over the hashmap. Old implementation collected clones of all the keys into a vector and then did a separate lookup for each key.
- Some simplification and deduplication was made by unifying all subscription parameters into a single type (`SubscriptionParams`) and implementing single subscribe/unsubscribe functions instead of separate functions for each subscription type. That will make adding new subscription types much easier.
- Default values of optional parameters are now exclusively handled in `RpcSolPubSubImpl`. The rest of the code deals with concrete values of all the parameters.

#### Compatibility notes

- Subscription ID is now global for all clients. New subscription ID is generated using a global counter instead of a per-client counter. If a client makes the same subscription twice within a connection, the same subscription ID will be returned.
- Key order in notification values now follows field declaration order. Previous implementation produced sorted keys because of an intermediate conversion to `serde_json::Value`.


CriesofCarrots (commented):
Thanks! Generally, I'm liking the look of this approach, especially the cleanup in rpc_subscriptions. Have you done any benchmarking or other testing that could give us a concrete idea of the perf improvements?
If you can rebase on master, we can also give it a spin on a public cluster.

I've made a few cursory comments (including some little nits that happened to catch my eye), but will need to spend more time on a comprehensive review, if the perf looks great.
@jstarry can you please give this a look as well?

giftal:
hey, we've done the rebase and checked all unit-tests.

rpc/src/rpc_subscriptions.rs
```diff
@@ -113,175 +108,123 @@ impl std::fmt::Debug for NotificationEntry {
...
-        {
-            let slot = if commitment.is_finalized() {
-                commitment_slots.highest_confirmed_root
-            } else if commitment.is_confirmed() {
-                commitment_slots.highest_confirmed_slot
-            } else {
-                commitment_slots.slot
-            };
+    let commitment = if let Some(c) = subscription.commitment() {
+        c
```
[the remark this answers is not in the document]
00nktk: Fixed

rpc/src/rpc_pubsub_service.rs
```diff
@@ -110,6 +80,189 @@ impl PubSubService {
     }
 }
 
+type StdResult<T> = std::result::Result<T, Box<dyn std::error::Error>>;
```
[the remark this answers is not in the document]
00nktk: Just removed the alias, it seemed redundant.

core/tests/rpc.rs
```diff
@@ -32,6 +42,168 @@ use std::{
 };
 use tokio::runtime::Runtime;
 
+// Suppress needless_return due to
+//   https://github.com/paritytech/jsonrpc/blob/2d38e6424d8461cdf72e78425ce67d51af9c6586/derive/src/lib.rs#L204
+// Once https://github.com/paritytech/jsonrpc/issues/418 is resolved, try to remove this clippy allow
+#[allow(clippy::needless_return)]
+#[rpc]
+pub trait RpcSolPubSub {
```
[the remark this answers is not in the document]
00nktk: I didn't manage to come up with an easy way to do this without code duplication. We could introduce a macro that will expand into two trait definitions, but that will significantly increase the code compexity/readability and I doubt that helps the case.

On the other hand, `rpc_pubsub.rs` is a relatively small module and two traits (one of which is only for internal use) with a short and comprehensive comment will suffice. 

Personally, I am a fan of the macro approach, but I'm unsure about adding a relatively complex macros to such a big codebase.

validator/src/main.rs
```diff
@@ -1038,15 +1038,16 @@ pub fn main() {
...
-    let default_rpc_pubsub_max_in_buffer_capacity =
-        PubSubConfig::default().max_in_buffer_capacity.to_string();
-    let default_rpc_pubsub_max_out_buffer_capacity =
-        PubSubConfig::default().max_out_buffer_capacity.to_string();
+
+    // There params are no longer in use.
+    let default_rpc_pubsub_max_connections = 1000.to_string();
+    let default_rpc_pubsub_max_fragment_size = (50 * 1024).to_string();
+    let default_rpc_pubsub_max_in_buffer_capacity = (50 * 1024).to_string();
+    let default_rpc_pubsub_max_out_buffer_capacity = (15 * 1024 * 1024).to_string();
```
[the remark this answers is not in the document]
00nktk: Removed the variables, but did not the corresponding args since that can break many validator startup scripts.
We might want to print a deprecation warning/message if any of the values are set, but I did not find any elegant way of doing so via `clap` crate. Still can add the warnings if neccessary.

00nktk:
> Thanks! Generally, I'm liking the look of this approach, especially the cleanup in rpc_subscriptions. Have you done any benchmarking or other testing that could give us a concrete idea of the perf improvements?
> If you can rebase on master, we can also give it a spin on a public cluster.

Thanks for the review!
We did test the correctness in terms of consistency with the documented API, as well as somewhat subjective perfomance testing.
Regarding bugs/non-desirable behaviour that got fixed:
- Disconnection problem as described in #18670.
- PubSub lagging behind RPC within the same node.

We're currently running a number of tests, as well as mainnet nodes with this patch. Folks from our testing team will provide a better overview of the progress. @Fedor-sa please leave a comment.

Fedor-sa:
As for the testing part, we executed several tests comprising:
- Multiple connections (bring up 1000+ unique socket connections, subscribe in each connection separately for the same account, once message received compare messages to each other ensuring all messages are identical)
- WS/RPC consistency (subscribe to WS, receive notification, immediately call RPC method for the same account, compare received date taking into account slot). This test allows us to verify, that WS notifications are not very late. And even if slot in WS notification is behind slot, received upon RPC call, ensure that WS gets up to speed and reduces this gap
- Generic regression testing - other subscription methods, various parameters etc. In fact this issue was found during regression testing https://github.com/solana-labs/solana/issues/18929
Tests are currently in progress
- 50k+ simultaneous subscription (goal is to check validator load)
- 'heavyweight' subscription concurrent with dozen others. Ensure that 'heavyweight' subscription (e.g. programAccount with numerous accounts) does not impact other subscription performance 
You welcome if you have any questions
Fedor

CriesofCarrots:
> _Tests are currently in progress_
> 
>     * 50k+ simultaneous subscription (goal is to check validator load)
> 
>     * 'heavyweight' subscription concurrent with dozen others. Ensure that 'heavyweight' subscription (e.g. programAccount with numerous accounts) does not impact other subscription performance

This is great! I'm looking forward to seeing the data.
How did the WS/RPC consistency test compare to the current implementation?

core/tests/rpc.rs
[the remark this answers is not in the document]
CriesofCarrots: I suppose another option would be a procedural macro that compares `RpcSolPubSub` pubsub trait with the new`RpcSolPubSub` rpc trait and ensures they include the same methods.

> On the other hand, `rpc_pubsub.rs` is a relatively small module and two traits (one of which can probably be `pub(crate)`) with a short and comprehensive comment will suffice.

But fair point, it is a small module, and probably not worth the overhead of writing a complex macro. Also, I suppose it's not the end of the world if the client lags behind. Okay, two traits and comprehensive comments sounds good.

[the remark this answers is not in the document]
00nktk: Done. However I'm not sure if that's the best way to express the issue, so it would be great if you could provide any corrections.

00nktk:
> If you can rebase on master, we can also give it a spin on a public cluster.

Hey @CriesofCarrots, I rebased the branch, we think it would be great to give it a try. Anything else you need to make this happen?

CriesofCarrots:
> > If you can rebase on master, we can also give it a spin on a public cluster.
> 
> Hey @CriesofCarrots, I rebased the branch, we think it would be great to give it a try. Anything else you need to make this happen?

Perfect, thank you! I've got a node set up, so I'll give it a test over the next couple days.

Fedor-sa:
From testing point of view
-  1000+ connections :white_check_mark:
- 40000+ simultaneous subscriptions `accountSubscribe` in 10 connections :white_check_mark: (also combined with additional "heavyweight" program subscription)
- Data consistency
    1.  WS data vs RPC data on patched validator - :white_check_mark: - this means that there are only small (1-5 Slots) lags observed between WS notification and RPC responses for single pubkey and WS reduces this gap while account data is not being changed rapidly. This test was done with `encoding: base64+zstd`  since data size for `base64` is much larger 
    2. Data  comparison between Mainnet and patched validator - *in progress* 
We found few cases of improvement  (see below), but still trying to check that all data is fully identical for the same slot and `commitment: finalized`


_Mainnet_ 
```
WS SLOT:  90230152 | First element in section "data": 6ad5ebc0e0ed70d27eb5c5c352b5937a |  10:26:15.443227 | api.mainnet-beta.solana.com
RPC SLOT:  90230152 | First element in section "data": 6ad5ebc0e0ed70d27eb5c5c352b5937a |  10:26:16.040630 | api.mainnet-beta.solana.com

WS SLOT:  90230155 | First element in section "data": c37e186ee30bba237c866e1aa7c666f1 |  10:26:17.069676 | api.mainnet-beta.solana.com
RPC SLOT:  90230156 | First element in section "data": 6ad5ebc0e0ed70d27eb5c5c352b5937a |  10:26:18.070193 | api.mainnet-beta.solana.com
```

_Updated validator_
```
WS SLOT:  90230139 | First element in section "data": 6ad5ebc0e0ed70d27eb5c5c352b5937a |  10:26:01.795825 | pub-validator-01.zubr.one:8900
RPC SLOT:  90230139 | First element in section "data": 6ad5ebc0e0ed70d27eb5c5c352b5937a |  10:26:02.004863 | pub-validator-01.zubr.one:8899

WS SLOT:  90230152 | First element in section "data": c37e186ee30bba237c866e1aa7c666f1 |  10:26:15.429553 | pub-validator-01.zubr.one:8900
RPC SLOT:  90230152 | First element in section "data": c37e186ee30bba237c866e1aa7c666f1 |  10:26:15.638836 | pub-validator-01.zubr.one:8899

WS SLOT:  90230156 | First element in section "data": 4d35d0f10d32536237e41886f0899a16 |  10:26:17.732359 | pub-validator-01.zubr.one:8900
RPC SLOT:  90230156 | First element in section "data": 4d35d0f10d32536237e41886f0899a16 |  10:26:17.937133 | pub-validator-01.zubr.one:8899
```

Comments/questions are welcome

rpc/src/rpc_subscriptions.rs
```diff
@@ -1139,179 +654,133 @@ impl RpcSubscriptions {
...
+                ("num_logs_subscriptions", num_logs_found, i64),
                 ("num_logs_notified", num_logs_notified, i64),
-                ("logs_time", logs_time.as_us() as i64, i64),
-                ("num_program_subscriptions", programs.len(), i64),
+                ("num_program_subscriptions", num_programs_found, i64),
                 ("num_programs_notified", num_programs_notified, i64),
-                ("programs_time", programs_time.as_us() as i64, i64),
-                ("num_signature_subscriptions", signatures.len(), i64),
+                ("num_signature_subscriptions", num_signatures_found, i64),
                 ("num_signatures_notified", num_signatures_notified, i64),
```
CriesofCarrots, line 779:
```suggestion
                ("num_signatures_notified", num_signatures_notified, i64),
                ("notifications_time", total_time.as_us() as i64, i64),
```

CriesofCarrots (commented):
@00nktk @Fedor-sa , thanks for your patience! This is testing well on our end.
Could you please add the metric below? I'd love to know how it compares to the accounts_time metric in your 40000+ simultaneous subscriptions test.

Also, in addition to the num_account_subscriptions/num_logs_subscriptions/num_signature_subscriptions that now report the number of unique `SubscriptionParams` of each type, would it be possible to emit a metric for the total number of broadcast subscribers of each type?

I think we're close :)

rpc/src/rpc_subscriptions.rs
00nktk: Hey, I've added `num_*_subscribers` metrics that represent the number of references to tokens of a certain subscription kind. That means though if a client is subscribed to `n` different subscriptions of a same kind, they will be count as `n` separate subscribers.

If you meant counting the amount of clients that *participate* in a certain subscription kind, then it's not that easy and would likely require a separate struct (maybe even two) with some atomic counters, which might seems like an overkill for a metric. But still doable.

`notifications_time` is added as well.

CriesofCarrots: Thank you!
Yes, I did mean number of clients that would receive notifications of a certain subscription kind, so that it would be parallel to the old `num_accounts_notified`/`num_signatures_notified`/`num_programs_notified`. It would be a lot easier to compare apples to apples if we had that. I'll ponder whether it's worth adding atomics for that, though.

In the meantime, any word on how `notifications_time` compares to the old `accounts_time` metric in your 40000+ simultaneous subscriptions test?

Riateche:
We've done some further improvements to the implementation:
- The pubsub server thread is now more decoupled from the notification generation thread. Instead of sharing the subscription tracker between the threads and accessing it via a mutex, the subscription tracker is now exclusively managed by the notification thread. The server thread pushes deduplicated subscriptions and unsubscriptions to the notification queue. This eliminates contention between the threads.
- There is now a limit on the total memory usage for items in the broadcast queue. 

We've added more metrics and are doing more extensive performance testing. We'll share the results once they're ready.

CriesofCarrots:
Oof, we're deep into testing the previous changeset with real cluster traffic/load; I did not expect more changes of this magnitude. I'll hold off on any more until your performance testing is done. Please let me know if you expect more sizable changes.

giftal:
> Oof, we're deep into testing the previous changeset with real cluster traffic/load; I did not expect more changes of this magnitude.
> I'll hold off on any more until your performance testing is done. Looking forward to seeing the results!
> Please let me know if you expect more sizable changes.



Hey Tyera,

Please go on with your tests, no more changes expected here.

CriesofCarrots (commented):
Thanks for the rebase. Can you run `cargo fmt` to pass first CI checks?
One review comment; would love to make "plain" more self-explanatory.

rpc/src/rpc_subscription_tracker.rs
```diff
@@ -359,18 +369,18 @@ impl LogsSubscriptionsIndex {
...
 pub struct SubscriptionsTracker {
     logs_subscriptions_index: LogsSubscriptionsIndex,
-    by_params: HashMap<SubscriptionParams, Arc<SubscriptionInfo>>,
     by_signature: HashMap<Signature, HashMap<SubscriptionId, Arc<SubscriptionInfo>>>,
     // Accounts, logs, programs, signatures (not gossip)
     commitment_watchers: HashMap<SubscriptionId, Arc<SubscriptionInfo>>,
     // Accounts, logs, programs, signatures (gossip)
     gossip_watchers: HashMap<SubscriptionId, Arc<SubscriptionInfo>>,
+    // Slots, slots updates, roots, votes.
+    plain: HashMap<SubscriptionParams, Arc<SubscriptionInfo>>,
```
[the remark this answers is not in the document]
Riateche: Sounds good.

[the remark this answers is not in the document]
CriesofCarrots: Please see CI failure

Riateche:
Not sure what's going wrong on the CI. `cargo test -p solana-rpc` succeeds and takes about a minute total on my system.

CriesofCarrots:
> Not sure what's going wrong on the CI. `cargo test -p solana-rpc` succeeds and takes about a minute total on my system.

Hmm, okay, I'll try to troubleshoot this tomorrow. What is your rust version?

Fedor-sa:
Regarding testing: we compared performance between old (1.6.20) and new versions 
_Methodology:_
1. Bring up within 1 minute 1000 connection and send `accountSubscribe` for 1 key in each connection. We used only 4 keys, but we regularly observe notification for them, therefore processing all notifications is a load for ws on validator
2. Keep all connections and subscriptions active for 15 minutes and then disconnect them within 1 minute
_Results_
### Old validator
```
data_received........................: 35 GB   33 MB/s
received_accountNotification_count...: 256090  240.459488/s
ws_msgs_received.....................: 257092  241.400331/s
ws_sessions..........................: 1000    0.938965/s
```
### New validator
```
data_received........................: 153 GB  143 MB/s
received_accountNotification_count...: 801912  752.966887/s
ws_msgs_received.....................: 802919  753.912424/s
ws_sessions..........................: 1000    0.938964/s
```
As we can see from results, validator with new WS sends around 5x times more data and around 3x times more notifications messages in the same test scenario

Riateche:
> Hmm, okay, I'll try to troubleshoot this tomorrow. What is your rust version?

1.54.0.

metrics/src/lib.rs
```diff
@@ -3,3 +3,55 @@ pub mod counter;
...
+        // new_count = strong_count
+        //    - 1 (in TokenCounter)
+        //    - 1 (token that's being dropped)
+        datapoint_info!(*self.0, ("count", Arc::strong_count(&self.0) - 2, i64));
+    }
+}
+
+impl Drop for TokenCounter {
+    fn drop(&mut self) {
+        datapoint_info!(*self.0, ("count", Arc::strong_count(&self.0) - 2, i64));
```
CriesofCarrots, line 55: This caused a panic in CI: https://buildkite.com/solana-labs/solana/builds/53940#46908081-e10f-441a-a548-d2b1ecec726b
Is this math right?
Also, please use checked or saturating math throughout this module to be safe. A node shouldn't panic submitting metrics.

Riateche: Thanks for pinpointing the issue. I've made a fix. The RPC tests now pass, but there is a timeout in CI.

CriesofCarrots: Thanks. Yes, we still need to troubleshoot that timeout issue.
On the plus side, we have the code on a v1.6 backport receiving real mainnet-beta traffic. Looking great so far.

CriesofCarrots:
@Riateche , an update from my end. The metrics on the staging server running this code look really good.
However, we need to figure out the unit test timeout issue. On my local MacOS machine and gce linux dev machine, the rpc_pubsub and rpc_subscription unit tests are taking 2-3s, vs <0.5s on current master. That's a pretty big regression. It's even worse inside a docker container (which is how CI runs), with some tests hanging for 50s for reasons I haven't been able to identify.
Do you have thoughts as to the source of the slow-down?

CriesofCarrots:
Thanks for the updates here! The solana-rpc tests look good now. Current timeout is because of tests in solana-core. I'll take a look and try to figure out the right direction.

rpc/src/rpc_pubsub_service.rs
```diff
@@ -39,10 +40,17 @@ pub struct PubSubConfig {
 
 impl Default for PubSubConfig {
     fn default() -> Self {
+        let queue_capacity_items = if cfg!(test) {
```
CriesofCarrots, line 43: Oh. This is the issue. `cfg!(test)` doesn't propagate across crates, and `RpcSubscriptions::new()` is used in quite a few solana-core module unit tests.
Although it will be a lot of churn, `RpcSubscriptions::new_for_tests()` might be the best way to go. I'm happy to help with this, if you give me permission to push to your branch.

CriesofCarrots: https://github.com/CriesofCarrots/solana/tree/turbo-ws1

Riateche: Thank you for the help! Are we now good to merge?

CriesofCarrots: Yep, let's pull the trigger

PR #18947, opened 2021-07-28
shares cluster-nodes between retransmit threads

behzadnouri:
#### Problem
cluster_nodes and last_peer_update are not shared between retransmit
threads, as each thread have its own value:
https://github.com/solana-labs/solana/blob/65ccfed86/core/src/retransmit_stage.rs#L476-L477

Additionally, with shared references, this code:
https://github.com/solana-labs/solana/blob/0167daa11/core/src/retransmit_stage.rs#L315-L328
has a concurrency bug where the thread which does compare_and_swap,
updates cluster_nodes much later after other threads have run with
outdated cluster_nodes for a while. In particular, the write-lock there
may block.

#### Summary of Changes
This commit shares cluster-nodes between retransmit threads.

### DIFF
--- a/core/benches/retransmit_stage.rs
+++ b/core/benches/retransmit_stage.rs
@@ -3,34 +3,42 @@
 extern crate solana_core;
 extern crate test;
 
-use log::*;
-use solana_core::retransmit_stage::retransmitter;
-use solana_entry::entry::Entry;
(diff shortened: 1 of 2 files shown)


sakridge (approved):
lgtm

PR #18971, opened 2021-07-29
unifies cluster-nodes computation & caching across turbine stages

behzadnouri:
#### Problem
1. Broadcast-stage is using `epoch_staked_nodes` based on the same slot that shreds belong to:
https://github.com/solana-labs/solana/blob/049fb0417/core/src/broadcast_stage/standard_broadcast_run.rs#L208-L228
https://github.com/solana-labs/solana/blob/0cf52e206/core/src/broadcast_stage.rs#L342-L349

    But retransmit-stage is using bank-epoch of the working-bank:
https://github.com/solana-labs/solana/blob/19bd30262/core/src/retransmit_stage.rs#L272-L289

    So the two are not consistent at epoch boundaries where some nodes may have a working bank (or similarly a root bank) lagging other nodes.

2. Both retransmit and broadcast stages cache _peers sorted by stake_, but they do not update the cache when the epoch (and so epoch staked nodes, and therefore order of peers) changes; The order of peers is crucial in order to consistently shuffle nodes over the propagation tree.
https://github.com/solana-labs/solana/blob/19bd30262/core/src/broadcast_stage/standard_broadcast_run.rs#L332-L344
https://github.com/solana-labs/solana/blob/19bd30262/core/src/retransmit_stage.rs#L272-L309

3. Cache update has a concurrency bug in which the thread which does `compare_and_swap` may be blocked when it tries to obtain the write-lock on cache, while other threads will keep running ahead with the outdated cache (since the atomic timestamp is already updated):
https://github.com/solana-labs/solana/blob/19bd30262/core/src/broadcast_stage/standard_broadcast_run.rs#L332-L344 

#### Summary of Changes
* This commit adds a new `ClusterNodesCache` which caches cluster-nodes for each epoch.
* The cache entries are keyed by epoch, and so if epoch changes cluster-nodes will be recalculated.
* A time-to-live eviction policy is enforced to refresh entries in case gossip contact-infos are updated.
    * Follow up changes will reduce turbine peer computations dependency on gossip propagation of contact-infos.
* Broadcast and retransmit stages will compute stakes (and so cluster nodes) consistently as both stages call into the same implementation.

### DIFF
--- a/core/src/broadcast_stage.rs
+++ b/core/src/broadcast_stage.rs
@@ -124,7 +124,7 @@ impl BroadcastStageType {
     }
 }
 
-pub type TransmitShreds = (Option<Arc<HashMap<Pubkey, u64>>>, Arc<Vec<Shred>>);
+type TransmitShreds = (Slot, Arc<Vec<Shred>>);
 trait BroadcastRun {
     fn run(
         &mut self,
@@ -339,27 +339,25 @@ impl BroadcastStage {
         }
 
         for (_, bank) in retransmit_slots.iter() {
-            let bank_epoch = bank.get_leader_schedule_epoch(bank.slot());
-            let stakes = bank.epoch_staked_nodes(bank_epoch);
-            let stakes = stakes.map(Arc::new);
+            let slot = bank.slot();
             let data_shreds = Arc::new(
                 blockstore
-                    .get_data_shreds_for_slot(bank.slot(), 0)
+                    .get_data_shreds_for_slot(slot, 0)
                     .expect("My own shreds must be reconstructable"),
             );
 
             if !data_shreds.is_empty() {
-                socket_sender.send(((stakes.clone(), data_shreds), None))?;
+                socket_sender.send(((slot, data_shreds), None))?;
             }
 
             let coding_shreds = Arc::new(
                 blockstore
-                    .get_coding_shreds_for_slot(bank.slot(), 0)
+                    .get_coding_shreds_for_slot(slot, 0)
                     .expect("My own shreds must be reconstructable"),
             );
 
             if !coding_shreds.is_empty() {
-                socket_sender.send(((stakes.clone(), coding_shreds), None))?;
+                socket_sender.send(((slot, coding_shreds), None))?;
             }
         }
 
@@ -464,10 +462,9 @@ pub mod test {
     };
 
     #[allow(clippy::implicit_hasher)]
-    pub fn make_transmit_shreds(
+    fn make_transmit_shreds(
         slot: Slot,
         num: u64,
-        stakes: Option<Arc<HashMap<Pubkey, u64>>>,
     ) -> (
         Vec<Shred>,
         Vec<Shred>,
@@ -489,11 +486,11 @@ pub mod test {
             coding_shreds.clone(),
             data_shreds
                 .into_iter()
-                .map(|s| (stakes.clone(), Arc::new(vec![s])))
+                .map(|s| (slot, Arc::new(vec![s])))
                 .collect(),
             coding_shreds
                 .into_iter()
-                .map(|s| (stakes.clone(), Arc::new(vec![s])))
+                .map(|s| (slot, Arc::new(vec![s])))
                 .collect(),
         )
     }
@@ -537,7 +534,7 @@ pub mod test {
         // Make some shreds
         let updated_slot = 0;
         let (all_data_shreds, all_coding_shreds, _, _all_coding_transmit_shreds) =
-            make_transmit_shreds(updated_slot, 10, None);
+            make_transmit_shreds(updated_slot, 10);
         let num_data_shreds = all_data_shreds.len();
         let num_coding_shreds = all_coding_shreds.len();
         assert!(num_data_shreds >= 10);
--- a/core/src/broadcast_stage.rs
+++ b/core/src/broadcast_stage.rs
@@ -44,6 +44,9 @@ pub(crate) mod broadcast_utils;
 mod fail_entry_verification_broadcast_run;
 mod standard_broadcast_run;
 
+const CLUSTER_NODES_CACHE_NUM_EPOCH_CAP: usize = 8;
+const CLUSTER_NODES_CACHE_TTL: Duration = Duration::from_secs(5);
+
 pub(crate) const NUM_INSERT_THREADS: usize = 2;
 pub(crate) type RetransmitSlotsSender = CrossbeamSender<HashMap<Slot, Arc<Bank>>>;
 pub(crate) type RetransmitSlotsReceiver = CrossbeamReceiver<HashMap<Slot, Arc<Bank>>>;
--- a/core/src/broadcast_stage/broadcast_duplicates_run.rs
+++ b/core/src/broadcast_stage/broadcast_duplicates_run.rs
@@ -49,70 +49,6 @@ impl BroadcastDuplicatesRun {
             num_slots_broadcasted: 0,
         }
     }
-
-    fn get_non_partitioned_batches(
-        &self,
-        my_pubkey: &Pubkey,
-        bank: &Bank,
-        data_shreds: Arc<Vec<Shred>>,
-    ) -> TransmitShreds {
-        let bank_epoch = bank.get_leader_schedule_epoch(bank.slot());
-        let mut stakes: HashMap<Pubkey, u64> = bank.epoch_staked_nodes(bank_epoch).unwrap();
-        stakes.retain(|pubkey, _stake| pubkey != my_pubkey);
-        (Some(Arc::new(stakes)), data_shreds)
-    }
-
-    fn get_partitioned_batches(
-        &self,
-        my_pubkey: &Pubkey,
-        bank: &Bank,
-        original_shreds: Arc<Vec<Shred>>,
-        partition_shreds: Arc<Vec<Shred>>,
-    ) -> (TransmitShreds, TransmitShreds) {
-        // On the last shred, partition network with duplicate and real shreds
-        let bank_epoch = bank.get_leader_schedule_epoch(bank.slot());
-        let mut original_recipients = HashMap::new();
-        let mut partition_recipients = HashMap::new();
-
-        let mut stakes: Vec<(Pubkey, u64)> = bank
-            .epoch_staked_nodes(bank_epoch)
-            .unwrap()
-            .into_iter()
-            .filter(|(pubkey, _)| pubkey != my_pubkey)
-            .collect();
-        stakes.sort_by(|(l_key, l_stake), (r_key, r_stake)| {
-            if r_stake == l_stake {
-                l_key.cmp(r_key)
-            } else {
-                l_stake.cmp(r_stake)
-            }
-        });
-
-        let mut cumulative_stake: u64 = 0;
-        for (pubkey, stake) in stakes.into_iter() {
-            cumulative_stake += stake;
-            if cumulative_stake <= self.config.stake_partition {
-                partition_recipients.insert(pubkey, stake);
-            } else {
-                original_recipients.insert(pubkey, stake);
-            }
-        }
-
-        warn!(
-            "{} sent duplicate slot {} to nodes: {:?}",
-            my_pubkey,
-            bank.slot(),
-            &partition_recipients,
-        );
-
-        let original_recipients = Arc::new(original_recipients);
-        let original_transmit_shreds = (Some(original_recipients), original_shreds);
-
-        let partition_recipients = Arc::new(partition_recipients);
-        let partition_transmit_shreds = (Some(partition_recipients), partition_shreds);
-
-        (original_transmit_shreds, partition_transmit_shreds)
-    }
 }
 
 impl BroadcastRun for BroadcastDuplicatesRun {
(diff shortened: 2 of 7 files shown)


core/src/broadcast_stage/broadcast_fake_shreds_run.rs
```diff
@@ -107,15 +101,16 @@ impl BroadcastRun for BroadcastFakeShredsRun {
...
-                if i <= self.partition && stakes.is_some() {
+                if i <= self.partition {
                     // Send fake shreds to the first N peers
                     data_shreds.iter().for_each(|b| {
                         sock.send_to(&b.payload, &peer.tvu_forwards).unwrap();
                     });
-                } else if i > self.partition && stakes.is_none() {
+                } else if i > self.partition {
                     data_shreds.iter().for_each(|b| {
                         sock.send_to(&b.payload, &peer.tvu_forwards).unwrap();
```
jbiseda, line 115: duplicate code?

core/src/broadcast_stage/broadcast_fake_shreds_run.rs
```diff
@@ -107,15 +101,16 @@ impl BroadcastRun for BroadcastFakeShredsRun {
         sock: &UdpSocket,
         _bank_forks: &Arc<RwLock<BankForks>>,
     ) -> Result<()> {
-        for ((stakes, data_shreds), _) in receiver.lock().unwrap().iter() {
+        for ((_slot, data_shreds), _) in receiver.lock().unwrap().iter() {
             let peers = cluster_info.tvu_peers();
+            // TODO: stakes.is_some was used here to identify fakes!
```
jbiseda, line 106: What should change to address the TODO?

core/src/retransmit_stage.rs
```diff
@@ -58,6 +58,9 @@ const DEFAULT_LRU_SIZE: usize = 10_000;
 // it doesn't pull up too much work.
 const MAX_PACKET_BATCH_SIZE: usize = 100;
 
+const CLUSTER_NODES_CACHE_CAP: usize = 8; // Number of epochs.
+const CLUSTER_NODES_CACHE_TTL: Duration = Duration::from_secs(5);
```
jbiseda, line 62: All callers are using the same values, move these consts into ClusterNodesCache::new()?

core/src/cluster_nodes.rs
```diff
@@ -211,6 +229,67 @@ fn get_nodes(cluster_info: &ClusterInfo, stakes: &HashMap<Pubkey, u64>) -> Vec<N
...
+            }
+        }
+        let epoch_staked_nodes = root_bank.epoch_staked_nodes(epoch);
+        if epoch_staked_nodes.is_none() {
+            inc_new_counter_info!("cluster_nodes-unknown_epoch_staked_nodes", 1);
+            if epoch != root_bank.get_leader_schedule_epoch(root_bank.slot()) {
+                return self.get(root_bank.slot(), root_bank, cluster_info);
+            }
+            inc_new_counter_info!("cluster_nodes-unknown_epoch_staked_nodes_root", 1);
+        }
```
carllin, line 283: If this function were to return an option, I think it's simpler to return `None` here and drop retransmit of a packet if the epoch stakes are so far ahead that we don't have an entry in the root bank for the epoch yet. 

Also, we already filter out these types of shreds that are too far ahead of the leader schedule in retransmit since we cannot determine the leader for signature verification, and it's good avoid retransmitting these types of shreds in case it's some sort of DOS vector anyways

core/src/cluster_nodes.rs
```diff
@@ -211,6 +229,67 @@ fn get_nodes(cluster_info: &ClusterInfo, stakes: &HashMap<Pubkey, u64>) -> Vec<N
...
+        root_bank: &Bank,
+        cluster_info: &ClusterInfo,
+    ) -> Arc<ClusterNodes<T>> {
+        let epoch = root_bank.get_leader_schedule_epoch(shred_slot);
+        let entry = self.get_cache_entry(epoch);
+        // Hold the lock on the entry here so that, if needed, only
+        // one thread recomputes cluster-nodes for this epoch.
+        let mut entry = entry.lock().unwrap();
+        if let Some((asof, nodes)) = entry.deref() {
+            if asof.elapsed() < self.ttl {
```
carllin, line 272: If we enforce: https://github.com/solana-labs/solana/pull/18971/files#r681451853 the `ttl` doesn't seem necessary since the set of epoch staked nodes for the known epochs <  `root_bank.get_leader_schedule_epoch(root_slot)` should always return the same result

core/src/broadcast_stage/broadcast_duplicates_run.rs
```diff
@@ -47,70 +55,19 @@ impl BroadcastDuplicatesRun {
...
-            &partition_recipients,
-        );
-
-        let original_recipients = Arc::new(original_recipients);
-        let original_transmit_shreds = (Some(original_recipients), original_shreds);
-
-        let partition_recipients = Arc::new(partition_recipients);
-        let partition_transmit_shreds = (Some(partition_recipients), partition_shreds);
-
+        // TODO: This is no longer doing what it is used to do!
```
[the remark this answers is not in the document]
carllin: Selfishly, it would be nice to maintain this old behavior if possible, since I have a few pending changes based on this 😃.

If you already have an idea of how to fixup this behavior, feel free to share. If not, I might have to take some creative freedoms to get this old behavior back 😈

core/src/retransmit_stage.rs
jbiseda: You could do something like ClusterNodesCache::new(Option<(cap, ttl)>) and use these as defaults if no values are provided. At least it would be explicit that the values can be chosen by the caller. That said, I'm not opposed to leaving it as written.

core/src/broadcast_stage.rs
```diff
@@ -44,6 +44,9 @@ pub(crate) mod broadcast_utils;
 mod fail_entry_verification_broadcast_run;
 mod standard_broadcast_run;
 
+const CLUSTER_NODES_CACHE_CAP: usize = 8; // Number of epochs.
```
jbiseda, line 47: s/CLUSTER_NODES_CACHE_CAP/CLUSTER_NODES_CACHE_NUM_EPOCH_CAP/ and remove the comment?

jbiseda (dismissed):
lgtm

core/src/broadcast_stage/broadcast_duplicates_run.rs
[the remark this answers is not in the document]
carllin: Awesome, thanks!

core/src/cluster_nodes.rs
carllin: > Can you please point me to where in the code-base above "2 epoch ahead" is implemented?
I find below in bank::new_from_parent:
https://github.com/solana-labs/solana/blob/cde146155/runtime/src/bank.rs#L1293-L1294
but this seems to be only one epoch ahead.

Ha you don't need me, you hit it right on the nose! Yeah it's 1 epoch, I'm misremembering information from the 🦖 times.

> Should we have fallback here that if the root-bank did not have the stakes for the epoch, then use the working-bank

I think we should fail here by returning `None`. If we are an epoch behind then we can't make an accurate decision so I think it's better to let the caller handle this

core/src/cluster_nodes.rs
carllin: Ah shucks, I misread this, I thought we were caching the `epoch_staked_nodes` here: https://github.com/solana-labs/solana/pull/18971/files#diff-ef69c0f93244cfe4b45f4d6ca5af4e91b59c123ef228cb0e73ac80a21c078eafR276 which never changes. 

Does it makes sense to cache the `epoch_staked_ndoes` as well so we can reuse it when reconstructing the cache entry based on new gossip contact-infos? i.e. changing the `CacheEntry<T>` to be `type CacheEntry<T> = Option<(/*as of:*/ Instant, /*epoch staked nodes*/ HashMap<Pubkey, u64>, Arc<ClusterNodes<T>>)>;`

core/src/cluster_nodes.rs
behzadnouri: > If we are an epoch behind then we can't make an accurate decision so I think it's better to let the caller handle this

But it is not actually that we are behind. In the example I gave above the node was actually catching up with the cluster. But at the epoch boundary, it does not have epoch-stakes for the needed epoch.

The current master code is using:
```rust
bank.get_leader_schedule_epoch(shred_slot)
```
which is **one epoch after**
```rust
epoch_schedule.get_epoch(shred_slot)
```
And at the epoch boundary shred is already one epoch after the root-slot. So in aggregate it will be 2 epochs ahead.

carllin: > And at the epoch boundary shred is already one epoch after the root-slot. So in aggregate it will be 2 epochs ahead.

Ah I see, that's gross, thanks for the explanation. 

My mental model was flawed because I didn't realize we were using the `leader_schedule_epoch` for the given slot to determine the `epoch_slots`: https://github.com/solana-labs/solana/blob/v1.7/core/src/broadcast_stage/standard_broadcast_run.rs#L225. This makes me wonder why we're not just using the epoch of the slot itself, i.e. `bank.epoch_schedule.get_epoch(slot)`

core/src/cluster_nodes.rs
behzadnouri: https://github.com/solana-labs/solana/pull/19090 will implement copy-on-write for staked-nodes and so wouldn't there be much room left here to optimize. Even without that commit I doubt that it will make much of difference. Leaning on keeping things simple for now.

core/src/broadcast_stage.rs
behzadnouri: done

core/src/cluster_nodes.rs
behzadnouri: > This makes me wonder why we're not just using the epoch of the slot itself, i.e. bank.epoch_schedule.get_epoch(slot); the epoch of the current slot should already be fixed in broadcast, and this should reduce the gap between the root bank and the slot to 1 epoch even on epoch boundaries.

Agree. I do not know if there is some specific reasons (or some corner cases) that requires to use `get_leader_schedule_epoch` instead. :thinking: 

> Rather than gambling with another layer of potential inconsistency, should we featurize a change to change the cluster to use bank.epoch_schedule.get_epoch(slot) instead of the bank.get_leader_schedule_epoch(slot) first, and then push this change?

I think if we just fallback on the working-bank here it will be backward compatible because that is already what the current master code is doing (i.e. it uses working-bank). I added this logic and I feel like we should be safe to merge this now (again due to backward compatibility). I did some further testing as well on test gce clusters.

Also, as you mentioned, earlier in the pipeline the shreds are sig-verified by the known leader of the slot. So it has already passed some sanity checks.

Once this is merged in, we can investigate if it makes sense to switch to `get_epoch` instead and remove working-bank entirely.

core/src/cluster_nodes.rs
carllin: Sounds good, thanks!

core/src/cluster_nodes.rs
carllin: > I think if we just fallback on the working-bank here it will be backward compatible because that is already what the current master code is doing

Yeah the only difference I was thinking of was when the working bank crossed the epoch boundary into the next epoch, then we got a shred `S` from the old epoch.

Then in the old code we would be using the stakes from the next epoch
In this new code we would be using the stakes based on the old epoch for shred `S`

Perhaps this is a small enough window where things are relatively fine?

behzadnouri: That is actually an inconsistency that this commit is intentionally trying to remove.
Please see "1." in the pull request description.

carllin: Right, but was more worried about the upgrade between people with and without this code change.

behzadnouri: We can monitor on testnet first and see how it goes. I am expecting it to be fine as the fanout is pretty large.

carllin (approved):

core/src/broadcast_stage/broadcast_duplicates_run.rs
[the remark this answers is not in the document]
carllin: @behzadnouri  I'm currently working to fix this up for `BroadcastDuplicatesRun`, will let you know if I run into any issues.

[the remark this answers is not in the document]
behzadnouri: Hey, this is the code to create partition on the last data shreds: https://github.com/behzadnouri/solana/commit/broadcast-duplicate

Feel free to use that code, or I can just send it as a pull-request to you. Either way is fine with me.

Please note that the old code:
https://github.com/solana-labs/solana/blob/b67ffab37/core/src/broadcast_stage/broadcast_duplicates_run.rs#L77-L99
was aiming to create a partition by manipulating the stakes effectively setting some of them to zero. But (even before all my changes to turbine) this does not actually create a partition. Because unstaked nodes were always part of the turbine broadcast and retransmit stages as long as their contact-info exists in gossip (ie. in cluster-info).

To actually create a partition you need to intercept where packets are sent over the socket as it is done here: https://github.com/behzadnouri/solana/commit/broadcast-duplicate

I am actively testing changes on turbine so if you plan to make changes here, please let me know in advance so that we do not step on each others' foot. cheers.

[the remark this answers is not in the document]
behzadnouri: sent out https://github.com/solana-labs/solana/pull/19253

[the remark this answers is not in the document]
carllin: @behzadnouri thanks!

PR #18983, opened 2021-07-29
Add snapshot_utils::bank_from_latest_snapshot_archives()

### LINKED ISSUE #18973: [Incremental Snapshots] Create fn to load from latest snapshots and then verify
After `bank_forks_utils` calls `snapshot_utils::bank_from_snapshot_archives()`, it verifies the bank matches the SnapshotArchiveInfo. This functionality should be moved to `snapshot_utils`.

Also, write a unit test for it, and update the tests in `core/tests/snapshots.rs` to use it.

_Original PR comment below:_

This flag to not check the hash was introduced here: https://github.com/solana-labs/solana/pull/17939/files#diff-b5dead0a38e5f63447595bbeda1428d3aa851d33954343bb1bc38b4b00add5fdR149, not sure why it's not safe to check the hash here. 

Maybe the hash is verified elsewhere already? Would be good to check the accounts state is being verified to match the snapshot hash somewhere in the test.

_Originally posted by @carllin in https://github.com/solana-labs/solana/pull/18565#discussion_r677816510_

brooksprumo:
While reviewing PR #18565, as issue was brought up to refactor some code
around verifying the bank after rebuilding from snapshots.  A new
top-level function has been added to get the latest snapshot archives
and load the bank then verify.  Additionally, new tests have been
written and existing tests have been updated to use this new function.

Fixes #18973

While resolving the issue, it became clear there was some additional
low-hanging fruit this change enabled.  Specifically, the functions
`bank_to_xxx_snapshot_archive()` now return their respective
`SnapshotArchiveInfo`.  And on the flip side,
`bank_from_snapshot_archives()` now takes `SnapshotArchiveInfo`s instead
of separate paths and archive formats.  This bundling simplifies bank
rebuilding.


lijunwangs (approved):
Looks good to me overall.

core/tests/snapshots.rs
```diff
@@ -169,22 +178,16 @@ mod tests {
...
-            ArchiveFormat::TarBzip2,
             old_genesis_config,
             None,
             None,
             AccountSecondaryIndexes::default(),
             false,
             None,
             accounts_db::AccountShrinkThreshold::default(),
-            check_hash_calculation,
+            false,
```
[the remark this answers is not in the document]
brooksprumo: The variable `check_hash_calculation` is defined above as the constant `false`, so I thought this would be more clear.

Here's the line in particular:
https://github.com/solana-labs/solana/blob/e4be00fece487d5413f7569b2e46c770b9b4d12e/core/tests/snapshots.rs#L163

Do you have a preference? I've found that a lot of the tests just use the constants instead of creating variables for each one.

runtime/src/snapshot_utils.rs
```diff
@@ -958,6 +974,87 @@ where
...
+        &snapshot_archives_dir,
+        *full_snapshot_archive_info.slot(),
+    );
+
+    info!(
+        "Loading bank from full snapshot: {}, and incremental snapshot: {:?}",
+        full_snapshot_archive_info.path().display(),
+        incremental_snapshot_archive_info
+            .as_ref()
+            .map(|isai| isai.path().display())
```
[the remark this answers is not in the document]
brooksprumo: I agree. I'll change this.

runtime/src/snapshot_utils.rs
```diff
@@ -958,6 +974,87 @@ where
...
+/// Check to make sure the deserialized bank's slot and hash matches the snapshot archive's slot
+/// and hash
+fn check_bank_with_snapshot_archive_info(
+    bank: &Bank,
+    snapshot_archive_info_slot: Slot,
+    snapshot_archive_info_hash: Hash,
+) -> Result<()> {
+    let bank_slot = bank.slot();
+    let bank_hash = bank.get_accounts_hash();
+    (bank_slot == snapshot_archive_info_slot && bank_hash == snapshot_archive_info_hash)
```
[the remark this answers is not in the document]
brooksprumo: OK will do!

core/tests/snapshots.rs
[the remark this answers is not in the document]
lijunwangs: For me, Boolean value is always confusing, true for what, false for what? You do not know it until you check out the function signature -- and worse for return values: the implementation. So I always prefer some variables. But I also agree we have the hard coded one everywhere -- as it is easier to write -- so it is not you who is initiating it -- but just saw the change and thus the comment.

runtime/src/snapshot_utils.rs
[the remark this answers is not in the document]

runtime/src/snapshot_utils.rs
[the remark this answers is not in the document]

brooksprumo:
@carllin @ryoqun Let me know if you'd like more time to review this PR. If yes, leave a comment here. If not, no worries! I won't merge before tomorrow mid-day.

core/tests/snapshots.rs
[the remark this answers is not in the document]
carllin: yeah let's keep the named variables where possible, this is also the benefit of having argument structs instead of long lists of arguments.

runtime/src/snapshot_utils.rs
```diff
@@ -958,6 +974,96 @@ where
...
+            |incremental_snapshot_archive_info| incremental_snapshot_archive_info.hash(),
+        ),
+    )?;
+
+    Ok((bank, timings))
+}
+
+/// Check to make sure the deserialized bank's slot and hash matches the snapshot archive's slot
+/// and hash
+fn check_bank_with_snapshot_archive_info(
```
carllin, line 1049: nit: rename to `verify_bank_against_expected_hash`

runtime/src/snapshot_utils.rs
```diff
@@ -958,6 +974,96 @@ where
...
+    bank: &Bank,
+    snapshot_archive_info_slot: Slot,
+    snapshot_archive_info_hash: Hash,
```
carllin, line 1052: I think we can just rename these to `expected_slot` and `expected_hash`

runtime/src/snapshot_utils.rs
```diff
@@ -1722,12 +1823,19 @@ pub fn package_process_and_archive_full_snapshot(
...
         maximum_snapshots_to_retain,
-    )
+    )?;
+
+    Ok(FullSnapshotArchiveInfo::new(
+        package.tar_output_file,
+        package.slot,
+        package.hash,
+        package.archive_format,
+    ))
```
carllin, line 1838: can we embed  `SnapshotArchiveInfo` inside `AccountsPackage` (replacing these 4 fields `tar_output_file, slot, hash, archive_format`, then this call could look like 

```
Ok(FullSnapshotArchiveInfo::new(package.snapshot_archive_info))
```

runtime/src/snapshot_utils.rs
```diff
@@ -145,11 +159,11 @@ impl IncrementalSnapshotArchiveInfo {
         &self.inner.slot
     }
 
-    fn _hash(&self) -> &Hash {
+    fn hash(&self) -> &Hash {
         &self.inner.hash
     }
 
-    fn _archive_format(&self) -> &ArchiveFormat {
+    fn archive_format(&self) -> &ArchiveFormat {
```
carllin, line 166: Side note, I think a good way to reduce this code duplication between `FullSnapshotArchiveInfo` and `IncrementalSnapshotArchiveInfo` is to create a trait `SnapshotArchiveInfo` like so:

```
trait SnapshotArchiveInfo {
   // constructor 
    fn from_snapshot_archive_info(snapshot_archive_info: SnapshotArchiveInfo) -> Self;
    // getter
    fn snapshot_archive_info(&self) -> &SnapshotArchiveInfo;

  // default methods to avoid duplication
    fn hash(&self) -> &Hash {
        &self.snapshot_archive_info().hash
    }

    fn archive_format(&self) -> &ArchiveFormat {
      &self.snapshot_archive_info().archive_format
      }
}

```

runtime/src/snapshot_utils.rs
```diff
@@ -958,6 +974,96 @@ where
     Ok((bank, timings))
 }
 
+/// Rebuild bank from snapshot archives.  This function searches `snapshot_archives_dir` for the
+/// highest full snapshot and highest corresponding incremental snapshot, then rebuilds the bank.
+#[allow(clippy::too_many_arguments)]
+pub fn bank_from_latest_snapshot_archives(
```
carllin, line 980: nit: want to communicate this function also verifies the hash

What do you think of `get_and_verify_bank_from_latest_snapshot_archives()`, too verbose?

runtime/src/snapshot_utils.rs
```diff
@@ -958,6 +974,96 @@ where
...
+        incremental_snapshot_archive_info
+            .as_ref()
+            .map(
+                |incremental_snapshot_archive_info| incremental_snapshot_archive_info
+                    .path()
+                    .display()
+            )
+    );
+
+    let (bank, timings) = bank_from_snapshot_archives(
```
carllin, line 1015: heh just realized the real hashing/verification of the account state happens in here, when we call down to

`verify_snapshot_bank() -> verify_bank_hash() -> verify_bank_hash_and_lamports() -> calculate_accounts_hash_helper_with_verify()`, and we verify the 
`calculated_hash` of the accounts state against the expected serialized `bank_hashes` in the snapshot

runtime/src/snapshot_utils.rs
```diff
@@ -2676,4 +2804,101 @@ mod tests {
 
         assert_eq!(*bank4, roundtrip_bank);
     }
+
+    /// Test rebuilding bank from the latest snapshot archives
+    #[test]
+    fn test_bank_from_latest_snapshot_archives() {
```
carllin, line 2810: thanks for the test!

core/tests/snapshots.rs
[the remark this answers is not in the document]
brooksprumo: Got it. Done!

runtime/src/snapshot_utils.rs
brooksprumo: Love it. Done!

runtime/src/snapshot_utils.rs
brooksprumo: I like this idea! One thing is that the `IncrementalSnapshotArchiveInfo` needs a `base_slot` parameter too. Do you know if there's a way to handle this within the Trait?

runtime/src/snapshot_utils.rs

runtime/src/snapshot_utils.rs
brooksprumo: Ooo I love it! Will do.

brooksprumo: OK, done!

runtime/src/snapshot_utils.rs
brooksprumo: I'm implemented the trait but _without_ a constructor. I think for now that should be good.

The next thing I'm working on is adding in incremental snapshot support to the background services (AccountsBackgroundService, AcccountsHashVerifier, SnapshotPackagerService). I'll by building out the SnapshotArchiveInfo more. In particular, moving Full and Incremental under a single enum, so that way when a background service makes a package, it'll contain the info about full vs incremental, and it'll carry through to the next service.

So yeah, is this OK for now? I did change `slot()` and `archive_format()` to not return references anymore, since they are `Copy` and small, and always used as values anyway.

runtime/src/snapshot_utils.rs
```diff
@@ -39,54 +39,69 @@ use {
...
+    fn archive_format(&self) -> ArchiveFormat {
+        self.snapshot_archive_info_inner().archive_format
+    }
+}
+
 /// Common information about a snapshot archive
-#[derive(PartialEq, Eq, Debug)]
-struct SnapshotArchiveInfo {
+#[derive(PartialEq, Eq, Debug, Clone)]
+pub struct SnapshotArchiveInfoInner {
```
carllin, line 65: hmmm I think I like the `SnapshotArchiveInfo` name here, and should instead rename the trait

runtime/src/snapshot_utils.rs
```diff
@@ -39,54 +39,69 @@ use {
     thiserror::Error,
 };
 
+/// Trait to query the snapshot archive information
+pub trait SnapshotArchiveInfo {
```
carllin, line 43: maybe rename to `SnapshotArchiveInfoGetter`

carllin (approved):

runtime/src/snapshot_utils.rs
brooksprumo: OK, will do!

runtime/src/snapshot_utils.rs

Issue #18993, opened 2021-07-30
Define all cost related const values in one place


taozhu-chicago:
#### Problem
there are several `const` variables defined few places, eg `cost_model.rs`, `block_cost_limits.rs`, maybe other places. Should put them under single roof.

#### Proposed Solution
Propose to move all `const` into `block_cost_limits.rs`, which is currently under `ledger/src`, since this file will be shared cross multiple crates (core, ledger, etc).

PR #18994, opened 2021-07-30
Reject blocks for costs above the max block cost

### LINKED ISSUE #18944: Reject blocks for costs above the max block cost
#### Problem
Validator should check and reject too-large blocks.

#### Proposed Solution
- Determine Max Block Cost
   Use the worst case scenario to determine Max Block Cost (in order to prevent additional forking); find the costliest instruction on mainnet, also find the number of bpf instructions per block; multiply these two would be Max Block Cost as in worst case scenario. 
- Determine block cost in realtime
  `blockstore_process::confirm_slot()` executes batches of transactions in slot in parallel, we can share a `block_cost_cap` with execution threads, each thread will deduct from it the executed batch's cost, as soon as the fall below zero (eg, the max allowed block limit is consumed), `execute_batch()` errors out, error code passed up to `confirm_slot()` which will eventually reject and mark the block as dead.

taozhu-chicago:
#### Problem
We don't want a block that is too large (say takes more than 400ms to process). 
While producer limits block size using best-effort approach with `cost model`, validator should reject blocks as soon as it determined its cost exceeds max limit, including those blocks with malicious intention. 

#### Summary of Changes
added realtime cost checking logic to reject block that would exceed max limit:
- defines max limits at `block_cost_limits.rs`, also removed associated cost for `signer` for now as it is not well defined, and makes very small difference.
- right after each bath's execution, accumulate its cost and check again
  limit, return error if limit is exceeded
- add feature `gate_large_block` for this change 
- moved cost `const` def into `block_cost_limits.rs`

Fixes #18944, #18993

### DIFF
--- a/core/src/replay_stage.rs
+++ b/core/src/replay_stage.rs
@@ -1935,7 +1935,6 @@ impl ReplayStage {
                     replay_vote_sender,
                     verify_recyclers,
                 );
-                execute_timings.accumulate(&bank_progress.replay_stats.execute_timings);
                 match replay_result {
                     Ok(replay_tx_count) => tx_count += replay_tx_count,
                     Err(err) => {
@@ -1961,6 +1960,12 @@ impl ReplayStage {
             }
             assert_eq!(*bank_slot, bank.slot());
             if bank.is_complete() {
+                execute_timings.accumulate(&bank_progress.replay_stats.execute_timings);
+                debug!("bank {} is completed replay from blockstore, contribute to update cost with {:?}",
+                       bank.slot(),
+                       bank_progress.replay_stats.execute_timings
+                       );
+
                 bank_progress.replay_stats.report_stats(
                     bank.slot(),
                     bank_progress.replay_progress.num_entries,
--- a/core/src/replay_stage.rs
+++ b/core/src/replay_stage.rs
@@ -2040,9 +2040,11 @@ impl ReplayStage {
         }
 
         // send accumulated excute-timings to cost_update_service
-        cost_update_sender
-            .send(execute_timings)
-            .unwrap_or_else(|err| warn!("cost_update_sender failed: {:?}", err));
+        if !execute_timings.details.per_program_timings.is_empty() {
+            cost_update_sender
+                .send(execute_timings)
+                .unwrap_or_else(|err| warn!("cost_update_sender failed: {:?}", err));
+        }
 
         inc_new_counter_info!("replay_stage-replay_transactions", tx_count);
         did_complete_bank
--- a/ledger/src/blockstore_processor.rs
+++ b/ledger/src/blockstore_processor.rs
@@ -179,7 +179,8 @@ fn execute_batch(
         .unwrap()
         .accumulate(execution_cost_units);
     debug!(
-        "batch of {} transactions, total execute cu {}, remaining block cost cap {}",
+        "bank {:?} batch of {} transactions, total execute cu {}, remaining block cost cap {}",
+        bank.slot(),
         batch.sanitized_transactions().len(),
         execution_cost_units,
         remaining_block_cost_cap
--- a/ledger/src/blockstore_processor.rs
+++ b/ledger/src/blockstore_processor.rs
@@ -1,6 +1,7 @@
 use crate::{
-    block_error::BlockError, blockstore::Blockstore, blockstore_db::BlockstoreError,
-    blockstore_meta::SlotMeta, leader_schedule_cache::LeaderScheduleCache,
+    block_cost_limits::*, block_error::BlockError, blockstore::Blockstore,
+    blockstore_db::BlockstoreError, blockstore_meta::SlotMeta,
+    leader_schedule_cache::LeaderScheduleCache,
 };
 use chrono_humanize::{Accuracy, HumanTime, Tense};
 use crossbeam_channel::Sender;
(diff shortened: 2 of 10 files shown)


taozhu-chicago:
broke a coverage test ... looking into it ...

ledger/src/blockstore_processor.rs
```diff
@@ -127,6 +172,22 @@ fn execute_batch(
...
+    debug!(
+        "batch of {} transactions, total execute cu {}, remaining block cost cap {}",
+        batch.sanitized_transactions().len(),
+        execution_cost_units,
+        remaining_block_cost_cap
+    );
+    if remaining_block_cost_cap == 0_u64 {
+        return Err(TransactionError::WouldExceedMaxBlockCostLimit);
+    }
+
```
carllin, line 190: This should be gated behind a feature so that validators don't fork off

taozhu-chicago: Yea, it'd be safer to featurize it. Will make the change.

aeyakovenko:
@taozhu-chicago @sakridge is this totally deterministic and based on some globally defined threshold?

taozhu-chicago:
> @taozhu-chicago @sakridge is this totally deterministic and based on some globally defined threshold?

the Max Block Limit is a fixed number, determined as the "worst case scenario", eg. = worst_instruction_compute_unit * max_instruction_per_block, observed from mainnet today.

carllin:
> the Max Block Limit is a fixed number, determined as the "worst case scenario", eg. = worst_instruction_compute_unit * max_instruction_per_block, observed from mainnet today.

@taozhu-chicago, is the `worst_instruction_compute_unit` consistent across all the validators as well, i.e. they are all using the same cost function?

taozhu-chicago:
> > the Max Block Limit is a fixed number, determined as the "worst case scenario", eg. = worst_instruction_compute_unit * max_instruction_per_block, observed from mainnet today.
> 
> @taozhu-chicago, is the `worst_instruction_compute_unit` consistent across all the validators as well, i.e. they are all using the same cost function?

Yes, computer_unit is calculated by @jackcmay 's `tx_wide_compute_cap` feature, at https://github.com/solana-labs/solana/blob/master/runtime/src/message_processor.rs#L1276-L1308 
It should be consistent cross nodes.

core/src/replay_stage.rs
```diff
@@ -1949,6 +1948,12 @@ impl ReplayStage {
             }
             assert_eq!(*bank_slot, bank.slot());
             if bank.is_complete() {
+                execute_timings.accumulate(&bank_progress.replay_stats.execute_timings);
```
sakridge, line 1951: So if `bank.is_complete()`, then it just sends completely empty `execute_timings`?

taozhu-chicago:
in determining the compute-unit for signature and read/write account ops: 
1. say `max block time` = 400ms / block
2. on average, a instruction takes 1ms, with 10X parallelism, `max instruction per block ` = ( `max block time` / 1ms ) * 10 = 4,000
3. `max block cost (cu) ` = `max instruction per block` * `max instruction cost (cu) ` = 4,000 * 200,000 = 800,000,000
4. therefore get `CU / us ratio` = ` max block cost (cu) ` / ( 1000 * `max block time` ) = 2,000

- signature takes 10us, with #4, it comes to 20K CU, which equals max 40K signatures per block;
- read account takes 5us, with #4, it comes to 10K CU, which equals max 80K reads per block;
- write account takes 25us, with #4, it comes to 50K CU, which equals max 16K writes per block.

core/src/replay_stage.rs
taozhu-chicago: yea, it'd send empty `execute_timings` if no bank is complete in this loop. However, on the receiving side, empty timings are ignored (https://github.com/solana-labs/solana/blob/master/core/src/cost_update_service.rs#L140-L143).

sakridge: Why waste the cycles of putting the thing in the channel and receiving if it just ignores it though? Hard to follow that kind of logic too, where it does extra work and nothing happens.

Can we not just send at the end of the block?

taozhu-chicago: yea, fair enough. Originally thought was to keep the flow uniform, at cost of wasted cycles. Can add a small check at the end of function before sending

sakridge (approved):
lgtm

behzadnouri:
@taozhu-chicago There is a 4x regression in `GCE 1 Hour perf stability - 5 nodes, 1 region, CPU only` job, and `git bisect` is pointing to this change. Would you please take a look?

Also this is the 2nd time cost-model has caused regressions in that job; So it seems like benign looking cost-model changes can easily introduce performance regressions. In the future, can you please run the perf-test before merging such changes?

taozhu-chicago:
> @taozhu-chicago There is a 4x regression in `GCE 1 Hour perf stability - 5 nodes, 1 region, CPU only` job, and `git bisect` is pointing to this change. Would you please take a look?
> 
> Also this is the 2nd time cost-model has caused regressions in that job; So it seems like benign looking cost-model changes can easily introduce performance regressions. In the future, can you please run the perf-test before merging such changes?

@behzadnouri thanks for bringing it up. Will pay closer attention to the perf-test. I'll look into the regression now

PR #18998, opened 2021-07-30
Fix serialization of parameters in RpcClient::get_block_production_with_config

brson:
#### Problem

RpcClient::get_block_production_with_config always fails because the RPC params must be an array.

This function is not used anywhere in Solana or the SPL.

#### Summary of Changes

Change the function to provide the correct type for the RPC params.

Add a unit test to capture the fix.

I also have an integration test for this using TestValidator in another patch to be submitted later.


CriesofCarrots (approved):
lgtm, thank you for the fix!

PR #18999, opened 2021-07-30
allows sendmmsg api taking owned values (as well as references)

behzadnouri:
#### Problem
Current signature of api in sendmmsg requires a slice of inner
references:
https://github.com/solana-labs/solana/blob/fe1ee4980/streamer/src/sendmmsg.rs#L130-L152

That forces the call-site to convert owned values to references even
though doing so is redundant and adds an extra level of indirection:
https://github.com/solana-labs/solana/blob/fe1ee4980/core/src/repair_service.rs#L291

#### Summary of Changes
This commit expands the api using AsRef and Borrow traits to allow
calling the method with owned values (as well as references like
before).


jbiseda (approved):
lgtm

PR #19021, opened 2021-08-03
Add more API documentation for Rust RpcClient

brson:
This adds more API docs and examples for the Rust RpcClient.

Followup to https://github.com/solana-labs/solana/pull/18748

The biggest changes here are relatively complete docs for `get_signature_statuses`, `get_signature_statuses_with_history`, `confirm_transaction`, and `confirm_transaction_with_commitment`. I have also added preliminary examples for more methods as I familiarize myself with the code. Some of these examples need improvement still, and I intend to revisit them when I write their full docs.

It's been about two weeks of work since my last patch, and this patch is pretty sizable, so I am looking for feedback. After this PR I plan to continue documenting this module. I also have a patch of integration tests for RpcClient that I intend to submit eventually.

### DIFF
--- a/sdk/src/transaction.rs
+++ b/sdk/src/transaction.rs
@@ -113,9 +113,15 @@ impl From<SanitizeError> for TransactionError {
 #[frozen_abi(digest = "AAeVxvWiiotwxDLxKLxsfgkA6ndW74nVbaAEb6cwJYqR")]
 #[derive(Debug, PartialEq, Default, Eq, Clone, Serialize, Deserialize, AbiExample)]
 pub struct Transaction {
-    /// A set of digital signatures of `account_keys`, `program_ids`, `recent_blockhash`, and `instructions`, signed by the first
-    /// signatures.len() keys of account_keys
-    /// NOTE: Serialization-related changes must be paired with the direct read at sigverify.
+    /// A set of digital signatures of the [`account_keys`], [`recent_blockhash`], and [`instructions`] fields
+    /// of [`message`], signed by the first `signatures.len()` keys of `account_keys`.
+    ///
+    /// [`account_keys`]: Message::account_keys
+    /// [`recent_blockhash`]: Message::recent_blockhash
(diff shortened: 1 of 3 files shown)


brson:
Another thing this patch does is add an "RPC Referenc" section to some of the methods that links to the RPC spec for the underlying protocol. The intent is to avoid having to reproduce the protocol documentation in the API.

sdk/src/transaction.rs
```diff
@@ -113,9 +113,15 @@ impl From<SanitizeError> for TransactionError {
...
-    /// A set of digital signatures of `account_keys`, `program_ids`, `recent_blockhash`, and `instructions`, signed by the first
-    /// signatures.len() keys of account_keys
-    /// NOTE: Serialization-related changes must be paired with the direct read at sigverify.
+    /// A set of digital signatures of the [`account_keys`], [`recent_blockhash`], and [`instructions`] fields
+    /// of [`message`], signed by the first `signatures.len()` keys of `account_keys`.
+    ///
+    /// [`account_keys`]: Message::account_keys
+    /// [`recent_blockhash`]: Message::recent_blockhash
+    /// [`instructions`]: Message::instructions
+    /// [`message`]: Transaction::message
```
CriesofCarrots, line 122: I know the original docs implied otherwise, but the signatures actually include the complete Message, not just the specified fields. How about simply...

```suggestion
    /// A set of digital signatures of a serialized [`Message`],
    /// signed by the first `signatures.len()` keys of `account_keys`.
    ///
    /// [`account_keys`]: Message::account_keys
```

CriesofCarrots (commented):
Thank you! Just a couple suggestions and nits. (Sorry that there's no good way to preserve 80-char wrapping in github suggestions!)

brson:
Updated with all suggestions resolved. Thanks for the good review @CriesofCarrots

CriesofCarrots (approved):
Thanks! lgtm
I'll hand-hold it through CI

PR #19025, opened 2021-08-03
Plug two leaks around dirty_stores

ryoqun:
#### Problem

there is regression since #17601

1. _at start-up (`verify_snapshot_bank`)_, dirty_stores accumulates `Arc`s to all shrunken stores because `clean_accounts` aren't run in `verify_snapshot_bank` after the initial `clean_accounts()`.
So that keeps mmap leaks. This surge of mmap use here is rather transient; after all shrinking and next clean_accounts(), it's usually `munmap()`ed en masse.

2. also the initial `clean_accounts()` isn't complete as was before. That's because `candidate_clean_keys` are created from empty `dirty_stores`. So, ultimately, this leaves more appendvecs than ideally minimized.

BEFORE 

note the almost instant `dirty_store_processing_us`:

```
[2021-08-03T15:17:40.939120664Z INFO  solana_metrics::metrics] datapoint: clean_accounts collect_delta_keys_us=38365i dirty_store_processing_us=33i accounts_scan=521229i clean_old_rooted=5963348i store_counts=417554i purge_filter=57567i calc_deps=107738i reclaims=69i delta_key_count=1076528i dirty_pubkeys_count=0i total_keys_count=1076528i
```

#### Summary of Changes

1. make dirty_stores population logic more selective.

2. correctly populate dirty_stores in generate_index().

AFTER

```
[2021-08-03T15:33:58.804400247Z INFO  solana_metrics::metrics] datapoint: clean_accounts collect_delta_keys_us=35957i dirty_store_processing_us=50367291i accounts_scan=31850539i clean_old_rooted=22471009i store_counts=952847i purge_filter=229714i calc_deps=395295i reclaims=993662i delta_key_count=79001842i dirty_pubkeys_count=79001842i total_keys_count=79001842i

```

LEFT is before, right is after

image

#### Remarks:

Also, I considered following approaches but turned them down:

- completely delay populating `dirty_stores` while booting
  - somewhat unknown risks for the introduction of bi-modal behavior; also this patch needs to go v1.7 rather quickly for timely mainnet-beta update
- make `dirty_stores`: `DashMap<_, std:sync::Arc<_>>`
  - generally doesn't work, particularly bad with store recycler (as it recycles when count == 1).

note that there is still very slow leaks, which is occurring even on v1.6.... this pr doesn't address that leak.

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -2097,7 +2097,7 @@ impl AccountsDb {
         );
     }
 
-    fn do_shrink_slot_stores<'a, I>(&'a self, slot: Slot, stores: I, _is_startup: bool) -> usize
+    fn do_shrink_slot_stores<'a, I>(&'a self, slot: Slot, stores: I, is_startup: bool) -> usize
     where
         I: Iterator<Item = &'a Arc<AccountStorageEntry>>,
     {
@@ -2254,8 +2254,10 @@ impl AccountsDb {
             if let Some(slot_stores) = self.storage.get_slot_stores(slot) {
                 slot_stores.write().unwrap().retain(|_key, store| {
                     if store.count() == 0 {
-                        self.dirty_stores
-                            .insert((slot, store.append_vec_id()), store.clone());
+                        if !is_startup {
+                            self.dirty_stores
+                                .insert((slot, store.append_vec_id()), store.clone());
+                        }
                         dead_storages.push(store.clone());
                         false
                     } else {
@@ -5233,8 +5235,11 @@ impl AccountsDb {
                 );
                 let count = store.remove_account(account_info.stored_size, reset_accounts);
                 if count == 0 {
-                    self.dirty_stores
-                        .insert((*slot, store.append_vec_id()), store.clone());
+                    // expected_slot is none should exclude write path from flush/shrink/bank drop
+                    if expected_slot.is_none() {
+                        self.dirty_stores
+                            .insert((*slot, store.append_vec_id()), store.clone());
+                    }
                     dead_slots.insert(*slot);
                 } else if self.caching_enabled
                     && Self::is_shrinking_productive(*slot, &[store.clone()])
@@ -5738,11 +5743,16 @@ impl AccountsDb {
         }
     }
 
-    pub fn add_root(&self, slot: Slot) {
-        self.accounts_index.add_root(slot, self.caching_enabled);
+    pub fn add_root_to_squash_bank(&self, slot: Slot) {
+        self.add_root_to_update_index(slot, self.caching_enabled);
         if self.caching_enabled {
             self.accounts_cache.add_root(slot);
         }
+    }
+
+    pub fn add_root_to_update_index(&self, slot: Slot, caching_enabled: bool) {
+        self.accounts_index.add_root(slot, caching_enabled);
+
         if let Some(slot_stores) = self.storage.get_slot_stores(slot) {
             for (store_id, store) in slot_stores.read().unwrap().iter() {
                 self.dirty_stores.insert((slot, *store_id), store.clone());
@@ -6009,7 +6019,7 @@ impl AccountsDb {
             if pass == 0 {
                 // Need to add these last, otherwise older updates will be cleaned
                 for slot in &slots {
-                    self.accounts_index.add_root(*slot, false);
+                    self.add_root_to_update_index(*slot, false);
                 }
 
                 self.initialize_storage_count_and_alive_bytes(&mut timings);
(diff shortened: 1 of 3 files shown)


runtime/src/accounts_db.rs
```diff
@@ -5233,8 +5235,11 @@ impl AccountsDb {
                 );
                 let count = store.remove_account(account_info.stored_size, reset_accounts);
                 if count == 0 {
-                    self.dirty_stores
-                        .insert((*slot, store.append_vec_id()), store.clone());
+                    // expected_slot is none should exclude write path from flush/shrink/bank drop
```
ryoqun, line 5238: kind of hack, but `handle_reclaims()` code path is literally maze.

runtime/src/accounts_db.rs
```diff
@@ -2254,8 +2254,10 @@ impl AccountsDb {
             if let Some(slot_stores) = self.storage.get_slot_stores(slot) {
                 slot_stores.write().unwrap().retain(|_key, store| {
                     if store.count() == 0 {
-                        self.dirty_stores
-                            .insert((slot, store.append_vec_id()), store.clone());
+                        if !is_startup {
```
ryoqun, line 2257: admittedly, this is another hack. xD

glad that we have managed to revive the almost-dying flag (`_`-prefixed) for quick win. :)

ryoqun:
note: there is no test...

also, here is a crude way to detect leak when shrink_all is running:

```

$ pmap -p 217758 > pmap.bad.maybe-fixed.$(date '+%Yy%mm%dd%Hh%Mm%Ss%Nns')

$ wc -l pmap.bad.maybe-fixed.2021y08m03d0*
   480052 pmap.bad.maybe-fixed.2021y08m03d05h16m26s137495180ns
   546836 pmap.bad.maybe-fixed.2021y08m03d05h16m37s707483373ns
   561374 pmap.bad.maybe-fixed.2021y08m03d05h17m10s011257814ns
   578413 pmap.bad.maybe-fixed.2021y08m03d05h17m48s074379085ns
   400871 pmap.bad.maybe-fixed.2021y08m03d05h43m43s268314673ns
   400867 pmap.bad.maybe-fixed.2021y08m03d05h43m52s882543697ns
   400873 pmap.bad.maybe-fixed.2021y08m03d05h44m47s872216956ns
   400872 pmap.bad.maybe-fixed.2021y08m03d05h45m21s393793566ns
   520249 pmap.bad.maybe-fixed.2021y08m03d06h48m20s214631714ns
   583508 pmap.bad.maybe-fixed.2021y08m03d06h48m30s747138617ns
   615597 pmap.bad.maybe-fixed.2021y08m03d06h48m41s378189396ns
   645764 pmap.bad.maybe-fixed.2021y08m03d06h49m26s753318034ns
   516872 pmap.bad.maybe-fixed.2021y08m03d07h51m44s319430684ns
   573594 pmap.bad.maybe-fixed.2021y08m03d07h51m52s085619175ns
   637344 pmap.bad.maybe-fixed.2021y08m03d07h52m23s614841435ns
   447513 pmap.bad.maybe-fixed.2021y08m03d08h18m08s857074170ns
   447513 pmap.bad.maybe-fixed.2021y08m03d08h18m29s111814438ns
   447514 pmap.bad.maybe-fixed.2021y08m03d08h18m55s404999949ns

```

runtime/src/accounts_db.rs
```diff
@@ -6009,7 +6019,7 @@ impl AccountsDb {
             if pass == 0 {
                 // Need to add these last, otherwise older updates will be cleaned
                 for slot in &slots {
-                    self.accounts_index.add_root(*slot, false);
+                    self.add_root_to_update_index(*slot, false);
```
sakridge, line 6022: We add any keys which have more than 1 key: https://github.com/solana-labs/solana/blob/d8984cb0f40eafbbf8aa2a95f3c0e43343e80230/runtime/src/accounts_db.rs#L5904-L5906

This uses the `uncleaned_pubkeys` and not `dirty_stores` which is why the `dirty_store` time is small, but I don't think we need to perform a clean on keys which only have 1 entry in them, no?

The other case is 0-lamport accounts. We can detect that in generate_index and also add it to the clean set.

sakridge: https://github.com/solana-labs/solana/pull/19041

ryoqun, line 6022: here

ryoqun: yeah, now i view things clearly. I totally agree with you.

ryoqun:
superceded by #19041 and #19042

PR #19028, opened 2021-08-03
add validator option --accounts-db-skip-shrink

jeffwashington:
#### Problem
When restarting validators with large snapshots, shrink can take hours. This option is useful to more quickly get to other troublesome parts of validator startup while debugging the system.
#### Summary of Changes

Fixes #

### DIFF
--- a/validator/src/main.rs
+++ b/validator/src/main.rs
@@ -1816,7 +1816,7 @@ pub fn main() {
             Arg::with_name("accounts_db_skip_shrink")
                 .long("accounts-db-skip-shrink")
                 .help("Enables faster starting of validators by skipping shrink. \
-                      This option if for use during testing."),
+                      This option is for use during testing."),
         )
         .arg(
             Arg::with_name("accounts_db_test_hash_calculation")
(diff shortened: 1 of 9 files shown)


jeffwashington:
verified this disables shrink on a validator

validator/src/main.rs
```diff
@@ -1812,6 +1812,12 @@ pub fn main() {
                 .long("no-accounts-db-caching")
                 .help("Disables accounts caching"),
         )
+        .arg(
+            Arg::with_name("accounts_db_skip_shrink")
+                .long("accounts-db-skip-shrink")
+                .help("Enables faster starting of validators by skipping shrink. \
+                      This option if for use during testing."),
```
CriesofCarrots, line 1819:
```suggestion
                      This option is for use during testing."),
```

CriesofCarrots (dismissed):
Lgtm! one typo nit
I'm thinking of plumbing this into the `multinode/net` scripts. Are there any other accounts db or index validator parameters that you think would be useful to be able to toggle from the testnet run commands?

jeffwashington:
> Any plans to backport this one? If not, can we add a hidden stub flag on v1.7 for compatibility

I have no objections to back porting it. It should not add risk if I did it right ;-)

PR #19032, opened 2021-08-03
move towards accounts index being dynamic

jeffwashington:
#### Problem
Currently, the number of account index bins is a constant. This will need to change to be a tunable parameter over time. When we move to a disk-based index, we will not want to create many buckets on disk for each test that creates an accounts index.
#### Summary of Changes
Make # bins more dynamic on the accounts index.
Fixes #


brooksprumo (approved):
lgtm

PR #19034, opened 2021-08-03
Dynamic Account index creation

jeffwashington:
#### Problem
#### Summary of Changes
work on construction of AccountsIndex, the root of all startup problems related to # bins.
Fixes #

### DIFF
--- a/runtime/src/accounts_index.rs
+++ b/runtime/src/accounts_index.rs
@@ -676,15 +676,15 @@ pub struct AccountsIndex<T> {
     pub removed_bank_ids: Mutex<HashSet<BankId>>,
 }
 
-impl<T> Default for AccountsIndex<T> {
+impl<
+        T: 'static + Clone + IsCached + ZeroLamport + std::marker::Sync + std::marker::Send + Debug,
+    > Default for AccountsIndex<T>
+{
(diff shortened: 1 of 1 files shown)


runtime/src/accounts_index.rs
```diff
@@ -705,6 +705,19 @@ impl<
...
 {
+    fn allocate_accounts_index(bins: usize) -> (LockMapType<T>, PubkeyBinCalculator16) {
+        let account_maps = (0..bins)
+            .into_iter()
+            .map(|_| RwLock::new(AccountMap::default()))
+            .collect::<Vec<_>>();
+        let bin_calculator = PubkeyBinCalculator16::new(bins);
+        (account_maps, bin_calculator)
+    }
+    pub fn allocate_index(&mut self, bins: usize) {
```
brooksprumo, line 716: Where does `allocate_index()` get used? I'm guessing that this will be the non-test way to allocate the index, and _not_ the Default way above?

jeffwashington: Got rid of it - it isn't used in the further work built on this anyway. Good catch.

brooksprumo (approved):
Looks good to me!

PR #19041, opened 2021-08-03
Handle 0-lamport account in index generation

sakridge:
#### Problem

0-lamport keys could be cleaned from the snapshot, but are not added to the clean set.

#### Summary of Changes

Add 0-lamport entries to the clean set if they are not added by being a duplicate key.

Fixes #

### DIFF
--- a/runtime/src/accounts_index.rs
+++ b/runtime/src/accounts_index.rs
@@ -1425,8 +1425,9 @@ impl<
         items.for_each(|(pubkey, account_info)| {
             let bin = get_bin_pubkey(&pubkey);
             // this value is equivalent to what update() below would have created if we inserted a new item
+            let is_zero_lamport = account_info.is_zero_lamport();
             let info = WriteAccountMapEntry::new_entry_after_update(slot, account_info);
-            binned[bin].1.push((pubkey, info));
+            binned[bin].1.push((pubkey, info, is_zero_lamport));
         });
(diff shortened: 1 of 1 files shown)


carllin (dismissed):

runtime/src/accounts_index.rs
```diff
@@ -1442,17 +1443,21 @@ impl<
...
+                        let already_exists = self.insert_new_entry_if_missing_with_lock(
+                            pubkey,
+                            &mut w_account_maps,
+                            new_item,
+                        );
+                        if let Some((mut w_account_entry, account_info, pubkey)) = already_exists {
+                            w_account_entry.update(slot, account_info, &mut _reclaims);
+                            duplicate_keys.push(pubkey);
+                        } else if is_zero_lamport {
+                            duplicate_keys.push(pubkey);
```
ryoqun, line 1458: nit: how about renaming `duplicate_keys` to `dirty_pubkeys`:

- _duplicate_ is now misnomer as it technically can contain not-duplicate zero-lamport keys.
- this is what is called at the sole call site (= `generate_index_for_slot`)

ryoqun (approved):
you beat me. :)

way finer solution than my coarse method (https://github.com/solana-labs/solana/pull/19025/files#r682565220).

jeffwashington (dismissed):
lgtm

PR #19044, opened 2021-08-03
expose retransmit_stage and window_service params, add window_service metrics

jbiseda:
- expose additional window_service metrics
- collect constants into param structures for use as configurable defaults
- expose configurable parameters for performance experiments


core/src/window_service.rs
```diff
@@ -68,6 +92,44 @@ impl WindowServiceMetrics {
...
+    num_packets_received: usize,
+    verified_receiver_recv_elapsed_us: u64,
+    handle_packets_elapsed_us: u64,
+    recv_window_elapsed_us: u64,
+}
+
+impl RecvWindowMetrics {
+    pub fn report_metrics(&self, metric_name: &'static str) {
+        datapoint_info!(
+            metric_name,
```
lijunwangs, line 107: Any value making this a parameter compared to a name like "recv_window_metrics"? The last_report logic can be implemented using the AtomicInterval? See ShrinkStats in accounts_db and other patterns.

validator/src/main.rs
```diff
@@ -1883,6 +1897,34 @@ pub fn main() {
...
+                .takes_value(true)
+                .default_value(&default_retransmit_packet_batch_threshold)
+                .hidden(true),
+        )
+        .arg(
+            Arg::with_name("window_service_shred_batch_threshold")
+                .long("window-service-shred-batch-threshold")
+                .takes_value(true)
+                .default_value(&default_window_service_shred_batch_threshold)
+                .hidden(true),
```
lijunwangs, line 1926: Should we worry about checks silly input and guard against them?

validator/src/main.rs
```diff
@@ -1883,6 +1897,34 @@ pub fn main() {
...
+        )
+        .arg(
+            Arg::with_name("retransmit_packet_batch_threshold")
+                .long("retransmit-packet-batch-threshold")
+                .takes_value(true)
+                .default_value(&default_retransmit_packet_batch_threshold)
+                .hidden(true),
+        )
+        .arg(
```
behzadnouri, line 1921: Do we see a use-case for the users running validators with different values for these arguments?

There is usually some risk with adding these arguments at a high level to main validator binary. cc @mvines since previously we had discussions over these flags.

behzadnouri (commented):
Adding more metrics is generally good, so I am +1 on those.

However, I am not convinced about changing `const`s to configurable parameters. Would you please clarify on the intended performance experiments?

Generally we can run performance experiments by changing the code, recompiling, and comparing metrics. Once we arrive at some reasonable parameters, we can leave them as simple `const`s in the code. There is not much gain as leaving them as configurable parameters in the code after that.

Unless there is a need to frequently update some parameter (i.e more often than new release cycles), at that cannot be achieved at compile time by updating the code, I feel like we better avoid that extra code-noise and verbosity or the risk of some future code changes using inconsistent values in different call-sites.

validator/src/main.rs
jbiseda: I was introducing these to make it easier to test the impact of changing these values without recompiling. I didn't intend for them to be generally used by validator operators. They're marked as hidden. I could prefix these flags with something. I see "experimental-poh-pinned-cpu-core". Thoughts?

jbiseda:
> Adding more metrics is generally good, so I am +1 on those.
> 
> However, I am not convinced about changing `const`s to configurable parameters. Would you please clarify on the intended performance experiments?
> 
> Generally we can run performance experiments by changing the code, recompiling, and comparing metrics. Once we arrive at some reasonable parameters, we can leave them as simple `const`s in the code. There is not much gain as leaving them as configurable parameters in the code after that.
> 
> Unless there is a need to frequently update some parameter (i.e more often than new release cycles), at that cannot be achieved at compile time by updating the code, I feel like we better avoid that extra code-noise and verbosity or the risk of some future code changes using inconsistent values in different call-sites.

PR #19061, opened 2021-08-04
Fix wrong old snapshot archives getting purged

### LINKED ISSUE #19057: Wrong snapshot archives are purged as "old"
#### Problem

When purging old snapshot archives, the most recent archives are purged first, instead of the oldest.

#### Proposed Solution

1. Purge the oldest snapshot archives first
2. Add a test to snapshot_utils to catch this in the future

brooksprumo:
I introduced a bug where old snapshot archives were incorrectly purged.
Instead of purged to oldest, I was purged the newest...

The fix is to add a `reverse()` in the purge logic, and I've added a
test to catch this bug in the future.

Fixes #19057

#### Summary of Changes

Fixes #


runtime/src/snapshot_utils.rs
```diff
@@ -2373,6 +2374,53 @@ mod tests {
...
+            assert_eq!(
+                *full_snapshot_archives.first().unwrap().slot(),
+                starting_slot
+            );
+            assert_eq!(*full_snapshot_archives.last().unwrap().slot(), slot);
+            for (i, full_snapshot_archive) in
+                full_snapshot_archives.iter().skip(1).rev().enumerate()
+            {
+                assert_eq!(*full_snapshot_archive.slot(), slot - i as Slot);
+            }
```
brooksprumo, line 2420: The previous tests don't sufficiently check for the remain snapshot archives that are _not_ the first and last ones.  This loop ensures all the non-purged snapshot archives match what we're expecting.

runtime/src/snapshot_utils.rs
```diff
@@ -1270,6 +1270,7 @@ where
     );
     let mut snapshot_archives = get_full_snapshot_archives(&snapshot_archives_dir);
     snapshot_archives.sort_unstable();
+    snapshot_archives.reverse();
```
brooksprumo, line 1273: And the fix. Below you'll see the pop(), which is supposed to pop the oldest snapshot archive, but was popping the newest instead. Then in the for loop, with the skip, this would always skip at least 1, so the oldest snapshot archive was correctly not purged, but for the incorrect logic, hence why the tests were passing...

jeffwashington (dismissed):
lgtm

PR #19063, opened 2021-08-04
Bank::new -> Bank::new_for_benches

jeffwashington:
#### Problem
#### Summary of Changes

Fixes #


brooksprumo (approved):
lgtm

PR #19064, opened 2021-08-04
for all tests, bank::new -> bank::new_for_tests

jeffwashington:
#### Problem

#### Summary of Changes
remove Bank::new()
replace all callers, which were only in tests (not benches), with Bank::new_for_tests()
Fixes #

### DIFF
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -1047,11 +1047,6 @@ impl Default for BlockhashQueue {
 }
 
 impl Bank {
-    pub fn new(genesis_config: &GenesisConfig) -> Self {
-        // this will go away in a coming pr where many replacements in test code will get made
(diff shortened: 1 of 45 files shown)


runtime/src/bank.rs
```diff
@@ -1047,11 +1047,6 @@ impl Default for BlockhashQueue {
 }
 
 impl Bank {
-    pub fn new(genesis_config: &GenesisConfig) -> Self {
```
jeffwashington, line 1050: this is the big change

jeffwashington:
note that compiling (without --tests) passes when Bank::new is removed. bench calls have already been replaced. this pr is just the calls from tests.

brooksprumo (approved):
lgtm

PR #19066, opened 2021-08-04
add traits required by IsCached

jeffwashington:
#### Problem
AccountsIndex is undergoing changes. The slot list contains items of generic type. It is cumbersome to pass along a grab bag of traits to all functions or to add a new required trait (such as clone or debug).

#### Summary of Changes
Add required traits to the existing trait IsCached. We could also create a new trait AccountsIndexItem or something, which requires IsCached. That change seems a little messy and unnecessary atm.

Fixes #

### DIFF
--- a/runtime/src/accounts_index.rs
+++ b/runtime/src/accounts_index.rs
@@ -41,7 +41,7 @@ pub type AccountMap<K, V> = BTreeMap<K, V>;
 
 type AccountMapEntry<T> = Arc<AccountMapEntryInner<T>>;
 
-pub trait IsCached {
+pub trait IsCached: 'static + Clone + Debug {
     fn is_cached(&self) -> bool;
 }
 


runtime/src/accounts_index.rs
```diff
@@ -41,7 +41,7 @@ pub type AccountMap<K, V> = BTreeMap<K, V>;
 
 type AccountMapEntry<T> = Arc<AccountMapEntryInner<T>>;
 
-pub trait IsCached {
+pub trait IsCached: 'static + Clone + Debug {
```
brooksprumo, line 44: I was also expecting to see where `'static + Clone + Debug` was _removed_ from current uses of `IsCached` as well. Is that meant for a later PR then?

jeffwashington: Yes, it's just messy and not required yet. Really, later I have to add, sync + send. That change will now be one line. And, new code written against these types will now not have to string along ~7 traits. I'll clean that all up at some point.

jeffwashington: but your expectation was reasonable...

brooksprumo (dismissed):
lgtm

PR #19083, opened 2021-08-05
rework defaults and construction of accountsdb, accounts, bank

jeffwashington:
#### Problem
#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/accounts.rs
+++ b/runtime/src/accounts.rs
@@ -87,7 +87,7 @@ impl AccountLocks {
 }
 
 /// This structure handles synchronization for db
-#[derive(Default, Debug, AbiExample)]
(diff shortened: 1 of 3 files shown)


runtime/src/accounts.rs
```diff
@@ -87,7 +87,7 @@ impl AccountLocks {
 }
 
 /// This structure handles synchronization for db
-#[derive(Default, Debug, AbiExample)]
```
brooksprumo, line 90: Does the default impl for Default not work here?

jeffwashington: derive impl Default won't once AccountsDb::default() disappears, which is coming soon.

brooksprumo (approved):

PR #19085, opened 2021-08-05
cli for num account index bins

jeffwashington:
#### Problem
validator or ledger-tool verify can pass `--accounts-index-bins 128` to specify the number of in-mem and on-disk buckets to use. This allows tuning a validator to a machine's disk and memory situation.
#### Summary of Changes

Fixes #

### DIFF
--- a/ledger-tool/src/main.rs
+++ b/ledger-tool/src/main.rs
@@ -11,7 +11,7 @@ use serde_json::json;
 use solana_clap_utils::{
     input_parsers::{cluster_type_of, pubkey_of, pubkeys_of},
     input_validators::{
-        is_parsable, is_pubkey, is_pubkey_or_keypair, is_slot, is_valid_percentage,
+        is_bin, is_parsable, is_pubkey, is_pubkey_or_keypair, is_slot, is_valid_percentage,
     },
 };
 use solana_core::cost_model::CostModel;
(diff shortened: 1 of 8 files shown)


ledger-tool/src/main.rs
```diff
@@ -849,6 +849,12 @@ fn main() {
         .long("no-accounts-db-caching")
         .takes_value(false)
         .help("Disables accounts-db caching");
+    let accounts_index_bins = Arg::with_name("accounts_index_bins")
+        .long("accounts-index-bins")
+        .value_name("BINS")
+        .validator(is_bin)
+        .takes_value(true)
+        .help("Number of bins to divide the account index into");
```
brooksprumo, line 857: Is this supposed to be "accounts index" (with the "s") here instead?

validator/src/main.rs
```diff
@@ -1820,6 +1820,14 @@ pub fn main() {
                 .help("Enables faster starting of validators by skipping shrink. \
                       This option is for use during testing."),
         )
+        .arg(
+            Arg::with_name("accounts_index_bins")
+                .long("accounts-index-bins")
+                .value_name("BINS")
+                .validator(is_bin)
+                .takes_value(true)
+                .help("Number of bins to divide the account index into"),
```
brooksprumo, line 1829: Same thing here w.r.t. "accounts index"

brooksprumo (approved):

PR #19087, opened 2021-08-05
rework bank::new_with_paths

jeffwashington:
#### Problem
#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -1065,7 +1065,7 @@ impl Bank {
 
     pub fn new_for_tests(genesis_config: &GenesisConfig) -> Self {
         // this will diverge
-        Self::new_with_paths(
+        Self::new_with_paths_for_tests(
             genesis_config,
             Vec::new(),
             &[],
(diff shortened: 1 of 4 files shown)


runtime/src/bank.rs
```diff
@@ -1104,7 +1104,31 @@ impl Bank {
...
+            debug_keys,
+            additional_builtins,
+            account_indexes,
+            accounts_db_caching_enabled,
+            shrink_ratio,
+            debug_do_not_add_builtins,
+        )
+    }
+
+    pub fn new_with_paths_production(
```
brooksprumo, line 1131: In all the previous PRs the functions without `_for_tests` meant for production. Why the switch here? (I am assuming the "production" here means not tests/benches.)

jeffwashington: Yes, that function is about to be refactored away. I did not want to leave the old, non-test function around and allow other callers to call it in the meantime accidentally. It is confusing. Trying to take clear, obvious steps towards cleaving the paths into thirds: benches, tests, production/normal/non-test-or-bench

brooksprumo (approved):

PR #19090, opened 2021-08-05
implements copy-on-write for staked-nodes

behzadnouri:
#### Problem
`Bank::staked_nodes` and `Bank::epoch_staked_nodes` redundantly clone
staked-nodes `HashMap` even though an immutable reference will suffice:
https://github.com/solana-labs/solana/blob/a9014cece/runtime/src/vote_account.rs#L77

#### Summary of Changes
This commit implements copy-on-write semantics for staked-nodes by
wrapping the underlying `HashMap` in `Arc<...>`.

### DIFF
--- a/runtime/src/vote_account.rs
+++ b/runtime/src/vote_account.rs
@@ -1,16 +1,19 @@
-use serde::de::{Deserialize, Deserializer};
-use serde::ser::{Serialize, Serializer};
-use solana_sdk::{
-    account::Account, account::AccountSharedData, instruction::InstructionError, pubkey::Pubkey,
-};
-use solana_vote_program::vote_state::VoteState;
-use std::{
-    borrow::Borrow,
-    cmp::Ordering,
-    collections::{hash_map::Entry, HashMap},
-    iter::FromIterator,
-    ops::Deref,
-    sync::{Arc, Once, RwLock, RwLockReadGuard},
+use {
+    itertools::Itertools,
+    serde::de::{Deserialize, Deserializer},
+    serde::ser::{Serialize, Serializer},
+    solana_sdk::{
+        account::Account, account::AccountSharedData, instruction::InstructionError, pubkey::Pubkey,
+    },
+    solana_vote_program::vote_state::VoteState,
+    std::{
+        borrow::Borrow,
+        cmp::Ordering,
+        collections::{hash_map::Entry, HashMap},
+        iter::FromIterator,
+        ops::Deref,
+        sync::{Arc, Once, RwLock, RwLockReadGuard},
+    },
 };
 
 // The value here does not matter. It will be overwritten
@@ -31,10 +34,14 @@ pub struct VoteAccount {
 #[derive(Debug, AbiExample)]
 pub struct VoteAccounts {
     vote_accounts: HashMap<Pubkey, (u64 /*stake*/, ArcVoteAccount)>,
+    // Inner Arc is meant to implement copy-on-write semantics as opposed to
+    // sharing mutations (hence RwLock<Arc<...>> instead of Arc<RwLock<...>>).
     staked_nodes: RwLock<
-        HashMap<
-            Pubkey, // VoteAccount.vote_state.node_pubkey.
-            u64,    // Total stake across all vote-accounts.
+        Arc<
+            HashMap<
+                Pubkey, // VoteAccount.vote_state.node_pubkey.
+                u64,    // Total stake across all vote-accounts.
+            >,
         >,
     >,
     staked_nodes_once: Once,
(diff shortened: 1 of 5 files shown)


runtime/src/vote_account.rs
```diff
@@ -119,9 +125,9 @@ impl VoteAccounts {
     fn add_node_stake(&mut self, stake: u64, vote_account: &ArcVoteAccount) {
         if stake != 0 && self.staked_nodes_once.is_completed() {
             if let Some(node_pubkey) = vote_account.node_pubkey() {
-                self.staked_nodes
-                    .write()
-                    .unwrap()
+                let mut staked_nodes = self.staked_nodes.write().unwrap();
+                let staked_nodes = Arc::make_mut(&mut staked_nodes);
+                staked_nodes
```
carllin, line 130: hmmmm does this only add the new entry to the local copy of `staked_nodes`? This won't persist the insert into the `self.staked_nodes` for other callers to observe right?

From what I can tell this function is only called from `Bank::store_account()` which is used in ledger-tool and probably expects the store to be persisted.

behzadnouri: It clones the inner pointer, so the changes are persisted.
Please see the added `test_staked_nodes_cow` or the example in `Arc::make_mut`:
https://doc.rust-lang.org/std/sync/struct.Arc.html#method.make_mut

carllin: I see, I mixed up what was being cloned, thanks for the clarification.

And the cost of cloning is acceptable because `Bank::store_account()` is only used in ledger-tool where there are no other outstanding references that would trigger this clone?

To avoid silent perf issues in the future if somebody calls `store_account` more intensively,  should we be more explicit about this and use `get_mut()` instead of `make_mut()` and return a boolean from this function to indicate success/failure,

behzadnouri: This code only makes a clone if:
* `.staked_nodes` was invoked since last mutation (i.e. since last `make_mut`).
* That returned `Arc<...>` is still alive when you want to mutate the hash-map again (i.e. when you call `make_mut`).

Current master code is making a clone every time `.staked_ndoes` is invoked:
https://github.com/solana-labs/solana/blob/fd937548a/runtime/src/vote_account.rs#L77

So this code always does fewer clones than current master code, and `make_mut` does not necessarily make a clone.

sakridge (approved):
lgtm

runtime/src/vote_account.rs
carllin: > That returned Arc<...> is still alive when you want to mutate the hash-map again (i.e. when you call make_mut).

Right, I'm worried about these calls https://github.com/solana-labs/solana/blob/fd937548a01ecbfae4fd0fef25efe4ad029dcffe/ledger-tool/src/main.rs#L2064-L2067 silently incurring a clone in the future if some other thread is holding another reference to this `Arc`. I feel like being explicit with a result from something like a `store_without_clone` that we can assert on makes this more transparent, and avoids silently incurring this cost.

behzadnouri: Can you please describe the scenario that previously it was not making a clone and now it does?

1. Previously [*every time* one calls `.staked_nodes()` it had to make a clone](https://github.com/solana-labs/solana/blob/fd937548a/runtime/src/vote_account.rs#L77). No exceptions there.
2. Now it clones only if *all* these happen:
    * you call `.staked_nodes()`, 
    * and hold on to the returned `Arc<...>` 
    * and then also try to mutate stake-nodes while holding to that `Arc`.

(2) is a subset of (1), so for every time that this code ends up making a clone, it was making a clone previously anyways; and so this code is removing clones, not adding clones.

carllin: >Can you please describe the scenario that previously it was not making a clone and now it does?

I don't think there is one currently, was referring to future-proofing calls like https://github.com/solana-labs/solana/blob/fd937548a01ecbfae4fd0fef25efe4ad029dcffe/ledger-tool/src/main.rs#L2064-L2067 in case anybody does add a case where they hold a returned `Arc`. This way anybody calling these functions will be cognizant of when a clone is incurred.

behzadnouri: Still like current call-sites, this code cannot introduce any *new* clones (where with the old code there was no clone).
I would be happy to discuss this more over a call or something next time we chat. For now, any objections to merge this?

carllin: > I would be happy to discuss this more over a call or something next time we chat
Sounds good, no other objections!

carllin (approved):

runtime/src/vote_account.rs
behzadnouri: thank you :pray:

Issue #18587, opened 2021-07-11
Program account notifications are not sent for every confirmed block


Arrowana:
#### Problem

It appears that programSubscribe often drops notifications.

Code snippet to reproduce: https://github.com/Arrowana/sol-rpc-race-example
It listens to the stake program for Authorized staker being the Keypair pubKey,
in a loop it sends a transaction to create a new stake account being the Authorized staker and withdrawer on a new account.

Example on mainnet-beta, it was notified "received" for a few created stake accounts but then dropped a few other:
```
received 1 BrXRZDA3acbRwYdve1T9sijhUmxNQ28GsZpx2aAio4EW
Sent 1
received 2 HFXDJQdJqovNxX5cA56A1XGPuiNTiUdyenstEfsAK6yF
Sent 2
Sent 3
received 4 2ztWtqFwoRVY2g6CjKUC63Yv7dtvMPxUu92CUPiQy3FG
Sent 4
Sent 5
received 6 2HvLYWYoXwYuzaeDKWXWr5ZUoC5vNXe6DcuEhjjwokPh
Sent 6
```

devnet: seems to notify of every single new program account without any problem
testnet: even more unreliable

Note: I tried removing the filter and it does not help, it still drops the expected transactions

#### Proposed Solution

TBD

jstarry:
This is probably not limited to program account subscriptions, but there are occasionally confirmed slots with program account changes which do not emit any notifications. Behavior observed on testnet using a modified version of your demo: 

Received notifications:
```
received 3 FLDzc4FxDAZdLT1fFjDnxb9iW7x7XmDspnwi4EQXVmbj: 85270787
received 11 8GMYKN9K6tGazLbS2EYSRRTJr1kiwc4EzTa7aP3Dux5D: 85270793
received 12 7g3aWPmpW5e97rjrXsfDthj1Y3sJgsyyRvK8jqG2Eok5: 85270793
received 13 5LwXst2gCMWAPbcWuwCzgEsWAMVHnV7QiGkBbWxBGFCB: 85270794
```

Confirmed transactions:
```
Sent 3: (85270787) https://explorer.solana.com/tx/bAHtLSXDFsbG9U4L3ragPT1RPY4vGGZnQPkM9Ybf3JN2W7qpWwCBQ9p5wh3wMP5cGmuyZg7dvRN3QnpwwTinH2w?cluster=testnet
[...]

PR #19092, opened 2021-08-05
Fixing missing pubsub notification for programSubscribe and logsSubscribe

lijunwangs:
#### Problem

https://github.com/solana-labs/solana/issues/18587: programSubscribe is missing notifications randomly. The issue is because of two reasons

1. Not all rooted slots get OptimisticallyConfirmed notifications
2. The OptimisticallyConfirmed notifications can be out of order for slots: slot A and B with A < B can see notification for B first before A.

#### Summary of Changes

Changed OptimisticallyConfirmedBankTracker to send notifications for parent banks if they have not been notified yet. We use a new variable last_notified_slot to track that.

Tests:
With my validator running against testnet, before the fix, it was failing 75% of time, with the fix, it is passing consistently. Using the program mentioned in https://github.com/solana-labs/solana/issues/18587.

Fixes # 18587

### DIFF
--- a/rpc/src/optimistically_confirmed_bank_tracker.rs
+++ b/rpc/src/optimistically_confirmed_bank_tracker.rs
@@ -157,14 +157,18 @@ impl OptimisticallyConfirmedBankTracker {
                         );
 
                         for parent in bank.parents().iter() {
-                            info!("notify_or_defer for parent {:?}", parent.slot());
-                            Self::notify_or_defer(
-                                subscriptions,
-                                bank_forks,
-                                parent,
-                                *last_notified_slot,
-                                &mut pending_optimistically_confirmed_banks,
-                            );
+                            if parent.slot() > *last_notified_slot {
+                                info!("notify_or_defer for parent {:?}", parent.slot());
+                                Self::notify_or_defer(
+                                    subscriptions,
+                                    bank_forks,
+                                    parent,
+                                    *last_notified_slot,
+                                    &mut pending_optimistically_confirmed_banks,
+                                );
+                            } else {
+                                break;
+                            }
                         }
                         *last_notified_slot = slot;
                     }
--- a/rpc/src/optimistically_confirmed_bank_tracker.rs
+++ b/rpc/src/optimistically_confirmed_bank_tracker.rs
@@ -143,9 +143,9 @@ impl OptimisticallyConfirmedBankTracker {
         mut last_notified_confirmed_slot: &mut Slot,
         mut pending_optimistically_confirmed_banks: &mut HashSet<Slot>,
     ) {
-        for parent in bank.clone().parents_inclusive().iter().rev() {
-            if parent.slot() > slot_threshold {
-                debug!("Calling notify_or_defer for parent {:?}", parent.slot());
+        for confirmed_bank in bank.clone().parents_inclusive().iter().rev() {
+            if confirmed_bank.slot() > slot_threshold {
+                debug!("Calling notify_or_defer for confirmed_bank {:?}", confirmed_bank.slot());
                 Self::notify_or_defer(
                     subscriptions,
                     bank_forks,
--- a/rpc/src/optimistically_confirmed_bank_tracker.rs
+++ b/rpc/src/optimistically_confirmed_bank_tracker.rs
@@ -66,6 +66,7 @@ impl OptimisticallyConfirmedBankTracker {
     ) -> Self {
         let exit_ = exit.clone();
         let mut pending_optimistically_confirmed_banks = HashSet::new();
+        let mut last_notified_confirmed_slot: Slot = 0;
         let mut highest_confirmed_slot: Slot = 0;
         let thread_hdl = Builder::new()
             .name("solana-optimistic-bank-tracker".to_string())
@@ -80,6 +81,7 @@ impl OptimisticallyConfirmedBankTracker {
                     &optimistically_confirmed_bank,
                     &subscriptions,
                     &mut pending_optimistically_confirmed_banks,
+                    &mut last_notified_confirmed_slot,
                     &mut highest_confirmed_slot,
                 ) {
                     break;
@@ -95,6 +97,7 @@ impl OptimisticallyConfirmedBankTracker {
         optimistically_confirmed_bank: &Arc<RwLock<OptimisticallyConfirmedBank>>,
         subscriptions: &Arc<RpcSubscriptions>,
         mut pending_optimistically_confirmed_banks: &mut HashSet<Slot>,
+        mut last_notified_confirmed_slot: &mut Slot,
         mut highest_confirmed_slot: &mut Slot,
     ) -> Result<(), RecvTimeoutError> {
         let notification = receiver.recv_timeout(Duration::from_secs(1))?;
@@ -104,6 +107,7 @@ impl OptimisticallyConfirmedBankTracker {
             optimistically_confirmed_bank,
             subscriptions,
             &mut pending_optimistically_confirmed_banks,
+            &mut last_notified_confirmed_slot,
             &mut highest_confirmed_slot,
         );
         Ok(())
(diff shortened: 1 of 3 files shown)


jstarry:
@lijunwangs can you please add a test that would fail without your fix? Also, it looks like you send notifications for a given slot before sending notifications for the parent slots, shouldn't the slots be notified in order?

lijunwangs:
@jstarry Not sure how do I add a failed test without my fix. My fix is already here in this PR. That would require a separate PR. I will look into the notification order issue.

carllin:
> Not sure how do I add a failed test without my fix. My fix is already here in this PR. That would require a separate PR. I will look into the notification order issue.

@lijunwangs I think Justin means a test that would fail if this fix wasn't made, but now passes with the fix in this PR. For example a test that sets up:

1) a solana-optimistic-bank-tracker 
2) an rpc service
3) a subscription for program id `X`,  and shows:

```
1 -> 2 (optimistic notify) -> 3 (create account `Y`) -> 4 (update account `Z` AND  optimistic notify)
```

That the RPC subscriber gets a notification for the last created account in slot `4` before the update in slot `3`

lijunwangs:
@carllin Is that unit test or integration tests. I don't think we check that subscription is received in unit tests. I did not find integration tests for the RPC module. Where do we keep them?

lijunwangs:
I saw something similar to what we are looking for in rpc_subscriptions.rs -- will look into it next. Thanks @carllin  and @jstarry for your suggestions.

rpc/src/optimistically_confirmed_bank_tracker.rs
```diff
@@ -101,31 +104,74 @@ impl OptimisticallyConfirmedBankTracker {
...
+                        }
+                        Self::notify_or_defer(
+                            subscriptions,
+                            bank_forks,
+                            bank,
+                            *last_notified_slot,
+                            &mut pending_optimistically_confirmed_banks,
+                        );
+
+                        *last_notified_slot = slot;
```
jstarry, line 174: How about whenever `subscriptions.notify_gossip_subscribers` is called, `last_notified_slot` is checked and updated accordingly? This would mean that `last_notified_slot` is updated below when pending opt confirmed banks are notified rather when they are put in the queue. Functionally the same result but I think it would make the code more resilient and readable

rpc/src/rpc_subscriptions.rs
```diff
@@ -1619,6 +1622,172 @@ pub(crate) mod tests {
             .contains_key(&stake::program::id()));
     }
 
+    #[test]
+    #[serial]
+    fn test_check_program_subscribe_for_missing_optimistically_confirmed_slot() {
```
jstarry, line 1627: Great test, thanks!

rpc/src/optimistically_confirmed_bank_tracker.rs
lijunwangs: I think it would be different. We actually want to update it either the slot is notified via `subscriptions.notify_gossip_subscribers` or put in the pending queue to prevent it from being reprocessed.

lijunwangs: But I agree the name last_notified_slot is a little misleading as it might not be really notified. It is just processed.

jstarry: Changing to `last_observed_slot` would be fine too

lijunwangs: I think we do not want to unnecessarily put them into the queue. I will rename it to highest_confirmed_slot -- it is what it really is.

jstarry (approved):

rpc/src/optimistically_confirmed_bank_tracker.rs
```diff
@@ -101,31 +104,74 @@ impl OptimisticallyConfirmedBankTracker {
...
+                            bank_forks,
+                            bank,
+                            *highest_confirmed_slot,
+                            &mut pending_optimistically_confirmed_banks,
+                        );
+
+                        *highest_confirmed_slot = slot;
                     }
                     drop(w_optimistically_confirmed_bank);
                 } else if slot > bank_forks.read().unwrap().root_bank().slot() {
```
CriesofCarrots, line 177: Hmm, I think this solution doesn't handle this case properly.
Imagine BankForks contains slots 0-2, root 0. The first optimistic confirmation we receive is for slot 3, but it's not actually in BankForks yet. We hit this case and add slot 3 to the pending list. However, slots 1-2 are not in the pending list, so when slot 3 is frozen, only slot 3 is notified.

I think that we need to use `highest_confirmed_slot` to notify parents in `BankNotification::Frozen(bank)` case below, and only update it on `notify` (not `defer`).
Then, rather than adding all the parents to the pending list in `defer` above, just add `bank.slot()`.

CriesofCarrots: You could likely demonstrate this failing in your test by not inserting bank3 into bank_forks until somewhere around here: https://github.com/solana-labs/solana/pull/19092/files#diff-836b3beddc73d5a3dbe035ef96380aab3ac7caadd149ab405b1c2534dd90b2f3R1775

lijunwangs: Good point, in the tests I have done against testnet. I was not able to see such conditions. @carllin Your thought? Can we have a OptimisticallyConfirmed notification without having the slot in the BankForks?

jstarry: Great catch @CriesofCarrots! @lijunwangs yes, as @CriesofCarrots has pointed out, it is possible to have a notification without having the slot in bank forks if the node is processing slightly behind the rest of the cluster

lijunwangs: Okay. I will look into how to handle this case.

carllin: >  Can we have a OptimisticallyConfirmed notification without having the slot in the BankForks

@lijunwangs yeah, this can happen if we get the notification before we start replaying hte slot

lijunwangs: Thanks for catching and confirming this use case @CriesofCarrots, @jstarry and @carllin -- it will be handled as well.

lijunwangs: This is addressed in the latest update

rpc/src/optimistically_confirmed_bank_tracker.rs
```diff
@@ -101,31 +107,95 @@ impl OptimisticallyConfirmedBankTracker {
...
+    fn notify_or_defer_confirmed_banks(
+        subscriptions: &Arc<RpcSubscriptions>,
+        bank_forks: &Arc<RwLock<BankForks>>,
+        bank: &Arc<Bank>,
+        slot_threshold: Slot,
+        mut last_notified_confirmed_slot: &mut Slot,
+        mut pending_optimistically_confirmed_banks: &mut HashSet<Slot>,
+    ) {
+        for parent in bank.parents().iter().rev() {
+            if parent.slot() > slot_threshold {
```
CriesofCarrots, line 147:
```suggestion
        for parent in bank.parents().iter().take_while(|slot| slot > slot_threshold).rev() {
```
Just noticed, I think we can save a few iterations?

CriesofCarrots: Also, what do you think about using `Bank::parents_inclusive()` to avoid the dupe of `notify_or_defer` below?

lijunwangs: Thanks for the suggestion on take_while. But I am not convinced  take_while actually saves any real cycles -- looking at is implementation using TakeWhile --  it uses the wrapped iterator and apply the closure in the same way a normal iterator. Will look into using parents_inclusive.

CriesofCarrots: The difference is that `take_while` stops iterating when it reaches the first element that doesn't satisfy the predicate, whereas your code will continue on to check every parent in the list.
Here's a code snippet that demonstrates perf difference over a larger set of elements:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=73168b2c8b67710d0015cb8869a08bee

lijunwangs: Good point @CriesofCarrots -- will change that.

CriesofCarrots: Sadly, `take_while` doesn't support double-ended iterator; as-such, current code is more performative for the likely parents.len()

CriesofCarrots (approved):
lgtm! One nit, take it or leave it.

PR #19093, opened 2021-08-06
get rid of WriteAccountMapEntry instance

jeffwashington:
#### Problem
Part of refactoring AccountsIndex to be backed by disk. Keeping account data open longer should be discouraged. All users of WriteAccountMapEntry were only modifying the slot list.
Later, this can be refactored to move into in_mem_accounts_index to not require the arc clone either. Clean and shrink are both slower with disk based indexes. This is an attempt to improve those code paths.
#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/accounts_index.rs
+++ b/runtime/src/accounts_index.rs
@@ -284,37 +284,6 @@ impl<T: IndexValue> PreAllocatedAccountMapEntry<T> {
     }
 }
 
-#[self_referencing]
-pub struct WriteAccountMapEntry<T: IndexValue> {
-    owned_entry: AccountMapEntry<T>,
-    #[borrows(owned_entry)]
-    #[covariant]
-    slot_list_guard: RwLockWriteGuard<'this, SlotList<T>>,
-}
-
-impl<T: IndexValue> WriteAccountMapEntry<T> {
-    pub fn from_account_map_entry(account_map_entry: AccountMapEntry<T>) -> Self {
-        WriteAccountMapEntryBuilder {
-            owned_entry: account_map_entry,
-            slot_list_guard_builder: |lock| lock.slot_list.write().unwrap(),
-        }
-        .build()
-    }
-
-    pub fn slot_list(&self) -> &SlotList<T> {
-        &*self.borrow_slot_list_guard()
-    }
-
-    pub fn slot_list_mut<RT>(
-        &mut self,
-        user: impl for<'this> FnOnce(&mut RwLockWriteGuard<'this, SlotList<T>>) -> RT,
-    ) -> RT {
-        let result = self.with_slot_list_guard_mut(user);
-        self.borrow_owned_entry().set_dirty(true);
-        result
-    }
-}
-
 #[derive(Debug, Default, AbiExample, Clone)]
 pub struct RollingBitField {
     max_width: u64,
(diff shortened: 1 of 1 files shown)


runtime/src/accounts_index.rs
```diff
@@ -1130,36 +1090,16 @@ impl<T: IsCached> AccountsIndex<T> {
...
-            }
-        }
+        pubkey: &Pubkey,
+        user: impl for<'a> FnOnce(&mut RwLockWriteGuard<'a, SlotList<T>>) -> RT,
+    ) -> Option<RT> {
+        let read_lock = self.account_maps[self.bin_calculator.bin_from_pubkey(pubkey)]
+            .read()
+            .unwrap();
+        let get = read_lock.get(pubkey);
+        get.map(|entry| user(&mut entry.slot_list.write().unwrap()))
```
carllin, line 1102: Right now, this seems similar to the the existing `get_account_write_entry()`: https://github.com/solana-labs/solana/blob/v1.7/runtime/src/accounts_index.rs#L1009-L1014, but with more scoping because the `read_lock` is released as soon as this function returns. However, looking at the places we call `get_account_write_entry`, we don't currently hold the lock longer than we should, so this seems this functionally equivalent. Is this to avoid the possibility of holding the read lock longer than we should? Is there some other benefit I'm missing, or is this in preparation for another change?

jeffwashington: Backing up, this pr is expressing an idea in a rough code implementation. I'm trying to understand the api we want to support: what we need, what exists today, what clients do with it, difficulties/constraints that arise due to the api, etc.
So, if I pull on this thread (change this api), what happens?
In this case, if I get rid of `WriteAccountMapEntry` and replace it with this call:
`slot_list_mut(pubkey, fn)` the callers are still satisfied, and then here are some possible advantages I can see:

1. the api gets simplified
2. we could collapse/rework the lock strategy (to remove arc/rwlock) and no caller would know or care
3. we could make slot_list_mut use a write lock at the bin level and no caller would care.
4. we eliminate the possibility that a caller could in the future get a write lock and keep it indefinitely accidentally or on purpose. With the new api, they don't even see a lock at all now.
5. we simplify the callers
6. the implementation gets simplified - there is 1 less self-referencing struct
7. simpler implementations are often followed by other simplifications and optimizations once we have boiled down something to its essential operations.

The biggest downsides I see to a refactor like this:
1. if we want to add more behavior (such as addref/release/delete) to `WriteAccountMapEntry`, then we may want to resurrect this struct.
2. We have to do some code moving since the existing struct with no members is likely no longer the right organization of the static methods that are in it.
3. Someone who knows the code today (like you) loses some familiarity with the current api and implementation of AccountsIndex.

jeffwashington: An example: If I have to support the current model of WriteAccountMapEntry, then in some cases (such as bucket maps or eliminating the rwlock per entry), we may require a copy of the slot list. This is a rough assertion - not a specific one. Copies of slot lists are something I continually run into as I imagine and explore different paths looking for the ideal set of tradeoffs for 10B accounts.

carllin: ok, thanks for the context, I don't have any objections to this change as it seems functionally equivalent.

brooksprumo (approved):
lgtm

PR #19098, opened 2021-08-06
remove AccountsDb::new() from public api

jeffwashington:
#### Problem
#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -1407,7 +1407,8 @@ impl AccountsDb {
         }
     }
 
-    pub fn new(paths: Vec<PathBuf>, cluster_type: &ClusterType) -> Self {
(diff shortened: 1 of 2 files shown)


runtime/src/accounts_db.rs
```diff
@@ -6186,6 +6182,10 @@ impl AccountsDb {
 
 #[cfg(test)]
 impl AccountsDb {
+    pub fn new(paths: Vec<PathBuf>, cluster_type: &ClusterType) -> Self {
+        Self::new_for_tests(paths, cluster_type)
+    }
```
brooksprumo, line 6187: Would it make sense to call `Self::new_with_config()` and then just delete the `new_for_tests()` function?

brooksprumo: Spoke to soon. I see the reason in `runtime/tests/accounts.rs` below. Sorry!

brooksprumo (approved):

PR #19110, opened 2021-08-07
Fix deserialization of RPC errors in HttpSender

### LINKED ISSUE #15576: RpcRequestError("Failed to deserialize RPC error response
#### Problem
When using the CLI to check a vote-account on mainnet, I receive an RPC parsing error:

```
$ solana --url [MAINNET] vote-account 5m2h5u5gaCb8d4iG2K1oKGrxDu99UZ1GJbFMJNFbCPmV
Failed to fetch epoch rewards: ClientError { request: Some(GetConfirmedBlock), kind: RpcError(RpcRequestError("Failed to deserialize RPC error response: {\"code\":-32004,\"message\":\"Block not available for slot 66960001\"} [missing field `data`]")) }
```

This validator uses the same pubkey & vote account on both mainnet & testnet. I do not receive that error on testnet, just mainnet.

#### Proposed Solution
I identify the cause of the parsing error. It looks like we are missing the 'data' element.

brson:
#### Problem

HttpSender, and in turn RpcClient, fails to deserialize some, if not many, RPC errors, resulting in generic "failed to deserialize RPC error" errors.

The bug comes from expecting every error to contain a `data` field, but that field is optional.

#### Summary of Changes

Remove the data field from the `RpcErrorObject` type. This is an internal intermediate type, and its data field is not even used. The presence of the `data` field here only serves to create this deserialization error.

Fixes #15576

### DIFF
--- a/core/tests/rpc.rs
+++ b/core/tests/rpc.rs
@@ -423,7 +423,6 @@ fn test_tpu_send_transaction() {
     }
 }
 
-// #15576 - correctly deserialize RPC errors
 #[test]
 fn deserialize_rpc_error() -> ClientResult<()> {
     solana_logger::setup();
--- a/core/tests/rpc.rs
+++ b/core/tests/rpc.rs
@@ -6,8 +6,10 @@ use reqwest::{self, header::CONTENT_TYPE};
 use serde_json::{json, Value};
 use solana_account_decoder::UiAccount;
 use solana_client::{
+    client_error::{ClientErrorKind, Result as ClientResult},
     rpc_client::RpcClient,
     rpc_config::{RpcAccountInfoConfig, RpcSignatureSubscribeConfig},
+    rpc_request::RpcError,
     rpc_response::{Response, RpcSignatureResult, SlotUpdate},
     tpu_client::{TpuClient, TpuClientConfig},
 };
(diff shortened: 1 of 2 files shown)


brson:
This is an API breaking change since users will see more correct error types now.

brson:
Here is the source of the error:

https://github.com/solana-labs/solana/pull/19110/files#diff-3f885c47815e09d9f2db1c3279c664170d9123687ea344cdc9baa75fbb718dd3R157

brson:
Here is the code that causes the error, since it isn't obvious from the diff:

https://github.com/solana-labs/solana/blob/813628d833b518c9a8ee65afa109a342af376386/client/src/http_sender.rs#L126-L161

core/tests/rpc.rs
```diff
@@ -420,3 +422,35 @@ fn test_tpu_send_transaction() {
         }
     }
 }
+
+// #15576 - correctly deserialize RPC errors
```
mvines, line 426:
```suggestion
```

mvines:
Thanks, looks great.  I just removed the comment at the top of the new test

mvines:
Cargo audit failure is unrelated

PR #19112, opened 2021-08-07
Add EpochSlots frozen state transition

carllin:
#### Problem
AncestorHashesService sends ReplayStage updates about cluster frozen slots: https://github.com/solana-labs/solana/blob/00890957ee608e41b25f6c79b4cd6ad7b52b9373/core/src/ancestor_hashes_service.rs#L323-L325, but ReplayStage doesn't process them

#### Summary of Changes
1. process_epoch_slots_frozen_dead_slots() processes updates from `AncestorHashesService`
2.  add state transition in `cluster_slot_state_verifier` to handle updates from `AncestorHashesService ` and potentially dump/repair dead slots
3. 
Fixes #

### DIFF
--- a/core/src/cluster_slot_state_verifier.rs
+++ b/core/src/cluster_slot_state_verifier.rs
@@ -292,9 +292,6 @@ pub enum ResultingStateChange {
     // Hash of our current frozen version of the slot
     DuplicateConfirmedSlotMatchesCluster(Hash),
     SendAncestorHashesReplayUpdate(AncestorHashesReplayUpdate),
-    // Hash returned from EpochSlots does not match the duplicate
-    // confirmed hash detected from replay or gossip.
-    EpochSlotsMismatchedHash,
 }
 
 impl SlotStateUpdate {
@@ -695,7 +692,6 @@ fn apply_state_changes(
     duplicate_slots_to_repair: &mut DuplicateSlotsToRepair,
     blockstore: &Blockstore,
     ancestor_hashes_replay_update_sender: &AncestorHashesReplayUpdateSender,
-    epoch_slots_frozen_slots: &mut EpochSlotsFrozenSlots,
     state_changes: Vec<ResultingStateChange>,
 ) {
     // Handle cases where the bank is frozen, but not duplicate confirmed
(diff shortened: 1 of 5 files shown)


core/src/cluster_slot_state_verifier.rs
```diff
@@ -5,12 +5,21 @@ use crate::{
...
+use std::collections::{BTreeMap, BTreeSet, HashMap};
 
 pub(crate) type DuplicateSlotsTracker = BTreeSet<Slot>;
-pub(crate) type DuplicateSlotsToRepair = HashSet<(Slot, Hash)>;
+pub(crate) type DuplicateSlotsToRepair = HashMap<Slot, Hash>;
+pub(crate) type EpochSlotsFrozenSlots = BTreeMap<Slot, Hash>;
 pub(crate) type GossipDuplicateConfirmedSlots = BTreeMap<Slot, Hash>;
 
+#[derive(PartialEq, Clone, Debug)]
+pub enum ClusterConfirmedHash {
```
carllin, line 16: Introduce `ClusterConfirmedHash` to track what level of certainty we have about a slot

core/src/cluster_slot_state_verifier.rs
```diff
@@ -287,28 +357,77 @@ fn check_duplicate_confirmed_hash_against_frozen_hash(
...
 
+fn check_epoch_slots_hash_against_frozen_hash(
+    state_changes: &mut Vec<ResultingStateChange>,
+    slot: Slot,
+    epoch_slots_frozen_hash: Hash,
+    bank_frozen_hash: Hash,
+    is_dead: bool,
+) {
+    if epoch_slots_frozen_hash != bank_frozen_hash {
+        if is_dead {
```
AshwinSekar, line 368: In this case we could still have the correct version right? If we had an incorrect ancestor, and this block was marked dead it is not necessary that this block is duplicate right?

carllin: yeah, its not necessarily this block that is duplicate, that's correct, this block just gets dumped in the crossfire :'(.

The hope is we eventually make an ancestor request for the earliest dead slot. After we get the ancestors for that earliest dead slot, that should tell us which of the frozen/not dead ancestors is bad.

AshwinSekar:
Looks good to me

AshwinSekar (approved):

PR #19114, opened 2021-08-07
Move SnapshotArchiveInfo and friends into its own module

brooksprumo:
`SnapshotArchiveInfo` has grown quite a bit, plus I will be making more additions to support Incremental Snapshots in the background services. Moving `SnapshotArchiveInfo` and friends into its own module makes sense at this point.


lijunwangs (approved):
Good refactoring. Looks good.

Issue #19079, opened 2021-08-05
(wip) packet drop investigation


ryoqun:
#### Problem

discord discussions for reference:
https://discord.com/channels/428295358100013066/478692221441409024/870707709639548989

ryoqun:
v1.7:

![image](https://user-images.githubusercontent.com/117807/128316896-c93c7c57-d0e7-4d3b-8e60-f727b20be8b9.png)

master:

```
$  /proc/$(pgrep -f solana-validator | awk '{print $1}')/exe --version
solana-validator 1.8.0 (src:dfe99efa; feat:1693936682)
$ git show dfe99efa -q
commit dfe99efa7cfaa14390844e7c6cb6594eec3d1afc (origin/master, origin/HEAD)
Author: Jeff Washington (jwash) <75863576+jeffwashington@users.noreply.github.com>
Date:   Wed Aug 4 21:58:53 2021 -0500

    introduce AccountsIndex::default_for_tests() (#19067)
```

![image](https://user-images.githubusercontent.com/117807/128317220-9c6be1a1-7cf7-43ae-adfe-e8c495fc9390.png)

ryoqun:
status: after another round of test, my packet drop monitoring mechanism had a bug leading to false positive of packet drops.... specifically, `conntrack -Z` don't work as intended..

so, it turns out nodes located inside gce don't exhibit packet drops. hope the script is finally got robust; will soon field-test on testnet using it.

ryoqun:
well, there is really a mysterious phenomenon on testnet.

Clearly, testnet's skip rate corresponds to mysterious broadcast/retransmit incoming bandwidth reduction.

here's skip rate:

![image](https://user-images.githubusercontent.com/117807/129360607-8ff88e86-bda8-4296-a36e-4abe3953c81d.png)

and this is incoming packets (violet (local listening port 8002) is `tvu` and skyblue (local listening port 8003) is `tvu_forward`), collected manually from my testnet staked node.

it seems that there was 3 incidents where testnet node experienced quite sudden incoming packet reduction for some reason. and it causes cluster-wide repairs and ultimately bad skip rate.

![image](https://user-images.githubusercontent.com/117807/129360730-8c2a25db-d39f-499c-9b52-5420120b3229.png)

behzadnouri:
> it seems that there was 3 incidents where testnet node experienced quite sudden incoming packet reduction for some reason.

are those near epoch boundaries? or new releases?

[...]

PR #19115, opened 2021-08-07
batch retransmit_to in retransmit_stage

jbiseda:
#### Problem
Computation interleaved with packet transmission makes measuring computation vs transmission time more difficult.

#### Summary of Changes
Consolidate packet retransmission in retransmit_stage.

./gce.sh create -p testnet-dev-jbiseda -n 10 -c 1
./net.sh start -c "bench-tps=1=--tx_count 5000 --thread-batch-sleep-ms 250"

changes left, master right:
Screen Shot retransmit_stage_batch-master


behzadnouri (commented):
I am not sure if you saw the ongoing discord discussions in `#network-protocols` channel regarding packets drops: https://github.com/solana-labs/solana/issues/19079
https://discord.com/channels/428295358100013066/478692221441409024/870707709639548989

It might be the case that excessive batching is actually causing higher bursts and overall more packet drops; and we might need to in fact go in the other direction and implement some rate-limiting, back-pressure, etc.

So it will be very helpful if with such changes we can get some before/after measurements of the impact and ensure that it does in fact improve things:
* There are already some `RetransmitStats` metrics in this code:
   https://github.com/solana-labs/solana/blob/c7b986220/core/src/retransmit_stage.rs#L64-L79
   https://github.com/solana-labs/solana/blob/c7b986220/core/src/retransmit_stage.rs#L135-L188
   We can start a validator on testnet running the current master code, leave it run for a while, then restart it with this change included and compare metrics; similar to: https://github.com/solana-labs/solana/pull/14781#issuecomment-766391155 
   * This would not lead any metrics on packet drops though.
* For packet drops, one (imperfect) proxy are the repair-metrics. We can start a GCE cluster of maybe 10 nodes or so, with and without this code and compare repair metrics similar to: https://github.com/solana-labs/solana/pull/14910#issuecomment-805107406

CriesofCarrots:
I'll defer to Behzad on the review (I'm in the `git blame` here because I moved a bunch of stuff to `solana-gossip`; haven't worked in this area much otherwise.)

core/src/retransmit_stage.rs
```diff
@@ -294,31 +294,25 @@ fn retransmit(
...
-        );
+            ClusterInfo::get_retransmit_to_dests(&neighbors[1..], true, socket_addr_space)
+        } else {
+            vec![]
+        };
+        let child_addrs: Vec<_> =
+            ClusterInfo::get_retransmit_to_dests(&children, !anchor_node, socket_addr_space);
+        dest_addrs.extend(child_addrs);
+        compute_turbine_peers.stop();
+        compute_turbine_peers_total += compute_turbine_peers.as_us();
```
behzadnouri, line 312: why change this metric?

behzadnouri:
Thank you for adding the metrics plots. Can you please describe the cluster set-up you used to obtain the metrics? and, what is your conclusion as the impact of change?

Also, it would be great if:
* use the "Problem/Summary of Changes" template for pull-request description: https://github.com/solana-labs/solana/blob/master/CONTRIBUTING.md#the-pr-problem-statement
* include before/after metrics in the same plot similar to:
  https://github.com/solana-labs/solana/pull/14910#issuecomment-805107406
  right now even the axes are not the same; which makes the comparison more difficult. e.g. 0-100 vs 0-600.

jbiseda:
Closing per test results

PR #19130, opened 2021-08-09
allows private addresses if not public network

behzadnouri:
#### Problem
`./net.sh` needs `--allow-private-addr` if `gce.sh create` was invoked without `-P`.

#### Summary of Changes
When loading config file, if not a public network add `--allow-private-addr`.


behzadnouri:
@mvines any objection to merging this?

mvines (approved):

PR #19135, opened 2021-08-09
Use private IP addresses for the partition gce test

AshwinSekar:
#### Problem
The recent gossip change requires an additional flag `--allow-private-addr` in order to discover private IP addresses.
We modify the gce cluster tests to pass this additional flag and set IP addresses to private. 

#### Summary of Changes
Use private IP addresses for this test.

Fixes #


behzadnouri:
As mentioned in the chat:
`USE_PUBLIC_IP_ADDRESSES = false` is the desired state: https://github.com/solana-labs/solana/pull/18728#issuecomment-884290209
Instead we need to add `ALLOW_PRIVATE_ADDR: "true"` similar to https://github.com/solana-labs/solana/pull/18876

behzadnouri (approved):
LGTM
Please update the pull-request description.

PR #19152, opened 2021-08-10
Banking_stage drops TooOld vote transactions

### LINKED ISSUE #17877: Cost model to handle redundant votes
#### Problem
Found in testnet, periodically, blocks of vote transactions have lower success rate (< 90%), largely due to many of the votes in block are "too old".

cost_model as it is right now wouldn't help in this case, because votes usually being evenly distributed across accounts, and vote itself is rather inexpensive, therefore falls below the limits that would trigger cost_model function.  

#### Proposed Solutions
possible approaches to handle redundant votes are:  

1. Leader to identify votes that are "too old", and drop these votes (eg., not add to entries/blocks for broadcasting). Leader can potentially borrow "vote instruction" logic in determine redundant votes. However, the performance impact must be evaluated, as well as the potential downstream impact of dropping vote transactions; 

2. or cost_model can pack potentially redundant vote transactions into different blocks, therefore improve each block's success rate. cost_model can do so by making the cost for a *potentially* redundant votes in single block more expensive. This approach does not require to accurately identify "too old" vote transaction since all votes will eventually be sent, hence reduce the extra work to put on to block producer.

taozhu-chicago:
#### Problem
A lot of vote transactions are transmitted and replayed only to fail at instruction_process stage. Leader could not include TooOld Vote transactions into block to improve efficiency.

#### Summary of Changes
1. add function to drop redundant votes at banking_stage,
2. the logic of defining redundancy yet to be refined

Fixes #17877


carllin:
>  Leader could not include TooOld Vote transactions into block to improve efficiency.

Yeah, iirc on our last call the actions we thought to take were:

1) Order the votes in the queue beforehand to try and maximize in order processing of the votes for each validator
2) Drop votes that are less than the latest vote in the vote state of the validator. This means introspecting the vote state in the banking stage *before*  instruction execution, as you suggested 😃

taozhu-chicago:
> 2. Drop votes that are less than the latest vote in the vote state of the validator. This means introspecting the vote state in the banking stage _before_  instruction execution, as you suggested 😃

Yea, I am exploring if there are light-weight options to detect "old" votes without having to invoke validator's VoteState entirely. Maybe banking_stage keeps a HashMap<validator_key, its_latest_voted_slot> , any votes from a validator has last vote earlier than `its_latest_voted_slot` will be dropped. Wondering if such approach worth the effort.

taozhu-chicago:
I ran today PR build in `testnet-dev`, which utilizes `VoteState` in banking_stage to check vote transactions against its validator's latest_voted_slot in detecting redundant votes; comparing to master build. I expected the number of transactions Banking_stage adds to block to be reduced somewhat, and hopefully also resulted in reduced timing on converting transactions (due to reduced number of TXs). The grafana seems to agree, just not in super convincing way. 

image

image

@carllin @sakridge What do you feel about this approach and result?

carllin:
> The grafana seems to agree - the left part of graph that drops vote is bit more sparse that master build, just not in super convincing way.

@taozhu-chicago, what is the metric the first graph is measuring?

core/src/vote_redundancy_checker.rs
```diff
@@ -0,0 +1,146 @@
...
+        }
+    }
+
+    pub fn check(
+        &mut self,
+        transaction: &Transaction,
+        bank: &Arc<Bank>,
+    ) -> Result<(), &'static str> {
+        if let Some((vote_account_pubkey, vote, _hash)) =
+            vote_transaction::parse_vote_transaction(transaction)
```
carllin, line 25: This should probably be done outside of holding the lock around the `VoteRedundancyChecker` and the results of the parsing passed in here

core/src/vote_redundancy_checker.rs
```diff
@@ -0,0 +1,146 @@
...
+                            "Vote on slot 0, skip checking redundancy for vote {:?}",
+                            vote
+                        );
+                        return Ok(());
+                    }
+                    slot
+                }
+            };
+
+            // TODO TAO - Should ignore VoteSwitch too?
```
carllin, line 48: yeah right now voteswitch is treated like any other vote

core/src/banking_stage.rs
```diff
@@ -1060,6 +1078,11 @@ impl BankingStage {
                 let tx: Transaction = limited_deserialize(&p.data[0..p.meta.size]).ok()?;
                 tx.verify_precompiles(libsecp256k1_0_5_upgrade_enabled)
                     .ok()?;
+                vote_redundancy_checker
+                    .write()
+                    .unwrap()
+                    .check(&tx, bank)
+                    .ok()?;
```
carllin, line 1085: Hmm I think there's a race here. Even if the `vote_redundancy_checker` see slots `1->2->3->4` in order, the execution could still be out of order because ultimately what matters is the order in which these banking threads hit the `process_and_record_transactions_locked()` and grab the locks for the accounts

core/src/vote_redundancy_checker.rs
```diff
@@ -0,0 +1,146 @@
+use log::*;
+use solana_runtime::bank::Bank;
+use solana_sdk::{clock::Slot, pubkey::Pubkey, transaction::Transaction};
+use solana_vote_program::vote_transaction;
+use std::{collections::HashMap, sync::Arc};
+
+#[derive(Default, Debug)]
+pub struct VoteRedundancyChecker {
+    latest_voted_slots: HashMap<Pubkey, Slot>,
```
carllin, line 9: If you make this a `DashMap<Pubkey, AtomicU64>`, you can get rid of the lock around this `VoteRedundancyChecker`, which should be a lot more performant.

taozhu-chicago:
> > The grafana seems to agree - the left part of graph that drops vote is bit more sparse that master build, just not in super convincing way.
> 
> @taozhu-chicago, what is the metric the first graph is measuring? Do we have a metric for measuring `VoteTooOld` failures to see the efficacy?

Sorry, me sloppy here. The first metric is the count of verified transactions (https://github.com/solana-labs/solana/pull/19152/files#diff-bb8464c2f02c8863fb60e6e29f023f3e6da93a93fbb9199a70a44c9c1b41c7f3L1069-L1072), if redundant votes are dropped, this counter should drop too. 

The second n=metric is the time takes to convert packets (https://github.com/solana-labs/solana/pull/19152/files#diff-bb8464c2f02c8863fb60e6e29f023f3e6da93a93fbb9199a70a44c9c1b41c7f3L1163), with less transactions to convert, this metric should go down too.

core/src/banking_stage.rs
taozhu-chicago: on second thought, if banking_stage tracks last voted slot, and assume all transaction it let through will eventually landed in blockstore, it should be able to drop votes that voted for older slots (as implemented as solution 2 at https://github.com/solana-labs/solana/pull/19152/files#diff-bb8464c2f02c8863fb60e6e29f023f3e6da93a93fbb9199a70a44c9c1b41c7f3R1103-R1118 ). Do you think so?

runtime/src/bank.rs
```diff
@@ -3085,6 +3087,104 @@ impl Bank {
...
+                );
+                return false;
+            }
+            Some(slot) => {
+                if 0 == *slot {
+                    debug!(
+                        "Vote on slot 0, skip checking redundancy for vote {:?}",
+                        vote
+                    );
+                    return false;
```
carllin, line 3110: Couldn't a separate vote for 0 have already landed? This special case seems unnecessary.

carllin, line 3110: Couldn't a later vote have already landed, why do we make an exception for slot 0?

core/src/banking_stage.rs
carllin: Hmm, this still seems to run into the same issue. I can have two threads, one gets a vote for slot 1, one for slot 2 from the same validator. They insert the slots in order into `latest_voted_slots`, but the thread with slot 2 processes it's vote transaction first. Then the thread adding `slot 1` will incur the `TooOld` error when it tries to be process the transaction.

runtime/src/bank.rs
```diff
@@ -3085,6 +3087,104 @@ impl Bank {
         }
     }
 
+    // TODO - solution 1, check vote redundancy by VoteState within bank,
+    //     this solution does not alight with bankless leader goal;
+    //     mainly using it to compare with solution #2 to confirm
+    //     behavior and performance.
+    fn is_redundant_vote_by_vote_state(&self, vote_account_pubkey: &Pubkey, vote: &Vote) -> bool {
```
carllin, line 3094: Would be good to add a test for this function

runtime/src/bank.rs
```diff
@@ -3340,6 +3440,7 @@ impl Bank {
     ) {
         let sanitized_txs = batch.sanitized_transactions();
         debug!("processing transactions: {}", sanitized_txs.len());
+        // TODO - this is the 1st metric
```
carllin, line 3443: not too clear what this TODO is addressing

runtime/src/bank.rs
```diff
@@ -3085,6 +3087,104 @@ impl Bank {
...
+    }
+
+    fn check_redundant_votes(
+        &self,
+        sanitized_txs: &[SanitizedTransaction],
+        lock_results: &[Result<()>],
+    ) -> Vec<Result<()>> {
+        sanitized_txs
+            .as_transactions_iter()
+            .zip(lock_results.to_vec())
```
carllin, line 3159: I think you could pass in this zip as an iterator for more clarity, i.e. something like `txs_and_locks: impl Iterator<Item=(&SanitizedTransaction, &Result<()>)>`

runtime/src/bank.rs
taozhu-chicago: So when the validator starts up, I noticed the first votes are for slot 0, at that moment `last_voted_slot` is `0` too. Hence the special path for slot 0 to avoid it being dropped as redundant.

core/src/banking_stage.rs
taozhu-chicago: right! the "solution 1", which banking_stage checks vote without involving `bank`, inherently has chances to pass old votes to bank for processing, though the possibilities are largely reduced (by filtering out out-of-order votes received by banking_stage). The benefit of this approach, as mentioned in Summary of Change above, is it doesn't require `bank`, which alights with bank-less leader goal. 

The other solution implemented in this PR (https://github.com/solana-labs/solana/pull/19152/files#diff-ed47b4a0198313377e091bb3957bbbc63d937805426d1b2b6de39d0a50d32a0cR3461) checks vote in bank without race condition, but my concern is if this approach will be invalidated when implementing bank-less leader (I haven't thought it through, just a general concern) 

At the end, we only need one of these two implementations in this PR. The debate is "which one" :)

runtime/src/bank.rs
taozhu-chicago: Yes! Once we settled on final solution, Will add test for it if this is the chosen one

runtime/src/bank.rs
taozhu-chicago: Ah, forgot to remove this comment. Added to "bookmark" the metric count I need for grafana. No real use.

core/src/banking_stage.rs
carllin: A few things:

1) Doesn't the bankless leader design also require access to a bank in order to figure out whether an account has sufficient funds to pay the transaction fee?

2) It seems like the stateless `VoteRedundancyChecker` could theoretically also be used in solution 2 *after* we grab the locks? 

3) If we go with the `VoteRedundancyChecker`,  I also think we may need a separate `VoteRedundancyChecker` per bank because of forking. For example if you had

```
      1
  /       \
2          3
```

A vote for slot 1 may have landed in the bank for slot 2, but not for the bank in slot 3.

taozhu-chicago: 1. Yeah, will need bank to check account balance, would be great is just simply in-n-out for that purpose only. But I am not complete sure, not opposing to do check after account lock at all. 
2. Yep, the Bank solution currently does the same thing `VoteRedundantChecker` does. 
3. This is another point for using bank-based solution, so current implement (https://github.com/solana-labs/solana/pull/19152/files#diff-ed47b4a0198313377e091bb3957bbbc63d937805426d1b2b6de39d0a50d32a0cR3117) covers the forking scenario, i think

Answering though these questions, I started to see the benefit of bank-based solution, I'll remove the banking_stage based code, to make the PR easier for review.

taozhu-chicago:
settled on checking vote using current bank's VoteState, after accounts were locked, and right before TXs were loaded and executed. Still working on manipulating test bank instances to test redundancy check function. But the implementation is ready for review. 

ran testnet cluster, the metrics shows almost half of votes were identified. as redundant and dropped (Orange are vote TX count, and Blue is dropped vote TX count):

runtime/src/bank.rs
taozhu-chicago: thanks @carllin for helping setup test.

taozhu-chicago:
rebased master

taozhu-chicago:
The PR is consistently failing on CI local-cluster test `test_fork_choice_refresh_old_votes`, it panicked at [this call](https://github.com/solana-labs/solana/blob/master/local-cluster/tests/local_cluster.rs#L387-L394) during CI. (it doesn't fail if run the test individually on local machine, but @AshwinSekar was able to reproduce it.) It seems possible the block with TXs in it was not voted on properly, due to changes in this PR. 

I traced the test by log, notice there are mainly two type vote tx are being dropped: 
1. the duplicated vote:
- processed vote `Vote { slots: [0, 1, 2, 3, 4], hash: 2QBgKPpebAQqRoL6BiL7UBR6wExB
u3ydsVRUjsZij8x7, timestamp: Some(1630448707) } is good for slot 32 vote account BvmUKAJSXytTDdFc48xALWAYM7cgXrzKkb5DLZBQn2dv`
- drops dup vote `Vote { slots: [0, 1, 2, 3, 4], hash: 2QBgKPpebAQqRoL6BiL7UBR6wExB
u3ydsVRUjsZij8x7, timestamp: Some(1630448707) } is redundant for slot 41 vote account BvmUKAJSXytTDdFc48xALWAYM7cgXrzKkb5DLZBQn2dv`

2. the 'too old' vote: 
- processed vote ` Vote { slots: [0, 1, 2, 3], hash: yTjm9L8BGPjM6qKXYeCv8zKenVWpRcz
LCS59fdnHa6u, timestamp: Some(1630448707) } is good for slot 32 vote account BvmUKAJSXytTDdFc48xALWAYM7cgXrzKkb5DLZBQn2dv`
- drop TooOld vote `Vote { slots: [0, 1, 2], hash: 8WKp7SFz4YUb5QsySbNh8vkixJouwJ7ZAX
P9AUmcagNH, timestamp: Some(1630448707) } is redundant for slot 32 vote account BvmUKAJSXytTDdFc48xALWAYM7cgXrzKkb5DLZBQn2dv`

Are any of these case could possibly impact consensus negatively? @carllin

carllin:
> processed vote Vote { slots: [0, 1, 2, 3, 4], hash: 2QBgKPpebAQqRoL6BiL7UBR6wExB u3ydsVRUjsZij8x7, timestamp: Some(1630448707) } is good for slot 32 vote account BvmUKAJSXytTDdFc48xALWAYM7cgXrzKkb5DLZBQn2dv
drops dup vote Vote { slots: [0, 1, 2, 3, 4], hash: 2QBgKPpebAQqRoL6BiL7UBR6wExB u3ydsVRUjsZij8x7, timestamp: Some(1630448707) } is redundant for slot 41 vote account BvmUKAJSXytTDdFc48xALWAYM7cgXrzKkb5DLZBQn2dv

Hmmm dup vote shouldn't affect consensus, since it only needs one copy to be ingested

> the 'too old' vote:
processed vote Vote { slots: [0, 1, 2, 3], hash: yTjm9L8BGPjM6qKXYeCv8zKenVWpRcz LCS59fdnHa6u, timestamp: Some(1630448707) } is good for slot 32 vote account BvmUKAJSXytTDdFc48xALWAYM7cgXrzKkb5DLZBQn2dv
drop TooOld vote Vote { slots: [0, 1, 2], hash: 8WKp7SFz4YUb5QsySbNh8vkixJouwJ7ZAX P9AUmcagNH, timestamp: Some(1630448707) } is redundant for slot 32 vote account BvmUKAJSXytTDdFc48xALWAYM7cgXrzKkb5DLZBQn2dv

I would be worried if the older, dropped vote had slots that were *not* present in the processed vote, but the old vote here is a strict subset of the newer vote, so all the slot information should have landed.

It seems like certain validators' towers aren't making progress with this change, it would be helpful to log the vote state that has *landed in a bank* on each bank in the validator: https://github.com/solana-labs/solana/blob/8ad52fa095c0d51f212c7ecfac4c4f4b459e763d/core/src/consensus.rs#L223. Here, you can log the 

1. bank slot
2. the `vote_account_pubkey`
3. Slots in the vote state and their lockout

Goal here is to see if roots are making progress.

Then compare to the local tower of each validator with that `vote_account_pubkey`, which can be observed by logging the `local_vote_state`  every time the validator votes after this line here: https://github.com/solana-labs/solana/blob/8ad52fa095c0d51f212c7ecfac4c4f4b459e763d/core/src/consensus.rs#L369

1. If it's an issue with votes not landing, the local tower will most likely look much different than the vote state that's landed on chain. 
2. If the local tower is also not making progress, this means the validator is also not voting, in which case look for these logs: 
3.  https://github.com/solana-labs/solana/blob/8ad52fa095c0d51f212c7ecfac4c4f4b459e763d/core/src/replay_stage.rs#L586-L590from the validator that's not voting. This would most likely be due to the failed threshold check https://github.com/solana-labs/solana/blob/8ad52fa095c0d51f212c7ecfac4c4f4b459e763d/core/src/replay_stage.rs#L2450-L2452, which is computed here: https://github.com/solana-labs/solana/blob/8ad52fa095c0d51f212c7ecfac4c4f4b459e763d/core/src/replay_stage.rs#L2230-L2231. This check is also based on on-chain vote state, so can also be debugged by the printing of the on-chain vote state recommended above.

taozhu-chicago:
Thanks @carllin ! It makes good sense and super helpful. I'll do your suggestions -- helps me to understand consensus much more too.

PR #19154, opened 2021-08-10
Pass SnapshotConfig to AccountsHashVerifier

brooksprumo:
AccountsHashVerifier will need access to both the full and incremental
snapshot archive interval slots config values, which is in the
SnapshotConfig.

Also, cleanup some `Option<>` params and their references.

### DIFF
--- a/core/src/tvu.rs
+++ b/core/src/tvu.rs
@@ -220,7 +220,7 @@ impl Tvu {
             tvu_config.trusted_validators.clone(),
             tvu_config.halt_on_trusted_validators_accounts_hash_mismatch,
             tvu_config.accounts_hash_fault_injection_slots,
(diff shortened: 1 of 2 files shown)


core/src/tvu.rs
```diff
@@ -220,7 +220,7 @@ impl Tvu {
             tvu_config.trusted_validators.clone(),
             tvu_config.halt_on_trusted_validators_accounts_hash_mismatch,
             tvu_config.accounts_hash_fault_injection_slots,
-            snapshot_interval_slots,
+            snapshot_config.clone(),
```
jeffwashington, line 223: fyi, I spent some time figuring out why this was a clone. Looks like it does need its own copy.

jeffwashington (approved):
lgtm

PR #19155, opened 2021-08-10
Move Bank::get_incremental_snapshot_storages() into snapshot_utils

brooksprumo:
Filtering out storages for incremental snapshots will be needed by the
background services for incremental snapshot support, but there is not a
Bank at that point.  Since the filtering doesn't apply only to Bank, and
more to snapshots, move the functionality into snapshot_utils.

### DIFF
--- a/runtime/src/snapshot_utils.rs
+++ b/runtime/src/snapshot_utils.rs
@@ -1536,7 +1536,7 @@ pub fn snapshot_bank(
     archive_format: &ArchiveFormat,
     hash_for_testing: Option<Hash>,
 ) -> Result<()> {
-    let storages: Vec<_> = root_bank.get_snapshot_storages();
+    let storages = root_bank.get_snapshot_storages();
     let mut add_snapshot_time = Measure::start("add-snapshot-ms");
     add_bank_snapshot(snapshots_dir, root_bank, &storages, snapshot_version)?;
     add_snapshot_time.stop();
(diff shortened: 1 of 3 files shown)


runtime/src/snapshot_utils.rs
```diff
@@ -1795,6 +1799,19 @@ pub fn process_accounts_package_pre(
...
+/// `incremental_snapshot_base_slot`.
+pub fn filter_snapshot_storages_for_incremental_snapshot(
+    snapshot_storages: &mut SnapshotStorages,
+    incremental_snapshot_base_slot: Slot,
+) {
+    snapshot_storages.retain(|storage| {
+        storage
+            .first()
+            .map_or(false, |entry| entry.slot() > incremental_snapshot_base_slot)
+    });
```
brooksprumo, line 1812: The algorithm is different than the original one in Bank. This is for two reasons:
1. You cannot do a `retain()` within a `retain()` because of an oversight in Rust. Since `retain()` doesn't give mutable access to its iterator, the inner Vec cannot call `retain()`.
2. Since the `SnapshotStorages` is a Vec of Slots, the inner vector will always be for a single slot. So we can more easily filter by just checking the first item.

runtime/src/snapshot_utils.rs
```diff
@@ -1536,7 +1536,7 @@ pub fn snapshot_bank(
     archive_format: &ArchiveFormat,
     hash_for_testing: Option<Hash>,
 ) -> Result<()> {
-    let storages: Vec<_> = root_bank.get_snapshot_storages();
+    let storages = root_bank.get_snapshot_storages();
```
jeffwashington, line 1539: Mainly as an fyi, and not the biggest deal in the world, but when we get_snapshot_storages more deeply in AccountsDb, we also return a corresponding vector of Slots. It turns out to be expensive to ask each storage for its slot. Definitely would tackle this in a different pr. More of an optimization possibility for later.

runtime/src/snapshot_utils.rs
jeffwashington: seems reasonable.

jeffwashington (approved):
lgtm

PR #19163, opened 2021-08-10
Use last_valid_block_height in services and client apps

CriesofCarrots:
#### Problem
`last_valid_slot` is deprecated, and all clusters now support block-height RPCs.

#### Summary of Changes
Update SendTransactionService, `solana program`, and `solana-tokens` CLI to use last_valid_block_height

### DIFF
--- a/banks-client/src/lib.rs
+++ b/banks-client/src/lib.rs
@@ -62,7 +62,7 @@ impl BanksClient {
         &mut self,
         ctx: Context,
         commitment: CommitmentLevel,
-    ) -> impl Future<Output = io::Result<(FeeCalculator, Hash, Slot)>> + '_ {
+    ) -> impl Future<Output = io::Result<(FeeCalculator, Hash, u64)>> + '_ {
         self.inner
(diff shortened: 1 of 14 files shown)


CriesofCarrots:
@mvines , can you please look at the SendTransactionService changes? I don't think there should be any breakage for anyone, but would appreciate the sanity check.
Also, should I include BanksClient/BanksServer? Those changes will be more breaking (updating `get_fees_with_commitment_and_context` to return a block height, instead of a slot). But maybe nbd, if those crates are only being used for program tests.

jackcmay (approved):

mvines:
> @mvines , can you please look at the SendTransactionService changes? I don't think there should be any breakage for anyone, but would appreciate the sanity check.

lgtm

> Also, should I include BanksClient/BanksServer? Those changes will be more breaking (updating `get_fees_with_commitment_and_context` to return a block height, instead of a slot). But maybe nbd, if those crates are only being used for program tests.

yeah that's probably the right thing to do even though it's a little breaky.  It'll only affect program tests so shouldn't be too painful 🤞🏼

PR #19169, opened 2021-08-11
[OLD] Make SnapshotPackagerService aware of Incremental Snapshots

### LINKED ISSUE #19166: Make SnapshotPackagerService aware of Incremental Snapshots
`SnapshotPackagerService` needs to know if the PendingSnapshotPackage is a full snapshot or an incremental snapshot, and to do the right thing.

brooksprumo:
Instead of `PendingSnapshotPackage` being an `AccountsPackage`, I've
created a new enum for either a FullSnapshotPackage or an
IncrementalSnapshotPackage.  This type with handle the actual call to
`snapshot_utils::archive_snapshot_package()` so it knows about full
snapshots vs incremental snapshots.

Fixes #19166

### DIFF
--- a/core/src/accounts_hash_verifier.rs
+++ b/core/src/accounts_hash_verifier.rs
@@ -4,14 +4,16 @@
 // hash on gossip. Monitor gossip for messages from validators in the --trusted-validators
 // set and halt the node if a mismatch is detected.
 
-use crate::snapshot_packager_service::PendingSnapshotPackage;
 use rayon::ThreadPool;
 use solana_gossip::cluster_info::{ClusterInfo, MAX_SNAPSHOT_HASHES};
 use solana_runtime::{
     accounts_db,
     snapshot_archive_info::SnapshotArchiveInfoGetter,
     snapshot_config::SnapshotConfig,
-    snapshot_package::{AccountsPackage, AccountsPackagePre, AccountsPackageReceiver},
+    snapshot_package::{
+        AccountsPackage, AccountsPackagePre, AccountsPackageReceiver, PendingSnapshotPackage,
+        SnapshotPackage,
+    },
 };
 use solana_sdk::{clock::Slot, hash::Hash, pubkey::Pubkey};
 use std::collections::{HashMap, HashSet};
@@ -161,7 +163,8 @@ impl AccountsHashVerifier {
                 == 0
             {
                 if let Some(pending_snapshot_package) = pending_snapshot_package {
-                    *pending_snapshot_package.lock().unwrap() = Some(accounts_package);
+                    *pending_snapshot_package.lock().unwrap() =
+                        Some(SnapshotPackage::FullSnapshotPackage(accounts_package));
                 }
             }
         }
--- a/runtime/src/snapshot_package.rs
+++ b/runtime/src/snapshot_package.rs
@@ -4,10 +4,7 @@ use crate::{
 };
 use crate::{
     snapshot_archive_info::{SnapshotArchiveInfo, SnapshotArchiveInfoGetter},
-    snapshot_utils::{
-        ArchiveFormat, BankSnapshotInfo, Result, SnapshotVersion, TMP_FULL_SNAPSHOT_PREFIX,
-        TMP_INCREMENTAL_SNAPSHOT_PREFIX,
-    },
+    snapshot_utils::{self, ArchiveFormat, BankSnapshotInfo, Result, SnapshotVersion},
 };
 use log::*;
 use solana_sdk::clock::Slot;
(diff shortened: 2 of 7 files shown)


brooksprumo:
@sakridge @carllin This is the gist for how I'm thinking about adding incremental snapshot support to the background services. The channels be updated to send enum types that will either be FullSnapshot or IncrementalSnapshot. This way at creation time, when the background service picks which snapshot type, it'll be carried forward until the snapshot is ultimately archived. This should eliminate the need to query these details again at each stage.

With the enums, I'm also planning to add traits that'll handle the archiving/packaging/etc. You will see an impl here for archiving with the `SnapshotPackageArchiver`.

How's this sound? Any potential downsides or corner cases I should be aware of? Thanks in advance for your thoughts!

core/src/accounts_hash_verifier.rs
```diff
@@ -161,7 +163,8 @@ impl AccountsHashVerifier {
                 == 0
             {
                 if let Some(pending_snapshot_package) = pending_snapshot_package {
-                    *pending_snapshot_package.lock().unwrap() = Some(accounts_package);
+                    *pending_snapshot_package.lock().unwrap() =
+                        Some(SnapshotPackage::FullSnapshotPackage(accounts_package));
```
brooksprumo, line 167: FYI: The larger `if` block will end up with an additional `else if` to see if AccountsHashVerifier should take in incremental snapshot here instead.

runtime/src/snapshot_package.rs
```diff
@@ -16,14 +13,73 @@ use solana_sdk::hash::Hash;
...
+            SnapshotPackage::IncrementalSnapshotPackage(accounts_package) => {
+                snapshot_utils::archive_snapshot_package(
+                    accounts_package,
+                    maximum_snapshot_archives_to_retain,
+                    snapshot_utils::TMP_INCREMENTAL_SNAPSHOT_PREFIX,
+                )
+            }
+        }
+    }
+}
```
brooksprumo, line 68: Here is the main new bits. It's got the updated `PendingSnapshotPacakge`, the `SnapshotPackage` enum, and the trait + impl for `SnapshotPackageArchiver`.

runtime/src/snapshot_utils.rs
```diff
@@ -239,10 +239,11 @@ pub fn remove_tmp_snapshot_archives(snapshot_archives_dir: &Path) {
     }
 }
 
-/// Make a full snapshot archive out of the AccountsPackage
-pub fn archive_snapshot_package(
+/// Make a snapshot archive out of the AccountsPackage
+pub(crate) fn archive_snapshot_package(
```
brooksprumo, line 243: A nice fallout was that I can reduce the visibility into some `snapshot_utils` functions, since this functionality is contained within the traits in the previous comment.

runtime/src/snapshot_package.rs
```diff
@@ -16,14 +13,73 @@ use solana_sdk::hash::Hash;
...
+
+/// This trait provides the `archive_snapshot_package()` function that facilitates archiving a
+/// snapshot package
+pub trait SnapshotPackageArchiver {
+    /// Archive a SnapshotPackage
+    fn archive_snapshot_package(&self, maximum_snapshot_archives_to_retain: usize) -> Result<()>;
+}
+
+impl SnapshotPackageArchiver for SnapshotPackage {
```
carllin, line 49: hmmm is there any reason this has to be a trait? Looks like it could just be in an impl block for `SnapshotPackage`

runtime/src/snapshot_package.rs
```diff
@@ -16,14 +13,73 @@ use solana_sdk::hash::Hash;
...
+                    accounts_package,
+                    maximum_snapshot_archives_to_retain,
+                    snapshot_utils::TMP_INCREMENTAL_SNAPSHOT_PREFIX,
+                )
+            }
+        }
+    }
+}
+
+impl SnapshotArchiveInfoGetter for SnapshotPackage {
```
carllin, line 70: Same as in https://github.com/solana-labs/solana/pull/19169/files#r687120040, seems like this could just be in an `impl` block

runtime/src/snapshot_package.rs
brooksprumo: Ah yes, good call. Got wrapped up only thinking in traits!

runtime/src/snapshot_package.rs
```diff
@@ -16,14 +13,73 @@ use solana_sdk::hash::Hash;
...
 pub type AccountsPackageSender = Sender<AccountsPackagePre>;
+
+/// The receiver side of the AccountsPackage channel, used by AccountsHashVerifier
 pub type AccountsPackageReceiver = Receiver<AccountsPackagePre>;
+
+/// The error type when sending an AccountsPackage over the channel fails
 pub type AccountsPackageSendError = SendError<AccountsPackagePre>;
 
+/// The PendingSnapshotPackage passes a SnapshotPackage from AccountsHashVerifier to SnapshotPackagerService for archiving
+pub type PendingSnapshotPackage = Arc<Mutex<Option<SnapshotPackage>>>;
```
carllin, line 33: Does this make more sense to leave in `snapshot_packager_service.rs` since it defines the interface between  the `SnapshotPackagerService` itself and `AccountsHashVerifier`?

brooksprumo: Yeah, I was wondering about this myself... I moved it because `snapshot_package.rs` defines the interface types between `AccountsBackgroundService` and `AccountsHashVerifier`, so I was thinking it'd be nice to have it all in one spot. Should I move `PendingSnapshotPackage` back?

https://github.com/solana-labs/solana/blob/9dfeee2993312accb61b3d48270671de2774eb81/runtime/src/snapshot_package.rs#L23-L25

runtime/src/snapshot_package.rs
brooksprumo: Done!

runtime/src/snapshot_package.rs
brooksprumo: Done!

runtime/src/snapshot_package.rs
carllin: ah kk, I think that's fine then, prefer to keep them all in one spot

runtime/src/snapshot_package.rs
```diff
@@ -16,14 +13,70 @@ use solana_sdk::hash::Hash;
...
+
+/// This enum wraps an AccountsPackage to differentiate between a full snapshot package and an
+/// incremental snapshot package
+pub enum SnapshotPackage {
+    FullSnapshotPackage(AccountsPackage),
+    IncrementalSnapshotPackage(AccountsPackage),
+}
```
carllin, line 40: Because the struct fields are basically the same here, and we just want to distinguish between `FullSnapshotPackage` and `IncrementalSnapshotPackage`, it might make more sense to use the following pattern

```
struct AccountsPackage<T> {
     <rest of the fields>
     _phantom: PhantomData<T>,
}
```

Here's a similar example: https://github.com/solana-labs/solana/blob/v1.7/core/src/cluster_nodes.rs#L26-L36

This makes common functions like `fn snapshot_archive_info(&self)` easier to implement without needing to case on every possible enum option.

brooksprumo: I agree! I've been working on the next part, with AccountsHashVerifier, and I've basically done that, but as an enum field instead of the Phantom so that I can limit (plus easily match on) the variants:

https://github.com/solana-labs/solana/blob/129dab9cdf74f43931c87c2d2a0f36faeb64e033/runtime/src/snapshot_package.rs#L188-L189

(Here's all the changes: https://github.com/solana-labs/solana/pull/19215) _<--- that_

What do you think about _that_ version? I can pull that into this PR if you like it.

carllin: The advantage of the Phantom type is it's zero length and you can have separate `impl` blocks for the different variants, which introduces more clarity.

brooksprumo: Cool! Will do.

brooksprumo:
Due to a change in design, plus the merge conflicts, I started a new branch and have a new PR. Closing this one. Head over to #19254 for the continuation.

PR #19172, opened 2021-08-11
Split solana_sdk.h

jackcmay:
#### Problem

`solana_sdk.h` has become a kitchen sink for everything BPf program C related

#### Summary of Changes

Split the mono-header into logical units

Fixes #


programs/bpf/c/src/deprecated_loader/deprecated_loader.c
```diff
@@ -2,8 +2,9 @@
  * @brief Example C-based BPF program that prints out the parameters
  * passed to it
  */
-#include <solana_sdk.h>
-#include <deserialize_deprecated.h>
+#include <sol_types.h>
+#include <sol_log.h>
+#include <sol_deserialize_deprecated.h>
```
mvines, line 7: totally just a style thing, but did you consider this?
```suggestion
#include <sol/types.h>
#include <sol/log.h>
#include <sol/deserialize_deprecated.h>
```

jackcmay: heh, I thought you were in the anti-directory camp

sdk/bpf/c/inc/solana_sdk.h
```diff
@@ -1,741 +1,20 @@
...
-}
-#endif
-
-#ifdef __cplusplus
-}
-#endif
+ * @brief Solana C-based BPF program types utility functions
+ */
+
+#include <deserialize_deprecated.h>
```
mvines, line 6: Should this be `#include <sol_deserialize_deprecated.h>`?  If not the usage of `sol_deserialize_deprecated.h` elsewhere in this PR is probably wrong

mvines: oops, nm

mvines (approved):
much improvement!

programs/bpf/c/src/deprecated_loader/deprecated_loader.c
mvines: I'm in the anti-shift key camp!  So / and - always wins over _ on my keyboard.

sdk/bpf/c/inc/solana_sdk.h
jackcmay: Yeah, it should be

programs/bpf/c/src/deprecated_loader/deprecated_loader.c
jackcmay: heh, yeah, old ways.  how about `sol-types.h`?

mvines: kebab case for C headers isn't really a norm though I think, so snake feels better in this case sadly.  That's why I didn't suggest `sol-deserialize-deprecated.h`.  I can take or leave the change of `sol_` to `sol/`, totally your call

mvines (approved):
:bowtie:

PR #19183, opened 2021-08-11
removes raw indexing from streamer

behzadnouri:
#### Problem
Raw indexing is verbose and error-prone. In fact this same code had
indexing bug causing validator nodes panic just a few months ago:
https://github.com/solana-labs/solana/commit/482b8c6be

#### Summary of Changes
Use iterators instead of raw indices.


streamer/src/recvmmsg.rs
```diff
@@ -34,76 +41,78 @@ pub fn recv_mmsg(socket: &UdpSocket, packets: &mut [Packet]) -> io::Result<(usiz
     Ok((total_size, i))
 }
 
+#[cfg(target_os = "linux")]
+fn cast_socket_addr(addr: &sockaddr_storage, hdr: &mmsghdr) -> Option<InetAddr> {
+    use libc::{sa_family_t, sockaddr_in, sockaddr_in6};
+    const SOCKADDR_IN_SIZE: usize = std::mem::size_of::<sockaddr_in>();
+    const SOCKADDR_IN6_SIZE: usize = std::mem::size_of::<sockaddr_in6>();
+    if addr.ss_family == AF_INET as sa_family_t {
+        if hdr.msg_hdr.msg_namelen == SOCKADDR_IN_SIZE as socklen_t {
```
jbiseda, line 50: combine the ss_family and msg_namelen checks with &&

streamer/src/sendmmsg.rs
```diff
@@ -143,22 +122,13 @@ where
...
-    for (pkt, dest) in packets.iter() {
-        mmsghdr_for_packet(
-            pkt.as_ref(),
-            dest.borrow(),
-            &mut iovs,
-            &mut addrs,
-            &mut hdrs,
-        );
+    let size = packets.len();
+    let mut iovs = vec![unsafe { std::mem::zeroed() }; size];
```
jbiseda, line 126: iovs could use mem::MaybeUninit::uninit().assume_init() as these will always be completely specified

jbiseda (approved):
looks good, a couple small comments

streamer/src/recvmmsg.rs
behzadnouri: done

streamer/src/sendmmsg.rs
behzadnouri: done

PR #19184, opened 2021-08-11
removes use of public ip addresses from perf tests

behzadnouri:
#### Problem
Using global IPs causes outbound traffic which costs money: https://github.com/solana-labs/solana/pull/18728#issuecomment-884290209

Also, following https://github.com/solana-labs/solana/pull/19130 if `gce.sh creat` is invoked without `-P` then `--allow-private-addr` is implied:
https://github.com/solana-labs/solana/blob/4cc1b1504/net/common.sh#L68-L73

Therefore tests only need to specify:
```yaml
USE_PUBLIC_IP_ADDRESSES: "false"
```

#### Summary of Changes
* Set `USE_PUBLIC_IP_ADDRESSES: "false"`.
* Remove redundant `ALLOW_PRIVATE_ADDR`


behzadnouri:
@sakridge any objection to merging this?

sakridge (approved):
no, looks good

PR #19186, opened 2021-08-11
Add EtcdTowerStorage

mvines:
Add ability to store tower in an etcd cluster instead of the tower .bin file on the local disk
```
solana-validator ... \
  --tower-storage etcd \
  --etcd-cacert-file certs/etcd-ca.pem \
  --etcd-cert-file certs/client.pem \
  --etcd-key-file certs/client-key.pem \
  --etcd-endpoint localhost:2379
```

Notes:
* The saved tower itself is still signed
* If multiple validators run with the same identity, etcd transactions are used to lock out the older validator from voting as soon as the newer validator loads the tower from etcd.  This can be used in conjunction with the `solana-validator set-identity ...` command to affect a hot switch from the primary to backup validator *without* the risk of double signing/slashing.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1309,6 +1309,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "etcd-client"
+version = "0.7.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


mvines:
@leoluk - FYI.  Might want to wait until I get a chance to write up some usage docs though

PR #19191, opened 2021-08-11
ABI v2 Proposal

Lichtso:
Fixes: #14523

### Design
See [document](https://github.com/Lichtso/solana/blob/docs/abi-v2-proposal/docs/src/proposals/sbf-program-abi-v2.md) of this PR.

### Preparation
#15410, #17898, https://github.com/solana-labs/rbpf/pull/196 and #19469, #19762, #20034, #20165, #20290, #20308, #20448, #20540, #20598, #20785, #20881, #20954, #21108, #21180, #21185, #21205, #21395, #21404, #21545, #21563, #21574, #21671, #21882, #21927, #22102, #22107, #22111, #22165

### Runtime
Implementation of ABI v2 internals and inter-operation with ABI v1 by mounting the old ABI on top of the new one.
#21706, #22226, #22274

### Native programs
Replace the use of `InvokeContext` and `KeyedAccount` of ABI v1 by `InstructionContext` and `BorrowedAccount` of ABI v2.

### BPF programs
Add a new loader with a different entrypoint and syscalls: CPI, return data, realloc.
See [dev branch](https://github.com/Lichtso/solana/tree/feature/direct_mapping).

### Testing
Migrate and adjust the entire test suite around message processing.

### Benchmarking
Measure the limits of ABIv2.

### DIFF
--- a/docs/src/proposals/abi-v2.md
+++ b/docs/src/proposals/abi-v2.md
@@ -16,10 +16,10 @@ Information about each entry (attribute value pair) which is either encoded expl
 
 Some [binary data-serialization formats](https://en.wikipedia.org/wiki/Comparison_of_data-serialization_formats#Comparison_of_binary_formats) are [CBOR](https://cbor.io), [Ion](https://amzn.github.io/ion-docs/), [MsgPack](https://msgpack.org), [Protobuf](https://github.com/protocolbuffers/protobuf).
 We currently only have CBOR in our Cargo.lock, so everything else would add a dependency.
-However, non of these support indirections / references / offsets / pointers and they are all in row-major order (see next point). Therefor, we should design and implement our own extensibile serialization format.
+However, none of these support indirections / references / offsets / pointers and they are all in row-major order (see next point). Therefore, we should design and implement our own extensibile serialization format.
 
 ## Nesting order
-Borrowing the terms from table based data base systems:
+Borrowing the terms from table based database systems:
 - Row-major encoding (inner loop over attributes, outer loop over accounts) means first the values of all attributes of one account are encoded, then comes the next account.
 - Column-major encoding (inner loop over accounts, outer loop over attributes) means first the values of one attribute for all accounts are encoded, then comes the next attribute.
 
@@ -34,45 +34,53 @@ This section describes how the new encoding works in general and how we would us
 ### Encoding
 Everything is in little endian.
 
-Associative map:
-- `u16`: number_of_entries (attribute value pairs)
-- `[u16; number_of_entries]`: Attributes sorted ascending (for binary search)
-- `[u8; number_of_entries]`: Branching of values: Direct nesting, indirect nesting, direct leaf, indirect leaf
-- `[u32; number_of_entries]`: Offsets into values (cummulative value lengths)
-- Values
+```Rust
+struct Map {
+  number_of_entries: u16,
+  attribute: [u16; number_of_entries],
+  value_offset: [u32; number_of_entries + 1],
+  is_value_indirect: [u8; number_of_entries],
+  values: [Value],
+}
+
+enum Value {
+  Direct([u8]),
+  Indirect({ pointer: u64, length: u32 }),
+}
+```
 
 ### Memory Mapping / Regions
 Additional to the mandatory RBPF memory regions (null, program, stack, heap) we would add `2 + number_of_accounts_in_instruction` memory regions to the VM mapping.
 
 #### Read-only meta data region:
 - Transaction context:
-  - `number_of_accounts`: Natural number
-  - `account_key`: Vector of pubkeys for all accounts
-  - `account_is_executable`: Bit vector for all accounts
-  - `account_owner`: Vector of pubkeys for all accounts
-  - `account_lamports`: Vector of natural numbers for all accounts
-  - `account_data_length`: Vector of natural numbers for all accounts
-  - `account_data`: Vector of pointers for all accounts
-  - `account_rent_epoch` (deprecated): Vector of natural numbers for all accounts
+  - `NumberOfAccountsInTransaction = 1`: `u32`
+  - `AccountKey = 2`: `[Pubkey; NumberOfAccountsInTransaction]`
+  - `AccountIsExecutable = 3`: `[bool; NumberOfAccountsInTransaction]`
+  - `AccountOwner = 4`: `[Pubkey; NumberOfAccountsInTransaction]`
+  - `AccountLamports = 5`: `[u64; NumberOfAccountsInTransaction]`
+  - `AccountData = 6`: `[&[u8]; NumberOfAccountsInTransaction]`
+  - `InvocationStackFrame = 7`: `Map`
 - Instruction context:
-  - `invoke_stack_frame`: Offset into nested structure
-  - Invoke stack frames:
-    - `program_account_index`: Natural number index into the transaction accounts
-    - `instruction_data`: Blob
-    - `number_of_accounts`: Natural number
-    - `account_indices`: Vector of natural number indices into the transaction accounts
-    - `account_is_signer`: Bit vector for all accounts referenced by account_indices
-    - `account_is_writable`: Bit vector for all accounts referenced by account_indices
+  - `ParentStackFrame = 8`: `Map`
+  - `InstructionData = 9`: `[u8]`
+  - `NumberOfAccountsInInstruction = 10`: `u32`
+  - `InstructionAccountIndices = 11`: `[u32; NumberOfAccountsInInstruction]`
+  - `ProgramAccountIndex = 12`: `u32`
+  - `AccountIsSigner = 13`: `[bool; NumberOfAccountsInInstruction]`
+  - `AccountIsWritable = 14`: `[bool; NumberOfAccountsInInstruction]`
+  - `WritableAttributes = 15`: `Map`
 
 #### Read-write meta data region:
 - Instruction context:
-    - `account_owner`: Vector of pubkeys for writable accounts
-    - `account_lamports`: Vector of natural numbers for writable accounts
+  - `AccountOwner`,
+  - `AccountLamports`,
 
 #### Individual account data regions:
-- One for each account in the instruction
-- Read-only or read-write depends on instruction context
-- `data`: Blob
+- One mapped memory region for each account in the instruction
+- But virtual and physical addresses stay the same throughout the transaction
+- Read-only or read-write depends on the instruction context
+- Content: `[u8]`
 
 ### Usage / Integration
 Replace `KeyedAccounts`, `serialize_parameters`, `deserialize_parameters` and parts of the `InvokeContext` with a new interface which directly operates on the new encoding. It would be used by the runtime and programs alike.
--- a/docs/src/proposals/abi-v2.md
+++ b/docs/src/proposals/abi-v2.md
@@ -1,0 +1,80 @@
+---
+title: ABI v2
+---
+
+ABI between loader and program entrypoint, as well as syscalls such as (cross program) invocation and account reallocation.
+
+## Motivation
+The reason we need a new serialization format and thus also a new ABI is because the existing one is not extensibile and stores the account data in-place / in-stream. This prevents accounts from being memory mapped individually and also hinders reallocation of accounts (changing their size). Furthermore, there is a lot of potential for optimizations such as encoding the meta data only once per transaction and partially per instruction.
+
+## Existing Encoding Standards
+Information about each entry (attribute value pair) which is either encoded explicitly or can be implicitly gained from the context:
+- Attribute (semantics) as a tag or name string
+- Data type / encoding
+- Value length
+- Value content
+
+Some [binary data-serialization formats](https://en.wikipedia.org/wiki/Comparison_of_data-serialization_formats#Comparison_of_binary_formats) are [CBOR](https://cbor.io), [Ion](https://amzn.github.io/ion-docs/), [MsgPack](https://msgpack.org), [Protobuf](https://github.com/protocolbuffers/protobuf).
+We currently only have CBOR in our Cargo.lock, so everything else would add a dependency.
+However, non of these support indirections / references / offsets / pointers and they are all in row-major order (see next point). Therefor, we should design and implement our own extensibile serialization format.
+
+## Nesting order
+Borrowing the terms from table based data base systems:
+- Row-major encoding (inner loop over attributes, outer loop over accounts) means first the values of all attributes of one account are encoded, then comes the next account.
+- Column-major encoding (inner loop over accounts, outer loop over attributes) means first the values of one attribute for all accounts are encoded, then comes the next attribute.
+
+Column-major encoding has the following advantages in our use case:
+- Extensibility on an attribute level (but not per account level as that would be wasteful).
+- Read / write access control per attribute (not per account meta data, but the account data is handled separately).
+- Splitting meta data into a local instruction context and a global transaction context.
+
+## Design
+This section describes how the new encoding works in general and how we would use it for the new ABI specifically.
+
+### Encoding
+Everything is in little endian.
+
+Associative map:
+- `u16`: number_of_entries (attribute value pairs)
+- `[u16; number_of_entries]`: Attributes sorted ascending (for binary search)
+- `[u8; number_of_entries]`: Branching of values: Direct nesting, indirect nesting, direct leaf, indirect leaf
+- `[u32; number_of_entries]`: Offsets into values (cummulative value lengths)
+- Values
+
+### Memory Mapping / Regions
+Additional to the mandatory RBPF memory regions (null, program, stack, heap) we would add `2 + number_of_accounts_in_instruction` memory regions to the VM mapping.
+
+#### Read-only meta data region:
+- Transaction context:
+  - `number_of_accounts`: Natural number
+  - `account_key`: Vector of pubkeys for all accounts
+  - `account_is_executable`: Bit vector for all accounts
+  - `account_owner`: Vector of pubkeys for all accounts
+  - `account_lamports`: Vector of natural numbers for all accounts
+  - `account_data_length`: Vector of natural numbers for all accounts
+  - `account_data`: Vector of pointers for all accounts
+  - `account_rent_epoch` (deprecated): Vector of natural numbers for all accounts
+- Instruction context:
+  - `invoke_stack_frame`: Offset into nested structure
+  - Invoke stack frames:
+    - `program_account_index`: Natural number index into the transaction accounts
+    - `instruction_data`: Blob
+    - `number_of_accounts`: Natural number
+    - `account_indices`: Vector of natural number indices into the transaction accounts
+    - `account_is_signer`: Bit vector for all accounts referenced by account_indices
+    - `account_is_writable`: Bit vector for all accounts referenced by account_indices
+
+#### Read-write meta data region:
+- Instruction context:
+    - `account_owner`: Vector of pubkeys for writable accounts
+    - `account_lamports`: Vector of natural numbers for writable accounts
+
+#### Individual account data regions:
+- One for each account in the instruction
+- Read-only or read-write depends on instruction context
+- `data`: Blob
+
+### Usage / Integration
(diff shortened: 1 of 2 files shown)


docs/src/proposals/abi-v2.md
```diff
@@ -0,0 +1,80 @@
...
+## Existing Encoding Standards
+Information about each entry (attribute value pair) which is either encoded explicitly or can be implicitly gained from the context:
+- Attribute (semantics) as a tag or name string
+- Data type / encoding
+- Value length
+- Value content
+
+Some [binary data-serialization formats](https://en.wikipedia.org/wiki/Comparison_of_data-serialization_formats#Comparison_of_binary_formats) are [CBOR](https://cbor.io), [Ion](https://amzn.github.io/ion-docs/), [MsgPack](https://msgpack.org), [Protobuf](https://github.com/protocolbuffers/protobuf).
+We currently only have CBOR in our Cargo.lock, so everything else would add a dependency.
+However, non of these support indirections / references / offsets / pointers and they are all in row-major order (see next point). Therefor, we should design and implement our own extensibile serialization format.
```
dmakarov, line 19: However, non -> However, none

docs/src/proposals/abi-v2.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Nesting order
+Borrowing the terms from table based data base systems:
```
dmakarov, line 22: data base -> database

docs/src/proposals/abi-v2.md
```diff
@@ -0,0 +1,80 @@
...
+
+#### Individual account data regions:
+- One for each account in the instruction
+- Read-only or read-write depends on instruction context
+- `data`: Blob
+
+### Usage / Integration
+Replace `KeyedAccounts`, `serialize_parameters`, `deserialize_parameters` and parts of the `InvokeContext` with a new interface which directly operates on the new encoding. It would be used by the runtime and programs alike.
+
+As we have already seen in [#15410](https://github.com/solana-labs/solana/pull/15410) this will most likely cause problems with the Rust borrow checker, because we can not have multiple mutable references to the `InvokeContext`. So the workaround using indices as account handles might continue to be necessary.
```
dmakarov, line 80: can not -> cannot

dmakarov (dismissed):
Looks good to me.

jon-chuang:
To piggyback on ABI v2, we've also identified the need for `AccountInfo.owner` to be `Rc<RefCell<&'a mut Pubkey>>` instead of immutable `&'a Pubkey` in order to avoid `unsafe` when reassigning the owner. 

Not sure if this should be squeezed in to this proposal as "miscellaneous changes"

docs/src/proposals/abi-v2.md
Lichtso: Thx for proofreading

ryoqun:
dumb question: why all accounts needs to be passed wholly including account.data, always? why can't we provide set of api (syscalls?) to returns handles to n-th account's (address, owner, is_signed, data, etc) as separate functions?

I mean, as our defi composition increases, i think there will be more situations where programs are just passing accounts down to CPI-ed instructions. for example, raydium instructions are just passing serum dex's (large) event queue account and its program account. In these cases, just account address are needed, if i'm right? there could be wasted cycles even with this proposal? Or, behind-the-scene ondemand (page fault in traditional mmu sense) MemoryRegion could alleviate that?

also my (long-neglected) traced dynamic account loading proposal (#17796 ) would be more fit with the my dumb api approach.

Lichtso:
@ryoqun I admit that the the proposal document does not make the most important point totally obvious and needs a better formulation: The proposed solution would serialize the accounts meta data just once per **transaction** (instead of twice for every **instruction** call and return) and then map it **by reference** together with the account payload data / content.

> dumb question: why all accounts needs to be passed wholly including account.data, always? why can't we provide set of api (syscalls?) to returns handles to n-th account's (address, owner, is_signed, data, etc) as separate functions?

Syscalls are harder to optimize and more expensive than memory accesses (even with address translation). Thus, we try to shift accessors from syscalls to memory if they don't need to trigger anything in the runtime immediately.

> I mean, as our defi composition increases, i think there will be more situations where programs are just passing accounts down to CPI-ed instructions. for example, raydium instructions are just passing serum dex's (large) event queue account and its program bpf account. In these cases, just account address are needed, if i'm right?
> there could be wasted cycles even with this proposal?

This way there would be no more memory copies involved and passing everything (mind the pointer to everything, not a copy thereof) is cheaper than selecting what is needed (as that requires interaction between programs and the runtime via syscalls).

> Or, behind-the-scene ondemand (page fault in traditional mmu sense) MemoryRegion could alleviate that?

We don't even need to do the mapping on demand as we have constant time (O(1)) address translation in RBPF now.

> also my (long-neglected) traced dynamic account loading proposal (#17796 ) would be more fit with the my dumb api approach.

Not sure why these two are in conflict. Because dynamically adding / removing accounts is harder if their meta-data is serialized once at the beginning of a transaction?

ryoqun:
> I admit that the the proposal document does not make the most important point totally obvious and needs a better formulation: The proposed solution would serialize the accounts meta data just once per **transaction** (instead of twice for every **instruction** call and return) and then map it **by reference** together with the account payload data / content.

aha! now cleared my mind. :) so, this is also friendly with justin's transaction v2 (moar accounts!) and gpu (avoid system memory access from gpu by syscalls), i guess?

> Because dynamically adding / removing accounts is harder if their meta-data is serialized once at the beginning of a transaction?

yeah. but come to think of, column-major serialization scheme maybe easy to extend to dynamically load accounts at end of areas for each field, right?

Also, as for the traced dynamic account proposal (not about abi v2), mutating state of `&[AccountInfo]` (which could be onstack) like `.len()` could be daunting as execution progresses to maintain no observability between simulation or on-chain execution... maybe dynamically loaded accounts special marker (tx format change, ugh)...?

docs/src/proposals/abi-v2.md
```diff
@@ -0,0 +1,88 @@
...
+  Direct([u8]),
+  Indirect({ pointer: u64, length: u32 }),
+}
+```
+
+### Memory Mapping / Regions
+Additional to the mandatory RBPF memory regions (null, program, stack, heap) we would add `2 + number_of_accounts_in_instruction` memory regions to the VM mapping.
+
+#### Read-only meta data region:
+- Transaction context:
```
jackcmay, line 56: Are both the transaction context and the instruction context passed to each program?

docs/src/proposals/abi-v2.md
```diff
@@ -0,0 +1,88 @@
...
+  - `AccountKey = 2`: `[Pubkey; NumberOfAccountsInTransaction]`
+  - `AccountIsExecutable = 3`: `[bool; NumberOfAccountsInTransaction]`
+  - `AccountOwner = 4`: `[Pubkey; NumberOfAccountsInTransaction]`
+  - `AccountLamports = 5`: `[u64; NumberOfAccountsInTransaction]`
+  - `AccountData = 6`: `[&[u8]; NumberOfAccountsInTransaction]`
+  - `InvocationStackFrame = 7`: `Map`
+- Instruction context:
+  - `ParentStackFrame = 8`: `Map`
+  - `InstructionData = 9`: `[u8]`
+  - `NumberOfAccountsInInstruction = 10`: `u32`
```
jackcmay, line 67: These could be `u16`

docs/src/proposals/abi-v2.md
```diff
@@ -0,0 +1,88 @@
...
+```
+
+### Memory Mapping / Regions
+Additional to the mandatory RBPF memory regions (null, program, stack, heap) we would add `2 + number_of_accounts_in_instruction` memory regions to the VM mapping.
+
+#### Read-only meta data region:
+- Transaction context:
+  - `NumberOfAccountsInTransaction = 1`: `u32`
+  - `AccountKey = 2`: `[Pubkey; NumberOfAccountsInTransaction]`
+  - `AccountIsExecutable = 3`: `[bool; NumberOfAccountsInTransaction]`
```
jackcmay, line 59: Executable is read-write because the loader could mark an account as executable

docs/src/proposals/abi-v2.md
```diff
@@ -0,0 +1,88 @@
...
+  - `AccountOwner = 4`: `[Pubkey; NumberOfAccountsInTransaction]`
```
jackcmay, line 60: Are these duplicates of the rw fields below?

docs/src/proposals/abi-v2.md
```diff
@@ -0,0 +1,88 @@
...
+  - `AccountLamports = 5`: `[u64; NumberOfAccountsInTransaction]`
+  - `AccountData = 6`: `[&[u8]; NumberOfAccountsInTransaction]`
+  - `InvocationStackFrame = 7`: `Map`
+- Instruction context:
+  - `ParentStackFrame = 8`: `Map`
```
jackcmay, line 65: What is this?

docs/src/proposals/abi-v2.md
```diff
@@ -0,0 +1,88 @@
...
+  - `InstructionData = 9`: `[u8]`
+  - `NumberOfAccountsInInstruction = 10`: `u32`
+  - `InstructionAccountIndices = 11`: `[u32; NumberOfAccountsInInstruction]`
+  - `ProgramAccountIndex = 12`: `u32`
+  - `AccountIsSigner = 13`: `[bool; NumberOfAccountsInInstruction]`
+  - `AccountIsWritable = 14`: `[bool; NumberOfAccountsInInstruction]`
+  - `WritableAttributes = 15`: `Map`
```
jackcmay, line 72: Why isn't this the same as `AccountIsSigner`?

docs/src/proposals/abi-v2.md
```diff
@@ -0,0 +1,88 @@
+---
+title: ABI v2
```
jackcmay, line 2: "SBF Program ABI v2"

docs/src/proposals/abi-v2.md
```diff
@@ -0,0 +1,88 @@
...
+---
+
+ABI between loader and program entrypoint, as well as syscalls such as (cross program) invocation and account reallocation.
```
jackcmay, line 5: I don't see anything syscall related in this document

docs/src/proposals/abi-v2.md
Lichtso: Yes, behind the global transaction context follows a stack with a local instruction context in each frame. All in the same memory region. That also means that deeper invocations can see their parent frames, but I don't think that is a problem, because it is public via the network protocol anyway.

docs/src/proposals/abi-v2.md
Lichtso: Yes, and I am not happy with that yet. The problem arises because some values of this attribute are read-only while others are writable.

docs/src/proposals/abi-v2.md
Lichtso: Good catch, will add it to the writable list.

docs/src/proposals/abi-v2.md
Lichtso: Currently the stack is modeled as a linked list (because every stack frame can have a different size). But I will add support for arrays of pointers in values, which is needed for account data anyway. Then we can also refer to an invocation stack frame by index.

docs/src/proposals/abi-v2.md
Lichtso: `WritableAttributes` is a pointer to the read-write meta data region.
If that was the question.

docs/src/proposals/abi-v2.md
Lichtso: Sure but saving two bytes here won't really make a dent.
Compressing all boolean arrays into bit vectors might help a lot more.

docs/src/proposals/abi-v2.md
Lichtso: Yep, better title, agreed.

docs/src/proposals/abi-v2.md
Lichtso: Right, will add a section for them as well.

docs/src/proposals/abi-v2.md
jackcmay: Any reason not to share the entire region (transaction context and all frames) with each layer of the call stack but pass each program an index to their instruction context within that region?

Also, what does this look like when making a CPI call?  Do we copy and modify the instruction contexts to include the current CPI instruction?

docs/src/proposals/abi-v2.md
jackcmay: I was referring to all the u32s in these contexts.  But vector boolean would save space, though unpacking bit-arrays in the program might not be worth it.

docs/src/proposals/abi-v2.md
jackcmay: I see, the duplicate entries are sure to cause confusion.  Executable, owner, lamports, etc... are all together writeable or not, can they be in the same region with a marker of whether the region (account) is rw or not?

docs/src/proposals/abi-v2.md
jackcmay: Ah, maybe a pointer to the "possibly writable" attributes and `AccountIsWritable` indicates if they are writable or not?

docs/src/proposals/sbf-program-abi-v2.md
```diff
@@ -0,0 +1,103 @@
...
+  value_offset: [u32; number_of_entries],
+  values: [Value],
+}
+
+struct FatPtr {
+  address: u64,
+  length: u32,
+}
+
+enum Value {
```
jackcmay, line 50: These require special representation in C?

docs/src/proposals/sbf-program-abi-v2.md
```diff
@@ -0,0 +1,103 @@
...
+  - `NumberOfAccountsInTransaction = 0`: `u32`
+  - `AccountKey = 1`: `[Pubkey; NumberOfAccountsInTransaction]`
+  - `AccountIsExecutable = 2`: `[bool; NumberOfAccountsInTransaction]`
+  - `AccountOwner = 3`: `[Pubkey; NumberOfAccountsInTransaction]`
+  - `AccountLamports = 4`: `[u64; NumberOfAccountsInTransaction]`
+  - `AccountData = 5`: `[&[u8]; NumberOfAccountsInTransaction]`
+  - `InvocationStackHeight = 6`: `u16`
+  - `InvocationStack = 7`: `[Map; InvocationStackHeight]`
+- Instruction context:
+  - `InstructionData = 8`: `[u8]`
```
jackcmay, line 72: where is the length of the instruction data communicated?

docs/src/proposals/sbf-program-abi-v2.md
```diff
@@ -0,0 +1,103 @@
...
+  - `WritableAttributes = 14`: `Map`
+
+#### Read-write meta data region:
+- Instruction context:
+  - `AccountIsExecutable`,
+  - `AccountOwner`,
+  - `AccountLamports`,
+
+#### Individual account data regions:
+- One mapped memory region for each account in the instruction
```
jackcmay, line 87: Where is the length of the data communicated and the length itself maybe be writeable to support allocation/deallocation?

Lichtso: The data is a `Value::Indirect` thus in the `FatPtr`.
It cannot be writable directly as we need to trigger the reallocation of the backing storage and we have no page fault mechanism.

docs/src/proposals/sbf-program-abi-v2.md
Lichtso: Currently only by the value slice (difference between adjacent `value_offset`s).
However that can might include padding, so yes, we should add an explicit `InstructionDataLength` attribute.

docs/src/proposals/sbf-program-abi-v2.md
Lichtso: Ah yes, this is outdated and should be described as an `union` now.
Because whether a value is direct or indirect is not encoded explicitly anymore but contextual from the usage.

docs/src/proposals/abi-v2.md
Lichtso: The idea is to replace the internal CPI [`invoke_stack`](https://github.com/solana-labs/solana/blob/5bb43280eb531e4702d3ed1ddc37dd0dd65d3668/runtime/src/message_processor.rs#L290) of the `InvokeContext` by the `InvocationStack` in the encoded structure and have it operate directly on that as well.

docs/src/proposals/abi-v2.md
Lichtso: The invocation stack is now an array of fatpointers.

brooksprumo:
I know this PR has been around for a while, so pardon the late comment. In a previous role I used [Flatbuffers](https://google.github.io/flatbuffers/index.html) to do serialization from an embedded IoT device to send to the cloud for processing. Without fully understanding our use-case here, I wonder if Flatbuffers may work. Then that's one fewer thing we need to create and maintain ourselves.

Flatbuffers is designed to be memory and runtime efficient, designed for mobile/games where those are a premium. Data in in little endian. It can support ABI changes over time. You can access the flatbuffer directly without needing to parse/unpack first.

It looks like the Rust version has a few features missing compared to the C++ version, but figured I'd pass thing along in case it was helpful!

Lichtso:
Thanks, I just scanned over it and will read it in detail once I get back to this PR.

One thing which almost all modern serialization formats (not older formats such as [Property list](https://en.wikipedia.org/wiki/Property_list)) are missing, and which we require for memory mapping of accounts, is an approach to indirection such as references and pointers. It seems tables in Flatbuffers might go in that direction, but I have to read further.

The other problem is that essential all serialization formats are row major encodings, and a column major encoding might be better for defining memory access permissions. But, I am not entirely convinced either way yet.

docs/src/proposals/sbf-program-abi-v2.md
```diff
@@ -0,0 +1,125 @@
...
+### Syscalls
+
+#### Cross Program Invocation (CPI)
+The runtime automatically allocates an empty invocation frame (mapped as writable) above the currently active one. If a program wants to call another one it simply writes directly into that empty invocation frame and configures it. Then the program only needs a syscall to trigger the execution and that syscall does not need to copy data forth and back between userspace and runtime anymore. It only has to validate the invocation frame, record the instruction and run the called program.
+
+#### Return Data Accessors
+In the old ABI return data was implemented by a setter and a getter syscall for copying data forth and back between userspace and runtime. But now, we can use shared memory in the transaction context instead. So, the two syscalls will be deprecated as there is no need to trigger anything in the runtime.
+
+#### Account Reallocation / Resizing
+TODO
```
jstarry, line 125: @Lichtso do you have any initial thoughts about how new account allocations will work? Could it be potentially dynamic such that CPI's could allocate many MB's of account data if needed? cc @brooksprumo

aeyakovenko: maybe if the TX specifies the expected allocations in the TX

docs/src/proposals/sbf-program-abi-v2.md
```diff
@@ -0,0 +1,121 @@
...
+Everything is in little endian and padded to alignment for fast access.
+Tight packing makes no sense as the encoded structure does not need to be copied in and out of the VMs, making the additional memory footprint of padding in the meta data irrelevant.
+
+### Memory Mapping / Regions
+Additional to the mandatory RBPF memory regions (null, program, stack, heap) we would add the following memory regions to the VM mapping:
+- Readonly attributes:
+  - Account Pubkeys
+  - Owner Pubkeys
+  - Is-executable flags
+  - Rent epochs
```
ryoqun, line 56: i think we can stop passing this field (`rent_epoch`) as we're effectively discontinuing rent-paying account support to squeeze cycles and reduce attack surface.

Lichtso: Agreed, however currently it is still necessary as ABIv1 is now running on-top of ABIv2 so it needs to be passed through for backward compatibility. But, we might be able to hide the attribute / property field in ABIv2 BPF programs.

PR #19192, opened 2021-08-11
Fix fork race condition in optimistic violation tower tests

### LINKED ISSUE #19017: Optimistic confirmation tests flakiness
#### Problem
`test_no_optimistic_confirmation_violation_with_tower` and `test_optimistic_confirmation_violation_without_tower` fail in about 1 in 5 runs.

In order to prime the test, we wait until the validators have voted on at least 27 slots. However because of the staking weights, and the fact that one validator C is offline, if for any reason there is a fork, validator A will be unable to switch over to validator B's fork because it cannot generate a switching proof (needs 38%, however the stake of validator B is only 36% and validator C is offline). 

If validator C is taken offline while leader, a fork could arise, and validator A will be stuck thus timing out the test.

#### Proposed Solution
Instead of priming by waiting for the required votes, we manually create the tower for the validators. We also submit the votes in gossip, so that the nodes see the optimistic confirmation. Then we truncate the ledgers and start the test. This avoids the issue of any potential forks during the priming stage. 

We need to make sure that the votes for slot 27 are present in gossip, otherwise validator A will not see the optimistic confirmation.

AshwinSekar:
#### Problem
The optimistic violation tower tests warmup by letting two validators run for a pre determined # of slots. Because of the stake weighting, this warmup fails if there is ever a fork. 

#### Summary of Changes
Instead run only a single validator and manually create a tower and ledger for the other validator in order to warmup

Fixes #19017


AshwinSekar:
Not stale, this commit fixes the original issue however it exposed another flakiness in the test which i am currently trying to address.

PR #19194, opened 2021-08-11
Block Ancestors Proposal

carllin:
#### Problem
EpochSlots takes too much bandwidth in gossip

#### Summary of Changes
1) Summary of our last discussion related to encoding a block's ancestors in its shreds.
2) Details related to changes to the duplicate slots protocols discussed separately between @AshwinSekar and myself.

Fixes #


docs/src/proposals/block_ancestors.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Removing EpochSlots from gossip
+
+EpochSlots takes a lot of bandwidth in gossip, and we would ultimately like to remove this
+burden on the network. Removing EpochSlots has a few fallouts described in the following
+sections.
+
+### Propagation Check
+
+TBD
```
carllin, line 58: This is a major blocker, I'm not sure how we get a safe equivalent to the current propagation check without `EpochSlots`

PR #19195, opened 2021-08-12
v1.6: Add token program hack

joncinque:
Hack at the level of the BPF loader instead of the account db load.  It doesn't have the neatness of using the exact account from the db, but `include_bytes!` makes this still pretty easy.


upload-perf/src/upload-perf.rs
```diff
@@ -74,7 +74,6 @@ fn main() {
                         ("commit", git_commit_hash.trim().to_string(), String)
                     );
                     */
-
```
mvines, line 77: 🔍  What is this stray whitespace!

mvines (commented):
looks like it should work!

upload-perf/src/upload-perf.rs
joncinque: :weary: this is from the new version of `cargo fmt` -- the old version forces this whitespace (not the first time this has happened to me)

mvines: r-

PR #19204, opened 2021-08-12
filters for recent contact-infos when checking for live stake

behzadnouri:
#### Problem
As spotted by @CriesofCarrots, wait-for-supermajority does not wait for anyone.

Contact-infos are saved to disk:
https://github.com/solana-labs/solana/blob/9dfeee299/gossip/src/cluster_info.rs#L1678-L1683

and restored on validator start-up:
https://github.com/solana-labs/solana/blob/9dfeee299/core/src/validator.rs#L450

Staked nodes entries will not expire until an epoch after. So when the
validator checks for online stake it is erroneously picking up
contact-infos restored from disk, which breaks the entire
wait-for-supermajority logic:
https://github.com/solana-labs/solana/blob/9dfeee299/core/src/validator.rs#L1515-L1561

#### Summary of Changes
This commit adds an extra check for the age of contact-info entries and
filters out old ones.

### DIFF
--- a/core/src/validator.rs
+++ b/core/src/validator.rs
@@ -1,93 +1,98 @@
 //! The `validator` module hosts all the validator microservices.
 
-use crate::{
-    broadcast_stage::BroadcastStageType,
-    cache_block_meta_service::{CacheBlockMetaSender, CacheBlockMetaService},
-    cluster_info_vote_listener::VoteTracker,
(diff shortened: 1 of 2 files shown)


CriesofCarrots (approved):
Gave this a spin; working great. Lgtm

PR #19209, opened 2021-08-12
Explorer: price can be unexpectedly missing even under Coingecko success 

oJshua:
#### Problem
It's possible for price data to be missing under a Coingecko query, even when status is success.

#### Summary of Changes
Only display price data if price exists (in TokenAccountSection)


oJshua:
Example: https://explorer.solana.com/address/2prC8tcVsXwVJAinhxd2zeMeWMWaVyzPoQeLKyDZRFKd

PR #19210, opened 2021-08-12
removes repeated bank-forks locking in window-service

behzadnouri:
#### Problem
Window service is repeatedly locking bank-forks to look-up working-bank
for every single shred:
https://github.com/solana-labs/solana/blob/5fde4ee3a/core/src/window_service.rs#L597-L606

#### Summary of Changes
* This commit updates shred_filter signature in recv_window so that where
we already obtain the lock on bank-forks, we can also look-up
working-bank once for all packets:
https://github.com/solana-labs/solana/blob/5fde4ee3a/core/src/window_service.rs#L256-L277
* Also cleaning up redundant uses of `&Arc` and `pub`s.

### DIFF
--- a/core/src/window_service.rs
+++ b/core/src/window_service.rs
@@ -5,7 +5,6 @@ use crate::{
     ancestor_hashes_service::AncestorHashesReplayUpdateReceiver,
     cluster_info_vote_listener::VerifiedVoteReceiver,
     completed_data_sets_service::CompletedDataSetsSender,
-    outstanding_requests::OutstandingRequests,
     repair_response,
     repair_service::{OutstandingShredRepairs, RepairInfo, RepairService},
     result::{Error, Result},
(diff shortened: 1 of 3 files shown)


core/src/window_service.rs
```diff
@@ -265,7 +265,7 @@ fn recv_window<F>(
     thread_pool: &ThreadPool,
 ) -> Result<()>
 where
-    F: Fn(&Shred, u64) -> bool + Sync,
+    F: Fn(&Shred, Arc<Bank>, /*last root:*/ u64) -> bool + Sync,
```
carllin, line 268: nit: `u64 -> Slot`

core/src/window_service.rs
```diff
@@ -286,8 +289,9 @@ where
         // call to `new_from_serialized_shred` is safe.
         assert_eq!(packet.data.len(), PACKET_DATA_SIZE);
         let serialized_shred = packet.data.to_vec();
+        let working_bank = Arc::clone(&working_bank);
```
carllin, line 292: 👍

carllin (dismissed):
Thanks for the cleanup!

jbiseda (dismissed):
lgtm

core/src/window_service.rs
behzadnouri: done

core/src/window_service.rs
```diff
@@ -386,12 +390,11 @@ impl WindowService {
     ) -> WindowService
     where
         F: 'static
-            + Fn(&Pubkey, &Shred, Option<Arc<Bank>>, u64) -> bool
+            + Fn(&Pubkey, &Shred, Option<Arc<Bank>>, /*last root:*/ u64) -> bool
```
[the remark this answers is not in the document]
behzadnouri: done

PR #19211, opened 2021-08-12
docs: create key with docker

dubalda:
#### Problem

When installing the client, there may be dependency issues, for example:
```
$ sh -c "$(curl -sSfL https://release.solana.com/v1.7.10/install)"
downloading v1.7.10 installer
/tmp/tmp.oIAuUMaDxs/solana-install-init: error while loading shared libraries: libssl.so.1.1: cannot open shared object file: No such file or directory

yum install openssl11-libs

$ sh -c "$(curl -sSfL https://release.solana.com/v1.7.10/install)"
downloading v1.7.10 installer
/tmp/tmp.rS1nYVEjKR/solana-install-init: /lib64/libc.so.6: version `GLIBC_2.18' not found (required by /tmp/tmp.rS1nYVEjKR/solana-install-init)

can`t install on CentOS7
```

#### Summary of Changes

How to use docker instead of cli installation

### DIFF
--- a/docs/src/wallet-guide/file-system-wallet.md
+++ b/docs/src/wallet-guide/file-system-wallet.md
@@ -23,6 +23,15 @@ mkdir ~/my-solana-wallet
 solana-keygen new --outfile ~/my-solana-wallet/my-keypair.json
 ```
 
+or use docker instead of cli install:
+```
+SOLANA_WALLET_DIR=~/my-solana-wallet
+SOLANA_KEYPAIR_FILENAME=my-keypair.json
+mkdir -p ${SOLANA_WALLET_DIR}
+docker run -it --rm --entrypoint "solana-keygen" --volume ${SOLANA_WALLET_DIR}:/tmp/solana \
+  solanalabs/solana:stable new --outfile /tmp/solana/${SOLANA_KEYPAIR_FILENAME}
+```
+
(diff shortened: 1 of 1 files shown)


dubalda:
@mvines review pls )

mvines:
I don't feel great about this direction:
1. The solanalabs/solana:stable Docker container is something we use for CI but isn't really something we strongly support in the community. It works and we have no plans to yank it, but I'd rather not increase its exposure.
2. Recommending people download a 300MB docker binary blob to create an ed25519 keypair seems to deviate from security best practices of minimizing the amount of code between you and your private key

Instead I think we'd recommend users build from source: https://docs.solana.com/cli/install-solana-cli-tools#build-from-source

dubalda:
> 
> 
> I don't feel great about this direction:
> 
>     1. The solanalabs/solana:stable Docker container is something we use for CI but isn't really something we strongly support in the community. It works and we have no plans to yank it, but I'd rather not increase its exposure.
> 
>     2. Recommending people download a 300MB docker binary blob to create an ed25519 keypair seems to deviate from security best practices of minimizing the amount of code between you and your private key
> 
> 
> Instead I think we'd recommend users build from source: https://docs.solana.com/cli/install-solana-cli-tools#build-from-source

I agree. But you need to give a choice: install rust, cargo and compile, or use docker. For some, using docker would be preferable

PR #19212, opened 2021-08-12
replace btree with hashmap

jeffwashington:
#### Summary of Changes

Fixes #


runtime/src/bank.rs
```diff
@@ -7266,8 +7264,10 @@ pub(crate) mod tests {
         }
     }
 
-    fn map_to_test_bad_range() -> AccountMap<Pubkey, i8> {
-        let mut map: AccountMap<Pubkey, i8> = AccountMap::new();
+    fn map_to_test_bad_range() -> std::collections::BTreeMap<Pubkey, i8> {
+        // It looks like this is used to test both bank range functions as well as range errors from BTreeMap.
+        // We will likely refactor AccountsIndex to make a higher level fn range() that mimics the previous single BTreeMap::range.
+        let mut map: std::collections::BTreeMap<Pubkey, i8> = std::collections::BTreeMap::new();
```
jeffwashington, line 7270: @ryoqun, AccountsIndex already doesn't work quite like this since we are now binning the index by pubkeys. Now that we plan to no longer use btree, that makes this code even less applicable. I could add a fn on AccountsIndex that returns a range, but I think references are going to get difficult to deal with since individual bins are behind their own rwlocks. If the purpose of this code is to verify that the ranges are being created correctly, then using a btree here is fine. If the purpose is to do more integration testing and test that the AccountsIndex is checking ranges correctly, then we have to rework this. Thoughts?

jeffwashington: Actually, see: #19225

ryoqun: @jeffwashington this is only used to unit-test `pubkey_range_from_partition`. #19225 seems fine to merge in that regard. :)

jeffwashington: I reworked this separately in #19225.

jeffwashington:
running gce perf tests now. First few didn't run because of user error...

carllin:
> running gce perf tests now. First few didn't run because of user error...

I think to accurately simulate perf, we should run a node against mainnet, or a testnet with a similar number of accounts. Specifically the call to `collect_rent_eagerly()`: https://github.com/solana-labs/solana/blob/140abec6ef80e64c8fa58254bf99ad8ad8d6e6c2/runtime/src/bank.rs#L3928-L3939

I don't know if the number of accounts rent is collected from is uniform over all banks frozen in an epoch (I think banks that have nonconsecutive parents i.e. skipped a slot might have more work to do), but maybe we can measure the average amount of time spent in rent collection per `X` banks. @ryoqun is that a reasonable way to measure this performance?

jeffwashington:
perf looks bad for this. not sure why yet.

ryoqun:
> I don't know if the number of accounts rent is collected from is uniform over all banks frozen in an epoch (I think banks that have nonconsecutive parents i.e. skipped a slot might have more work to do), but maybe we can measure the average amount of time spent in rent collection per `X` banks. @ryoqun is that a reasonable way to measure this performance?

yeah. i think so. unless  there is violent skip rate in the given cluster, it should be ok to ignore skipped banks.

jeffwashington:
different discussion in #19239

jeffwashington:
[bench-tps, duration: 600, mean: 24601, max: 49047](https://solanalabs.slack.com/archives/CP2L2S4KV/p1629307087025000)
[master]()

jeffwashington:
#19299 provides a mechanism to avoid sorting altogether.

runtime/src/accounts.rs
```diff
@@ -787,6 +787,7 @@ impl Accounts {
             "load_to_collect_rent_eagerly_scan_elapsed",
             ancestors,
             range,
+            true,
```
jeffwashington, line 790: This 'true' causes us to find all account indexes that match the range. The closure loads all the accounts and returns (pubkey, AccountSharedData). So, it doesn't matter at all how many accounts are returned per iteration of the deep scan. The 'true' also avoids the sort completely. So, this change is equivalent (at least for rent collection) to eliminating the sort and returning a full bucket of account index items at a time. #19299. But, this change also causes fallback rpc scans (when there is no secondary index) to work as well.

jeffwashington:
I reverted the # bins to 16. It seems to be better performance on bench-tps. Since we are no longer sorting on the only use case we care about performance for (rent), then it doesn't matter to this change how many bins there are.

jeffwashington:
@sakridge please take a look at the tps results.

carllin:
hmmm, because rent perf scales with the number of accounts, maybe bench-tps doesn't accurately reflect the impact on rent perf without onboarding the same number of accounts as mainnet beta?

jeffwashington:
> I reverted the # bins to 16. It seems to be better performance on bench-tps. Since we are no longer sorting on the only use case we care about performance for (rent), then it doesn't matter to this change how many bins there are.

This was incorrect. It does matter how many buckets there are even without sorting. 16 bins with 80M accounts results in 5M accounts per bin. We must linearly search the entire bucket to find all keys matching rent's range. Thus, smaller bucket sizes make the time reasonable.

jeffwashington:
Tests demonstrate out to 235M accounts that 8k buckets have equivalent rent performance.
We need disk buckets to get to higher account levels much higher than this. That change is coming. Disk buckets are expected to work like a HashMap, not like a BTreeMap. We will also likely need higher bin/bucket counts as we move to disk buckets and higher account counts. So, this change seems to be >= current performance, equivalent in functionality, and on the path towards where we are trying to go. It removes uncertainty from removing the sorting behavior of the map when we do go to disk buckets. Rent collection will also be refactored in the hopefully near term.

PR #19214, opened 2021-08-12
fix compile error in master

jeffwashington:
#### Problem
Not sure how this got through.
#### Summary of Changes

Fixes #


jeffwashington:
@brooksprumo did we collide somehow?

PR #19216, opened 2021-08-12
change untar to use unpack instead of unpack_in

jeffwashington:
#### Problem
tar crate changed its behavior as part of a security update. An internal call in `unpack_in` call was changed from:
`fs::create_dir_all(...)` // does NOT return error if path already exists
to
`fs::create_dir(...)` // returns error if path already exists

The goal was to fix a security hole where items could be unpacked in folders outside the destination folder.

There are possibly concerns with how we unpack to several configurable directories, which may be on different drives completely. That would seem to also trigger an error.

This pr switches from using unpack_in to unpack so that we control sanitizing.

#19175
[tar crate change](https://github.com/alexcrichton/tar-rs/commit/d90a02f582c03dfa0fd11c78d608d0974625ae5d#diff-3dcefa956e75e2171b83e5134b542405a2adb7909a16dc03fad7fd92e8e2d945R776)
[PR by @CriesofCarrots against tar crate](https://github.com/alexcrichton/tar-rs/pull/260)

#### Summary of Changes

Closes #19230

### DIFF
--- a/runtime/src/hardened_unpack.rs
+++ b/runtime/src/hardened_unpack.rs
@@ -162,16 +162,13 @@ where
         )?;
         total_count = checked_total_count_increment(total_count, limit_count)?;
 
-        let target = sanitize_path(&entry.path()?, unpack_dir)?;
+        let target = sanitize_path(&entry.path()?, unpack_dir)?; // ? handles file system errors
         if target.is_none() {
-            continue; // failed
+            continue; // skip it
         }
         let target = target.unwrap();
 
         let unpack = entry.unpack(target);
-        if unpack.is_err() {
-            error!("failing to unpack: {:?}", unpack_dir);
-        }
         check_unpack_result(unpack.map(|_unpack| true)?, path_str)?;
 
         // Sanitize permissions.
@@ -208,12 +205,14 @@ where
     }
 }
 
-// return true if path is good
+// return Err on file system error
+// return Some(path) if path is good
+// return None if we should skip this file
 fn sanitize_path(entry_path: &Path, dst: &Path) -> Result<Option<PathBuf>> {
+    // We cannot call unpack_in because it errors if we try to use 2 account paths.
+    // So, this code is borrowed from unpack_in
     // code from: unpack_in does its own sanitization
     // ref: https://docs.rs/tar/*/tar/struct.Entry.html#method.unpack_in
-    // we cannot call unpack_in because it errors if an unpack dir already exists
-
     let mut file_dst = dst.to_path_buf();
     {
         let path = entry_path;
@@ -238,7 +237,7 @@ fn sanitize_path(entry_path: &Path, dst: &Path) -> Result<Option<PathBuf>> {
     // Skip cases where only slashes or '.' parts were seen, because
     // this is effectively an empty filename.
     if *dst == *file_dst {
-        return Ok(None); // should this be false?
+        return Ok(None);
     }
 
     // Skip entries without a parent (i.e. outside of FS root)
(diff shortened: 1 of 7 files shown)


jeffwashington:
@mvines @CriesofCarrots

jeffwashington:
@CriesofCarrots 
I haven't cleaned this up yet, but it passes. I guess I should rebase it off the new security update.

jeffwashington:
#19216

CriesofCarrots:
Passed ✅ Is it ready for a close look?

jeffwashington:
I'm not confident in the sanitization partial port that I did. This has been a side project. Having someone else look at it would be swell.

jeffwashington:
Here is the code we're adapting:
https://github.com/alexcrichton/tar-rs/blob/d90a02f582c03dfa0fd11c78d608d0974625ae5d/src/entry.rs#L364

jeffwashington:
To make things more confusing, here is a change that just changes unpack_in to unpack and is a little cleaned up. I tried to more closely run the same sanitization code. It currently is failing the ci. Not sure why.
#19258

runtime/src/hardened_unpack.rs
```diff
@@ -9,6 +9,7 @@ use {
         fs::{self, File},
         io::{BufReader, Read},
         path::{
+            Component,
             Component::{CurDir, Normal},
```
CriesofCarrots, line 13:
```suggestion
            Component::{self, CurDir, Normal},
```

runtime/src/hardened_unpack.rs
```diff
@@ -199,6 +205,76 @@ where
     }
 }
 
+// return Err on file system error
+// return Some(path) if path is good
+// return None if we should skip this file
+fn sanitize_path(entry_path: &Path, dst: &Path) -> Result<Option<PathBuf>> {
+    // We cannot call unpack_in because it errors if we try to use 2 account paths.
+    // So, this code is borrowed from unpack_in
+    // code from: unpack_in does its own sanitization
```
CriesofCarrots, line 214: Delete this line?

runtime/src/hardened_unpack.rs
```diff
@@ -199,6 +205,76 @@ where
     }
 }
 
+// return Err on file system error
+// return Some(path) if path is good
+// return None if we should skip this file
+fn sanitize_path(entry_path: &Path, dst: &Path) -> Result<Option<PathBuf>> {
+    // We cannot call unpack_in because it errors if we try to use 2 account paths.
```
CriesofCarrots, line 212: It would be helpful to call out and justify differences between our implementation and unpack_in, perhaps in code below?

runtime/src/hardened_unpack.rs
```diff
@@ -199,6 +205,76 @@ where
...
+    };
+
+    fs::create_dir_all(parent)?;
+    validate_inside_dst(dst, parent)?;
+    let target = parent.join(entry_path.file_name().unwrap());
+
+    Ok(Some(target))
+}
+
+fn validate_inside_dst(dst: &Path, file_dst: &Path) -> Result<PathBuf> {
```
CriesofCarrots, line 257: Perhaps a comment here mentioning that this duplicates the tar-rs method?

CriesofCarrots (approved):
Thanks for taking this @jeffwashington . Implementation looks good to me; one nit and a couple ideas about more explicit documentation

t-nelson (approved):
lgtm!

CriesofCarrots (approved):
:shipit:

Issue #19217, opened 2021-08-13
`SystemInstruction::Deallocate`


jon-chuang:
#### Problem
The current idiom applied in the Solanaverse for closing accounts is to defund the account, whereupon it will be loaded as an empty account for subsequent transactions.

This is because if address has `lamports == 0`, `Bank.accounts` returns `AccountSharedData::default()` for all loading paths.

However, within the same transaction, this account address could either be hijacked or kept in permanent limbo through various means by other instructions.

#### Proposed Solution
Expose `SystemInstruction::Deallocate`, which can instantly set the account to be equivalent to `AccountSharedData::default()`, i.e. with 0 lamports and len 0 data slice. This forces further use of that address to obey the rules of authorisation for system accounts.

#### Discussion
There are several directions possible.

1. The calling program needs to set the system program to be the account owner, and thus must first zero out the account data, before calling `SystemInstruction::Deallocate`. Currently, zeroing out the account data is probably cheapest to do with a `sol_memset_`, which does the following thing, and is hopefully turned into an efficient implementation by the rust compiler:

<details>

```rust
let s = question_mark!(
    translate_slice_mut::<u8>(memory_mapping, s_addr, n, self.loader_id, true),
    result
);
for val in s.iter_mut().take(n as usize) {
    *val = c as u8;
}
```
Although we know for sure that [`std::ptr::write_bytes`]( https://doc.rust-lang.org/std/ptr/fn.write_bytes.html) does a memset under the hood. According to [this stackoverflow thread](https://stackoverflow.com/questions/51732596/what-is-the-equivalent-of-a-safe-memset-for-slices), the compiler can do memset for large arrays for the above code snippet already.

Testing this, here is the rust.godbolt output with the -O flag:
```rust
pub fn thing(buffer: &mut [u8]) {
    for i in &mut buffer[..] { *i = 240 }
}
```
```asm
example::thing:
        test    rsi, rsi
        je      .LBB1_2
        push    rax
        mov     rdx, rsi
        mov     esi, 240
        call    qword ptr [rip + memset@GOTPCREL]
        add     rsp, 8
```
So I think we're good.

</details>

This costs at the current default compute budget is 1 compute unit per 250 bytes, so a grand total of 40,000 CUs per 10MB, definitely very doable.

[...]

PR #19226, opened 2021-08-13
Allocate can re-allocate accounts that are non-empty, only not nonces

jon-chuang:
#### Problem
We'd like non-nonce accounts to be able to reallocate their data

#### Summary of Changes
Allow arbitrary reallocations as long as system account is not nonce account

Fixes https://github.com/solana-labs/solana/issues/19217

@jstarry

### DIFF
--- a/runtime/src/system_instruction_processor.rs
+++ b/runtime/src/system_instruction_processor.rs
@@ -79,7 +79,7 @@ fn allocate(
 
     // if it looks like the `to` account is already in use, bail
     //   (note that the id check is also enforced by message_processor)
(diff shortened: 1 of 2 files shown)


jon-chuang:
Actually, this is a terrible idea. 

For instance, `sysvar::Instructions` is considered a system account. Now I sincerely hope that the construction of the nonce account doesn't allow construction from `sysvar::Instructions`. That would permit changing the `sysvar::Instructions` data.

jstarry:
@jon-chuang we won't need this if we add support for dealloc right?

jon-chuang:
Yes, that, or account resizing by owning programs

PR #19233, opened 2021-08-13
retransmits shreds recovered from erasure codes

behzadnouri:
#### Problem
Shreds recovered from erasure codes have not been received from turbine
and have not been retransmitted to other nodes downstream. This results
in more repairs across the cluster which is slower.

#### Summary of Changes
* 4th commit reworks retransmit stage to simply receive shreds instead of packets; because converting recovered "shreds" to "packets" just to send them to retransmit-stage is superfluous. This will also avoid [partially deserializing packet in retransmit stage](https://github.com/solana-labs/solana/blob/32501866b/core/src/retransmit_stage.rs#L233) when the entire packet was already deserialized to shred earlier in the window-stage.
* In preparation of the above change, first 2 commits move a number of metrics related to packet-counts from retransmit stage to window-service earlier in the pipeline. Metrics for repair and discard packets are emitted in the window stage and, with the 4th commit, those packets are no longer sent down to retransmit-stage (which was pointless since they will not be retransmitted).
* Last commit, channels through recovered shreds to retransmit stage in order to further broadcast the shreds to downstream nodes in the tree.


core/src/window_service.rs
```diff
@@ -316,29 +354,32 @@ where
...
-            let _ = retransmit.send(packets);
-        }
+    // Exclude repair packets from retransmit.
+    let _ = retransmit_sender.send(
+        shreds
+            .iter()
+            .zip(&repair_infos)
+            .filter(|(_, repair_info)| repair_info.is_none())
+            .map(|(shred, _)| shred)
+            .cloned()
```
carllin, line 368: Wonder if we could avoid this clone if we switched over to some `Arc<Shred>`, but that's a story for another time

core/src/retransmit_stage.rs
```diff
@@ -560,21 +454,23 @@ impl RetransmitStage {
...
         completed_data_sets_sender: CompletedDataSetsSender,
-        max_slots: &Arc<MaxSlots>,
+        max_slots: Arc<MaxSlots>,
         rpc_subscriptions: Option<Arc<RpcSubscriptions>>,
         duplicate_slots_sender: Sender<Slot>,
         ancestor_hashes_replay_update_receiver: AncestorHashesReplayUpdateReceiver,
     ) -> Self {
         let (retransmit_sender, retransmit_receiver) = channel();
+        // https://github.com/rust-lang/rust/issues/39364#issuecomment-634545136
+        let _retransmit_sender = retransmit_sender.clone();
```
carllin, line 464: ick, I've run into this too.

carllin:
> Shreds recovered from erasure codes have not been received from turbine
and have not been retransmitted to other nodes downstream. This results
in more repairs across the cluster which is slower.

Hmm, what if in cases where we just happen to receive some combination of erasure shreds + data shreds first such that we recover some data shreds that we *eventually* would have gotten from turbine. Will this cause a massive spike in bandwidth if we then retransmit these recovered shreds, even though turbine had no problem circulating them? 

To avoid this, should we wait for a bit to see that these recovered shreds don't ultimately arrive through turbine?

carllin (approved):
Nice, this is so much cleaner, awesome! 😃

core/src/window_service.rs
```diff
@@ -546,16 +589,16 @@ impl WindowService {
         insert_sender: CrossbeamSender<(Vec<Shred>, Vec<Option<RepairMeta>>)>,
         verified_receiver: CrossbeamReceiver<Vec<Packets>>,
         shred_filter: F,
-        leader_schedule_cache: Arc<LeaderScheduleCache>,
         bank_forks: Arc<RwLock<BankForks>>,
-        retransmit: PacketSender,
+        retransmit_sender: Sender<Vec<Shred>>,
```
jbiseda, line 593: thanks for renaming these :)

jbiseda (approved):
looks good!

KeKo6988:
> #### Problem
> Shreds recovered from erasure codes have not been received from turbine
> and have not been retransmitted to other nodes downstream. This results
> in more repairs across the cluster which is slower.
> 
> #### Summary of Changes
> * 4th commit reworks retransmit stage to simply receive shreds instead of packets; because converting recovered "shreds" to "packets" just to send them to retransmit-stage is superfluous. This will also avoid [partially deserializing packet in retransmit stage](https://github.com/solana-labs/solana/blob/32501866b/core/src/retransmit_stage.rs#L233) when the entire packet was already deserialized to shred earlier in the window-stage.
> * In preparation of the above change, first 2 commits move a number of metrics related to packet-counts from retransmit stage to window-service earlier in the pipeline. Metrics for repair and discard packets are emitted in the window stage and, with the 4th commit, those packets are no longer sent down to retransmit-stage (which was pointless since they will not be retransmitted).
> * Last commit, channels through recovered shreds to retransmit stage in order to further broadcast the shreds to downstream nodes in the tree.

core/benches/retransmit_stage.rs
```diff
@@ -17,11 +17,10 @@ use {
         shred::Shredder,
     },
     solana_measure::measure::Measure,
-    solana_perf::packet::{Packet, Packets},
     solana_runtime::{bank::Bank, bank_forks::BankForks},
     solana_sdk::{
         hash::Hash,
-        pubkey,
+        pubkey::Pubkey,
```
KeKo6988, line 23: 5rZrRNA9NyAig5B6F8V98EfUFPxS3T6hKa4rT5cdvbJC

core/src/retransmit_stage.rs
```diff
@@ -33,17 +33,13 @@ use {
         pubkey::Pubkey,
```
KeKo6988, line 33: 5rZrRNA9NyAig5B6F8V98EfUFPxS3T6hKa4rT5cdvbJC

KeKo6988 (approved):

behzadnouri:
> Hmm, what if in cases where we just happen to receive some combination of erasure shreds + data shreds first such that we recover some data shreds that we _eventually_ would have gotten from turbine. Will this cause a massive spike in bandwidth if we then retransmit these recovered shreds, even though turbine had no problem circulating them?
> 
> To avoid this, should we wait for a bit to see that these recovered shreds don't ultimately arrive through turbine?

That should be picked up by duplicates check in retransmit stage (i.e. `check_if_already_received`) and so the 2nd copy of the shred will be skipped:
https://github.com/solana-labs/solana/blob/0e5ea36cc/core/src/retransmit_stage.rs#L227-L253

PR #19235, opened 2021-08-14
Add AsRef<AccountInfo> for AccountInfo

fanatid:
Ref https://github.com/project-serum/anchor/issues/608

#### Problem

In [project-serum/anchor](https://github.com/project-serum/anchor/) we want to optimize code generation with something like:
```rust
pub trait KeyRef<'info> {
    fn key(&self) -> &'info Pubkey;
}

impl<'info, T> KeyRef<'info> for T
where
    T: AsRef<AccountInfo<'info>>,
{
    fn key(&self) -> &'info Pubkey {
        self.as_ref().key
    }
}
```

of course this produce error:
```
upstream crates may add a new impl of trait `std::convert::AsRef<solana_program::account_info::AccountInfo<'_>>` for type `solana_program::account_info::AccountInfo<'_>` in future versions
```

#### Summary of Changes

Implement `AsRef<AccountInfo<'a>>` for `AccountInfo<'a>`.

### DIFF
--- a/sdk/program/src/account_info.rs
+++ b/sdk/program/src/account_info.rs
@@ -266,4 +266,13 @@ mod tests {
         assert_eq!(k4, *info2_3_4[2].key);
         assert_eq!(k5, *info5.key);
     }
+
+    #[test]
+    fn test_account_info_as_ref() {
+        let k = Pubkey::new_unique();
+        let l = &mut 0;
+        let d = &mut [0u8];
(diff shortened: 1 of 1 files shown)


mvines:
lgtm but would you mind also adding a quick test for this, one that'll yell if somebody removes this in the future

fanatid:
Good idea :+1:  Test added.

mvines:
Thanks!  This'll ship in Solana v1.7.11

PR #19239, opened 2021-08-16
wip: Hashmap accounts index perf

jeffwashington:
#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/accounts_index.rs
+++ b/runtime/src/accounts_index.rs
@@ -615,12 +615,14 @@ pub struct AccountsIndexIterator<'a, T> {
     start_bound: Bound<Pubkey>,
     end_bound: Bound<Pubkey>,
     is_finished: bool,
+    collect_all_unsorted: bool,
 }
 
 impl<'a, T> AccountsIndexIterator<'a, T> {
     fn range<'b, R>(
         map: &'b AccountMapsReadLock<'b, T>,
         range: R,
+        collect_all_unsorted: bool,
     ) -> Vec<(&'b Pubkey, &'b AccountMapEntry<T>)>
     where
         R: RangeBounds<Pubkey>,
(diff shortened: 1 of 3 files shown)


runtime/src/accounts_index.rs
```diff
@@ -631,7 +633,9 @@ impl<'a, T> AccountsIndexIterator<'a, T> {
                 result.push((k, v));
             }
         }
-        result.sort_unstable_by(|a, b| a.0.cmp(b.0));
+        if !collect_all_unsorted {
```
jeffwashington, line 636: @ryoqun, thoughts on this? We don't need to sort for rent collection.

runtime/src/accounts_index.rs
```diff
@@ -707,10 +712,12 @@ impl<'a, T: 'static + Clone> Iterator for AccountsIndexIterator<'a, T> {
...
-            for (pubkey, account_map_entry) in
-                Self::range(&i.read().unwrap(), (self.start_bound, self.end_bound))
-            {
-                if chunk.len() >= ITER_BATCH_SIZE {
+            for (pubkey, account_map_entry) in Self::range(
+                &i.read().unwrap(),
+                (self.start_bound, self.end_bound),
+                self.collect_all_unsorted,
+            ) {
+                if chunk.len() >= ITER_BATCH_SIZE && !self.collect_all_unsorted {
```
jeffwashington, line 720: @ryoqun, and for rent collection we don't need to limit the # of results we return at once.

ryoqun: yes too. (ideally, we would like to parallelize the eager rent collection.. or change the whole rent collection mechanism altogether with @aeyakovenko 's account.deletion_slot based rent collection approch...)

jeffwashington: @carllin @sakridge - this is a commit that gets rid of the sorting for rent collection to understand that more deeply. Initial bench-tps/perf-test numbers indicate this is better than current master. I don't have accurate numbers yet. There is a regression in master that behzad is looking at.

PR #19253, opened 2021-08-16
adds back cluster partitioning to broadcast-duplicates

behzadnouri:
#### Problem
An earlier version of this code was aiming to create a partition by
manipulating stakes, and setting some of them to zero:
https://github.com/solana-labs/solana/blob/cde146155/core/src/broadcast_stage/broadcast_duplicates_run.rs#L65-L116

https://github.com/solana-labs/solana/pull/18971
moved stakes computation further down the stream, and so that logic
could no longer live there. 

#### Summary of Changes
* This commit adds back cluster partitioning by intercepting packets. This truly creates a partition because demonstrably some nodes are skipped right before sending packets.
* However the old code:
https://github.com/solana-labs/solana/blob/b67ffab37/core/src/broadcast_stage/broadcast_duplicates_run.rs#L77-L99
was aiming to create a partition by setting some of the stakes to zero. But that does not actually create a partition. Because unstaked nodes were always part of the turbine broadcast and retransmit stages as long as their contact-info exists in gossip (ie. in cluster-info).

### DIFF
--- a/core/src/broadcast_stage/broadcast_duplicates_run.rs
+++ b/core/src/broadcast_stage/broadcast_duplicates_run.rs
@@ -1,6 +1,7 @@
 use {
     super::*,
     crate::cluster_nodes::ClusterNodesCache,
+    itertools::Itertools,
     solana_entry::entry::Entry,
     solana_ledger::shred::Shredder,
     solana_runtime::blockhash_queue::BlockhashQueue,
@@ -9,6 +10,7 @@ use {
         signature::{Keypair, Signer},
         system_transaction,
     },
+    std::collections::HashSet,
 };
 
 pub const MINIMUM_DUPLICATE_SLOT: Slot = 20;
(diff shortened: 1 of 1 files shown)


core/src/broadcast_stage/broadcast_duplicates_run.rs
```diff
@@ -198,19 +204,22 @@ impl BroadcastRun for BroadcastDuplicatesRun {
 
         // Special handling of last shred to cause partition
         if let Some((original_last_data_shred, partition_last_data_shred)) = last_shreds {
+            self.original_last_data_shreds.extend(
+                original_last_data_shred
+                    .iter()
+                    .map(|shred| shred.payload.clone()),
```
jbiseda, line 210: Can you use the signature here instead of the payload?

behzadnouri: done

jbiseda (approved):

PR #19254, opened 2021-08-16
Make SnapshotPackagerService aware of Incremental Snapshots

### LINKED ISSUE #19166: Make SnapshotPackagerService aware of Incremental Snapshots

brooksprumo:
#### Problem

SnapshotPackagerService does not know about Incremental Snapshots

#### Summary of Changes

Add a field to SnapshotPackage that is an enum for SnapshotType, so `archive_snapshot_package()` will do the right thing.

Fixes #19166

_Note: This is a continuation of #19169_

### DIFF
--- a/core/src/snapshot_packager_service.rs
+++ b/core/src/snapshot_packager_service.rs
@@ -1,19 +1,17 @@
 use solana_gossip::cluster_info::{ClusterInfo, MAX_SNAPSHOT_HASHES};
 use solana_runtime::{
-    snapshot_archive_info::SnapshotArchiveInfoGetter, snapshot_package::SnapshotPackage,
+    snapshot_archive_info::SnapshotArchiveInfoGetter, snapshot_package::PendingSnapshotPackage,
     snapshot_utils,
 };
 use solana_sdk::{clock::Slot, hash::Hash};
 use std::{
     sync::{
         atomic::{AtomicBool, Ordering},
-        Arc, Mutex,
+        Arc,
     },
     thread::{self, Builder, JoinHandle},
     time::Duration,
 };
-
-pub type PendingSnapshotPackage = Arc<Mutex<Option<SnapshotPackage>>>;
 
 pub struct SnapshotPackagerService {
     t_snapshot_packager: JoinHandle<()>,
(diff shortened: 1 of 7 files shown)


core/src/snapshot_packager_service.rs
```diff
@@ -175,6 +176,7 @@ mod tests {
             Hash::default(),
             ArchiveFormat::TarBzip2,
             SnapshotVersion::default(),
+            SnapshotType::FullSnapshot,
```
lijunwangs, line 179: Add test cases for incremental snapshot.

runtime/src/snapshot_package.rs
```diff
@@ -16,14 +16,26 @@ use solana_sdk::hash::Hash;
...
 };
 use tempfile::TempDir;
 
+/// The sender side of the AccountsPackage channel, used by AccountsBackgroundService
 pub type AccountsPackageSender = Sender<AccountsPackage>;
+
+/// The receiver side of the AccountsPackage channel, used by AccountsHashVerifier
 pub type AccountsPackageReceiver = Receiver<AccountsPackage>;
+
+/// The error type when sending an AccountsPackage over the channel fails
```
carllin, line 32: Thanks for the new set of comments!

runtime/src/snapshot_utils.rs
```diff
@@ -268,10 +269,14 @@ pub fn archive_snapshot_package(
         .map_err(|e| SnapshotError::IoWithSource(e, "create archive path"))?;
 
     // Create the staging directories
+    let staging_dir_prefix = match snapshot_package.snapshot_type {
+        SnapshotType::FullSnapshot => TMP_FULL_SNAPSHOT_PREFIX,
+        SnapshotType::IncrementalSnapshot => TMP_INCREMENTAL_SNAPSHOT_PREFIX,
+    };
```
carllin, line 275: could maybe be a method `to_prefix()` on `SnapshotType`

core/src/snapshot_packager_service.rs
brooksprumo: Sure thing! What would you like the tests to cover? Currently, all the "real" snapshot tests are either under `snapshot_utils` or `core/tests/snapshots.rs`. Are you thinking of basically a duplicate of `snapshot_packager_service::create_and_verify_snapshot()`?

lijunwangs: yes. Maybe we can just tag on this test case to add one more incremental snapshot?

brooksprumo: OK, will do.

runtime/src/snapshot_utils.rs

core/src/snapshot_packager_service.rs
brooksprumo: @lijunwangs So, looking into this test here, it's kind of funny. There is actually nothing specific about the SnapshotPackagerService being tested here. It's mostly reimplementations of the functions in snapshot_utils, then creating fake bank snapshots, building a SnapshotPackage, then passing it to functions in snapshot_utils to the actual achiving and verifying.

Since there's nothing specific to the SnapshotPackagerService being tested, and because the test itself isn't particularly "good", I'd rather not modify it here, as I don't see any added benefit. The tests in snapshot_utils and `core/tests/snapshots.rs` do already cover the test here, plus actually testing SnapshotPackagerService.

What do you think?

lijunwangs: Sounds good

lijunwangs (approved):
LGTM

PR #19256, opened 2021-08-16
discards epoch-slots epochs ahead of the current root

behzadnouri:
#### Problem
Cross cluster gossip contamination is causing cluster-slots hash map to
contain a lot of bogus values and consume too much memory:
https://github.com/solana-labs/solana/issues/17789

If a node is using the same identity key across clusters, then these
erroneous values might not be filtered out by shred-versions check,
because one of the variants of the contact-info will have matching
shred-version:
https://github.com/solana-labs/solana/issues/17789#issuecomment-896304969

#### Summary of Changes
The cluster-slots hash-map is bounded and trimmed at the lower end by
the current root. This commit also discards slots epochs ahead of the
root.

### DIFF
--- a/core/src/cluster_slots.rs
+++ b/core/src/cluster_slots.rs
@@ -10,7 +10,6 @@ use {
     },
     std::{
         collections::{BTreeMap, HashMap},
-        ops::Range,
         sync::{Arc, Mutex, RwLock},
     },
 };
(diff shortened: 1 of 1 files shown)


core/src/cluster_slots.rs
```diff
@@ -40,6 +44,8 @@ impl ClusterSlots {
     }
 
     fn update_internal(&self, root: Slot, epoch_slots_list: Vec<EpochSlots>) {
+        // Discard slots at or before current root or epochs ahead.
+        const SLOTS_DIFF_RANGE: Range<u64> = 1..2 * DEFAULT_SLOTS_PER_EPOCH;
```
sakridge, line 48: Ideally we get this from the runtime value: `bank.get_slots_in_epoch()`

behzadnouri: done.
added an extra `.max(DEFAULT_SLOTS_PER_EPOCH)` just to be a bit more conservative in discarding epoch-slots.

core/src/cluster_slots.rs
```diff
@@ -54,13 +58,20 @@ impl ClusterSlots {
                 })
                 .collect()
         };
+        // Discard slots at or before current root or epochs ahead.
+        let slot_range = (root + 1)
```
carllin, line 62: hmmm I think we need some buffer for the slots before the root for nodes that are catching up. There's also a use case currently for handling duplicate slots where nodes that have gotten the wrong version of a block rely on seeing `EpochSlots` reach a certain percentage of the threshold before initiating the duplicate repair protocol. cc @AshwinSekar 

Is it feasible to keep around at least the latest epoch of `EpochSlots`?

behzadnouri: That sounds like a separate change independent of this commit.
Current master code is already discarding epoch slots at or before current root:
https://github.com/solana-labs/solana/blob/692aa9914/core/src/cluster_slots.rs#L61-L63
https://github.com/solana-labs/solana/blob/692aa9914/core/src/cluster_slots.rs#L82

carllin: oh shoot this is the internal update for *other* nodes' epoch slots wer're receiving, but we're still persisting our own updates that are earlier than the root. Sorry for the confusion!

carllin (approved):

PR #19257, opened 2021-08-16
accounts_index_bins to AccountsIndexConfig

jeffwashington:
#### Problem
Rework validator options to prepare to pass through more accounts index configuration options.
#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/accounts_index.rs
+++ b/runtime/src/accounts_index.rs
@@ -773,8 +773,8 @@ impl<T: IsCached> AccountsIndex<T> {
         Self::new(Some(BINS_FOR_TESTING))
     }
 
-    pub fn new(bins: Option<AccountsIndexConfig>) -> Self {
-        let (account_maps, bin_calculator) = Self::allocate_accounts_index(bins);
+    pub fn new(config: Option<AccountsIndexConfig>) -> Self {
+        let (account_maps, bin_calculator) = Self::allocate_accounts_index(config);
         Self {
             account_maps,
             bin_calculator,
@@ -794,9 +794,11 @@ impl<T: IsCached> AccountsIndex<T> {
     }
 
     fn allocate_accounts_index(
-        bins: Option<AccountsIndexConfig>,
+        config: Option<AccountsIndexConfig>,
     ) -> (LockMapType<T>, PubkeyBinCalculator16) {
-        let bins = bins.and_then(|config| config.bins).unwrap_or(BINS_DEFAULT);
+        let bins = config
+            .and_then(|config| config.bins)
+            .unwrap_or(BINS_DEFAULT);
         let account_maps = (0..bins)
             .into_iter()
             .map(|_| RwLock::new(AccountMap::default()))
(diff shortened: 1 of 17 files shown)


runtime/src/accounts_index.rs
```diff
@@ -78,6 +82,11 @@ pub struct AccountSecondaryIndexesIncludeExclude {
     pub keys: HashSet<Pubkey>,
 }
 
+#[derive(Debug, Default, Clone, Copy)]
```
jeffwashington, line 85: A lot of noise to introduce this struct.
This struct will contain more options as we move forward.

runtime/src/accounts_index.rs
```diff
@@ -33,8 +33,12 @@ use thiserror::Error;
...
 pub const ITER_BATCH_SIZE: usize = 1000;
 pub const BINS_DEFAULT: usize = 16;
-pub const BINS_FOR_TESTING: usize = BINS_DEFAULT;
-pub const BINS_FOR_BENCHMARKS: usize = BINS_DEFAULT;
+pub const BINS_FOR_TESTING: AccountsIndexConfig = AccountsIndexConfig {
+    bins: Some(BINS_DEFAULT),
+};
+pub const BINS_FOR_BENCHMARKS: AccountsIndexConfig = AccountsIndexConfig {
+    bins: Some(BINS_DEFAULT),
+};
```
brooksprumo, line 41: If you anticipate adding more fields, maybe it makes sense to rename these constants to indicate the type is now `AccountsIndexConfig`. Maybe like `ACCOUNTS_INDEX_CONFIG_FOR_TESTING`/etc?

I'll leave it up to you if you'd like to address that in this PR or not.

runtime/src/accounts_index.rs
brooksprumo: Worth it! I think this is a great idea to prevent runaway function parameters from the start.

runtime/src/accounts_index.rs
```diff
@@ -764,7 +773,7 @@ impl<T: IsCached> AccountsIndex<T> {
         Self::new(Some(BINS_FOR_TESTING))
     }
 
-    pub fn new(bins: Option<usize>) -> Self {
+    pub fn new(bins: Option<AccountsIndexConfig>) -> Self {
```
brooksprumo, line 776: Similar to the other comment, maybe the `bins` parameter should be `accounts_index_config`.

runtime/src/accounts_index.rs
```diff
@@ -784,8 +793,10 @@ impl<T: IsCached> AccountsIndex<T> {
         }
     }
 
-    fn allocate_accounts_index(bins: Option<usize>) -> (LockMapType<T>, PubkeyBinCalculator16) {
-        let bins = bins.unwrap_or(BINS_DEFAULT);
+    fn allocate_accounts_index(
+        bins: Option<AccountsIndexConfig>,
```
brooksprumo, line 797: Same here: `bins` to `accounts_index_config`.

brooksprumo (approved):
Love the idea. I think there are a few things you could address here, but I'll leave it up to you whether you want to make the changes here or in a subsequent PR. Looks good.

Issue #18746, opened 2021-07-17
RPC simulations should allow for passing in user provided account data


jon-chuang:
#### Problem
Users may want to simulate transactions based on hypothetical account state.

#### Proposed Solution
Use a json format to pass account data, bank ingests this data and writes over any data fetched from accountsDB when executing transaction.

Related https://github.com/solana-labs/solana/pull/17796

ryoqun:
yeah, I like this additional facility. :)

PR #19263, opened 2021-08-17
Rpc simulation with injected accounts

jon-chuang:
#### Problem
Users may want to simulate transactions based on hypothetical data.

For instance, they may like to see the program logs when running governance proposal instructions.
Else, they might like to simulate the effective price when executing a swap on Serum or Raydium.

There are many examples one can think of where this would be useful.

#### Summary of Changes
Implement an RPC->Bank pathway for injecting account data into transaction simulation.

Fixes https://github.com/solana-labs/solana/issues/18746

@ryoqun @jstarry @mvines

### DIFF
--- a/account-decoder/src/lib.rs
+++ b/account-decoder/src/lib.rs
@@ -30,7 +30,7 @@ pub type StringAmount = String;
 pub type StringDecimals = String;
 
 /// A duplicate representation of an Account for pretty JSON serialization
-#[derive(Serialize, Deserialize, Clone, Debug)]
+#[derive(Serialize, Deserialize, PartialEq, Clone, Debug)]
 #[serde(rename_all = "camelCase")]
 pub struct UiAccount {
     pub lamports: u64,
--- a/client/src/rpc_config.rs
+++ b/client/src/rpc_config.rs
@@ -28,7 +28,14 @@ pub struct RpcSendTransactionConfig {
 pub struct RpcSimulateTransactionAccountsConfig {
     pub encoding: Option<UiAccountEncoding>,
     pub addresses: Vec<String>,
-    pub injected_accounts: Option<Vec<(String, UiAccount)>>,
+    pub injected_accounts: Option<Vec<RpcSimulateTransactionInjectedAccount>>,
+}
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct RpcSimulateTransactionInjectedAccount {
+    pub address: String,
+    pub account_data: UiAccount,
 }
 
 #[derive(Debug, Default, Clone, PartialEq, Serialize, Deserialize)]
--- a/client/src/rpc_config.rs
+++ b/client/src/rpc_config.rs
@@ -1,6 +1,6 @@
 use {
     crate::rpc_filter::RpcFilterType,
-    solana_account_decoder::{UiAccountEncoding, UiDataSliceConfig},
+    solana_account_decoder::{UiAccountEncoding, UiAccount, UiDataSliceConfig},
     solana_sdk::{
         clock::{Epoch, Slot},
         commitment_config::{CommitmentConfig, CommitmentLevel},
@@ -28,6 +28,7 @@ pub struct RpcSendTransactionConfig {
 pub struct RpcSimulateTransactionAccountsConfig {
     pub encoding: Option<UiAccountEncoding>,
     pub addresses: Vec<String>,
+    pub injected_accounts: Option<Vec<(String, UiAccount)>>,
 }
 
 #[derive(Debug, Default, Clone, PartialEq, Serialize, Deserialize)]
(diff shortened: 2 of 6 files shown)


mschneider:
This would be really helpful for previewing mango DAO governance proposals.

jon-chuang:
I'm getting the same build error locally, basically due to OOM. Not sure what's the root cause.

jstarry:
@jon-chuang this feature looks really useful but I'm not confident at all that the runtime can handle whatever injected accounts you want without breaking a critical assumption somewhere that causes a panic or OOM.

Here are a few things off the top of my head:
1. Account data is unbounded, could be over 10MB
2. Create or override arbitrary sysvars
3. Create new native programs which can read from a user specified path to dynamically load executables from disk

Maybe we can start with a safe limited set of injection capabilities, but even then we need to be careful.

jon-chuang:
Hi @jstarry, thanks for the feedback.  

I agree. Here are the proposed restrictions:

#### 1. The account data will be smaller or equal to 10MB.
1. I’d prefer if this were enforced at the RPC stage, but then again, RPC nodes are already liable to DoS by unbounded types like `<string>`, aren’t they? 
2. Are there any safeguards that exist at the RPC layer to limit data ingress? If not, would you prefer if overly long data is rejected in the `rpc` processing layer or in the `bank`? Personally, I’d prefer if overly long data never touches the bank, so would side with the former.

#### 2. Forbid injecting executable accounts.

At most, we would have wanted bpf_loader-owned accounts to be injectable. However, `CachedExecutors` would write over injected accounts, so injecting bpf_loader-owned executables would result in inconsistent behaviour.

<details>

> Still, there is a valid use case of testing out programs prior to their being deployed on live data as a final stage of integration testing. Although there could be other ways of doing this though, such as pulling account data into a local test-validator, recreating exact on-chain conditions could be difficult for an ecosystem that is increasingly composed. 
> 
> Thankfully, simulating conditions under which, say, a DAO programs approves a program upgrade, one would allow for simulating instructions executed by an upgraded program. That's because upon program `Update` or `DeployWithMaxLen` by `bpfupgradableloader111...`, the newly created executor is added to the invoke context.
</details>

#### 3. Handle sysvars as per usual
For sysvars, I can’t say I can think of anything particularly bad you can do with them. As far as the runtime is concerned, they are ordinary account data owned by the `SysVar111…` address, so ordinary write-protection rules apply.

In general, I’m not too worried about write-protection since the `simulate_transaction_with_injected_accounts` method never tries to commit any account data.

ryoqun:
(alternatively, support group of txes as a single batch of simulation? so there should only runtime-sanctioned state modifications. but if you're mucking to progress time (including slot/epoch), this can't be used....)

jstarry:
Yeah I like @ryoqun's idea of simulating a **batch** of transactions more than injecting arbitrary account state. Simulation can also skip signer checks which adds to the flexibility. @jon-chuang and @mschneider would that approach be sufficient to cover your desired use-cases?

Account state injection is still probably useful but in my opinion should probably be limited to custom clusters

t-nelson:
> (alternatively, support group of txes as a single batch of simulation? so there should only runtime-sanctioned state modifications. but if you're mucking to progress time (including slot/epoch), this can't be used....)

This could potentially be implemented as an external protocol.  Similar to flashbots on Ethereum

jon-chuang:
> Yeah I like @ryoqun's idea of simulating a **batch** of transactions more than injecting arbitrary account state. Simulation can also skip signer checks which adds to the flexibility. @jon-chuang and @mschneider would that approach be sufficient to cover your desired use-cases?
> 
> Account state injection is still probably useful but in my opinion should probably be limited to custom clusters

Hmm, skipping signer checks could make this direction feasible. For instance, to simulate the result of a majority vote, one could simply look up all of the addresses of the locked governance tokens, and simulate as though a majority had voted a certain way.

Still, implementing this could be rather complex. For one, the RPC node needs to be able to keep the intermediate context alive for the duration of the simulation.

If we're still considering injecting accounts, which could be a much simpler and direct solution in most cases, if you're worried about the attack surface, we could possibly restrict to a more manageable set:
- `data.len()` <= account size limit in RPC decoding layer.
- `data.len()` of accounts remains the same
- `owner` of accounts remain the same.
- we don't allow injecting sysvar addresses, except `Clock`. 
- injected accounts cannot be executable. (the original accounts cannot be executable as well). By definition this excludes all "sensitive" native program addresses.

This restricts injectable accounts to be non-executable, non-sysvar accounts for which only the data (but not the len) and the lamports (and rent_epoch, but that could be gotten rid of too) could be modified.

If it turns out that a use case needs more features than this, the injected accounts can be slowly extended in a conservative fashion in the future.

mschneider:
To simulate the outcome of a vote without signer checks we wouldn’t need to
run the governance program at all during simulation but only the individual
instructions. Which is probably good enough.

Wondering if we’d possibly miss any other signer checks that could prevent
execution of the instruction post vote. I guess it’s tolerable as it merely
makes people approve a not-runnable hence impossible proposal that doesn’t
pose harm.

jon-chuang:
> To simulate the outcome of a vote without signer checks we wouldn’t need to run the governance program at all during simulation but only the individual instructions. Which is probably good enough. Wondering if we’d possibly miss any other signer checks that could prevent execution of the instruction post vote. I guess it’s tolerable as it merely makes people approve a not-runnable hence impossible proposal that doesn’t pose harm.

Hmm yeah, simulation is definitely not foolproof. 

But yeah skipping the governance program entirely would make it possible to give an idea of the execution result without any changes to the runtime. 

Somehow hadn't thought of that. Skipping signer checks opens up a lot of possibilities of simulation I hadn't thought of before. Such as minting arbitrary amount of SPL tokens to an account for testing purposes.

So maybe the changes in this PR aren't necessary. By contrast, allowing for submission of a batch of transactions to be executed sequentially does seem more enticing/have long-term use. 

---

I was thinking that for batching, the execution context is also ephemeral, as it is now, and to limit the number of transactions (to 128), and maintaining a hashmap to store the intermediate state. Thoughts? 

Maybe this still opens up to DoS attacks on RPC nodes since 256 * 10MB * 128 = 327GB. With the new 200+ account format, transaction simulation already opens up a 2.5GB per simulation OOM attack.

PR #19266, opened 2021-08-17
Add ed25519 sig verify program

seanyoung:
#### Summary of Changes

Fixes #

### DIFF
--- a/programs/ed25519/Cargo.toml
+++ b/programs/ed25519/Cargo.toml
@@ -1,0 +1,26 @@
+[package]
+name = "solana-ed25519-program"
+description = "Solana Ed25519 program"
+version = "1.8.0"
+homepage = "https://solana.com/"
+documentation = "https://docs.rs/solana-ed25519-program"
+repository = "https://github.com/solana-labs/solana"
+authors = ["Solana Maintainers <maintainers@solana.foundation>"]
+license = "Apache-2.0"
+edition = "2018"
+
+[dependencies]
+solana-sdk = { path = "../../sdk", version = "=1.8.0" }
+
+[dev-dependencies]
+bytemuck = { version = "1.7.2", features = ["derive"] }
+ed25519-dalek = "=1.0.1"
+rand = "0.7.0"
+solana-logger = { path = "../../logger", version = "=1.8.0" }
+
+[lib]
+crate-type = ["lib"]
+name = "solana_ed25519_program"
+
+[package.metadata.docs.rs]
+targets = ["x86_64-unknown-linux-gnu"]
--- a/sdk/program/src/fee_calculator.rs
+++ b/sdk/program/src/fee_calculator.rs
@@ -1,5 +1,6 @@
 #![allow(clippy::integer_arithmetic)]
 use crate::clock::DEFAULT_MS_PER_SLOT;
+use crate::ed25519_program;
 use crate::message::Message;
 use crate::secp256k1_program;
 use log::*;
@@ -32,20 +33,22 @@ impl FeeCalculator {
         note = "Please do not use, will no longer be available in the future"
     )]
     pub fn calculate_fee(&self, message: &Message) -> u64 {
-        let mut num_secp256k1_signatures: u64 = 0;
+        let mut num_signatures: u64 = 0;
         for instruction in &message.instructions {
             let program_index = instruction.program_id_index as usize;
             // Message may not be sanitized here
             if program_index < message.account_keys.len() {
                 let id = message.account_keys[program_index];
-                if secp256k1_program::check_id(&id) && !instruction.data.is_empty() {
-                    num_secp256k1_signatures += instruction.data[0] as u64;
+                if (secp256k1_program::check_id(&id) || ed25519_program::check_id(&id))
+                    && !instruction.data.is_empty()
+                {
+                    num_signatures += instruction.data[0] as u64;
                 }
             }
         }
 
         self.lamports_per_signature
-            * (u64::from(message.header.num_required_signatures) + num_secp256k1_signatures)
+            * (u64::from(message.header.num_required_signatures) + num_signatures)
     }
 }
 
--- a/sdk/src/ed25519_instruction.rs
+++ b/sdk/src/ed25519_instruction.rs
@@ -1,0 +1,328 @@
+#![cfg(feature = "full")]
+
+use crate::{decode_error::DecodeError, instruction::Instruction};
+use bytemuck::{bytes_of, Pod, Zeroable};
+use ed25519_dalek::{ed25519::signature::Signature, Signer, Verifier};
+use thiserror::Error;
+
+#[derive(Error, Debug, Clone, PartialEq)]
+pub enum Ed25519Error {
+    #[error("ed25519 public key is not valid")]
+    InvalidPublicKey,
+    #[error("ed25519 signature is not valid")]
+    InvalidSignature,
+    #[error("offset not valid")]
+    InvalidDataOffsets,
+    #[error("instruction is incorrect size")]
+    InvalidInstructionDataSize,
+}
+
+impl<T> DecodeError<T> for Ed25519Error {
+    fn type_of() -> &'static str {
+        "Ed25519Error"
+    }
+}
+
+pub const PUBKEY_SERIALIZED_SIZE: usize = 32;
+pub const SIGNATURE_SERIALIZED_SIZE: usize = 64;
+pub const SIGNATURE_OFFSETS_SERIALIZED_SIZE: usize = 14;
+// bytemuck requires structures to be aligned
+pub const SIGNATURE_OFFSETS_START: usize = 2;
+pub const DATA_START: usize = SIGNATURE_OFFSETS_SERIALIZED_SIZE + SIGNATURE_OFFSETS_START;
+
+#[derive(Default, Debug, Copy, Clone, Zeroable, Pod)]
+#[repr(C)]
+pub struct Ed25519SignatureOffsets {
+    signature_offset: u16,             // offset to ed25519 signature of 64 bytes
+    signature_instruction_index: u16,  // instruction index to find signature
+    public_key_offset: u16,            // offset to public key of 32 bytes
+    public_key_instruction_index: u16, // instruction index to find public key
+    message_data_offset: u16,          // offset to start of message data
+    message_data_size: u16,            // size of message data
+    message_instruction_index: u16,    // index of instruction data to get message data
+}
+
+pub fn new_ed25519_instruction(keypair: &ed25519_dalek::Keypair, message: &[u8]) -> Instruction {
+    let signature = keypair.sign(message).to_bytes();
+    let pubkey = keypair.public.to_bytes();
+
+    assert_eq!(pubkey.len(), PUBKEY_SERIALIZED_SIZE);
+    assert_eq!(signature.len(), SIGNATURE_SERIALIZED_SIZE);
+
+    let mut instruction_data = Vec::with_capacity(
+        DATA_START
+            .saturating_add(SIGNATURE_SERIALIZED_SIZE)
+            .saturating_add(PUBKEY_SERIALIZED_SIZE)
+            .saturating_add(message.len()),
+    );
+
+    let num_signatures: u8 = 1;
+    let public_key_offset = DATA_START;
+    let signature_offset = public_key_offset.saturating_add(PUBKEY_SERIALIZED_SIZE);
+    let message_data_offset = signature_offset.saturating_add(SIGNATURE_SERIALIZED_SIZE);
+
+    // add padding byte so that offset structure is aligned
+    instruction_data.extend_from_slice(bytes_of(&[num_signatures, 0]));
+
+    let offsets = Ed25519SignatureOffsets {
+        signature_offset: signature_offset as u16,
+        signature_instruction_index: 0,
+        public_key_offset: public_key_offset as u16,
+        public_key_instruction_index: 0,
+        message_data_offset: message_data_offset as u16,
+        message_data_size: message.len() as u16,
+        message_instruction_index: 0,
+    };
+
+    instruction_data.extend_from_slice(bytes_of(&offsets));
+
+    debug_assert_eq!(instruction_data.len(), public_key_offset);
+
+    instruction_data.extend_from_slice(&pubkey);
+
+    debug_assert_eq!(instruction_data.len(), signature_offset);
+
+    instruction_data.extend_from_slice(&signature);
+
+    debug_assert_eq!(instruction_data.len(), message_data_offset);
+
+    instruction_data.extend_from_slice(message);
+
+    Instruction {
+        program_id: solana_sdk::ed25519_program::id(),
+        accounts: vec![],
+        data: instruction_data,
+    }
+}
+
+pub fn verify_signatures(data: &[u8], instruction_datas: &[&[u8]]) -> Result<(), Ed25519Error> {
+    if data.len() < SIGNATURE_OFFSETS_START {
+        return Err(Ed25519Error::InvalidInstructionDataSize);
+    }
+    let num_signatures = data[0] as usize;
+    if num_signatures == 0 && data.len() > SIGNATURE_OFFSETS_START {
+        return Err(Ed25519Error::InvalidInstructionDataSize);
+    }
+    let expected_data_size = num_signatures
+        .saturating_mul(SIGNATURE_OFFSETS_SERIALIZED_SIZE)
+        .saturating_add(SIGNATURE_OFFSETS_START);
+    if data.len() < expected_data_size {
+        return Err(Ed25519Error::InvalidInstructionDataSize);
+    }
+    for i in 0..num_signatures {
+        let start = i
+            .saturating_mul(SIGNATURE_OFFSETS_SERIALIZED_SIZE)
+            .saturating_add(SIGNATURE_OFFSETS_START);
+        let end = start.saturating_add(SIGNATURE_OFFSETS_SERIALIZED_SIZE);
+
+        // bytemuck wants structures aligned
+        let offsets: &Ed25519SignatureOffsets = bytemuck::try_from_bytes(&data[start..end])
+            .map_err(|_| Ed25519Error::InvalidDataOffsets)?;
+
+        // Parse out signature
+        let signature_index = offsets.signature_instruction_index as usize;
+        if signature_index >= instruction_datas.len() {
+            return Err(Ed25519Error::InvalidDataOffsets);
+        }
+        let signature_instruction = instruction_datas[signature_index];
+        let sig_start = offsets.signature_offset as usize;
+        let sig_end = sig_start.saturating_add(SIGNATURE_SERIALIZED_SIZE);
+        if sig_end >= signature_instruction.len() {
+            return Err(Ed25519Error::InvalidDataOffsets);
+        }
+
+        let signature =
+            ed25519_dalek::Signature::from_bytes(&signature_instruction[sig_start..sig_end])
+                .map_err(|_| Ed25519Error::InvalidSignature)?;
+
+        // Parse out pubkey
+        let pubkey = get_data_slice(
+            instruction_datas,
+            offsets.public_key_instruction_index,
+            offsets.public_key_offset,
+            PUBKEY_SERIALIZED_SIZE,
+        )?;
+
+        let publickey = ed25519_dalek::PublicKey::from_bytes(pubkey)
+            .map_err(|_| Ed25519Error::InvalidPublicKey)?;
+
+        // Parse out message
+        let message = get_data_slice(
+            instruction_datas,
+            offsets.message_instruction_index,
+            offsets.message_data_offset,
+            offsets.message_data_size as usize,
+        )?;
+
+        publickey
+            .verify(message, &signature)
+            .map_err(|_| Ed25519Error::InvalidSignature)?;
+    }
+    Ok(())
+}
+
+fn get_data_slice<'a>(
+    instruction_datas: &'a [&[u8]],
+    instruction_index: u16,
+    offset_start: u16,
+    size: usize,
+) -> Result<&'a [u8], Ed25519Error> {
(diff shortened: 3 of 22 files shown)


seanyoung:
The tests keep on failing in CI at different points, even though the same tests work locally.

seanyoung:
Should this feature be gated?

jackcmay:
> Should this feature be gated?

Yes

seanyoung:
I've added `feature_set::ed25519_program_enabled::id()`

programs/ed25519/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+repository = "https://github.com/solana-labs/solana"
+authors = ["Solana Maintainers <maintainers@solana.foundation>"]
+license = "Apache-2.0"
+edition = "2018"
+
+[dependencies]
+solana-sdk = { path = "../../sdk", version = "=1.8.0" }
+
+[dev-dependencies]
+bincode = "1.3.3"
```
jackcmay, line 16: Would borsh be more efficient?

sdk/src/ed25519_instruction.rs
```diff
@@ -0,0 +1,287 @@
+#![cfg(feature = "full")]
+
+use crate::instruction::Instruction;
+use ed25519_dalek::{ed25519::signature::Signature, Signer, Verifier};
+use serde_derive::{Deserialize, Serialize};
+
+#[derive(Debug, PartialEq)]
+pub enum Ed25519Error {
```
jackcmay, line 8: `thiserror` is helpful and what we've using on recently added error types

sdk/src/ed25519_instruction.rs
```diff
@@ -0,0 +1,287 @@
...
+}
+
+pub fn new_ed25519_instruction(keypair: &ed25519_dalek::Keypair, message: &[u8]) -> Instruction {
+    let signature = keypair.sign(message).to_bytes();
+    let pubkey = keypair.public.to_bytes();
+
+    assert_eq!(pubkey.len(), PUBKEY_SERIALIZED_SIZE);
+    assert_eq!(signature.len(), SIGNATURE_SERIALIZED_SIZE);
+
+    let mut instruction_data = vec![];
```
jackcmay, line 38: You could use `with_capacity()` here?

sdk/src/ed25519_instruction.rs
```diff
@@ -0,0 +1,287 @@
...
+    instruction_data.resize(
+        DATA_START
+            .saturating_add(signature.len())
+            .saturating_add(pubkey.len())
+            .saturating_add(message.len())
+            .saturating_add(1),
```
jackcmay, line 44: magic number "1"?

sdk/src/ed25519_instruction.rs
```diff
@@ -0,0 +1,287 @@
...
+    public_key_offset: u16,           // offset to public key of 32 bytes
+    public_key_instruction_index: u8, // instruction index to find public key
+    message_data_offset: u16,         // offset to start of message data
+    message_data_size: u16,           // size of message data
+    message_instruction_index: u8,    // index of instruction data to get message data
+}
+
+pub fn new_ed25519_instruction(keypair: &ed25519_dalek::Keypair, message: &[u8]) -> Instruction {
+    let signature = keypair.sign(message).to_bytes();
+    let pubkey = keypair.public.to_bytes();
```
jackcmay, line 33: Public key is variable lenght?

sdk/src/ed25519_instruction.rs
```diff
@@ -0,0 +1,287 @@
...
+
+    assert_eq!(pubkey.len(), PUBKEY_SERIALIZED_SIZE);
+    assert_eq!(signature.len(), SIGNATURE_SERIALIZED_SIZE);
+
+    let mut instruction_data = vec![];
+    instruction_data.resize(
+        DATA_START
+            .saturating_add(signature.len())
```
jackcmay, line 41: signature and pubkey are variable lengths?

sdk/src/ed25519_instruction.rs
```diff
@@ -0,0 +1,287 @@
...
+    instruction_data[public_key_offset..public_key_offset.saturating_add(pubkey.len())]
+        .copy_from_slice(&pubkey);
+
+    let signature_offset = DATA_START.saturating_add(pubkey.len());
+    instruction_data[signature_offset..signature_offset.saturating_add(signature.len())]
+        .copy_from_slice(&signature);
+
+    let message_data_offset = signature_offset
+        .saturating_add(signature.len())
+        .saturating_add(1);
```
jackcmay, line 58: magic number "1"

sdk/src/ed25519_instruction.rs
```diff
@@ -0,0 +1,287 @@
...
+    instruction_data[message_data_offset..].copy_from_slice(message);
```
jackcmay, line 59: These are guarenteed to be the same length?

sdk/src/ed25519_instruction.rs
```diff
@@ -0,0 +1,287 @@
...
+    if data.is_empty() {
+        return Err(Ed25519Error::InvalidInstructionDataSize);
+    }
+    let count = data[0] as usize;
+    if count == 0 && data.len() > 1 {
+        return Err(Ed25519Error::InvalidInstructionDataSize);
+    }
+    let expected_data_size = count
+        .saturating_mul(SIGNATURE_OFFSETS_SERIALIZED_SIZE)
+        .saturating_add(1);
```
jackcmay, line 92: "magic number"

sdk/src/ed25519_instruction.rs
```diff
@@ -0,0 +1,287 @@
...
+
+pub const PUBKEY_SERIALIZED_SIZE: usize = 32;
+pub const SIGNATURE_SERIALIZED_SIZE: usize = 64;
+pub const SIGNATURE_OFFSETS_SERIALIZED_SIZE: usize = 11;
+pub const DATA_START: usize = SIGNATURE_OFFSETS_SERIALIZED_SIZE + 1;
+
+#[derive(Default, Serialize, Deserialize, Debug)]
+pub struct Ed25519SignatureOffsets {
+    signature_offset: u16,            // offset to ed25519 signature of 64 bytes
+    signature_instruction_index: u8,  // instruction index to find signature
```
jackcmay, line 23: Know why the "instruction" in "signature_instruction_index"?  Couldn't it just be "signature_index"?

sdk/src/feature_set.rs
```diff
@@ -237,7 +241,8 @@ lazy_static! {
         (gate_large_block::id(), "validator checks block cost against max limit in realtime, reject if exceeds."),
         (mem_overlap_fix::id(), "memory overlap fix"),
         (versioned_tx_message_enabled::id(), "enable versioned transaction message processing"),
-        (libsecp256k1_fail_on_bad_count::id(), "Fail libsec256k1_verify if count appears wrong")
+        (libsecp256k1_fail_on_bad_count::id(), "Fail libsec256k1_verify if count appears wrong"),
```
jackcmay, line 244: Can you lowercase the "F" at the beginning of this line please, must have missed it in a previous change.

sdk/src/ed25519_instruction.rs
```diff
@@ -0,0 +1,287 @@
...
+    bincode::serialize_into(writer, &offsets).unwrap();
+
+    Instruction {
+        program_id: solana_sdk::ed25519_program::id(),
+        accounts: vec![],
+        data: instruction_data,
+    }
+}
+
+pub fn verify_signatures(data: &[u8], instruction_datas: &[&[u8]]) -> Result<(), Ed25519Error> {
```
mvines, line 82: I'll be experimenting with the [bytemuck](https://crates.io/crates/bytemuck) crate a little tomorrow, but if I'm reading the docs.rs right I think we might be able to use that crate here as well to reduce the copies (maybe not quite zero copy) and drop the bincode

sdk/src/ed25519_instruction.rs
seanyoung: Renamed `Ed25519Error` to `ThisError`

sdk/src/ed25519_instruction.rs
seanyoung: Replaced with constants

sdk/src/ed25519_instruction.rs
seanyoung: replaced with `vec![0u8, len]`

sdk/src/ed25519_instruction.rs
seanyoung: Well, they should be. I've added `message_data_offset + message.len()` to be sure

sdk/src/ed25519_instruction.rs
seanyoung: Replaced with constants

sdk/src/ed25519_instruction.rs
seanyoung: Not necessary

sdk/src/feature_set.rs
seanyoung: done

sdk/src/ed25519_instruction.rs
seanyoung: Not necessary, removed.

sdk/src/ed25519_instruction.rs
seanyoung: Removed

programs/ed25519/Cargo.toml
seanyoung: I've left it as for now while we decided on bytemuck (I like the name btw)

sdk/program/src/pubkey.rs
```diff
@@ -379,6 +379,7 @@ impl Pubkey {
             stake::config::id(),
             vote::program::id(),
             secp256k1_program::id(),
+            ed25519_program::id(),
```
mvines, line 382: Oof, so I think this change is a consensus footgun.  We probably need a big warning at the top of `is_native_program_id()` that says, DON'T ADD NEW IDs.

The problem is that the `sol_try_find_program_address()` syscall invokes `create_program_address`, which invokes `is_native_program_id()`.  Since there's no feature gating in `is_native_program_id()`, I could fork nodes off the cluster by invoking `sol_try_find_program_address()` from a program with `ed25519_program::id()` as nodes with/without this line will behave differently.  Time it right and you can take out the entire cluster

sdk/src/transaction/mod.rs
```diff
@@ -452,6 +453,16 @@ impl Transaction {
                     libsecp256k1_fail_on_bad_count,
                 );
                 e.map_err(|_| TransactionError::InvalidAccountIndex)?;
+            } else if crate::ed25519_program::check_id(program_id) {
```
mvines, line 456: There's a similar consensus issue here as well.  Since `FeatureSet` is not available, we'll need new parameter into `verify_precompiles()` that indicates whether  `ed25519_program` has been enabled yet or not.  

It might be better to just actually pass `FeatureSet` into `verify_precompiles()`, perhaps as a separate PR that lands before this one.  More context: soon after this PR lands I'll likely be adding another precompile program in here too!

sdk/src/ed25519_instruction.rs
```diff
@@ -0,0 +1,329 @@
+#![cfg(feature = "full")]
+
+use crate::{decode_error::DecodeError, instruction::Instruction};
+use bytemuck::{bytes_of, Pod, Zeroable};
+use ed25519_dalek::{ed25519::signature::Signature, Signer, Verifier};
+use serde_derive::{Deserialize, Serialize};
+use thiserror::Error;
+
+#[derive(Error, Debug, Serialize, Clone, PartialEq)]
```
mvines, line 9: Do we need `Serialize` here at all?

sdk/src/ed25519_instruction.rs
```diff
@@ -0,0 +1,329 @@
...
+    }
+}
+
+pub const PUBKEY_SERIALIZED_SIZE: usize = 32;
+pub const SIGNATURE_SERIALIZED_SIZE: usize = 64;
+pub const SIGNATURE_OFFSETS_SERIALIZED_SIZE: usize = 14;
+pub const NUM_SIGNATURES_SIZE: usize = 2;
+pub const DATA_START: usize = SIGNATURE_OFFSETS_SERIALIZED_SIZE + NUM_SIGNATURES_SIZE;
+
+#[derive(Default, Debug, Deserialize, Copy, Clone, Zeroable, Pod)]
```
mvines, line 33: `Deserialize` can be removed now that bincode has been given the 👢

seanyoung: Indeed it can, removed

sdk/src/ed25519_instruction.rs
seanyoung: No, we don't. Removed.

sdk/src/transaction/mod.rs
seanyoung: I've added another commit to do exactly this.

sdk/program/src/pubkey.rs
seanyoung: I'm glad you caught that before we had a disaster on our hands :pray: 

So I've started working on a solution where so`sol_try_find_program_address()` and `sol_create_program_address()` use the invoke context list of native programs to check. This requires `Pubkey::create_program_address()` to not check `Pubkey::is_native_program_id()` (which is unchanged). I haven't quite settled on this yet, and it needs more tests, but would you mind giving it a quick look and see what you think (last commit in PR now)

sdk/src/transaction/mod.rs
seanyoung: The commits should really be re-ordered.

sdk/src/transaction/sanitized.rs
```diff
@@ -220,10 +219,22 @@ impl SanitizedTransaction {
                 let e = verify_eth_addresses(
                     data,
                     &instruction_datas,
-                    libsecp256k1_0_5_upgrade_enabled,
-                    libsecp256k1_fail_on_bad_count,
+                    feature_set.is_active(&feature_set::libsecp256k1_0_5_upgrade_enabled::id()),
+                    feature_set.is_active(&feature_set::libsecp256k1_fail_on_bad_count::id()),
                 );
                 e.map_err(|_| TransactionError::InvalidAccountIndex)?;
+            } else if feature_set.is_active(&feature_set::ed25519_program_enabled::id())
```
mvines, line 226: I think it makes sense to check `crate::ed25519_program::check_id(program_id)` first.  This check will be performed for every single instruction that is ever executed, and short-circuiting the `&&` on the pubkey equality check before the `feature_set.is_active()` check should be faster

sdk/src/transaction/mod.rs
```diff
@@ -448,10 +447,22 @@ impl Transaction {
                 let e = verify_eth_addresses(
                     data,
                     &instruction_datas,
-                    libsecp256k1_0_5_upgrade_enabled,
-                    libsecp256k1_fail_on_bad_count,
+                    feature_set.is_active(&feature_set::libsecp256k1_0_5_upgrade_enabled::id()),
+                    feature_set.is_active(&feature_set::libsecp256k1_fail_on_bad_count::id()),
                 );
                 e.map_err(|_| TransactionError::InvalidAccountIndex)?;
+            } else if feature_set.is_active(&feature_set::ed25519_program_enabled::id())
```
mvines, line 454: Same comment as https://github.com/solana-labs/solana/pull/19266/files#r699066382 here

sdk/program/src/pubkey.rs
mvines: The `create_program_address_program_id_checked()` approach looks good to me.  It'd be helpful to add a function comment to `create_program_address_program_id_checked()` that describes when it's safe to use this function over the original

sdk/src/transaction/sanitized.rs
seanyoung: Yes, good point.

There is something to be said for the feature set to be reduced down to a bit set rather than comparing 32 bytes every time. Having said that check_id compiles down to:
```
_ZN10solana_sdk13native_loader8check_id17hee92c47a82a05162E:
        .cfi_startproc
        movdqu  (%rdi), %xmm0
        movdqu  16(%rdi), %xmm1
        pcmpeqb .LCPI659_0(%rip), %xmm1
        pcmpeqb .LCPI659_1(%rip), %xmm0
        pand    %xmm1, %xmm0
        pmovmskb        %xmm0, %eax
        cmpl    $65535, %eax
        sete    %al
        retq
.Lfunc_end659:
```
So should be pretty quick.

seanyoung: Fixed

sdk/src/transaction/mod.rs
seanyoung: Fixed

sdk/program/src/pubkey.rs
seanyoung: So I've cleaned up the code and it's ready for review. However, I think it would be nice to have a test that checks that `sol_try_find_program_address()` and `sol_create_program_address()` work with the ed25519 program id when the feature is not enabled. Is there an easy way to do this or should I just leave it?

Note that `Pubkey::create_program_address()` is still used by CPI, but since the caller can not be the ed25519 program, I assume this is safe.

sdk/src/transaction/sanitized.rs
mvines: > There is something to be said for the feature set to be reduced down to a bit set rather than comparing 32 bytes every time

Definitely.  Optimising `FeatureSet::is_active()` has been on my wish list for a while.  Thanks for the assembly :)

sdk/program/src/pubkey.rs
mvines: I think the latest commits look good, @jackcmay can you please carefully review as well: `syscalls.rs` and `pubkey.rs`

jackcmay: Yup, on my list for today :-)

programs/bpf_loader/src/syscalls.rs
```diff
@@ -857,6 +854,14 @@ impl<'a> SyscallObject<BpfError> for SyscallCreateProgramAddress<'a> {
         memory_mapping: &MemoryMapping,
         result: &mut Result<u64, EbpfError<BpfError>>,
     ) {
+        let invoke_context = match self.invoke_context.try_borrow() {
```
jackcmay, line 857: How about:
```suggestion
    let invoke_context = invoke_context
        .try_borrow()
        .map_err(|_| SyscallError::InvokeContextBorrowFailed)?;
```

seanyoung: This function doesn't return anything, the result goes into the `result` argument.

seanyoung: The other option is:
```
 let invoke_context = question_mark!(invoke_context
        .try_borrow()
        .map_err(|_| SyscallError::InvokeContextBorrowFailed), result);
```

seanyoung: That might be better.

seanyoung: Fixed now

sdk/program/src/pubkey.rs
jackcmay: I don't think the native program id checks in this pr are correct.  `sol_try_find_program_address()` and `sol_create_program_address` is enforcing that the program id is not one of the native programs ids but that is not the same as built-in programs.  For example, ids like `feature::id()` and `sysvar::id()` are not included in the builtins.

I think the existing native program id checks in pubkey are not ideal, it creates another place where we have to add a program id when programs are added and can easily be missed (like it was when I added `compute_budget_program`).  I'm not sure yet what the better solution looks like, at the very least the program id check in the syscalls should include the missing ids in addition to the builtins.  I'd rather not but seems like we need a third list where `is_native_program_id` = `builtins` + non-builtin-ids.

I also think we should make these changes in its own PR so folks can review them on their own.

programs/bpf_loader/src/syscalls.rs
jackcmay: Yup, that is what I meant, thanks for the fix :-)

sdk/program/src/pubkey.rs
jackcmay: I'm putting together a pr to address this, let me know if you have any thoughts

seanyoung: I was thinking we should make an explicit allow-list of native programs for these two functions at vm creation time. This both makes it very explicit and also means the list is empty when all features are enabled, making it performant.

seanyoung: Fixed

jackcmay (dismissed):

seanyoung:
@jackcmay I had to push some changes because the documentation changes still had `u8` where they should have been `u16`

sdk/program/src/fee_calculator.rs
```diff
@@ -32,20 +33,22 @@ impl FeeCalculator {
...
             let program_index = instruction.program_id_index as usize;
             // Message may not be sanitized here
             if program_index < message.account_keys.len() {
                 let id = message.account_keys[program_index];
-                if secp256k1_program::check_id(&id) && !instruction.data.is_empty() {
-                    num_secp256k1_signatures += instruction.data[0] as u64;
+                if (secp256k1_program::check_id(&id) || ed25519_program::check_id(&id))
+                    && !instruction.data.is_empty()
+                {
+                    num_signatures += instruction.data[0] as u64;
```
jackcmay, line 45: This is at best confusing since the count is represented as a u16.  Might be more straightforward and simple to go back to a u8 count as the first byte, a padding byte, and then the bytemuck serialized data?  Unless you think a u8 would overflow at some point in the future

seanyoung: Good catch.

seanyoung: Fixed. I've added a single padding byte.

jackcmay (dismissed):

PR #19269, opened 2021-08-17
--accounts-index-memory-limit-mb

jeffwashington:
#### Problem
As we move to a disk-based accounts index, we will want to allow a budget for the in-memory portion of the index.
This will get hooked up to accounts index, which will be all in memory initially (as it is now).
#### Summary of Changes
Allow a validator option for mb budget for accounts index.
atm, missing means 'all in memory'. Any limit will soon cause us to put accounts on disk. Later this option will be more precise in its tuning.
Fixes #


brooksprumo (approved):
lgtm

Issue #16146, opened 2021-03-25
RPC: make nonCirculatingAccounts optional on getSupply request


oJshua:
#### Problem
The front page of Explorer now requires a getSupply RPC call, and it would be helpful (in terms of payload size) if we could optionally omit nonCirculatingAccounts from the response.

#### Proposed Solution
Introduce an argument that prevents nonCirculatingAccounts from being returned.

PR #19270, opened 2021-08-17
RPC: add option to exclude accounts from get_supply

jstarry:
#### Problem
Explorer home page fetches circulating supply stats on load but the perf is really bad because the full list of non-circulating accounts is returned. 

#### Summary of Changes
- Add `excludeNonCirculatingAccountsList` param to `get_supply` RPC method

Fixes https://github.com/solana-labs/solana/issues/16146


CriesofCarrots (approved):
lgtm, thanks!

PR #19271, opened 2021-08-17
Make AccountsHashVerifier aware of Incremental Snapshots

### LINKED ISSUE #19167: Make AccountsHashVerifier make aware of Incremental Snapshots
`AccountsHashVerifier` needs to know if it should generate a full snapshot package or an incremental snapshot package when sending over to `SnapshotPackagerService`.

brooksprumo:
#### Problem

AccountsHashVerifier does not know to take Incremental Snapshots

#### Summary of Changes

AccountsHashVerifier now creates SnapshotPackages that are either Full or Incremental Snapshots depending on the SnapshotConfig's full/incremental snapshot interval.

Additionally:

- Move some fields to SnapshotPackage and remove from AccountsPackage
- Move snapshot storages filtering into SnapshotPackage, since the AccountsPackage doesn't know
- SnapshotPackage knows about full snapshots vs incremental snapshots, and AccountsPackage no longer does
- Remove the "processing" of AccountsPackage, as (1) it only conditional verifies the hash, and (2), the verification is only done in AccountsHashVerifier
- Refactor AccountsHashVerifier
- Update snapshots background services test to now take/expect incremental snapshots

Fixes #19167

_NOTE: This PR is on the large side, due to the renaming/refactoring. I'm happy to walk through/explain more anything/everything here! Thanks in advance!_

### DIFF
--- a/core/src/accounts_hash_verifier.rs
+++ b/core/src/accounts_hash_verifier.rs
@@ -15,7 +15,6 @@ use solana_runtime::{
         AccountsPackage, AccountsPackageReceiver, PendingSnapshotPackage, SnapshotPackage,
         SnapshotType,
     },
-    snapshot_utils,
     sorted_storages::SortedStorages,
 };
 use solana_sdk::{clock::Slot, hash::Hash, pubkey::Pubkey};
@@ -206,38 +205,39 @@ impl AccountsHashVerifier {
                     % snapshot_config.incremental_snapshot_archive_interval_slots
                     == 0
                 {
-                    let last_full_snapshot_slot =
-                        snapshot_utils::get_highest_full_snapshot_archive_slot(
-                            &snapshot_config.snapshot_package_output_path,
-                        );
+                    // Conditionally submit an incremental snapshot package for archiving.
+                    // Submit the package _unless_ the pending snapshot package contains a
+                    // full snapshot package.
+                    let can_submit = pending_snapshot_package.lock().unwrap().as_ref().map_or(
+                        true,
+                        |snapshot_package| {
+                            snapshot_package.snapshot_type == SnapshotType::IncrementalSnapshot
+                        },
+                    );
 
-                    match last_full_snapshot_slot {
-                        None => warn!("No full snapshot archive found; will not process incremental snapshot package!"),
-                        Some(incremental_snapshot_base_slot) => {
-                            let snapshot_package = SnapshotPackage::new(
-                                accounts_package,
-                                snapshot_config.snapshot_package_output_path.clone(),
-                                snapshot_config.archive_format,
-                                Some(incremental_snapshot_base_slot),
-                            );
+                    if can_submit {
+                        // NOTE: Read the value of the last full snapshot slot _after_ checking
+                        // `can_submit`.  This way we know a full snapshot is not in the middle of
+                        // being archived, and therefore the value of the last full snapshot slot
+                        // is correct.
+                        let last_full_snapshot_slot =
+                            *snapshot_config.last_full_snapshot_slot.read().unwrap();
+                        match last_full_snapshot_slot {
+                            None => warn!("No full snapshot archive found; will not process incremental snapshot package!"),
+                            Some(incremental_snapshot_base_slot) => {
+                                let snapshot_package = SnapshotPackage::new(
+                                    accounts_package,
+                                    snapshot_config.snapshot_package_output_path.clone(),
+                                    snapshot_config.archive_format,
+                                    Some(incremental_snapshot_base_slot),
+                                );
+                                *pending_snapshot_package.lock().unwrap() = Some(snapshot_package);
 
-                            let mut pending_snapshot_package = pending_snapshot_package.lock().unwrap();
-
-                            // Conditionally submit an incremental snapshot package for archiving.
-                            // Submit the package _unless_ the pending snapshot package contains a
-                            // full snapshot package.
-                            let can_submit = pending_snapshot_package.as_ref().map_or(true, |pending_snapshot_package| {
-                                pending_snapshot_package.snapshot_type == SnapshotType::IncrementalSnapshot
-                            });
-
-                            if can_submit {
-                                *pending_snapshot_package = Some(snapshot_package);
                             }
-                            else {
-                                info!("Not submitting incremental snapshot for archiving; there is still a pending full snapshot package waiting to be archived.");
-                            }
-                        }
-                    };
+                        };
+                    } else {
+                        info!("Not submitting incremental snapshot for archiving; there is still a pending full snapshot package waiting to be archived.");
+                    }
                 }
             };
         };
@@ -370,6 +370,7 @@ mod tests {
             archive_format: ArchiveFormat::Tar,
             snapshot_version: SnapshotVersion::default(),
             maximum_snapshots_to_retain: usize::MAX,
+            last_full_snapshot_slot: Default::default(),
         };
         for i in 0..MAX_SNAPSHOT_HASHES + 1 {
             let accounts_package = AccountsPackage {
--- a/core/src/accounts_hash_verifier.rs
+++ b/core/src/accounts_hash_verifier.rs
@@ -6,14 +6,16 @@
 
 use rayon::ThreadPool;
 use solana_gossip::cluster_info::{ClusterInfo, MAX_SNAPSHOT_HASHES};
+use solana_measure::measure::Measure;
 use solana_runtime::{
-    accounts_db,
-    snapshot_archive_info::SnapshotArchiveInfoGetter,
+    accounts_db::{self, AccountsDb},
+    accounts_hash::HashStats,
     snapshot_config::SnapshotConfig,
     snapshot_package::{
         AccountsPackage, AccountsPackageReceiver, PendingSnapshotPackage, SnapshotPackage,
     },
     snapshot_utils,
+    sorted_storages::SortedStorages,
 };
 use solana_sdk::{clock::Slot, hash::Hash, pubkey::Pubkey};
 use std::collections::{HashMap, HashSet};
@@ -39,7 +41,7 @@ impl AccountsHashVerifier {
         cluster_info: &Arc<ClusterInfo>,
         trusted_validators: Option<HashSet<Pubkey>>,
         halt_on_trusted_validators_accounts_hash_mismatch: bool,
-        fault_injection_rate_slots: u64,
+        fault_injection_rate_slots: Slot,
         snapshot_config: Option<SnapshotConfig>,
     ) -> Self {
         let exit = exit.clone();
@@ -48,7 +50,7 @@ impl AccountsHashVerifier {
             .name("solana-hash-accounts".to_string())
             .spawn(move || {
                 let mut hashes = vec![];
-                let mut thread_pool_storage = None;
+                let mut thread_pool = None;
                 loop {
                     if exit.load(Ordering::Relaxed) {
                         break;
@@ -56,11 +58,9 @@ impl AccountsHashVerifier {
 
                     match accounts_package_receiver.recv_timeout(Duration::from_secs(1)) {
                         Ok(accounts_package) => {
-                            if accounts_package.hash_for_testing.is_some()
-                                && thread_pool_storage.is_none()
+                            if accounts_package.hash_for_testing.is_some() && thread_pool.is_none()
                             {
-                                thread_pool_storage =
-                                    Some(accounts_db::make_min_priority_thread_pool());
+                                thread_pool = Some(accounts_db::make_min_priority_thread_pool());
                             }
 
                             Self::process_accounts_package(
@@ -73,7 +73,7 @@ impl AccountsHashVerifier {
                                 &exit,
                                 fault_injection_rate_slots,
                                 snapshot_config.as_ref(),
-                                thread_pool_storage.as_ref(),
+                                thread_pool.as_ref(),
                             );
                         }
                         Err(RecvTimeoutError::Disconnected) => break,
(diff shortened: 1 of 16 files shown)


core/src/accounts_hash_verifier.rs
```diff
@@ -155,17 +179,53 @@ impl AccountsHashVerifier {
...
         }
 
+        cluster_info.push_accounts_hashes(hashes.clone());
+    }
+
+    fn handle_snapshot(
+        accounts_package: AccountsPackage,
+        pending_snapshot_package: Option<&PendingSnapshotPackage>,
+        snapshot_config: Option<&SnapshotConfig>,
+    ) {
```
brooksprumo, line 189: This function has the code for now doing an Incremental Snapshot.

A nice thing that came out of the changes, is now we don't unconditionally make a `SnapshotPackage`. Instead, only when it's time. This should save a bit of time copying/moving data around. And in the case of an Incremental Snapshot, this is where the snapshot storages filtering takes place, and not anywhere else.

core/src/accounts_hash_verifier.rs
```diff
@@ -96,49 +96,73 @@ impl AccountsHashVerifier {
...
-            snapshot_config,
+        );
+
+        Self::handle_snapshot(accounts_package, pending_snapshot_package, snapshot_config);
+    }
+
+    fn verify_accounts_package_hash(
+        accounts_package: &AccountsPackage,
+        thread_pool: Option<&ThreadPool>,
+    ) {
```
brooksprumo, line 121: This function used to basically be the code inside `snapshot_utils::process_accounts_package()`, but it only really did anything for AccountsHashVerifier, since `expected_hash` was `None` for all other uses. So now it only lives here :)

runtime/src/snapshot_package.rs
```diff
@@ -44,28 +36,31 @@ pub struct AccountsPackage {
     pub snapshot_links: TempDir,
     pub storages: SnapshotStorages,
     pub hash: Hash, // temporarily here while we still have to calculate hash before serializing bank
-    pub archive_format: ArchiveFormat,
     pub snapshot_version: SnapshotVersion,
-    pub snapshot_output_dir: PathBuf,
```
brooksprumo, line 49: I was able to remove these two fields, since only `SnapshotPackage` needs them. This will be a small perf boost to not copy a path into here each time, and instead only do it when actually making a SnapshotPackage.

runtime/src/snapshot_package.rs
```diff
@@ -85,100 +80,12 @@ impl AccountsPackage {
...
-            snapshot_output_dir: snapshot_package_output_path.as_ref().to_path_buf(),
             expected_capitalization: bank.capitalization(),
             hash_for_testing,
             cluster_type: bank.cluster_type(),
         })
     }
-
-    /// Package up bank snapshot files, snapshot storages, and slot deltas for a full snapshot.
-    #[allow(clippy::too_many_arguments)]
-    pub fn new_for_full_snapshot(
```
brooksprumo, line 99: Can get rid of these two `new_for_full/incremental_snapshot()` functions. `AccountsPackage` no longer knows about full vs incremental snapshots, just `SnapshotPackage`.

runtime/src/snapshot_package.rs
```diff
@@ -192,31 +99,50 @@ pub struct SnapshotPackage {
...
+                    snapshot_archives_dir,
+                    incremental_snapshot_base_slot,
+                    accounts_package.slot,
+                    &accounts_package.hash,
+                    archive_format,
+                );
+                snapshot_utils::filter_snapshot_storages_for_incremental_snapshot(
+                    &mut accounts_package.storages,
+                    incremental_snapshot_base_slot,
+                );
```
brooksprumo, line 129: Here's where the storage filtering happens for incremental snapshots.

core/src/accounts_hash_verifier.rs
```diff
@@ -155,17 +179,53 @@ impl AccountsHashVerifier {
...
+                        None => warn!("No full snapshot archive found; will not process incremental snapshot package!"),
+                        Some(incremental_snapshot_base_slot) => {
+                            let snapshot_package = SnapshotPackage::new(
+                                accounts_package,
+                                snapshot_config.snapshot_package_output_path.clone(),
+                                snapshot_config.archive_format,
+                                Some(incremental_snapshot_base_slot),
+                            );
+
+                            *pending_snapshot_package.lock().unwrap() = Some(snapshot_package);
```
carllin, line 223: If the last full snapshot hasn't been ingested yet, maybe we don't want to overwrite it with an incremental snapshot?

core/src/accounts_hash_verifier.rs
```diff
@@ -155,17 +179,53 @@ impl AccountsHashVerifier {
...
                     *pending_snapshot_package.lock().unwrap() = Some(snapshot_package);
-                }
-            }
-        }
+                } else if accounts_package.block_height
+                    % snapshot_config.incremental_snapshot_archive_interval_slots
+                    == 0
+                {
+                    let last_full_snapshot_slot =
+                        snapshot_utils::get_highest_full_snapshot_archive_slot(
```
carllin, line 209: Is there a race here if the packaging service is *just* about to finish packaging the next full snapshot, but we read off the old last full snapshot?

core/src/accounts_hash_verifier.rs
```diff
@@ -155,17 +179,53 @@ impl AccountsHashVerifier {
             }
         }
 
+        cluster_info.push_accounts_hashes(hashes.clone());
+    }
+
+    fn handle_snapshot(
```
carllin, line 185: nit: maybe rename to `queue_snapshot_for_packaging()`

runtime/src/snapshot_package.rs
carllin: hmmm I feel like this is a little late in the pipeline, i.e. there was never a need for AccountsBackgroundService to waste the resources cloning out the unnecessary storages (which can be quite a few) only for them to be filtered out here. 

Seems more efficient to have the `AccountsBackgroundService` rather than the `AccountsHashVerifier` determine the incremental vs full snapshot and only clone + pass on those storages that are necessary.

brooksprumo: Yeah, cloning the unnecessary storages just to filter them out later is a bummer. But, the only way to fix that is to change `AccountsDb::get_snapshot_storages()`, which is where we do the cloning.

For the changes here, this just moves where the filtering is, so there isn't a change in how much gets cloned. Maybe I should open a new issue to update `AccountsDb::get_snapshot_storages()` to just return snapshot storages above a slot? What do you think?

core/src/accounts_hash_verifier.rs
brooksprumo: Eek! Yeah I agree. Will fix!

core/src/accounts_hash_verifier.rs
brooksprumo: Yeah... 

Since SnapshotPackagerService is the only place that knows if `snapshot_utils::archive_snapshot_package()` succeeds or not, I think I need to use that result to pass back info to AccountsHashVerifier to get the _true_ last full snapshot slot. This will also be useful for AccountsBackgroundService, because that's where `clean_accounts()` gets called, and it needs the last full snapshot slot as well.

Back to AccountsHashVerifier here, luckily this race doesn't produce broken incremental snapshots, just ones that are too big (because the base slot could've been higher).

core/src/accounts_hash_verifier.rs
brooksprumo: Done!

core/src/accounts_hash_verifier.rs
brooksprumo: Done!

core/src/accounts_hash_verifier.rs
brooksprumo: Fixed!

core/src/accounts_hash_verifier.rs
```diff
@@ -39,7 +41,7 @@ impl AccountsHashVerifier {
         cluster_info: &Arc<ClusterInfo>,
         trusted_validators: Option<HashSet<Pubkey>>,
         halt_on_trusted_validators_accounts_hash_mismatch: bool,
-        fault_injection_rate_slots: u64,
+        fault_injection_rate_slots: Slot,
```
lijunwangs, line 44: I feel better to leave this u64 -- it is the rate of slots, right -- fundamentally a number?

core/src/accounts_hash_verifier.rs
```diff
@@ -96,49 +96,73 @@ impl AccountsHashVerifier {
...
+        measure_hash.stop();
+        datapoint_info!(
+            "accounts_hash_verifier",
+            ("calculate_hash", measure_hash.as_us(), i64),
         );
     }
 
-    fn process_snapshot_package(
-        snapshot_package: SnapshotPackage,
+    fn push_hashes_to_cluster(
```
lijunwangs, line 144: push_accounts_hashes_to_cluster is clearer

core/src/accounts_hash_verifier.rs
brooksprumo: Yeah, I agree. Ideally I'd like this to be some sort of difference type too.

Since Rust doesn't let you do implicit conversions within arithmetic, and the actual slot is compared to the interval, one of the two needs to be cast to the other. I think some of the older code uses `u64` instead of `Slot` likely before the `Slot` alias existed. It seems other parts of the code uses `Slot` for the various rates as well, so I was trying to move in that direction here. Like here:

https://github.com/solana-labs/solana/blob/f59a55be171e64aa7deb021cad756ca3c69968cc/runtime/src/snapshot_config.rs#L8-L13

I'm fine using any type here, but I would like to understand what the overall best practice is/should be for this codebase, so I don't further bifurcate it :)

@carllin What do you think about this?

core/src/accounts_hash_verifier.rs
```diff
@@ -155,17 +179,68 @@ impl AccountsHashVerifier {
...
+                        snapshot_config.archive_format,
+                        None,
+                    );
+
                     *pending_snapshot_package.lock().unwrap() = Some(snapshot_package);
-                }
-            }
-        }
+                } else if accounts_package.block_height
+                    % snapshot_config.incremental_snapshot_archive_interval_slots
```
lijunwangs, line 205: I am not sure how to configure when the snapshots is taken. I think it would be simpler to say that we take a snapshot for every X amount of intervals. And among these we take a full snapshot for every Y snapshots.
The double intervals based on the block_height is difficult to understand -- at least to me.

core/src/accounts_hash_verifier.rs
brooksprumo: @carllin I fixed this by adding in a shared `last_full_snapshot_slot` to `SnapshotConfig`:

https://github.com/solana-labs/solana/blob/a2491c73865a19e596162b5cc05d67cece943251/runtime/src/snapshot_config.rs#L33-L37

`SnapshotPackagerService` updates this value:

https://github.com/solana-labs/solana/blob/a2491c73865a19e596162b5cc05d67cece943251/core/src/snapshot_packager_service.rs#L48-L56

and `AccountsHashVerifier` reads this value:

https://github.com/solana-labs/solana/blob/a2491c73865a19e596162b5cc05d67cece943251/core/src/accounts_hash_verifier.rs#L223-L224

I'll also end up using this in `AccountsBackgroundService` as well, where `clean_accounts()` is called.

core/src/accounts_hash_verifier.rs
```diff
@@ -96,49 +96,73 @@ impl AccountsHashVerifier {
...
         thread_pool: Option<&ThreadPool>,
     ) {
-        let snapshot_package =
-            snapshot_utils::process_accounts_package(accounts_package, thread_pool, None);
-        Self::process_snapshot_package(
-            snapshot_package,
+        Self::verify_accounts_package_hash(&accounts_package, thread_pool);
+
+        Self::push_hashes_to_cluster(
+            &accounts_package,
```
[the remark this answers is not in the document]
brooksprumo: Correct, just a cosmetic thing. In the old version, the `hash` field in the SnapshotPackage was just copied from the AccountsPackage. In the new version, we only conditionally make a SnapshotPackage, so instead we get the hash from the AccountsPackage. The actual hash value is unchanged.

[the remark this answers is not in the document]
lijunwangs: Thanks for confirming!

core/src/accounts_hash_verifier.rs
brooksprumo: Yeah, this took me a bit to understand too. There are three intervals at play, and two "slots". First the intervals:

- accounts hash interval
- incremental snapshot interval
- full snapshot interval

The accounts hash interval is the interval used to take a bank snapshot, and also what gets pushed to AccountsHashVerifier and out to the cluster.

The full and incremental snapshot intervals each just say when to take their respective snapshots. Based on the code, full and incremental snapshots _must_ be a multiple of the accounts hash interval (otherwise the modulus may never hit zero). This is enforced in the CLI parameters validation.

One additional thing is that full snapshots take precedence over incremental snapshots. I.e. if the current slot ends up being an interval for _both_ full and incremental snapshots, then we only take a full snapshot.

And that does it for the intervals. Next are the two slots: the "slot" and the "block height".

You'll see that block height is used everywhere when comparing to the intervals. This is because it's possible for a slot number to just not exist. I.e. slots could validly be 1, 2, 3, 4, 6, 7, 8, 12, 13, 14, 17. If one of the intervals was 5, it would never get hit. Thus, block height. The block height never has gaps, so the interval checks never get skipped.

Does that make sense? Let me know if there are further questions. I'd love to make this part simpler if possible too!

core/src/accounts_hash_verifier.rs
brooksprumo: Done!

runtime/src/snapshot_package.rs
carllin: Yeah I think introducing an option to `get_snapshot_storages` to only clone out slots > a minimum slot is the right move. 

I think this also means `AccountsBackgroundService` should be the one setting the incremental/vs full snapshot flag here: https://github.com/solana-labs/solana/pull/19271/files#diff-387c1093824250b8106cb4207b95cf0b3ab875cd24cd1c2deed8d9fcd938c5aeR191-R205, and passing it downstream to the rest of the pipeline.

core/src/accounts_hash_verifier.rs
carllin: It's counted in slots, so I see no problem with using that as the unit of measurement

core/src/accounts_hash_verifier.rs
carllin: Maybe a. test for `submit_for_packaging()`?

core/src/accounts_hash_verifier.rs
carllin: Hmm this doesn't seem quite right because we could have an incremental snapshot that arrives before the latest full snapshot `B` is finished. This incremental snapshot will mark the latest full snapshot is some `A < B`, but has already filtered out the storages between `A` and `B`.

I think maybe the right thing to do is use modular arithmetic on the full snapshot interval to figure out what the *expected* last snapshot should be  for an incremental snapshot, and just use that?

runtime/src/snapshot_package.rs
brooksprumo: Yeah, I agree. I think that's ultimately the right way to do this. OK. I think that means I should close out this PR. I'll get working on AccountsBackgroundService!

I'll pull out some of the changes in this PR into smaller PRs as well (like the LastFullSnapshotSlot addition to SnapshotConfig).

brooksprumo:
Closing this PR (see [here](https://github.com/solana-labs/solana/pull/19271#discussion_r692947264) for more context). I will instead move the decision making about Full vs Incremental snapshot as early as possible (into AccountsBackgroundService).

PR #19272, opened 2021-08-17
Send request from ReplayStage to repair for dumped slots

carllin:
#### Problem
Replay doesn't send a request to repair a dumped slot

#### Summary of Changes
Add request from replay -> repair to fetch dumped slot from a peer who has it.

Fixes #

### DIFF
--- a/core/src/replay_stage.rs
+++ b/core/src/replay_stage.rs
@@ -18,7 +18,7 @@ use {
         heaviest_subtree_fork_choice::HeaviestSubtreeForkChoice,
         latest_validator_votes_for_frozen_banks::LatestValidatorVotesForFrozenBanks,
         progress_map::{ForkProgress, ProgressMap, PropagatedStats},
-        repair_service::DuplicateSlotsResetReceiver,
+        repair_service::{DuplicateSlotRepairRequestSender, DuplicateSlotsResetReceiver},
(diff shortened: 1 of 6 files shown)


core/src/replay_stage.rs
```diff
@@ -4921,6 +4928,179 @@ pub mod tests {
...
+
+        let VoteSimulator {
+            mut progress,
+            bank_forks,
+            ..
+        } = vote_simulator;
+
+        // Create situation where slots with *common ancestors* are marked dead,
+        // but then we detect the rest of the cluster has frozen the slot via EpochSlots.
+        // There should not be multiple repair requests generated for the common ancestors.
```
AshwinSekar, line 4960: Could there still be multiple requests due to dump and repair trying to fix these at the same time?

AshwinSekar (dismissed):
LGTM

core/src/replay_stage.rs
carllin: `duplicate_slot_repair_request_sender` is a hashmap so should de-duplicate any outstanding requests. 

Once the request has been made, however, and the slot removed from `duplicate_slot_repair_request_sender`, we may add it again, but only if the bank has been frozen again (another wrong hash for instance).

AshwinSekar (approved):

PR #19274, opened 2021-08-17
CalculateHashIntermediate::new_without_slot -> new

jeffwashington:
#### Problem
follow up rename on earlier refactor
#### Summary of Changes
move the function name back to 'new'
Fixes #


steviez (approved):
I was initially inclined to look to see if there was a `new_with_slot()` function, but doesn't make so, so makes sense to make it less verbose

Issue #19275, opened 2021-08-17
Transaction Cost updates


taozhu-chicago:
#### Problem
With the established calculation for cost unit: https://github.com/solana-labs/solana/pull/18994#issuecomment-896128992, should update transaction cost calculation accordingly. Few things to do: 

- adjust signature cost, re-measure its performance, currently is 10us, probably should be 25-50us, update to proper CU
- include signature cost into transaction cost calculation
- revisit need of `account_max_limit` logic
- add built-in instruction cost, add to transaction cost calc too. (need turn on `tx_wide_compute_cap` feature?)
- re-measure the block max cost. 
- Make sure when block limit is breached, and transactions are being rejected, system performance will not degrade significantly. 
- run perf test, and bench-tps

taozhu-chicago:
out dated

PR #19276, opened 2021-08-17
Update block max cost limit to fix performance regession

taozhu-chicago:
#### Problem
During performance test, `bench-tps` generated transfer transactions started to hit block max limit, therefore being rejected and retried by `banking_stage`. 

This is the result of block max limit was set too low for now, partly due to increasing account read/write cost. Created issue #19275 to resync cost limit

#### Summary of Changes
increase block_cost_max by 10 times to accommodate updated account read and write costs; 

Fixes #

### DIFF
--- a/ledger/src/block_cost_limits.rs
+++ b/ledger/src/block_cost_limits.rs
@@ -15,7 +15,7 @@ pub const fn max_instructions_per_block() -> u64 {
 }
 
 pub const fn block_cost_max() -> u64 {
-    MAX_INSTRUCTION_COST * max_instructions_per_block()
+    MAX_INSTRUCTION_COST * max_instructions_per_block() * 10
 }
 
 pub const fn account_cost_max() -> u64 {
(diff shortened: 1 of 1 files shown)


jeffwashington (approved):
The code looks fine to me, but I don't know enough about the effects.

behzadnouri:
I think someone who has followed cost-model changes more closely better to review the code.
But my concern is that if we expect that these constants to be reasonable for foreseeable future or just a matter of time until they cause an issue on the  cluster? and, if anything better than add-hoc tweaks can be done here?

Previous regression was caught by bench-tps. But that job is executing very specific type of transactions. How can we re-assure that a different kind of transactions (probably more similar to mainnet txs) will not be negatively impacted by this change or similar other changes in cost-model?

taozhu-chicago:
> I think someone who has followed cost-model changes more closely better to review the code.
> But my concern is that if we expect that these constants to be reasonable for foreseeable future or just a matter of time until they cause an issue on the cluster? and, if anything better than add-hoc tweaks can be done here?
> 
> Previous regression was caught by bench-tps. But that job is executing very specific type of transactions. How can we re-assure that a different kind of transactions (probably more similar to mainnet txs) will not be negatively impacted by this change or similar other changes in cost-model?

PR #19278, opened 2021-08-17
Fix flaky test test_fork_choice_refresh_old_votes

AshwinSekar:
#### Problem
During our leader slot we might enter a situation where we have voted for our slot but the slot has not yet reached blockstore through the broadcast thread. This causes the test to fail as we do not have the ancestor information for the fork in blockstore. 

#### Summary of Changes
Use the latest vote in tower that has also landed in blockstore.

Fixes #

### DIFF
--- a/local-cluster/tests/local_cluster.rs
+++ b/local-cluster/tests/local_cluster.rs
@@ -1042,9 +1042,9 @@ fn test_kill_partition_switch_threshold_progress() {
     \-------- 4 (38%, heavier fork)
 */
 // where the 2% that voted on slot 1 don't see their votes land in a block
-// and thus without integrating votes from gossip into fork choice, will
-// deem slot 4 the heavier fork and try to switch to slot 4, which doesn't pass the
-// switch threshold. This stalls the network.
+// due to blockhash expiration, and thus without resigning their votes with
+// a newer blockhash, will deem slot 4 the heavier fork and try to switch to
+// slot 4, which doesn't pass the switch threshold. This stalls the network.
 
 // We do this by:
 // 1) Creating a partition so all three nodes don't see each other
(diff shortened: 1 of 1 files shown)


local-cluster/tests/local_cluster.rs
```diff
@@ -1137,14 +1137,21 @@ fn test_fork_choice_refresh_old_votes() {
 
             info!("Opened blockstores");
 
-            // Get latest votes
-            let (lighter_fork_latest_vote, _) = last_vote_in_tower(
+            // Get latest votes. We additionally check to make sure the vote has landed in
```
carllin, line 1140: Would also be nice to fixup the bad test comment I left above:

```
// where the 2% that voted on slot 1 don't see their votes land in a block
// and thus without integrating votes from gossip into fork choice, will
```

```
// where the 2% that voted on slot 1 don't see their votes land in a block
// due to blockhash expiration, and thus without resigning their votes  with
// a newer blockhash, will
```

local-cluster/tests/local_cluster.rs
```diff
@@ -1137,14 +1137,27 @@ fn test_fork_choice_refresh_old_votes() {
...
             let (heaviest_fork_latest_vote, _) =
                 last_vote_in_tower(&heaviest_ledger_path, &context.heaviest_validator_key).unwrap();
+            while lighter_fork_blockstore
+                .meta(lighter_fork_latest_vote)
+                .unwrap()
+                .is_none()
+                || heaviest_blockstore
+                    .meta(heaviest_fork_latest_vote)
+                    .unwrap()
+                    .is_none()
```
carllin, line 1157: I think we should also wait for the block to be full, might be easier to just call `while !blockstore.is_full()`

carllin (dismissed):

PR #19285, opened 2021-08-18
replace function with const var for better readability

taozhu-chicago:
#### Problem
`const fn foo()` isn't as readable as `const FOO`

#### Summary of Changes
To use `const VAR`, no functional change

Fixes #

### DIFF
--- a/core/src/cost_model.rs
+++ b/core/src/cost_model.rs
@@ -68,7 +68,7 @@ pub struct CostModel {
 
 impl Default for CostModel {
     fn default() -> Self {
-        CostModel::new(account_cost_max(), block_cost_max())
+        CostModel::new(ACCOUNT_COST_MAX, BLOCK_COST_MAX)
     }
 }
 
@@ -126,9 +126,9 @@ impl CostModel {
 
             if is_writable {
                 self.transaction_cost.writable_accounts.push(*k);
-                self.transaction_cost.account_access_cost += account_write_cost();
+                self.transaction_cost.account_access_cost += ACCOUNT_WRITE_COST;
             } else {
-                self.transaction_cost.account_access_cost += account_read_cost();
+                self.transaction_cost.account_access_cost += ACCOUNT_READ_COST;
             }
         });
         debug!(
@@ -388,8 +388,7 @@ mod tests {
                 .try_into()
                 .unwrap();
 
-        let expected_account_cost =
-            account_write_cost() + account_write_cost() + account_read_cost();
+        let expected_account_cost = ACCOUNT_WRITE_COST + ACCOUNT_WRITE_COST + ACCOUNT_READ_COST;
         let expected_execution_cost = 8;
 
         let mut cost_model = CostModel::default();
@@ -445,7 +444,7 @@ mod tests {
 
         let number_threads = 10;
         let expected_account_cost =
-            account_write_cost() + account_write_cost() * 2 + account_read_cost() * 2;
+            ACCOUNT_WRITE_COST + ACCOUNT_WRITE_COST * 2 + ACCOUNT_READ_COST * 2;
         let cost1 = 100;
         let cost2 = 200;
         // execution cost can be either 2 * Default (before write) or cost1+cost2 (after write)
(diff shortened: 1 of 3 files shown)


steviez (approved):
For the sake of capturing our quick chat the other day (and partially for confirming my understanding) ...

A couple chunks from the [Rust reference](https://doc.rust-lang.org/reference/const_eval.html),
> Certain forms of expressions, called constant expressions, can be evaluated at compile time. In const contexts, these are the only allowed expressions, and are always evaluated at compile time. In other places, such as let statements, constant expressions may be, but are not guaranteed to be, evaluated at compile time

> The following expressions are constant expressions, so long as any operands are also constant expressions and do not cause any Drop::drop calls to be run.
> - Literals.
> - ...
> - Built-in negation, arithmetic, logical, comparison or lazy boolean operators used on integer and floating point types, bool, and char.

> A const context is one of the following:
> ...
> - The initializer of
>     - constants

> A const fn is a function that one is permitted to call from a const context ... When called from a const context, the function is interpreted by the compiler at compile time

So my understanding is that:
- Defining a constant (ie `pub const seven = 3 + 4`) is a const context (constant initializer), and this is guaranteed to be evaluated at compile-time.
- A const fn can be used in a const context, in which case it would be evaluated at compile time.
- A const fn can also be used in a non-const context, in which case it *may be* evaluated at compile time.

Given the above, I don't think the const fn's would be guaranteed to be resolved at compile-time (although probably likely would) since they aren't all used in const contexts, so using constants gives us that guarantee too.

Interested to hear your thought, but regardless, change looks good to me!

taozhu-chicago:
> For the sake of capturing our quick chat the other day (and partially for confirming my understanding) ...
> 
> A couple chunks from the [Rust reference](https://doc.rust-lang.org/reference/const_eval.html),
> 
> > Certain forms of expressions, called constant expressions, can be evaluated at compile time. In const contexts, these are the only allowed expressions, and are always evaluated at compile time. In other places, such as let statements, constant expressions may be, but are not guaranteed to be, evaluated at compile time
> 
> > The following expressions are constant expressions, so long as any operands are also constant expressions and do not cause any Drop::drop calls to be run.
> > 
> > * Literals.
> > * ...
> > * Built-in negation, arithmetic, logical, comparison or lazy boolean operators used on integer and floating point types, bool, and char.
> 
> > A const context is one of the following:
> > ...
> > 
> > * The initializer of
> >   
> >   * constants
> 
> > A const fn is a function that one is permitted to call from a const context ... When called from a const context, the function is interpreted by the compiler at compile time
> 
> So my understanding is that:
> 
> * Defining a constant (ie `pub const seven = 3 + 4`) is a const context (constant initializer), and this is guaranteed to be evaluated at compile-time.
> * A const fn can be used in a const context, in which case it would be evaluated at compile time.
> * A const fn can also be used in a non-const context, in which case it _may be_ evaluated at compile time.
> 
> Given the above, I don't think the const fn's would be guaranteed to be resolved at compile-time (although probably likely would) since they aren't all used in const contexts, so using constants gives us that guarantee too.
> 
> Interested to hear your thought, but regardless, change looks good to me!

Nice summary! I honestly not certain what `non-const context` exactly is. In my use case:

steviez:
> Nice summary! I think it is within `const context` in my use of these `const function`. The performance tests I did before/after suggests this is the case. But honestly, I am not sure what does `non-const context` mean, as in `const fn(variable: &mut u64) {}`?

Gotcha - I think I was mixed up on the which context is relevant here. Originally, I was thinking the caller of the constant/const function was the one that mattered. Such as:
```
self.transaction_cost.account_access_cost += ACCOUNT_READ_COST; // or account_read_cost()
```
But, as you pointed out to me in DM, I think the context of the const function itself is probably what is relevant.
```
pub const fn account_read_cost() -> u64 {
    // read account averages 5us
    compute_unit_to_us_ratio() * 5
}
```
This makes sense as the caller of a function shouldn't dictate how / when the compiler evaluates that function. Your quick test agrees with this so I'm all good / think I'm straightened out ... enough of this rabbit hole and back to real things ha

PR #19286, opened 2021-08-18
use 32k bins in accounts index

jeffwashington:
#### Summary of Changes

Fixes #

PR #19287, opened 2021-08-18
ITER_BATCH_SIZE from 1k -> 10k

jeffwashington:
#### Problem
AccountsIndex is changing. We are likely to move to a non-sorted map structure. We already bin into 16 bins in master. We will likely set the # bins to 32k(#19286) or more. 80M accounts across 32k bins is 2.4k pubkeys per bin.
When iterating through the accounts in the index, we currently start at the beginning and iterate, stopping at ITER_BATCH_SIZE. We then remember the pubkey of the last element. The next time we continue the iteration, we start after the last pubkey. This mechanism works fine for a sorted map.
For an unsorted map, if we arbitrarily stop in the middle of a bin iter (because we hit ITER_BATCH_SIZE), one easy method is to sort all the items in the bin to figure out how to reliably continue next time without repeating or skipping keys.
Sorts are expensive.
So, one option is to keep the bin sizes to a manageable size over time, as # accounts grow, and set ITER_BATCH_SIZE to generally higher than that bin size.
I can imagine other ways to accomplish this.
My question here is what problems do you see with ITER_BATCH_SIZE being 10k, for example.
We may be loading 10k accounts during an rpc scan perhaps.
#### Summary of Changes
This is a step in the direction of more items being returned per scan. Soon, I want to return all items in a bin (#19299). I'm hoping this simple PR will help us figure out if there are any concerns I'm not thinking of.

We are currently [tracking max items in a bin (currently ~5M with 16 bins in mnb)](https://metrics.solana.com:8888/sources/23/chronograf/data-explorer?query=SELECT%20mean%28%22max_bin_size%22%29%20AS%20%22mean_max_bin_size%22%2C%20mean%28%22min_bin_size%22%29%20AS%20%22mean_min_bin_size%22%20FROM%20%22mainnet-beta%22.%22autogen%22.%22generate_index%22%20WHERE%20time%20%3E%20%3AdashboardTime%3A%20AND%20time%20%3C%20%3AupperDashboardTime%3A%20GROUP%20BY%20time%28%3Ainterval%3A%29%20FILL%28null%29).

Fixes #

### DIFF
--- a/runtime/src/accounts_index.rs
+++ b/runtime/src/accounts_index.rs
@@ -31,7 +31,7 @@ use std::{
 };
 use thiserror::Error;
 
-pub const ITER_BATCH_SIZE: usize = 1000;
+pub const ITER_BATCH_SIZE: usize = 10_000;
 pub const BINS_DEFAULT: usize = 16;
 pub const ACCOUNTS_INDEX_CONFIG_FOR_TESTING: AccountsIndexConfig = AccountsIndexConfig {
     bins: Some(BINS_DEFAULT),


carllin:
>My question here is what problems do you see with ITER_BATCH_SIZE being 10k, for example.
We may be loading 10k accounts during an rpc scan perhaps.

For all the current RPC scans, I believe we scan using the secondary index.  This means the step size here doesn't matter as that scan simply looks up every pubkey in the secondary index, without any chunking: https://github.com/solana-labs/solana/blob/1d375ff2da642ac9a24f4d1fe88160f37c4b24c5/runtime/src/accounts_index.rs#L1119-L1123.

I think the primary concern is probably the rent collection. I think perf there will depend on the number of accounts we're collecting in rent on each call here to `load_to_collect_rent_eagerly()`. https://github.com/solana-labs/solana/blob/1d375ff2da642ac9a24f4d1fe88160f37c4b24c5/runtime/src/accounts.rs#L786-L792. If the number of accounts we're collecting rent from is a lot higher than this step size of `10_000`, it probably won't affect perf very much. If it's a lot lower, we may eat a little more overhead.

> 80M accounts across 32k bins is 2.4k pubkeys per bin.

Given this, it's probably good to check in the change to bump the bin number alongside this change, as this one is dependent on more bins to reduce lock contention?

jeffwashington:
Lock contention isn't the issue here. It may be later. There are several intersecting and related ideas going on. The larger bin count was to reduce the count per bin. That reduces the amount of items that have to be sorted if we have to sort a bin.
If we're on disk, then the larger bin count reduces how much goes in each file and reduces how much searching goes on, in theory. That is all yet to be tuned. Bin count is now a command line arg so we can optimize for the right set of tradeoffs among multiple other tuning parameters.

jeffwashington:
If we aren't using this scan for anything but rent collection, then can I propose that we remove this scan code path completely so that it can't be used for other than rent? That means we don't have to support it or worry about its performance at all.

jeffwashington:
@carllin making sure you see the comments. Thanks for engaging!

carllin:
> If we aren't using this scan for anything but rent collection, then can I propose that we remove this scan code path completely so that it can't be used for other than rent? That means we don't have to support it or worry about its performance at all.

Hmmm I think that's fine if we turn off the ability for nodes to do serve queries when they don't have the secondary index enabled. Right now I think, though generally unused, the fallback for not having secondary indexes enabled is this default scan. For example: https://github.com/solana-labs/solana/blob/1d375ff2da642ac9a24f4d1fe88160f37c4b24c5/rpc/src/rpc.rs#L1851-L1855

jeffwashington:
> if we turn off the ability for nodes to do serve queries

Can we discuss this? Is this a problem? Does it error? Have we turned off other things like this previously?

PR #19290, opened 2021-08-18
Plumb accounts-db-skip-shrink through testnet scripts

CriesofCarrots:
#### Problem
Booting simulations from snapshots with millions of accounts, initial shrink can take hours. Currently the only way to take advantage of #19028 is by manually editing the multinode-demo bash.

#### Summary of Changes
Plumb args to allow testnet boots to easily skip shrink


CriesofCarrots:
Flaky `test_fork_choice_refresh_old_votes`

PR #19298, opened 2021-08-18
Enable booting testnet validators without tower required

CriesofCarrots:
#### Problem
When rebooting a testnet from a previous snapshot -- eg. in the case of cluster failure, or to test new code against many existing accounts -- the previous tower file may no longer be relevant. Currently the only way to allow the boot to proceed is to manually remove the `--require-tower` arg from the multinode-demo scripts.

#### Summary of Changes
Add a `--skip-require-tower` option to skip the `--require-tower` arg.

This is foundational work to enable complete automation of rebooting testnet from previous snap.
Needs rebase on #19290

### DIFF
--- a/multinode-demo/bootstrap-validator.sh
+++ b/multinode-demo/bootstrap-validator.sh
@@ -20,6 +20,7 @@ else
 fi
 
 no_restart=0
+maybeRequireTower=true
 
 args=()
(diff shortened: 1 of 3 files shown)


CriesofCarrots:
Flaky `test_fork_choice_refresh_old_votes`

PR #19300, opened 2021-08-18
Fail secp256k1 if the instruction data looks incorrect

### LINKED ISSUE #17961: secp256k1 program succeeds even if the instruction data indicates there is no sigs
#### Problem

If a secp256k1 instruction data is large enough to contain signatures but that signature count (first byte) is zero the secp256k1 program return success.  This can be a footgun because the the issuer might be encoding the data incorrectly but has clearly encoded something more than the byte.   An instruction success, in this case, could lead to a false positive that is not easily recognized.

#### Proposed Solution

If the instruction data is larger than one byte (the count) but the count is zero, throw an error.

jackcmay:
#### Problem

If a secp256k1 instruction data is large enough to contain signatures but that signature count (first byte) is zero the secp256k1 program return success. This can be a footgun because the the issuer might be encoding the data incorrectly but has clearly encoded something more than the byte. An instruction success, in this case, could lead to a false positive that is not easily recognized.

#### Summary of Changes
If the instruction data is larger than one byte (the count) but the count is zero, throw an error.

Fixes #17961

### DIFF
--- a/banks-server/src/banks_server.rs
+++ b/banks-server/src/banks_server.rs
@@ -131,10 +131,14 @@ impl BanksServer {
 fn verify_transaction(
     transaction: &Transaction,
     libsecp256k1_0_5_upgrade_enabled: bool,
+    libsecp256k1_fail_on_bad_count: bool,
 ) -> transaction::Result<()> {
     if let Err(err) = transaction.verify() {
         Err(err)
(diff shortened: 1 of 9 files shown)


sakridge (approved):
lgtm

PR #19304, opened 2021-08-18
v1.6: Handle new security advisories

CriesofCarrots:
https://github.com/solana-labs/solana/pull/19126 for v1.6


CriesofCarrots:
Rolled into #19302

PR #19307, opened 2021-08-19
Fix null case in getMultipleAccountsInfo

### LINKED ISSUE #19301: web3.js - getMultipleAccountsInfo throws exception if any account is missing
#### Problem
When getting many accounts via Connection.getMultipleAccountsInfo, if any of the accounts do not exist, the entire promise throws an exception and rejects with an error of similar format to this:

ERROR	StructError: At path: value.4 -- Expected an object, but received: null
    at validate (/var/task/node_modules/superstruct/lib/index.cjs:363:19)
    at Object.create (/var/task/node_modules/superstruct/lib/index.cjs:319:18)
    at /var/task/node_modules/@solana/web3.js/lib/index.cjs.js:2562:29
    at Struct.coercer (/var/task/node_modules/superstruct/lib/index.cjs:948:52)

This error causes the entire promise to reject, and accounts which were successfully retrieved are not returned. This forces the user to call Connection.getAccountInfo in a loop if they are not certain of the status of the accounts they're querying.

#### Proposed Solution
If some accounts do not exist, still resolve the promise, but place 'null' at the corresponding index in the resolved AccountInfo array. This would make the behavior effectively the same as calling Connection.getAccountInfo in a loop. However, this might be considered a breaking change, as the returned arrays currently don't have nulls.

Arrowana:
#### Problem

#19301

#### Summary of Changes

Fixes #19301

Question:
In what case the response value is actually null? It seems like it 

https://docs.solana.com/developing/clients/jsonrpc-api#getmultipleaccounts

### DIFF
--- a/web3.js/src/connection.ts
+++ b/web3.js/src/connection.ts
@@ -2415,13 +2415,13 @@ export class Connection {
   async getMultipleAccountsInfo(
     publicKeys: PublicKey[],
     commitment?: Commitment,
-  ): Promise<AccountInfo<Buffer>[] | null> {
+  ): Promise<(AccountInfo<Buffer> | null)[] | null> {
     const keys = publicKeys.map(key => key.toBase58());
     const args = this._buildArgs([keys], commitment, 'base64');
     const unsafeRes = await this._rpcRequest('getMultipleAccounts', args);
     const res = create(
       unsafeRes,
-      jsonRpcResultAndContext(nullable(array(AccountInfoResult))),
+      jsonRpcResultAndContext(nullable(array(nullable(AccountInfoResult)))),
     );
     if ('error' in res) {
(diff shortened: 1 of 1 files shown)


oJshua (approved):

CriesofCarrots:
@oJshua , this failed the commit lint. I think it's going to have issues releasing

oJshua:
@CriesofCarrots thanks, let me verify. I fixed the commit title in the squash, so had assumed it would pass.

oJshua:
@CriesofCarrots we're good to go! 1.24.1 released

CriesofCarrots:
Oh cool! Thanks @oJshua TIL, including that reworded squash commits don't show up anywhere on the PR. I'll look for that in future!

PR #19308, opened 2021-08-19
stake: Remove v2 program references

joncinque:
#### Problem

While doing work on `credits_observed`, it made sense to also eliminate the `fix_stake_deactivate` flag everywhere, enabled in slot 55728001.  This will make it easier to read the other (coming) PR.

#### Summary of Changes

Remove references to stake program v2, but NOT the feature itself, so that it can still be queried with `solana feature status`.  Let me know if it makes sense to remove that too.

Note that this PR makes no logic changes at all, for your reviewing pleasure.


mvines:
> Remove references to stake program v2, but NOT the feature itself, so that it can still be queried with solana feature status. Let me know if it makes sense to remove that too.

Removing the feature flag itself is fine as well.  Do you need the v1.7 backport on this PR though?

joncinque:
> Removing the feature flag itself is fine as well.

Done, `rewrite_stake` went away with it

> Do you need the v1.7 backport on this PR though?

Yes, since it's a prereq for #19309

t-nelson (approved):
lgtm!

PR #19309, opened 2021-08-19
stake: Advance `credits_observed` on activation epoch

joncinque:
#### Problem

A stake does not receive rewards during its activation epoch, since `stake.delegation.stake(...` always returns 0 for the activation epoch [1].  On the other hand its `credits_observed` is not updated, which can make reward calculation confusing.

[1] https://github.com/solana-labs/solana/blob/b59b55716365be5c69a84d601e046ee37ccdb42c/sdk/program/src/stake/state.rs#L395

#### Summary of Changes

Advance `credits_observed` during the activation epoch.  My approach was to reuse the `rewarded_epoch` being used elsewhere rather than going off the `vote_state.epoch_credits()` as used in `calculate_stake_points_and_credits`.  It means there's some redundancy, but it felt clearer to me.

Edit: this was built on #19308, and rebased since then to isolate the changes

### DIFF
--- a/program-test/tests/warp.rs
+++ b/program-test/tests/warp.rs
@@ -2,7 +2,8 @@
 use {
     assert_matches::assert_matches,
     bincode::deserialize,
-    solana_program_test::{processor, ProgramTest, ProgramTestError},
+    solana_banks_client::BanksClient,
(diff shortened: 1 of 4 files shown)


t-nelson (commented):
lgtm!

t-nelson (approved):

PR #19318, opened 2021-08-19
Proposal: return data from BPF programs

seanyoung:
See docs/src/proposals/return-data.md in PR for details.

### DIFF
--- a/docs/src/proposals/return-data.md
+++ b/docs/src/proposals/return-data.md
@@ -93,19 +93,37 @@ certainly work for the CPI case, but this would not work RPC or Transaction case
 
 ## Proposed Solution
 
-The callee can set the return data using a new system call `sol_set_return_data(u8 *buf, u64 length)`.
+The callee can set the return data using a new system call `sol_set_return_data(buf: *const u8, length: u64)`.
 There is a limit of 1024 bytes for the returndata. This function can be called multiple times, and
 will simply overwrite what was written in the last call.
 
-The return data can be retrieved with `sol_get_return_data(u8 mut *buf, u64 length) -> u64`. This function
-returns the length of the return data.
+The return data can be retrieved with `sol_get_return_data(buf: *mut u8, length: u64, program_id: *mut u8) -> u64`. This function copies the return buffer, and the program_id that set the return data, and
+returns the length of the return data, or `u64::MAX` if no return data is set.
 
 When an instruction calls `sol_invoke()`, the return data of the callee is copied into the return data
 of the current instruction. This means that any return data is automatically passed up the call stack,
 to the callee of the current instruction (or the RPC call).
 
-Note that `sol_invoke()` clears the returns data, so that any return data from
-a previous invoke is not reused if the invoked fails to set a return data.
+Note that `sol_invoke()` clears the returns data before invoking the callee, so that any return data from
+a previous invoke is not reused if the invoked fails to set a return data. For example:
+
+ - A invokes B
+ - Before entry to B, return data is cleared.0
+ - B sets some return data and returns
+ - A invokes C
+ - Before entry to C, return data is cleared.
+ - C does not set return data and returns
+ - A checks return data and finds it empty
+
+Another scenario to consider:
+
+ - A invokes B
+ - B invokes C
+ - C sets return data and returns
+ - B does not touch return data and returns
+ - A gets return data from C
+ - A does not touch return data
+ - Return data from transaction is what C set.
 
 The compute costs are calculated for getting and setting the return data using
 the syscalls.
--- a/docs/src/proposals/return-data.md
+++ b/docs/src/proposals/return-data.md
@@ -104,7 +104,11 @@ When an instruction calls `sol_invoke()`, the return data of the callee is copie
 of the current instruction. This means that any return data is automatically passed up the call stack,
 to the callee of the current instruction (or the RPC call).
 
-Passing a return buffer invoke the same compute cost as passing data as input.
+Note that `sol_invoke()` clears the returns data, so that any return data from
+a previous invoke is not reused if the invoked fails to set a return data.
+
+The compute costs are calculated for getting and setting the return data using
+the syscalls.
 
 For a normal RPC or Transaction, the returndata is base64-encoded and stored along side the sol_log
 strings in the [stable log](https://github.com/solana-labs/solana/blob/95292841947763bdd47ef116b40fc34d0585bca8/sdk/src/process_instruction.rs#L275-L281).
--- a/docs/src/proposals/return-data.md
+++ b/docs/src/proposals/return-data.md
@@ -1,0 +1,139 @@
+# Return data from BPF programs
+
+## Problem
+
+In the Solidity langauge it is permitted to return any number of values from a function,
+for example a variable length string can be returned:
+
+```
+function foo1() public returns (string) {
+    return "Hello, world!\n";
+}
+```
+
+Multiple values, arrays and structs are permitted too.
+
+```
+struct S {
+    int f1;
+    bool f2
+};
+
+function foo2() public returns (string, int[], S) {
+    return (a, b, c);
+}
+```
+
+All the return values are eth abi encoded to a variable-length byte array.
+
+On ethereum errors can be returned too:
+
+```
+function withdraw() public {
+    require(msg.sender == owner, "Permission denied");
+}
+
+function failure() public {
+    revert("I afraid I can't do that dave");
+}
+```
+These errors help the developer debug any issue they are having, and can
+also be caught in a Solidity `try` .. `catch` block. Outside of a `try` .. `catch`
+block, any of these would cause the transaction or rpc to fail.
+
+## Existing solution
+
+The existing solution that Solang uses, writes the return data to the callee account data.
+The caller's account cannot be used, since the callee may not be the same BPF program, so
+it will not have permission to write to the callee's account data.
+
+Another solution would be to have a single return data account which is passed
+around through CPI. Again this does not work for CPI as the callee may not have
+permission to write to it.
+
+The problem with this solution is:
+
+- It does not work for RPC calls
+- It is very racey; a client has to submit the Tx and then retrieve the account
+  data. This is not atomic so the return data can be overwritten by another transaction.
+
+## Requirements for Solution
+
+It must work for:
+
+- RPC: An RPC should be able to return any number of values without writing to account data
+- Transaction: An transaction should be able to return any number of values without needing to write them account data
+- CPI: The callee must "set" return value, and the caller must be able to retrieve it.
+
+## Review of other chains
+
+### Ethereum (EVM)
+
+The `RETURN` opcode allows a contract to set a buffer as a returndata. This opcode takes a pointer to memory and a size. The `REVERT` opcode works similarly but signals that the call failed, and all account data changes must be reverted.
+
+For CPI, the caller can retrieve the returned data of the callee using the `RETURNDATASIZE` opcode which returns the length, and the `RETURNDATACOPY` opcode, which takes a memory destination pointer, offset into the returndata, and a length argument.
+
+Ethereum stores the returndata in blocks.
+
+### Parity Substrate
+
+The return data can be set using the `seal_return(u32 flags, u32 pointer, u32 size)` syscall.
+- Flags can be 1 for revert, 0 for success (nothing else defined)
+- Function does not return
+
+CPI: The `seal_call()` syscall takes pointer to buffer and pointer to buffer size where return data goes
+ - There is a 32KB limit for return data.
+
+Parity Substrate does not write the return data to blocks.
+
+## Rejected Solution
+
+The concept of ephemeral accounts has been proposed a solution for this. This would
+certainly work for the CPI case, but this would not work RPC or Transaction case.
+
+## Proposed Solution
+
+The callee can set the return data using a new system call `sol_returndata(u8 *buf, u64 length)`.
+This function does not return and signals success for execution of the contract. There is a limit
+of 1024 bytes for the returndata.
+
+The caller can retrieve the return data of the callee by a new system call:
+
+```
+fn sol_invoke_signed_rust_return_data(
+    instruction_addr: *const u8,
+    account_infos_addr: *const u8,
+    account_infos_len: u64,
+    signers_seeds_addr: *const u8,
+    signers_seeds_len: u64,
+    return_data_addr: *mut u8,
+    return_data_length: *mut u64,
+) -> u64;
+
+uint64_t sol_invoke_signed_c_return_data(
+  const SolInstruction *instruction,
+  const SolAccountInfo *account_infos,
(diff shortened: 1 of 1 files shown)


docs/src/proposals/return-data.md
```diff
@@ -0,0 +1,139 @@
...
+Parity Substrate does not write the return data to blocks.
+
+## Rejected Solution
+
+The concept of ephemeral accounts has been proposed a solution for this. This would
+certainly work for the CPI case, but this would not work RPC or Transaction case.
+
+## Proposed Solution
+
+The callee can set the return data using a new system call `sol_returndata(u8 *buf, u64 length)`.
```
ryoqun, line 96: :tada:

docs/src/proposals/return-data.md
```diff
@@ -0,0 +1,139 @@
...
+This function does not return and signals success for execution of the contract. There is a limit
+of 1024 bytes for the returndata.
+
+The caller can retrieve the return data of the callee by a new system call:
```
mvines, line 100:
```suggestion
The caller can retrieve the return data of the callee by using a new variant of the `sol_invoke` system call:
```

docs/src/proposals/return-data.md
```diff
@@ -0,0 +1,139 @@
...
+
+```
```
mvines, line 102: Should we put the language at the end?  `sol_invoke_signed_with_return_data_rust` and `sol_invoke_signed_with_return_data_c`

mvines (commented):
@armaniferrante / @jstarry - would be great if you have some time to review this proposal, it should be quick!

docs/src/proposals/return-data.md
seanyoung: Good point, I'll apply that.

docs/src/proposals/return-data.md
```diff
@@ -0,0 +1,139 @@
...
+
+CPI: The `seal_call()` syscall takes pointer to buffer and pointer to buffer size where return data goes
+ - There is a 32KB limit for return data.
+
+Parity Substrate does not write the return data to blocks.
+
+## Rejected Solution
+
+The concept of ephemeral accounts has been proposed a solution for this. This would
+certainly work for the CPI case, but this would not work RPC or Transaction case.
```
jstarry, line 92: Not suggesting we take this path but I think it would work if we had special handling for a `ReturnValueProgram` which could support a "write" instruction that programs could use via CPI and its accounts could be cleared by the runtime after each transaction instruction once the return value has been recorded in tx logs. Think of it as a ["shared memory program"](https://spl.solana.com/shared-memory) which has special runtime handling.

docs/src/proposals/return-data.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Proposed Solution
+
+The callee can set the return data using a new system call `sol_set_return_data(u8 *buf, u64 length)`.
+This function does not return and signals success for execution of the contract. There is a limit
+of 1024 bytes for the returndata.
```
jstarry, line 98: How was this 1024 byte limit decided? Seems plenty big enough but I could imagine protocols bumping up against this limit in the future if they use multiple programs which pass computed data around or need to bubble up and combine return values from nested CPI's.

docs/src/proposals/return-data.md
```diff
@@ -0,0 +1,139 @@
...
+Parity Substrate does not write the return data to blocks.
+
+## Rejected Solution
+
+The concept of ephemeral accounts has been proposed a solution for this. This would
+certainly work for the CPI case, but this would not work RPC or Transaction case.
+
+## Proposed Solution
+
+The callee can set the return data using a new system call `sol_set_return_data(u8 *buf, u64 length)`.
```
jstarry, line 96: Maybe this syscall could include "success" in its name to dissuade devs from mistakenly using it to return errors. Something like: `sol_return_success_with_data`

docs/src/proposals/return-data.md
```diff
@@ -0,0 +1,139 @@
...
+  const SolInstruction *instruction,
+  const SolAccountInfo *account_infos,
+  int account_infos_len,
+  const SolSignerSeeds *signers_seeds,
+  int signers_seeds_len
+  uint8_t *return_data,
+  uint64_t *return_data_length,
+);
+```
+On entry, `return_data_length` should point to the size of the buffer at `return_data`. If the callee
```
jstarry, line 123: Does the size of the return data impact compute unit consumption?

docs/src/proposals/return-data.md
mvines: 1024 was something smaller than https://github.com/solana-labs/solana/blob/22a18a68e3ee68ae013d647e62e12128433d7230/runtime/src/log_collector.rs#L3, but that's mostly it.

For top-level instructions, the idea was that the return value could just get surfaced in the transaction log instead of plumbing new storage/RPC for it. But even with new plumbing, we don't want programs to be dumping a ton of return metadata into all the RPC node's rocksdb and BigTable. An alternative to this limitation would be for them to allocate an account to hold the data

docs/src/proposals/return-data.md
mvines: I think the idea was that the non-success case could still return data, but it's obviously then only available by reading the transaction log over RPC

docs/src/proposals/return-data.md
armaniferrante: Using logs to return values SGTM.

jstarry: Cool, sounds good. Agreed that accounts can be used if needed

docs/src/proposals/return-data.md
jstarry: > I think the idea was that the non-success case could still return data

The line below says:
> This function does not return and signals success for execution of the [instruction]

Does that not contradict what you're saying? Or are you referring to runtime verification failures?

docs/src/proposals/return-data.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Rejected Solution
+
+The concept of ephemeral accounts has been proposed a solution for this. This would
+certainly work for the CPI case, but this would not work RPC or Transaction case.
+
+## Proposed Solution
+
+The callee can set the return data using a new system call `sol_set_return_data(u8 *buf, u64 length)`.
+This function does not return and signals success for execution of the contract. There is a limit
```
jstarry, line 97:
```suggestion
This function does not return and signals success for execution of the program instruction. There is a limit
```

docs/src/proposals/return-data.md
mvines: Indeed it does contradict what I said. @seanyoung, wdyt? I recall us discussing that it could be desirable for programs to have the ability to return data when they fail as well.

docs/src/proposals/return-data.md
```diff
@@ -0,0 +1,139 @@
...
+  int signers_seeds_len
+  uint8_t *return_data,
+  uint64_t *return_data_length,
+);
+```
+On entry, `return_data_length` should point to the size of the buffer at `return_data`. If the callee
+sets more data than this buffer, an error should be returned. On successful exit, `return_data_length`
+will contain the actual length returned by the callee.
+
+For a normal RPC or Transaction, the returndata is base64-encoded and stored along side the sol_log
```
mvines, line 127: Wouldn't hurt to mention that the log format for return data would end up as a stable log declaration in https://github.com/solana-labs/solana/blob/95292841947763bdd47ef116b40fc34d0585bca8/sdk/src/process_instruction.rs#L275-L281

seanyoung: Good point, I've mentioned it with a link.

docs/src/proposals/return-data.md
seanyoung: > I think the idea was that the non-success case could still return data

Returning an error/non-success via this return buffer is a bad idea.
- Success/non-success becomes more complicated; program instruction can return success return code and an error in the return data
- On non-success the account data state changes should be reverted. We don't want to to this either on the VM side or program instruction side
- Only transaction that succeed should be valid, we don't intend on supporting recovery from errors (i.e. try-catch)
- Non-success cases can surface errors via program log.

docs/src/proposals/return-data.md
seanyoung: That might work. However, I do think this more complicated for both the program instruction side and the vm side rather than this return data solution.

docs/src/proposals/return-data.md
seanyoung: That's an interesting question but I don't know what the answer to that is. @jackcmay @mvines wdyt?

jackcmay: Rather than overloading invoke, why not break this out into a separate syscall that retrieves the return data?

docs/src/proposals/return-data.md
jackcmay: Any reason this syscall shouldn't be separated from success/fail and exit or not?  We could instead provide a syscall that allows the program to set some amount of "information data" which the program can set at any time but only the last data set is retained.  Effectively a single "return data" blob per tx.  This data could be stored next to the log data and available via rpc or the calling program and replaced within a tx every time an instruction processor calls the `set` syscall.  This would allow programs to return richer data on failure via rpc or return data to calling programs and rpc on success.

docs/src/proposals/return-data.md
jackcmay: If multiple instructions return data does each one's data get stored in the stable log?

jackcmay: "stored along side" or do you mean "embedded in"?

docs/src/proposals/return-data.md
seanyoung: That sounds like a great idea. I'll amend the proposal.

docs/src/proposals/return-data.md
seanyoung: Overloading invoke saves the program making two syscalls (invoke & get_return_data) rather than just invoke.

Maybe the separate syscall makes more sense. Happy to move on that and amend the proposal.

jackcmay: Yeah, that's the trade-off.  I'm leaning toward separating the two so that the two API's don't get entwined.  Especially if we change invoke, we might need to create two new invokes.  But the advantage of a single syscall isn't negligible so I could be swayed either way.

jackcmay: And yes, we have a common way of charging for data passed in and out of the program, these bytes should incur a similar cost

ronielex (approved):
I didn't know about this promotion or being completely?

docs/src/proposals/return-data.md
seanyoung: BTW it would be great if we had a way of separating log data from return data. Could we have a special return data entry in the stable log? Ideally we could also have an event type entry too for solidity events.

docs/src/proposals/return-data.md
seanyoung: I've updated the proposal.

ronielex (commented):
This kind of idea means if someone to missed 
The proposal didn't succeed it's means there is a lot of pending of what did done for.

docs/src/proposals/return-data.md
mvines: > Could we have a special return data entry in the stable log? Ideally we could also have an event type entry too for solidity events.

Yeah this is what I was thinking we'd do here

mvines (dismissed):
Latest iteration looks great to me, nice and simple

docs/src/proposals/return-data.md
```diff
@@ -0,0 +1,121 @@
...
+The callee can set the return data using a new system call `sol_set_return_data(u8 *buf, u64 length)`.
+There is a limit of 1024 bytes for the returndata. This function can be called multiple times, and
+will simply overwrite what was written in the last call.
+
+The return data can be retrieved with `sol_get_return_data(u8 mut *buf, u64 length) -> u64`. This function
+returns the length of the return data.
+
+When an instruction calls `sol_invoke()`, the return data of the callee is copied into the return data
+of the current instruction. This means that any return data is automatically passed up the call stack,
+to the callee of the current instruction (or the RPC call).
```
joncinque, line 105: I'm coming in a bit late here, but here's an idea, take it or leave it.  Would it make sense to also include the program id that set the return data?

As you say here, if you have program A which calls program B which calls program C, ie. A -> B -> C, program C can set a return value, B can use that return value but *not* set a return value itself, then A can also read C's return value.  As a safety check, A may want to know who set the return value, B or C.

If program A relies on C's return value, for example to calculate how many tokens to transfer, then a malicious program B could update itself to change that value, and steal funds.  It's a bit far-fetched, but at least we'd give devs the tools to program defensively.

mvines: I think that's a really good idea

mvines: @seanyoung - wdyt about providing the program id that set the return data in `sol_get_return_data()`?  This'll allow the caller to add a defensive check like what Jon mentioned if they want.
Something like `sol_get_return_data(u8 mut *buf, u64 length, Pubkey mut *from) -> u64`.

ronielex (approved):
I hope it works for us

docs/src/proposals/return-data.md
```diff
@@ -0,0 +1,125 @@
...
+
+Note that `sol_invoke()` clears the returns data, so that any return data from
```
jackcmay, line 107: Can you elaborate this, A -> B -> C, does this mean that if B sets return data it would be cleared when `invoke`ing C?  So if B sets return data, calls C, C does not set return data, then A will not receive B's return value?

This behavior should be well documented next to the `sol_set_return_data` system call so folks have a clear understanding of what to expect

docs/src/proposals/return-data.md
seanyoung: I totally agree this is a good idea. I've changed the syscall as suggested. Thanks @joncinque !

docs/src/proposals/return-data.md
seanyoung: I've expanded on this.

ronielex (approved):

seanyoung:
Merged via https://github.com/solana-labs/solana/pull/19548

Lichtso:
Unfortunately, I saw it just now that it was merged, and I have some concerns:
* I don't think a syscall is the right way to implement this. We should avoid syscalls if possible, because of:
  * their context switch overhead
  * they are hard to get rid off again
  * simply not necessary here as a program can only return once, if I understand the purpose correctly
* Global (transaction wide) state in `InvokeContext` is what I am trying to get rid off as it requires explicit copies everywhere.

Instead, as it needs to be implemented in the new ABI (see #19191) anyway, I think that reserving a shared memory area there would be a better approach.

seanyoung:
@Lichtso I should have added you as a reviewer, sorry about that. I think your criticisms are all valid. 

Your proposed ABI requires re-building BPF programs anyway, how about if we removed the syscall in ABI v2.

PR #19319, opened 2021-08-19
Allow closing upgradeable program accounts

jackcmay:
#### Problem

- Developers can never reclaim the lamports from an upgradeable program account
- If someone transfers ownership of an account that holds lamports to the upgradeable loader they can never reclaim those lamports

#### Summary of Changes

- Allow upgradeable program accounts to be closed.  The actual executable "Program" account is not closed or drained, but the "ProgramData" account is.  The "Program" account holds very few lamports and is executable so is immutable so will be left forever.  When the ProgramData account is closed its lamports are drained and it is set to the `Uninitialized` state.  This means that any invoke, upgrade, etc.. of that program in the future will fail.
- Allow loader accounts in the `Uninitialized` state to be drained.  This does not require an authority so there is a chance that someone else could drain it first but this should be rare because `Uninitialized` accounts owned by the loader should never carry any lamports.

Fixes #

### DIFF
--- a/cli/src/program.rs
+++ b/cli/src/program.rs
@@ -14,7 +14,7 @@ use solana_clap_utils::{self, input_parsers::*, input_validators::*, keypair::*}
 use solana_cli_output::{
     display::new_spinner_progress_bar, CliProgram, CliProgramAccountType, CliProgramAuthority,
     CliProgramBuffer, CliProgramId, CliUpgradeableBuffer, CliUpgradeableBuffers,
-    CliUpgradeableProgram,
+    CliUpgradeableProgram, CliUpgradeableProgramClosed,
 };
 use solana_client::{
     client_error::ClientErrorKind,
@@ -860,15 +860,17 @@ fn process_program_deploy(
                     false
                 } else {
                     return Err(format!(
-                        "{} is not an upgradeable loader ProgramData account",
+                        "Program {} has been closed, use a new Program Id",
                         programdata_address
                     )
                     .into());
                 }
             } else {
-                return Err(
-                    format!("ProgramData account {} does not exist", programdata_address).into(),
-                );
+                return Err(format!(
+                    "Program {} has been closed, use a new Program Id",
+                    programdata_address
+                )
+                .into());
             }
         } else {
             return Err(format!("{} is not an upgradeable program", program_pubkey).into());
@@ -1196,17 +1198,10 @@ fn process_show(
                                         - UpgradeableLoaderState::programdata_data_offset()?,
                                 }))
                         } else {
-                            Err(format!("Invalid associated ProgramData account {} found for the program {}",
-                                        programdata_address, account_pubkey)
-                                    .into(),
-                            )
+                            Err(format!("Program {} has been closed", account_pubkey).into())
                         }
                     } else {
-                        Err(format!(
-                            "Failed to find associated ProgramData account {} for the program {}",
-                            programdata_address, account_pubkey
-                        )
-                        .into())
+                        Err(format!("Program {} has been closed", account_pubkey).into())
                     }
                 } else if let Ok(UpgradeableLoaderState::Buffer { authority_address }) =
                     account.state()
@@ -1298,18 +1293,10 @@ fn process_dump(
                             f.write_all(program_data)?;
                             Ok(format!("Wrote program to {}", output_location))
                         } else {
-                            Err(
-                                format!("Invalid associated ProgramData account {} found for the program {}",
-                                        programdata_address, account_pubkey)
-                                    .into(),
-                            )
+                            Err(format!("Program {} has been closed", account_pubkey).into())
                         }
                     } else {
-                        Err(format!(
-                            "Failed to find associated ProgramData account {} for the program {}",
-                            programdata_address, account_pubkey
-                        )
-                        .into())
+                        Err(format!("Program {} has been closed", account_pubkey).into())
                     }
                 } else if let Ok(UpgradeableLoaderState::Buffer { .. }) = account.state() {
                     let offset = UpgradeableLoaderState::buffer_data_offset().unwrap_or(0);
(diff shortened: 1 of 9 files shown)


jackcmay:
@jstarry Does this method of closing accounts fall into the same traps as #19217 is trying to avoid?

jstarry:
> @jstarry Does this method of closing accounts fall into the same traps as #19217 is trying to avoid?

Yeah, it does. Any uninitialized upgradeable loader account can be initialized as a program or a buffer. It's fine if buffers are reused but it's not great that this applies to programdata accounts too. Your implementation in this PR looks sound as is, but it would be nice if a programdata account could never be initialized as a different type of upgradeable loader account since it's a PDA

jackcmay:
> > @jstarry Does this method of closing accounts fall into the same traps as #19217 is trying to avoid?
> 
> Yeah, it does. Any uninitialized upgradeable loader account can be initialized as a program or a buffer. It's fine if buffers are reused but it's not great that this applies to programdata accounts too. Your implementation in this PR looks sound as is, but it would be nice if a programdata account could never be initialized as a different type of upgradeable loader account since it's a PDA

I'm still not seeing the issue with reinitializing uninitialized loader owned accounts, can you shed more light?

jstarry:
> I'm still not seeing the issue with reinitializing uninitialized loader owned accounts, can you shed more light?

Any uninitialized programdata account can be reused as a buffer account or as a new program account without requiring the programdata account to sign. So after the programdata account is closed, it can be reused if it is initialized in the same transaction. I don't see any ways that this can cause any issues, but I'm of the opinion that it would be weird if a program's programdata account becomes an executable program itself with its own programdata account.

jackcmay:
Hmm, yeah I guess that would be kinda weird to see in the explorer, but the loader doesn't enforce any signing on the Program accounts during deploy or at any other time so why does it care if the account is a PDA or not?

jstarry:
> so why does it care if the account is a PDA or not?

It doesn't, I don't see anything problematic about this PR besides pointing out the weirdness of allowing program data's to turn into programs.

jackcmay:
> > so why does it care if the account is a PDA or not?
> 
> It doesn't, I don't see anything problematic about this PR besides pointing out the weirdness of allowing program data's to turn into programs.

Ok, cool, just checking to make sure I wasn't missing anything related to the larger closing issue

jackcmay:
@mvines A set of eyes would be appreciated
@jstarry Can you take a look from a program deployment workflow perspective?

Thanks!

jstarry:
@jackcmay we need to ensure that the program account has a write lock when closing a programdata account just like we do for upgrades. Otherwise a close programdata tx could be parallelized with a tx that invokes the program.

jstarry:
I have a few no-action-needed comments about this PR:
- Allowing programs to be "closed" now is not much different from upgrading to a no-op implementation and then freezing (by clearing the upgrade authority). So adding the ability to also drain the lampposts seems like a net-positive, I don't see an argument someone could make about not wanting this feature at this point.
- I imagine a protocol would want to have a different authority for "freezing" and "closing" since those are both one-way operations and could be disastrous if performed unintentionally. But this could easily be bolted on by setting the upgrade authority to another program which uses different authorities to gate different loader instructions.

programs/bpf_loader/src/lib.rs
```diff
@@ -620,37 +632,70 @@ fn process_loader_upgradeable_instruction(
...
+                _ => {
+                    ic_logger_msg!(logger, "Account does not support closing");
+                    return Err(InstructionError::InvalidArgument);
                 }
-            } else {
-                ic_logger_msg!(logger, "Account does not support closing");
-                return Err(InstructionError::InvalidArgument);
             }
 
             ic_logger_msg!(logger, "Closed {}", close_account.unsigned_key());
```
jstarry, line 701: What do you think of changing this key to the program id if the programdata is being closed? Also, it would be nice if the log specified what type of account was closed.

jstarry:
Some CLI feedback:
- `solana program close -h` is too buffer specific, it should mention programs too or have a separate subcommand
- when deploying to a closed program, the error message uses the programdata address instead of the program id:
```
$ solana program close 7ZjUKQ7G2knCUFRu46EVGMCBmWRhegyg8NQaq3FvZZQd
Closed Program Id 7ZjUKQ7G2knCUFRu46EVGMCBmWRhegyg8NQaq3FvZZQd, 2.2108092 SOL reclaimed

$ solana program deploy ./my_program.so
Error: Program 6aWnZC5sazX4HgdiNzMsE8JxBHB7dEBad1fMXipE1aUL has been closed, use a new Program Id
```

jstarry:
Should we add the close instruction to `check_authorized_program` so that programs which use multisigs and governance as their upgrade authority can close the program?
https://github.com/solana-labs/solana/blob/ac8bd587ecdee2870ba25a7639291719b1c27b3c/programs/bpf_loader/src/syscalls.rs#L2243-L2245

sdk/cargo-build-bpf/src/main.rs
```diff
@@ -587,6 +587,8 @@ fn build_bpf_package(config: &Config, target_directory: &Path, package: &cargo_m
         check_undefined_symbols(config, &program_so);
 
         println!();
+        println!("Deploying this program will use the following generated keypair by default for the program's address:");
+        println!("  {}", program_keypair.display());
         println!("To deploy this program:");
         println!("  $ solana program deploy {}", program_so.display());
```
mvines, line 593: I got the new message under 80 characters thusly: 
```suggestion
        println!("To deploy this program:");
        println!("  $ solana program deploy {}", program_so.display());
        println!("The program address will default to this keypair (override with --program-id):");
        println!("  {}", program_keypair.display());
```

programs/bpf_loader/src/lib.rs
```diff
@@ -198,6 +201,15 @@ fn process_instruction_common(
                     ic_logger_msg!(logger, "Wrong ProgramData account for this Program account");
                     return Err(InstructionError::InvalidArgument);
                 }
+                if let UpgradeableLoaderState::ProgramData {
+                    slot: _,
+                    upgrade_authority_address: _,
+                } = programdata.state()?
+                {
+                } else {
```
mvines, line 209: How about this:
```suggestion
                if !matches!(programdata.state()?, UpgradeableLoaderState::ProgramData { .. }) {
```

programs/bpf_loader/src/lib.rs
```diff
@@ -620,37 +632,70 @@ fn process_loader_upgradeable_instruction(
...
-            if let UpgradeableLoaderState::Buffer { authority_address } = close_account.state()? {
-                if authority_address.is_none() {
-                    ic_logger_msg!(logger, "Buffer is immutable");
-                    return Err(InstructionError::Immutable);
-                }
-                if authority_address != Some(*authority.unsigned_key()) {
-                    ic_logger_msg!(logger, "Incorrect buffer authority provided");
-                    return Err(InstructionError::IncorrectAuthority);
+            if !invoke_context.is_feature_active(&close_upgradeable_program_accounts::id()) {
+                if let UpgradeableLoaderState::Buffer {
```
mvines, line 645: `matches!()` could be used here as well

programs/bpf_loader/src/lib.rs
jackcmay: We don't know the program id at this point, there is no reverse lookup and I wasn't planning on changing the close instruction to include the extra Program account.

jackcmay:
> Should we add the close instruction to `check_authorized_program` so that programs which use multisigs and governance as their upgrade authority can close the program?
> 
> https://github.com/solana-labs/solana/blob/ac8bd587ecdee2870ba25a7639291719b1c27b3c/programs/bpf_loader/src/syscalls.rs#L2243-L2245

Great idea!

programs/bpf_loader/src/lib.rs
jackcmay: After making the changes to require the program write lock I also made these changes.

cli/src/program.rs
```diff
@@ -333,29 +333,31 @@ impl ProgramSubCommands for App<'_, '_> {
                 )
                 .subcommand(
                     SubCommand::with_name("close")
-                        .about("Close an account and withdraw all lamports")
+                        .about("Close an program orbuffer account, and withdraw all lamports")
```
jstarry, line 336:
```suggestion
                        .about("Close an program or buffer account, and withdraw all lamports")
```

sdk/program/src/loader_upgradeable_instruction.rs
```diff
@@ -117,8 +117,11 @@ pub enum UpgradeableLoaderInstruction {
     /// withdraws all the lamports
     ///
     /// # Account references
-    ///   0. `[writable]` The account to close.
+    ///   0. `[writable]` The account to close, if closing a program must be the
+    ///      ProgramData account.
     ///   1. `[writable]` The account to deposit the closed account's lamports.
     ///   2. `[signer]` The account's authority.
+    ///   0. `[writable]` The associated Program account if the account to close
```
jstarry, line 124:
```suggestion
    ///   3. `[writable]` The associated Program account if the account to close
```

sdk/program/src/loader_upgradeable_instruction.rs
```diff
@@ -117,8 +117,11 @@ pub enum UpgradeableLoaderInstruction {
     /// withdraws all the lamports
     ///
     /// # Account references
-    ///   0. `[writable]` The account to close.
+    ///   0. `[writable]` The account to close, if closing a program must be the
+    ///      ProgramData account.
     ///   1. `[writable]` The account to deposit the closed account's lamports.
     ///   2. `[signer]` The account's authority.
```
jstarry, line 123:
```suggestion
    ///   2. `[signer]` The account's authority, if the account in not uninitialized.
```

programs/bpf_loader/src/syscalls.rs
```diff
@@ -2242,7 +2242,8 @@ fn check_authorized_program(
         || bpf_loader_deprecated::check_id(program_id)
         || (bpf_loader_upgradeable::check_id(program_id)
             && !(bpf_loader_upgradeable::is_upgrade_instruction(instruction_data)
-                || bpf_loader_upgradeable::is_set_authority_instruction(instruction_data)))
+                || bpf_loader_upgradeable::is_set_authority_instruction(instruction_data)
+                || bpf_loader_upgradeable::is_close_instruction(instruction_data)))
```
jstarry, line 2246: need a feature switch here because closing a buffer could succeed before the `close_upgradeable_program_accounts` feature switch is enabled

PR #19325, opened 2021-08-19
Accountsdb replication installment 2

lijunwangs:
#### Problem
This is the 2nd installment for the AccountsDb replication.

#### Summary of Changes
The basic google protocol buffer protocol for replicating updated slots and accounts. tonic/tokio is used for transporting the messages. 

The basic framework of the client and server for replicating slots and accounts -- the persisting of accounts in the replica-side will be done at the next PR -- right now -- the accounts are streamed to the replica-node and dumped. Replication for information about Bank is also not done in this PR -- to be addressed in the next PR to limit the change size.

Functionality used by both the client and server side are encapsulated in the replica-lib crate.

There is no impact to the existing validator by default.

Tests:

Observe the confirmed slots replicated to the replica-node.
Observe the accounts for the confirmed slot are received at the replica-node side.

Fixes #


replica-lib/src/accountsdb_repl_server.rs
```diff
@@ -0,0 +1,177 @@
...
+                .worker_threads(worker_threads)
+                .thread_name("sol-accountsdb-repl-wrk")
+                .enable_all()
+                .build()
+                .expect("Runtime"),
+        );
+
+        let server_cloned = accountsdb_repl_server.clone();
+        let (exit_signal_sender, exit_signal_receiver) = oneshot::channel::<()>();
+        // let sender = Arc::new(Mutex::new(Some(tx)));
```
lijunwangs, line 115: remove

replica-lib/src/replica_accounts_server.rs
```diff
@@ -0,0 +1,88 @@
...
+    fn get_slot_accounts(
+        &self,
+        request: &accountsdb_repl_server::ReplicaAccountsRequest,
+    ) -> Result<accountsdb_repl_server::ReplicaAccountsResponse, tonic::Status> {
+        let slot = request.slot;
+
+        match self.bank_forks.read().unwrap().get(slot) {
+            None => Err(tonic::Status::not_found("The slot is not found")),
+            Some(bank) => {
+                let accounts = bank.rc.accounts.scan_slot(slot, |account| match account {
```
lijunwangs, line 30: refactor and put them into separate functions?

replica-node/src/accountsdb_repl_service.rs
```diff
@@ -0,0 +1,81 @@
+/// Module responsible for replicating AccountsDb data from its peer to its local AccountsDb
```
lijunwangs, line 1: At the replica-node side/

validator/src/main.rs
```diff
@@ -1691,6 +1693,37 @@ pub fn main() {
                 .requires("enable_rpc_transaction_history")
                 .help("Verifies blockstore roots on boot and fixes any gaps"),
         )
+        .arg(
```
lijunwangs, line 1696: Hide all of these parameters until fully functional.

replica-lib/src/accountsdb_repl_server.rs
```diff
@@ -0,0 +1,177 @@
...
+    tokio::{
+        runtime::Runtime,
+        sync::oneshot::{self, Receiver, Sender},
+    },
+    tonic::{self, transport},
+};
+
+tonic::include_proto!("accountsdb_repl");
+
+pub trait ReplicaUpdatedSlotsServer {
```
carllin, line 18: nit: I think "updated" -> "confirmation" might be a little clearer here, i.e.

`ReplicaUpdatedSlotsServer -> ReplicaSlotConfirmationServer`

replica-lib/src/replica_accounts_server.rs
carllin: hmmm how does this compare to just sending the storage entry and having the replica parse out each account? Could be less work for the validator.

replica-node/src/accountsdb_repl_service.rs
```diff
@@ -0,0 +1,81 @@
...
+        mut accountsdb_repl_client: AccountsDbReplClientService,
+    ) {
+        loop {
+            match accountsdb_repl_client.get_updated_slots(last_replicated_slot) {
+                Ok(slots) => {
+                    info!("Received updated slots: {:?}", slots);
+                    if !slots.is_empty() {
+                        for slot in slots.iter() {
+                            Self::replicate_accounts_for_slot(&mut accountsdb_repl_client, *slot);
+                        }
```
carllin, line 66: A bit worried here that before  we've fetched all the accounts, the root will have moved past the confirmed slots we just fetched, and we may have cleaned some necessary information.

For example:

`1 - > 2 -> 3 -> 4 -> 5`

Let's say the slots we returned from `get_updated_slots()` are `1` and `2`.

While we are running this loop `for slot in slots.iter() {`, the validator makes a root at 
slot `3` and cleans up information from slots `1` and `2`. Now we have an inconsistent view of slot `2`.

It seems like when we make a request for the accounts of `requested_slot` we need to fetch *all* information up to and including `safe_slot = max(requested_slot, current_root)`

replica-lib/src/replica_accounts_server.rs
lijunwangs: You mean just stream out the binary directly?I was mulling over the same. One I am not sure how to handle different versions when the structure changes? In addition, some of the fields only applies locally.

replica-node/src/accountsdb_repl_service.rs
lijunwangs: Agreed with the concern of race condition. I was planning to address it in a following-on PR. We have to address gaps, using your example,
1 - > 2 -> 3 -> 4 -> 5

Suppose we successfully replicated information for slot 1 and 2. Bank for slot 3 is dropped. And we process at 4. Will we not have consistency issue as well?

validator/src/main.rs
lijunwangs: Done

replica-lib/src/accountsdb_repl_server.rs
lijunwangs: Yes. Will address

replica-node/src/accountsdb_repl_service.rs
carllin: > Suppose we successfully replicated information for slot 1 and 2. Bank for slot 3 is dropped at the server. And we process at 4. Will we not have consistency issue as well?

The server should return all the parents of the `safe_slot`, so we will know if there are any gaps missing.

replica-lib/src/accountsdb_repl_server.rs
lijunwangs: Done

replica-lib/src/accountsdb_repl_client.rs
```diff
@@ -0,0 +1,119 @@
...
+
+impl From<tonic::transport::Error> for ReplicaRpcError {
+    fn from(err: tonic::transport::Error) -> Self {
+        ReplicaRpcError::ConnectionError(err.to_string())
+    }
+}
+
+impl AccountsDbReplClient {
+    pub async fn connect(rpc_peer: &SocketAddr) -> Result<Self, ReplicaRpcError> {
+        let url = format!("http://{}", rpc_peer);
```
brooksprumo, line 31: Do you know if we can use https here and get some added security (some TLS maybe)?

brooksprumo (commented):
I think everything looked good to me, just had one question where I left a comment. I don't have a full picture of all the moving parts, so it's likely worthwhile to get another set of eyes on the _why_'s of the code as well.

replica-lib/src/accountsdb_repl_client.rs
lijunwangs: I think the replication is mostly done between nodes trust each other in the same data center -- so communication between them is trusted. As for detecting data corruption in-transit -- I would look into the hashing of the account and see if we can piggyback on that. Using http would be introduce less CPU overhead and better performance.

mvines: A TLS tunnel can also be added at the server ops level as well using existing off-the-shelf software.

replica-lib/src/replica_confirmed_slots_server.rs
```diff
@@ -0,0 +1,121 @@
+use tonic;
+
+// tonic::include_proto!("accountsdb_repl");
```
sakridge, line 3: Do we need this?

replica-lib/src/accountsdb_repl_client.rs
brooksprumo: Noted. Thanks for the replies. Marking as resolved.

replica-lib/src/replica_confirmed_slots_server.rs
lijunwangs: No -- I will remove the comment. The types are imported via accountsdb_repl_server

replica-node/src/accountsdb_repl_service.rs
lijunwangs: This issue will be handled in installment 3.

replica-lib/src/accountsdb_repl_server_factory.rs
```diff
@@ -0,0 +1,29 @@
...
+
+pub struct AccountsDbReplServerFactory {}
+
+impl AccountsDbReplServerFactory {
+    pub fn build_accountsdb_repl_server(
+        config: AccountsDbReplServiceConfig,
+        confirmed_bank_receiver: Receiver<Slot>,
+        bank_forks: Arc<RwLock<BankForks>>,
+    ) -> AccountsDbReplService {
+        AccountsDbReplService::new(
```
sakridge, line 21: This threw me off a bit. We don't really use factory pattern elsewhere and I'm not sure how much value this wrapper adds, but I'm not opposed especially if you had further plans to expand it.

sakridge (dismissed):
lgtm solid work

replica-lib/src/accountsdb_repl_server_factory.rs
lijunwangs: This pattern is used to achieve two goals: the AccountsDbReplService does not need to know about the real implementations of ReplicaSlotConfirmationServer and  ReplicaAccountsServer and the caller to instantiate AccountsDbReplService does not need to know about (ReplicaSlotConfirmationServerImpl and ReplicaAccountsServerImpl. So that we could change the implementation freely. As such we could unit test the AccountsDbReplService and code below its stack without a real validator.

PR #19326, opened 2021-08-19
Add runtime address map cache

jstarry:
#### Problem
Need an efficient way to map the addresses of mapped transactions (v2)

#### Summary of Changes
- Defined state and serialization for an `AddressMap` account
- Added skeleton for address map program
- Created a deterministic address map cache in the runtime which is populated from snapshots, updated on rooted banks, and pruned on epoch boundaries.

#### Implementation Notes:
1. Address map cache additions from rooted banks are not immediately usable. The current implementation requires a warm up period of 2 epochs to ensure that all nodes have the same view of activated address maps.
2. Address map account entries are immutable while an address map is activated so there is no need to acquire locks when using the address map cache for address lookups.

Fixes #


runtime/src/accounts_db.rs
```diff
@@ -6019,6 +6033,11 @@ impl AccountsDb {
                     stored_account,
                 },
             )| {
+                if solana_address_map_program::check_id(&stored_account.account_meta.owner) {
+                    self.address_map_cache
+                        .populate_cache(&stored_account.meta.pubkey, stored_account.data);
+                }
```
jstarry, line 6039: @jeffwashington as discussed, this is where I stuck in the address map cache changes. My understanding is that this will pick up every stored address map account from the snapshot.

programs/address_map/build.rs
```diff
@@ -0,0 +1,27 @@
+extern crate rustc_version;
+use rustc_version::{version_meta, Channel};
+
+fn main() {
+    // Copied and adapted from
+    // https://github.com/Kimundi/rustc-version-rs/blob/1d692a965f4e48a8cb72e82cda953107c0d22f47/README.md#example
+    // Licensed under Apache-2.0 + MIT
+    match version_meta().unwrap().channel {
```
t-nelson, line 8: :eyes:

runtime/src/address_map/cache.rs
```diff
@@ -0,0 +1,581 @@
...
+            if let Some(deactivation_epoch) = address_map.deactivation_epoch {
+                rooted_bank_epoch < deactivation_epoch.saturating_add(DEACTIVATION_COOLDOWN)
+            } else {
+                true
+            }
+        });
+    }
+
+    /// Called during snapshot processing to populate the cache from rooted slots
+    pub fn populate_cache(&self, map_key: &Pubkey, map_data: &[u8]) {
```
t-nelson, line 186: Any reason not to make this fallible?  Seems like it'd help debugging deserialization failures

runtime/src/address_map/pending_changes.rs
```diff
@@ -0,0 +1,184 @@
...
+        self.activations.iter()
+    }
+
+    /// Iterator over all address maps deactivated in a single bank
+    pub fn deactivations_iter(&self) -> impl Iterator<Item = SetEntry<Pubkey>> {
+        self.deactivations.iter()
+    }
+
+    /// Record pending change for address map account if it was newly activated or deactivated
+    pub fn record(&self, key: &Pubkey, map_data: &[u8]) {
```
t-nelson, line 61: Seems like this one could also be fallible

t-nelson (commented):
Looking good!

runtime/src/accounts_db.rs
jeffwashington: Seems reasonable to populate `address_map_cache` like this.

programs/address_map/build.rs
jstarry: Ha, well this is just copy pasta from other build scripts to get the abi digest macro working

ryoqun: symlink please? or is some tweak needed?:)

programs/address_map/src/lib.rs
```diff
@@ -0,0 +1,189 @@
...
+#[derive(Debug, Serialize, Deserialize, PartialEq, Clone, AbiExample)]
+pub struct AddressMap {
+    /// Authority address which must sign for each modification.
+    pub authority: Option<Pubkey>,
+    /// Activation slot. Address map entries may not be modified once activated.
+    pub activation_slot: Slot,
+    /// Deactivation slot. Address map accounts cannot be closed unless they have been deactivated.
+    pub deactivation_slot: Slot,
+    /// Number of stored address entries. Limited by `MAX_ADDRESS_MAP_ENTRIES`.
+    pub num_entries: u8,
```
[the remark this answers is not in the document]
jstarry: Maybe it was in an early draft but it was merged using `u8`. Do you have any objections to this approach?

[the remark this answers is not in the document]
t-nelson: Nope, 256 seems plenty.  I just missed the spec change.  All good here :pray:

runtime/src/address_map/pending_changes.rs
jstarry: Updated to return a result

runtime/src/address_map/cache.rs
jstarry: No reason, updated to return a result

programs/address_map/build.rs
jstarry: Updated to use a symlink instead!

runtime/src/address_map/cache.rs
```diff
@@ -0,0 +1,581 @@
...
+    }
+
+    /// Map a message's address map indexes to full addresses if the address
+    /// maps referenced in the message are valid and active for the current
+    /// epoch.
+    pub fn map_message_addresses(
+        &self,
+        message: &v0::Message,
+        current_epoch: Epoch,
+    ) -> Option<MappedAddresses> {
```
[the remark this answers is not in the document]
jstarry: Updated, thanks

jstarry:
@t-nelson this is ready for another pass

jstarry:
> Those seem to be requirements for the in-memory cache as well. You don't get around conflicting address access either way. Only 1 thread can make a modifying update to an account.

I implemented an in-memory cache because the transaction processing pipeline would need significant changes to charge fees for invalid transactions. By having the cache in-memory, the overhead from fee-less invalid transactions is minimized. Also, address map entries are immutable while the account is active so no locking is required for address map lookups.

> If the accounts are clearly AccountMap references, then you don't have any overhead of checking each one, no?

I was referring to needing to load the accounts for AccountMap references. We need to verify that these accounts exist, are owned by the address map program, and are active for the current epoch before using them to lookup addresses. That overhead can be reduced by using the in-memory cache.

jstarry:
> I implemented an in-memory cache because the transaction processing pipeline would need significant changes to charge fees for invalid transactions. By having the cache in-memory, the overhead from fee-less invalid transactions is minimized. Also, address map entries are immutable while the account is active so no locking is required for address map lookups.

After the new cost model and [fee changes](https://github.com/solana-labs/solana/pull/16984) are implemented, we will be in a better position to charge fees for invalid transactions. If we charge fees for invalid address map lookups, it shouldn't be an issue to use system resources for random lookups and so this approach of having a special in-memory cache is probably not needed. By not using a special cache, we also avoid the need to worry about how much active memory is needed to keep the full cache in memory. For these reasons, I will change the implementation to use accounts db directly for address mapping.

PR #19331, opened 2021-08-20
don't skip shrinking slots that would go to 0

jeffwashington:
#### Summary of Changes

Fixes #


jeffwashington:
@lijunwangs Here's a theory. Testing it now.

runtime/src/accounts_db.rs
```diff
@@ -2524,7 +2524,9 @@ impl AccountsDb {
         let mut shrink_slots: ShrinkCandidates = HashMap::new();
         for usage in &store_usage {
             let alive_ratio = (total_alive_bytes as f64) / (total_bytes as f64);
-            if alive_ratio > shrink_ratio {
+            let store = &usage.store;
+            let after_shrink_size = Self::page_align(store.alive_bytes() as u64);
+            if after_shrink_size > 0 && alive_ratio > shrink_ratio {
```
lijunwangs, line 2529: Good find!

lijunwangs (approved):

jeffwashington:
#19320

sakridge:
So is the point to actually shrink those stores with 0 alive bytes? Or is it that we were not getting much past those with 0 alive bytes to shrink other stores?

sakridge:
We don't shrink stores with alive_bytes==0 anyway:
https://github.com/solana-labs/solana/blob/f8fefc9c5ab77567a7f8cec672e5f5c72c08a714/runtime/src/accounts_db.rs#L2316

lijunwangs:
I could not reproduce the condition with a validator against devnet over night. So the increased usage has some other reasons

lijunwangs:
> We don't shrink stores with alive_bytes==0 anyway:
> 
> https://github.com/solana-labs/solana/blob/f8fefc9c5ab77567a7f8cec672e5f5c72c08a714/runtime/src/accounts_db.rs#L2316

Yes. AccountStorageEntry with 0 alive bytes is not processed during shrink -- they have nothing to shrink. The question is when they are added to the `RecycleStores`?

PR #19341, opened 2021-08-20
Add LastFullSnapshotSlot to SnapshotConfig

brooksprumo:
#### Problem

There is not a way to communicate what the last full snapshot slot is.

#### Summary of Changes

Add `last_full_snapshot_slot` to `SnapshotConfig` to enable passing runtime information between threads about the last full snapshot slot.

Related to #17088 

_(To see how LastFullSnapshotSlot will be used, see this next PR #19342)_


jeffwashington (approved):
lgtm

PR #19342, opened 2021-08-20
Use LastFullSnapshotSlot

brooksprumo:
#### Problem

Threads/background services do not communicate the last full snapshot slot, and therefore do not call `clean_accounts()` correctly once adding Incremental Snapshots.

#### Summary of Changes

- SnapshotPackagerService now gets a SnapshotConfig
- SnapshotPackagerService writes/updates last_full_snapshot_slot after successfully taking a full snapshot
- AccountsBackgroundService reads last_full_snapshot_slot and passes that value when it calls `clean_accounts()`
- Validator also writes last_full_snapshot_slot after it calls `bank_to_full_snapshot_archive()`.

Related to #17088


brooksprumo:
@carllin and I spoke, and he pointed out a way to get the last full snapshot slot without needing the lock, so I'm going to go down that route for now. Closing this PR.

PR #19348, opened 2021-08-20
Add base_slot to get_snapshot_storages()

brooksprumo:
#### Problem

When taking a bank snapshot that is destined to be an incremental snapshot, we collect _all_ the snapshot storages, then immediately filter/throw out the ones below the incremental snapshot base slot (aka last full snapshot slot). Ideally, we'd only collect the snapshot storages that we needed in the first place.

#### Summary of Changes

Add a `snapshot_base_slot` optional parameter to `get_snapshot_storages()`, and filter/retain only slots higher.

Related to #17088

### DIFF
--- a/core/tests/snapshots.rs
+++ b/core/tests/snapshots.rs
@@ -282,7 +282,7 @@ mod tests {
             snapshot_path,
             last_bank.src.slot_deltas(&last_bank.src.roots()),
             &snapshot_config.snapshot_package_output_path,
-            last_bank.get_snapshot_storages(),
+            last_bank.get_snapshot_storages(None),
             ArchiveFormat::TarBzip2,
             snapshot_version,
             None,
@@ -364,7 +364,7 @@ mod tests {
 
         // Take snapshot of zeroth bank
         let bank0 = bank_forks.get(0).unwrap();
-        let storages = bank0.get_snapshot_storages();
+        let storages = bank0.get_snapshot_storages(None);
         snapshot_utils::add_bank_snapshot(snapshot_path, bank0, &storages, snapshot_version)
             .unwrap();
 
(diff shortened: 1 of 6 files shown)


core/tests/snapshots.rs
```diff
@@ -788,7 +788,7 @@ mod tests {
             .find(|elem| elem.slot == slot)
             .ok_or_else(|| Error::new(ErrorKind::Other, "did not find snapshot with this path"))?;
         let storages = {
-            let mut storages = bank.get_snapshot_storages();
+            let mut storages = bank.get_snapshot_storages(Some(incremental_snapshot_base_slot));
             snapshot_utils::filter_snapshot_storages_for_incremental_snapshot(
```
brooksprumo, line 792: The call to `filter_snapshot_storages_for_incremental_snapshot()` will be removed in the next PR. I am trying to keep PRs as minimal as possible :)

runtime/src/bank.rs
```diff
@@ -483,13 +483,6 @@ impl BankRc {
             bank_id_generator: Arc::new(AtomicU64::new(0)),
         }
     }
-
-    pub fn get_snapshot_storages(&self, slot: Slot) -> SnapshotStorages {
-        self.accounts
-            .accounts_db
-            .get_snapshot_storages(slot, None)
-            .0
-    }
```
brooksprumo, line 492: This function is never called, except by `Bank` below, so I've just moved it there.

runtime/src/snapshot_utils.rs
```diff
@@ -1636,7 +1636,7 @@ pub fn bank_to_incremental_snapshot_archive(
 
     let temp_dir = tempfile::tempdir_in(snapshots_dir)?;
     let storages = {
-        let mut storages = bank.get_snapshot_storages();
+        let mut storages = bank.get_snapshot_storages(Some(full_snapshot_slot));
         filter_snapshot_storages_for_incremental_snapshot(&mut storages, full_snapshot_slot);
```
brooksprumo, line 1640: The call to `filter_snapshot_storages_for_incremental_snapshot()` will be removed in the next PR.

carllin (approved):

PR #19354, opened 2021-08-20
Dedupe SendTransactionService(s)

CriesofCarrots:
#### Problem
https://github.com/solana-labs/solana/blob/master/rpc/src/send_transaction_service.rs#L1
Copy-pasted SendTransactionService requires changes be made multiple places

#### Summary of Changes
Rework SendTransactionService with a generic to only require solana-gossip and solana-poh for rpc_service usage
Remove duplicate solana-banks-server version of service

Last 2 commits can be skipped: purely cosmetic


CriesofCarrots:
@mvines , I don't love adding a new crate for this, but the highest common dependency would be runtime, and it didn't seem like an obvious fit there. Thoughts?

transaction/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "solana-transaction"
+version = "1.8.0"
+description = "Solana transaction"
+authors = ["Solana Maintainers <maintainers@solana.foundation>"]
+repository = "https://github.com/solana-labs/solana"
+homepage = "https://solana.com/"
+documentation = "https://docs.rs/solana-transaction-status"
```
mvines, line 8: Doc link needs an update

mvines (commented):
> I don't love adding a new crate for this, but the highest common dependency would be runtime, and it didn't seem like an obvious fit there. Thoughts?

runtime definitely doesn't seem like a nice fit.  I don't mind the new crate, I'm not hot on the crate name `solana-transaction` as that feels much too generic.  `solana-send-transaction-service` is a mouthful but I don't hate it and it's accurate at least.  `solana-transaction-service` doesn't seem terrible either.

CriesofCarrots:
> `solana-send-transaction-service` is a mouthful but I don't hate it and it's accurate at least.

@mvines , went with this. Everything else look okay?

mvines (approved):
yep, looks good!

PR #19355, opened 2021-08-20
Correct JSON-RPC docs for getSlot

brson:
#### Problem

From my reading and experiments, the description of the `getSlot` method is not quite correct. It says it returns the slot the node is processing, but what it really does is returns the slot that corresponds to a provided commitment level. That is not (ever?) the slot that is being processed, but a slot some number back.

#### Summary of Changes

Use more precise language.

### DIFF
--- a/docs/src/developing/clients/jsonrpc-api.md
+++ b/docs/src/developing/clients/jsonrpc-api.md
@@ -2247,10 +2247,7 @@ Result:
 
 ### getSlot
 
-Returns the slot associated with the given commitment level
-
-If the commitment level is not provided, then this method returns the highest
-finalized slot.
+Returns the slot that has reached the [given or default commitment level](jsonrpc-api.md#configuring-state-commitment)
 
 #### Parameters:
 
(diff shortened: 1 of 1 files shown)


docs/src/developing/clients/jsonrpc-api.md
```diff
@@ -2247,7 +2247,10 @@ Result:
 
 ### getSlot
 
-Returns the current slot the node is processing
+Returns the slot associated with the given commitment level
+
+If the commitment level is not provided, then this method returns the highest
+finalized slot.
```
CriesofCarrots, line 2253:
```suggestion
Returns the slot that has reached the [given or default commitment level](jsonrpc-api.md#configuring-state-commitment)
```

jstarry:
Yup, much better, thanks!

CriesofCarrots (commented):
Great catch! Vestigial docs from before we had different commitment levels 😝 
Wdyt about deduping/linking to the commitment section for info on the default level?

brson:
> Great catch! Vestigial docs from before we had different commitment levels 😝
> Wdyt about deduping/linking to the commitment section for info on the default level?

wfm. I committed your suggestion.

CriesofCarrots (approved):

PR #19362, opened 2021-08-22
implements copy-on-write for vote-accounts

behzadnouri:
#### Problem
`Bank::vote_accounts` redundantly clones vote-accounts HashMap even though
an immutable reference will suffice:
https://github.com/solana-labs/solana/blob/95c998a19/runtime/src/bank.rs#L5174-L5186

#### Summary of Changes
This commit implements copy-on-write semantics for vote-accounts by
wrapping the underlying HashMap in `Arc<...>`.

### DIFF
--- a/runtime/src/vote_account.rs
+++ b/runtime/src/vote_account.rs
@@ -31,7 +31,7 @@ struct VoteAccountInner {
 
 #[derive(Debug, AbiExample)]
 pub struct VoteAccounts {
-    vote_accounts: HashMap<Pubkey, (/*stake:*/ u64, VoteAccount)>,
+    vote_accounts: Arc<HashMap<Pubkey, (/*stake:*/ u64, VoteAccount)>>,
     // Inner Arc is meant to implement copy-on-write semantics as opposed to
     // sharing mutations (hence RwLock<Arc<...>> instead of Arc<RwLock<...>>).
     staked_nodes: RwLock<
@@ -46,7 +46,7 @@ pub struct VoteAccounts {
 }
 
 impl VoteAccount {
-    pub fn lamports(&self) -> u64 {
+    pub(crate) fn lamports(&self) -> u64 {
         self.0.account.lamports
     }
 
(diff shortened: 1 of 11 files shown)


runtime/src/vote_account.rs
```diff
@@ -83,37 +83,43 @@ impl VoteAccounts {
...
             let vote_account = vote_account.clone();
             self.add_node_stake(delta, &vote_account);
         }
     }
 
-    pub fn sub_stake(&mut self, pubkey: &Pubkey, delta: u64) {
-        if let Some((stake, vote_account)) = self.vote_accounts.get_mut(pubkey) {
+    pub(crate) fn sub_stake(&mut self, pubkey: &Pubkey, delta: u64) {
+        let vote_accounts = Arc::make_mut(&mut self.vote_accounts);
+        if let Some((stake, vote_account)) = vote_accounts.get_mut(pubkey) {
```
carllin, line 122: Nice! I think this works

Quick sanity check, this should be safe even if we have interleaving updates between Replay and BankingStage.

For example, imagine we have:

```
     /--------5 (Banking Stage)
0
   \-----------4 (ReplayStage)
```

So the sequence of events:
1)  We start a leader block for our own slot `5` , inheriting the shared `stakes` object from the slot `0`: https://github.com/solana-labs/solana/blob/84db04ce6c9bce7e1bb556ec44ec26e0461af5dc/runtime/src/bank.rs#L1344.

2) At some later point in time Replay concurrently creates and starts replaying block `4`, also inheriting the shared `stakes` object from the slot `0`. 

The situation we want to avoid is either of the slot-specific updates to `stakes` via `update_cached_accounts()`: https://github.com/solana-labs/solana/blob/84db04ce6c9bce7e1bb556ec44ec26e0461af5dc/runtime/src/bank.rs#L5152 from being visible to the other child. 

Now we know this is safe because if we do an update in slot `5` before or after `4` is created, we clone the vote accounts for `5`, so these updates are hidden from `4`.

behzadnouri: Yes, it will be safe because as soon as any mutations are about to be made (and there is still another Arc reference to the same hash-map), a clone is made and so the mutations are not leaked to other Arc references.
The added `test_staked_nodes_cow` test is testing one such scenario.

carllin (approved):

PR #19365, opened 2021-08-23
Fix duplicate broadcast test

carllin:
#### Problem
1.  `partition_last_data_shreds` isn't shared between the run() and transmit() shreds
2. `partition_last_data_shreds` is currently only transmitted if the selected broadcast node via `get_broadcast_peer()` is in the partition. However, when the partition stake is small such as in `test_duplicate_shreds_broadcast_leader()`, then the partition node is never selected by `get_broadcast_peer()`

#### Summary of Changes
1. Share `partition_last_data_shreds` via a `Mutex`
2. Always broadcast `partition_last_data_shreds` to the partition nodes, even if they are not selected via `get_broadcast_peer()`

Fixes #

### DIFF
--- a/core/src/broadcast_stage/broadcast_duplicates_run.rs
+++ b/core/src/broadcast_stage/broadcast_duplicates_run.rs
@@ -60,8 +60,8 @@ impl BroadcastDuplicatesRun {
             prev_entry_hash: None,
             num_slots_broadcasted: 0,
             cluster_nodes_cache,
-            original_last_data_shreds: Arc::new(Mutex::new(HashSet::default())),
-            partition_last_data_shreds: Arc::new(Mutex::new(HashSet::default())),
+            original_last_data_shreds: Arc::<Mutex<HashSet<Signature>>>::default(),
+            partition_last_data_shreds: Arc::<Mutex<HashSet<Signature>>>::default(),
         }
     }
 }
@@ -281,15 +281,11 @@ impl BroadcastRun for BroadcastDuplicatesRun {
         let socket_addr_space = cluster_info.socket_addr_space();
         let packets: Vec<_> = shreds
             .iter()
-            .flat_map(|shred| {
+            .filter_map(|shred| {
                 let seed = shred.seed(Some(self_pubkey), &root_bank);
-                let node = cluster_nodes.get_broadcast_peer(seed);
-                if node.is_none() {
-                    return vec![];
-                }
-                let node = node.unwrap();
+                let node = cluster_nodes.get_broadcast_peer(seed)?;
                 if !socket_addr_space.check(&node.tvu) {
-                    return vec![];
+                    return None;
                 }
                 if self
                     .original_last_data_shreds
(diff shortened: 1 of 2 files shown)


core/src/broadcast_stage/broadcast_duplicates_run.rs
```diff
@@ -269,44 +273,58 @@ impl BroadcastRun for BroadcastDuplicatesRun {
...
+                    // If the shred is part of the partition, broadcast it directly to
+                    // the partition node
+                    return cluster_partition
+                        .iter()
+                        .filter_map(|pubkey| {
+                            let tvu = cluster_info
+                                .lookup_contact_info(pubkey, |contact_info| contact_info.tvu)?;
+                            Some((&shred.payload, tvu))
+                        })
+                        .collect();
```
carllin, line 324: Always broadcast partition shreds to the validators in the partition

core/src/broadcast_stage/broadcast_duplicates_run.rs
```diff
@@ -269,44 +273,58 @@ impl BroadcastRun for BroadcastDuplicatesRun {
...
+
         // Broadcast data
         let cluster_nodes =
             self.cluster_nodes_cache
                 .get(slot, &root_bank, &working_bank, cluster_info);
         let socket_addr_space = cluster_info.socket_addr_space();
         let packets: Vec<_> = shreds
             .iter()
-            .filter_map(|shred| {
+            .flat_map(|shred| {
```
behzadnouri, line 284: Probably better to keep `filter_map` here so that we have the `?` operator inside, and just add `.flatten()` right before `.collect()` at the end.

core/src/broadcast_stage/broadcast_duplicates_run.rs
```diff
@@ -269,44 +273,58 @@ impl BroadcastRun for BroadcastDuplicatesRun {
...
+                        return vec![];
                     }
+                } else if self
+                    .partition_last_data_shreds
+                    .lock()
+                    .unwrap()
+                    .remove(&shred.signature())
+                {
+                    // If the shred is part of the partition, broadcast it directly to
+                    // the partition node
```
behzadnouri, line 316: This is explaining what the code doing is; not why it is doing that. This comment from the change description is more helpful:
> when the partition stake is small ..., then the partition node is never selected by get_broadcast_peer()

Can you please add that instead here as the comment why `get_broadcast_peer` is ignored in this case?

core/src/broadcast_stage/broadcast_duplicates_run.rs
behzadnouri: This seems to add an asymmetry between when it has to broadcast to the partition vs the non-partition?

Should this wide fanout be conditional on `self.config.stake_partition` being less than some threshold? e.g. `self.config.stake_partition < 10%` or something.
What if `get_broadcast_peer` already returns a node within the partition?

Also, not related to this specific change, but seems like this whole partitioning mechanism only happens during broadcast stage. The retransmit stage is not aware of the partition and will fanout to its subtree. So in the case that we stick with `get_broadcast_peer` (or, otherwise include the very first node in first neighborhood) it will fanout to the entire tree.
Is that fine for purpose of the tests? or, should this always do a wide fantout to all the target nodes and then not use retransmit-stage at all for the tests?

behzadnouri (approved):
lgtm

PR #19373, opened 2021-08-23
Shrink mmap leakage

lijunwangs:
#### Problem
The shrink by total usage may skip shrink candidates in the next round.

#### Summary of Changes
If the candidate's own alive ratio is lower than the threshold add it back to the candidate so that it can be processed next time.

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -2500,7 +2500,7 @@ impl AccountsDb {
         let mut total_alive_bytes: u64 = 0;
         let mut candidates_count: usize = 0;
         let mut total_bytes: u64 = 0;
-        let mut total_candidate_stores: i64 = 0;
+        let mut total_candidate_stores: usize = 0;
         for (slot, slot_shrink_candidates) in shrink_slots {
             candidates_count += slot_shrink_candidates.len();
             for store in slot_shrink_candidates.values() {
@@ -2563,24 +2563,8 @@ impl AccountsDb {
             measure.as_ms() as usize
         );
         inc_new_counter_info!("select_top_sparse_storage_entries-seeds", candidates_count);
-        datapoint_info!(
-            "shrink-candidates",
-            (
-                "total_preliminary_candidate_stores",
-                total_candidate_stores,
-                i64
-            ),
-            (
-                "select_top_sparse_storage_entries-ms",
-                measure.as_ms() as u64,
-                i64
-            ),
-            (
-                "select_top_sparse_storage_entries-seeds",
-                candidates_count,
-                i64
-            ),
-        );
+        inc_new_counter_info!("total_preliminary_candidate_stores", total_candidate_stores);
+
         (shrink_slots, shrink_slots_next_batch)
     }
 
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -2489,7 +2489,7 @@ impl AccountsDb {
         &self,
         shrink_slots: &ShrinkCandidates,
         shrink_ratio: f64,
-    ) -> (ShrinkCandidates, Option<ShrinkCandidates>) {
+    ) -> (ShrinkCandidates, ShrinkCandidates) {
         struct StoreUsageInfo {
             slot: Slot,
             alive_ratio: f64,
@@ -2527,6 +2527,8 @@ impl AccountsDb {
         for usage in &store_usage {
             let store = &usage.store;
             let alive_ratio = (total_alive_bytes as f64) / (total_bytes as f64);
+            debug!("alive_ratio: {:?} store_id: {:?}, store_ratio: {:?} requirment: {:?}, total_bytes: {:?} total_alive_bytes: {:?}",
+                alive_ratio, usage.store.append_vec_id(), usage.alive_ratio, shrink_ratio, total_bytes, total_alive_bytes);
             if alive_ratio > shrink_ratio {
                 // we have reached our goal, stop
                 debug!(
@@ -2541,7 +2543,7 @@ impl AccountsDb {
                         .insert(store.append_vec_id(), store.clone());
                 } else {
                     break;
-		}
+                }
             } else {
                 let current_store_size = store.total_bytes();
                 let after_shrink_size = Self::page_align(store.alive_bytes() as u64);
@@ -2559,7 +2561,7 @@ impl AccountsDb {
             measure.as_ms() as usize
         );
         inc_new_counter_info!("select_top_sparse_storage_entries-seeds", candidates_count);
-        (shrink_slots, Some(shrink_slots_next_batch))
+        (shrink_slots, shrink_slots_next_batch)
     }
 
     pub fn shrink_candidate_slots(&self) -> usize {
@@ -2567,7 +2569,9 @@ impl AccountsDb {
             std::mem::take(&mut *self.shrink_candidate_slots.lock().unwrap());
         let (shrink_slots, shrink_slots_next_batch) = {
             if let AccountShrinkThreshold::TotalSpace { shrink_ratio } = self.shrink_ratio {
-                self.select_candidates_by_total_usage(&shrink_candidates_slots, shrink_ratio)
+                let (shrink_slots, shrink_slots_next_batch) =
+                    self.select_candidates_by_total_usage(&shrink_candidates_slots, shrink_ratio);
+                (shrink_slots, Some(shrink_slots_next_batch))
             } else {
                 (shrink_candidates_slots, None)
             }
@@ -2589,12 +2593,8 @@ impl AccountsDb {
             measure_shrink_all_candidates.as_ms() as usize
         );
         inc_new_counter_info!("shrink_all_candidate_slots-count", shrink_candidates_count);
-        if let AccountShrinkThreshold::TotalSpace {
-            shrink_ratio: _shrink_ratio,
-        } = self.shrink_ratio
-        {
+        if let Some(shrink_slots_next_batch) = shrink_slots_next_batch {
             let mut shrink_slots = self.shrink_candidate_slots.lock().unwrap();
-            let shrink_slots_next_batch = shrink_slots_next_batch.unwrap();
             for (slot, stores) in shrink_slots_next_batch {
                 shrink_slots.entry(slot).or_default().extend(stores);
             }
@@ -9943,10 +9943,11 @@ pub mod tests {
         let accounts = AccountsDb::new_single_for_tests();
 
         let mut candidates: ShrinkCandidates = HashMap::new();
-        let output_candidates =
+        let (output_candidates, next_candidates) =
             accounts.select_candidates_by_total_usage(&candidates, DEFAULT_ACCOUNTS_SHRINK_RATIO);
 
         assert_eq!(0, output_candidates.len());
+        assert_eq!(0, next_candidates.len());
 
         // case 2: two candidates, only one selected
         let dummy_path = Path::new("");
@@ -9957,7 +9958,7 @@ pub mod tests {
         let entry1 = Arc::new(AccountStorageEntry::new(
             dummy_path, dummy_slot, dummy_id1, dummy_size,
         ));
-        entry1.alive_bytes.store(8000, Ordering::Relaxed);
+        entry1.alive_bytes.store(0, Ordering::Relaxed);
 
         candidates
             .entry(dummy_slot)
@@ -9968,17 +9969,35 @@ pub mod tests {
         let entry2 = Arc::new(AccountStorageEntry::new(
             dummy_path, dummy_slot, dummy_id2, dummy_size,
         ));
-        entry2.alive_bytes.store(3000, Ordering::Relaxed);
+        entry2.alive_bytes.store(0, Ordering::Relaxed);
         candidates
             .entry(dummy_slot)
             .or_default()
             .insert(entry2.append_vec_id(), entry2.clone());
 
-        let output_candidates =
-            accounts.select_candidates_by_total_usage(&candidates, DEFAULT_ACCOUNTS_SHRINK_RATIO);
+        let dummy_id3 = 55;
+        let entry3 = Arc::new(AccountStorageEntry::new(
+            dummy_path, dummy_slot, dummy_id3, dummy_size,
+        ));
+        entry3.alive_bytes.store(1000, Ordering::Relaxed);
+        candidates
+            .entry(dummy_slot)
+            .or_default()
+            .insert(entry3.append_vec_id(), entry3.clone());
+
+        let (output_candidates, next_candidates) =
+            accounts.select_candidates_by_total_usage(&candidates, 0.17);
         assert_eq!(1, output_candidates.len());
         assert_eq!(1, output_candidates[&dummy_slot].len());
-        assert!(output_candidates[&dummy_slot].contains(&entry2.append_vec_id()));
+        assert!(
+            output_candidates[&dummy_slot].contains(&entry1.append_vec_id())
+                || output_candidates[&dummy_slot].contains(&entry2.append_vec_id())
+        );
+        assert_eq!(1, next_candidates.len());
+        assert!(
+            next_candidates[&dummy_slot].contains(&entry1.append_vec_id())
+                || next_candidates[&dummy_slot].contains(&entry2.append_vec_id())
+        );
 
         // case 3: two candidates, both are selected
         candidates.clear();
@@ -10009,7 +10028,7 @@ pub mod tests {
             .or_default()
             .insert(entry2.append_vec_id(), entry2.clone());
 
-        let output_candidates =
+        let (output_candidates, next_candidates) =
             accounts.select_candidates_by_total_usage(&candidates, DEFAULT_ACCOUNTS_SHRINK_RATIO);
         assert_eq!(2, output_candidates.len());
         assert_eq!(1, output_candidates[&dummy_slot].len());
@@ -10017,6 +10036,7 @@ pub mod tests {
 
         assert!(output_candidates[&dummy_slot].contains(&entry1.append_vec_id()));
         assert!(output_candidates[&dummy_slot2].contains(&entry2.append_vec_id()));
+        assert_eq!(0, next_candidates.len());
     }
 
     #[test]
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -2559,11 +2559,11 @@ impl AccountsDb {
         }
         measure.stop();
         inc_new_counter_info!(
-            "select_top_sparse_storage_entries-ms",
+            "shrink_select_top_sparse_storage_entries-ms",
             measure.as_ms() as usize
         );
-        inc_new_counter_info!("select_top_sparse_storage_entries-seeds", candidates_count);
-        inc_new_counter_info!("total_preliminary_candidate_stores", total_candidate_stores);
+        inc_new_counter_info!("shrink_select_top_sparse_storage_entries-seeds", candidates_count);
+        inc_new_counter_info!("shrink_total_preliminary_candidate_stores", total_candidate_stores);
 
         (shrink_slots, shrink_slots_next_batch)
     }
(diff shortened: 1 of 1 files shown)


runtime/src/accounts_db.rs
```diff
@@ -9948,17 +9994,35 @@ pub mod tests {
...
-        assert!(output_candidates[&dummy_slot].contains(&entry2.append_vec_id()));
+        assert!(
+            output_candidates[&dummy_slot].contains(&entry1.append_vec_id())
+                || output_candidates[&dummy_slot].contains(&entry2.append_vec_id())
+        );
+        assert_eq!(1, next_candidates.len());
+        assert!(
+            next_candidates[&dummy_slot].contains(&entry1.append_vec_id())
+                || next_candidates[&dummy_slot].contains(&entry2.append_vec_id())
+        );
```
carllin, line 10025: Hmmm why is this not deterministic between `entry1` and `entry2`, is it because their alive ratios are the same?

Could we make this more deterministic by making one of them have a greater alive ratio

runtime/src/accounts_db.rs
```diff
@@ -9948,17 +9994,35 @@ pub mod tests {
...
+            dummy_path, dummy_slot, dummy_id3, dummy_size,
+        ));
+        entry3.alive_bytes.store(1000, Ordering::Relaxed);
+        candidates
+            .entry(dummy_slot)
+            .or_default()
+            .insert(entry3.append_vec_id(), entry3.clone());
+
+        let (output_candidates, next_candidates) =
+            accounts.select_candidates_by_total_usage(&candidates, 0.17);
```
carllin, line 10014: A comment here on what the expected behavior is and why, along the lines of
       1) Listing alive ratio of each store
       2) the `shrink_ratio` of .17 and why it was chosen
       3) How a) and b) together lead to the expected output

would be very helpful for maintaining this test in the future.

runtime/src/accounts_db.rs
```diff
@@ -9948,17 +9994,35 @@ pub mod tests {
...
             .or_default()
             .insert(entry2.append_vec_id(), entry2.clone());
 
-        let output_candidates =
-            accounts.select_candidates_by_total_usage(&candidates, DEFAULT_ACCOUNTS_SHRINK_RATIO);
+        let dummy_id3 = 55;
+        let entry3 = Arc::new(AccountStorageEntry::new(
+            dummy_path, dummy_slot, dummy_id3, dummy_size,
+        ));
+        entry3.alive_bytes.store(1000, Ordering::Relaxed);
```
carllin, line 10007: the magic numbers here like `1000` are a bit hard to reason about, should this be a function of `dummy_size` and the previous `alive_bytes` in the previous entries.

runtime/src/accounts_db.rs
```diff
@@ -9948,17 +9994,35 @@ pub mod tests {
         let entry2 = Arc::new(AccountStorageEntry::new(
             dummy_path, dummy_slot, dummy_id2, dummy_size,
         ));
-        entry2.alive_bytes.store(3000, Ordering::Relaxed);
+        entry2.alive_bytes.store(0, Ordering::Relaxed);
```
carllin, line 9997: instead of `0`, maybe use `entry_2_alive_bytes`, then we can use this variable to calculate how many alive bytes are needed in `entry3`'s `entry_3_alive_bytes`, so that we can avoid the magic number here: https://github.com/solana-labs/solana/pull/19373/files#r697780367 as well.

runtime/src/accounts_db.rs
```diff
@@ -9923,10 +9968,11 @@ pub mod tests {
         let accounts = AccountsDb::new_single_for_tests();
 
         let mut candidates: ShrinkCandidates = HashMap::new();
-        let output_candidates =
+        let (output_candidates, next_candidates) =
             accounts.select_candidates_by_total_usage(&candidates, DEFAULT_ACCOUNTS_SHRINK_RATIO);
 
         assert_eq!(0, output_candidates.len());
+        assert_eq!(0, next_candidates.len());
```
carllin, line 9975: I think as this list of test cases grows, it's good to break these tests into separate tests with the prefix `test_select_candidates_{description of test}`

runtime/src/accounts_db.rs
```diff
@@ -9948,17 +9994,35 @@ pub mod tests {
...
         candidates
             .entry(dummy_slot)
             .or_default()
             .insert(entry2.append_vec_id(), entry2.clone());
 
-        let output_candidates =
-            accounts.select_candidates_by_total_usage(&candidates, DEFAULT_ACCOUNTS_SHRINK_RATIO);
+        let dummy_id3 = 55;
+        let entry3 = Arc::new(AccountStorageEntry::new(
+            dummy_path, dummy_slot, dummy_id3, dummy_size,
```
carllin, line 10005: these names are a bit hard to track. Maybe

```
dummy_path -> common_storage_path
dummy_slot -> common_shrink_slot
dummy_id3 -> storage_id3
dummy_size -> storage3_total_storage_size  (feel like this should be specific to each storage so we can use it to compute other variables below)
```

carllin:
A test for this `break` case: https://github.com/solana-labs/solana/blob/4d91d6922149725452dbc687b398e60bca8902d8/runtime/src/accounts_db.rs#L2541-L2548 where we have two storage entries that satisfy the `usage.alive_ratio < shrink_ratio` *after*  the total shrink ratio has already been hit https://github.com/solana-labs/solana/blob/4d91d6922149725452dbc687b398e60bca8902d8/runtime/src/accounts_db.rs#L2534 would be good.

I.e. if the shrink ratio is 50%, then

1. One storage entry with 1000 bytes and 10 alive bytes, so it gets prioritized first
2. Two storage entries each with 10 bytes and 5 alive bytes. Make sure *both* of these get picked for shrink

runtime/src/accounts_db.rs
```diff
@@ -2522,44 +2523,57 @@ impl AccountsDb {
...
+                alive_ratio, usage.store.append_vec_id(), usage.alive_ratio, shrink_ratio, total_bytes, total_alive_bytes);
             if alive_ratio > shrink_ratio {
                 // we have reached our goal, stop
                 debug!(
                     "Shrinking goal can be achieved at slot {:?}, total_alive_bytes: {:?} \
                     total_bytes: {:?}, alive_ratio: {:}, shrink_ratio: {:?}",
                     usage.slot, total_alive_bytes, total_bytes, alive_ratio, shrink_ratio
                 );
-                break;
+                if usage.alive_ratio < shrink_ratio {
```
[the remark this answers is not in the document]
carllin: also it's worth noting dead slots are removed from the shrink candidate set here: https://github.com/solana-labs/solana/blob/4d91d6922149725452dbc687b398e60bca8902d8/runtime/src/accounts_db.rs#L2202

runtime/src/accounts_db.rs
```diff
@@ -2579,6 +2597,16 @@ impl AccountsDb {
...
         );
         inc_new_counter_info!("shrink_all_candidate_slots-count", shrink_candidates_count);
+        let mut pended_counts: usize = 0;
+        if let Some(shrink_slots_next_batch) = shrink_slots_next_batch {
+            let mut shrink_slots = self.shrink_candidate_slots.lock().unwrap();
+            for (slot, stores) in shrink_slots_next_batch {
+                pended_counts += stores.len();
+                shrink_slots.entry(slot).or_default().extend(stores);
+            }
+        }
```
carllin, line 2607: Maybe a test for this as well that the stores are actually added back to the shrink candidates list and are triggered on the next shrink

runtime/src/accounts_db.rs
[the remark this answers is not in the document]
carllin: This shrink candidate set should be bounded by the number of alive bytes satisfying this condition here: https://github.com/solana-labs/solana/blob/4d91d6922149725452dbc687b398e60bca8902d8/runtime/src/accounts_db.rs#L2531-L2534. 

I.e. the only way  for the shrink candidate set to keep growing in size is if:
1.  there are a bunch of little stores that all satisfy `usage.alive_ratio < shrink_ratio`
2.  None of those little stores are dead, so still remain in the shrink candidate set.
3.  `alive_ratio > shrink_ratio` for the entire system, even without cleaning these little stores.

If zero lamport account cleaning is working properly, all slots should eventually be marked dead within an epoch so that should bound the number of candidates to 430K. The only time this wouldn't happens is if we have a chain of zero lamport accounts that takes *multiple* cleans to purge because they are mutually dependent.

In this latter case if zero-lamport accounts are not being cleaned up, I think the shrink candidate set can in worst case be at least as large as the number of zero-lamport keys, which in degenerate cases as we saw this week, could be millions

runtime/src/accounts_db.rs
```diff
@@ -4547,6 +4589,9 @@ impl AccountsDb {
                 self.recycle_stores.read().unwrap().entry_count() as u64,
                 i64
             ),
+            ("total_bytes", total_bytes, i64),
```
carllin, line 4592: do we need a metric for the number of shrink candidates as well?

runtime/src/accounts_db.rs
lijunwangs: Yes. It took me a while to come up with the number to trigger the condition.

runtime/src/accounts_db.rs
lijunwangs: Will address

runtime/src/accounts_db.rs
lijunwangs: sure

runtime/src/accounts_db.rs
lijunwangs: will do

lijunwangs:
I have measure overall AccountsDb alive ratio of two validators with and without the fix: The top line is the validator with the fix. It shows the alive_ratio degrades at much slower pace than the one without the fix. 
Screen Shot 2021-08-28 at 10 39 44 AM

lijunwangs:
Here is the graph comparing the total_preliminary_candidate_stores with and without the fixes. The top line is the one with the fix. As expected it has higher counts than without the fixes. The counts in both cases is not significant.

Screen Shot 2021-08-28 at 10 42 50 AM

lijunwangs:
The following is the graph of total AccountsDb store total_bytes for two validators: one with the fix and one without. The bottom one is the one with the fix. The total bytes grows at a less rate than the validator without the fixes.
Screen Shot 2021-08-28 at 10 49 30 AM

runtime/src/accounts_db.rs
lijunwangs: Fixed

runtime/src/accounts_db.rs
```diff
@@ -9916,87 +9967,229 @@ pub mod tests {
...
+        let store1_id = 22;
+        let store1 = Arc::new(AccountStorageEntry::new(
+            common_store_path,
+            slot1,
+            store1_id,
+            store_file_size,
+        ));
+
+        // store1 has 1 page-aligned alive bytes, its alive ratio is 1/4: 0.25
+        let store1_alive_bytes = 3 * 1024;
```
carllin, line 10152: can we just use `store_file_size / 4`?

runtime/src/accounts_db.rs
```diff
@@ -9916,87 +9967,229 @@ pub mod tests {
...
+        let slot2 = 44;
+        let store2 = Arc::new(AccountStorageEntry::new(
+            common_store_path,
+            slot2,
+            store2_id,
+            store_file_size,
+        ));
+
+        // store2 has 2 page-aligned bytes, its alive ratio is 2/4: 0.5
+        let store2_alive_bytes = 5 * 1024;
```
carllin, line 10172: Can we just do `can we just use store_file_size / 2?

lijunwangs: I intentionally keep them not the rounded to the page size so that we are examining the logic is doing page aligned calculations

runtime/src/accounts_db.rs
```diff
@@ -9916,87 +9967,229 @@ pub mod tests {
...
 
-        let dummy_id2 = 44;
-        let dummy_slot2 = 44;
-        let entry2 = Arc::new(AccountStorageEntry::new(
-            dummy_path,
-            dummy_slot2,
-            dummy_id2,
-            dummy_size,
+        // The store2's alive_ratio is 0.5: as its page aligned alive size is 1 page.
+        let store2_alive_bytes = 1024;
```
carllin, line 10093: can we just use PAGE_SIZE - 1?

runtime/src/accounts_db.rs
```diff
@@ -9916,87 +9967,229 @@ pub mod tests {
...
+        let entry3 = Arc::new(AccountStorageEntry::new(
+            common_store_path,
+            common_slot_id,
+            store3_id,
+            store_file_size,
         ));
-        entry2.alive_bytes.store(3000, Ordering::Relaxed);
+
+        // The store3's alive ratio is 1.0 as its page-aligned alive size is 2 pages
+        let store3_alive_bytes = 1024 * 7;
```
carllin, line 10111: can we just use PAGE_SIZE + 1?

runtime/src/accounts_db.rs
```diff
@@ -9916,87 +9967,229 @@ pub mod tests {
...
-        assert_eq!(1, output_candidates[&dummy_slot].len());
-        assert!(output_candidates[&dummy_slot].contains(&entry2.append_vec_id()));
-
-        // case 3: two candidates, both are selected
-        candidates.clear();
-        let dummy_size = 4 * PAGE_SIZE;
-        let dummy_id1 = 22;
-        let entry1 = Arc::new(AccountStorageEntry::new(
-            dummy_path, dummy_slot, dummy_id1, dummy_size,
+        // Set the target alive ratio to 0.6 so that we can just get rid of store1, the remaing two stores
```
carllin, line 10045: nit: `remaing` -> `remaining`

runtime/src/accounts_db.rs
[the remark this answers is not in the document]
lijunwangs: Thanks for the analysis Carl!

runtime/src/accounts_db.rs
lijunwangs: Addressed

runtime/src/accounts_db.rs
lijunwangs: Good idea

runtime/src/accounts_db.rs
lijunwangs: Done

runtime/src/accounts_db.rs
lijunwangs: Done

runtime/src/accounts_db.rs
lijunwangs: Done

runtime/src/accounts_db.rs
lijunwangs: Done

lijunwangs:
> > Can also see a comparison in total accounts size before/after this change? and bytes shrunk/s?
> 
> Will do
See the graphs posted

runtime/src/accounts_db.rs
lijunwangs: Fixed

lijunwangs:
> A test for this `break` case:
> 
> https://github.com/solana-labs/solana/blob/4d91d6922149725452dbc687b398e60bca8902d8/runtime/src/accounts_db.rs#L2541-L2548
> 
> where we have two storage entries that satisfy the `usage.alive_ratio < shrink_ratio` _after_ the total shrink ratio has already been hit
> https://github.com/solana-labs/solana/blob/4d91d6922149725452dbc687b398e60bca8902d8/runtime/src/accounts_db.rs#L2534
> 
> would be nice, as it's a bit subtle
> I.e. if the shrink ratio is 50%, then
> 
> 1. One storage entry with 1000 bytes and 10 alive bytes, so it gets prioritized first
> 2. Two storage entries each with 10 bytes and 5 alive bytes. Make sure _both_ of these get picked for shrink

test_select_candidates_by_total_usage_3_way_split_condition covers the break case.

carllin:
@lijunwangs looks great, thanks for all the cleanup and diligence! 😃 

Last question is if we want to have a failsafe for this case: https://github.com/solana-labs/solana/pull/19373/files#r697786319. Just in case we introduce something that makes the shrink set grow very large, as illustrated by my long-winded comment here: https://github.com/solana-labs/solana/pull/19373#discussion_r697786319

lijunwangs:
> @lijunwangs looks great, thanks for all the cleanup and diligence! 😃
> 
> Last question is if we want to have a failsafe for this case: https://github.com/solana-labs/solana/pull/19373/files#r697786319. Just in case we introduce something that makes the shrink set grow very large, as illustrated by my long-winded comment here: [#19373 (comment)](https://github.com/solana-labs/solana/pull/19373#discussion_r697786319)

From the capture of the total input candidates count (before we filter them down via total space optimization)  -- the count is really small as shown in the graph https://github.com/solana-labs/solana/pull/19373#issuecomment-907663006.  Another way to reason about this is with the reasonable alive ratio 80% -- we would have to work on even larger number of stores to shrink to counter them. And in the next round it will be likely become the candidates to be selected for cleanup and hence dropped from the queue.

carllin:
> > @lijunwangs looks great, thanks for all the cleanup and diligence! 😃
> > Last question is if we want to have a failsafe for this case: https://github.com/solana-labs/solana/pull/19373/files#r697786319. Just in case we introduce something that makes the shrink set grow very large, as illustrated by my long-winded comment here: [#19373 (comment)](https://github.com/solana-labs/solana/pull/19373#discussion_r697786319)
> 
> From the capture of the total input candidates count (before we filter them down via total space optimization) -- the count is really small as shown in the graph [#19373 (comment)](https://github.com/solana-labs/solana/pull/19373#issuecomment-907663006). Another way to reason about this is with the reasonable alive ratio 80% -- we would have to work on even larger number of stores to shrink to counter them. And in the next round it will be likely become the candidates to be selected for cleanup and hence dropped from the queue.

Oh I was imagining a situation where if there's a bug in the zero lamport cleaning, you can have a bunch of very small stores all less than the alive ratio. For instance, let's say each store has one zero lamport account with only a few bytes of data, and one dead account with an equal number bytes of data, so then this store is 79% < 80% alive.

Now if there are bigger stores that are are only 50% full, because we sort by fullness, we always shrink those big ones first and then because we've met the quota, we add these small stores back to the queue (assuming these small stores' total size is very small compared to the total account size). 

But this is an edge case that assumes there's a bug with zero-lamport cleaning, so I'm fine merging this as is with any follow-ups if necessary.

carllin (approved):

lijunwangs:
> > > @lijunwangs looks great, thanks for all the cleanup and diligence! 😃
> > > Last question is if we want to have a failsafe for this case: https://github.com/solana-labs/solana/pull/19373/files#r697786319. Just in case we introduce something that makes the shrink set grow very large, as illustrated by my long-winded comment here: [#19373 (comment)](https://github.com/solana-labs/solana/pull/19373#discussion_r697786319)
> > 
> > 
> > From the capture of the total input candidates count (before we filter them down via total space optimization) -- the count is really small as shown in the graph [#19373 (comment)](https://github.com/solana-labs/solana/pull/19373#issuecomment-907663006). Another way to reason about this is with the reasonable alive ratio 80% -- we would have to work on even larger number of stores to shrink to counter them. And in the next round it will be likely become the candidates to be selected for cleanup and hence dropped from the queue.
> 
> Oh I was imagining a situation where if there's a bug in the zero lamport cleaning, you can have a bunch of very small stores all less than the alive ratio. For instance, let's say each store has one zero lamport account with only a few bytes of data, and one dead account with an equal number bytes of data, so then this store is 79% < 80% alive.
> 
> Now if there are bigger stores that are are only 50% full, because we sort by fullness, we always shrink those big ones first and then because we've met the quota, we add these small stores back to the queue (assuming these small stores' total size is very small compared to the total account size).
> 
> But this is an edge case that assumes there's a bug with zero-lamport cleaning, so I'm fine merging this as is with any follow-ups if necessary.

Thanks @carllin  for your thoughtful review -- I got your point that under very skewed condition it could happen. We could bound on that count of max number stores deferred and let the extra one go through regular shrink -- I will raise an issue and have it addressed in a separate PR.

PR #19385, opened 2021-08-23
in clean, keep ref count and account info in sync wrt reclaims

jeffwashington:
#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -7862,20 +7862,24 @@ pub mod tests {
         let pubkey2 = solana_sdk::pubkey::new_rand();
         let zero_lamport_account =
             AccountSharedData::new(0, 0, AccountSharedData::default().owner());
+        let non_zero_lamport_account =
+            AccountSharedData::new(1, 0, AccountSharedData::default().owner());
 
         // Store 2 accounts in slot 0, then update account 1 in two more slots
         accounts.store_uncached(0, &[(&pubkey1, &zero_lamport_account)]);
         accounts.store_uncached(0, &[(&pubkey2, &zero_lamport_account)]);
         accounts.store_uncached(1, &[(&pubkey1, &zero_lamport_account)]);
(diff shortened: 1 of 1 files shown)


jeffwashington:
@carllin @lijunwangs @sakridge 
Here's a theory. This cleans immediately to 432k slots on the snapshot I got.

jeffwashington:
Here's the 1.6 version of this:
#19384

jeffwashington:
This code causes us to not do the clean because the account info list is not the same len as the refcount:
https://github.com/solana-labs/solana/blob/0b3fad19f40dd8e444254007fc2e5986f1c32b7d/runtime/src/accounts_db.rs#L1562-L1571

runtime/src/accounts_db.rs
```diff
@@ -1944,6 +1944,7 @@ impl AccountsDb {
                     .map(|store_removed| store_removed.contains(&account_info.offset))
                     .unwrap_or(false);
                 if was_reclaimed {
+                    *ref_count -= 1; // since we aren't including this account info, we also have to decrement the expected refcount so the list len matches the expected ref count
```
carllin, line 1947: hmmm I don't think you want to decrement the ref count here because the `was_reclaimed` means the entry for this `(pubkey, slot)` was removed from the AccountsIndex, but still exists in the storage entry until a shrink happens

carllin:
First of note is that the refcount tracks the number of slot storage entries in which a pubkey exists, it doesn't track the number of entries in the AccountsIndex (you can be removed from the index and still exist in the storage entry until shrink)

> Note that the refcount was grabbed early in clean, during the scan:

The refcount is also updated to account for any dead slots (which completely removes the storage entry for those slots, hence changing the refcount): https://github.com/solana-labs/solana/blob/0b3fad19f40dd8e444254007fc2e5986f1c32b7d/runtime/src/accounts_db.rs#L1927-L1929

PR #19387, opened 2021-08-23
Add parameter to allow setting max-retries for SendTransaction rpc

### LINKED ISSUE #19347: SendTransactionService retries all transactions indiscriminately
#### Problem
Some transactions are only relevant if they execute promptly, but an rpc node will retry every transaction it receives until the recent blockhash expires, potentially wasting rpc and validator banking-stage cycles.

#### Proposed Solution
Consider adding a no-retry parameter to `sendTransaction` rpc, directing the SendTransactionService to only attempt to submit the transaction to upcoming leaders once.

Note that such a transaction would still be valid until the blockhash expires; the transaction could still land due to validator forwarding or a duplicate submission. If guaranteed reduced liveness is needed, it would be incumbent on the user to include on-chain program logic to reject transactions to enforce the expiration.
Using the no-retry parameter would be a courtesy to the rpc node and the cluster.

CriesofCarrots:
#### Problem
Some transactions are only relevant if they execute promptly, but an rpc node will retry every transaction it receives until the recent blockhash expires, potentially wasting rpc and validator banking-stage cycles.

#### Summary of Changes
Add parameter to allow setting max-retries for SendTransaction rpc
Add tracking in SendTransactionService for transactions dropped due to reaching max-retries

Fixes #19347

### DIFF
--- a/banks-server/src/banks_server.rs
+++ b/banks-server/src/banks_server.rs
@@ -167,6 +167,7 @@ impl Banks for BanksServer {
             serialize(&transaction).unwrap(),
             last_valid_block_height,
             None,
+            None,
         );
         self.transaction_sender.send(info).unwrap();
     }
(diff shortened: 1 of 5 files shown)


mvines (approved):
lgtm with JSON RPC API documentation updated

PR #19400, opened 2021-08-24
Banking Stage drop too_old votes

### LINKED ISSUE #17877: Cost model to handle redundant votes

taozhu-chicago:
#### Problem
banking_stage inserts all Vote transactions, includes `TooOld` votes, into bank. Those redundant votes should be dropped.

#### Summary of Changes
1. add check_redundant_votes function to check vote by current bank's vote_state;

2. banking_stage calls above function after accounts were locked, and before
   calling bank.load_and_execute_transactions, redundant vote transactions
   are marked as TransactionError::ProcessedAlready in result;

Fixes #17877

### DIFF
--- a/core/src/banking_stage.rs
+++ b/core/src/banking_stage.rs
@@ -1,7 +1,7 @@
 //! The `banking_stage` processes Transaction messages. It is intended to be used
 //! to contruct a software pipeline. The stage uses all available CPU cores and
 //! can do its processing in parallel with signature verification on the GPU.
-use crate::{cost_tracker::CostTracker, packet_hasher::PacketHasher};
+use crate::{cost_tracker::CostTracker, packet_hasher::PacketHasher, vote_redundancy_checker};
 use crossbeam_channel::{Receiver as CrossbeamReceiver, RecvTimeoutError};
 use itertools::Itertools;
 use lru::LruCache;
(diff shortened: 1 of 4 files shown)


taozhu-chicago:
Create this PR from #19152 to check why CI fails on Stable test, which succeeds locally.

taozhu-chicago:
this is a duplicate of #19152

PR #19401, opened 2021-08-24
Make background services aware of incremental snapshots

### LINKED ISSUE #19167: Make AccountsHashVerifier make aware of Incremental Snapshots

### LINKED ISSUE #19168: Make AccountsBackgroundService make aware of Incremental Snapshots
`AccountsBackgroundService` needs to know about incremental snapshots to make sure that when it calls `clean_accounts()` that it passes the correct value for `last_full_snapshot_slot`. Additionally, hopefully will be able to filter the correct snapshot storages here before passing along to the AccountsHashVerifier.

brooksprumo:
AccountsBackgroundService now knows about incremental snapshots.  It is
now also in charge of deciding if an AccountsPackage is destined to be a
SnapshotPackage or not (or just used by AccountsHashVerifier).

!!! New behavior changes !!!

Full snapshots (both bank and archive) **MUST** succeed.

This is required because of how the last full snapshot slot is
calculated, which is used by AccountsBackgroundService when calling
`clean_accounts()`.

File system calls are now unwrapped and will result in a crash. As Trent told me:

>Well I think if a snapshot fails due to some IO error, it's very likely that the operator is going to have to intervene before it works.  We should exit error in this case, otherwise the validator might happily spin for several more hours, never successfully writing a complete snapshot, before something else brings it down.  This would leave the validator's last local snapshot many more slots behind than it would be had we exited outright and potentially force the operator to abandon ledger continuity in favor of a quick catchup

Fixes #19167 
Fixes #19168

### DIFF
--- a/core/src/accounts_hash_verifier.rs
+++ b/core/src/accounts_hash_verifier.rs
@@ -96,7 +96,7 @@ impl AccountsHashVerifier {
         pending_snapshot_package: Option<&PendingSnapshotPackage>,
         hashes: &mut Vec<(Slot, Hash)>,
         exit: &Arc<AtomicBool>,
-        fault_injection_rate_slots: Slot,
+        fault_injection_rate_slots: u64,
         snapshot_config: Option<&SnapshotConfig>,
         thread_pool: Option<&ThreadPool>,
     ) {
@@ -148,7 +148,7 @@ impl AccountsHashVerifier {
         halt_on_trusted_validator_accounts_hash_mismatch: bool,
         hashes: &mut Vec<(Slot, Hash)>,
         exit: &Arc<AtomicBool>,
-        fault_injection_rate_slots: Slot,
+        fault_injection_rate_slots: u64,
     ) {
         let hash = accounts_package.hash;
         if fault_injection_rate_slots != 0
--- a/core/src/accounts_hash_verifier.rs
+++ b/core/src/accounts_hash_verifier.rs
@@ -207,10 +207,7 @@ impl AccountsHashVerifier {
                 .unwrap()
                 .as_ref()
                 .map_or(true, |snapshot_package| {
-                    matches!(
-                        snapshot_package.snapshot_type,
-                        SnapshotType::IncrementalSnapshot(_)
-                    )
+                    snapshot_package.snapshot_type.is_incremental_snapshot()
                 }),
         };
 
--- a/core/src/accounts_hash_verifier.rs
+++ b/core/src/accounts_hash_verifier.rs
@@ -41,7 +41,7 @@ impl AccountsHashVerifier {
         cluster_info: &Arc<ClusterInfo>,
         trusted_validators: Option<HashSet<Pubkey>>,
         halt_on_trusted_validators_accounts_hash_mismatch: bool,
-        fault_injection_rate_slots: Slot,
+        fault_injection_rate_slots: u64,
         snapshot_config: Option<SnapshotConfig>,
     ) -> Self {
         let exit = exit.clone();
--- a/core/src/accounts_hash_verifier.rs
+++ b/core/src/accounts_hash_verifier.rs
@@ -6,14 +6,16 @@
 
 use rayon::ThreadPool;
 use solana_gossip::cluster_info::{ClusterInfo, MAX_SNAPSHOT_HASHES};
+use solana_measure::measure::Measure;
 use solana_runtime::{
-    accounts_db,
-    snapshot_archive_info::SnapshotArchiveInfoGetter,
+    accounts_db::{self, AccountsDb},
+    accounts_hash::HashStats,
     snapshot_config::SnapshotConfig,
     snapshot_package::{
         AccountsPackage, AccountsPackageReceiver, PendingSnapshotPackage, SnapshotPackage,
+        SnapshotType,
     },
-    snapshot_utils,
+    sorted_storages::SortedStorages,
 };
 use solana_sdk::{clock::Slot, hash::Hash, pubkey::Pubkey};
 use std::collections::{HashMap, HashSet};
@@ -39,7 +41,7 @@ impl AccountsHashVerifier {
         cluster_info: &Arc<ClusterInfo>,
         trusted_validators: Option<HashSet<Pubkey>>,
         halt_on_trusted_validators_accounts_hash_mismatch: bool,
-        fault_injection_rate_slots: u64,
+        fault_injection_rate_slots: Slot,
         snapshot_config: Option<SnapshotConfig>,
     ) -> Self {
         let exit = exit.clone();
@@ -48,7 +50,7 @@ impl AccountsHashVerifier {
             .name("solana-hash-accounts".to_string())
             .spawn(move || {
                 let mut hashes = vec![];
-                let mut thread_pool_storage = None;
+                let mut thread_pool = None;
                 loop {
                     if exit.load(Ordering::Relaxed) {
                         break;
@@ -56,11 +58,9 @@ impl AccountsHashVerifier {
 
                     match accounts_package_receiver.recv_timeout(Duration::from_secs(1)) {
                         Ok(accounts_package) => {
-                            if accounts_package.hash_for_testing.is_some()
-                                && thread_pool_storage.is_none()
+                            if accounts_package.hash_for_testing.is_some() && thread_pool.is_none()
                             {
-                                thread_pool_storage =
-                                    Some(accounts_db::make_min_priority_thread_pool());
+                                thread_pool = Some(accounts_db::make_min_priority_thread_pool());
                             }
 
                             Self::process_accounts_package(
@@ -73,7 +73,7 @@ impl AccountsHashVerifier {
                                 &exit,
                                 fault_injection_rate_slots,
                                 snapshot_config.as_ref(),
-                                thread_pool_storage.as_ref(),
+                                thread_pool.as_ref(),
                             );
                         }
                         Err(RecvTimeoutError::Disconnected) => break,
@@ -96,49 +96,73 @@ impl AccountsHashVerifier {
         pending_snapshot_package: Option<&PendingSnapshotPackage>,
         hashes: &mut Vec<(Slot, Hash)>,
         exit: &Arc<AtomicBool>,
-        fault_injection_rate_slots: u64,
+        fault_injection_rate_slots: Slot,
         snapshot_config: Option<&SnapshotConfig>,
         thread_pool: Option<&ThreadPool>,
     ) {
-        let snapshot_package =
-            snapshot_utils::process_accounts_package(accounts_package, thread_pool, None);
-        Self::process_snapshot_package(
-            snapshot_package,
+        Self::verify_accounts_package_hash(&accounts_package, thread_pool);
+
+        Self::push_accounts_hashes_to_cluster(
+            &accounts_package,
             cluster_info,
             trusted_validators,
             halt_on_trusted_validator_accounts_hash_mismatch,
-            pending_snapshot_package,
             hashes,
             exit,
             fault_injection_rate_slots,
-            snapshot_config,
+        );
+
+        Self::submit_for_packaging(accounts_package, pending_snapshot_package, snapshot_config);
+    }
+
+    fn verify_accounts_package_hash(
+        accounts_package: &AccountsPackage,
+        thread_pool: Option<&ThreadPool>,
+    ) {
+        let mut measure_hash = Measure::start("hash");
+        if let Some(expected_hash) = accounts_package.hash_for_testing {
+            let sorted_storages = SortedStorages::new(&accounts_package.snapshot_storages);
+            let (hash, lamports) = AccountsDb::calculate_accounts_hash_without_index(
+                &sorted_storages,
+                thread_pool,
+                HashStats::default(),
+                false,
+                None,
+            )
+            .unwrap();
+
+            assert_eq!(accounts_package.expected_capitalization, lamports);
+            assert_eq!(expected_hash, hash);
+        };
+        measure_hash.stop();
+        datapoint_info!(
+            "accounts_hash_verifier",
+            ("calculate_hash", measure_hash.as_us(), i64),
         );
     }
 
-    fn process_snapshot_package(
-        snapshot_package: SnapshotPackage,
+    fn push_accounts_hashes_to_cluster(
+        accounts_package: &AccountsPackage,
         cluster_info: &ClusterInfo,
         trusted_validators: Option<&HashSet<Pubkey>>,
         halt_on_trusted_validator_accounts_hash_mismatch: bool,
-        pending_snapshot_package: Option<&PendingSnapshotPackage>,
         hashes: &mut Vec<(Slot, Hash)>,
         exit: &Arc<AtomicBool>,
-        fault_injection_rate_slots: u64,
-        snapshot_config: Option<&SnapshotConfig>,
+        fault_injection_rate_slots: Slot,
     ) {
-        let hash = *snapshot_package.hash();
+        let hash = accounts_package.hash;
         if fault_injection_rate_slots != 0
-            && snapshot_package.slot() % fault_injection_rate_slots == 0
+            && accounts_package.slot % fault_injection_rate_slots == 0
         {
             // For testing, publish an invalid hash to gossip.
             use rand::{thread_rng, Rng};
             use solana_sdk::hash::extend_and_hash;
-            warn!("inserting fault at slot: {}", snapshot_package.slot());
+            warn!("inserting fault at slot: {}", accounts_package.slot);
             let rand = thread_rng().gen_range(0, 10);
             let hash = extend_and_hash(&hash, &[rand]);
-            hashes.push((snapshot_package.slot(), hash));
+            hashes.push((accounts_package.slot, hash));
         } else {
-            hashes.push((snapshot_package.slot(), hash));
+            hashes.push((accounts_package.slot, hash));
         }
 
         while hashes.len() > MAX_SNAPSHOT_HASHES {
@@ -155,17 +179,44 @@ impl AccountsHashVerifier {
             }
         }
 
-        if let Some(snapshot_config) = snapshot_config {
-            if snapshot_package.block_height % snapshot_config.full_snapshot_archive_interval_slots
-                == 0
-            {
-                if let Some(pending_snapshot_package) = pending_snapshot_package {
-                    *pending_snapshot_package.lock().unwrap() = Some(snapshot_package);
-                }
-            }
-        }
-
         cluster_info.push_accounts_hashes(hashes.clone());
+    }
+
+    fn submit_for_packaging(
+        accounts_package: AccountsPackage,
+        pending_snapshot_package: Option<&PendingSnapshotPackage>,
+        snapshot_config: Option<&SnapshotConfig>,
+    ) {
+        if accounts_package.snapshot_type.is_none()
+            || pending_snapshot_package.is_none()
+            || snapshot_config.is_none()
+        {
+            return;
+        };
+
+        let snapshot_package = SnapshotPackage::from(accounts_package);
+        let pending_snapshot_package = pending_snapshot_package.unwrap();
+        let _snapshot_config = snapshot_config.unwrap();
+
+        // If the snapshot package is an Incremental Snapshot, do not submit it if there's already
+        // a pending Full Snapshot.
+        let can_submit = match snapshot_package.snapshot_type {
+            SnapshotType::FullSnapshot => true,
+            SnapshotType::IncrementalSnapshot(_) => pending_snapshot_package
+                .lock()
+                .unwrap()
+                .as_ref()
+                .map_or(true, |snapshot_package| {
+                    matches!(
+                        snapshot_package.snapshot_type,
+                        SnapshotType::IncrementalSnapshot(_)
+                    )
+                }),
+        };
+
+        if can_submit {
+            *pending_snapshot_package.lock().unwrap() = Some(snapshot_package);
+        }
     }
 
     fn should_halt(
@@ -225,10 +276,10 @@ mod tests {
     use solana_gossip::{cluster_info::make_accounts_hashes_message, contact_info::ContactInfo};
     use solana_runtime::{
         snapshot_config::LastFullSnapshotSlot,
-        snapshot_package::SnapshotType,
         snapshot_utils::{ArchiveFormat, SnapshotVersion},
     };
     use solana_sdk::{
+        genesis_config::ClusterType,
         hash::hash,
         signature::{Keypair, Signer},
     };
@@ -301,30 +352,24 @@ mod tests {
             last_full_snapshot_slot: LastFullSnapshotSlot::default(),
         };
         for i in 0..MAX_SNAPSHOT_HASHES + 1 {
-            let slot = full_snapshot_archive_interval_slots + i as u64;
-            let block_height = full_snapshot_archive_interval_slots + i as u64;
-            let slot_deltas = vec![];
-            let snapshot_links = TempDir::new().unwrap();
-            let storages = vec![];
-            let snapshot_archive_path = PathBuf::from(".");
-            let hash = hash(&[i as u8]);
-            let archive_format = ArchiveFormat::TarBzip2;
-            let snapshot_version = SnapshotVersion::default();
-            let snapshot_package = SnapshotPackage::new(
-                slot,
-                block_height,
-                slot_deltas,
-                snapshot_links,
-                storages,
-                snapshot_archive_path,
-                hash,
-                archive_format,
-                snapshot_version,
-                SnapshotType::FullSnapshot,
-            );
+            let accounts_package = AccountsPackage {
+                slot: full_snapshot_archive_interval_slots + i as u64,
+                block_height: full_snapshot_archive_interval_slots + i as u64,
+                slot_deltas: vec![],
+                snapshot_links: TempDir::new().unwrap(),
+                snapshot_storages: vec![],
+                hash: hash(&[i as u8]),
+                archive_format: ArchiveFormat::TarBzip2,
+                snapshot_version: SnapshotVersion::default(),
+                snapshot_archives_dir: PathBuf::default(),
+                expected_capitalization: 0,
+                hash_for_testing: None,
+                cluster_type: ClusterType::MainnetBeta,
+                snapshot_type: None,
+            };
 
-            AccountsHashVerifier::process_snapshot_package(
-                snapshot_package,
+            AccountsHashVerifier::process_accounts_package(
+                accounts_package,
                 &cluster_info,
                 Some(&trusted_validators),
                 false,
@@ -333,7 +378,9 @@ mod tests {
                 &exit,
                 0,
                 Some(&snapshot_config),
+                None,
             );
+
             // sleep for 1ms to create a newer timestmap for gossip entry
             // otherwise the timestamp won't be newer.
             std::thread::sleep(Duration::from_millis(1));
--- a/core/src/snapshot_packager_service.rs
+++ b/core/src/snapshot_packager_service.rs
@@ -14,8 +14,6 @@ use std::{
     time::Duration,
 };
 
-const ARCHIVE_SNAPSHOT_PACKAGE_MAX_ATTEMPTS: usize = 10;
-
 pub struct SnapshotPackagerService {
     t_snapshot_packager: JoinHandle<()>,
 }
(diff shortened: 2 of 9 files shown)


runtime/src/accounts_background_service.rs
```diff
@@ -169,23 +174,60 @@ impl SnapshotRequestHandler {
...
+                } else if block_height
+                    % self
+                        .snapshot_config
+                        .incremental_snapshot_archive_interval_slots
+                    == 0
+                {
+                    Some(SnapshotType::IncrementalSnapshot(last_full_snapshot_slot))
+                } else {
+                    None
+                };
```
brooksprumo, line 192: Now AccountsBackgroundService is the one responsible for setting the snapshot type (previously this was in AccountsHashVerifier).

Note that the `None` here is for AccountsPackages that will _not_ be archived, which is possible if the incremental snapshot interval does _not_ equal the accounts hash interval.

runtime/src/accounts_background_service.rs
```diff
@@ -169,23 +174,60 @@ impl SnapshotRequestHandler {
...
+                    // continue on.
+                    if snapshot_type != Some(SnapshotType::FullSnapshot) {
+                        break;
+                    }
+
+                    assert!(
+                        attempts <= SNAPSHOT_BANK_MAX_ATTEMPTS,
+                        "Unable to snapshot bank intended for full snapshot!"
                     );
                 }
```
brooksprumo, line 232: This new loop is to retry full snapshots. They must succeed since we are calculating the last full snapshot slot instead of having SnapshotPackagerService tell us.

Notice that there's a `clone()` on the `status_cache_slot_deltas` now, which is a bummer. Maybe there's another way to do this?

runtime/src/accounts_background_service.rs
```diff
@@ -155,12 +156,16 @@ impl SnapshotRequestHandler {
...
+                    snapshot_root_bank.slot(),
+                    self.snapshot_config.full_snapshot_archive_interval_slots,
+                );
                 let mut clean_time = Measure::start("clean_time");
                 // Don't clean the slot we're snapshotting because it may have zero-lamport
                 // accounts that were included in the bank delta hash when the bank was frozen,
                 // and if we clean them here, the newly created snapshot's hash may not match
                 // the frozen hash.
-                snapshot_root_bank.clean_accounts(true, false, None);
+                snapshot_root_bank.clean_accounts(true, false, Some(last_full_snapshot_slot));
```
brooksprumo, line 168: Calculate the last full snapshot slot and use it when calling `clean_accounts()`. This happens again later in the file too.

core/src/accounts_hash_verifier.rs
```diff
@@ -155,19 +179,46 @@ impl AccountsHashVerifier {
...
+        if accounts_package.snapshot_type.is_none()
+            || pending_snapshot_package.is_none()
+            || snapshot_config.is_none()
+        {
+            return;
+        };
+
+        let snapshot_package = SnapshotPackage::from(accounts_package);
+        let pending_snapshot_package = pending_snapshot_package.unwrap();
+        let _snapshot_config = snapshot_config.unwrap();
```
brooksprumo, line 199: I may be able to remove SnapshotConfig from AccountsHashVerifier entirely too. I'm thinking about taking some things out of AccountsPackage the live in SnapshotConfig, which I may then use here, so I'm going to leave the SnapshotConfig here for now.

core/src/snapshot_packager_service.rs
```diff
@@ -42,24 +45,46 @@ impl SnapshotPackagerService {
...
+                        // continue on.
+                        if snapshot_package.snapshot_type != SnapshotType::FullSnapshot {
+                            break;
+                        }
+
+                        assert!(
+                            attempts <= ARCHIVE_SNAPSHOT_PACKAGE_MAX_ATTEMPTS,
+                            "Unable to archive full snapshot package!"
+                        );
                     }
```
brooksprumo, line 88: Here's the new loop where we retry archiving full snapshot packages if there's a failure. Is `assert`ing the best option here? Will that bring down the whole node, or just this thread?

core/src/accounts_hash_verifier.rs
```diff
@@ -96,49 +96,73 @@ impl AccountsHashVerifier {
...
+                HashStats::default(),
+                false,
+                None,
+            )
+            .unwrap();
+
+            assert_eq!(accounts_package.expected_capitalization, lamports);
+            assert_eq!(expected_hash, hash);
+        };
+        measure_hash.stop();
```
brooksprumo, line 137: This used to be in `snapshot_utils::process_accounts_package()`, but nothing used it except for AccountsHashVerifier. So, moving it here. As a plus, now we don't need to pass a TheadPool into other snapshot_utils functions.

core/tests/snapshots.rs
```diff
@@ -255,14 +257,14 @@ mod tests {
...
-            let mut bank = Bank::new_from_parent(&bank_forks[slot], &Pubkey::default(), slot + 1);
+        for slot in 1..=last_slot {
+            let mut bank = Bank::new_from_parent(&bank_forks[slot - 1], &Pubkey::default(), slot);
             f(&mut bank, mint_keypair);
             let bank = bank_forks.insert(bank);
             // Set root to make sure we don't end up with too many account storage entries
             // and to allow snapshotting of bank and the purging logic on status_cache to
             // kick in
-            if slot % set_root_interval == 0 || slot == last_slot - 1 {
+            if slot % set_root_interval == 0 || slot == last_slot {
```
brooksprumo, line 267: Had to change around the slot numbers so they were proper multiples of the snapshot intervals. Previously the check on line 267 (265 old) was true on `slot = interval + 1`, which mean the bank snapshot would never be sent to be archived.

core/tests/snapshots.rs
```diff
@@ -1019,9 +1010,14 @@ mod tests {
...
+        assert_eq!(deserialized_bank.slot(), LAST_SLOT,);
         assert_eq!(
-            deserialized_bank.slot(),
-            EXPECTED_SLOT_FOR_LAST_SNAPSHOT_ARCHIVE
+            deserialized_bank,
+            **bank_forks
+                .read()
+                .unwrap()
+                .get(deserialized_bank.slot())
+                .unwrap()
```
brooksprumo, line 1020: Yay! Now we check the banks are equal too!

runtime/src/snapshot_package.rs
```diff
@@ -228,18 +180,10 @@ impl SnapshotArchiveInfoGetter for SnapshotPackage {
...
-
-impl SnapshotType {
-    /// Get the string prefix of the snapshot type
-    pub fn to_prefix(&self) -> &'static str {
-        match self {
-            SnapshotType::FullSnapshot => TMP_FULL_SNAPSHOT_PREFIX,
-            SnapshotType::IncrementalSnapshot => TMP_INCREMENTAL_SNAPSHOT_PREFIX,
-        }
-    }
+    IncrementalSnapshot(Slot),
```
brooksprumo, line 188: By putting the `Slot` in the `IncrementalSnapshot`, I can always guarantee there is a last full snapshot if the snapshot type is Incremental. Otherwise I'd have asserts all over the place to ensure that was the case.

runtime/src/snapshot_package.rs
```diff
@@ -228,18 +180,10 @@ impl SnapshotArchiveInfoGetter for SnapshotPackage {
...
 }
```
brooksprumo, line 189: While looking through the actual archive and unarchive functions more, I realized that the prefix didn't need to be unique between full and incremental snapshots. So that just makes this a bit simpler.

runtime/src/snapshot_utils.rs
```diff
@@ -1538,29 +1535,45 @@ pub fn snapshot_bank(
...
-    let storages = root_bank.get_snapshot_storages(None);
+    let snapshot_storages = snapshot_type.map_or_else(SnapshotStorages::default, |snapshot_type| {
+        let incremental_snapshot_base_slot = match snapshot_type {
+            SnapshotType::IncrementalSnapshot(incremental_snapshot_base_slot) => {
+                Some(incremental_snapshot_base_slot)
+            }
+            _ => None,
+        };
+        root_bank.get_snapshot_storages(incremental_snapshot_base_slot)
+    });
```
brooksprumo, line 1550: Based on the snapshot type, get the snapshot storages. In the `None` case, that means the "snapshot" is not meant to be archived, and will only be used to send to AccountsHashVerifier to push the hash out to the cluster. In that case, we don't need to snapshot any of the storages even.

runtime/src/snapshot_utils.rs
```diff
@@ -1538,29 +1535,45 @@ pub fn snapshot_bank(
...
+        &snapshot_storages,
+        snapshot_version,
+    )?;
     add_snapshot_time.stop();
     inc_new_counter_info!("add-snapshot-ms", add_snapshot_time.as_ms() as usize);
 
     // Package the relevant snapshots
     let highest_bank_snapshot_info = get_highest_bank_snapshot_info(&bank_snapshots_dir)
         .expect("no snapshots found in config bank_snapshots_dir");
+    assert_eq!(highest_bank_snapshot_info.slot, root_bank.slot());
```
brooksprumo, line 1564: Added this assert, since it was just assumed before.

brooksprumo (commented):
There's a lot going on in this PR. I've tried to call out most important bits!

core/src/snapshot_packager_service.rs
carllin: hehe yeah its a puzzle whether the web of channels we have implemented will correctly shut down a node in cases where a specific thread panics. Generally, threads that are connected via channels will detect the other side has closed and also close, but that is not the case here (no channels to any other threads). 

I think you might have to set `exit` to true here and log an error. It's the global validator exit flag, so should shut down the full validator (maybe a good test is to spin up a validator with this exit set to true after 30 seconds)

runtime/src/accounts_background_service.rs
```diff
@@ -169,23 +174,60 @@ impl SnapshotRequestHandler {
...
+                        &self.snapshot_config.bank_snapshots_dir,
+                        &self.snapshot_config.snapshot_archives_dir,
+                        self.snapshot_config.snapshot_version,
+                        self.snapshot_config.archive_format,
+                        hash_for_testing,
+                        snapshot_type,
+                    );
+
+                    if result.is_ok() {
+                        break;
```
carllin, line 212: `snapshot_bank` can throw an error on channel send as well, if the thread on the receiving side of the `accounts_package_sender`  closes first. This is possible behavior when the validator is shutting down, so some options:

1. maybe we need check if result `is_ok` OR it's an expected send error, then break.

2. Eat and don't the send error at all: https://github.com/solana-labs/solana/blob/6d246031d06b81ee21d591a267a50f9f76d93eac/runtime/src/snapshot_utils.rs#L1579 with something like `let _ = accounts_package_sender.send(accounts_package)`, 

3. Have the `snapshot_bank` handle some input `max_retries` and wrap that logic in there.

I think 1 is cleanest, but is annoying to maintain because if we add additional errors in `snapshot_bank` we could run into this again. Maybe 1) AND we define some `enum SnapshotBankErrors` returned by `snapshot_bank` such that if we introduce any new errors, we'll have to handle the new member in the enum

runtime/src/accounts_background_service.rs
carllin: Yeah I think you would have to return the `status_cache_slot_deltas` in the error result of `snapshot_bank`. Maybe a `SnapshotBankArgs` that is the `snapshot_bank` arguments and is also that returned on error?

runtime/src/snapshot_package.rs
```diff
@@ -228,18 +180,10 @@ impl SnapshotArchiveInfoGetter for SnapshotPackage {
     }
 }
 
+/// Snapshots come in two flavors, Full and Incremental.  The IncrementalSnapshot has a Slot field,
+/// which is the incremental snapshot base slot.
```
carllin, line 184: nit: base -> base full snapshot

runtime/src/snapshot_utils.rs
carllin: sanity check here for my understanding 😃. Who's populating the `bank_snapshots_dir` here: https://github.com/solana-labs/solana/blob/6d246031d06b81ee21d591a267a50f9f76d93eac/runtime/src/snapshot_utils.rs#L1562

It can't have gotten a new root while AccountsBackgroundService was calliing `snapshot_bank()`?

runtime/src/accounts_background_service.rs
carllin: Yeah using the `block_height` is a lot simpler than trying to find the first slot that crossed the full snapshot boundary. Interesting bit here is the full snapshots will be correlated with skip rate, i.e. the higher the skip rate, the less often - in wallclock time - we will take full snapshots. I think this is fine for v1.

runtime/src/accounts_background_service.rs
carllin: You could track the `last_full_snapshot_slot` every time this condition is met: https://github.com/solana-labs/solana/blob/6d246031d06b81ee21d591a267a50f9f76d93eac/runtime/src/accounts_background_service.rs#L178-L182, and just pass it in here.

On startup you infer this `last_full_snapshot_slot` from the snapshot you unpack

runtime/src/accounts_background_service.rs
brooksprumo: Ah! That is interesting. Good to note.

runtime/src/accounts_background_service.rs
brooksprumo: Thanks, I'll give that a whirl!

runtime/src/accounts_background_service.rs
brooksprumo: Good idea, will do.

core/src/snapshot_packager_service.rs
brooksprumo: Good idea, will do.

runtime/src/snapshot_utils.rs
brooksprumo: The SnapshotConfig is what ultimately sets the `bank_snapshots_dir`. The bank snapshots created by `snapshot_utils::add_bank_snapshot()` is what ends up in this directory. I've added this comment, which hopefully describes the issue sufficiently:

```
    // The result of get_highest_bank_snapshot_info() must be the same as the bank snapshot created
    // by add_bank_snapshot() above.  Otherwise, the AccountsPackage would hardlink the wrong bank
    // snapshot relative to the root_bank's slot and snapshot storages.
```

Does that look OK?

runtime/src/accounts_background_service.rs
brooksprumo: Done!

core/src/snapshot_packager_service.rs
brooksprumo: OK! I confirmed that the asserts will bring down the whole node. I ran bootstrap-validator locally and inserted `panic!()`s into both AccountsBackgroundService and SnapshotPackagerService separately. The panics fired and immediately killed the process (then the script restarted the validator, as expected, successfully). So looks like this is all good.

runtime/src/snapshot_package.rs
brooksprumo: Is your suggestion to rename "incremental snapshot base slot" or "incremental snapshot base full snapshot slot"?

If yes, then I do not think we want to include "full snapshot". There are two contexts where incremental snapshots are talked about. One is when an incremental snapshot is on its own (i.e. when taking an incremental snapshot, or when getting snapshot storages for an incremental snapshot), and Two is when an incremental snapshot is talked about in reference to a full snapshot (i.e. when loading a bank from both a full and incremental snapshot, or `bank_to_incremental_snapshot_archive()`).

But first, we have `IncrementalSnapshotArchiveInfo` and `FullSnapshotArchiveInfo`, which are passed around and set by various functions. The `IncrementalSnapshotArchiveInfo` has an additional field called `base_slot`, so it's like the incremental snapshot's base slot. When I named this, I felt this communicated the idea of "incremental"; in my head this asks "what slot is this incremental snapshot based on?". The answer is the last full snapshot slot, yes. The comment on the field hopefully is clear:

https://github.com/solana-labs/solana/blob/6a53ec28e03f0e6368e3719daa50166bc21e979d/runtime/src/snapshot_archive_info.rs#L87-L97

So when constructing or using an `IncrementalSnapshotArchiveInfo`, you'll see params named `incremental_snapshot_base_slot`. (like building the incremental snapshot archive path, or getting snapshot storages for an incremental snapshot.)

And when using incremental snapshots with the context of a full snapshot also, you'll see params called `full_snapshot_slot`. Like when calling `bank_to_incremental_snapshot_archive`, or `bank_from_snapshot_archives`.

Hopefully this makes sense! Let me know what you think.

core/src/accounts_hash_verifier.rs
```diff
@@ -155,19 +179,46 @@ impl AccountsHashVerifier {
...
+        // a pending Full Snapshot.
+        let can_submit = match snapshot_package.snapshot_type {
+            SnapshotType::FullSnapshot => true,
+            SnapshotType::IncrementalSnapshot(_) => pending_snapshot_package
+                .lock()
+                .unwrap()
+                .as_ref()
+                .map_or(true, |snapshot_package| {
+                    matches!(
+                        snapshot_package.snapshot_type,
```
lijunwangs, line 211: a simple snapshot_package.snapshot_type == SnapshotType::IncrementalSnapshot does not work?

runtime/src/accounts_background_service.rs
```diff
@@ -169,23 +174,60 @@ impl SnapshotRequestHandler {
...
+
+                    // Only require success for full snapshots.  Otherwise just log the error and
+                    // continue on.
+                    if snapshot_type != Some(SnapshotType::FullSnapshot) {
+                        break;
+                    }
+
+                    assert!(
+                        attempts <= SNAPSHOT_BANK_MAX_ATTEMPTS,
+                        "Unable to snapshot bank intended for full snapshot!"
```
lijunwangs, line 231: Do we want to panic? Earlier we only logged an error

runtime/src/accounts_background_service.rs
```diff
@@ -399,7 +441,18 @@ impl AccountsBackgroundService {
                                 // slots >= bank.slot()
                                 bank.force_flush_accounts_cache();
                             }
-                            bank.clean_accounts(true, false, None);
+                            let last_full_snapshot_slot = request_handler
+                                .snapshot_request_handler
+                                .as_ref()
+                                .map(|snapshot_request_handler| {
+                                    snapshot_utils::calculate_last_full_snapshot_slot(
```
lijunwangs: Where is this defined?

core/src/accounts_hash_verifier.rs
```diff
@@ -96,49 +96,73 @@ impl AccountsHashVerifier {
         pending_snapshot_package: Option<&PendingSnapshotPackage>,
         hashes: &mut Vec<(Slot, Hash)>,
         exit: &Arc<AtomicBool>,
-        fault_injection_rate_slots: u64,
+        fault_injection_rate_slots: Slot,
```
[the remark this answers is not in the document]
brooksprumo: Maybe I misunderstood? I thought @carllin said `Slot` was ok in the other PR:

https://github.com/solana-labs/solana/pull/19271#discussion_r692570703

runtime/src/snapshot_package.rs
```diff
@@ -42,34 +41,61 @@ pub struct AccountsPackage {
...
-    pub storages: SnapshotStorages,
+    pub snapshot_storages: SnapshotStorages,
     pub hash: Hash, // temporarily here while we still have to calculate hash before serializing bank
     pub archive_format: ArchiveFormat,
     pub snapshot_version: SnapshotVersion,
     pub snapshot_archives_dir: PathBuf,
     pub expected_capitalization: u64,
     pub hash_for_testing: Option<Hash>,
     pub cluster_type: ClusterType,
+    pub snapshot_type: Option<SnapshotType>,
```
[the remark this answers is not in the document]
brooksprumo: This comment should hopefully give an answer: 
https://github.com/solana-labs/solana/pull/19401#discussion_r695117482

And some more context here too:
https://github.com/solana-labs/solana/pull/19401#discussion_r695105658

core/src/accounts_hash_verifier.rs
brooksprumo: Yah, since `SnapshotType::IncrementalSnapshot` is a tuple struct with a base slot field, I got a compile error for not accounting for the field. The `matches!` macro is basically just a `switch` under the hood.

runtime/src/accounts_background_service.rs
brooksprumo: Correct! There is new behavior; I put it in the description, but maybe I need it in the code too?

>!!! New behavior changes !!!
>
>Full snapshots (both bank and archive) **MUST** succeed.  If there are
>failures, the functions will be retried.  If they still fail, an assert
>will be triggered.
>
>This is required because of how the last full snapshot slot is
>calculated, which is used by AccountsBackgroundService when calling
>`clean_accounts()`.

runtime/src/accounts_background_service.rs
brooksprumo: Which in particular? If you're asking about `calculate_last_full_snapshot_slot()`, it's in `snapshot_utils.rs` here:
https://github.com/solana-labs/solana/blob/5985f4f0b9fa17cd00c078194ad4ed89c7465f31/runtime/src/snapshot_utils.rs#L1739-L1746

But I'm no longer using it in this implementation.

runtime/src/snapshot_package.rs
```diff
@@ -42,34 +41,61 @@ pub struct AccountsPackage {
...
+            snapshot_storages.iter().all(|storage| storage
+                .iter()
+                .all(|entry| entry.slot() > incremental_snapshot_base_slot)),
+            "Incremental snapshot package must only contain storage entries where slot > incremental snapshot base slot (i.e. full snapshot slot)!"
+            );
+        }
+
         // Hard link the snapshot into a tmpdir, to ensure its not removed prior to packaging.
+        let snapshot_links = tempfile::Builder::new()
+            .prefix(&format!("{}{}-", TMP_BANK_SNAPSHOT_PREFIX, bank.slot()))
```
lijunwangs, line 94: Do full snapshot and incremental snapshot look any different now for the tmp dir entry?

runtime/src/snapshot_package.rs
lijunwangs: Nice!

core/src/accounts_hash_verifier.rs
lijunwangs: Got it -- saw you have changed the enum def -- maybe we should define a utility function check is_incremental?

runtime/src/snapshot_package.rs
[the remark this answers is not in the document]
lijunwangs: hmm. We are overloading the snapshot type for some other purpose -- maybe a separate flag for the Hash only. And when it is None, what is really the type? Is it a full or incremental? Does not matter?

core/src/accounts_hash_verifier.rs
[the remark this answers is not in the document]
lijunwangs: What I means is the count of something and the something can be of different unit. For example 'count of car' and car is of different types. What we really want here is the count or number, not the unit of the thing we are counting. Why it worked by chance is -- the Slot by design itself is a number:)

runtime/src/accounts_background_service.rs
```diff
@@ -169,23 +171,64 @@ impl SnapshotRequestHandler {
...
-                    self.snapshot_config.snapshot_version,
-                    &self.snapshot_config.archive_format,
-                    hash_for_testing,
-                );
-                if r.is_err() {
+                loop {
+                    attempts += 1;
+                    let result = snapshot_utils::snapshot_bank(
+                        &snapshot_root_bank,
+                        status_cache_slot_deltas.clone(),
```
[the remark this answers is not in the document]
brooksprumo: Yep! I'm looking to avoid that. See this thread for previous discussion:
https://github.com/solana-labs/solana/pull/19401#discussion_r695107247

But now I am going to do a different implementation _without_ the retry loop, so I'll be able to remove the `.clone()` here.

runtime/src/snapshot_package.rs
brooksprumo: Nope. None of the different temp directories need different prefixes actually, but now we're only differentiating between bank snapshots and snapshot archives.

runtime/src/accounts_background_service.rs
lijunwangs: Got it -- I did not realize it was hidden in the file not expanded. Sorry for the confusion

runtime/src/snapshot_package.rs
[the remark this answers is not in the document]
brooksprumo: Yeah, I debated how far to take this... There's a lot of coupling and implied behavior between bank snapshots, snapshot archives, and the hashing... plus how those three interact with the background services... Ultimately I figured that to do it "right" would likely require redoing the whole background services from the ground up, and that would probably be out of scope for this task, so I tried to make it _least bad_ :)

This `snapshot_type` is used to decide which snapshot storages to get, when to take a snapshot/what time, and if it should be archived by SnapshotPackagerService.

runtime/src/accounts_background_service.rs
brooksprumo: Now we aren't doing retries in ABS, so no more clone. Yay!

runtime/src/accounts_background_service.rs
brooksprumo: Carl and I talked offline. Retrying in ABS has been removed. Filesystem errors shouldn't happen, so any fs:: calls now are `expect()` instead of `?`.

core/src/accounts_hash_verifier.rs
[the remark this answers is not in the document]
brooksprumo: Ok, reverted back to u64!

core/src/accounts_hash_verifier.rs
brooksprumo: I like it. Done!

runtime/src/accounts_background_service.rs
brooksprumo: Heh, and now this has changed again. There's no more retrying in ABS, so this is all gone now.

runtime/src/accounts_background_service.rs
brooksprumo: No worries! Thanks for being detailed.

runtime/src/accounts_background_service.rs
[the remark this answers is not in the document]
brooksprumo: And now retry is gone, so `.clone()` is gone. heh.

brooksprumo:
@carllin @lijunwangs Thanks for the reviews so far! I've fixed or commented on everything, so this PR should be ready for another round. Thanks in advance!

runtime/src/snapshot_package.rs
carllin: makes sense, thanks for the explanation!

runtime/src/snapshot_utils.rs
carllin: hmmm I see, does it make sense to just use the result of the `add_bank_snapshot` then instead of throwing away that result and refetching it via `get_highest_bank_snapshot_info()`

runtime/src/accounts_background_service.rs
```diff
@@ -169,26 +170,52 @@ impl SnapshotRequestHandler {
...
+                // if there was an error while snapshotting the bank, it was due to the
+                // channel, and thus the node is coming down.  Can log it here and continue on.
+                if result.is_err() {
                     warn!(
-                        "Error generating snapshot for bank: {}, err: {:?}",
+                        "Error taking bank snapshot. slot: {}, snapshot type: {:?}, err: {:?}",
                         snapshot_root_bank.slot(),
-                        r
+                        snapshot_type,
+                        result
```
carllin, line 216: Should we also just return here from AccountsBackgroundService and shut the node down?

 Otherwise reasoning what happens if this fails for a full snapshot and then somehow works for the next incremental snapshot is kind of annoying. We would have an incremental snapshot with a base slot that we don't have a full snapshot for xD

runtime/src/snapshot_utils.rs
```diff
@@ -1538,30 +1535,52 @@ pub fn snapshot_bank(
...
+    });
     let mut add_snapshot_time = Measure::start("add-snapshot-ms");
-    add_bank_snapshot(&bank_snapshots_dir, root_bank, &storages, snapshot_version)?;
+    add_bank_snapshot(
+        &bank_snapshots_dir,
+        root_bank,
+        &snapshot_storages,
+        snapshot_version,
+    )
+    .expect("failed to serialize bank snapshot to disk");
```
carllin, line 1558: should we add some safeguards to make sure `add_bank_snapshot` *only* returns nonretryable errors? It seems like `add_bank_snapshot` returns type `SnapshotError`. Right now it seems all of them are fatal, but should we introduce a `is_fatal()` function on `SnapshotError` in case somebody adds a new variant to it? Then we can say 

```
if let Err(e) = add_bank_snapshot {
       if e.is_fatal() {
           panic!("failed to serialize bank snapshot to disk");
       }
}
```

runtime/src/snapshot_utils.rs
brooksprumo: Yep, love it! I wasn't sure if this behavior as written was originally intended as a check for the filesystem or something, but with the checks now, that's not an issue. Done!

runtime/src/snapshot_utils.rs
```diff
@@ -1538,30 +1535,43 @@ pub fn snapshot_bank(
...
-        storages,
-        *archive_format,
+        snapshot_storages,
+        archive_format,
         snapshot_version,
         hash_for_testing,
-    )?;
+        snapshot_type,
+    )
+    .expect("failed to hard link bank snapshot into a tmpdir");
```
carllin, line 1574: Hmmm circling back to this `expect`, maybe we should return an error here and have ABS handle it so we're sure it exits properly? This ensures that ABS can set the exit variable so the validator shuts down when ABS shuts down.

carllin (approved):

runtime/src/snapshot_utils.rs
brooksprumo: Handled offline.

runtime/src/snapshot_utils.rs
brooksprumo: Handled offline.

runtime/src/accounts_background_service.rs
brooksprumo: Handled offline.

PR #19402, opened 2021-08-24
Non clean 1.6 jwash2

jeffwashington:
#### Summary of Changes

Fixes #


runtime/src/accounts_db.rs
```diff
@@ -2619,6 +2623,16 @@ impl AccountsDb {
         }
         remove_storages_elapsed.stop();
 
+        for s in &all_removed_slot_storages {
+            for (_store_id, store) in s.read().unwrap().iter() {
+                if !store.unref_done.swap(true, Ordering::Relaxed) {
```
jeffwashington, line 2628: @sakridge here's my first attempt. All tests pass with the exception of a race condition I'm debugging in:
test_store_scan_consistency_unrooted

sakridge: I see. So if we may go through the unref path twice in some cases. This would prevent that by keeping a bool per store to prevent it.

runtime/src/accounts_db.rs
```diff
@@ -4032,6 +4050,9 @@ impl AccountsDb {
         for slot in dead_slots.iter() {
             if let Some(slot_storage) = self.storage.get_slot_stores(*slot) {
                 for store in slot_storage.read().unwrap().values() {
+                    let already_unrefd = store.unref_done.swap(true, Ordering::Relaxed);
```
jeffwashington, line 4053: @sakridge here's the other path I found.

runtime/src/accounts_db.rs
jeffwashington: All runtime tests now pass locally. I wasn't handling `AccountStorageEntry::recycle`.

jeffwashington:
#19406 #19407 
running ci for 1.7 and master

jeffwashington:
we believe @carllin has a 1.7 solution for this that he will back port to 1.6. I have separated the test out for consideration separately in 1.7 and master.

carllin:
Here is the proposed backport: https://github.com/solana-labs/solana/pull/19414. It was a bit of a messy merge as there are quite a few fixes on 1.7 that had not been backported to 1.6

PR #19408, opened 2021-08-24
Non clean 1.7 test only

jeffwashington:
#### Summary of Changes

Fixes #


jeffwashington:
@sakridge @carllin @lijunwangs here is just the new test stephen added for 1.6. I added the test to 1.7. This test passes. This confirms that the problem identified in 1.6 has already been fixed in 1.7. Carl will work on backporting his fix to 1.6 instead of the fix that stephen and I worked on.

jeffwashington:
here is the test against master:
#19409

PR #19414, opened 2021-08-25
Backport #17319

carllin:
#### Problem
`purge_slots()` removes the storage entries without unrefing the containing accounts. This may lead to a bad ref count for zero lamport accounts which will then never get removed.

#### Summary of Changes
Backport #17319 which reworked `purge_slots()` to use a common `handle_reclaims()` path which takes care of the dead slot and unrefing accounts

Fixes #


carllin:
@ryoqun as you were the primary reviewer on https://github.com/solana-labs/solana/pull/17319, if you could sanity check this that would be awesome!

PR #19418, opened 2021-08-25
added bank_cost_trackers module to support each bank to have its own cost tracker

taozhu-chicago:
#### Problem
`cost tracker` is shared between banking stage threads, there is potential race condition where one thread may reset `cost tracker` to new bank therefore resets tracked costs to zero while other threads are still working on previous bank.

#### Summary of Changes
- add `bank_cost_trackers.rs` to track `cost_tracker` per slot;
- release `cost_tracker` when its bank is no longer the current working bank, reducing memory footprint;
- banking_stage threads using `BankCostTrackers` to access bank's unique `cost-tracker`, avoiding race condition

Fixes #

### DIFF
--- a/banking-bench/src/main.rs
+++ b/banking-bench/src/main.rs
@@ -4,7 +4,9 @@ use crossbeam_channel::unbounded;
 use log::*;
 use rand::{thread_rng, Rng};
 use rayon::prelude::*;
-use solana_core::{banking_stage::BankingStage, cost_model::CostModel, cost_tracker::CostTracker};
+use solana_core::{
+    bank_cost_trackers::BankCostTrackers, banking_stage::BankingStage, cost_model::CostModel,
+};
 use solana_gossip::{cluster_info::ClusterInfo, cluster_info::Node};
 use solana_ledger::{
     blockstore::Blockstore,
@@ -230,9 +232,9 @@ fn main() {
             vote_receiver,
             None,
             replay_vote_sender,
-            Arc::new(RwLock::new(CostTracker::new(Arc::new(RwLock::new(
+            Arc::new(BankCostTrackers::new(Arc::new(RwLock::new(
                 CostModel::default(),
-            ))))),
+            )))),
         );
         poh_recorder.lock().unwrap().set_bank(&bank);
 
--- a/core/benches/banking_stage.rs
+++ b/core/benches/banking_stage.rs
@@ -7,9 +7,9 @@ use crossbeam_channel::unbounded;
 use log::*;
 use rand::{thread_rng, Rng};
 use rayon::prelude::*;
+use solana_core::bank_cost_trackers::BankCostTrackers;
 use solana_core::banking_stage::{BankingStage, BankingStageStats};
 use solana_core::cost_model::CostModel;
-use solana_core::cost_tracker::CostTracker;
 use solana_entry::entry::{next_hash, Entry};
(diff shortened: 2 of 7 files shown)


jstarry:
Why not add a `cost_tracker` field to the `Bank` struct? It would eliminate the race condition issue and you wouldn't have to handle the possibility of a bank not having an associated cost tracker, which doesn't make sense anyways.

Also, this is not related to this PR but it feels really strange to me that we have locking in `CostModel` just that multiple threads can reuse the `transaction_cost` field. Seems like it would affect performance negatively to grab a lock for each processed transaction.

taozhu-chicago:
> Why not add a `cost_tracker` field to the `Bank` struct? It would eliminate the race condition issue and you wouldn't have to handle the possibility of a bank not having an associated cost tracker, which doesn't make sense anyways.

This is a good point, it'd make sense to let `bank` has a `cost tracker`. It's just I am still nervous about modifying `bank`, which comes with backward compatibility issues. My thought is there are still a lot of changes needed for cost_tracking when it'd be put out to testnet in 1.8, might be a good idea to do all that incremental changes (like this one) outside `bank`. Can integrate the function into bank when it matures somewhat. 

> Also, this is not related to this PR but it feels really strange to me that we have locking in `CostModel` just that multiple threads can reuse the `transaction_cost` field. Seems like it would affect performance negatively to grab a lock for each processed transaction.

Yeah, this is up for refactoring too (right next on my todo list). The reason the `transaction_cost` being reused was stats showed heap allocating `transaction_cost` object for each `calculate_cost()` call was expensive. I don't remember the exact micros, but it was too expensive to be ignored. I did a patch to reuse the object for each calculation, which seems to put perf-test back on track. The cost of that is the awkwardness of needing write lock for apparent read operation. Going to refactor it out soon.

jstarry:
> This is a good point, it'd make sense to let bank has a cost tracker. It's just I am still nervous about modifying bank, which comes with backward compatibility issues. My thought is there are still a lot of changes needed for cost_tracking when it'd be put out to testnet in 1.8, might be a good idea to do all that incremental changes (like this one) outside bank. Can integrate the function into bank when it matures somewhat.

Now is definitely the time because as you've already discovered, separating the cost model from the bank is prone to race conditions. It'd be much safer to associate the cost tracker directly with the bank.

> Yeah, this is up for refactoring too (right next on my todo list). The reason the transaction_cost being reused was stats showed heap allocating transaction_cost object for each calculate_cost() call was expensive. I don't remember the exact micros, but it was too expensive to be ignored. I did a patch to reuse the object for each calculation, which seems to put perf-test back on track. The cost of that is the awkwardness of needing write lock for apparent read operation. Going to refactor it out soon.

Got it, fixing this later is fine!

taozhu-chicago:
> > This is a good point, it'd make sense to let bank has a cost tracker. It's just I am still nervous about modifying bank, which comes with backward compatibility issues. My thought is there are still a lot of changes needed for cost_tracking when it'd be put out to testnet in 1.8, might be a good idea to do all that incremental changes (like this one) outside bank. Can integrate the function into bank when it matures somewhat.
> 
> Now is definitely the time because as you've already discovered, separating the cost model from the bank is prone to race conditions. It'd be much safer to associate the cost tracker directly with the bank.
> 

Yeah, I see your point, it might as well be the best time. Let's leave this PR as-is while I making the change

PR #19420, opened 2021-08-25
Experiment with removing zero lamport accounts from non-empty storages

carllin:
#### Problem

As a refresher, we add keys to the candidate set on:
1. Dead slots, we add the storage entry to the candidate set
2. Shrinking slots, we add the storage entry to the candidate set
3. New roots, we add the storage entry to the candidate set
4.  Bank freeze we add the dirty pubkeys to the candidate set

So a zero lamport key will never be cleaned if there ever is ever a scenario where:
1) We add a zero lamport key to the candidate set
2) The zero lamport key is not removed via clean
3) We then never add the zero lamport key back to the candidate set

I think this can happen if:
1) We add the zero lamport key to the candidate set via one of the four options above
2) We don't clean the zero lamport key, *even though the ref count is 1*, due to this check: https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_db.rs#L1594-L1597
3) The slot for the zero lamport key is never shrunk again because it never goes under 80% capacity (Maybe the zero lamport account data is large or there are a lot of zero lamport keys in this slot that can't be cleaned for the same reason as 2) above, *even if rent overwrites all the remaining non-zero pubkeys*.

#### Summary of Changes
Remove zero lamport keys if ref_count is 1, even if the storage entry is not empty

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -1565,21 +1565,8 @@ impl AccountsDb {
                 );
                 true
             } else {
-                let mut no_delete = false;
-                for (_slot, account_info) in account_infos {
-                    debug!(
-                        "calc_delete_dependencies()
-                        storage id: {},
-                        count len: {}",
-                        account_info.store_id,
-                        store_counts.get(&account_info.store_id).unwrap().0,
-                    );
-                    if store_counts.get(&account_info.store_id).unwrap().0 != 0 {
-                        no_delete = true;
-                        break;
-                    }
-                }
-                no_delete
+                assert_eq!(account_infos.len(), 1);
+                false
             };
             if no_delete {
                 let mut pending_store_ids: HashSet<usize> = HashSet::new();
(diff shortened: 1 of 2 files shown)


runtime/src/accounts_db.rs
```diff
@@ -1565,21 +1565,8 @@ impl AccountsDb {
...
-                        account_info.store_id,
-                        store_counts.get(&account_info.store_id).unwrap().0,
-                    );
-                    if store_counts.get(&account_info.store_id).unwrap().0 != 0 {
-                        no_delete = true;
-                        break;
-                    }
-                }
-                no_delete
+                assert_eq!(account_infos.len(), 1);
```
lijunwangs, line 1568: Why account_infos.len(). would be = 1? 
I thought the issue could be the pub key for some 0 lamports are not returned at all in construct_candidate_clean_keys and hence we will not catch them either.

sakridge:
> We don't clean the zero lamport key, even though the ref count is 1, due to this check: https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_db.rs#L1594-L1597

Remember that we already decremented the ref counts here. So if it was 1 originally, it should be 0 here.

carllin:
Closed in favor of https://github.com/solana-labs/solana/pull/19570 which will merge to master, as 1.7 has no urgency to be fixed currently

PR #19426, opened 2021-08-25
cache account hash info

jeffwashington:
#### Problem
As # accounts grows, we cannot afford to scan all storages to calculate the hash. If we do, we use up all free ram paging in storages. This results in poor performance for the entire system as likely every storage/disk access becomes a page fault. Kin simulations indicated this problem causes system performance to grind to a halt once we exceed free ram. Lobotomizing hash calculation completely allowed kin to go much further - limited by the accounts index itself.
Analysis indicates that from 1 hash calculation to the next (minimum of of 30-40s) almost all storage changes take place at the top and bottom ends of the 432k slot range. So, the middle slot storages largely remain constant. With shrink throttling, this is likely even more true.
See the summary of changes below.
Note that there are further optimizations for this solution as well. For example, we could choose larger ranges in the middle of the slot range and smaller ones close to the ends. This would prevent misses to cause us to scan 5,000 storages and reduce that down to 10s, 100s, or even 1,000. We could also sort (and reduce?) across multiple ranges and store indexes into these files such that we retain even more of the required work from one hash calculation to the next at a lower cost of data (some indices instead of (pubkey, hash, lamports))

#### Summary of Changes
Break storage scanning into ranges of slots (slot mod 5_000, for example). Scan that range of slots/storages. Store the reduced and sorted results for that range of slots in a separate file. Next time, before scanning all storages in that range, check mod dates (is this sufficient?) on all storage files and check for existence of a cached results file generated from the same storages. If it exists, skip scanning those storages and sorting the results and just used the previously saved data.
Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -29,6 +29,7 @@ use crate::{
     },
     ancestors::Ancestors,
     append_vec::{AppendVec, StoredAccountMeta, StoredMeta, StoredMetaWriteVersion},
+    cache_hash_data::CacheHashData,
     contains::Contains,
     pubkey_bins::PubkeyBinCalculator16,
     read_only_accounts_cache::ReadOnlyAccountsCache,
@@ -61,6 +62,7 @@ use std::{
     boxed::Box,
     collections::{hash_map::Entry, BTreeSet, HashMap, HashSet},
     convert::TryFrom,
+    hash::{Hash as StdHash, Hasher as StdHasher},
     io::{Error as IoError, Result as IoResult},
     ops::{Range, RangeBounds},
     path::{Path, PathBuf},
@@ -101,7 +103,7 @@ pub const BINS_PER_PASS: usize = PUBKEY_BINS_FOR_CALCULATING_HASHES / NUM_SCAN_P
 // If this is too big, we don't get enough parallelism of scanning storages.
 // If this is too small, then we produce too many output vectors to iterate.
 // Metrics indicate a sweet spot in the 2.5k-5k range for mnb.
-const MAX_ITEMS_PER_CHUNK: Slot = 5_000;
+const MAX_ITEMS_PER_CHUNK: Slot = 2_500;
 
 // A specially reserved storage id just for entries in the cache, so that
 // operations that take a storage entry can maintain a common interface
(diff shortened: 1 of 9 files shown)


jeffwashington:
I did analysis on chunk/cache file size vs perf. Larger chunks/cache file sizes perform better. But, the larger the size, the more likely a change would cause the entire file to be invalid. > 1k is definitely an inflection point where the improvement starts to flatten out.

jeffwashington:
Kin sims indicate that we get much further with this approach. We are at this moment at 519M with this hash calculation occurring correctly and an in-mem, HashMap accounts index.

runtime/src/accounts_db.rs
```diff
@@ -956,6 +961,12 @@ pub struct AccountsDb {
     /// Set of storage paths to pick from
     pub(crate) paths: Vec<PathBuf>,
 
+    ledger_path: PathBuf,
```
sakridge, line 964: Can we call it `accounts_hash_cache_path` or something like that? It's not really holding ledger.

runtime/src/accounts_db.rs
```diff
@@ -4932,32 +4963,113 @@ impl AccountsDb {
...
+                if start == end {
+                    return after_func(retval);
+                }
+
+                let mut file_name = String::default();
+                if accounts_cache_and_ancestors.is_none()
+                    && end.saturating_sub(start) == MAX_ITEMS_PER_CHUNK
+                {
+                    let mut load_from_cache = true;
+                    let mut hasher = std::collections::hash_map::DefaultHasher::new(); // wrong one?
```
sakridge, line 5009: Just to give my 2 cents on the comments...
Obviously this is not cryptographic so you could encounter collisions, although it seems you would have to have also a collision with the same slot and bin ranges which seems very unlikely.

Blake3 hash is quite fast though. It might be worth just testing to see if it's the same speed because the amount of data being hashed here is quite small so the speed difference might also not be noticeable at all and has the advantage that one doesn't need to consider collisions at all.

runtime/src/accounts_db.rs
```diff
@@ -4932,32 +4963,113 @@ impl AccountsDb {
...
+                        bin_range.end.hash(&mut hasher);
+                        if let Some(sub_storages) = sub_storages {
+                            if sub_storages.len() > 1 {
+                                load_from_cache = false;
+                                break;
+                            }
+                            let storage_file = sub_storages.first().unwrap().accounts.get_path();
+                            slot.hash(&mut hasher);
+                            storage_file.hash(&mut hasher);
+                            // check alive_bytes, etc. here?
```
sakridge, line 5023: It does make me a bit nervous to rely just on the fs modify time, but I can't think of a scenario where it shouldn't work.

sakridge (approved):
looks good

PR #19457, opened 2021-08-27
Web3 docs updated with quickstart guide

cryptogosu:
#### Problem

The Solana Web3 documentation was lacking some good examples of how to get started.

#### Summary of Changes

Created a Quickstart guide for using Solana Web3. This includes how to use a Keypair, create and sign a transaction, and basic interaction with Programs on Solana.

Any feedback is welcome.

### DIFF
--- a/docs/src/developing/clients/javascript-api.md
+++ b/docs/src/developing/clients/javascript-api.md
@@ -2,4 +2,331 @@
 title: Web3 JavaScript API
 ---
 
-See [solana-web3](https://solana-labs.github.io/solana-web3.js/).
+## What is Solana-Web3.js?
+
+The Solana-Web3.js library aims to provide complete coverage of Solana. The library was built on top of the [Solana JSON RPC API](https://docs.solana.com/developing/clients/jsonrpc-api).
+
+## Common Terminology
+
+| Term | Definition |
+|-------------|------------------------|
+| Program     | Stateless executable code written to interpret instructions. Programs are capable of performing actions based on the instructions provided. |
+| Instruction | The smallest unit of a program that a client can include in a transaction. Within its processing code, an instruction may contain one or more cross-program invocations. |
+| Transaction | One or more instructions signed by the client using one or more Keypairs and executed atomically with only two possible outcomes: success or failure. |
+
+For the full list of terms, see [Solana terminology](https://docs.solana.com/terminology#cross-program-invocation)
+
+## Getting Started
+
+### Installation
+
+#### yarn
+
+```bash
+$ yarn add @solana/web3.js
+```
+
+#### npm
+
+```bash
+$ npm install --save @solana/web3.js
+```
+
+#### Bundle
+
+```html
+<!-- Development (un-minified) -->
+<script src="https://unpkg.com/@solana/web3.js@latest/lib/index.iife.js"></script>
+
+<!-- Production (minified) -->
+<script src="https://unpkg.com/@solana/web3.js@latest/lib/index.iife.min.js"></script>
+```
+
+### Usage
+
+#### Javascript
+
+```javascript
+const solanaWeb3 = require('@solana/web3.js');
+console.log(solanaWeb3);
+```
+
+
+#### ES6
+
+```javascript
+import * as solanaWeb3 from '@solana/web3.js';
+console.log(solanaWeb3);
+```
+
+
+#### Browser Bundle
+
+```javascript
+// solanaWeb3 is provided in the global namespace by the bundle script
+console.log(solanaWeb3);
+```
+
+## Quickstart
+
+### Connecting to a Wallet
+
+To allow users to use your dApp or application on Solana, they will need to get access to their Keypair. A Keypair is a private key with a matching public key, used to sign transactions.
+
+There are two ways to obtain a Keypair:
+1. Generate a new Keypair
+2. Obtain a Keypair using the secret key
+
+You can obtain a new Keypair with the following:
+
+```javascript
+const {Keypair} = require("@solana/web3.js");
+
+let keypair = Keypair.generate();
+```
+
+This will generate a brand new Keypair for a user to fund and use within your application.
+
+You can allow entry of the secretKey using a textbox, and obtain the Keypair with `Keypair.fromSecretKey(secretKey)`.
+
+```javascript
+const {Keypair} = require("@solana/web3.js");
+
+let secretKey = Uint8Array.from([
+  202, 171, 192, 129, 150, 189, 204, 241, 142,  71, 205,
+  2,  81,  97,   2, 176,  48,  81,  45,   1,  96, 138,
+  220, 132, 231, 131, 120,  77,  66,  40,  97, 172,  91,
+  245,  84, 221, 157, 190,   9, 145, 176, 130,  25,  43,
+  72, 107, 190, 229,  75,  88, 191, 136,   7, 167, 109,
+  91, 170, 164, 186,  15, 142,  36,  12,  23
+]);
+
+let keypair = Keypair.fromSecretKey(secretKey);
+```
+
+Many wallets today allow users to bring their Keypair using a variety of extensions or web wallets. The general recommendation is to use wallets, not Keypairs, to sign transactions. The wallet creates a layer of separation between the dApp and the Keypair, ensuring that the dApp never has access to the secret key. You can find ways to connect to external wallets with the [wallet-adapter](https://github.com/solana-labs/wallet-adapter) library.
+
+### Creating and Sending Transactions
+
+To interact with programs on Solana, you create, sign, and send transactions to the network. Transactions are collections of instructions with signatures. The order that instructions exist in a transaction determines the order they are executed.
+
+A transaction in Solana-Web3.js is created using the [`Transaction`](javascript-api.md#Transaction) object and adding desired messages, addresses, or instructions.
+
+Take the example of a transfer transaction:
+
+```javascript
+const {Keypair, Transaction, SystemProgram, LAMPORTS_PER_SOL} = require("@solana/web3.js");
+
+let fromKeypair = Keypair.generate();
+let toKeypair = Keypair.generate();
+let transaction = new Transaction();
+
+transaction.add(
+  SystemProgram.transfer({
+    fromPubkey: fromKeypair.publicKey,
+    toPubkey: toKeypair.publicKey,
+    lamports: LAMPORTS_PER_SOL
+  })
+);
+```
+
+The above code achieves creating a transaction ready to be signed and broadcasted to the network. The `SystemProgram.transfer` instruction was added to the transaction, containing the amount of lamports to send, and the `to` and `from` public keys.
+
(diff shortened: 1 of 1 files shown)


nimmneun:
@cryptogosu y not fix the 5 whitespaces that prevent CI from passing. People might skip the review just bcs of that 🙈 Appreciate the quick start guide 👍

cryptogosu:
@nimmneun I was not able to update the past few days, but the update should now be pushed with the fix.

docs/src/developing/clients/javascript-api.md
```diff
@@ -2,4 +2,308 @@
...
-See [solana-web3](https://solana-labs.github.io/solana-web3.js/).
+## What is Solana-Web3.js?
+
+The Solana-Web3.js library aims to provide complete coverage of Solana. The library was built on top of the [Solana JSON RPC API](https://docs.solana.com/developing/clients/jsonrpc-api).
+
+## Common Terminology
+
+| Term | Definition |
+|-------------|------------------------|
+| Program     | Stateless executeable code written to interpret instructions. Programs are capable of performing actions in based on the instructions provided. |
```
paul-schaaf, line 13: > Programs are capable of performing actions in based on the instructions provided

grammar^

docs/src/developing/clients/javascript-api.md
```diff
@@ -2,4 +2,308 @@
...
+```javascript
+// solanaWeb3 is provided in the global namespace by the bundle script
+console.log(solanaWeb3);
+```
+
+## Quickstart
+
+### Connecting to a Wallet
+
+To allow users to use your dApp or application on Solana, they will need to connect their wallet. A Keypair is a private key with a matching public key, used to sign transactions.
```
paul-schaaf, line 74: these two sentences feel disconnected to me. it's not obvious that the wallet is basically the keypair from this. could be connected by using Keypair instead of wallet in the first sentence or adding a clause saying sth like "...their wallet, which is just a keypair."

docs/src/developing/clients/javascript-api.md
```diff
@@ -2,4 +2,308 @@
...
+  transaction,
+  [keypair]
+);
+```
+
+The above code takes in a `TransactionInstruction` using `SystemProgram`, creates a `Transaction`, and sends it over the network. You use `Connection` in order to define with Solana network you are connecting to, namely `mainnet-beta`, `testnet`, or `devnet`.
+
+### Interacting with Programs
+
+The previous section visits sending basic transactions. Interacting with programs is a more complex transaction, and can be done similarly. At the time of writing programs on Solana are either written in Rust or C.
```
paul-schaaf, line 152: > Interacting with programs is a more complex transaction, and can be done similarly

the previous section was also about interacting with a program

docs/src/developing/clients/javascript-api.md
```diff
@@ -2,4 +2,308 @@
...
+  72, 107, 190, 229,  75,  88, 191, 136,   7, 167, 109,
+  91, 170, 164, 186,  15, 142,  36,  12,  23
+]);
+
+let keypair = Keypair.fromSecretKey(secretKey);
+```
+
+Many wallets today allow users to bring their Keypair using a variety of extensions or web wallets. The general recommendation is to use wallets, not text entry, to obtain the Keypair for signing. You can find ways to connect to external wallets with the [wallet-adapter](https://github.com/solana-labs/wallet-adapter) library.
+
+### Creating and Sending Transactions
```
paul-schaaf, line 109: personally, I would combine this and the next section since both of them are about interacting with programs and splitting them like this gives the impression that transfer transactions are _not_ interacting with programs.

paul-schaaf: Id definitely keep the SystemProgram.transfer example tho cause it shows that in practice many of the common calls do not to be built by hand but can be abstracted away with libraries. alternatively, you could use the SystemProgram.allocate builder function as an example too.

docs/src/developing/clients/javascript-api.md
```diff
@@ -2,4 +2,308 @@
...
+    pubkey: &Pubkey,
+    space: u64
+) -> Instruction
+```
+
+In Solana when you want to interact with a program you must first know all the accounts you will be interacting with.
+
+You must always provide every account that the program will be interacting within the instruction. Not only that, but you must provide whether or not the account is `isSigner` or `isWritable`.
+
+In the `allocate` method above, a single account `pubKey` is required, as well as an amount of `space` for allocation. We know that the `allocate` method writes to the account by allocating space within it, making the `pubKey` required to be `isWriteable`. `isSigner` is required when you are designating the account that is running the instruction. In this case, the signer is the account calling to allocate space within itself.
```
paul-schaaf, line 167: > isWriteable

isWritable

docs/src/developing/clients/javascript-api.md
```diff
@@ -2,4 +2,308 @@
...
+let layoutFields = Object.assign({instruction: allocateStruct.index}, params);
+allocateStruct.layout.encode(layoutFields, data);
+
+allocateTransaction.add(new web3.TransactionInstruction({
+  keys,
+  programId: web3.SystemProgram.programId,
+  data,
+}));
+
+web3.sendAndConfirmTransaction(connection, allocateTransaction, [keypair]);
```
paul-schaaf, line 308: consider adding a second keypair here that pays for the tx. otherwise it would mean that the account getting allocated would also be the feepayer which is not a realistic example. this would also show that it's possible to pay for a tx with an account that is not part of any of the instructions

docs/src/developing/clients/javascript-api.md
```diff
@@ -2,4 +2,308 @@
...
+let allocateStruct = {
+  index: 8,
+  layout: struct([
+    u32('instruction'),
+    ns64('space'),
+  ])
+};
+```
+
+The above is created using `@solana/buffer-layout` to facilitate the payload creation. The `allocate` function takes in the parameter `space`. To interact with the function we must provide the data as a Buffer format. The `buffer-layout` library helps with allocating the buffer and encoding it correctly for Rust programs on Solana to interpret.
```
joncinque, line 196: nit: can clarify to say: "The above is created using `u32` and `ns64` from `@solana/buffer-layout`..."

docs/src/developing/clients/javascript-api.md
```diff
@@ -2,4 +2,308 @@
...
+  programId: web3.SystemProgram.programId,
+  data,
+}));
+
+web3.sendAndConfirmTransaction(connection, allocateTransaction, [keypair]);
+```
+
+Finally, we add the transaction instruction with all the account keys, data, and programId and broadcast the transaction to the network.
+
+The full code can be found below. **Note**: You may need to fund the `Keypair` to get it to run on your local network.
```
joncinque, line 276: Nit: can clarify here to say "You may need to fund the `Keypair` with an airdrop or transfer to get it to run on your local network"

joncinque (commented):
Looks great to me! Just a few small comments

docs/src/developing/clients/javascript-api.md
cryptogosu: Updated to `Stateless executeable code written to interpret instructions. Programs are capable of performing actions based on the instructions provided.`

docs/src/developing/clients/javascript-api.md
cryptogosu: Added some flavor to explain it was also writing about interacting with a program

docs/src/developing/clients/javascript-api.md
cryptogosu: Updated to remove wallet - Keypair interchangeable language. This was based on advice of someone from Solana

docs/src/developing/clients/javascript-api.md
cryptogosu: I didn't combine them, but added some clarification. It may be even better to show both transfer or allocate in both forms too.

docs/src/developing/clients/javascript-api.md
cryptogosu: Added payer

docs/src/developing/clients/javascript-api.md
```diff
@@ -2,4 +2,331 @@
...
+To interact with programs on Solana, you create, sign, and send transactions to the network. Transactions are collections of instructions with signatures. The order that instructions exist in a transaction determines the order they are executed.
+
+A transaction in Solana-Web3.js is created using the [`Transaction`](javascript-api.md#Transaction) object and adding desired messages, addresses, or instructions.
+
+Take the example of a transfer transaction:
+
+```javascript
+const {Keypair, Transaction, SystemProgram, LAMPORTS_PER_SOL} = require("@solana/web3.js");
+
+let keypair = Keypair.generate();
```
joncinque, line 120: This can get moved down to the `sendAndConfirm` part, since it isn't used here

docs/src/developing/clients/javascript-api.md
```diff
@@ -2,4 +2,331 @@
...
+    pubkey: &Pubkey,
+    space: u64
+) -> Instruction
+```
+
+In Solana when you want to interact with a program you must first know all the accounts you will be interacting with.
+
+You must always provide every account that the program will be interacting within the instruction. Not only that, but you must provide whether or not the account is `isSigner` or `isWritable`.
+
+In the `allocate` method above, a single account `pubKey` is required, as well as an amount of `space` for allocation. We know that the `allocate` method writes to the account by allocating space within it, making the `pubKey` required to be `isWritable`. `isSigner` is required when you are designating the account that is running the instruction. In this case, the signer is the account calling to allocate space within itself.
```
joncinque, line 169: nit
```suggestion
In the `allocate` method above, a single account `pubkey` is required, as well as an amount of `space` for allocation. We know that the `allocate` method writes to the account by allocating space within it, making the `pubkey` required to be `isWritable`. `isSigner` is required when you are designating the account that is running the instruction. In this case, the signer is the account calling to allocate space within itself.
```

docs/src/developing/clients/javascript-api.md
```diff
@@ -2,4 +2,331 @@
...
+
+```javascript
+let allocateTransaction = new web3.Transaction({
+    feePayer: payer.publicKey
+  });
+let keys = [{pubkey: keypair.publicKey, isSigner: true, isWritable: true}];
+let params = { space: 100 };
+```
+
+We create the transaction `allocateTransaction`, keys, and params objects. `feePayer` is an optional field when creating a transaction that specifies who is paying for the transaction. `keys` represents all accounts that the program's `allocate` function will interact with. Since the `allocate` function also required space, we created `params` to be used later when invoking the `allocate` function.
```
joncinque, line 196: nit for clarification:
```suggestion
We create the transaction `allocateTransaction`, keys, and params objects. `feePayer` is an optional field when creating a transaction that specifies who is paying for the transaction, defaulting to the pubkey of the first signer in the transaction. `keys` represents all accounts that the program's `allocate` function will interact with. Since the `allocate` function also required space, we created `params` to be used later when invoking the `allocate` function.
```

joncinque (commented):
Just a couple more nits, then this is good to go from my end

cryptogosu:
@joncinque Updated all with requested changes

joncinque (approved):
looks good, thanks for the contribution!

PR #19469, opened 2021-08-27
Index loaders / executable accounts

Lichtso:
#### Problem
Loaders / executable accounts are still separate and need to be concatenated with the regular accounts in order to be indexed (transaction wide).
See #14523.

#### Summary of Changes
* Replaces `loaders` by `executable_indices`.
* Moves `MessageProcessor::create_keyed_accounts()` into `InvokeContext::push()`.
* WIP feature gate: Fill executable account gaps with the real deal directly (requires them to be read-only).

### DIFF
--- a/runtime/src/accounts.rs
+++ b/runtime/src/accounts.rs
@@ -101,11 +101,11 @@ pub struct Accounts {
 // for the load instructions
 pub type TransactionAccounts = Vec<(Pubkey, AccountSharedData)>;
 pub type TransactionRent = u64;
-pub type TransactionLoaders = Vec<Vec<(Pubkey, AccountSharedData, usize)>>;
+pub type TransactionProgramIndices = Vec<Vec<usize>>;
 #[derive(PartialEq, Debug, Clone)]
 pub struct LoadedTransaction {
     pub accounts: TransactionAccounts,
-    pub loaders: TransactionLoaders,
+    pub program_indices: TransactionProgramIndices,
     pub rent: TransactionRent,
     pub rent_debits: RentDebits,
 }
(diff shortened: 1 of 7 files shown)


Lichtso:
@CriesofCarrots While resolving merge conflicts I touched some of your recent changes. A quick assessment if the three tests ([see last commit](https://github.com/solana-labs/solana/pull/19469/commits/f04289226c04c0fd0f20add5f7cc6f0d45154218)) are still correct suffices.

CriesofCarrots (commented):
Tests in last commit look correct to me. Thanks!

jstarry:
@Lichtso I don't understand what problem this change fixed. Can you explain?

runtime/src/accounts.rs
```diff
@@ -338,44 +338,46 @@ impl Accounts {
...
+                    .checked_sub_lamports(fee)
+                    .map_err(|_| TransactionError::InsufficientFundsForFee)?;
+
+                let program_indices = message
+                    .instructions()
+                    .iter()
+                    .map(|instruction| {
+                        self.load_executable_accounts(
+                            ancestors,
+                            &mut accounts,
```
jstarry, line 369: Why do we need to add executable accounts to `accounts` here? Couldn't we add them above?

Lichtso: That is a preparation for the next step:
Currently we leave gaps for the executable accounts (so that the indices stay correct) and then load them for each instruction. Thus, they are loaded duplicate if multiple instructions use the same executable. And that could be avoided here, once we can make sure that accounts used as programs are not writable. (see #19629)

Lichtso:
> @Lichtso I don't understand what problem this change fixed. Can you explain?

You are right, the PR description was not particularly insightful.
I updated it and added some details, hope it is easier to understand now.

runtime/src/accounts.rs
jstarry: Ah yeah that is tricky. Someone could update program and then call it later within the same transaction, thanks for the explanation

jstarry:
Ah yeah that makes sense. I like that all accounts can be indexed in the same way but I think it would be nice to have a wrapper type which keeps account inputs and account deps separate internally but implements the [index trait](https://doc.rust-lang.org/std/ops/trait.Index.html#). This way the code is more straightforward and no one has to remember what is actually in the accounts vec

PR #19473, opened 2021-08-27
Add address map program

jstarry:
#### Problem
Users and protocols need a way to store addresses on-chain for transaction v2 address lookups.

#### Summary of Changes
Added address map program
- Number of entries is capped at `u8::MAX` since transaction v2 account indices are typed as `u8`
- Activation is instant but deactivation of on-chain map accounts has a cool down period to ensure that address maps can't be re-initialized at the same address.
- Support closing address map accounts but use derived addresses based on the current epoch to prevent address maps from being re-initialized at the same address.

Fixes #

### DIFF
--- a/programs/address-map-tests/tests/processor.rs
+++ b/programs/address-map-tests/tests/processor.rs
@@ -1,0 +1,246 @@
+use {
+    solana_address_map_program::{
+        id,
+        instruction::{
+            activate, close_account, deactivate, initialize_account, insert_entries, set_authority,
+        },
+        processor::process_instruction,
+        state::{AddressMap, AddressMapState},
+    },
+    solana_program_test::*,
+    solana_sdk::{
+        clock::{Epoch, Slot},
+        epoch_schedule::EpochSchedule,
+        instruction::InstructionError,
+        native_token::LAMPORTS_PER_SOL,
+        pubkey::Pubkey,
+        signature::{Keypair, Signer},
+        system_instruction::transfer,
+        transaction::{Transaction, TransactionError},
+    },
+};
+
+const WARP_EPOCH: Epoch = 1;
+
+async fn setup_warped_context() -> ProgramTestContext {
+    let program_test = ProgramTest::new("", id(), Some(process_instruction));
+    let mut context = program_test.start_with_context().await;
+    let epoch_schedule = EpochSchedule::default();
+    let warp_slot = epoch_schedule.get_first_slot_in_epoch(WARP_EPOCH);
+    context.warp_to_slot(warp_slot).unwrap();
+    context
+}
+
+#[tokio::test]
+async fn test_initialize_account() {
+    let mut context = setup_warped_context().await;
+    let client = &mut context.banks_client;
+    let payer = &context.payer;
+    let recent_blockhash = context.last_blockhash;
+
+    // First initialization should succeed
+    {
+        let authority_address = Pubkey::new_unique();
+        let (init_account_instruction, map_address) =
+            initialize_account(payer.pubkey(), authority_address, WARP_EPOCH, 0);
+        let transaction = Transaction::new_signed_with_payer(
+            &[init_account_instruction],
+            Some(&payer.pubkey()),
+            &[payer],
+            recent_blockhash,
+        );
+
+        assert!(client.process_transaction(transaction).await.is_ok());
+        let map_account = client.get_account(map_address).await.unwrap().unwrap();
+        let map = AddressMap::deserialize(&map_account.data).unwrap();
+        assert_eq!(map.authority, Some(authority_address));
+        assert_eq!(map.activation_slot, Slot::MAX);
+        assert_eq!(map.deactivation_slot, Slot::MAX);
+        assert_eq!(map.num_entries, 0u8);
+    }
+
+    // Second initialization should fail
+    {
+        let authority_address = Pubkey::new_unique();
+        let transaction = Transaction::new_signed_with_payer(
+            &[initialize_account(payer.pubkey(), authority_address, WARP_EPOCH, 0).0],
+            Some(&payer.pubkey()),
+            &[payer],
+            recent_blockhash,
+        );
+
+        assert_eq!(
+            client
+                .process_transaction(transaction)
+                .await
+                .unwrap_err()
+                .unwrap(),
+            TransactionError::InstructionError(0, InstructionError::AccountAlreadyInitialized),
+        );
+    }
+}
+
+#[tokio::test]
+async fn test_initialize_account_wrong_epoch() {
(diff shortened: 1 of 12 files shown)


jackcmay:
Should these be exposed via the sdk?

jstarry:
> Should these be exposed via the sdk?

Yeah, definitely. I think I can use cargo features to have the sdk directly depend on the address map program crate without bringing in the program runtime.

programs/address-map/src/instruction.rs
```diff
@@ -0,0 +1,177 @@
...
+    ///
+    /// # Account references
+    ///   0. `[WRITE]` Address map account to close
+    ///   1. `[WRITE]` Recipient of closed account lamports
+    ///   2. `[SIGNER]` Current authority, if required.
+    CloseAccount,
+}
+
+/// Derive an address map address from a wallet address and the current epoch.
+pub fn derive_address_map_address(payer_address: &Pubkey, current_epoch: Epoch) -> (Pubkey, u8) {
```
joncinque, line 64: I'm worried that this will make discoverability of address maps difficult.  One of the great features of the associated token account is that you can immediately find someone's token account with just their wallet and the mint.  Here, you may have to iterate through a bunch of epochs in order to find their address map, or you'll just use `find_program_address` and do it the "easy" way.

What if, instead of an epoch, this is another `Pubkey`? That way, you can setup a different address map for each program that you interact with.  Since you've cleverly put this on a PDA dependent on the signer, there's no way for one of these to get hijacked after `CloseAccount`.

programs/address-map/src/instruction.rs
```diff
@@ -0,0 +1,177 @@
...
+        ],
+    );
+
+    (instruction, map_address)
+}
+
+/// Returns an instruction that updates the authority of an address map account.
+/// If the new authority is `None`, the address map account will be immutable.
+/// Inactive address maps cannot be made immutable.
+pub fn set_authority(
```
joncinque, line 100: Similar to an associated token account, should we discourage people from reassigning authority?  Immutability makes sense, since you may want to setup an address map for your program as a one-time procedure, and have it hard-coded into the program.

What are the use-cases for reassigning authority?  I'm thinking that a user may create an address map for a program, then give the authority over to one of the program addresses which performs the setup and adjusts it as needed after updates.

programs/address-map/src/instruction.rs
```diff
@@ -0,0 +1,177 @@
...
+    current_epoch: Epoch,
+    num_entries: u8,
+) -> (Instruction, Pubkey) {
+    let (map_address, bump_seed) = derive_address_map_address(&payer_address, current_epoch);
+    let instruction = Instruction::new_with_bincode(
+        id(),
+        &AddressMapInstruction::InitializeAccount {
+            bump_seed,
+            num_entries,
+            authority: authority_address,
```
joncinque, line 85: take it or leave it: Since we've had potential hijackings in the past, we may want to have a signature from the authority as well.  This way, bad integrations that just look for any account map for an authority can't get duped.

For example, if it's a defi application, the attacker sets up trap address map accounts for a bunch of addresses, using their own token accounts as the receivers, and just hopes that a bad integration uses the trap address map.

programs/address-map-tests/tests/processor.rs
```diff
@@ -0,0 +1,246 @@
...
+        let (init_account_instruction, map_address) =
+            initialize_account(payer.pubkey(), authority_address, WARP_EPOCH, 0);
+        let transaction = Transaction::new_signed_with_payer(
+            &[init_account_instruction],
+            Some(&payer.pubkey()),
+            &[payer],
+            recent_blockhash,
+        );
+
+        assert!(client.process_transaction(transaction).await.is_ok());
```
joncinque, line 53: nit, for better error messages in case there's an error, you can `unwrap()` these
```suggestion
        client.process_transaction(transaction).await.unwrap();
```

programs/address-map-tests/tests/processor.rs
```diff
@@ -0,0 +1,246 @@
...
+        ],
+        Some(&payer.pubkey()),
+        &[payer, &authority],
+        recent_blockhash,
+    );
+
+    assert!(client.process_transaction(transaction).await.is_ok());
+    let closed_account = client.get_account(map_address).await.unwrap().unwrap();
+    let uninitialized_data = bincode::serialize(&AddressMapState::Uninitialized).unwrap();
+    assert_eq!(
```
joncinque, line 241: clever!

joncinque (commented):
Looks really good! Most of my questions are nits and ideas about the interface, but I did look at the implementation as well, which looks safe

programs/address-map/src/processor.rs
```diff
@@ -0,0 +1,294 @@
...
+                }
+                map
+            } else {
+                return Err(InstructionError::UninitializedAccount);
+            };
+
+            map.authority = new_authority;
+            map_account.set_state(&AddressMapState::Initialized(map))?;
+        }
+        AddressMapInstruction::InsertEntries { offset, entries } => {
```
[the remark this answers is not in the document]
jstarry: In the new PR (https://github.com/solana-labs/solana/pull/21616), I renamed to "extend lookup table" which should imply that new addresses are appended

programs/address-map/src/state.rs
```diff
@@ -0,0 +1,190 @@
...
+#[allow(clippy::large_enum_variant)]
+pub enum AddressMapState {
+    /// Account is not initialized.
+    Uninitialized,
+    /// Initialized `AddressMap` account.
+    Initialized(AddressMap),
+}
+
+/// Data structure of address map
+#[derive(Debug, Serialize, Deserialize, PartialEq, Clone, AbiExample)]
```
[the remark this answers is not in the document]
jstarry: This program is a native program so BPF compute isn't at play

programs/address-map/src/instruction.rs
jstarry: I agree that it's not ideal and I did consider this point when reimplementing the program here: https://github.com/solana-labs/solana/pull/21616. However, I think indexing or creating a known registry of on-chain lookup tables is going to have to be the way forward here. I think it's more likely that bots and protocols will manage their own lookup table accounts and users will never have to manage them.

programs/address-map/src/instruction.rs
jstarry: Great call, I modified the create instruction here (https://github.com/solana-labs/solana/pull/21616) to use the authority in the PDA and enforce that it's a signer. It's also no longer possible to set the authority address to another address.

jstarry:
Addressed all relevant comments and closing this implementation in favor of a more dev-friendly version here: https://github.com/solana-labs/solana/pull/21616

PR #19477, opened 2021-08-27
Fix slot-to-hours estimation

CriesofCarrots:
#### Problem
We refer to this comment block to explain the effect of various `--limit-ledger-size` values based on cluster tps. But math in the 50k tps case seems off.

#### Summary of Changes
```
200_000_000 shreds / 5_000 shreds/slot = 40_000 slots
((40_000 slots * .4/s/slot) / 60s/min) / 60min/hr = 4.4444 hr
```
Also:
```
200_000_000 shreds / 60 shreds/slot = 3.33M slots
(((3.33M slots * .4/s/slot) / 60s/min) / 60min/hr) / 24h/day) = 15.4 days

### DIFF
--- a/core/src/ledger_cleanup_service.rs
+++ b/core/src/ledger_cleanup_service.rs
@@ -15,8 +15,8 @@ use std::time::Duration;
 
 // - To try and keep the RocksDB size under 400GB:
 //   Seeing about 1600b/shred, using 2000b/shred for margin, so 200m shreds can be stored in 400gb.
-//   at 5k shreds/slot at 50k tps, this is 500k slots (~5 hours).
-//   At idle, 60 shreds/slot this is about 4m slots (18 days)
+//   at 5k shreds/slot at 50k tps, this is 40k slots (~4.4 hours).
+//   At idle, 60 shreds/slot this is about 3.33m slots (~15 days)
 // This is chosen to allow enough time for
 // - A validator to download a snapshot from a peer and boot from it
 // - To make sure that if a validator needs to reboot from its own snapshot, it has enough slots locally

### CODE REVIEW

File: core/src/ledger_cleanup_service.rs [hunk 1]
```diff
@@ -15,7 +15,7 @@ use std::time::Duration;
 
 // - To try and keep the RocksDB size under 400GB:
 //   Seeing about 1600b/shred, using 2000b/shred for margin, so 200m shreds can be stored in 400gb.
-//   at 5k shreds/slot at 50k tps, this is 500k slots (~5 hours).
+//   at 5k shreds/slot at 50k tps, this is 500k slots (~55 hours).
```
steviez, line 18: This line tripped me up as well; however, my correction is held up as part of a larger PR so let's come to a consensus here and I'll rebase :).

I think the time value of (roughly) 5 hours is correct, but the slot value (500k) is incorrect.
```
400 GB = 400,000,000,000 bytes * (1 shred / 2000 bytes) = 200,000,000 shreds // current comment agrees
200,000,000 shreds * (1 slot / 5,000 shreds) = 40,000 slots @ 50k TPS load
40,000 slots * 0.4 sec / slot * (1 min / 60 sec) * (1 hour / 60 min) = 4.44... hours
```

CriesofCarrots: Ohhhh, this makes so much more sense. I just trusted the slots, and didn't go further back.
I agree with your math and will fix up.

CriesofCarrots: @steviez , pushed. Please take a look

steviez (approved):
LGTM

CriesofCarrots:
@sakridge , would appreciate a sanity check on this, if you can.

sakridge (approved):
lgtm

PR #19478, opened 2021-08-27
collect rent in parallel for partitions

jeffwashington:
#### Problem
Work on parallelizing rent collection. It grows large with large # accounts in kin sim.
#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -65,6 +65,8 @@ use crate::{
 use byteorder::{ByteOrder, LittleEndian};
 use itertools::Itertools;
 use log::*;
+use rayon::iter::IntoParallelIterator;
+use rayon::iter::ParallelIterator;
 use rayon::ThreadPool;
 use solana_measure::measure::Measure;
 use solana_metrics::{datapoint_debug, inc_new_counter_debug, inc_new_counter_info};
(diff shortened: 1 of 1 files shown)


runtime/src/bank.rs
```diff
@@ -3890,9 +3892,11 @@ impl Bank {
         }
 
         let mut measure = Measure::start("collect_rent_eagerly-ms");
-        for partition in self.rent_collection_partitions() {
-            self.collect_rent_in_partition(partition);
-        }
+        self.rent_collection_partitions()
+            .into_par_iter()
```
sakridge, line 3896: I think this would use the full rayon threads, right? Any idea if it might impact other things going on? I suppose there is no transaction processing happening at this point.

runtime/src/bank.rs
```diff
@@ -65,6 +65,8 @@ use crate::{
 use byteorder::{ByteOrder, LittleEndian};
 use itertools::Itertools;
 use log::*;
+use rayon::iter::IntoParallelIterator;
+use rayon::iter::ParallelIterator;
```
sakridge, line 69: I think we're using the combined `use` pattern here. Do you mind combining the `rayon` entries to keep it consistent?

runtime/src/bank.rs
jeffwashington: I think the # of partitions will be small. I think we get > 1 when we have skipped slots? There are things here I don't really understand. I could limit the threads with a thread pool or chunk it or something else.

jeffwashington: I could add a metric for # partitions, but it should be 1 partition per slot, which should be 1 partition per 400ms, or 2.5 per second, right? If we get behind, it could spike. I think?

jeffwashington: using acct db's thread pool now.

sakridge (approved):

PR #19486, opened 2021-08-29
uses rayon thread-pool for retransmit-stage parallelization

behzadnouri:
#### Problem
* Retransmit threads are under utilized when there are bursts of shreds.
* Duplicate work across threads:
https://github.com/solana-labs/solana/blob/84db04ce6/core/src/retransmit_stage.rs#L253-L262

#### Summary of Changes
* The commit uses rayon thread-pool to dynamically batch shreds into tasks and assign to all available threads.

### DIFF
--- a/core/src/retransmit_stage.rs
+++ b/core/src/retransmit_stage.rs
@@ -11,11 +11,11 @@ use {
         completed_data_sets_service::CompletedDataSetsSender,
         packet_hasher::PacketHasher,
         repair_service::{DuplicateSlotsResetSender, RepairInfo},
-        result::{Error, Result},
         window_service::{should_retransmit_and_persist, WindowService},
     },
     crossbeam_channel::{Receiver, Sender},
     lru::LruCache,
+    rayon::{prelude::*, ThreadPool, ThreadPoolBuilder},
     solana_client::rpc_response::SlotUpdate,
     solana_gossip::cluster_info::{ClusterInfo, DATA_PLANE_FANOUT},
     solana_ledger::{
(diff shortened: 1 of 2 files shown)


behzadnouri:
This code is on the left vs current code on the right:
`retransmit-stage::epoch_cache_update`:
retransmit-stage epoch_cache_update
`retransmit-stage::epoch_fetch`:
retransmit-stage epoch_fetch

Other timing metrics are not directly comparable because they would be cpu time vs real time. But this chart compares the lag between the time the shreds are sent to retransmit channel until they are actually retransmitted:
retransmit-stage avg_shreds_done_elapsed_micros_2

core/src/retransmit_stage.rs
```diff
@@ -224,57 +175,51 @@ fn maybe_reset_shreds_received_cache(
...
     let mut shreds = shreds_receiver.recv_timeout(RECV_TIMEOUT)?;
     let mut timer_start = Measure::start("retransmit");
-    while let Ok(more_shreds) = shreds_receiver.try_recv() {
-        shreds.extend(more_shreds);
-        if shreds.len() >= MAX_SHREDS_BATCH_SIZE {
-            break;
-        }
-    }
-    drop(shreds_receiver);
+    shreds.extend(shreds_receiver.try_iter().flatten());
```
jbiseda, line 195: Is there no longer a need to limit the batch size received before starting to process the shreds?

core/src/retransmit_stage.rs
```diff
@@ -361,74 +297,56 @@ pub fn retransmitter(
...
-                    }
-                    trace!("exiting retransmitter");
-                })
-                .unwrap()
+    );
+    let mut hasher_reset_ts = Instant::now();
+    let mut stats = RetransmitStats::default();
+    let shreds_received = Mutex::new((LruCache::new(DEFAULT_LRU_SIZE), PacketHasher::default()));
+    let first_shreds_received = Mutex::<BTreeSet<Slot>>::default();
+    let num_threads = get_thread_count().min(8).max(sockets.len());
```
jbiseda, line 312: Was `8` chosen to match the value in `new_with_external_ip`? Could this be a shared constant?

core/src/retransmit_stage.rs
behzadnouri: With the previous code, one thread was locking the channel and, without a batch-size limit, if there was a burst of shreds on the channel, they would all be assigned to that single thread. So the batch-size limit was mitigating unbalanced workload distribution across the threads. Here, we don't have that issue anymore because assigning shreds to threads is done after consuming from the channel.

The other aspect is that if adding a batch-size limit will improve latencies for shreds processing; 
I think it will actually do more harm than good. Because rayon works better with larger batches (due to fixed-cost amortization), and adding a batch-size limit will force rayon to work sub-optimally causing overall larger average latencies when there is a heavy load by cutting the batches short.

core/src/retransmit_stage.rs
behzadnouri: I was borrowing that `.min(8)` from gossip:
https://github.com/solana-labs/solana/blob/d7051b0d2/gossip/src/cluster_info.rs#L1643-L1647

in order not to take too much cpu for one process. It does not actually need to match `sockets.len()`, and we can have more  threads than sockets.

behzadnouri:
@jbiseda @sakridge @carllin any objections to merging this?

jbiseda (approved):
looks good

PR #19506, opened 2021-08-30
parallelize index_read in shrink

jeffwashington:
#### Problem
shrink time gets large with high # of accounts
#### Summary of Changes
break up account index lookups into smaller chunks and process in parallel
Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -2266,36 +2266,64 @@ impl AccountsDb {
         }
 
         let mut index_read_elapsed = Measure::start("index_read_elapsed");
-        let mut alive_total = 0;
+        let alive_total_collect = AtomicUsize::new(0);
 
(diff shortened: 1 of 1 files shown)


sakridge:
Might be a sign that 1 store per slot is maybe too little once we get to these huge sizes. Shrinking may be very expensive with these large stores.

runtime/src/accounts_db.rs
```diff
@@ -2266,36 +2266,64 @@ impl AccountsDb {
...
-                        .fetch_add(1, Ordering::Relaxed);
-                }
-            }
-        }
+        let alive_total_collect = AtomicUsize::new(0);
+
+        let len = stored_accounts.len();
+        let alive_accounts_collect = Mutex::new(Vec::with_capacity(len));
+        let unrefed_pubkeys_collect = Mutex::new(Vec::with_capacity(len));
+        self.thread_pool.install(|| {
```
sakridge, line 2274: maybe a good time to refactor this part into it's own function?

PR #19507, opened 2021-08-30
parallel shrink

jeffwashington:
#### Problem
make shrink faster by using available cores
#### Summary of Changes

Fixes #


jeffwashington:
mnb, lemond:
earlier time is master, later time is with this pr:
image

lijunwangs (approved):
Looks good -- do we have a graph indicating what is the difference between with and without this change?

PR #19512, opened 2021-08-31
Populate memo in bigtable transaction structs

CriesofCarrots:
#### Problem
Payment providers would like to use memos to annotate transactions, but wallets can't easily display these memos in the tx-history list. `ConfirmedTransactionStatusWithSignature` and `RpcConfirmedTransactionStatusWithSignature` contain memo fields for this purpose, but these are not being populated. 

#### Summary of Changes
Populate memo field in bigtable structs `TransactionInfo` and `TransactionByAddrInfo`, which will by turn populate the memo field for `getSignaturesForAddress` response objects that are queried from bigtable.

Transactions will multiple memos will return them in one semicolon-separated String. I thought of other potential ways to handle/disambiguate multiple memo ixs in the same tx, but this seemed sufficient, especially since most use-cases will only have one memo/tx.

```
$ solana transaction-history F8vKjn13shKnef6pGfoHuWfpWP6DqmWT3xZK7oahnGuw -v

RPC URL: http://localhost:8899
Default Signer Path: id.json
Commitment: confirmed
3HtfiPUR67EFvkJia4u3TUYGQTVCva1jGpNpaT44UCKbcnSmUd6RFw5HNRtRd8Rxfz1mEe79u56JKhB6HZJGAT2i [slot=47 timestamp=2021-08-31T00:18:38Z status=Finalized] Test; This is a test second memo
5gESbb4gP15oy3U3RvHE9StEMzv55Qq67dVDqwcmV5AGvUTRiiEgc8HmtcYw2QXSV7SWp6T2WMmcqZsSfHx5kruC [slot=30 timestamp=2021-08-31T00:18:30Z status=Finalized]
52zpvpGSVGN58VfhXkf4duy67a3FpMcT9YC5QUTDEDCg7jbwsUFST2ByppWMdVGM84jKZHZMeDPLPjneruFdfPLj [slot=18 timestamp=2021-08-31T00:18:24Z status=Finalized] Beep beep boop boop; This is a test second memo
3 transactions found
```

Support for blockstore-queried transactions will require db changes to store the memo; separate PR.

PR #19515, opened 2021-08-31
Populate memo in blockstore signatures-for-address

CriesofCarrots:
#### Problem
cc #19512  Payment providers would like to use memos to annotate transactions, but wallets can't easily display these memos in the tx-history list. `ConfirmedTransactionStatusWithSignature` and `RpcConfirmedTransactionStatusWithSignature` contain memo fields for this purpose, but these are not being populated. 

#### Summary of Changes
Store memos in new blockstore TransactionMemos column via the TransactionStatusService, and use it to populate the memo field for `getSignaturesForAddress` response objects that are queried from blockstore.

This will require a stubbed TransactionMemos column in v1.6, or backporting both memo-population PRs to v1.6. Preference?

I also considered the alternate solution of a new AddressSignatures column to store the memo alongside each address, similar to the schema in bigtable, but ultimately decided that the very small perf savings in `Blockstore::get_confirmed_signatures_for_address2()` are not worth the additional disk to duplicate memos for every address. (However, it may still be worth plumbing a new AddressSignatures column sometime in the future to fix the column key to return signatures in proper execution order.)

### DIFF
--- a/transaction-status/src/extract_memos.rs
+++ b/transaction-status/src/extract_memos.rs
@@ -27,6 +27,14 @@ pub fn extract_and_fmt_memos<T: ExtractMemos>(message: &T) -> Option<String> {
     }
 }
 
+fn maybe_push_parsed_memo(memos: &mut Vec<String>, program_id: Pubkey, data: &[u8]) {
+    if program_id == spl_memo_id_v1() || program_id == spl_memo_id_v3() {
+        let memo_len = data.len();
+        let parsed_memo = parse_memo_data(data).unwrap_or_else(|_| "(unparseable)".to_string());
+        memos.push(format!("[{}] {}", memo_len, parsed_memo));
+    }
+}
+
 pub trait ExtractMemos {
     fn extract_memos(&self) -> Vec<String>;
 }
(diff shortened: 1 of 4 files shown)


transaction-status/src/extract_memos.rs
```diff
@@ -15,29 +18,55 @@ pub fn spl_memo_id_v3() -> Pubkey {
...
+            || self.account_keys.contains(&spl_memo_id_v3())
+        {
+            for instruction in &self.instructions {
+                let program_id = self.account_keys[instruction.program_id_index as usize];
+                if program_id == spl_memo_id_v1() || program_id == spl_memo_id_v3() {
+                    let memo_len = instruction.data.len();
+                    let parsed_memo = parse_memo_data(&instruction.data)
+                        .unwrap_or_else(|_| "(unparseable)".to_string());
+                    memos.push(format!("[{}] {}", memo_len, parsed_memo));
+                }
```
mvines, line 47: Worth de-duping this code with lines 62-66?

mvines (dismissed):
> This will require a stubbed TransactionMemos column in v1.6, or backporting both memo-population PRs to v1.6. Preference?

A full backport seems worthwhile as this looks low risk and gets us nice memos in RPC sooner

> However, it may still be worth plumbing a new AddressSignatures column sometime in the future to fix the column key to return signatures in proper execution order

Can you please remind me of the limitation here?  Signatures from the same block aren't returned in chronological entry order?

CriesofCarrots:
> Can you please remind me of the limitation here? Signatures from the same block aren't returned in chronological entry order?

@mvines , yes, that's right. The key on the AddressSignatures column is `(unused_u8, Pubkey, Slot, Signature`, which means that within each slot, the signatures are stored and return in order of Signature bytes. A fixed AddressSignatures column would be keyed: `(Pubkey, Slot, SlotIndex, Signature)`

mvines (approved):

mvines:
> > Can you please remind me of the limitation here? Signatures from the same block aren't returned in chronological entry order?
> 
> @mvines , yes, that's right. The key on the AddressSignatures column is `(unused_u8, Pubkey, Slot, Signature`, which means that within each slot, the signatures are stored and return in order of Signature bytes. A fixed AddressSignatures column would be keyed: `(Pubkey, Slot, SlotIndex, Signature)`

Thanks. This *may* become an issue with the confidential token transfer work that we're looking at, I'll make a note and we can revisit it soon-ish if so!

PR #19517, opened 2021-08-31
Packet/shred batch timer

jbiseda:
#### Problem
Metrics don't capture batching behavior as packets move through the system.

#### Summary of Changes
Record timings for incoming batches of packets. Propagate this information as the packets are processed through the system. For example, this enables the correlation of outgoing retransmitted packets to incoming packets.

Fixes #


jbiseda:
Some measurements which show interesting outliers:

window_service:
Screen Shot 2021-08-31 at 12 10 19 AM
Screen Shot 2021-08-31 at 12 11 00 AM

retransmit_stage:
Screen Shot 2021-08-31 at 12 07 53 AM
Screen Shot 2021-08-31 at 12 09 37 AM

ledger/src/shred.rs
```diff
@@ -211,6 +214,15 @@ pub struct Shred {
     pub payload: Vec<u8>,
 }
 
+#[derive(Debug, Default, Clone)]
+pub struct Shreds {
+    pub shreds: Vec<Shred>,
+    pub timer: PacketTimer,
+}
+
+pub type ShredReceiver = Receiver<Shreds>;
```
carllin, line 223: nit: `ShredReceiver -> ShredsReceiver`

ledger/src/shred.rs
```diff
@@ -211,6 +214,15 @@ pub struct Shred {
...
+pub type ShredSender = Sender<Shreds>;
```
carllin, line 224: nit: ShredSender -> ShredsSender

ledger/src/shred.rs
```diff
@@ -211,6 +214,15 @@ pub struct Shred {
     pub payload: Vec<u8>,
 }
 
+#[derive(Debug, Default, Clone)]
+pub struct Shreds {
+    pub shreds: Vec<Shred>,
```
carllin, line 219: nit: `shreds -> inner_shreds` so we can avoid calls that look like `shreds.shreds`

core/src/broadcast_stage/broadcast_duplicates_run.rs
```diff
@@ -341,8 +341,12 @@ impl BroadcastRun for BroadcastDuplicatesRun {
         blockstore: &Arc<Blockstore>,
     ) -> Result<()> {
         let (all_shreds, _) = receiver.lock().unwrap().recv()?;
+        let mut shreds = Shreds::default();
+        shreds.shreds = all_shreds.to_vec();
+        // TODO mark timer origin
```
carllin, line 346: Is this TODO already addressed? Might be good to create an issue for outstanding TODO's and reference the issue.

perf/src/packet.rs
```diff
@@ -3,29 +3,87 @@ use crate::{cuda_runtime::PinnedVec, recycler::Recycler};
...
 pub const PACKETS_PER_BATCH: usize = 128;
 pub const NUM_RCVMMSGS: usize = 128;
 
+#[derive(Copy, Clone, Debug, Default)]
+pub struct PacketTimer {
+    incoming_start: Option<Instant>,
+    incoming_initial_end: Option<Instant>,
+    incoming_end: Option<Instant>,
+    outgoing_start: Option<Instant>,
+    num_coalesced: usize,
```
carllin, line 19: maybe a comment on these, especially distinguishing `incoming_initial_end` and `incoming_end`

perf/src/packet.rs
```diff
@@ -3,29 +3,87 @@ use crate::{cuda_runtime::PinnedVec, recycler::Recycler};
...
+    }
+
+    pub fn get_incoming_end(&self) -> Option<Instant> {
+        self.incoming_end
+    }
+
+    pub fn get_num_coalesced(&self) -> usize {
+        self.num_coalesced
+    }
+}
```
carllin, line 62: nit:  all our getters in the codebase forgo the `get()` prefix

streamer/src/packet.rs
```diff
@@ -30,16 +30,20 @@ pub fn recv_from(obj: &mut Packets, socket: &UdpSocket, max_wait_ms: u64) -> Res
         );
         match recv_mmsg(socket, &mut obj.packets[i..]) {
             Err(_) if i > 0 => {
+                break;
+                /*
                 if start.elapsed().as_millis() as u64 > max_wait_ms {
                     break;
                 }
+                */
```
carllin, line 38: Is this safe to delete? Sorry, I'm lacking some context here

core/src/window_service.rs
```diff
@@ -358,28 +447,69 @@ where
...
+        .iter()
+        .map(|pkts| pkts.timer.get_num_coalesced())
+        .sum();
     stats.elapsed += now.elapsed();
+    stats.num_packet_batches += packets.len();
+    stats.num_coalesced_batches += coalesced_batches;
+    stats.elapsed_first_recv_to_start +=
+        now - packets.first().unwrap().timer.get_incoming_start().unwrap();
+    stats.elapsed_last_recv_to_start +=
+        now - packets.last().unwrap().timer.get_incoming_end().unwrap();
```
carllin, line 492: Will this aggregate information be useful without know how many `packets` we've processed? I see above we increment `num_packet_batches` by `packets.len()`, but we only add to `elapsed_first_recv_to_start` once per `packets`

perf/src/packet.rs
```diff
@@ -3,29 +3,87 @@ use crate::{cuda_runtime::PinnedVec, recycler::Recycler};
...
+    }
+
+    pub fn mark_outgoing_start(&mut self) {
+        if self.outgoing_start == None {
+            self.outgoing_start = Some(Instant::now());
+        }
+    }
+
+    pub fn extend_incoming_from(&mut self, pkt_timer: &PacketTimer) {
+        //debug_assert!(self.incoming_end.unwrap() < newer.incoming_end.unwrap());
```
carllin, line 44: dead code?

carllin (commented):
@jbiseda cool stuff, thanks for plumbing all this through 😃 

I think I get the rough idea of the goal here, but could you update the PR description with some of the specifics for:

For the newly added metrics like `elapsed_first_recv_to_start_micros`, `elapsed_last_recv_to_start_micros`, `batch_span_us_50pct`, `batch_span_us_90pct`, etc., what the goal is we're trying to measure / problems they could help us debug?

 For example, a comment describing:

* The `*_first_recv_to_*` style of metrics is measuring the time from pulling the packets off the wire to when they arrive at each of the components like ledger insert, retransmit, window service. This could be useful for debugging if components are lagging anywhere

* The `*span*` type of metrics are measuring if we have congestion in the pipe or we're spinning too long ingesting packets and they're all being coalesced together

streamer/src/packet.rs
jbiseda: This was just a quick test to see how this would show up in packet timing data. Added results to the description. I'll clean this up before moving the review out of draft status.

ledger/src/shred.rs
jbiseda: done

perf/src/packet.rs
jbiseda: removed incoming_initial_end

core/src/window_service.rs
jbiseda: removed these stats, they're captured by the histograms

perf/src/packet.rs
jbiseda: removed

jbiseda:
sample measurements from testnet with e2-highmem-16

receive_window:
Screen Shot 2021-09-09 at 4 33 25 AM
Screen Shot 2021-09-09 at 4 33 51 AM
Screen Shot 2021-09-09 at 4 34 50 AM

retransmit_stage:
Screen Shot 2021-09-09 at 4 35 26 AM
Screen Shot 2021-09-09 at 4 35 57 AM
Screen Shot 2021-09-09 at 4 37 05 AM
Screen Shot 2021-09-09 at 6 01 19 AM

PR #19522, opened 2021-08-31
remove redundant cancelled_shrink

jeffwashington:
#### Problem
Somehow I accidentally added a redundant metric.
#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -1222,7 +1222,6 @@ struct ShrinkStats {
     skipped_shrink: AtomicU64,
     dead_accounts: AtomicU64,
     alive_accounts: AtomicU64,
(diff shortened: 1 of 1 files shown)


runtime/src/accounts_db.rs
```diff
@@ -2311,9 +2305,6 @@ impl AccountsDb {
             self.shrink_stats
                 .skipped_shrink
                 .fetch_add(1, Ordering::Relaxed);
-            self.shrink_stats
```
lijunwangs, line 2314: oh. redundant as skipped_shrink?

lijunwangs (approved):

runtime/src/accounts_db.rs
jeffwashington: yep

PR #19524, opened 2021-08-31
add Copy trait to AccountInfo for fast copies to mmapped file

jeffwashington:
#### Problem
When we move to disk based accounts index, we would like to be able to copy a slotlist into the mmapped file. That requires the struct support the Copy trait unless perhaps if we want to do some unsafe tricks.
This change doesn't change anything yet - it will be valuable when we move to disk based index. However, it is very noisy with tests.
#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -212,7 +212,7 @@ impl GenerateIndexTimings {
     }
 }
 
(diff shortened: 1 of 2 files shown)


jeffwashington:
A side effect may be unintentional copies in new code.

carllin (approved):

PR #19529, opened 2021-08-31
Revert "Add LastFullSnapshotSlot to SnapshotConfig (#19341)"

brooksprumo:
#### Problem

LastFullSnapshotSlot is no longer used.

This was the meant to be the second version of how incremental snapshot support would be added to the background services, but a different implementation has been used instead (see PR #19401 if interested).

#### Summary of Changes

This reverts commit 4d361af976af79125271dd8c08cbea97711c3eb3, which removes LastFullSnapshotSlot.


jeffwashington (approved):
lgtm

PR #19530, opened 2021-08-31
Revert "Add snapshot_runtime_info module (#18199)"

brooksprumo:
#### Problem

The snapshot_runtime_info_module is not used. This was originally meant to be how incremental snapshot support would be added to the background services, but a different implementation has been used instead (see PR #19401 if interested).

#### Summary of Changes

This reverts commit c1b9d40a64e57375e4298d8d96345a4bb17b5de3, which removes the snapshot_runtime_info module.


jeffwashington (approved):
lgtm

PR #19531, opened 2021-08-31
add Debug trait to range for accounts index

jeffwashington:
#### Problem
It is frequently useful to log the range while debugging.
#### Summary of Changes
Add Debug trait.
Fixes #


brooksprumo (commented):
lgtm!

PR #19532, opened 2021-08-31
tests which assert or panic in other threads now will report an error

jeffwashington:
#### Problem
Tests calling this function have a tendency to fail in a thread. The result is the test hangs and we get no report of an error.
#### Summary of Changes
Only wait for the test to finish for 100s. If it takes longer than that, then join the other threads so we can get the failure message.
Fixes #

### DIFF
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -12976,21 +12976,22 @@ pub(crate) mod tests {
         // Let threads run for a while, check the scans didn't see any mixed slots
         let min_expected_number_of_scans = 5;
         std::thread::sleep(Duration::new(5, 0));
-        let mut loops = 1000;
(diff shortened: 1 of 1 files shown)


runtime/src/bank.rs
```diff
@@ -12976,12 +12976,17 @@ pub(crate) mod tests {
         // Let threads run for a while, check the scans didn't see any mixed slots
         let min_expected_number_of_scans = 5;
         std::thread::sleep(Duration::new(5, 0));
+        let mut loops = 1000;
```
carllin, line 12979: nit: loops -> remaining_loops

carllin (approved):

carllin (approved):

PR #19536, opened 2021-08-31
remove type specifics to make funciton more flexible

jeffwashington:
#### Problem
minor annoyance I keep running into. No need to specify type here.
#### Summary of Changes

Fixes #


brooksprumo (approved):
Looks good to me

PR #19539, opened 2021-09-01
Add compute_budget_program to the native_program_id list

jackcmay:
#### Problem

Compute budget program id missing from native program id list

#### Summary of Changes

Add it

Fixes #

### DIFF
--- a/sdk/program/src/compute_budget_program.rs
+++ b/sdk/program/src/compute_budget_program.rs
@@ -1,0 +1,1 @@
+crate::declare_id!("ComputeBudget111111111111111111111111111111");
--- a/sdk/program/src/lib.rs
+++ b/sdk/program/src/lib.rs
@@ -12,6 +12,7 @@ pub mod bpf_loader;
 pub mod bpf_loader_deprecated;
 pub mod bpf_loader_upgradeable;
 pub mod clock;
+pub mod compute_budget_program;
 pub mod decode_error;
 pub mod entrypoint;
 pub mod entrypoint_deprecated;
--- a/sdk/program/src/pubkey.rs
+++ b/sdk/program/src/pubkey.rs
@@ -1,7 +1,7 @@
 #![allow(clippy::integer_arithmetic)]
 use crate::{
-    bpf_loader, bpf_loader_deprecated, config, decode_error::DecodeError, feature, hash::hashv,
-    secp256k1_program, stake, system_program, sysvar, vote,
+    bpf_loader, bpf_loader_deprecated, compute_budget_program, config, decode_error::DecodeError,
+    feature, hash::hashv, secp256k1_program, stake, system_program, sysvar, vote,
 };
 
 use borsh::{BorshDeserialize, BorshSchema, BorshSerialize};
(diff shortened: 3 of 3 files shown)


mvines:
Do we really need to whitelist all new native programs here?  I feel it's overkill and causes feature activation pain (eg, what's happening over yonder in https://github.com/solana-labs/solana/pull/19266)

jackcmay:
@t-nelson Thoughts on whether we need to whitelist all native programs when computing program addresses?

jackcmay:
> @t-nelson Thoughts on whether we need to whitelist all native programs when computing program addresses?

Trent reminded me of: https://github.com/solana-labs/solana/security/advisories/GHSA-gh6m-fpgg-f9m8#advisory-comment-66920

@mvines @t-nelson 

Looks like we added these native program checks for good measure without citing a direct threat.  

I think we can probably exclude precompiles in that list since there aren't any accounts owned by them and they aren't programs that can be directly called via cpi.  This would mean removing the secp256k1 program and not including the compute_budget_program.

Thoughts on whether we should include the rest of the builtins?  If so I have a pr in progress that will allow us to featurize any changes to the list as well as keep the list in sync with the featurized one.

mvines:
> I think we can probably exclude precompiles in that list since there aren't any accounts owned by them and they aren't programs that can be directly called via cpi. This would mean removing the secp256k1 program and not including the compute_budget_program.

Keeping secp256k1 at this point seems harmless, I agree on not including compute_budget_program, or the upcoming ed25519_program

> Thoughts on whether we should include the rest of the builtins? If so I have a pr in progress that will allow us to featurize any changes to the list as well as keep the list in sync with the featurized one.

The current list again seems like we can just leave as it, the ROI on pruning it seems low.

jackcmay:
> The current list again seems like we can just leave as it, the ROI on pruning it seems low.

My concern is the longer-term confusion as to why we are including some ids and not others.  For example, we are working on a new program ABI that will come along with new loaders, should those loaders be added to the list?  They will interact with accounts they own and very probably use PDAs...

jackcmay:
> Yeah totally. Comments in the code on the motivation for that list could help as well

Yup, but that's the thing, we don't have a clear motivation besides, to be on the safe side, but the safe side of what? ;-)

Possible clear lines could one of the following:
-  all natively known ids except pre-compiles
-  any program that interacts with accounts it owns or depends on an account being signed

This list is also troubling because there isn't anything unique about these programs compared to developer deployed programs except they are known and some perform cluster critical duties.  Maybe that's enough, but one could raise the question of if we are protecting these programs against threat X, how are we protecting developer's programs against threat X.

mvines:
Amen brother! I'm not a fan of the list at all, because like you said it makes native programs subtly different from non-native programs.

jackcmay:
> Amen brother! I'm not a fan of the list at all, because like you said it makes native programs subtly different from non-native programs.

Check with @CherryWorm who initially brought up the threat related to this check being added and they said they don't see any reason to keep  the check either, so: https://github.com/solana-labs/solana/pull/19595

jackcmay:
We are removing the list altogether so closing his PR: https://github.com/solana-labs/solana/pull/19595

PR #19548, opened 2021-09-01
Add return data to VM

seanyoung:
Implementation for #19318 

Signed-off-by: Sean Young <sean@mess.org>

#### Summary of Changes

Fixes #


seanyoung:
I've created a PR for Solang to use this new system, and it works very well. The result is much nicer. :tada:

seanyoung:
Compared the proposal, there are two changes:
- return data is always cleared on entry (so we don't "return" data that the caller set)
- the syscalls charge  for their arguments, according to the length of data. Passing return data is not charged in itself, as the calls already charge.

programs/bpf_loader/src/syscalls.rs
```diff
@@ -2540,6 +2574,94 @@ fn call<'a>(
     Ok(SUCCESS)
 }
 
+// Return data hanlding
```
mvines, line 2577:
```suggestion
// Return data handling
```

sdk/src/process_instruction.rs
```diff
@@ -309,6 +313,19 @@ pub mod stable_log {
         ic_logger_msg!(logger, "Program log: {}", message);
     }
 
+    /// Log return data as from the program itself.
```
mvines, line 316: Just to be complete please also document here that the `Program return data:` will not be present if no return data was provided (as opposed to an empty `<program-generated-in-base64>` value)

sdk/src/process_instruction.rs
```diff
@@ -309,6 +313,19 @@ pub mod stable_log {
...
+    ///
+    /// The general form is:
+    ///
+    /// ```notrust
+    /// "Program return data: <program-generated-in-base64>"
```
mvines, line 321:
```suggestion
    /// "Program return data: <program-generated-data-in-base64>"
```

If we end up adding the program id to the return data (cc: https://github.com/solana-labs/solana/pull/19318#discussion_r704627320) then I think it'll be good to include the program id that generated the return data in this log as well

mvines (approved):
lgtm!  Would want @jackcmay's careful 👁️ on this too

jackcmay:
>     * the syscalls charge  for their arguments, according to the length of data. Passing return data is not charged in itself, as the calls already charge.

We want to avoid a program calling into the runtime and consuming cycles that are not accounted for.  For example, a program could call `sol_get_return_data` in a loop, and the program would incur a small cost of calling from BPF but a lot more unaccounted cost doing work in the syscall runtime.  That work in the runtime needs to be accounted for

programs/bpf/c/src/return_data/return_data.c
```diff
@@ -0,0 +1,35 @@
...
+  uint8_t buf[1024];
+
+  // There should be no return data on entry
+  uint64_t ret = sol_get_return_data(NULL, 0);
+
+  sol_assert(ret == 0);
+
+  // set some return data
+  sol_set_return_data((const uint8_t*)DATA, sizeof(DATA));
+
```
jackcmay, line 18: How about adding this pattern to the test after the data is set, I image folks will use it to get the length of the return value:
```rust
int64_t num_bytes = sol_get_return_data(NULL, 0);
```

sdk/program/src/program.rs
```diff
@@ -64,3 +64,42 @@ extern "C" {
...
+
+/// Get the return data from invoked program
+pub fn get_return_data() -> Vec<u8> {
+    #[cfg(target_arch = "bpf")]
+    {
+        let mut buf = [0u8; MAX_RETURN_DATA];
+
+        let size = unsafe { sol_get_return_data(buf.as_mut_ptr(), buf.len() as u64) };
+
+        buf[..size as usize].to_vec()
```
jackcmay, line 95: might want to take the min of `size` and `MAX_RETURN_DATA` for a bit of future proof and protection.

runtime/src/message_processor.rs
```diff
@@ -123,6 +124,7 @@ impl<'a> ThisInvokeContext<'a> {
             sysvars: RefCell::new(vec![]),
             blockhash,
             fee_calculator,
+            return_data: Vec::new(),
```
jackcmay, line 127: "Option` instead? to avoid allocs for invoke that don't use return data?

programs/bpf_loader/src/syscalls.rs
```diff
@@ -53,6 +53,9 @@ use thiserror::Error as ThisError;
 /// Maximum signers
 pub const MAX_SIGNERS: usize = 16;
 
+/// Maximum size return data
+pub const MAX_RETURN_DATA: u64 = 1024;
```
jackcmay, line 57: Can this be pulled from the SDK so it is defined in one place?

programs/bpf_loader/src/syscalls.rs
```diff
@@ -82,6 +85,8 @@ pub enum SyscallError {
     TooManyAccounts,
     #[error("Overlapping copy")]
     CopyOverlapping,
+    #[error("return data too large ({0} > {1})")]
```
jackcmay, line 88: caps, and would you please fix the other two lower case instances while you are in here?

programs/bpf_loader/src/syscalls.rs
```diff
@@ -2540,6 +2574,94 @@ fn call<'a>(
...
+        &mut self,
+        addr: u64,
+        len: u64,
+        _arg3: u64,
+        _arg4: u64,
+        _arg5: u64,
+        memory_mapping: &MemoryMapping,
+        result: &mut Result<u64, EbpfError<BpfError>>,
+    ) {
+        if len > MAX_RETURN_DATA {
```
jackcmay, line 2593: Consume first before doing anything else on the assumption that things will work out

programs/bpf_loader/src/syscalls.rs
```diff
@@ -2540,6 +2574,94 @@ fn call<'a>(
...
+        &mut self,
+        addr: u64,
+        len: u64,
+        _arg3: u64,
+        _arg4: u64,
+        _arg5: u64,
+        memory_mapping: &MemoryMapping,
+        result: &mut Result<u64, EbpfError<BpfError>>,
+    ) {
+        let invoke_context = question_mark!(
```
jackcmay, line 2637: consume first before any work

programs/bpf_loader/src/syscalls.rs
```diff
@@ -2540,6 +2574,94 @@ fn call<'a>(
...
+            self.invoke_context
+                .try_borrow()
+                .map_err(|_| SyscallError::InvokeContextBorrowFailed),
+            result
+        );
+
+        let return_data = invoke_context.get_return_data();
+
+        let length = std::cmp::min(return_data.len() as u64, len);
```
jackcmay, line 2646: nit: move `std::cmp::min` to a `use` statement at the top

programs/bpf_loader/src/syscalls.rs
```diff
@@ -2540,6 +2574,94 @@ fn call<'a>(
...
+        );
+
+        let return_data_result = question_mark!(
+            translate_slice_mut::<u8>(memory_mapping, addr, length, self.loader_id, true,),
+            result
+        );
+
+        return_data_result.copy_from_slice(&return_data[..length as usize]);
+
+        *result = Ok(return_data.len() as u64);
```
jackcmay, line 2661:
```suggestion
        *result = Ok(length);
```

sdk/bpf/c/inc/sol/return_data.h
```diff
@@ -0,0 +1,39 @@
+#pragma once
+/**
+ * @brief Solana keccak system call
```
jackcmay, line 3:
```suggestion
 * @brief Solana return data calls
```

sdk/program/src/program.rs
```diff
@@ -64,3 +64,42 @@ extern "C" {
         signers_seeds_len: u64,
     ) -> u64;
 }
+
+#[allow(unused)]
+const MAX_RETURN_DATA: usize = 1024;
```
jackcmay, line 69: Remove the attribute and make this `pub`?

jackcmay: Also add doc comment

sdk/program/src/program.rs
```diff
@@ -64,3 +64,42 @@ extern "C" {
...
+
+/// Set return data for program
```
jackcmay, line 71:
```suggestion
/// Set a program's return data
```

sdk/program/src/program.rs
```diff
@@ -64,3 +64,42 @@ extern "C" {
...
+        let size = unsafe { sol_get_return_data(buf.as_mut_ptr(), buf.len() as u64) };
+
+        buf[..size as usize].to_vec()
+    }
+
+    #[cfg(not(target_arch = "bpf"))]
+    crate::program_stubs::sol_get_return_data()
+}
+
+#[cfg(target_arch = "bpf")]
```
jackcmay, line 102: These can be placed inside their respective functions, take a look at examples in `Pubkey`

jackcmay: Go ahead and move `sol_invoke_signed_rust` in there while you are at it if you want

programs/bpf_loader/src/syscalls.rs
seanyoung: We would like the syscall to always return the actual length, even if less bytes are copied. This means that you can get the length with `sol_get_return_data(NULL, 0);`

programs/bpf_loader/src/syscalls.rs
seanyoung: Fixed

sdk/src/process_instruction.rs
seanyoung: So with the program_id, the return_data becomes an Option<>. Now it is present, even if it is empty.

sdk/src/process_instruction.rs
seanyoung: Absolutely. Now it included the program_id of the setter.

programs/bpf/c/src/return_data/return_data.c
seanyoung: Good idea, added.

sdk/program/src/program.rs
seanyoung: Fixed

runtime/src/message_processor.rs
seanyoung: So it has become an option now.

programs/bpf_loader/src/syscalls.rs
seanyoung: Yes, I missed that. Thanks for catching this.

seanyoung: Fixed

programs/bpf_loader/src/syscalls.rs
seanyoung: done

programs/bpf_loader/src/syscalls.rs
seanyoung: Changed

programs/bpf_loader/src/syscalls.rs
seanyoung: We need the length of the return buffer before we know how much to consume, can't change this.

programs/bpf_loader/src/syscalls.rs
seanyoung: Done

sdk/bpf/c/inc/sol/return_data.h
seanyoung: Fixed.

sdk/program/src/program.rs
seanyoung: Fixed.

sdk/program/src/program.rs
seanyoung: Fixed.

sdk/program/src/program.rs
seanyoung: This does look nicer, fixed.

seanyoung:
> We want to avoid a program calling into the runtime and consuming cycles that are not accounted for.  For example, a program could call `sol_get_return_data` in a loop, and the program would incur a small cost of calling from BPF but a lot more unaccounted cost doing work in the syscall runtime.  That work in the runtime needs to be accounted for

Yes, I agree. I've added a syscall_base_cost. 

Since cpi_bytes_per_unit is 250, and most of the time there will be less than 250 bytes return data, this part will usually be zero.

programs/bpf_loader/src/syscalls.rs
```diff
@@ -2540,6 +2573,137 @@ fn call<'a>(
...
+                ),
+                result
+            );
+
+            program_id_result[0] = *program_id;
+
+            // Return the actual length, rather the length returned
+            *result = Ok(return_data.len() as u64);
+        } else {
+            question_mark!(
```
jackcmay, line 2695: Charge base cost always, then charge byte cost if there are bytes to get

programs/bpf_loader/src/syscalls.rs
```diff
@@ -2540,6 +2573,137 @@ fn call<'a>(
...
+        let invoke_context = question_mark!(
+            self.invoke_context
+                .try_borrow()
+                .map_err(|_| SyscallError::InvokeContextBorrowFailed),
+            result
+        );
+
+        let budget = invoke_context.get_compute_budget();
+
+        if let Some((program_id, return_data)) = invoke_context.get_return_data() {
```
jackcmay, line 2655: Why not return 0 if `None`?

sdk/program/src/pubkey.rs
```diff
@@ -63,7 +63,7 @@ impl From<u64> for PubkeyError {
     Hash,
     AbiExample,
 )]
-pub struct Pubkey([u8; 32]);
+pub struct Pubkey(pub [u8; 32]);
```
jackcmay, line 66: What drove this change?

sdk/src/compute_budget.rs
```diff
@@ -95,6 +97,7 @@ impl ComputeBudget {
             cpi_bytes_per_unit: 250,        // ~50MB at 200,000 units
             sysvar_base_cost: 100,
             secp256k1_recover_cost: 25_000,
+            syscall_base_cost: 1,
```
jackcmay, line 100: 100 seems more reasonable

programs/bpf_loader/src/syscalls.rs
jackcmay: Agreed, I misread which length that was

programs/bpf_loader/src/syscalls.rs
seanyoung: The idea is that a program can set an empty return data.

Having said that, I'm still not sure about this. This is overloading the length return with  a special "nothing set" value.

jackcmay: Have a use case in mind where a program would set an empty return value?/

sdk/program/src/pubkey.rs
seanyoung: Looks like it isn't needed. I was trying to use `.as_mut_ptr()` on pubkey and failing. I did not realise that `&mut pubkey` will implicitly cast to `*mut Pubkey`.

sdk/src/compute_budget.rs
seanyoung: Fixed

programs/bpf_loader/src/syscalls.rs
seanyoung: No, I don't I must admit. The only thought I had was that with program_id recorded, we need an option anyway so now empty return data is easier to implement.

For solidity only have functions with no return value will return 0 bytes, in which case we wouldn't be looking for return data anyway.

What do you think?

jackcmay: Yeah, I'm thinking if len is zero we return zero and don't touch the output data or program id

seanyoung: I've changed it. If a return data of zero length is set, then the return_data Option becomes none and this function return zero.

sdk/program/src/program.rs
```diff
@@ -54,13 +66,46 @@ pub fn invoke_signed(
...
+        let mut buf = [0u8; MAX_RETURN_DATA];
+        let mut program_id = Pubkey::default();
+
+        let size =
+            unsafe { sol_get_return_data(buf.as_mut_ptr(), buf.len() as u64, &mut program_id) };
+
+        if size == 0 {
+            None
+        } else {
+            let size = std::cmp::min(size as usize, MAX_RETURN_DATA);
```
jackcmay, line 104: nit: use a `use` for `std::cmp`

jackcmay (dismissed):
Looks great besides one last nit ;-)

sdk/src/process_instruction.rs
```diff
@@ -309,6 +317,25 @@ pub mod stable_log {
         ic_logger_msg!(logger, "Program log: {}", message);
     }
 
+    /// Log return data as from the program itself. This line will not be present if no return
+    /// data was set, or if the return data was set to zero length.
+    ///
+    /// The general form is:
+    ///
+    /// ```notrust
+    /// "Program return data: <program-generated-data-in-base64>"
```
mvines, line 326:
```suggestion
    /// "Program return data: <program-id> <program-generated-data-in-base64>"
```

sdk/program/src/program.rs
seanyoung: I agree having the full namespace there is a bit much. Since this is in a target == bpf block, we can't have a `use std::cmp::min;` at the top without `#[allow(unused)]'. So I've put the use in the same block.

sdk/src/process_instruction.rs
seanyoung: Thanks for catching that, fixed.

PR #19553, opened 2021-09-01
Simplify process_blockstore tests

brooksprumo:
Some of the blockstore_processor tests weren't using
`test_process_blockstore()`, but could.  I updated those tests, and also
changed to using `..` for ignoring, instead of `_`, since I'll be adding
another return value here shortly (thus reducing the need to change
these again).


jeffwashington (approved):
lgtm

PR #19554, opened 2021-09-01
make test pass with 1 bin to remove noisy failure during testing

jeffwashington:
#### Problem
While tuning or debugging changes, sometimes you use 1 bin. This currently fails a test needlessly.
#### Summary of Changes
make the test pass if there is only 1 bin.
Fixes #


brooksprumo (approved):
lgtm

PR #19559, opened 2021-09-01
Add Incremental Snapshot support to RPC

### LINKED ISSUE #19579: Make RPC aware of incremental snapshots

brooksprumo:
#### Problem

There's no way to get incremental snapshot information from RPC.

#### Summary of Changes

- Add new RPC method, `getHighestSnapshotSlot` that returns a `SnapshotSlotInfo`, which contains both the highest full snapshot slot, and the highest incremental snapshot slot _based on_ the full snapshot.
- Deprecate old RPC method, `getSnapshotSlot`
- Update API docs

Fixes #19579


mvines:
Is there a reason to have three methods here?  It seems like just `getSnapshotSlots` would be sufficient?  

Aside: I don't love the name `getSnapshotSlots`.  To me that implies it'll return multiple slots that snapshots are available for, rather than a full/incremental snapshot slot pair

brooksprumo:
@mvines 

> Is there a reason to have three methods here? It seems like just `getSnapshotSlots` would be sufficient?

Good point. I this will likely come down to desired usage. I can see two use cases:
1. A node already has a full snapshot, and wants to get the highest incremental snapshot for their full snapshot. This would only be useful as long as the slot hasn't crossed a full snapshot interval.
2. A node just wants to get the latest and is OK downloading a new full snapshot if necessary.

Use-case 1 may not be all that valuable though. If a node `A` has a full snapshot at slot 800,000, and node `B` has a full snapshot at slot 810,000, then node `A` won't be able to use an incremental snapshot from node `B` anyway. So getting both the highest full snapshot and highest incremental snapshot would be the right way to go.

OK, I'm on board with just having a single RPC function!

> Aside: I don't love the name `getSnapshotSlots`. To me that implies it'll return multiple slots that snapshots are available for, rather than a full/incremental snapshot slot pair

Yeah, the name sucks. How about one of these:

- `getSnapshotInfo`
- `getSnapshotSlot2`
- `getSnapshotSlotPair`

I'm in favor of `getSnapshotInfo`.

mvines:
How about: `getHighestSnapshotSlot(full: Number, incremental: Number | undefined)`

mvines:
Oh, also lets turn the response from `getHighestSnapshotSlot` into a struct with these fields.  If `getSnapshotSlot` returned an object instead of a simple number, we'd be able to extend it in a backwards-compatible way and not be stuck with creating a new method!

client/src/rpc_response.rs
```diff
@@ -435,3 +435,9 @@ impl From<ConfirmedTransactionStatusWithSignature> for RpcConfirmedTransactionSt
         }
     }
 }
+
+#[derive(Serialize, Deserialize, Clone, Copy, Debug, PartialEq)]
+pub struct HighestSnapshotSlot {
+    pub full_snapshot_slot: Slot,
+    pub incremental_snapshot_slot: Option<Slot>,
+}
```
mvines, line 443: mebbie something like this:
```suggestion
pub struct SnapshotSlotInfo {
    pub full: Slot,
    pub incremental: Option<Slot>,
}
```

brooksprumo: Can do.

Also, it looks like most of the RPC responses that are structs are prefixed with `Rpc`; should I do that here (i.e. `RpcSnapshotSlotInfo`)?

docs/src/developing/clients/jsonrpc-api.md
```diff
@@ -2138,8 +2138,8 @@ None
 #### Results:
 
 - `<object>`
-  - `full_snapshot_slot: <u64>` - Highest full snapshot slot
-  - `incremental_snapshot_slot: <u64 | null>` - Highest incremental snapshot slot _based on_ `full_snapshot_slot`
+  - `full: <u64>` - Highest full snapshot slot
+  - `incremental: <u64 | null>` - Highest incremental snapshot slot _based on_ `full`
```
mvines, line 2142: I think `null` might actually be `undefined` here (serde_json serialization of `None`)?

brooksprumo: Done!

client/src/rpc_response.rs
brooksprumo: Done!

mvines (approved):

PR #19563, opened 2021-09-01
add debugging to test

jeffwashington:
#### Problem
when this test fails, it would be useful to know more information
#### Summary of Changes

Fixes #

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5530,7 +5530,7 @@ dependencies = [
  "bincode",
  "chrono",
  "crossbeam-channel",
- "futures 0.3.16",
+ "futures 0.3.17",
  "futures-util",
  "log 0.4.14",
  "prost",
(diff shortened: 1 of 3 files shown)


Cargo.lock
```diff
@@ -5541,7 +5541,7 @@ dependencies = [
  "solana-rpc",
  "solana-runtime",
  "solana-sdk",
- "tokio 1.10.1",
+ "tokio 1.11.0",
```
brooksprumo, line 5544: Interesting. Did your changes pull in these Cargo modifications? Or maybe this was to fix a CI check?

brooksprumo (approved):
Looks good other than the Cargo.lock question.

Cargo.lock
jeffwashington: i suspect someone will update master. I didn't cause it that I can imagine.

CriesofCarrots: I merged bumps to futures and tokio recently. Perhaps those prs missed bumping replica-lib, for some reason 🤔 
Anyway, shouldn't be an issue to pull in here.

PR #19587, opened 2021-09-02
Make wait_for_restart_window() aware of Incremental Snapshots

brooksprumo:
When the validator is waiting to restart, the snapshot slot is checked.
With Incremental Snapshots, the full snapshot interval is going to be
_much_ higher, which would make this function wait for a looooong time.

So, if there's an incremental snapshot slot when checking, use that
slot, otherwise fall back to the full snapshot slot.

Related to #17088

### DIFF
--- a/validator/src/main.rs
+++ b/validator/src/main.rs
@@ -278,14 +278,17 @@ fn wait_for_restart_window(
                     }
                 };
 
-                let full_snapshot_slot =
-                    snapshot_slot_info.map(|snapshot_slot_info| snapshot_slot_info.full);
+                let snapshot_slot = snapshot_slot_info.map(|snapshot_slot_info| {
+                    snapshot_slot_info
+                        .incremental
+                        .unwrap_or(snapshot_slot_info.full)
+                });
                 match in_leader_schedule_hole {
                     Ok(_) => {
                         if restart_snapshot == None {
-                            restart_snapshot = full_snapshot_slot;
+                            restart_snapshot = snapshot_slot;
                         }
-                        if restart_snapshot == full_snapshot_slot && !monitoring_another_validator {
+                        if restart_snapshot == snapshot_slot && !monitoring_another_validator {
                             "Waiting for a new snapshot".to_string()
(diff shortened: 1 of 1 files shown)


mvines:
For backwards compat, I think it'd be good to support the old RPC method as well as a fallback (until v1.9 perhaps).  Eg, I upgrade my validator software and then run `solana-validator exit`.  The new `solana-validator` binary is now talking to the old, running, `solana-validator` that doesn't know about the new RPC method.

brooksprumo:
> For backwards compat, I think it'd be good to support the old RPC method as well as a fallback (until v1.9 perhaps).

Ah, good point! Is there a common way to check for an RPC method/support old methods?

mvines:
Yeah I guess we should have done this as a part of #19559 'eh :)

`RpcClient::get_node_version()` is used internally for graceful fallback, so the implementation of `RpcClient::get_highest_snapshot_slot()` could check the node version and fall back to invoking `RpcRequest::GetSnapshotSlot` if less than 1.8.0

mvines (approved):

validator/src/main.rs
```diff
@@ -316,10 +319,18 @@ fn wait_for_restart_window(
                 "".to_string()
             } else {
                 format!(
-                    "| Full Snapshot Slot: {}",
+                    "| Full Snapshot Slot: {} | Incremental Snapshot Slot: {}",
```
behzadnouri, line 322: This is now printing a line exceeding terminal width, and apparently has broken progress-bar.
Instead of re-writing the same line, it is printing a new line each time.

Or maybe `42` below needs to be adjusted:
https://github.com/solana-labs/solana/blob/544a957ad/validator/src/lib.rs#L128

behzadnouri: @brooksprumo can we have this fixed?
`solana-validator monitor` is very annoying with this.

behzadnouri: Maybe the one showed in `monitor` is caused by https://github.com/solana-labs/solana/pull/19559
Either way this needs to be fixed.

Issue #19568, opened 2021-09-01
web3.js: Connection.simulateTransaction could be better


CriesofCarrots:
#### Problem
`Connection.simulateTransaction` doesn't support the `accounts` parameter exposed in RPC.
Also, it's difficult to use when starting with a serialized Message.

#### Proposed Solution
- Extend `simulateTransaction` to accept `Transaction | Message`; will likely need to fill in dummy signatures when extending the Message, as here: https://github.com/solana-labs/solana/blob/2f3df8cce0711ad33ecf1b5c9fe056c2f85a0f06/explorer/src/pages/inspector/SimulatorCard.tsx#L145-L150
- Add inputs to support passing the `accounts` param to RPC. Possible options could include: all accounts, non-program accounts, or specified array. ... I'm not sure that "all" is actually necessary.
- Add a helper fn or two to Message toward ^, possibly:
  - `isProgramId(index: number): boolean`
  - `programIds(): Array<PublicKey>`

Issue #19569, opened 2021-09-01
web3.js: Transaction.populate requires dummy signatures


CriesofCarrots:
#### Problem
If you call `Transaction.populate(message)` with no signers, all of the `accountKeys` in the the resulting TransactionInstructions will be marked `isSigner: false`, even though the Message.header contains the data needed to set that flag correctly.
This makes the method inconvenient for preparing a Transaction for simulation. Callers can work around this by providing dummy signatures, but not great ux. eg:
https://github.com/solana-labs/solana/blob/2f3df8cce0711ad33ecf1b5c9fe056c2f85a0f06/explorer/src/pages/inspector/SimulatorCard.tsx#L145-L150

#### Proposed Solution
Replace this logic
https://github.com/solana-labs/solana/blob/ed2c071fe1f8b24695ba392ba687214f34085c41/web3.js/src/transaction.ts#L691

PR #19590, opened 2021-09-02
web3.js: add accounts support to simulateTransaction

oJshua:
#### Problem
`Connection.simulateTransaction` doesn't support the `accounts` parameter exposed in RPC.
Also, it's difficult to use when starting with a serialized Message.

#### Summary of Changes

- Add support for Message
- include non-program accounts when includeAccounts is true or use specified PublicKey[]
- add helper functions to Message
- support default signatures when populating transaction from Message

Fixes https://github.com/solana-labs/solana/issues/19568, https://github.com/solana-labs/solana/issues/19569

### DIFF
--- a/web3.js/src/connection.ts
+++ b/web3.js/src/connection.ts
@@ -3517,8 +3517,8 @@ export class Connection {
         Array.isArray(includeAccounts)
           ? includeAccounts
           : message.accountKeys.filter(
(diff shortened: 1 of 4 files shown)


CriesofCarrots (approved):
Lgtm, thanks for taking this one!

PR #18373, opened 2021-07-01
Remove feature switch for demoting sysvar write locks

jstarry:
#### Summary of Changes
Removed feature switch for demoting sysvar write locks now that it is enabled on all clusters

PR #19593, opened 2021-09-02
Demote write locks on transaction program ids

CriesofCarrots:
#### Problem
Programs can't be written to, and taking a write lock on them can make batch sizes smaller than they should be.

#### Summary of Changes
Lower write lock to read lock for accounts marked as program-ids in a transaction.

### DIFF
--- a/cli-output/src/display.rs
+++ b/cli-output/src/display.rs
@@ -140,7 +140,7 @@ fn format_account_mode(message: &Message, index: usize) -> String {
         } else {
             "-"
         },
-        if message.is_writable(index) {
+        if message.is_writable(index, /*demote_program_write_locks=*/ true) {
             "w" // comment for consistent rust fmt (no joking; lol)
         } else {
             "-"
--- a/core/src/banking_stage.rs
+++ b/core/src/banking_stage.rs
@@ -1049,6 +1049,7 @@ impl BankingStage {
         libsecp256k1_fail_on_bad_count: bool,
         cost_tracker: &Arc<RwLock<CostTracker>>,
         banking_stage_stats: &BankingStageStats,
+        demote_program_write_locks: bool,
     ) -> (Vec<SanitizedTransaction>, Vec<usize>, Vec<usize>) {
         let mut retryable_transaction_packet_indexes: Vec<usize> = vec![];
 
@@ -1082,7 +1083,8 @@ impl BankingStage {
             verified_transactions_with_packet_indexes
                 .into_iter()
                 .filter_map(|(tx, tx_index)| {
-                    let result = cost_tracker_readonly.would_transaction_fit(&tx);
+                    let result = cost_tracker_readonly
+                        .would_transaction_fit(&tx, demote_program_write_locks);
                     if result.is_err() {
                         debug!("transaction {:?} would exceed limit: {:?}", tx, result);
                         retryable_transaction_packet_indexes.push(tx_index);
@@ -1165,6 +1167,7 @@ impl BankingStage {
                 bank.libsecp256k1_fail_on_bad_count(),
                 cost_tracker,
                 banking_stage_stats,
+                bank.demote_program_write_locks(),
             );
         packet_conversion_time.stop();
         inc_new_counter_info!("banking_stage-packet_conversion", 1);
@@ -1201,7 +1204,10 @@ impl BankingStage {
         let mut cost_tracking_time = Measure::start("cost_tracking_time");
         transactions.iter().enumerate().for_each(|(index, tx)| {
             if unprocessed_tx_indexes.iter().all(|&i| i != index) {
-                cost_tracker.write().unwrap().add_transaction_cost(tx);
+                cost_tracker
+                    .write()
+                    .unwrap()
+                    .add_transaction_cost(tx, bank.demote_program_write_locks());
             }
         });
         cost_tracking_time.stop();
@@ -1268,6 +1274,7 @@ impl BankingStage {
                 bank.libsecp256k1_fail_on_bad_count(),
                 cost_tracker,
                 banking_stage_stats,
+                bank.demote_program_write_locks(),
             );
         unprocessed_packet_conversion_time.stop();
 
--- a/program-runtime/src/instruction_processor.rs
+++ b/program-runtime/src/instruction_processor.rs
@@ -4,6 +4,7 @@ use solana_sdk::{
     account::{AccountSharedData, ReadableAccount, WritableAccount},
     account_utils::StateMut,
     bpf_loader_upgradeable::{self, UpgradeableLoaderState},
+    feature_set::demote_program_write_locks,
     ic_logger_msg, ic_msg,
     instruction::{CompiledInstruction, Instruction, InstructionError},
     keyed_account::{keyed_account_at_index, KeyedAccount},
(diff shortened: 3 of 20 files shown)


CriesofCarrots:
@taozhu-chicago would you mind sanity checking the banking_stage to cost_model changes?

sdk/program/src/message/legacy.rs
```diff
@@ -353,7 +353,7 @@ impl Message {
         self.program_position(i).is_some()
     }
 
-    pub fn is_writable(&self, i: usize) -> bool {
+    pub fn is_writable(&self, i: usize, demote_program_write_locks: bool) -> bool {
```
jackcmay, line 356: Is this a backward compatibility issue?

cli-output/src/display.rs
```diff
@@ -140,7 +140,7 @@ fn format_account_mode(message: &Message, index: usize) -> String {
         } else {
             "-"
         },
-        if message.is_writable(index) {
+        if message.is_writable(index, /*demote_program_write_locks=*/ true) {
```
jackcmay, line 143: What if instead we "correct" the message first, then we don't have to modify `is_writable`

core/src/banking_stage.rs
```diff
@@ -1201,7 +1204,10 @@ impl BankingStage {
         let mut cost_tracking_time = Measure::start("cost_tracking_time");
         transactions.iter().enumerate().for_each(|(index, tx)| {
             if unprocessed_tx_indexes.iter().all(|&i| i != index) {
-                cost_tracker.write().unwrap().add_transaction_cost(tx);
+                cost_tracker
```
jackcmay, line 1207: If we "correct" the tx here we don't have to pass in the bool

jackcmay:
Be nice to also demote write locks by default in tx/message/instruction construction

sdk/program/src/message/legacy.rs
CriesofCarrots: I don't think so, but tell me more about why you think there might be.

cli-output/src/display.rs
CriesofCarrots: I guess I'm not excited about mucking with the actual message. It requires changing both the header and potentially re-ordering the account_keys, plus Message/SanitizedMessage is generally not mutable.
Not a big deal just for the CLI, but ickier other places.

Incidentally, a number of these changes, including this one, are identical to how we did it for sysvars: https://github.com/solana-labs/solana/pull/15497

CriesofCarrots:
> Be nice to also demote write locks by default in tx/message/instruction construction

Good idea, I'll add that.

sdk/program/src/message/legacy.rs
jackcmay: I'm just thinking from an SDK ABI perspective

cli-output/src/display.rs
jackcmay: I agree, reforming the message isn't ideal, but it saves a lot of work scattered all over and is probably more sound doing it in one place (or a lot fewer places.  It could be part of the `sanitize` step.

CriesofCarrots: Hmm, yeah I see your point. Let me see how noodly that approach is.

taozhu-chicago:
> @taozhu-chicago would you mind sanity checking the banking_stage to cost_model changes?

the changes on banking_stage/cost_model path look good to me.

taozhu-chicago (commented):
It does take a lot of plumbing to get to banking_stage, maybe there is another way cost_model can check for TX's write locks without diving deep into `is_writable()` call?

sdk/program/src/message/legacy.rs
jstarry: Well, it used to have this ABI before I [removed it](https://github.com/solana-labs/solana/pull/18373). So if anything this makes it better :)

cli-output/src/display.rs
jstarry: I also prefer to implement this for how #15497 was done because otherwise we need to keep the original tx around to ensure that we record the message as it was signed. Otherwise it won't pass replay.

jackcmay: I guess its only temporary, but dang that's a lot of plumbing :-(

sdk/program/src/message/legacy.rs
jackcmay: I guess we are making exceptions to our backward compatibility policy for these cases?

cli-output/src/display.rs
CriesofCarrots: > otherwise we need to keep the original tx around

Nuts, that's true. I guess we're stuck with plumbing either way, and this is less scary to me.

sdk/program/src/message/legacy.rs
CriesofCarrots: Urg, I guess so? Fortunately, I don't remember anyone noticing the last time we added or removed the bool.

CriesofCarrots:
> It does take a lot of plumbing to get to banking_stage, maybe there is another way cost_model can check for TX's write locks without diving deep into `is_writable()` call?

Not a bad idea @taozhu-chicago , but no worries for this pr. We need the changes in `is_writable` anyway, and have easy access to the bank/feature-set.

CriesofCarrots:
> Be nice to also demote write locks by default in tx/message/instruction construction

I tried this out here: https://github.com/CriesofCarrots/solana/commit/19ee2de13a9d24ee7923a188640b14a1aef6205c
It's messy, and spurs the question whether we should do the same for sysvars and builtin programs (as per `is_writable`). So I vote no on including.

program-runtime/src/instruction_processor.rs
```diff
@@ -347,6 +348,7 @@ impl InstructionProcessor {
         instruction: &'a CompiledInstruction,
         executable_accounts: &'a [(Pubkey, Rc<RefCell<AccountSharedData>>)],
         accounts: &'a [(Pubkey, Rc<RefCell<AccountSharedData>>)],
+        demote_sysvar_write_locks: bool,
```
jstarry, line 351: nit: old name used

sdk/program/src/message/sanitized.rs
```diff
@@ -192,10 +189,13 @@ impl SanitizedMessage {
 
     /// Returns true if the account at the specified index is writable by the
     /// instructions in this message.
-    pub fn is_writable(&self, index: usize) -> bool {
+    pub fn is_writable(&self, index: usize, demote_program_write_locks: bool) -> bool {
+        if demote_program_write_locks && self.is_invoked(index) {
+            return false;
+        }
```
jstarry, line 195: nit: not needed

jstarry (approved):
looks good to me, just a few nits

program-runtime/src/instruction_processor.rs
CriesofCarrots: 🤦‍♀️

sdk/program/src/message/sanitized.rs
CriesofCarrots: Ah, thanks! Yep, with https://github.com/solana-labs/solana/pull/19593/commits/db0f8d0afde705faf80743e891d5f31b1e517966 added this is redundant

jackcmay (approved):

CriesofCarrots:
Feature_set conflicts make me sad

solana-grimes:

KeKo6988 (approved):

PR #19596, opened 2021-09-02
RPC: Add "empty" account data encoding option

mvines:
Sometimes you want to invoke [getMultipleAccounts](https://docs.solana.com/developing/clients/jsonrpc-api#getmultipleaccounts) or [getProgramAccounts](https://docs.solana.com/developing/clients/jsonrpc-api#getprogramaccounts) on a large dataset and you don't care about the account data.  

There's no way to prevent RPC from returning those useless bytes, so add one.

### DIFF
--- a/account-decoder/src/lib.rs
+++ b/account-decoder/src/lib.rs
@@ -58,6 +58,7 @@ pub enum UiAccountEncoding {
     JsonParsed,
     #[serde(rename = "base64+zstd")]
     Base64Zstd,
+    Empty,
 }
 
 impl UiAccount {
@@ -117,6 +118,9 @@ impl UiAccount {
                     )
                 }
             }
+            UiAccountEncoding::Empty => {
+                UiAccountData::Binary(String::new(), UiAccountEncoding::Empty)
+            }
         };
         UiAccount {
             lamports: account.lamports(),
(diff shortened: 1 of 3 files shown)


CriesofCarrots:
✨ Nice idea

mvines:
> ✨ Nice idea

Thanks, do you like the use of "empty" here?  I had "skip" for a while but the account data field does still end up getting returned, but as the empty string.

mvines:
Another idea that would achieve my goal of reducing RPC egress and add some additional utility: a "length" encoding that returns the account data length instead of the data itself.

CriesofCarrots:
Just recalling that we have the `dataSlice` parameter, which should be able to achieve the same as Empty: `"dataSlice": {"length":0, "offset":<whatever>}`. But that requires a lot more characters in the request... I'm not opposed to supporting Empty also.
Length seems pretty useful!

mvines:
> Just recalling that we have the `dataSlice` parameter, which should be able to achieve the same as Empty: `"dataSlice": {"length":0, "offset":<whatever>}`.

🤔 good point.  Lemme sleep on whether I want to add this extra knob for sugar over that

mvines:
I went for https://github.com/solana-labs/solana/pull/19614 instead

PR #19600, opened 2021-09-03
Make startup aware of Incremental Snapshots

### LINKED ISSUE #19297: Make validator startup aware of Incremental Snapshots
(1) When a validator starts up, there is a call stack of:

```
1. Bank::exhaustively_free_unused_resource()
2. blockstore_processor::load_frozen_banks()
3. blockstore_processor::do_process_blockstore_from_root()
4. blockstore_processor::process_blockstore_from_root()
5. bank_forks_utils::load_from_snapshot()
6. bank_forks_utils::load()
7. validator::new_banks_from_ledger()
8. Validator::new()
9. validator/src/main::main()
```
...and `Bank::exhaustively_free_unused_resource()` needs to have the `last_full_snapshot_slot` parameter set correctly (right now it is `None`). So I need to figure out how to get that value and pass it through the call stack.

This is related to #18973 

(2) Additionally, when loading the frozen banks, if a root crosses a full snapshot boundary, we need to take a full snapshot at this point, otherwise the clean from (1) maybe we wrong, and also when ABS starts up, its first incremental snapshots may be wrong.

So need to take full snapshots, _plus_ pass the last full snapshot slot to ABS when it starts up.

brooksprumo:
Now that the background services are aware of incremental snapshots,
they need (1) the correct last full snapshot slot in order to clean
accounts correctly, and (2) all expected full snapshots to be available
when created an incremental snapshot based on them.

This commit fixes startup so both requirements from above are met.

At startup, the blockstore processor loads frozen banks.  Some of these
banks may be roots, and some of these roots may cross the full snapshot
interval.  If/when that happens, take a bank snapshot and queue the full
snapshot in the AccountsPackageSender.  And at the end of startup,
return the last full snapshot slot to pass into the background services.

Fixes #19297

## Overview

High level overview for `blockstore_processor::do_process_blockstore_from_root()`. The new arg `last_full_snapshot_slot` is populated if we loaded from a full snapshot (vs starting from genesis). Then `load_frozen_forks()` processes the blocks, and updates the value (`last_full_snapshot_slot`) in case a new root is found that crosses the full snapshot interval. Finally, the value is returned and gets passed along to AccountsBackgroundService for its starting value of `last_full_snapshot_slot`.

Additionally, if a new root is found during `load_frozen_forks()` that crosses the full snapshot interval, now take a snapshot (with `snapshot_utils::snapshot_bank()`).

### DIFF
--- a/core/src/tvu.rs
+++ b/core/src/tvu.rs
@@ -44,10 +44,10 @@ use solana_runtime::{
     bank_forks::BankForks,
     commitment::BlockCommitmentCache,
     snapshot_config::SnapshotConfig,
-    snapshot_package::PendingSnapshotPackage,
+    snapshot_package::{AccountsPackageReceiver, AccountsPackageSender, PendingSnapshotPackage},
     vote_sender_types::ReplayVoteSender,
 };
-use solana_sdk::{pubkey::Pubkey, signature::Keypair};
+use solana_sdk::{clock::Slot, pubkey::Pubkey, signature::Keypair};
 use std::{
     boxed::Box,
     collections::HashSet,
@@ -135,6 +135,8 @@ impl Tvu {
         tvu_config: TvuConfig,
         max_slots: &Arc<MaxSlots>,
         cost_model: &Arc<RwLock<CostModel>>,
+        accounts_package_channel: (AccountsPackageSender, AccountsPackageReceiver),
+        last_full_snapshot_slot: Option<Slot>,
     ) -> Self {
         let Sockets {
             repair: repair_socket,
@@ -212,9 +214,9 @@ impl Tvu {
                 (Some(snapshot_config), Some(pending_snapshot_package))
             })
             .unwrap_or((None, None));
-        let (accounts_hash_sender, accounts_hash_receiver) = channel();
+        let (accounts_package_sender, accounts_package_receiver) = accounts_package_channel;
         let accounts_hash_verifier = AccountsHashVerifier::new(
-            accounts_hash_receiver,
+            accounts_package_receiver,
             pending_snapshot_package,
             exit,
             cluster_info,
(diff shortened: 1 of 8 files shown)


ledger/src/blockstore_processor.rs
```diff
@@ -1161,11 +1186,35 @@ fn load_frozen_forks(
...
+                        )
+                        .expect("Failed to snapshot bank while loading frozen banks");
+                        trace!(
+                            "took bank snapshot for new root bank, block height: {}, slot: {}",
+                            block_height,
+                            *root
+                        );
+                        *last_full_snapshot_slot = Some(*root);
+                    }
+                }
```
brooksprumo, line 1211: This is where snapshots are taken while processing blocks that are roots, and also sets the new value for `last_full_snapshot_slot`.

ledger-tool/src/main.rs
```diff
@@ -712,7 +712,7 @@ fn load_bank_forks(
         let snapshot_archives_dir =
             snapshot_archive_path.unwrap_or_else(|| blockstore.ledger_path().to_path_buf());
         Some(SnapshotConfig {
-            full_snapshot_archive_interval_slots: 0, // Value doesn't matter
+            full_snapshot_archive_interval_slots: Slot::MAX,
```
brooksprumo, line 715: Cannot be zero because this value is used when processing the blockstore and is the modulo of the new roots. A value of 0 here is a divide-by-zero error.

core/src/tvu.rs
```diff
@@ -135,6 +135,8 @@ impl Tvu {
         tvu_config: TvuConfig,
         max_slots: &Arc<MaxSlots>,
         cost_model: &Arc<RwLock<CostModel>>,
+        accounts_package_channel: (AccountsPackageSender, AccountsPackageReceiver),
+        last_full_snapshot_slot: Option<Slot>,
```
brooksprumo, line 139: The TVU now takes in an AccountsPackageSender/Receiver, and the last full snapshot slot. This comes from bank_forks::load() and blockstore_processor::do_process_blockstore_from_root().

brooksprumo (commented):
I'm calling out a few spots to highlight so they don't get lost in some of the plumbing noise.

ledger/src/blockstore_processor.rs
brooksprumo: TODO: call clean before snapshot_bank

carllin (approved):

PR #19602, opened 2021-09-03
Removed the --authorized-withdrawer argument to the cli create-vote-a…

bji:
…ccount

and made this parameter a required third argument.  This is because
authorized withdrawer should never be the same as vote account keypair or
validator identity keypair for security reasons.  Added a
--allow-unsafe-authorized-withdrawer to override this restriction if
necessary.

#### Summary of Changes

Fixes #


mvines:
Thanks @bji, do you want us to take this over? There's a bunch of little things that we'll need to patch up to get this mergeable such as:
* `cargo fmt` is angry
* I bet a test or two will fail
* `src/running-validator/validator-start.md` references create-vote-account so will need an update

cli/src/vote.rs
```diff
@@ -70,10 +79,13 @@ impl VoteSubCommands for App<'_, '_> {
                         "Public key of the authorized voter [default: validator identity pubkey]. "),
                 )
                 .arg(
-                    pubkey!(Arg::with_name("authorized_withdrawer")
-                        .long("authorized-withdrawer")
-                        .value_name("WITHDRAWER_PUBKEY"),
-                        "Public key of the authorized withdrawer [default: validator identity pubkey]. "),
+                    Arg::with_name("allow_unsafe_authorized_withdrawer")
+                        .long("allow-unsafe-authorized-withdrawer")
+                        .value_name("ALLOW_UNSAFE_WITHDRAWER_PUBKEY")
```
mvines, line 84:
```suggestion
```

bji:
Well I did fix and run the unit tests, so I think they work.  I wanted to add a few to test the failure case but I could not figure out how to get the assert_eq! macro to handle asserting that the result was an error.  Seems there is some Rust PartialEq magic that I could not get to work right.

bji:
cargo fmt I can do myself, I never remember what that command is called so I did not run it

bji:
> Thanks @bji, do you want us to take this over? There's a bunch of little things that we'll need to patch up to get this mergeable such as:
> 
>     * `cargo fmt` is angry

I can fix this.  Just run 'cargo fmt' and let it re-pretty the code, right?

> 
>     * I bet a test or two will fail

I already updated the unit tests in the original commit.  Would have liked to add some 'negative' tests to ensure that the command fails when a bad withdraw authority key is used but I couldn't figure out how to get assert_eq! to do what I wanted.

> 
>     * `src/running-validator/validator-start.md` references create-vote-account so will need an update

I already changed that, it was in the original commit.

mvines:
> > `src/running-validator/validator-start.md` references create-vote-account so will need an update

> I already changed that, it was in the original commit.

Ah, I was referring to this section: https://docs.solana.com/running-validator/validator-start#create-vote-account

===
As a part of this change, try removing the `Option` from here:
https://github.com/solana-labs/solana/blob/0b50bb2b2025b43c27bc755a95fbf114f2655d30/cli/src/cli.rs#L295
which'll ultimately allow for this `.unwrap_or()` to be removed:
https://github.com/solana-labs/solana/blob/0b50bb2b2025b43c27bc755a95fbf114f2655d30/cli/src/vote.rs#L566

bji:
OK I *think* I got all the i's dotted and t's crossed.  If you can do a quick check and if it looks like I'm just going to be more work than it's worth, then please feel free to take over.

cli/src/vote.rs
```diff
@@ -563,7 +593,7 @@ pub fn process_create_vote_account(
         let vote_init = VoteInit {
             node_pubkey: identity_pubkey,
             authorized_voter: authorized_voter.unwrap_or(identity_pubkey),
-            authorized_withdrawer: authorized_withdrawer.unwrap_or(identity_pubkey),
+            authorized_withdrawer: authorized_withdrawer,
```
mvines, line 596:
```suggestion
            authorized_withdrawer,
```

cli/tests/stake.rs
```diff
@@ -30,6 +30,7 @@ use solana_streamer::socket::SocketAddrSpace;
 fn test_stake_delegation_force() {
     let mint_keypair = Keypair::new();
     let mint_pubkey = mint_keypair.pubkey();
+    let withdraw_authority_pubkey = Keypair::new().pubkey();
```
mvines, line 33:
```suggestion
    let authorized_withdrawer = Keypair::new().pubkey();
```

cli/tests/stake.rs
```diff
@@ -53,7 +54,7 @@ fn test_stake_delegation_force() {
         seed: None,
         identity_account: 0,
         authorized_voter: None,
-        authorized_withdrawer: None,
+        authorized_withdrawer: withdraw_authority_pubkey,
```
mvines, line 57:
```suggestion
        authorized_withdrawer,
```

mvines (approved):
lgtm once CI is green

bji:
I changed "withdraw authority" to "authorized withdrawer" everywhere in code and documentation because having "withdraw authority" in some part of the docs and "authorized withdrawer" in others is partially what led to the confusion in variable names.  Trying to go for consistency here and it seems you prefer "authorized withdrawer".

bji:
(stake accounts still have a "withdraw authority", but vote accounts have an "authorized withdrawer")

mvines:
"withdraw authority" is the more modern term and is generally what I prefer in documentation and command-line help/arguments.

The code is littered with "authorized withdrawer" from an earlier time.  So until somebody takes a larger code cleanup effort, I try to keep the code consistent and use whatever the dominant term is for that particular file or function

bji:
So I changed a bunch of docs to the wrong one.  I thought "authorized 
withdrawer" is what you wanted ...

mvines:
Ugh now I remember going through this thought process before.  `solana stake-authorize` is perfect and uses the new names.  But for vote accounts the "authorized withdrawer" term is pretty well embedded in the vote-account-related command-line arguments, and given that vote accounts aren't used as commonly as stake accounts, it's maybe not worth the effort to try to clean this up.  So I think your latest docs changes are fine and do help improve overall consistency

mvines:
Looks like we also need an update at https://github.com/solana-labs/solana/blob/2a877ae06e11e8de5b0295e66b70e88cb9b5a860/multinode-demo/validator.sh#L303
The CI issue should be reproducible by running `ci/localnet-sanity.sh` locally

bji:
> Looks like we also need an update at
> 
> https://github.com/solana-labs/solana/blob/2a877ae06e11e8de5b0295e66b70e88cb9b5a860/multinode-demo/validator.sh#L303
> 
> The CI issue should be reproducible by running `ci/localnet-sanity.sh` locally

OK I did.  Not sure if you'll like what I did but it wasn't possible to make it as simple as the existing code because it's not an optional argument.

Also for completeness I grepped all files in the repo for "create-vote-account" (should have done that the first time, mea culpa) and it looks like there aren't any other instances to fix up.

mvines:
Pardon me for being rude, I pushed a commit here that simplifies the changes to `multinode-demo/validator.sh` a little.

bji (commented):
Yeah I don't really understand what you're doing here, if --authorized-withdrawer is specified as a pubkey you're going to write a file whose name is the pubkey of the authorized withdrawer, but I don't think it's really possible to shoe horn in the --authorized-withdrawer argument in a sane way so whatever you do that is fine with you, is fine with me!

bji (commented):
Actually I think you missed the setting of authorized_withdrawer_pubkey above on line 82, I think you want to change that to authorized_withdrawer?

mvines:
> Actually I think you missed the setting of authorized_withdrawer_pubkey above on line 82, I think you want to change that to authorized_withdrawer?

Yep, thanks and done!

mvines:
Thanks for the PR @bji!

bji:
> Thanks for the PR @bji!

You are welcome, thank you for your patience in working it into commit-ready condition.

PR #19611, opened 2021-09-03
Forwarding

sakridge:
#### Problem

Nodes can forward too many packets.

#### Summary of Changes

Add data budget for transaction forwarding

Fixes #


core/src/banking_stage.rs
```diff
@@ -315,11 +319,21 @@ impl BankingStage {
         socket: &std::net::UdpSocket,
         tpu_forwards: &std::net::SocketAddr,
         unprocessed_packets: &UnprocessedPackets,
+        data_budget: &DataBudget,
     ) -> std::io::Result<()> {
         let packets = Self::filter_valid_packets_for_forwarding(unprocessed_packets.iter());
         inc_new_counter_info!("banking_stage-forwarded_packets", packets.len());
+        const INTERVAL_MS: u64 = 100;
```
carllin, line 326: When layering the prioritization on top of this, do we want a global queue for prioritizing which packets to forward? Adding the prioritization here would at best sort packets with a particular banking thread's buffer.

KeKo6988 (approved):

core/src/banking_stage.rs
sakridge: Yea, I was thinking about that as well. A global list will do a better job of prioritization but it is a larger change with more risk.

mvines:
PR needs a rebase to drop 23d7064b8b43d549ef922661782b520c0742264e

jstarry:
I'm going to have this cleaned up by EOD fyi

PR #19614, opened 2021-09-03
Add RpcClient::get_multiple_accounts_with_config()

mvines:
The `dataSlice` input to [getMultipleAccounts](https://docs.solana.com/developing/clients/jsonrpc-api#getmultipleaccounts) isn't exposed in RpcClient

### DIFF
--- a/client/src/rpc_client.rs
+++ b/client/src/rpc_client.rs
@@ -2966,10 +2966,24 @@ impl RpcClient {
         pubkeys: &[Pubkey],
         commitment_config: CommitmentConfig,
     ) -> RpcResult<Vec<Option<Account>>> {
+        self.get_multiple_accounts_with_config(
+            pubkeys,
+            RpcAccountInfoConfig {
+                encoding: Some(UiAccountEncoding::Base64Zstd),
+                commitment: Some(self.maybe_map_commitment(commitment_config)?),
+                data_slice: None,
+            },
+        )
+    }
+
+    pub fn get_multiple_accounts_with_config(
+        &self,
+        pubkeys: &[Pubkey],
+        config: RpcAccountInfoConfig,
+    ) -> RpcResult<Vec<Option<Account>>> {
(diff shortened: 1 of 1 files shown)


client/src/rpc_client.rs
```diff
@@ -2966,11 +2966,22 @@ impl RpcClient {
...
+                commitment: Some(self.maybe_map_commitment(commitment_config)?),
+                data_slice: None,
+            },
+        )
+    }
+
+    pub fn get_multiple_accounts_with_config(
+        &self,
+        pubkeys: &[Pubkey],
+        mut config: RpcAccountInfoConfig,
```
CriesofCarrots, line 2982: Why is this `mut`, incidentally?

mvines: Because of this: 
```rust
config.commitment = config.commitment.or_else(|| Some(self.commitment()));
```

Default commitment can't be `None` ("finalized") and the more verbose version bugged me for no good reason:
```rust
let config = RpcAccountInfoConfig {
  commitment: config.commitment.or_else(|| Some(self.commitment())),
  ..config
};
```

CriesofCarrots: Oh thanks! I read right by that.

mvines: but wdyt?  👢 the `mut`?

CriesofCarrots: I don't mind the `mut` in principle, but it caught my attention because it was a different pattern than anything else in the crate. I'd either remove it, or update `get_program_accounts_with_config` to the same pattern to match?

(aside: maybe it's time to remove `maybe_map_commitment`? The public clusters have been supporting the new commitment variants for quite a while.)

mvines: >  it caught my attention because it was a different pattern than anything else in the crate.

Yep fair enough.  👢 ed

PR #19615, opened 2021-09-03
Make purge_old_snapshot_archives aware of maximum_incremental_snapshot_archives_to_retain

### LINKED ISSUE #18639: [Incremental Snapshots] Add field for maximum number of incremental snapshots to retain
I plan to address this TODO in a subsequent PR when adding integration tests for incremental snapshots.

_Originally posted by @brooksprumo in https://github.com/solana-labs/solana/pull/18504#discussion_r666414513_

brooksprumo:
#### Problem

An unlimited (not in reality though) number of incremental snapshots could be retained. We already have a maximum number of full snapshot archives to retain, but not one for incremental snapshots.

#### Summary of Changes

Update `purge_old_snapshot_archives()` to now take in a max for incremental snapshot archives, so they are bounded. Also updated tests.

Fixes #18639

NOTE: There will be a subsequent PR that will use the `maximum_incremental_snapshot_archives_to_retain` from SnapshotConfig, instead of all the default values currently in place.

### DIFF
--- a/runtime/src/snapshot_utils.rs
+++ b/runtime/src/snapshot_utils.rs
@@ -372,7 +372,11 @@ pub fn archive_snapshot_package(
     fs::rename(&archive_path, &snapshot_package.path())
         .map_err(|e| SnapshotError::IoWithSource(e, "archive path rename"))?;
 
-    purge_old_snapshot_archives(tar_dir, maximum_snapshot_archives_to_retain);
+    purge_old_snapshot_archives(
+        tar_dir,
+        maximum_snapshot_archives_to_retain,
+        DEFAULT_MAX_INCREMENTAL_SNAPSHOT_ARCHIVES_TO_RETAIN,
+    );
 
     timer.stop();
     info!(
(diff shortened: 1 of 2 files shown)


runtime/src/snapshot_utils.rs
```diff
@@ -52,6 +52,7 @@ const DEFAULT_SNAPSHOT_VERSION: SnapshotVersion = SnapshotVersion::V1_2_0;
 pub(crate) const TMP_BANK_SNAPSHOT_PREFIX: &str = "tmp-bank-snapshot-";
 pub const TMP_SNAPSHOT_ARCHIVE_PREFIX: &str = "tmp-snapshot-archive-";
 pub const DEFAULT_MAX_FULL_SNAPSHOT_ARCHIVES_TO_RETAIN: usize = 2;
+pub const DEFAULT_MAX_INCREMENTAL_SNAPSHOT_ARCHIVES_TO_RETAIN: usize = 100;
```
mvines, line 55: 100 seems like a lot! Why not like, say 2?

brooksprumo: Yeah, 100 is probably too many. I was thinking that the incremental snapshots would be very small. How about a few more, like 4?

mvines: sgtm. I don't have a major objection to 100, it just seemed like a lot but 🤷🏼

mvines (dismissed):
looks good to me overall, thanks!

runtime/src/snapshot_utils.rs
brooksprumo: Got it. Thanks for the review!

PR #19616, opened 2021-09-03
Pass SnapshotConfig to SnapshotPackagerService

brooksprumo:
SnapshotPackagerService will need more values that are from SnapshotConfig for incremental snapshot support. Instead of just passing in each one-by-one everywhere, gimme the whole SnapshotConfig to make life easier.


jeffwashington (approved):
lgtm

PR #19618, opened 2021-09-03
accounts index iterator takes &Range

jeffwashington:
#### Problem
Follow on changes will allow callers to not have to clone bounds to call range functions.
This is a little messy, so easier to do now.
#### Summary of Changes

Fixes #


brooksprumo (approved):
Interesting! Today I learned that `Range` is not `Copy`.

PR #19624, opened 2021-09-03
Add few more metrics data points 

taozhu-chicago:
#### Problem
Need few more data points to help come up reasonable cost for Fee Schedule (#16984)
 
#### Summary of Changes
- add slot, count and accumulated-units to per-program-timings; This allows to run influxDB query on cluster to get average `compute_unit to micro_sec` conversion ratio, which is used to convert execute_time from micro-sec to compute-unit. 

- bump sigverify stats from debug to info

Fixes #


core/src/sigverify_stage.rs
```diff
@@ -97,11 +97,12 @@ impl SigVerifyStage {
             (len as f32 / verify_batch_time.as_s())
         );
 
-        datapoint_debug!(
+        datapoint_info!(
```
sakridge, line 100: I think we need to accumulate these metrics, otherwise this will submit it too many times. That's why we have it as _debug by default.

core/src/sigverify_stage.rs
```diff
@@ -97,11 +97,12 @@ impl SigVerifyStage {
...
             "sigverify_stage-total_verify_time",
             ("num_batches", batch_len, i64),
             ("num_packets", len, i64),
             ("verify_time_ms", verify_batch_time.as_ms(), i64),
+            ("verify_time_us", verify_batch_time.as_us(), i64),
```
sakridge, line 105: let's just replace the `ms` one.

core/src/sigverify_stage.rs
taozhu-chicago: yeah, noticed a lot more loggings at info level. Looks like @jbiseda already has an PR to accumulate stats #19674. Should be able to merge with it

core/src/sigverify_stage.rs
taozhu-chicago: sounds good, thanks

core/src/sigverify_stage.rs
taozhu-chicago: restored this change in fav of Jeff's PR

core/src/sigverify_stage.rs
taozhu-chicago: restore this change in fav of Jeff's PR

core/src/progress_map.rs
```diff
@@ -124,18 +124,29 @@ impl ReplaySlotStats {
             .iter()
             .collect();
         per_pubkey_timings.sort_by(|a, b| b.1.accumulated_us.cmp(&a.1.accumulated_us));
-        let total: u64 = per_pubkey_timings.iter().map(|a| a.1.accumulated_us).sum();
+        let total_us: u64 = per_pubkey_timings.iter().map(|a| a.1.accumulated_us).sum();
+        let total_units: u64 = per_pubkey_timings
+            .iter()
+            .map(|a| a.1.accumulated_units)
+            .sum();
+        let total_count: u32 = per_pubkey_timings.iter().map(|a| a.1.count).sum();
```
jbiseda, line 132: you could collapse these sum()s with something like:

let (total_us, total_units, total_count) = per_pubkey_timings.iter().fold((0,0,0), |(sum_us, sum_units, sum_count), a| (sum_us + a.1.accumulated_us, sum_units + a.1.accumulated_units, sum_count + a.1.count));

taozhu-chicago: neat! Thanks!

jbiseda (approved):

PR #19629, opened 2021-09-03
Return error if Transaction contains writable executable or ProgramData accounts

CriesofCarrots:
#### Problem
Accounts marked executable can only be written to if being upgraded (ie. if the UpgradeableLoader is present in the transaction). As of #19593, we're demoting top-level program ids incorrectly marked as writable, but we can't detect other executables until the accounts are loaded.
Similarly, ProgramData accounts can only be written if the UpgradeableLoader is present in the transaction, but we can't identify these accounts until they are loaded.

#### Summary of Changes
Return `InvalidAccountIndex` error to fail these types of Transactions and release the write lock quickly. Open to creating a new error type, if this seems significant enough.

Needs rebase on #19593 
@jackcmay  @jstarry , the last 3 commits are ready for review

### DIFF
--- a/runtime/src/accounts.rs
+++ b/runtime/src/accounts.rs
@@ -276,16 +276,15 @@ impl Accounts {
                             })
                             .unwrap_or_default();
 
-                        if account.executable()
-                            && demote_program_write_locks
-                            && message.is_writable(i, demote_program_write_locks)
-                            && !is_upgradeable_loader_present
-                        {
-                            error_counters.invalid_writable_account += 1;
-                            return Err(TransactionError::InvalidWritableAccount);
-                        }
-
                         if bpf_loader_upgradeable::check_id(account.owner()) {
+                            if demote_program_write_locks
+                                && message.is_writable(i, demote_program_write_locks)
+                                && !is_upgradeable_loader_present
+                            {
+                                error_counters.invalid_writable_account += 1;
+                                return Err(TransactionError::InvalidWritableAccount);
+                            }
+
                             if account.executable() {
                                 // The upgradeable loader requires the derived ProgramData account
                                 if let Ok(UpgradeableLoaderState::Program {
@@ -306,19 +305,13 @@ impl Accounts {
                                     error_counters.invalid_program_for_execution += 1;
                                     return Err(TransactionError::InvalidProgramForExecution);
                                 }
-                            } else if demote_program_write_locks
-                                && message.is_writable(i, demote_program_write_locks)
-                                && !is_upgradeable_loader_present
-                            {
-                                // If a ProgramData account is present, it should only be writable
-                                // if the bpf_loader_upgradeable account is also present
-                                if let Ok(UpgradeableLoaderState::ProgramData { .. }) =
-                                    account.state()
-                                {
-                                    error_counters.invalid_writable_account += 1;
-                                    return Err(TransactionError::InvalidWritableAccount);
-                                }
                             }
+                        } else if account.executable()
+                            && demote_program_write_locks
+                            && message.is_writable(i, demote_program_write_locks)
+                        {
+                            error_counters.invalid_writable_account += 1;
+                            return Err(TransactionError::InvalidWritableAccount);
                         }
 
                         tx_rent += rent;
@@ -1751,32 +1744,9 @@ mod tests {
             (Err(TransactionError::InvalidWritableAccount), None)
         );
 
-        // Solution 1: include bpf_loader_upgradeable account
-        let mut account = AccountSharedData::new(40, 1, &native_loader::id()); // create mock bpf_loader_upgradeable
-        account.set_executable(true);
-        account.set_rent_epoch(1);
-        let accounts_with_upgradeable_loader = vec![
-            accounts[0].clone(),
-            accounts[1].clone(),
-            (bpf_loader_upgradeable::id(), account),
-        ];
-        message.account_keys = vec![key0, key1, bpf_loader_upgradeable::id()];
-        let tx = Transaction::new(&[&keypair], message.clone(), Hash::default());
-        let loaded_accounts =
-            load_accounts(tx, &accounts_with_upgradeable_loader, &mut error_counters);
-
-        assert_eq!(error_counters.invalid_writable_account, 1);
-        assert_eq!(loaded_accounts.len(), 1);
-        let result = loaded_accounts[0].0.as_ref().unwrap();
-        assert_eq!(result.accounts[..2], accounts_with_upgradeable_loader[..2]);
-        assert_eq!(
-            result.loaders[0],
-            vec![accounts_with_upgradeable_loader[2].clone()]
-        );
-
-        // Solution 2: mark programdata as readonly
+        // Mark executables as readonly
         message.account_keys = vec![key0, key1, key2]; // revert key change
-        message.header.num_readonly_unsigned_accounts = 2; // ark both executables as readonly
+        message.header.num_readonly_unsigned_accounts = 2; // mark both executables as readonly
         let tx = Transaction::new(&[&keypair], message, Hash::default());
         let loaded_accounts = load_accounts(tx, &accounts, &mut error_counters);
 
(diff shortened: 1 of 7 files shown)


jackcmay:
Last 3 commits look good so far!

runtime/src/accounts.rs
```diff
@@ -278,6 +278,14 @@ impl Accounts {
                             })
                             .unwrap_or_default();
 
+                        if account.executable()
+                            && demote_program_write_locks
+                            && message.is_writable(i, demote_program_write_locks)
+                        {
+                            error_counters.invalid_account_index += 1;
+                            return Err(TransactionError::InvalidAccountIndex);
+                        }
```
jstarry: Does this mean we can't write lock upgradeable programs during upgrades? I think we need to add a check to see if the upgradeable loader is present

runtime/src/accounts.rs
```diff
@@ -252,6 +252,7 @@ impl Accounts {
 
                     if solana_sdk::sysvar::instructions::check_id(key) {
                         if message.is_writable(i, demote_program_write_locks) {
+                            error_counters.invalid_account_index += 1;
```
jstarry, line 255: the message writable check will always return false for the instructions sysvar so this will never be hit anyways fyi

runtime/src/accounts.rs
CriesofCarrots: Oh duh, I even said that myself in the slack thread. Thanks for catching me!

CriesofCarrots: Fixed in updated https://github.com/solana-labs/solana/pull/19629/commits/08eed57dd468c7f9d8ea6058e09fe58729652234

runtime/src/accounts.rs
CriesofCarrots: True, I guess we should remove that check entirely?

jstarry: Yeah, I'm in favour

CriesofCarrots: ✅

KeKo6988 (approved):

CriesofCarrots:
Ready for final review!

runtime/src/accounts.rs
```diff
@@ -278,26 +275,49 @@ impl Accounts {
                             })
                             .unwrap_or_default();
 
-                        if account.executable() && bpf_loader_upgradeable::check_id(account.owner())
+                        if account.executable()
+                            && demote_program_write_locks
+                            && message.is_writable(i, demote_program_write_locks)
+                            && !is_upgradeable_loader_present(message)
```
sakridge, line 281: can we hoist `is_upgradeable_loader_present` outside this loop?

CriesofCarrots: Yes, ofc, thanks!

runtime/src/accounts.rs
```diff
@@ -278,26 +275,49 @@ impl Accounts {
...
-                            if let Ok(UpgradeableLoaderState::Program {
-                                programdata_address,
-                            }) = account.state()
-                            {
-                                if let Some(account) = self
-                                    .accounts_db
-                                    .load_with_fixed_root(ancestors, &programdata_address)
-                                    .map(|(account, _)| account)
+                            error_counters.invalid_account_index += 1;
+                            return Err(TransactionError::InvalidAccountIndex);
```
jackcmay, line 284: Would be helpful to return a new error here that better represents the failure

runtime/src/accounts.rs
```diff
@@ -278,26 +275,49 @@ impl Accounts {
...
+                                    }
                                 } else {
-                                    error_counters.account_not_found += 1;
-                                    return Err(TransactionError::ProgramAccountNotFound);
+                                    error_counters.invalid_program_for_execution += 1;
+                                    return Err(TransactionError::InvalidProgramForExecution);
+                                }
+                            }
+
+                            // If a ProgramData account is present, it should only be writable
```
jackcmay, line 310: Could this move into the success case of `load_with_fixed_root` above?

runtime/src/accounts.rs
CriesofCarrots: Yeah, I went back and forth on that, and ultimately went for less churn. But I'll add a unique error and error_counter

runtime/src/accounts.rs
jackcmay: Hmm, this is the 2nd case so no, but it could probably be inside an else so the account is not deserialized twice?

CriesofCarrots: Oh good, I see your 2nd note now. I was stuck trying to figure out how it would work in `load_with_fixed_root`. Yes, great point, I'll rewrite to deser once.

runtime/src/accounts.rs
```diff
@@ -278,26 +276,50 @@ impl Accounts {
...
-                                    return Err(TransactionError::ProgramAccountNotFound);
+                                    error_counters.invalid_program_for_execution += 1;
+                                    return Err(TransactionError::InvalidProgramForExecution);
                                 }
                             } else {
-                                error_counters.invalid_program_for_execution += 1;
-                                return Err(TransactionError::InvalidProgramForExecution);
+                                // If a ProgramData account is present, it should only be writable
+                                // if the bpf_loader_upgradeable account is also present
+                                if let Ok(UpgradeableLoaderState::ProgramData { .. }) =
```
jackcmay, line 312: Would it be more efficient to first check the books and then deserialize?

runtime/src/accounts.rs
```diff
@@ -251,9 +252,6 @@ impl Accounts {
                     }
 
                     if solana_sdk::sysvar::instructions::check_id(key) {
-                        if message.is_writable(i, demote_program_write_locks) {
```
jackcmay, line 254: Why was this removed?

CriesofCarrots: @jstarry pointed out that, because sysvar write locks are being demoted, this will never return true `if solana_sdk::sysvar::instructions::check_id(key)`.

CriesofCarrots: So just tidying up.

runtime/src/accounts.rs
```diff
@@ -278,26 +276,48 @@ impl Accounts {
                             })
                             .unwrap_or_default();
 
-                        if account.executable() && bpf_loader_upgradeable::check_id(account.owner())
```
jstarry, line 281: Looks like the owner check was removed in order to apply this write restriction to all executable accounts. I think it would be better to distinguish between non-upgradeable programs and upgradeable programs though to make it impossible to write lock the token program.

runtime/src/accounts.rs
jstarry: Could we skip the deserialization altogether and disallow writes to any account owned by the upgradeable loader? The only time a program other than the upgradeable loader can write to one of these accounts is if they top up the lamports which doesn't seem useful anyways.

runtime/src/accounts.rs
CriesofCarrots: Oh, only include the `is_upgradeable_loader_present` exception if the owner is the upgradeable loader? Yeah, makes sense

jackcmay:
Adding more lamports might be necessary when we add support for realloc...

jstarry:
True, do you think the bpf loader program could be responsible for that?

jackcmay:
I think so, the current plan is continue funding the account via the buffer account, so realloc to larger ProgramData account would require a larger buffer account with a larger associated number of tokens

CriesofCarrots:
I'm not clear, @jackcmay , are you saying the upgradeable-loader program need to be present to do that sort of realloc?
If so, then I think we can skip the deserialization, because any type of `UpgradeableLoaderState` woudl require the program be present for writes.

jackcmay:
In order for a program to be realloc'd larger the ProgramData account's balance will have to increase in order to still be rent-exempt.  The question was who/how transfers those lamports.  The current plan is for the larger buffer account to ferry along those lamports which means the upgradeable loader will always exist for all program reallocs.  aka, we can probably skip deserialization l like Justin suggested.

CriesofCarrots:
Okay, the last commit changes the logic to return an error if:
- A writable account is owned by the upgradeable loader, but the upgradeable loader is not included in the tx
- OR, a writable account is executable

Look sufficient?

PR #19638, opened 2021-09-04
Update to new Telegram link at docs

pierreneter:
#### Problem

There is an outdated telegram link of Solana at docs.

#### Summary of Changes

Update to new Telegram URL.


mvines:
Thanks!

PR #19612, opened 2021-09-03
Add maximum_incremental_snapshot_archives_to_retain to SnapshotConfig

brooksprumo:
Adding `maximum_incremental_snapshot_archives_to_retain` into `SnapshotConfig` has a good bit of plumbing noise, so pulled it out into its own PR.

The `maximum_incremental_snapshot_archives_to_retain` will be used by many functions in snapshot_utils.

PR #19640, opened 2021-09-04
Plumb `maximum_incremental_snapshot_archives_to_retain`

brooksprumo:
#### Problem

`purge_old_snapshot_archives()` could accept a `maximum_incremental_snapshot_archives_to_retain` parameter, but no functions could set that parameter with anything other than the default, since it was not plumbed through.

#### Summary of Changes

Plumb through `maximum_incremental_snapshot_archives_to_retain`

This is the final PR for this work. See also:

PR #19612 
PR #19615
PR #19616 

And the main issue: #18639


steviez (approved):
LGTM!

Issue #18629, opened 2021-07-13
Native CPI caller write privileges is incorrect


jstarry:
#### Problem
Caller write privileges are not created correctly for native cpi. The `caller_write_privileges `vector should match up with `message.account_keys` because they will be indexed into in the same way

#### Proposed Solution
- Remove incorrect write privilege value insertion
- Match up write privileges vector with the accounts vector built from `message.account_keys`

fix with no tests is in this branch: https://github.com/jstarry/solana/tree/fix-caller-write-privileges

jstarry:
Since native_invoke is only called for the bpf upgradeable program and that program doesn't modify caller accounts before invoking the native create account instruction, there shouldn't be any change of behavior before and after the fix.

jstarry:
@jackcmay I'm pretty sure https://github.com/solana-labs/solana/pull/19645 will be blocked on this issue FYI

jackcmay:
What specifically do you think is blocking?

jstarry:
`caller_write_privileges` is constructed in order of `keyed_account_indices` instead of `message.account_keys` which contrasts with how `accounts` and `keyed_account_indices_reordered` are constructed. This means that if `keyed_account_indices` is not monotonically increasing, the write privileges can be incorrect. This is problematic because if a native program modifies any accounts before invoking another program, it's possible that the native program could be allowed to write to a read-only account that it owns. Or, it may be prevented from writing to a writable account that it owns.

#19645 introduces a change which modifies a buffer account before the native invoke which means that a malicious user could trick the bpf loader program to successfully deplete the balance of a readonly buffer account.

Issue #9711, opened 2020-04-24
Cross-program balanced instruction handling is too restrictive


jackcmay:
#### Problem

The `MessageProcessor`'s balanced instruction check ensures the Lamport sum of all accounts in the instruction remains the same before and after the instruction.  Cross-program invocations only include the accounts required by the invoked instruction and thus leaves out any transfers the invoking program may have made to other accounts before the call to invoke.

For example:

For programs `A` and `B` and accounts `X`, `Y`, `Z`

- A transfers lamports from X to Y
- A invokes B passing Y, Z
- balanced instruction check fails because it does not take into account the lamports in X

It requires developers to be aware of when they transfer tokens relative to any invocations they are making.

#### Proposed Solution

Include all accounts in the balance calculations

jackcmay:
@jstarry

mvines:
@jackcmay - do you want to keep this in 1.2 or punt to 1.3?

jackcmay:
@mvines I think given the other priorities punting to 1.3 is more realistic

leoluk:
Looks like this is breaking Wormhole guardian set upgrades, which do:

* A CPI call.
* Transfer lamports from the bridge's subsidizer pool.
* Another CPI call.

```
[devnet] [2021-02-01T19:09:41.671877046Z DEBUG solana_runtime::message_processor] Program Bridge1p5gheXUvJ6jGWGeCsgPKgnE3YgdGKRVCMY9o invoke [1]
[devnet] [2021-02-01T19:09:41.672078448Z DEBUG solana_runtime::message_processor] Program log: Instruction: PostVAA
[devnet] [2021-02-01T19:09:41.673453480Z DEBUG solana_runtime::message_processor] Program log: deriving key
[devnet] [2021-02-01T19:09:41.673568861Z DEBUG solana_runtime::message_processor] Program log: deploying contract
[devnet] [2021-02-01T19:09:41.673794033Z DEBUG solana_runtime::message_processor] Program 11111111111111111111111111111111 invoke [2]
[devnet] [2021-02-01T19:09:41.673816843Z TRACE solana_runtime::system_instruction_processor] process_instruction: CreateAccount { lamports: 3814080, space: 420, owner: Bridge1p5gheXUvJ6jGWGeCsgPKgnE3YgdGKRVCMY9o }
[...]

PR #19645, opened 2021-09-04
Reduce payer balance needed to deploy programs

jstarry:
#### Problem
Program deploys require more SOL than necessary for a successful deploy.

The payer balance is used to pay for programdata account creation before the payer is credited for the deploy buffer account. The buffer account balance is about the same as the required programdata balance, so a payer basically has to cover twice the rent cost than they need to.

#### Summary of Changes
- Credit buffer balance before paying for programdata account creation

Fixes #

### DIFF
--- a/programs/bpf_loader/src/lib.rs
+++ b/programs/bpf_loader/src/lib.rs
@@ -35,7 +35,7 @@ use solana_sdk::{
         stop_verify_mul64_imm_nonzero,
     },
     ic_logger_msg, ic_msg,
-    instruction::InstructionError,
+    instruction::{AccountMeta, InstructionError},
     keyed_account::{from_keyed_account, keyed_account_at_index, KeyedAccount},
     loader_instruction::LoaderInstruction,
     loader_upgradeable_instruction::UpgradeableLoaderInstruction,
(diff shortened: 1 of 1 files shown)


jstarry:
Blocked on https://github.com/solana-labs/solana/issues/18629

jackcmay:
I'm working on #18629

jackcmay:
> I'm working on #18629

https://github.com/solana-labs/solana/pull/19750

jackcmay:
Btw, if this was blocked I would expect a test that shows that :-)

programs/bpf_loader/src/lib.rs
```diff
@@ -390,13 +390,27 @@ fn process_loader_upgradeable_instruction(
...
+
+            let mut instruction = system_instruction::create_account(
                 payer.unsigned_key(),
                 programdata.unsigned_key(),
                 1.max(rent.minimum_balance(programdata_len)),
                 programdata_len as u64,
                 program_id,
             );
+
+            // pass an extra account to avoid the overly strict UnbalancedInstruction error
```
jstarry, line 409: FYI @jackcmay I'm bumping into this issue now: https://github.com/solana-labs/solana/issues/9711

jstarry:
This is ready for review @jackcmay, as mentioned above I hit the overly restrictive balanced instruction check but it's easy to work around for now

jackcmay:
> This is ready for review @jackcmay, as mentioned above I hit the overly restrictive balanced instruction check but it's easy to work around for now

We are expecting to solve the overly restrictive balanced instruction issue in ABIv2 (FYI @Lichtso )

jackcmay (dismissed):

PR #19654, opened 2021-09-05
Speed up program deploys by writing larger chunks

jstarry:
#### Problem
CLI program deploys use a fixed chunk size of 229 bytes, which is far lower than what can fit in the 1232 bytes allowed for serialized transactions. When using different signers for the tx fee payer and the upgrade authority, the chunk size can be as high as 916 bytes, 4 times larger than the current size, which should decrease the deploy time of programs by about 75%.

#### Summary of Changes
- Dynamically compute the max chunk size allowed per write instruction during deploys and upgrades. The dynamic computation is needed because the number of signatures and keys used in the transaction could change depending on whether the payer and upgrade authority are different, for example.

Fixes #

### DIFF
--- a/cli/src/program.rs
+++ b/cli/src/program.rs
@@ -41,9 +41,10 @@ use solana_sdk::{
     loader_instruction,
     message::Message,
     native_token::Sol,
+    packet::PACKET_DATA_SIZE,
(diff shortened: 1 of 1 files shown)


jackcmay (approved):
Very nice!

PR #19655, opened 2021-09-05
fixup! test-validator: start logging asap

steviez:
#### Problem
I think the logic for when we make a logfile is inverted (from https://github.com/solana-labs/solana/commit/ee65ffb3c8); the current behavior is:
`./solana-test-validator` ==> logfile generated in `ledger_path` directory
`/solana-test-validator --log` ==> No logfile + output to console

### DIFF
--- a/validator/src/bin/solana-test-validator.rs
+++ b/validator/src/bin/solana-test-validator.rs
@@ -323,7 +323,7 @@ fn main() {
 
     let validator_log_symlink = ledger_path.join("validator.log");
 
-    let logfile = if output != Output::Log {
+    let logfile = if output == Output::Log {
         let validator_log_with_timestamp = format!(
             "validator-{}.log",
             SystemTime::now()


brooksprumo (approved):
lgtm

jstarry:
The `--quiet` and `--log` cli flags were/are really unintuitive. Here's the break down by `Output` where `--quiet` is `Output::None`, `--log` is `Output::Log`, and the default behavior is `Output::Dashboard`.

Previously:
`Output::None`: Hid dashboard and redirected logs to a file
`Output::Log`: Hid dashboard and displayed logs in terminal
`Output::Dashboard`: Showed dashboard and redirected logs to a file

Now (after this change):
`Output::None`: Hides dashboard and displays logs in terminal
`Output::Log`: Redirects logs to a file
`Output::Dashboard`: Interleaves dashboard and logs in terminal

I think we should copy over the `--logfile` cli arg behavior from `solana-validator` and remove the `--log` cli arg altogether. And I think the `--quiet` arg should redirect logs to `/dev/null`.

jstarry:
In the meantime, I'm going to revert this change so that `--quiet` is actually quiet again

PR #19656, opened 2021-09-06
Blockstore test cleanup

steviez:
#### Summary of Changes
This PR accomplishes several things; functionality is not changed on any of the tests in a meaningful way:
- Make it so that all the tests in ledger that create a blockstore cleanup the full ledger directory (blockstore + genesis + actual directory) on success
   - This reclaims disk space immediately and reduces clutter which helps if inspecting the ledger on disk of a failed test.
   - Done by using a TempDir for the ledger (cleans itself up when it goes out of scope)
- Removed some Arc's around blockstore that weren't actually needed
- Make naming consistent across tests to use `blockstore` for object and `ledger_path` for the directory that contains blockstore
- Avoid the following pattern of using braces to implicitly drop the blockstore before removing the files on disk
```
let blockstore = Blockstore::open(...)
{
   ...
}
Blockstore::destroy(...)
```
   - Instead, move the blockstore object into the common function and explicitly drop it in there.
   - Unfortunately, making this change resulted in a bloated diff for all the lines that got a tab chopped off since they were no longer in an extra set of brackets.

A lot of the diff is from clippy whitespace cleanup from removing the braces; definitely toggle the diff to ignore whitespace when viewing.

### DIFF
--- a/ledger/src/blockstore.rs
+++ b/ledger/src/blockstore.rs
@@ -3948,6 +3948,17 @@ pub fn make_chaining_slot_entries(
     slots_shreds_and_entries
 }
 
+// used for tests only
+// Any access to the DB outside of blockstore (ie cf handles) should be dropped prior to calling this
+pub fn destroy_test_ledger(ledger_path: &Path, blockstore: Blockstore) {
+    // Destroying the blockstore without closing it first is undefined behavior; any
+    // active
+    drop(blockstore);
+    Blockstore::destroy(&ledger_path).expect("Expected successful blockstore destruction");
+    // Destroy the ledger directory, which may contain files such as genesis.bin
+    fs::remove_dir_all(&ledger_path).expect("Expected successful ledger destruction");
+}
+
 #[cfg(not(unix))]
 fn adjust_ulimit_nofile(_enforce_ulimit_nofile: bool) -> Result<()> {
     Ok(())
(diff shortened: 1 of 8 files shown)


ledger/src/blockstore.rs
```diff
@@ -3871,9 +3889,23 @@ pub fn create_new_ledger_from_name(
     genesis_config: &GenesisConfig,
     access_type: AccessType,
 ) -> (PathBuf, Hash) {
-    let ledger_path = get_ledger_path_from_name(name);
+    let (ledger_path, blockhash) =
+        create_new_ledger_from_name_auto_delete(name, genesis_config, access_type);
+    (ledger_path.into_path(), blockhash)
```
steviez, line 3894: Doing `ledger_path.into_path()` converts it from a TempDir to a regular PathBuf; this allows using the same function and providing an auto-delete and "regular" version

ledger/src/blockstore.rs
```diff
@@ -3809,9 +3809,16 @@ macro_rules! get_tmp_ledger_path_auto_delete {
...
 pub fn get_ledger_path_from_name_auto_delete(name: &str) -> TempDir {
-    let path = get_ledger_path_from_name(name);
+    let mut path = get_ledger_path_from_name(name);
+    // path is a directory so .file_name() returns the last component of the path
+    let last = path.file_name().unwrap().to_str().unwrap().to_string();
+    path.pop();
     fs::create_dir_all(&path).unwrap();
-    TempDir::new_in(path).unwrap()
+    Builder::new()
+        .prefix(&last)
```
steviez, line 3818: Each test makes a new folder that has filename & line# in it as identifying information. We would then make a temporary directory inside of there; this would get cleaned up but we'd still have the outer directory.

Popping the last element off the path and using it as prefix instead makes the directory with filename & line# the TempDir itself, which means this extra directory also gets cleaned up.

carllin (approved):

PR #19670, opened 2021-09-06
Rename to is_valid instead of is_invalid

brooksprumo:
I'll be expanding snapshot configuration, and wanted to start with renaming this function to `is_snapshot_config_valid()` instead of _invalid_.


CriesofCarrots (approved):
Yay, my head is much happier

PR #19671, opened 2021-09-07
Add some common subcommands and arguments to all solana CLIs (such as shell completion)

theonekeyg:
#### Problem
Add some common subcommands and arguments to all solana CLIs
(such as shell completion).

This problem for sure has multiple solutions, this is my take on it.

I introduced a trait that implements itself on clap App, adding method
for adding common commands (`finalize`) and processing added commands
(`process_common`).

All the end user has to do is import the trait in their program, call
`finalize` at the end of creation of clap App, and call `process_common`
later in the program whenever he wants to process common commands.

Personally, i just want the shell completion functionality, but this could be
extended to add some other common subcommands and arguments.
I also think this will add some generality to the implementations of CLI utilities,
as right now it's a mess.

I implemented this new design for solana-keygen CLI as a reference to show
how little change is needed in order to setup new API.

### DIFF
--- a/clap-utils/src/app.rs
+++ b/clap-utils/src/app.rs
@@ -1,0 +1,49 @@
+use clap::{App, Arg, Shell, SubCommand};
+use std::io::stdout;
+
+/// Trait that adds common solana arguments and subcommands to clap App.
+pub trait SolanaApp<'ab, 'v> {
+    /// Add common arguments and subcommands to provided App.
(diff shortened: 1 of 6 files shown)


theonekeyg:
I think I was misunderstood, I didn't mean to volunteer to implement this new design for every CLI in solana repo (I guess I could do that as well, but I need to know that this PR is something yall interested in too first).

PR #19681, opened 2021-09-07
Add note about older rust versions on release branches

steviez:
#### Problem
The README states to use the latest stable version of rust; this may not be correct when building a release branch.

#### Summary of Changes
Add note to check rust version when building a release branch;.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,11 +19,17 @@ $ source $HOME/.cargo/env
 $ rustup component add rustfmt
 ```
 
-Please make sure you are always using the latest stable rust version by running:
+When building the master branch, please make sure you are using the latest stable rust version by running:
 
 ```bash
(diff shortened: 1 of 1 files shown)


CriesofCarrots (approved):
Thank you!

mvines:
I think we should document and encourage the use of `./cargo`.  It does the right thing for master and release branches

steviez:
> I think we should document and encourage the use of `./cargo`. It does the right thing for master and release branches

I like that, I'll tweak it again next time I'm waiting for on tests. I think we should keep the note about overrides tho; the script doesn't (and I'm not sure that it should) adjust the default toolchain, so running `cargo build` would use the wrong version if someone has a newer rust version on their machine.

PR #19685, opened 2021-09-07
precompiles are not verified when called via cpi

seanyoung:
#### Problem

Calling a secp256k1/ed25519 verify via CPI does not actually do any signature check, it just returns success.

#### Summary of Changes

Fixes #


mvines:
Sean, do you strictly need precompiles in a CPI?  The fact that this doesn't work currently is an intentional design decision. We want to keep the door open for a a future smarter sigverify stage that collects and offloads batches of precompiles to the GPU or CPU vector processing even before accounts are locked and the runtime starts getting involved

seanyoung:
Right. The whole point of adding ed25519 signature verify program is so that we can verify ed25519 signatures from Solidity, i.e. from BPF. I'm not sure how else this can be achieved.

seanyoung:
Alternatively I can add a `sol_ed25511(_sig_verify()` syscall like I originally submitted.

What's the reason someone would want to do signature verification in a tx and pay for it? Wouldn't be easier/quicker to do this client side?

mvines:
> Right. The whole point of adding ed25519 signature verify program is so that we can verify ed25519 signatures from Solidity, i.e. from BPF. I'm not sure how else this can be achieved.

I was imagining it'd look like this:
The signature to be verified would be placed in a precompile ed25519 instruction by the user, and verified by the sigverify stage.  When the solidity code executes later and wants to "verify" the ed25519 signature, it uses the [Instructions](https://docs.solana.com/developing/runtime-facilities/sysvars#instructions) sysvar to search for a matching signature in the precompiles for the transaction it's running in.  If the signature is found then it's known to be valid because the transaction wasn't rejected.

seanyoung:
> > Right. The whole point of adding ed25519 signature verify program is so that we can verify ed25519 signatures from Solidity, i.e. from BPF. I'm not sure how else this can be achieved.
> 
> I was imagining it'd look like this:
> The signature to be verified would be placed in a precompile ed25519 instruction by the user, and verified by the sigverify stage. When the solidity code executes later and wants to "verify" the ed25519 signature, it uses the [Instructions](https://docs.solana.com/developing/runtime-facilities/sysvars#instructions) sysvar to search for a matching signature in the precompiles for the transaction it's running in. If the signature is found then it's known to be valid because the transaction wasn't rejected.

Aha! That is perfect. I had not thought about that. I'll do that instead.

PR #19687, opened 2021-09-07
AccountsIndexConfig -> AccountsDbConfig

jeffwashington:
#### Problem
There will be at least a few more parameters immediately. This is better than passing collections of parameters.
#### Summary of Changes
pass a hierarchy of accounts config
accounts db:
  accounts index
Fixes #


brooksprumo (dismissed):
lgtm

PR #19694, opened 2021-09-07
 Add CLI args for incremental snapshots

### LINKED ISSUE #19296: [Incremental Snapshots] Plumb CLI options to SnapshotConfig, Validator, ledger-tool, etc

brooksprumo:
#### Problem

Incremental snapshots are not enable-able, nor configurable.

#### Summary of Changes

Add a new cli arg, `incremental-snapshots`, that turns on incremental snapshots! Also, repurpose `snapshot-interval-slots` based on whether incremental snapshots are enabled or not. Lastly, added `full-snapshot-interval` too.

Also added `maximum-incremental-snapshots-to-retain` option.

Fixes #19296

### DIFF
--- a/core/src/validator.rs
+++ b/core/src/validator.rs
@@ -640,6 +640,7 @@ impl Validator {
             if let Some(snapshot_config) = config.snapshot_config.clone() {
                 if !is_snapshot_config_valid(
                     snapshot_config.full_snapshot_archive_interval_slots,
+                    snapshot_config.incremental_snapshot_archive_interval_slots,
                     config.accounts_hash_interval_slots,
                 ) {
                     error!("Snapshot config is invalid");
@@ -1661,12 +1662,26 @@ fn cleanup_accounts_path(account_path: &std::path::Path) {
 }
 
 pub fn is_snapshot_config_valid(
-    snapshot_interval_slots: u64,
-    accounts_hash_interval_slots: u64,
+    full_snapshot_interval_slots: Slot,
+    incremental_snapshot_interval_slots: Slot,
+    accounts_hash_interval_slots: Slot,
 ) -> bool {
-    snapshot_interval_slots == 0
-        || (snapshot_interval_slots >= accounts_hash_interval_slots
-            && snapshot_interval_slots % accounts_hash_interval_slots == 0)
+    // if full snapshot interval is MAX, that means snapshots are turned off, so yes, valid
+    if full_snapshot_interval_slots == Slot::MAX {
+        return true;
+    }
+
+    let is_incremental_config_valid = if incremental_snapshot_interval_slots == Slot::MAX {
+        true
+    } else {
+        incremental_snapshot_interval_slots >= accounts_hash_interval_slots
+            && incremental_snapshot_interval_slots % accounts_hash_interval_slots == 0
+            && full_snapshot_interval_slots > incremental_snapshot_interval_slots
+    };
+
+    full_snapshot_interval_slots >= accounts_hash_interval_slots
+        && full_snapshot_interval_slots % accounts_hash_interval_slots == 0
+        && is_incremental_config_valid
 }
 
 #[cfg(test)]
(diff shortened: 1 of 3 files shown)


validator/src/main.rs
```diff
@@ -2680,36 +2725,54 @@ pub fn main() {
...
+    let (full_snapshot_archive_interval_slots, incremental_snapshot_archive_interval_slots) =
+        if snapshot_interval_slots > 0 {
+            if is_incremental_snapshot_enabled {
+                (
+                    value_t_or_exit!(matches, "full_snapshot_interval_slots", u64),
+                    value_t_or_exit!(matches, "incremental_snapshot_interval_slots", u64),
+                )
+            } else {
+                (snapshot_interval_slots, Slot::MAX)
+            }
```
mvines, line 2741: It took me a little while to convince myself that there wasn't a bug in this logic. This makes things more readable for me at least:
```suggestion
    let (full_snapshot_archive_interval_slots, incremental_snapshot_archive_interval_slots) =
        let incremental_snapshot_interval_slots = value_t_or_exit!(matches, "incremental_snapshot_interval_slots", u64);      
        if incremental_snapshot_interval_slots > 0 {
            if matches.is_present("incremental_snapshots") {
                (
                    value_t_or_exit!(matches, "full_snapshot_interval_slots", u64),
                    incremental_snapshot_interval_slots,
                )
            } else {
                (incremental_snapshot_interval_slots, Slot::MAX)
            }
```

brooksprumo: Yep, that looks good to me. I had something like this at one point as well, so happy to go with the version that makes sense to more than just me.

core/src/validator.rs
```diff
@@ -1661,12 +1662,26 @@ fn cleanup_accounts_path(account_path: &std::path::Path) {
 }
 
 pub fn is_snapshot_config_valid(
```
mvines, line 1664: This makes me wish that:
* `accounts_hash_interval_slots` remains unchanged
* `incremental_snapshot_interval` becomes in interval of `accounts_hash_interval_slots`
* `full_snapshot_interval` becomes in interval of `incremental_snapshot_interval`
then `is_snapshot_config_valid()` could be deleted as the config is always valid.

mvines: That'd be a larger change though, if you want to punt on that I'd be fine with that

brooksprumo: Yeah, I like that. Is that something to change for here/now though?

If yes, there's somethings I think would need to be figured out first:

1. How would the cli arg names look? Would there be new args? What would happen to the old ones?
2. How would someone indicate "no snapshots"? This one I think is pretty easy with a new `--no-snapshots` flag.

Also, I think the one difference is that right now the full snapshot interval doesn't have to be a multiple of the incremental snapshot interval. It _could_ be, but the logic later doesn't require that.

mvines: > Is that something to change for here/now though?

Na doesn't need to be here. However if we do want to go with that scheme then we'd want to replace the new args that are being introduced in this PR, fortunately there's no backwards compatibility requirement until master forks off to a release branch.

> I think the one difference is that right now the full snapshot interval doesn't have to be a multiple of the incremental snapshot interval. It could be, but the logic later doesn't require that.

Ah, right.  `full_snapshot_interval only` needs to be an interval of `accounts_hash_interval_slots`

mvines: Up to you if you want to pick that nit though, the current approach works well enough IMO

brooksprumo: I wonder what's more intuitive for the user/validator. Maybe keeping the interval in slots is the way to go...

I was thinking of how the help text would display this info, and how default values would show up. Also, because hopefully everyone uses the same full snapshot interval. So then if the user changes the accounts hash interval, that would change the default value for a full snapshot interval multiplier to hit the same interval in slots...

Bleh, these args are too intertwined, I think, to have an obvious "best" answer. With that, my plan is to leave the args as-is (assuming that's still OK).

brooksprumo:
@mvines Do you want me to wait until `downstream-projects` is fixed so that I can have a fully passing CI? (Or also is there more you'd like to review/more changes you'd like me to make?)

mvines (approved):
lgtm, no need to block on that busted downstream CI job

PR #19710, opened 2021-09-08
add --accounts-index-path to be used by disk accounts index

jeffwashington:
#### Problem
Disk-based accounts index is stored on disk. This is an argument to allow a validator to specify 0-n disks/folders/drives/nvmes to be used for the backing store. With no argument, the result is <ledger>/accounts_index

This will be connected to the disk index.
#### Summary of Changes
--accounts-index-path /dev1/accounts_index --accounts-index-path /dev1/accounts_index
Fixes #

### DIFF
--- a/validator/src/main.rs
+++ b/validator/src/main.rs
@@ -2012,6 +2012,14 @@ pub fn main() {
                 .help("Number of bins to divide the accounts index into"),
         )
         .arg(
+            Arg::with_name("accounts_index_path")
+                .long("accounts-index-path")
+                .value_name("PATH")
+                .takes_value(true)
+                .multiple(true)
+                .help("Persistent accounts-index location"),
+        )
+        .arg(
             Arg::with_name("accounts_db_test_hash_calculation")
                 .long("accounts-db-test-hash-calculation")
                 .help("Enables testing of hash calculation using stores in \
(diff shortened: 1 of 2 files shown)


jeffwashington:
[new options](https://docs.google.com/document/d/19PppfksKQ9vVrO8RBLWne4Tqb-0j-ITQmbeVa5aClvE/edit?usp=sharing)

validator/src/main.rs
```diff
@@ -2006,6 +2006,14 @@ pub fn main() {
                 .takes_value(true)
                 .help("Number of bins to divide the accounts index into"),
         )
+        .arg(
+            Arg::with_name("accounts_index_path")
+                .long("accounts-index-path")
+                .value_name("PATH")
+                .takes_value(true)
+                .multiple(true)
+                .help("Persistent accounts-index location"),
```
brooksprumo, line 2015: Would it be valuable to include in the help text that this option may be specified multiple times? A few of the options that accept multiple say "May be specified multiple times.". And maybe what the default value is too?

jeffwashington: good point. I added them.

validator/src/main.rs
```diff
@@ -2013,6 +2013,16 @@ pub fn main() {
...
         )
+        .arg(
+            Arg::with_name("accounts_index_path")
+                .long("accounts-index-path")
+                .value_name("PATH")
+                .takes_value(true)
+                .multiple(true)
+                .help("Persistent accounts-index location. \
+                       May be specified multiple times. \
+                       [default: [ledger]/accounts_index"),
```
brooksprumo, line 2024: Should there be a `]` at the end of the default line? (same for ledger-tool too)

brooksprumo (dismissed):
lgtm

PR #19717, opened 2021-09-08
TpuClient now uses the processed slot to seed its leader schedule cache

mvines:
Using `getMaxShredInsertSlot` (or `getMaxRetransmitSlot`) to seed the leader schedule cache is no good because forked nodes can transmit their slots from well above the real cluster height, and that breaks the cache.  Instead start the cache from the current processed slot, which may be slightly behind the cluster but it's better than starting from thousands of slots ahead of it

### DIFF
--- a/client/src/tpu_client.rs
+++ b/client/src/tpu_client.rs
@@ -5,7 +5,9 @@ use crate::{
 };
 use bincode::serialize;
 use log::*;
-use solana_sdk::{clock::Slot, pubkey::Pubkey, transaction::Transaction};
+use solana_sdk::{
+    clock::Slot, commitment_config::CommitmentConfig, pubkey::Pubkey, transaction::Transaction,
+};
 use std::{
     collections::{HashMap, HashSet, VecDeque},
     net::{SocketAddr, UdpSocket},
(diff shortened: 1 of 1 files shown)


jstarry:
Could we instead tweak the `estimated_current_slot` method? It is supposed to handle the problem you're describing. I feel that using "processed" will be worse because rpc clients commonly fall behind the cluster and so "processed" is often behind the current leader

mvines:
> Could we instead tweak the `estimated_current_slot` method? It is supposed to handle the problem you're describing. I feel that using "processed" will be worse because rpc clients commonly fall behind the cluster and so "processed" is often behind the current leader

`estimated_current_slot` seemed to be doing the right thing actually, the problem is the bad `start_slot ` (from `get_max_shred_insert_slot()`) seeds the cache here:
https://github.com/solana-labs/solana/blob/7a226dc03adc340995eedf8916dd05e44e661a73/client/src/tpu_client.rs#L255
which causes `TpuClient:send_transaction()` to fail because `LeaderTpuCache`'s `first_slot` is thousands of slots ahead of what `estimated_current_slot` returns:
https://github.com/solana-labs/solana/blob/7a226dc03adc340995eedf8916dd05e44e661a73/client/src/tpu_client.rs#L117
and the client falls back to using slowmo `RPC` to deploy programs/etc.

So seeding `LeaderTpuCache` starting from the processed slot may produce a `first_slot` that's behind a little but `LeaderTpuCache::run()` will fetch more schedule if/when `estimated_current_slot` shoots head.  But we never fetch older  leader schedule so starting from a slot in the future is basically fatal for however many slots any forked staked node is ahead of the cluster.

jstarry:
Ah makes perfect sense, sorry I missed that you were just changing the seed slot, looks good!

PR #19728, opened 2021-09-08
metrics on rent partitions

jeffwashington:
#### Problem
rent collection costs something. measure # of partitions to guide optimization.
#### Summary of Changes

Fixes #


brooksprumo (approved):
lgtm

PR #19746, opened 2021-09-09
Add test_incremental_snapshot_download() to local-cluster

### LINKED ISSUE #19065: [Incremental Snapshots] Add local_cluster tests
Add test to local_cluster that both creates incremental snapshots, and then loads another node with the full+incremental snapshots.

brooksprumo:
#### Problem

There were no local-cluster tests that used incremental snapshots.

#### Summary of Changes

Add a local-cluster test for incremental snapshots.

Fixes #19065 

P.S. Also note there will be a follow-on PR that adds another local-cluster test for dealing with the taking-a-snapshot-during-load_frozen_forks scenario.

### DIFF
--- a/download-utils/src/lib.rs
+++ b/download-utils/src/lib.rs
@@ -2,7 +2,10 @@
 use console::Emoji;
 use indicatif::{ProgressBar, ProgressStyle};
 use log::*;
-use solana_runtime::{snapshot_utils, snapshot_utils::ArchiveFormat};
+use solana_runtime::{
+    snapshot_package::SnapshotType,
+    snapshot_utils::{self, ArchiveFormat},
+};
 use solana_sdk::{clock::Slot, genesis_config::DEFAULT_GENESIS_ARCHIVE, hash::Hash};
 use std::fs::{self, File};
 use std::io;
@@ -244,13 +247,16 @@ pub fn download_genesis_if_missing(
     }
 }
 
-pub fn download_snapshot<'a, 'b>(
+/// Download a snapshot archive from `rpc_addr`.  Use `snapshot_type` to specify downloading either
+/// a full snapshot or an incremental snapshot.
+pub fn download_snapshot_archive<'a, 'b>(
     rpc_addr: &SocketAddr,
     snapshot_archives_dir: &Path,
     desired_snapshot_hash: (Slot, Hash),
-    use_progress_bar: bool,
+    snapshot_type: SnapshotType,
     maximum_full_snapshot_archives_to_retain: usize,
     maximum_incremental_snapshot_archives_to_retain: usize,
+    use_progress_bar: bool,
     progress_notify_callback: &'a mut DownloadProgressCallbackOption<'b>,
 ) -> Result<(), String> {
     snapshot_utils::purge_old_snapshot_archives(
@@ -259,20 +265,31 @@ pub fn download_snapshot<'a, 'b>(
         maximum_incremental_snapshot_archives_to_retain,
     );
 
-    for compression in &[
+    for archive_format in [
         ArchiveFormat::TarZstd,
         ArchiveFormat::TarGzip,
         ArchiveFormat::TarBzip2,
         ArchiveFormat::Tar, // `solana-test-validator` creates uncompressed snapshots
     ] {
-        let desired_snapshot_package = snapshot_utils::build_full_snapshot_archive_path(
-            snapshot_archives_dir.to_path_buf(),
-            desired_snapshot_hash.0,
-            &desired_snapshot_hash.1,
-            *compression,
-        );
+        let destination_path = match snapshot_type {
+            SnapshotType::FullSnapshot => snapshot_utils::build_full_snapshot_archive_path(
+                snapshot_archives_dir.to_path_buf(),
+                desired_snapshot_hash.0,
+                &desired_snapshot_hash.1,
+                archive_format,
+            ),
+            SnapshotType::IncrementalSnapshot(base_slot) => {
+                snapshot_utils::build_incremental_snapshot_archive_path(
+                    snapshot_archives_dir.to_path_buf(),
+                    base_slot,
+                    desired_snapshot_hash.0,
+                    &desired_snapshot_hash.1,
+                    archive_format,
+                )
+            }
+        };
 
-        if desired_snapshot_package.is_file() {
+        if destination_path.is_file() {
             return Ok(());
         }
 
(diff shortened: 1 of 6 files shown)


local-cluster/tests/local_cluster.rs
```diff
@@ -3515,22 +3669,47 @@ fn generate_account_paths(num_account_paths: usize) -> (Vec<TempDir>, Vec<PathBu
...
 ) -> SnapshotValidatorConfig {
+    setup_snapshot_validator_config_with(
+        snapshot_interval_slots,
+        Slot::MAX,
+        snapshot_interval_slots,
+        num_account_paths,
+    )
+}
+
+fn setup_snapshot_validator_config_with(
```
carllin, line 3690: nit: Should we just move this into a constructor for `SnapshotValidatorConfig`?

brooksprumo: Sure, can do!

local-cluster/tests/local_cluster.rs
```diff
@@ -3452,10 +3572,10 @@ fn run_test_load_program_accounts(scan_commitment: CommitmentConfig) {
     t_scan.join().unwrap();
 }
 
-fn wait_for_next_snapshot(
+fn wait_for_next_full_snapshot(
```
carllin, line 3575: How about coalescing this functionality with `wait_for_next_full_and_incremental_snapshot()` by

1. Creating top level function `wait_for_next_snapshot(wait_for_snapshot: Vec<SnapshotType>)`. This fetches the latest slot via the client and then calls

one or both of:
2. `wait_for_next_incremental_snapshot(last_slot: Slot)`
3. `wait_for_next_full_snapshot(last_slot: Slot)`

local-cluster/tests/local_cluster.rs
brooksprumo: Done!

local-cluster/tests/local_cluster.rs
brooksprumo: Yah, should be doable!

brooksprumo: So the reason why I have the current implementation is because `wait_for_next_incremental_snapshot()` likely does not do what we want it to. It's how I originally did the code, but ran into this problem:

full snapshot interval: 100
incremental snapshot interval: 10

last full snapshot slot: 500
last incremental slot: 590
last_slot: 595

If waiting for next incremental snapshot, which happens at slot 610, it would have a different base slot (full snapshot slot) of 600 and not 500. So then when the function returned, the previous full snapshot and the returned incremental snapshot would be incompatible.

I think I can solve this with an impl very similar to `wait_for_next_full_and_incremental_snapshot()`, but you'll only have two options: either waiting for the next full snapshot, or waiting for the next incremental snapshot (but this one will always return the associated full snapshot as well).

brooksprumo: Done!

download-utils/src/lib.rs
```diff
@@ -299,3 +305,91 @@ pub fn download_snapshot<'a, 'b>(
         desired_snapshot_hash.0, rpc_addr
     ))
 }
+
+pub fn download_full_snapshot<'a, 'b>(
+    rpc_addr: &SocketAddr,
+    snapshot_config: &SnapshotConfig,
+    desired_full_snapshot_archive_info: &FullSnapshotArchiveInfo,
```
carllin, line 312: Both `FullSnapshotArchiveInfo` and `IncrementalSnapshotArchiveInfo` implement `SnapshotArchiveInfoGetter`. Is there a way to leverage that to coalesce `download_full_snapshot` and `download_incremental_snapshot` into a single function that constructs  a `destination_path` and downloads it?

local-cluster/tests/local_cluster.rs
```diff
@@ -3452,10 +3572,61 @@ fn run_test_load_program_accounts(scan_commitment: CommitmentConfig) {
...
+    IncrementalAndFullSnapshot,
+}
+
+#[derive(Debug)]
+enum NextSnapshotResult {
+    FullSnapshot(FullSnapshotArchiveInfo),
+    IncrementalAndFullSnapshot(IncrementalSnapshotArchiveInfo, FullSnapshotArchiveInfo),
+}
+
+/// Return the next snapshot archive infos after the cluster's last processed slot
```
carllin, line 3624: As we grow the snapshot utilities for local cluster, I think we should move these to a new module.


1) `wait_for_next_incremental_snapshot`, `wait_for_next_full_snapshot` and `wait_for_next_snapshot()` should be in an `impl` block for `LocalCluster`. I think there's enough stuff here where we can actually create an `impl` block for `LocalCluster` in a separate `local_cluster_snapshot_utils.rs`, as shown here: https://hashnode.com/post/managing-big-rust-impls-by-splitting-them-up-ciutsechq0163p253r8kin48n

2. `NextSnapshotType` and `NextSnapshotResult` can be moved into that separate module as well

download-utils/src/lib.rs
brooksprumo: Done. I removed this version of the download and then refactored the existing `download_snapshot()` to work for both Full and Incremental snapshots.

local-cluster/tests/local_cluster.rs
brooksprumo: @carllin The `wait_for_()` functions are just in the test code. Do you want me to move this code out of the tests and into the src?

carllin: Yeah that works for me, all of the logic in the local_cluster/src is pretty much test-specific anyways

brooksprumo: @carllin Done! Let me know if the layout is what you were thinking.

local-cluster/src/local_cluster_snapshot_utils.rs
```diff
@@ -0,0 +1,115 @@
...
+    },
+    snapshot_utils,
+};
+use solana_sdk::{client::SyncClient, commitment_config::CommitmentConfig};
+use std::{path::Path, thread::sleep, time::Duration};
+
+impl LocalCluster {
+    /// Return the next full snapshot archive info after the cluster's last processed slot
+    pub fn wait_for_next_full_snapshot(
+        cluster: &LocalCluster,
```
carllin, line 15: I think these can be replaced with `&self`

brooksprumo: Done!

carllin (approved):

PR #19747, opened 2021-09-09
log accounts loaded for shrink stats

jeffwashington:
#### Problem
useful to know how many accounts (and when) we load accounts for shrink. This is a large user of the accounts index.
#### Summary of Changes

Fixes #


brooksprumo (approved):
lgtm

PR #19756, opened 2021-09-10
Add `solana-close-vote-account` to CLI

drewtaylors:
#### Problem
Attempting to remove all SOL from a vote account results in an intentional error to retain rent exemption. Adding the ability to close a vote account allows users to withdraw all SOL from a vote account.

#### Summary of Changes
* Added a command `solana close-vote-account` to the CLI
```
$ solana close-vote-account --help
solana-close-vote-account 
Close a vote account and withdraw all funds remaining.

USAGE:
    solana close-vote-account [FLAGS] [OPTIONS] <VOTE_ACCOUNT_ADDRESS> <RECIPIENT_ADDRESS>
```
* Added unit and functional tests for respective changes

Fixes #


cli/src/vote.rs
```diff
@@ -907,6 +968,41 @@ pub fn process_withdraw_from_vote_account(
...
+    rpc_client: &RpcClient,
+    config: &CliConfig,
+    vote_account_pubkey: &Pubkey,
+    withdraw_authority: SignerIndex,
+    destination_account_pubkey: &Pubkey,
+    memo: Option<&String>,
+) -> ProcessResult {
+    let latest_blockhash = rpc_client.get_latest_blockhash()?;
+    let withdraw_authority = config.signers[withdraw_authority];
+
```
mvines, line 981: I think `solana close-vote-account` should fail on vote accounts with active stake, since there are very few times when that would be desirable and it's most likely that somebody would accidentally close their staked vote account instead.

That check would look approximately like this (I didn't compile this snippet though, it might blow up!):
```rust
let vote_account_status = rpc_client.get_vote_accounts_with_config(
    RpcGetVoteAccountsConfig {
        vote_pubkey: Some(vote_account_pubkey),
        ..RpcGetVoteAccountsConfig::default()
    }
)?;
if let Some(vote_account) = 
        vote_account_status.current.into_iter().chain(vote_account_status.delinquent.into_iter()).next() {
    if vote_account.activated_stake != 0 {
        return Err(format!("Cannot close a vote account with active stake: {}", vote_account_pubkey).into());   
    }
} else {
    return Err(format!("Vote account does not exist: {}", vote_account_pubkey).into());
}
```

mvines (commented):
This looks great overall, thanks so much for the PR!

cli/src/vote.rs
drew-fisker: sounds good - and thanks for catching this for me.

will look into it!

drewtaylors:
functionality there - just need to add tests and update docs

mvines:
Running `cargo fmt` will make the CI build happy

mvines (dismissed):
Looks great to me, just need to run the CI gauntlet.  Please @ me if you need any help decoding what CI is complaining about!

drewtaylors:
@mvines tests are failing due to the nature of `if-else` block:
```
if let Some(vote_account) = vote_account_status
    .current
    .into_iter()
    .chain(vote_account_status.delinquent.into_iter())
    .next()
}
    if vote_account.activated_stake != 0 {
        return Err(format!(
            "Cannot close a vote account with active stake: {}",
            vote_account_pubkey
        )
        .into());
    }
} else {
    return Err(format!("Vote account does not exist: {}", vote_account_pubkey).into());
}
```

if there's no `root slot` assigned, a vote account fails the above `if-else` and this is what's occurring in the test. i'm not sure how/when a `root slot` is assigned, but please let me know how to do so programmatically. alt we can remove the `else` portion?

mvines:
> if there's no `root slot` assigned, a vote account fails the above `if-else` and this is what's occurring in the test. i'm not sure how/when a `root slot` is assigned, but please let me know how to do so programmatically. alternatively, we can remove the `else` portion?

Ah, I think the `else` can be removed entirely as a vote account in that state has no active stake.

`rpc_client.get_vote_accounts()` only returns the vote accounts that are either staked or actively voting with the rest of the network. Idle/abandoned vote accounts would hit the `else` clause but are totally safe to close.

Aside: a vote account obtains a `root slot` when it submits enough votes along a particular vote to reach a root: https://docs.solana.com/terminology#root

drewtaylors:
> > if there's no `root slot` assigned, a vote account fails the above `if-else` and this is what's occurring in the test. i'm not sure how/when a `root slot` is assigned, but please let me know how to do so programmatically. alternatively, we can remove the `else` portion?
> 
> Ah, I think the `else` can be removed entirely as a vote account in that state has no active stake.
> 
> `rpc_client.get_vote_accounts()` only returns the vote accounts that are either staked or actively voting with the rest of the network. Idle/abandoned vote accounts would hit the `else` clause but are totally safe to close.
> 
> Aside: a vote account obtains a `root slot` when it submits enough votes along a particular fork to create a root: https://docs.solana.com/terminology#root

cool, thank you! really appreciate all the help you've been giving me.

`v1.8.0` has brought up some changes to voting accounts due to `WITHDRAWER_PUBKEY` being required now, so looking into that.

mvines:
> `v1.8.0` has brought up some changes to voting accounts due to `WITHDRAWER_PUBKEY` being required now, so looking into that.

I quickly scanned the failing test and implementation again, nothing stuck out immediately. There's probably a minor keypair mismatch somewhere. Please let me know if you get stuck and I'll take a deeper look

drewtaylors:
hey @mvines after looking at things this morning I'm feeling pretty stumped. :/ 

After playing with the unit tests in `cli/src/tests/vote.rs` and doing some manual testing, taking out more than the minimum balance using the `CloseVoteAccount` renders a vote account unusable even if funds are added to get it back into the minimum. Is there something going on in the backend that removes the `withdrawal authority` after the `vote account` dips below the `rent exemption`?

The test that is failing is https://github.com/solana-labs/solana/blob/fed5b4715e4df710ea482ee7b71a7cc9ade305c6/cli/tests/vote.rs#L167.

mvines:
Ah, we just need this:
```diff
diff --git a/cli/tests/vote.rs b/cli/tests/vote.rs
index 7a5f2c7f4..3ec15df0e 100644
--- a/cli/tests/vote.rs
+++ b/cli/tests/vote.rs
@@ -151,6 +151,17 @@ fn test_vote_authorize_and_withdraw() {
     check_recent_balance(expected_balance, &rpc_client, &vote_account_pubkey);
     check_recent_balance(100, &rpc_client, &destination_account);
 
+    // Re-assign validator identity
+    let new_identity_keypair = Keypair::new();
+    config.signers.push(&new_identity_keypair);
+    config.command = CliCommand::VoteUpdateValidator {
+        vote_account_pubkey,
+        new_identity_account: 2,
+        withdraw_authority: 1,
+        memo: None,
+    };
+    process_command(&config).unwrap();
+
     // Close vote account
     let destination_account = solana_sdk::pubkey::new_rand(); // Send withdrawal to new account to make balance check easy
     config.signers = vec![&default_signer, &withdraw_authority];
@@ -164,14 +175,4 @@ fn test_vote_authorize_and_withdraw() {
     check_recent_balance(0, &rpc_client, &vote_account_pubkey);
     check_recent_balance(expected_balance, &rpc_client, &destination_account);
 
-    // Re-assign validator identity
-    let new_identity_keypair = Keypair::new();
-    config.signers.push(&new_identity_keypair);
-    config.command = CliCommand::VoteUpdateValidator {
-        vote_account_pubkey,
-        new_identity_account: 2,
-        withdraw_authority: 1,
-        memo: None,
-    };
-    process_command(&config).unwrap();
 }
```

Once the vote account is closed (its lamports hit zero), the runtime will delete the account so it's completely expected that the "Re-assign validator identity" setup would fail if it runs afterwards

drewtaylors:
> Once the vote account is closed (its lamports hit zero), the runtime will delete the account so it's completely expected that the "Re-assign validator identity" setup would fail if it runs afterwards

Okay, that makes a lot of sense! Thanks for explaining things to me, I'm pretty new to all of this but trying to learn. Keeping my fingers crossed now that everything is green!

mvines:
Thanks for working on this!

PR #19762, opened 2021-09-10
CPI Account Reuse

Lichtso:
#### Problem
CPI needs an overhaul for inter-operation between the two ABI versions (see #19191).
- The code around CPI is hard to read & understand, needs cleanup
- `native_invoke` performs two unnecessary account copy steps:
  - `Rc::new(keyed_account.account.clone())`
  - `dst_keyed_account.try_account_ref_mut()?.set_data(src_keyed_account.data().to_vec());`
- Syscall `translate_accounts` creates new Account structs as well:
  - `Rc::new(RefCell::new(AccountSharedData::from(Account { .. })))`
- Both syscall and `native_invoke` should be recycling the existing account structs instead of secretly having their own internally, in order to interoperate with ABI v2.

#### Summary of Changes
- General code cleanup, deleted about 200 LOC net
- Consolidated the code around `callee_keyed_accounts`, `caller_write_privileges` and `program_indices` of the CPI syscall and `native_invoke` both into one in `InstructionProcessor::create_message()`.
- Removed the two account copy steps from `native_invoke`.
- Recycles the existing account structs in both syscall and `native_invoke`.

Fixes #

### DIFF
--- a/program-runtime/src/instruction_processor.rs
+++ b/program-runtime/src/instruction_processor.rs
@@ -454,14 +454,14 @@ impl InstructionProcessor {
     pub fn native_invoke(
         invoke_context: &mut dyn InvokeContext,
         instruction: Instruction,
-        keyed_account_indices: &[usize],
+        keyed_account_indices_obsolete: &[usize],
         signers: &[Pubkey],
     ) -> Result<(), InstructionError> {
         let invoke_context = RefCell::new(invoke_context);
         let mut invoke_context = invoke_context.borrow_mut();
 
         let caller_keyed_accounts = invoke_context.get_keyed_accounts()?;
-        let callee_keyed_accounts = keyed_account_indices
+        let callee_keyed_accounts = keyed_account_indices_obsolete
             .iter()
             .map(|index| keyed_account_at_index(caller_keyed_accounts, *index))
             .collect::<Result<Vec<&KeyedAccount>, InstructionError>>()?;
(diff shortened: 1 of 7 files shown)


programs/bpf_loader/src/serialization.rs
```diff
@@ -247,9 +260,14 @@ pub fn serialize_parameters_aligned(
 pub fn deserialize_parameters_aligned(
     keyed_accounts: &[KeyedAccount],
     buffer: &[u8],
+    account_lengths: &[usize],
```
jackcmay, line 263: This is neccessary because a CPI call will change the length of the account within this call?

program-runtime/src/instruction_processor.rs
```diff
@@ -427,215 +447,118 @@ impl InstructionProcessor {
...
 
-        let message = Message::new(&[instruction.clone()], None);
-        let program_id_index = message.instructions[0].program_id_index as usize;
-
-        Ok((message, program_id, program_id_index))
+        Ok((message, caller_write_privileges, program_indices))
     }
 
     /// Entrypoint for a cross-program invocation from a native program
     pub fn native_invoke(
```
jackcmay, line 505: This was mainly a copy-paste from the BPF version, thanks for cleaning it up

programs/bpf_loader/src/serialization.rs
Lichtso: Yes, that was the missing puzzle piece I searched for the entire day.
The hidden purpose of the additional temporary account structs.

Because once the CPI calls recycle the existing account structs, it overwrites the length and so the deserialize of the caller program afterwards parses and writes back garbage.

PR #19764, opened 2021-09-10
Proposal: event data for Solidity

seanyoung:
#### Problem

Solana has no support for solidity events.

#### Summary of Changes

Fixes #

### DIFF
--- a/programs/bpf_loader/src/syscalls.rs
+++ b/programs/bpf_loader/src/syscalls.rs
@@ -22,7 +22,7 @@ use solana_sdk::{
         allow_native_ids, blake3_syscall_enabled, check_seed_length,
         close_upgradeable_program_accounts, demote_program_write_locks, disable_fees_sysvar,
         libsecp256k1_0_5_upgrade_enabled, mem_overlap_fix, return_data_syscall_enabled,
-        secp256k1_recover_syscall_enabled,
+        secp256k1_recover_syscall_enabled, sol_log_data_syscall_enabled,
     },
     hash::{Hasher, HASH_BYTES},
     ic_msg,
@@ -181,6 +181,11 @@ pub fn register_syscalls(
             .register_syscall_by_name(b"sol_set_return_data", SyscallSetReturnData::call)?;
         syscall_registry
             .register_syscall_by_name(b"sol_get_return_data", SyscallGetReturnData::call)?;
+    }
+
+    // Log data
+    if invoke_context.is_feature_active(&sol_log_data_syscall_enabled::id()) {
+        syscall_registry.register_syscall_by_name(b"sol_log_data", SyscallLogData::call)?;
     }
 
     Ok(syscall_registry)
@@ -357,6 +362,8 @@ pub fn bind_syscall_context_objects<'a>(
         !invoke_context.is_feature_active(&disable_fees_sysvar::id());
     let is_return_data_syscall_active =
         invoke_context.is_feature_active(&return_data_syscall_enabled::id());
+    let is_sol_log_data_syscall_active =
+        invoke_context.is_feature_active(&sol_log_data_syscall_enabled::id());
 
     let invoke_context = Rc::new(RefCell::new(invoke_context));
 
@@ -404,6 +411,16 @@ pub fn bind_syscall_context_objects<'a>(
         vm,
         is_return_data_syscall_active,
         Box::new(SyscallGetReturnData {
+            invoke_context: invoke_context.clone(),
+            loader_id,
+        }),
+    );
+
+    // sol_log_data
+    bind_feature_gated_syscall_context_object!(
+        vm,
+        is_sol_log_data_syscall_active,
+        Box::new(SyscallLogData {
             invoke_context: invoke_context.clone(),
             loader_id,
         }),
@@ -2374,6 +2391,72 @@ impl<'a> SyscallObject<BpfError> for SyscallGetReturnData<'a> {
         } else {
             *result = Ok(0);
         }
+    }
+}
+
+// Log data handling
+pub struct SyscallLogData<'a> {
+    invoke_context: Rc<RefCell<&'a mut dyn InvokeContext>>,
+    loader_id: &'a Pubkey,
+}
+impl<'a> SyscallObject<BpfError> for SyscallLogData<'a> {
+    fn call(
+        &mut self,
+        addr: u64,
+        len: u64,
+        _arg3: u64,
+        _arg4: u64,
+        _arg5: u64,
+        memory_mapping: &MemoryMapping,
+        result: &mut Result<u64, EbpfError<BpfError>>,
+    ) {
+        let invoke_context = question_mark!(
+            self.invoke_context
+                .try_borrow()
+                .map_err(|_| SyscallError::InvokeContextBorrowFailed),
+            result
+        );
+
+        let budget = invoke_context.get_compute_budget();
+
+        question_mark!(
+            invoke_context
+                .get_compute_meter()
+                .consume(budget.syscall_base_cost),
+            result
+        );
+
+        let untranslated_fields = question_mark!(
+            translate_slice::<&[u8]>(memory_mapping, addr, len, self.loader_id),
+            result
+        );
+
+        question_mark!(
+            invoke_context
+                .get_compute_meter()
+                .consume(untranslated_fields.iter().map(|e| e.len() as u64).sum()),
+            result
+        );
+
+        let mut fields = Vec::with_capacity(untranslated_fields.len());
+
+        for untranslated_field in untranslated_fields {
+            fields.push(question_mark!(
+                translate_slice::<u8>(
+                    memory_mapping,
+                    untranslated_field.as_ptr() as *const _ as u64,
+                    untranslated_field.len() as u64,
+                    self.loader_id,
+                ),
+                result
+            ));
+        }
+
+        let logger = invoke_context.get_logger();
+
+        stable_log::program_data(&logger, &fields);
+
+        *result = Ok(0);
     }
 }
 
--- a/sdk/bpf/c/inc/sol/log.h
+++ b/sdk/bpf/c/inc/sol/log.h
@@ -41,6 +41,11 @@ static void sol_log_array(const uint8_t *array, int len) {
 }
 
 /**
+ * Print the base64 representation of some arrays.
+ */
+void sol_log_data(SolBytes *fields, uint64_t fields_len);
+
+/**
  * Prints the program's input parameters
  *
  * @param params Pointer to a SolParameters structure
--- a/sdk/src/feature_set.rs
+++ b/sdk/src/feature_set.rs
@@ -211,6 +211,10 @@ pub mod reduce_required_deploy_balance {
     solana_sdk::declare_id!("EBeznQDjcPG8491sFsKZYBi5S5jTVXMpAKNDJMQPS2kq");
 }
 
+pub mod sol_log_data_syscall_enabled {
+    solana_sdk::declare_id!("HYPs7jyJ3KwQFdDpuSzMtVKf1MLJDaZRv3CSWvfUqdFo");
+}
+
 lazy_static! {
     /// Map of feature identifiers to user-visible description
     pub static ref FEATURE_NAMES: HashMap<Pubkey, &'static str> = [
@@ -259,6 +263,7 @@ lazy_static! {
         (return_data_syscall_enabled::id(), "enable sol_{set,get}_return_data syscall"),
         (fix_write_privs::id(), "fix native invoke write privileges"),
         (reduce_required_deploy_balance::id(), "reduce required payer balance for program deploys"),
+        (sol_log_data_syscall_enabled::id(), "enable sol_log_data syscall"),
         /*************** ADD NEW FEATURES HERE ***************/
     ]
     .iter()
--- a/sdk/src/process_instruction.rs
+++ b/sdk/src/process_instruction.rs
@@ -1,5 +1,6 @@
 #![cfg(feature = "full")]
 
+use itertools::Itertools;
 use solana_sdk::{
     account::AccountSharedData,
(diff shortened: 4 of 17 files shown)


docs/src/proposals/event_data.md
```diff
@@ -0,0 +1,155 @@
+# Event data
```
mvines, line 1:
```suggestion
# Program Events
```

sdk/src/process_instruction.rs
```diff
@@ -321,6 +322,24 @@ pub mod stable_log {
         ic_logger_msg!(logger, "Program log: {}", message);
     }
 
+    /// Log event data as from the program itself.
+    ///
+    /// The general form is:
+    ///
+    /// ```notrust
+    /// "Program event data: <program-id> <event-data-in-base64>+"
```
mvines, line 330:
```suggestion
    /// "Program event: <program-id> <event-data-in-base64>+"
```

sdk/src/process_instruction.rs
```diff
@@ -321,6 +322,24 @@ pub mod stable_log {
         ic_logger_msg!(logger, "Program log: {}", message);
     }
 
+    /// Log event data as from the program itself.
```
mvines, line 325:
```suggestion
    /// Emit a program event.
```

sdk/src/process_instruction.rs
```diff
@@ -321,6 +322,24 @@ pub mod stable_log {
...
+    ///
+    /// ```notrust
+    /// "Program event data: <program-id> <event-data-in-base64>+"
+    /// ```
+    ///
+    /// That is, any program-generated output is guaranteed to be prefixed by "Program event data: "
+    pub fn program_event(logger: &Rc<RefCell<dyn Logger>>, program_id: &Pubkey, data: &[&[u8]]) {
+        ic_logger_msg!(
+            logger,
+            "Program return data: {} {}",
```
mvines, line 337:
```suggestion
            "Program event: {} {}",
```

mvines: wdyt about shortening "Program return data:" to "Program return:" too?

docs/src/proposals/event_data.md
```diff
@@ -0,0 +1,155 @@
...
+
+This requires a new system call:
+
+```
+struct event_field {
+    const uint8_t *addr;
+    uint64_t length;
+};
+
+void sol_event_data(struct event_fields fields, uint64_t length);
```
mvines, line 148:
```suggestion
void sol_event(struct event_fields fields, uint64_t length);
```

sdk/src/process_instruction.rs
seanyoung: Good idea

sdk/src/feature_set.rs
```diff
@@ -258,7 +262,8 @@ lazy_static! {
         (ed25519_program_enabled::id(), "enable builtin ed25519 signature verify program"),
         (allow_native_ids::id(), "allow native program ids in program derived addresses"),
         (check_seed_length::id(), "Check program address seed lengths"),
-        (return_data_syscall_enabled::id(), "enable sol_{set,get}_return_data syscall")
+        (return_data_syscall_enabled::id(), "enable sol_{set,get}_return_data syscall"),
+        (event_data_syscall_enabled::id(), "enable sol_event_data syscall"),
```
mvines, line 266:
```suggestion
        (sol_event_syscall_enabled::id(), "enable sol_event_data syscall"),
```

mvines (approved):

docs/src/proposals/event_data.md
```diff
@@ -0,0 +1,155 @@
...
+Events should be added to the program log. The program log should include:
+
+ - program_id that generate the data
+ - base64 encoded data (more than one permitted)
+
+So if we encoding the topics first, followed by the data then the event in the
+tx above would look like:
+```
+program event: GaBtBJvmS4Arjj5W1NmFcyvPjsHN38UGYDq2MDwbs9Qu 3fJSrRviyJtpwrBo/DeNqpUrpFjxKEWKPVaTfUjs8AAAAAAAAAAAAAAACC+JDWOEeNIR7yII88FGa1q/g1UQAAAAAAAAAAAAAAAKEkMdC522QANLDN/O75zOFh5ivk AAAAAAAAAAAAAAAAAAAAAAAAAAAAAACgMNzrvS9MAAA=
+```
```
joncinque, line 138: To be clear, are all of the topics encoded as one continuous piece of data?  And is the idea that an event listener would slice this up into 32-byte chunks and then check against each one of those?  I wonder if some more structure might make it clearer, something like: 

```
program event: GaBtBJvmS4Arjj5W1NmFcyvPjsHN38UGYDq2MDwbs9Qu topics 3fJSrRviyJtpwrBo/DeNqpUrpFjxKEWKPVaTfUjs8 AAAAAAAAAAAAAAACC+JDWOEeNIR7yII88FGa1q/g1UQ AAAAAAAAAAAAAAAKEkMdC522QANLDN/O75zOFh5ivk data AAAAAAAAAAAAAAAAAAAAAAAAAAAAAACgMNzrvS9MAAA=
```

(I hope I got the spaces right in the topics part) Would that potentially be clearer for a solidity dev, or would they not care?  On the other hand, this approach would make it look a lot more like eth, which is probably what you were trying to avoid :smile:

seanyoung: I'm am concerned about making it too eth-centric.

Maybe we could do something like this:

```
struct event_field {
    const uint8_t *data;
    uint64_t data_length;
    const char *tag;
    uint64_t tag_length;
};
void sol_event(struct event_fields fields, uint64_t length);
```

For Solidity events, the BPF could add the "topics" add "data" tag. I think should be printed with a colon so it is clear that what the tag refers to:

```
program event: GaBtBJvmS4Arjj5W1NmFcyvPjsHN38UGYDq2MDwbs9Qu topics:3fJSrRviyJtpwrBo/DeNqpUrpFjxKEWKPVaTfUjs8AAAAAAAAAAAAAAACC+JDWOEeNIR7yII88FGa1q/g1UQAAAAAAAAAAAAAAAKEkMdC522QANLDN/O75zOFh5ivk data:AAAAAAAAAAAAAAAAAAAAAAAAAAAAAACgMNzrvS9MAAA=
```

jackcmay: As long as the event is formatted consistently I don't think folks need the tags, and "topics:" terminology is eth-centric

docs/src/proposals/event_data.md
```diff
@@ -0,0 +1,155 @@
...
+};
+
+void sol_event(struct event_fields fields, uint64_t length);
+```
+
+### Considerations
+
+- Not sure if "event" is a good name. It is a generic binary data log facility
+- Should there be text field in the program log so we can have a little bit of
+  metadata on the binary data
```
jackcmay, line 155: To make the "event" data more human readable?

docs/src/proposals/event_data.md
```diff
@@ -0,0 +1,155 @@
...
+
+This requires a new system call:
+
+```
+struct event_field {
+    const uint8_t *addr;
+    uint64_t length;
+};
+
+void sol_event(struct event_fields fields, uint64_t length);
```
jackcmay, line 148: How are "topics" and "data" delineated with this API?

programs/bpf_loader/src/syscalls.rs
```diff
@@ -2717,6 +2734,80 @@ impl<'a> SyscallObject<BpfError> for SyscallGetReturnData<'a> {
...
+        let invoke_context = question_mark!(
+            self.invoke_context
+                .try_borrow()
+                .map_err(|_| SyscallError::InvokeContextBorrowFailed),
+            result
+        );
+
+        let budget = invoke_context.get_compute_budget();
+
+        question_mark!(
```
jackcmay, line 2762: Can't we go ahead and consume `len` up front?

docs/src/proposals/event_data.md
seanyoung: So the API allows any number of fields to be specified. With solidity, there will always be two fields: first topics, then data.

programs/bpf_loader/src/syscalls.rs
seanyoung: We need to get the sum of all of the length of all the fields.

docs/src/proposals/event_data.md
seanyoung: Absolutely, I've added this to the proposal.

docs/src/proposals/event_data.md
jackcmay: Ah, and topics would be a single buffer consisting of multiple 32-byte sequences?

programs/bpf_loader/src/syscalls.rs
jackcmay: True dat

docs/src/proposals/event_data.md
seanyoung: Exactly

programs/bpf_loader/src/syscalls.rs
seanyoung: I've used the same costing model as `sol_log`

docs/src/proposals/event_data.md
jackcmay: Yeah, this does sound like more of a general-purpose logging facility.  I think we might want to keep the "event" terminology out of this api

seanyoung:
Solang side is also coming along, needs some more tests: https://github.com/hyperledger-labs/solang/pull/494

docs/src/proposals/event_data.md
seanyoung: I've struggled to come up with a better name. Any suggestions? `sol_log_binary()`?

jackcmay: I'm terrible with names but that, or `sol_log_data`?

seanyoung: I've renamed to `sol_log_data`. It now looks pretty similar to the other log functions.

docs/src/proposals/event_data.md
seanyoung: The idea is that that tags would be provided by the BPF program, not hardcoded in solana. I'm undecided though, it seems a bit complex.

sdk/bpf/c/inc/sol/log.h
```diff
@@ -40,6 +40,19 @@ static void sol_log_array(const uint8_t *array, int len) {
   }
 }
 
+/**
+ * A single data field for sol_log_data
+ */
+typedef struct {
+  const uint8_t *data;
+  uint64_t len;
+} SolLogDataField;
```
jackcmay, line 49: Any reason not to use `SolBytes` defined in `sdk/bpf/c/inc/sol/types.h`?

sdk/src/process_instruction.rs
```diff
@@ -321,20 +322,38 @@ pub mod stable_log {
...
     ///
     /// The general form is:
     ///
     /// ```notrust
-    /// "Program return data: <program-id> <program-generated-data-in-base64>"
+    /// "Program return: <program-id> <program-generated-data-in-base64>"
     /// ```
     ///
     /// That is, any program-generated output is guaranteed to be prefixed by "Program return data: "
-    pub fn program_return_data(logger: &Rc<RefCell<dyn Logger>>, program_id: &Pubkey, data: &[u8]) {
```
jackcmay, line 334: Why rename?

jackcmay (approved):

sdk/bpf/c/inc/sol/log.h
seanyoung: Good point, I've fixed that.

sdk/src/process_instruction.rs
seanyoung: That was a change proposed by @mvines 

Now that we're calling it "program log data:" I don't think it makes much sense to change this either, so I've reverted this change.

jackcmay (dismissed):

Lichtso:
This might have been answered already: What is the reason the existing logging syscall won't do?
The text / string encoding vs "raw" binary, or is there something else?

seanyoung:
> This might have been answered already: What is the reason the existing logging syscall won't do?
> The text / string encoding vs "raw" binary, or is there something else?

The existing syscalls don't do "raw" binary and it would be useful to have solidity events separated from program log, for client side filtering.

sdk/src/process_instruction.rs
```diff
@@ -321,6 +322,24 @@ pub mod stable_log {
         ic_logger_msg!(logger, "Program log: {}", message);
     }
 
+    /// Emit a program log data.
+    ///
+    /// The general form is:
+    ///
+    /// ```notrust
+    /// "Program log data: <program-id> <binary-data-in-base64>+"
```
mvines, line 330: 🤔 I have a couple nits about the log form:

1. The "Program log data:" feels a little wordy.  How about "Program data:" or "Program event:"?  
2. Turns out we don't need the "<program-id>" here at all because a [program invoke](https://github.com/solana-labs/solana/blob/b507715d449acd5f04b1563957e4d52392abe700/sdk/src/process_instruction.rs#L297) message is guaranteed to precede this message.  We make this same assumption to avoid repeating the program id in program log messages already.

seanyoung: Good points! 

1. "Program data: " is a bit better, I agree. I'm having trouble coming up with a great name here.
2. Aha! That's a great point. I've also update the program return data entries.

sdk/src/process_instruction.rs
seanyoung: I've flip flopped on this. Now they are:

```
Program data: <base64 event data>
Program return: <base64 return data>
```

sdk/src/process_instruction.rs
seanyoung: I've also renamed "Program return data: " to "Program return: "

PR #19767, opened 2021-09-10
accounts index iterator uses copy of pubkey and account map entry

jeffwashington:
#### Problem
Minor refactoring making accounts iteration return owned copies earlier in the call chain. This maps better to disk buckets later.
#### Summary of Changes

Fixes #


brooksprumo (approved):
lgtm

carllin (approved):

PR #19769, opened 2021-09-10
refactor generate_index zero lamports

jeffwashington:
#### Problem
This code will change more as we move to disk based accounts index. For the moment, this refactoring improves performance and will clear a way for other changes that are coming.
#### Summary of Changes

Fixes #


brooksprumo (approved):
looks good to me

PR #19771, opened 2021-09-10
Add more docs for RpcClient

brson:
This adds more docs and examples to the RpcClient Rust API.

Followup to https://github.com/solana-labs/solana/pull/19021.

Many of the methods documented in this patch are simple accessors, and their docs just restate the one-line description from the spec, while linking to the spec.

After this patch, there are only a handful of undocumented non-deprecated methods left, many of which deal with the token program. These are a bit tricky to understand and document, so I have left it to the future.

This is the last doc patch I intend to submit for RpcClient, at least for now, as I have been poking at it for months and want to move on to document other modules. I will submit separately a bunch of integration tests I wrote while writing these docs.

### DIFF
--- a/client/src/rpc_client.rs
+++ b/client/src/rpc_client.rs
@@ -80,7 +80,7 @@ impl RpcClientConfig {
 /// querying and transacting with the network from external programs.
 ///
 /// This type builds on the underlying RPC protocol, adding extra features such
-/// as timeout handling, retries, and waiting on transaction commitment levels.
+/// as timeout handling, retries, and waiting on transaction [commitment levels][cl].
 /// Some methods simply pass through to the underlying RPC protocol. Not all RPC
 /// methods are encapsulated by this type, but `RpcClient` does expose a generic
 /// [`send`](RpcClient::send) method for making any [`RpcRequest`].
@@ -1659,7 +1659,9 @@ impl RpcClient {
 
     /// Returns epoch activation information for a stake account.
     ///
-    /// This method uses the configured commitment level.
+    /// This method uses the configured [commitment level].
+    ///
+    /// [cl]: https://docs.solana.com/developing/clients/jsonrpc-api#configuring-state-commitment
     ///
     /// # RPC Reference
     ///
@@ -1745,7 +1747,9 @@ impl RpcClient {
 
     /// Returns information about the current supply.
     ///
-    /// This method uses the configured commitment level.
+    /// This method uses the configured [commitment level][cl].
+    ///
+    /// [cl]: https://docs.solana.com/developing/clients/jsonrpc-api#configuring-state-commitment
     ///
     /// # RPC Reference
     ///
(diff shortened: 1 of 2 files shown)


brson:
I'll investigate the ci failure tomorrow.

CriesofCarrots:
> I'll investigate the ci failure tomorrow.

My guess is that it's the change to the `getAccountInfo` mock that is causing the test failures. Let me know if it's too prickly; we can rewrite to use custom mocks (which probably should be done more broadly; unfortunately, a lot of tests predate the custom-mock approach).

brson:
@CriesofCarrots I've pushed two commits that I think will fix the failures. They have rpc_client use custom mocks to cover the case that broke existing tests.

brson:
I needed to create a support function, create_rpc_client_mocks, for the RpcClient examples, and I put it directly in the rpc_client module, with `#[doc(hidden)]`. There's not a great place to put this function since it needs to be shared by examples.

brson:
New CI failures. I'll give them a look soon.

CriesofCarrots:
> New CI failures. I'll give them a look soon.

I think the downstream-projects failure will be resolved if you rebase on master.
So I'll get to reviewing this in the meantime!

client/src/rpc_client.rs
```diff
@@ -2947,20 +3439,117 @@ impl RpcClient {
...
+    /// # let rpc_client = RpcClient::new_mock("succeeds".to_string());
+    /// let slot = rpc_client.get_max_shred_insert_slot()?;
+    /// # Ok::<(), ClientError>(())
     pub fn get_max_shred_insert_slot(&self) -> ClientResult<Slot> {
         self.send(RpcRequest::GetMaxShredInsertSlot, Value::Null)
     }
 
+    /// Returns the account information for a list of pubkeys.
+    ///
+    /// This method uses the configured default commitment level.
```
CriesofCarrots, line 3490:
```suggestion
    /// This method uses the configured commitment level.
```
And link?

client/src/rpc_client.rs
```diff
@@ -3022,13 +3703,65 @@ impl RpcClient {
         Ok(minimum_balance)
     }
 
-    /// Request the balance of the account `pubkey`.
+    /// Request the balance of the provided account pubkey.
+    ///
```
CriesofCarrots, line 3707: Add? 
```
/// This method uses the configured commitment level.
```

CriesofCarrots (commented):
These look great!
Most of my comments are around the phrase `commitment level`. Should we link them all for consistency?

client/src/rpc_client.rs
```diff
@@ -2297,12 +2446,29 @@ impl RpcClient {
...
+    /// [`Finalized`]: CommitmentLevel::Finalized.
+    ///
+    /// # Errors
+    ///
+    /// This method returns an error if the given commitment level is below
+    /// [`Confirmed`].
+    ///
+    /// [`Confirmed`]: CommitmentLevel::Confirmed
+    ///
```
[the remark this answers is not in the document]
brson: Great catch. Must have been a copy-paste error.

brson:
I accepted all suggestions, linked all first instances of "commitment level" to the appropriate docs, and rebased against master. Waiting for the CI to show if that fixed the downstream breakage.

Thanks for the good review.

CriesofCarrots:
Yep, all green. Thanks for your patient work on this!

PR #19773, opened 2021-09-10
factor out InMemAccountsIndex to prepare for disk index

jeffwashington:
#### Problem
Factor out one disk index bin so we can begin adding metrics and refactoring towards disk buckets. Previously, this was a hashmap directly used by AccountsIndex.
#### Summary of Changes

Fixes #


brooksprumo (approved):
lgtm

carllin (approved):

PR #19776, opened 2021-09-10
Add default() to SnapshotConfig

brooksprumo:
#### Problem

SnapshotConfig is about to get more parameters, and updating all its uses can be cumbersome. Adding default() simplifies future PRs.

#### Summary of Changes

Add default() to SnapshotConfig and then update all its uses.

### DIFF
--- a/runtime/src/snapshot_config.rs
+++ b/runtime/src/snapshot_config.rs
@@ -1,5 +1,4 @@
-use crate::snapshot_utils::ArchiveFormat;
-use crate::snapshot_utils::SnapshotVersion;
+use crate::snapshot_utils::{self, ArchiveFormat, SnapshotVersion};
 use solana_sdk::clock::Slot;
 use std::path::PathBuf;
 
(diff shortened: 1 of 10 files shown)


runtime/src/snapshot_config.rs
```diff
@@ -31,3 +30,22 @@ pub struct SnapshotConfig {
...
 }
+
+impl Default for SnapshotConfig {
+    fn default() -> Self {
+        Self {
+            full_snapshot_archive_interval_slots:
+                snapshot_utils::DEFAULT_FULL_SNAPSHOT_ARCHIVE_INTERVAL_SLOTS,
+            incremental_snapshot_archive_interval_slots:
+                snapshot_utils::DEFAULT_INCREMENTAL_SNAPSHOT_ARCHIVE_INTERVAL_SLOTS,
+            snapshot_archives_dir: PathBuf::default(),
```
jeffwashington, line 41: is this == snapshot_archives_dir: PathBuf::from("/"),?

brooksprumo: Good catch! No, it's basically an empty string, like `String::new()`, which ends up being `Vec::new()`.

In the rpc_service test that sets snapshot_archive_dir and bank_snapshots_dir to `PathBuf::from("/")`, the values don't matter at all, just that SnapshotConfig is `Some` and not `None`.

jeffwashington (approved):
lgtm

PR #19778, opened 2021-09-10
InMemAccountsIndex iter returns vec<owned>

jeffwashington:
#### Problem
this will be required by disk index
#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/in_mem_accounts_index.rs
+++ b/runtime/src/in_mem_accounts_index.rs
@@ -1,7 +1,7 @@
 use crate::accounts_index::{AccountMapEntry, IsCached};
 use solana_sdk::pubkey::Pubkey;
 use std::collections::{
-    hash_map::{Entry, Iter, Keys},
+    hash_map::{Entry, Keys},
     HashMap,
 };
 use std::fmt::Debug;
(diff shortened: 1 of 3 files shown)


brooksprumo:
I think this PR needs to rebased on master, now that #19775 has been merged.

runtime/src/in_mem_accounts_index.rs
```diff
@@ -26,8 +26,8 @@ impl<V: IsCached> InMemAccountsIndex<V> {
         self.map.entry(pubkey)
     }
 
-    pub fn iter(&self) -> Iter<K, AccountMapEntry<V>> {
-        self.map.iter()
+    pub fn iter(&self) -> Vec<(K, AccountMapEntry<V>)> {
+        self.map.iter().map(|(k, v)| (*k, v.clone())).collect()
```
brooksprumo, line 30: I understand the intent here, but I do find it odd to have a function called `iter()` that _doesn't_ return an iterator. I'll see if I can poke around other rust stuffs to see if there's already a name for this. Maybe something like `to_vec()` would work?

brooksprumo (approved):
`items()` looks good to me!

PR #19779, opened 2021-09-10
in_mem_accounts_index iter filters by range

jeffwashington:
#### Problem
Move items range filtering deeper into collection to prepare for disk based index.
#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/in_mem_accounts_index.rs
+++ b/runtime/src/in_mem_accounts_index.rs
@@ -12,7 +12,7 @@ use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
 
 use std::fmt::Debug;
-
+use std::ops::RangeBounds;
 type K = Pubkey;
 
 // one instance of this represents one bin of the accounts index.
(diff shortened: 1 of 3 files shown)


runtime/src/in_mem_accounts_index.rs
```diff
@@ -26,8 +26,17 @@ impl<V: IsCached> InMemAccountsIndex<V> {
         self.map.entry(pubkey)
     }
 
-    pub fn items(&self) -> Vec<(K, AccountMapEntry<V>)> {
-        self.map.iter().map(|(k, v)| (*k, v.clone())).collect()
+    pub fn items<R>(&self, range: &Option<&R>) -> Vec<(K, AccountMapEntry<V>)>
+    where
+        R: RangeBounds<Pubkey> + std::fmt::Debug,
```
brooksprumo, line 31: What is the `fmt::Debug` used for here (and same for the other fn)?

jeffwashington: so it is possible to print/log it. I have found this convenient to have present while debugging and it has no penalty, so I went ahead and added it now.

runtime/src/in_mem_accounts_index.rs
```diff
@@ -49,9 +49,18 @@ impl<T: IsCached> InMemAccountsIndex<T> {
...
     }
 
-    pub fn items(&self) -> Vec<(K, AccountMapEntry<T>)> {
+    pub fn items<R>(&self, range: &Option<&R>) -> Vec<(K, AccountMapEntry<T>)>
+    where
+        R: RangeBounds<Pubkey> + std::fmt::Debug,
+    {
         Self::update_stat(&self.stats().items, 1);
-        self.map.iter().map(|(k, v)| (*k, v.clone())).collect()
+        let mut result = Vec::with_capacity(self.map.len());
```
brooksprumo, line 57: If the range is `None`, I agree the capacity parameter should be `map.len()`. If the range is `Some`, would it be beneficial to set the capacity to the range's width instead?

(This is me assuming that keys are unique, so if they are not, then the current impl makes sense.)

brooksprumo (commented):
Looks good, just one question

runtime/src/in_mem_accounts_index.rs
jeffwashington: range width is specified in pubkey values. We don't know how many pubkeys are in a range width ex: (0xffde... to 0xffdf...). We only know that we will return <= map.len, which is all the entries present in the accounts index. I don't think we can even get the width of a range of pubkeys. Perhaps as a 32 byte integer value, which is bigger than usize at any rate. I think map.len is the only thing that makes sense.

brooksprumo: Ah, yeah, that makes sense. I was thinking small scale and none-sparse. Thanks for spelling it out for me.

brooksprumo (approved):
lgtm

PR #19793, opened 2021-09-11
Remove clippy exception

sakridge:
#### Problem

Unnecessary clippy exception in gossip code

#### Summary of Changes

Remove

Fixes #

### DIFF
--- a/gossip/src/gossip_service.rs
+++ b/gossip/src/gossip_service.rs
@@ -143,10 +143,7 @@ pub fn discover(
     Vec<ContactInfo>, // all gossip peers
     Vec<ContactInfo>, // tvu peers (validators)
 )> {
-    let keypair = {
-        #[allow(clippy::redundant_closure)]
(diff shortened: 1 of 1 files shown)


gossip/src/gossip_service.rs
```diff
@@ -143,10 +143,7 @@ pub fn discover(
     Vec<ContactInfo>, // all gossip peers
     Vec<ContactInfo>, // tvu peers (validators)
 )> {
-    let keypair = {
-        #[allow(clippy::redundant_closure)]
-        keypair.unwrap_or_else(|| Keypair::new())
```
sakridge, line 148: Maybe I'm missing something? Doesn't seem like we need this.

mvines: Looks like it, yeah. I got a clippy complaint about this at one point but glad it's gone

mvines (approved):

PR #19794, opened 2021-09-11
Weekend fun: Less deps

gilescope:
Just a hunch , but you might not need these deps.


mvines:
Thanks!  To make that CI build happy, try running:
```
$ cd programs/bpf
$ cargo check
```
and include the Cargo.lock changes in this PR

mvines:
Thanks for choosing Solana for your weekend fun!

PR #19797, opened 2021-09-12
accounts index stats

jeffwashington:
#### Problem
This is a little complicated. Goal is to create a struct to accumulate metrics for accounts index and then dump the metrics every 10s.
This reporting accomplished in a bg thread that wakes up every 10s. It could be done in a fg thread, in theory, but every account index operation may have to check if it is time to report stats, which is bad. We will need a bg thread for disk buckets soon anyway, so it seems like the right thing to do to create it now.
We eventually will want aggregated per-bin/bucket stats (for example, min/max bucket len) and that could be a traversal of the whole backing store. That could mean a non-trivial amount of time and lock management. Thus, the longer term strategy should be to report metrics in the bg.

#### Summary of Changes

Fixes #


jeffwashington:
sample metrics:
```
accounts_index deletes=0i gets_from_mem=1246131i get_mem_us=101i gets_missing=2452i get_missing_us=0i items=53i keys=0i
```

runtime/src/accounts_index_storage.rs
```diff
@@ -0,0 +1,65 @@
...
+
+impl Drop for AccountsIndexStorage {
+    fn drop(&mut self) {
+        self.exit.store(true, Ordering::Relaxed);
+        self.wait.notify_all();
+        if let Some(x) = self.handle.take() {
+            x.join().unwrap()
+        }
+    }
+}
```
brooksprumo, line 35: Does the order that these stores are visible to other threads matter? If so, then the `exit` ordering probably needs to change. If the other threads can handle any order of these stores being visible, then this is all good.

brooksprumo (commented):
Looks good, just a few questions. I also see in my email that new commits may've already been pushed, so I'll take a look again right afterwards.

brooksprumo:
@jeffwashington Since Github removes this PR from the list of "Review requests", can you re-request a review from me after commenting/resolving, so that I don't miss a PR to review? Thanks!

runtime/src/in_mem_accounts_index.rs
```diff
@@ -1,44 +1,80 @@
...
+    pub fn new(storage: &AccountsIndexStorage) -> Self {
         Self {
             map: HashMap::new(),
+            storage: storage.storage().clone(),
         }
     }
 
+    pub fn new_bucket_map_holder() -> Arc<BucketMapHolder> {
+        Arc::new(BucketMapHolder::new())
+    }
```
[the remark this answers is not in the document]
jeffwashington: no, apparently not.

runtime/src/accounts_index_storage.rs
jeffwashington: brooks and I discussed in slack. worst case is tests take 10s. this code will be reworked once we have more bg threads that are actually doing work.

brooksprumo (dismissed):
lgtm

PR #19799, opened 2021-09-12
move upsert and a handful of helpers to InMemAccountsIndex

jeffwashington:
#### Problem
pushing account index functionality into InMemAccountsIndex to prepare for disk buckets

#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/accounts_index.rs
+++ b/runtime/src/accounts_index.rs
@@ -16,7 +16,7 @@ use solana_sdk::{
     pubkey::{Pubkey, PUBKEY_BYTES},
 };
 use std::{
-    collections::{btree_map::BTreeMap, hash_map::Entry, HashSet},
+    collections::{btree_map::BTreeMap, HashSet},
     fmt::Debug,
     ops::{
         Bound,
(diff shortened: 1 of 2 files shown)


runtime/src/accounts_index.rs
```diff
@@ -2989,14 +2904,14 @@ pub mod tests {
...
-        WriteAccountMapEntry::upsert(
-            w_account_maps,
+        let mut w_account_maps = index.get_account_maps_write_lock(&key.pubkey());
+        w_account_maps.upsert(
             &key.pubkey(),
             new_entry,
             &mut SlotList::default(),
             UPSERT_PREVIOUS_SLOT_ENTRY_WAS_CACHED_FALSE,
         );
+        drop(w_account_maps);
```
brooksprumo, line 2914: Can this be collapsed into `index.get_account_maps_write_lock(&key.pubkey()).upsert(...)`? I think that would remove the need for the `drop` too maybe?

brooksprumo (commented):
Looks good, just the one comment. There's no functional changes, right, just moving the code around?

runtime/src/accounts_index.rs
jeffwashington: yes, could be collapsed. just test code and I was trying to change as few lines as possible in these refactorings.

brooksprumo (approved):
lgtm

PR #19802, opened 2021-09-12
get rid of 'entry' from InMemAccountsIndex

jeffwashington:
#### Problem
Disk based account index has a hard time supporting entry. It isn't necessary if the actual operations are operations on the account index directly. So, after moving them all, get rid of the pub for entry.
#### Summary of Changes

Fixes #


brooksprumo (approved):

jeffwashington:
abandoning this. It is better to keep the private function and add metrics to it. I did that elsewhere.

PR #19806, opened 2021-09-12
metrics on accounts index entry

jeffwashington:
#### Problem
entry (vs get) is a different type of accounts index operation. Reroute all internal entry calls to use the existing, now-private function so we get metrics on entry.
#### Summary of Changes

Fixes #


brooksprumo (approved):
lgtm

PR #19810, opened 2021-09-12
Track reset bank in PohRecorder

carllin:
#### Problem
BankingStage would like access to a bank to access validator stakes for QoS.

However, when the validator is not a leader, it PohRecorder does not currently have a bank.

This can be naturally solved if we always tracked in PohRecorder the last reset bank from startup/replay.

Extracted from https://github.com/solana-labs/solana/pull/19611.

#### Summary of Changes
Track a `start_bank` instead of a `start_slot` in `PohRecorder`

Most of the changes are just test changes, important bits are called out in comments.

Fixes #

### DIFF
--- a/poh/src/poh_recorder.rs
+++ b/poh/src/poh_recorder.rs
@@ -18,7 +18,11 @@ use {
     },
     log::*,
     solana_entry::{entry::Entry, poh::Poh},
-    solana_ledger::{blockstore::Blockstore, leader_schedule_cache::LeaderScheduleCache},
+    solana_ledger::{
+        blockstore::Blockstore,
+        genesis_utils::{create_genesis_config, GenesisConfigInfo},
+        leader_schedule_cache::LeaderScheduleCache,
+    },
     solana_runtime::bank::Bank,
     solana_sdk::{
         clock::NUM_CONSECUTIVE_LEADER_SLOTS, hash::Hash, poh_config::PohConfig, pubkey::Pubkey,
@@ -155,7 +159,7 @@ pub struct PohRecorder {
     pub poh: Arc<Mutex<Poh>>,
     tick_height: u64,
     clear_bank_signal: Option<SyncSender<bool>>,
-    start_slot: Slot,              // parent slot
+    last_reset_bank: Arc<Bank>,    // parent slot
     start_tick_height: u64,        // first tick_height this recorder will observe
     tick_cache: Vec<(Entry, u64)>, // cache of entry and its tick_height
     working_bank: Option<WorkingBank>,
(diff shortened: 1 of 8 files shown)


poh/src/poh_recorder.rs
```diff
@@ -364,28 +407,28 @@ impl PohRecorder {
...
+            self.start_slot(),
+            blockhash,
+            reset_bank.slot()
         );
 
         std::mem::swap(&mut cache, &mut self.tick_cache);
 
-        self.start_slot = start_slot;
-        self.tick_height = (start_slot + 1) * self.ticks_per_slot;
+        self.start_bank = reset_bank;
```
carllin, line 430: This is one spot where we change `start_slot`  to track to `start_bank` instead

poh/src/poh_recorder.rs
```diff
@@ -395,23 +438,19 @@ impl PohRecorder {
...
 
-    pub fn set_working_bank(&mut self, working_bank: WorkingBank) {
-        trace!("new working bank");
-        assert_eq!(working_bank.bank.ticks_per_slot(), self.ticks_per_slot());
-        self.working_bank = Some(working_bank);
-        // TODO: adjust the working_bank.start time based on number of ticks
-        // that have already elapsed based on current tick height.
-        let _ = self.flush_cache(false);
-    }
-
```
carllin, line 406: This functionality was only used over `set_bank()` in tests, which have been refactored to avoid using this function. Thus this function can now be deleted

poh/src/poh_recorder.rs
```diff
@@ -462,9 +501,8 @@ impl PohRecorder {
                 working_bank.max_tick_height,
                 working_bank.bank.slot()
             );
-            let working_slot =
-                (working_bank.max_tick_height / self.ticks_per_slot).saturating_sub(1);
-            self.start_slot = working_slot;
+            self.start_bank = working_bank.bank.clone();
+            let working_slot = self.start_slot();
```
carllin, line 505: This is equivalent to the original code because:

1) `working_bank` is only ever set by `self.set_bank()`, in which `working_bank.max_tick_height` is set to `working_bank.bank.max_tick_height()`: 
https://github.com/solana-labs/solana/blob/857fce13f9fc879b8811cefacb4939dab714781f/poh/src/poh_recorder.rs#L412

2) This means because `bank.max_tick_height == (bank.slot() + 1) * ticks_per_slot`: https://github.com/solana-labs/solana/blob/master/runtime/src/bank.rs#L1282, then the old code 

```
(working_bank.max_tick_height / self.ticks_per_slot).saturating_sub(1);
```

is exactly equivalent to `working_bank.bank.slot()`

PR #19846, opened 2021-09-13
Set is_startup=true for exhaustively_free_unused_resource()

brooksprumo:
#### Problem

`exhaustively_free_unused_resource()` is only called during startup, but it calls `clean_accounts()` with `is_startup` set to false.

#### Summary of Changes

Call `clean_accounts()` inside `exhaustively_free_unused_resource()` with `is_startup` set to true.


jeffwashington (approved):
lgtm

carllin (approved):

PR #19848, opened 2021-09-13
introduce bucket map

jeffwashington:
#### Problem
This is the working decision of disk-backing we want for the on-disk accounts index.
There are several other potential project overlaps to pursue. This code may need to be generalized or otherwise optimized.
There are currently no callers of this outside of its crate.
Having this in master will allow us to easily begin calling it and improving and testing it in the system behind tests and optional configurations.

#### Summary of Changes

Fixes #

### DIFF
--- a/bucket_map/src/bucket.rs
+++ b/bucket_map/src/bucket.rs
@@ -1,5 +1,6 @@
 use crate::bucket_map::{BucketItem, BucketMapError, MaxSearch, RefCount};
-use crate::data_bucket::{BucketMapStats, DataBucket};
+use crate::bucket_stats::BucketMapStats;
+use crate::bucket_storage::BucketStorage;
 use crate::index_entry::IndexEntry;
 use rand::thread_rng;
 use rand::Rng;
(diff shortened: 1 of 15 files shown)


bucket_map/src/bucket_map.rs
```diff
@@ -23,15 +23,15 @@ pub struct BucketMapKeyValue<T> {
 
 #[derive(Debug, Default)]
 pub struct BucketMapConfig {
-    pub num_buckets: usize,
```
jeffwashington: what is your thinking on max vs num here? That possibly not all buckets are allocated?

bucket_map/src/bucket_map.rs
```diff
@@ -137,10 +135,17 @@ impl<T: Clone + Copy + Debug> BucketMap<T> {
         Some(self.buckets[ix].read().unwrap().as_ref()?.range(range))
     }
 
+    /// Get the Pubkeys for bucket `ix`
+    pub fn keys(&self, ix: usize) -> Option<Vec<Pubkey>> {
```
jeffwashington, line 140: I think we get rid of option here and just return vec![] if there is no bucket. Caller doesn't need to know bucket doesn't exist. That isn't a case in the way we intend to use it anyway. All buckets should be allocated.Same on range I think.

bucket_map/src/bucket.rs
```diff
@@ -13,7 +13,7 @@ use std::path::PathBuf;
 use std::sync::atomic::Ordering;
 use std::sync::Arc;
 
-// >= 2 instances of this per 'bucket' in the bucket map. 1 for index, >= 1 for data
```
jeffwashington: I think I meant 'Bucket' here, not DataBucket

bucket_map/src/data_bucket.rs
```diff
@@ -79,7 +78,6 @@ pub struct DataBucket {
     mmap: MmapMut,
     pub cell_size: u64,
     pub capacity_pow2: u8,
```
jeffwashington, line 80: was it incorrect? I'm not seeing that at a quick glance.

brooksprumo (approved):
Looks good! I think the atomic instruction ordering in the header locking needs to be changed, but since this isn't used anywhere yet, I think it's fine to go in as-is.

PR #19849, opened 2021-09-13
clippy for test

jeffwashington:
#### Problem
clippy fires for an extra clone
#### Summary of Changes
Eliminate noise by eliminating extra clone.
Fixes #


Lichtso (commented):
Already happened in #19750.

PR #19851, opened 2021-09-13
build downstream spl bpf programs for tests

jackcmay:
#### Problem

Downstream CI tests are failing because not all the dependent bpf programs are being built

#### Summary of Changes

Build them.

Should these be built implicitly, maybe missing dep?

Fixes #

### DIFF
--- a/scripts/build-downstream-projects.sh
+++ b/scripts/build-downstream-projects.sh
@@ -68,6 +68,7 @@ spl() {
 
     $cargo build
     $cargo test
+    $cargo_build_bpf
     $cargo_test_bpf
   )
 }


CriesofCarrots:
Oh, I see, the issue is this block:
https://github.com/solana-labs/solana-program-library/blob/30406eea40671afefa18b5ae6c9b37ffcb6235d0/associated-token-account/program/tests/program_test.rs#L14-L21
If use_latest_spl_token, an spl_token binary is needed.

I'm a little confused by the comment, though, because Token v3.2.0 is already released.
@SebastianBor , can we delete that block and forgo this extra build for downstream-projects?

jackcmay:
@CriesofCarrots I'm going to commit this change for now to unblock CI, if/when @SebastianBor removes the block I'll revert.  Sound good?

CriesofCarrots:
Yep, thanks @jackcmay

SebastianBor:
> Oh, I see, the issue is this block:
> https://github.com/solana-labs/solana-program-library/blob/30406eea40671afefa18b5ae6c9b37ffcb6235d0/associated-token-account/program/tests/program_test.rs#L14-L21
> If use_latest_spl_token, an spl_token binary is needed.
> 
> I'm a little confused by the comment, though, because Token v3.2.0 is already released.
> @SebastianBor , can we delete that block and forgo this extra build for downstream-projects?

It can't be removed because the tests would fail. The code needs the latest `spl-token` version which supports `InitializeAccount3`. I checked the latest build `1.7.11` and the instruction is still not there. @joncinque do you know when it's going to be released? Is it already on-chain but not in the sdk yet?
Regardless of that it feels like a valid scenario when a program depends on unreleased version of the `spl-token` program,  or any other program, and hence the build should support this setup?

joncinque:
@SebastianBor I don't think anyone's started on the release, so it isn't *anywhere*.  Considering it's only new instructions, and no behavior change, we should be able to release it relatively quickly.  I could get started on it later this week once stake-o-matic is humming along.

If you need it sooner than that in testing environments, I can walk you through the steps.  Let's document the release process this time, since it's mainly in a couple of heads at the moment.

SebastianBor:
> @SebastianBor I don't think anyone's started on the release, so it isn't _anywhere_. Considering it's only new instructions, and no behavior change, we should be able to release it relatively quickly. I could get started on it later this week once stake-o-matic is humming along.
> 
> If you need it sooner than that in testing environments, I can walk you through the steps. Let's document the release process this time, since it's mainly in a couple of heads at the moment.

There is no rush from my side

CriesofCarrots:
Ah, should we update the comment, then? Sounds like those tests actually need Token v3.3.0
Also, release docs are in the readme: https://github.com/solana-labs/solana-program-library/#release-process
That would be awesome if you update those!

jackcmay:
I agree with sebastian that this kind of case (running a bpf based test) will come up again.  I think the better fix here is for `test-bpf` to also `build-bpf` its dependencies but I think that's non-trivial from a dependency perspective.

I'm going to go ahead and backport this to v1.7 since the same issue resides there too, any objections?

SebastianBor:
> Ah, should we update the comment, then? Sounds like those tests actually need Token v3.3.0
> Also, release docs are in the readme: https://github.com/solana-labs/solana-program-library/#release-process
> That would be awesome if you update those!

Since I don't know what exact version is going to be released (major/minor bump?) I've updated the comments with `>3.2.0` comment

SebastianBor:
> I agree with sebastian that this kind of case (running a bpf based test) will come up again. I think the better fix here is for `test-bpf` to also `build-bpf` its dependencies but I think that's non-trivial from a dependency perspective.

Can we publish the current version (master branch) of `spl-token` to crates.io and use it instead of the local dependency? If I understand the problem correctly it should solve this issue with building dependencies or I'm missing something here?

PR #19852, opened 2021-09-13
add <T> through accounts index

jeffwashington:
#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/accounts_index.rs
+++ b/runtime/src/accounts_index.rs
@@ -48,7 +48,9 @@ pub type AccountMap<V> = InMemAccountsIndex<V>;
 
 pub(crate) type AccountMapEntry<T> = Arc<AccountMapEntryInner<T>>;
 
-pub trait IsCached: 'static + Clone + Debug + PartialEq + ZeroLamport + Copy + Default {
+pub trait IsCached:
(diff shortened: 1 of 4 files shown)


runtime/src/accounts_index.rs
```diff
@@ -48,7 +48,7 @@ pub type AccountMap<V> = InMemAccountsIndex<V>;
 
 pub(crate) type AccountMapEntry<T> = Arc<AccountMapEntryInner<T>>;
 
-pub trait IsCached: 'static + Clone + Debug + PartialEq + ZeroLamport + Copy + Default {
+pub trait IsCached: 'static + Clone + Debug + PartialEq + ZeroLamport + Copy + Default + Sync + Send {
```
brooksprumo, line 51: This trait has grown quite a bit. Maybe in the next round `IsCached` can go back to just the `is_cached()` function, and a new trait can be added for this composite?

brooksprumo (commented):
Can you re-request the review once this PR is rebased on top of the 'per bin stats' commit?

runtime/src/accounts_index.rs
jeffwashington: sure, seems reasonable

brooksprumo (approved):
lgtm

PR #19853, opened 2021-09-13
Cli: check current authorities before attempting to change them

CriesofCarrots:
#### Problem
Submitting the same solana-cli checked commands repeatedly succeeds unexpectedly. For example:
```
$ solana stake-authorize-checked ~/stake.json --withdraw-authority ~/current_authority.json --new-withdraw-authority ~/new_authority.json
```
will succeed indefinitely. This is because the stake program checks that the current authority according to the program and the new authority according to the instruction are both signers, and they are -- the same `~/new_authority.json`.

`solana stake-set-lockup-checked`, `solana vote-authorize-voter-checked` and `solana-authorize-withdrawer-checked` evidence the same behavior.

#### Summary of Changes
When not in offline mode, check the stake or vote account ahead of time to ensure the provided current authority is consistent with the account

### DIFF
--- a/cli/src/cli.rs
+++ b/cli/src/cli.rs
@@ -1973,17 +1973,36 @@ mod tests {
         let result = process_command(&config);
         assert!(result.is_ok());
 
+        let vote_account_info_response = json!(Response {
+            context: RpcResponseContext { slot: 1 },
+            value: json!({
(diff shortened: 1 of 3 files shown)


mvines (approved):
lgtm once CI is happy

PR #19854, opened 2021-09-14
feat: getBlocks API is added

kamaci:
#### Problem

Documentation mentions getBlocks API but it is missing at web3.js

#### Summary of Changes

Implemented getBlocks API

### DIFF
--- a/web3.js/src/connection.ts
+++ b/web3.js/src/connection.ts
@@ -3072,6 +3072,26 @@ export class Connection {
   }
 
   /**
+   * Fetch confirmed blocks between two slots
+   */
+  async getBlocks(
+    startSlot: number,
+    endSlot?: number,
+    commitment?: Finality,
+  ): Promise<Array<number>> {
+    const args = this._buildArgsAtLeastConfirmed(
+      endSlot !== undefined ? [startSlot, endSlot] : [startSlot],
+      commitment,
+    );
+    const unsafeRes = await this._rpcRequest('getBlocks', args);
+    const res = create(unsafeRes, jsonRpcResult(array(number())));
+    if ('error' in res) {
+      throw new Error('failed to get blocks: ' + res.error.message);
+    }
+    return res.result;
+  }
+
+  /**
    * Fetch a list of Signatures from the cluster for a confirmed block, excluding rewards
    */
   async getConfirmedBlockSignatures(
--- a/web3.js/test/connection.test.ts
+++ b/web3.js/test/connection.test.ts
@@ -2077,6 +2077,54 @@ describe('Connection', () => {
     );
   });
 
+  it('get blocks between two slots', async () => {
+    await mockRpcResponse({
+      method: 'getBlocks',
+      params: [0, 10],
+      value: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10],
+    });
+
+    await mockRpcResponse({
+      method: 'getSlot',
+      params: [],
+      value: 10,
+    });
+
+    const latestSlot = await connection.getSlot();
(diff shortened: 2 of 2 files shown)


web3.js/src/connection.ts
```diff
@@ -3071,6 +3071,26 @@ export class Connection {
...
+  /**
+   * Fetch confirmed blocks between two slots
+   */
+  async getBlocks(
+    startSlot: number,
+    endSlot?: number,
+    commitment?: Finality,
+  ): Promise<Array<number>> {
+    const args = this._buildArgsAtLeastConfirmed(
+      endSlot ? [startSlot, endSlot] : [startSlot],
```
jstarry, line 3083:
```suggestion
      endSlot !== undefined ? [startSlot, endSlot] : [startSlot],
```

kamaci: I've changed that line 👍

jstarry:
@kamaci almost there! can you fix your test please? The tests are run against a local validator that starts from slot 0 so your test won't work

kamaci:
> @kamaci almost there! can you fix your test please? The tests are run against a local validator that starts from slot 0 so your test won't work

I've updated the PR.

kamaci:
Is there anything left to do?

kamaci:
> @kamaci almost there! can you fix your test please? The tests are run against a local validator that starts from slot 0 so your test won't work

Fixed the test

jstarry:
still not fixed

jstarry:
please run locally before submitting a new commit

web3.js/test/connection.test.ts
```diff
@@ -2077,6 +2077,36 @@ describe('Connection', () => {
...
 
+  it('get blocks between two slots', async () => {
+    await mockRpcResponse({
+      method: 'getBlocks',
+      params: [0, 10],
+      value: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10],
+    });
+
+    const blocks = await connection.getBlocks(0, 10);
+    expect(blocks).to.have.length(10);
```
jstarry, line 2088: This could still fail... how about you first request `connection.getSlot` to see what slots are expected and ensure that we have at least that many?

kamaci: Actually I tested it locally and it used to have 42 blocks but it is 24 now. However, I'll implement what you suggested.

kamaci: Hope this change is OK :)

jstarry: Looks great, could you apply the same fix to the other test please?

kamaci: Btw, I have this:

`expect(blocks).to.have.lengthOf.greaterThanOrEqual(latestSlot);`

However there may be less than last slot number slots, right? So, I need to change that line.

kamaci: Also, another test at connection.test.ts has these lines:

```
    while ((await connection.getSlot()) <= 0) {
      continue;
    }
```

should I follow the same approach?

jstarry: > However there may be less than last slot number slots, right? So, I need to change that line.

In this case, with one validator, there should always be a block for every slot

> should I follow the same approach?

Sure, sounds great!

kamaci: I used to think that block numbers may not be sequential.

jstarry:
Looks great, thanks for this change!

kamaci:
> Looks great, thanks for this change!

Thanks!

PR #19865, opened 2021-09-14
Simplify ed25519 instruction index

seanyoung:
Allow u16::MAX to be specified for the instruction index. This makes it
possible to specify the current instruction, so it is not necessary to
know the instruction number.

#### Summary of Changes

Fixes #


seanyoung:
@jstarry suggested that we improved the ed25519 program so that it accepts `u16::MAX` as an instruction index which means "current instruction". This simplifies instruction construction (both rust and js side).

I've implemented the suggested change here. What do you think @mvines @jackcmay ? I don't have strong opinions on this, as @jstarry points out this is a better devx experience.

jackcmay:
Fine by me, do we also want to apply this to secp256k1?

PR #19876, opened 2021-09-14
Add test for startup processing new roots past full snapshot interval

### LINKED ISSUE #19856: [Incremental Snapshots] Add local-cluster test when startup's blockstore has roots that cross the full snapshot interval

brooksprumo:
#### Problem

There's no test for the scenario that's in the title.

#### Summary of Changes

Here's the comment I added for the new test:

```
/// Test the scenario where a node starts up from a snapshot and its blockstore has enough new
/// roots that cross the full snapshot interval.  In this scenario, the node needs to take a full
/// snapshot while processing the blockstore so that once the background services start up, there
/// is the correct full snapshot available to take subsequent incremental snapshots.
```

Fixes #19856

### DIFF
--- a/local-cluster/tests/local_cluster.rs
+++ b/local-cluster/tests/local_cluster.rs
@@ -1928,19 +1928,6 @@ fn test_incremental_snapshot_download_with_crossing_full_snapshot_interval_at_st
 
     let mut cluster = LocalCluster::new(&mut config, SocketAddrSpace::Unspecified);
 
-    let leader_snapshot_archives_dir = &leader_snapshot_test_config
-        .validator_config
-        .snapshot_config
-        .as_ref()
-        .unwrap()
-        .snapshot_archives_dir;
-    let validator_snapshot_archives_dir = &validator_snapshot_test_config
-        .validator_config
-        .snapshot_config
-        .as_ref()
-        .unwrap()
-        .snapshot_archives_dir;
-
     debug!("snapshot config:\n\tfull snapshot interval: {}\n\tincremental snapshot interval: {}\n\taccounts hash interval: {}",
            full_snapshot_interval,
            incremental_snapshot_interval,
@@ -1970,7 +1957,10 @@ fn test_incremental_snapshot_download_with_crossing_full_snapshot_interval_at_st
 
     info!("Waiting for leader to create snapshots...");
     let (incremental_snapshot_archive_info, full_snapshot_archive_info) =
-        LocalCluster::wait_for_next_incremental_snapshot(&cluster, leader_snapshot_archives_dir);
+        LocalCluster::wait_for_next_incremental_snapshot(
+            &cluster,
+            leader_snapshot_test_config.snapshot_archives_dir.path(),
+        );
     debug!(
         "Found snapshots:\n\tfull snapshot: {}\n\tincremental snapshot: {}",
         full_snapshot_archive_info.path().display(),
@@ -1985,7 +1975,7 @@ fn test_incremental_snapshot_download_with_crossing_full_snapshot_interval_at_st
     info!("Downloading full snapshot to validator...");
     download_snapshot_archive(
         &cluster.entry_point_info.rpc,
-        validator_snapshot_archives_dir,
+        validator_snapshot_test_config.snapshot_archives_dir.path(),
         (
             full_snapshot_archive_info.slot(),
             *full_snapshot_archive_info.hash(),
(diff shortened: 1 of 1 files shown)


local-cluster/tests/local_cluster.rs
```diff
@@ -1876,6 +1876,359 @@ fn test_incremental_snapshot_download() {
...
+            let from_file_path = entry.path();
+            let to_file_path = to.join(from_file_path.file_name().unwrap());
+            trace!(
+                "\t\tcopying file from {} to {}...",
+                from_file_path.display(),
+                to_file_path.display()
+            );
+            fs::copy(from_file_path, to_file_path).unwrap();
+        }
+    };
```
carllin, line 2079: Might be able to use this utility: https://github.com/solana-labs/solana/blob/d0633c5547e4fd7a3099c45aedc6003397ff7db0/core/tests/snapshots.rs#L459-L460

local-cluster/tests/local_cluster.rs
```diff
@@ -1876,6 +1876,359 @@ fn test_incremental_snapshot_download() {
...
+
+    info!("Downloading incremental snapshot to validator...");
+    download_snapshot_archive(
+        &cluster.entry_point_info.rpc,
+        validator_snapshot_archives_dir,
+        (
+            incremental_snapshot_archive_info.slot(),
+            *incremental_snapshot_archive_info.hash(),
+        ),
+        SnapshotType::IncrementalSnapshot(incremental_snapshot_archive_info.base_slot()),
```
carllin, line 2025: Clarification: Where does the validator get the corresponding full snapshot for this incremental snapshot?

local-cluster/tests/local_cluster.rs
```diff
@@ -1876,6 +1876,359 @@ fn test_incremental_snapshot_download() {
...
+    // To ensure that a snapshot will be taken during startup, the blockstore needs to have roots
+    // that cross a full snapshot interval.
+    info!("Waiting for the validator to see enough slots to cross a full snapshot interval...");
+    let starting_slot = incremental_snapshot_archive_info.slot();
+    let timer = Instant::now();
+    loop {
+        let validator_current_slot = cluster
+            .get_validator_client(&validator_identity.pubkey())
+            .unwrap()
+            .get_slot_with_commitment(CommitmentConfig::finalized())
```
carllin, line 2128: nice! It's subtle here that we need this to be `finalized()`, and not simply `confirmed()`

local-cluster/tests/local_cluster.rs
```diff
@@ -1876,6 +1876,359 @@ fn test_incremental_snapshot_download() {
...
+                    info!("Success! Made new full and incremental snapshots!");
+                    trace!(
+                        "Full snapshot slot: {}, incremental snapshot slot: {}",
+                        highest_full_snapshot_slot,
+                        highest_incremental_snapshot_slot
+                    );
+                    break;
+                }
+            }
+        }
```
carllin, line 2190: Wondering if there's a way to verify that the hash of the snapshot that was created at startup here matches the snapshot on the leader slot.  That way we can verify the snapshot was correctly taken.

If we didn't purge the old bank snapshots on the leader, then maybe this would be possible since we could search its snapshot directory for such a snapshot

Unfortunately we do purge here: https://github.com/solana-labs/solana/blob/d0633c5547e4fd7a3099c45aedc6003397ff7db0/runtime/src/snapshot_utils.rs#L1567 😛, hmmm....

local-cluster/tests/local_cluster.rs
brooksprumo: Yah! So when the leader calls `wait_for_next_incremental_snapshot()`, it returns the tuple of full and incremental snapshots (SnapshotArchiveInfo). Since we now know the specific snapshots to download, the validator can they call `download_snapshot_archive()` and use the specific snapshots that the leader just said it created.

The incremental snapshot that `wait_for_next_incremental_snapshot()` returns is required to be based on the full snapshot it also returns; those two will never be out of sync.

And since the leader says "hey, I created this full & incremental snapshot", the validator can turn around and say "hey, I'd like to download those same full & incremental snapshots".

local-cluster/tests/local_cluster.rs
brooksprumo: Done! Unfortunately `fs_extra` doesn't have a function that's equivalent to my `delete_files()`, so that needs to stay.

local-cluster/tests/local_cluster.rs
brooksprumo: Yeah, this is a good test to add. Done!

local-cluster/tests/local_cluster.rs
```diff
@@ -1876,6 +1877,403 @@ fn test_incremental_snapshot_download() {
...
+    let (
+        validator_highest_full_snapshot_archive_info,
+        _validator_highest_incremental_snapshot_archive_info,
+    ) = loop {
+        if let Some(highest_full_snapshot_info) =
+            snapshot_utils::get_highest_full_snapshot_archive_info(
+                validator_snapshot_test_config.snapshot_archives_dir.path(),
+            )
+        {
+            if highest_full_snapshot_info.slot() > validator_previous_highest_full_snapshot_slot {
```
carllin, line 2201: Should we be waiting for the highest full snapshot slot to be equal to the `validator_previous_highest_full_snapshot_slot `: https://github.com/solana-labs/solana/blob/a23a0634da309efcd9702921e914faf4b8d9e3d1/local-cluster/tests/local_cluster.rs#L2175 to prevent races where `validator_previous_highest_full_snapshot_slot` is actually more than one full snapshot ahead?

carllin (approved):

PR #19879, opened 2021-09-14
introduce CachedBucket in accounts index model

jeffwashington:
#### Problem
The hashmap itself will move to CachedBucket.
It will also be behind a RwLock.
This gets messy when combined.
The purpose of this change is to create the right struct hierarchy that matches the locking hierarchy and prepares the way for the right pieces of data to be owned in the right code cleanly.

#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/bucket_map_holder.rs
+++ b/runtime/src/bucket_map_holder.rs
@@ -1,21 +1,43 @@
 use crate::accounts_index::IndexValue;
 use crate::bucket_map_holder_stats::BucketMapHolderStats;
 use crate::waitable_condvar::WaitableCondvar;
-use std::fmt::Debug;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 use std::time::Duration;
 
 // will eventually hold the bucket map
-#[derive(Debug, Default)]
+#[derive(Default)]
 pub struct BucketMapHolder<T: IndexValue> {
-    pub stats: BucketMapHolderStats,
+    pub stats: Arc<BucketMapHolderStats>,
+    pub buckets: Vec<Arc<CachedBucket<T>>>,
+}
+
+#[derive(Default)]
+pub struct CachedBucket<T: IndexValue> {
+    // soon, this will hold the hashmap itself: pub map: RwLock<HashMap<Pubkey, AccountMapEntry<T>>>,
+    _bin: usize,
+    _stats: Arc<BucketMapHolderStats>,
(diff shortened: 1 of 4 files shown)


runtime/src/bucket_map_holder.rs
```diff
@@ -9,13 +9,37 @@ use std::time::Duration;
...
-        Self::default()
+    pub fn new() -> Arc<Self> {
+        let bins = 0;
+        let stats = Arc::new(BucketMapHolderStats::default());
+        let buckets = (0..bins)
+            .into_iter()
+            .map(|bin| Arc::new(CachedBucket::new(stats.clone(), bin)))
+            .collect();
+        Arc::new(Self { stats, buckets })
     }
```
brooksprumo, line 43: As a user, I would be surprised if I called `new()` on a type `T` and it didn't return `T`, but instead `Arc<T>`. Does the Arc need to be here?

runtime/src/in_mem_accounts_index.rs
```diff
@@ -21,6 +21,7 @@ pub struct InMemAccountsIndex<T: IsCached> {
     map: HashMap<Pubkey, AccountMapEntry<T>>,
     storage: Arc<BucketMapHolder<T>>,
     bin: usize,
+    bucket: Arc<CachedBucket<T>>,
```
brooksprumo, line 24: It looks like the bucket is basically `storage[bin]`, yes? (or similar, based on `InMemAccountsIndex::new()`.) I assume adding this bucket param makes life easier?

brooksprumo (commented):
A few questions. Maybe they don't really matter much for the PRs that follow on after this, but figured I'd ask.

runtime/src/bucket_map_holder.rs
jeffwashington: Yes, I can clean that up. I changed direction and that was an artifact left over.

runtime/src/in_mem_accounts_index.rs
jeffwashington: Yes, I want a direct call to it. Probably will be reworked more. Soon, this struct won't use that storage at all.

brooksprumo (approved):
lgtm

jeffwashington:
@brooksprumo fyi I figured out I don't actually need this layer of indirection. I went a different way. Thankfully, I suppose, these prs have been fighting the ci for so many hours...
#19904 looks like a better solution. I can share an Arc to these as the buckets themselves. So, we're already closer to the desired end state.

PR #19884, opened 2021-09-14
Banking_stage prioritize vote transactions

taozhu-chicago:
#### Problem
Validator votes can be "blocked" by other transactions. 

#### Summary of Changes
- add QoS to banking_stage to pick up votes first

Fixes #


core/src/banking_stage.rs
```diff
@@ -1146,8 +1214,19 @@ impl BankingStage {
...
+        // process all votes in the pipeline first
+        Self::process_packets_vote(
+            &mut mms,
+            &mut vote_tx_count,
+            poh,
+            recorder,
+            transaction_status_sender.clone(),
+            gossip_vote_sender,
+        );
+
```
taozhu-chicago, line 1229: @sakridge maybe can add `vote-only` feature at here. This is the point where votes in pipeline have been processed, and about to process/forward other txs in received packets. It could check `bank.vote_only_bank()` and skip next step if necessary. wdyt?

sakridge: yea, that seems ok.

sakridge: That is reasonable. I'm starting to be a bit worried about the code duplication we introduce in this way. We'll have to have this process_votes on the buffered packet processing as well, right? And we would deserialize everything twice except votes.

Also, do we need some of the other checks that we make in packets_from_transactions?

core/src/banking_stage.rs
```diff
@@ -1114,6 +1121,67 @@ impl BankingStage {
             .collect()
     }
 
+    fn generate_vote_packet_indexes(vers: &PinnedVec<Packet>) -> Vec<usize> {
+        vers.iter()
+            .enumerate()
+            .filter_map(|(index, ver)| {
+                if !ver.meta.discard {
+                    let tx = limited_deserialize(&ver.data[0..ver.meta.size]);
+                    if tx.is_ok()
```
sakridge, line 1130: Can it just be this:
`let tx = limited_deserialize(&ver.data[0..ver.meta.size]).ok()?;`

core/src/banking_stage.rs
taozhu-chicago: Yeah, the duplication is a concern, folding vote-only logic here avoids some. I'm copying your code from #19916 to here. 

Not sure if necessary to do `process_packets_vote` again on `process_buffered_packets` again. Once votes are processed at point of receiving, they will not being put into buffer, except when `! get_bank_still_processing_txs`. Not sure how often this happens, will add a metric to check it out.

core/src/banking_stage.rs
taozhu-chicago: done! thanks

core/src/banking_stage.rs
sakridge: Yes, they can be received prior to the validator being the leader.

sakridge: So then it will process the buffered ones first. I think we need to handle both those pathways.

taozhu-chicago: Gotcha, that explains the metric. Receiving packets before becoming leader is solid need for apply same QoS on buffered. Thanks for pointing it out!

core/src/banking_stage.rs
```diff
@@ -1190,6 +1216,71 @@ impl BankingStage {
...
+        unprocessed_vote_tx_count: &mut usize,
+        poh: &Arc<Mutex<PohRecorder>>,
+        recorder: &TransactionRecorder,
+        transaction_status_sender: Option<TransactionStatusSender>,
+        gossip_vote_sender: &ReplayVoteSender,
+        banking_stage_stats: &BankingStageStats,
+    ) {
+        let mms_iter = mms.iter_mut();
+        for msgs in mms_iter {
+            let bank_start = poh.lock().unwrap().bank_start();
```
carllin, line 1250: I think it's safe to move this lock above outside of the `mms_iter` loop.  If we hit the end of a slot, we can then early exit and have the fallback logic handle it

core/src/replay_stage.rs
```diff
@@ -2515,9 +2519,10 @@ impl ReplayStage {
         root_slot: u64,
         leader: &Pubkey,
         subscriptions: &Arc<RpcSubscriptions>,
+        vote_only_bank: bool,
     ) -> Bank {
         subscriptions.notify_slot(slot, parent.slot(), root_slot);
-        Bank::new_from_parent(parent, leader, slot)
+        Bank::new_from_parent_with_vote_only(parent, leader, slot, vote_only_bank)
```
carllin, line 2525: might be good to log a warning and emit a datapoint whenever this happens so that we can track it in the metrics

core/src/banking_stage.rs
```diff
@@ -1190,6 +1216,71 @@ impl BankingStage {
...
+            if PohRecorder::get_bank_still_processing_txs(&bank_start).is_none() {
+                // do noting here, the main loop will take care of unprocessed packets
+                *skip_process_packets_vote_count += 1;
+                continue;
+            }
+            let (bank, bank_creation_time) = bank_start.unwrap();
+
+            let packet_indexes = Self::generate_vote_packet_indexes(&msgs.packets);
+            let (processed, _verified_txs_len, unprocessed_indexes) =
```
carllin, line 1259: So if this fails due to something like lock contention, does the vote just enter the regular processing pipeline again?

core/src/banking_stage.rs
```diff
@@ -1190,6 +1216,71 @@ impl BankingStage {
...
+                        None
+                    }
+                } else {
+                    None
+                }
+            })
+            .collect()
+    }
+
+    fn process_packets_vote(
```
carllin, line 1237: Maybe a basic unit test to test it correctly 
1) parses out vote transactions from a group
2) Runs the proper sanity checks

carllin:
Hmmm this seems like a good bandaid, but ultimately I think we want a separate vote queue and set of servicing voting threads. Some benefits seem like:

1) Don't affect regular transaction processing if the votes conflict
2) Won't block vote transactions behind expensive non-vote transactions. For instance here https://github.com/solana-labs/solana/blob/7440ee434483977376c2b63a7a89884ff97c2575/core/src/banking_stage.rs#L1302 if there's an expensive batch of non-votes, the vote transactions will be blocked again
3) Sigverify/intermediate stage can own this logic https://github.com/solana-labs/solana/blob/7440ee434483977376c2b63a7a89884ff97c2575/core/src/banking_stage.rs#L1219-L1235 for identifying vote transactions, and directly forward to those vote threads
4) Vote threads can own any vote-specific ordering/forwarding logic

core/src/banking_stage.rs
taozhu-chicago: yep, good call!

core/src/replay_stage.rs
taozhu-chicago: added counter `replay_stage-new_vote_only_bank` for each vote-only bank creation, another counter `banking_stage-loop-stats.vote_only_bank_dropped_tx_count` tracks the number of non-vote transactions dropped due to vote-only

core/src/banking_stage.rs
taozhu-chicago: this is the case for now, I was bit concerned about its likelihood, added counter `banking_stage-loop-stats::unprocessed_vote_tx_count`, monitored it over few bench-tps tests, it is always zero. Feels like this is an unlikely event. 

I also don't have a clear path for handling unprocessed votes, maybe can put them on top of buffered packets, feels like a lot of changes, not super comfortable about it. 

image

taozhu-chicago:
> Hmmm this seems like a good bandaid, but ultimately I think we want a separate vote queue and set of servicing voting threads. Some benefits seem like:
> 
> 1. Don't affect regular transaction processing if the votes conflict
> 2. Won't block vote transactions behind expensive non-vote transactions. For instance here https://github.com/solana-labs/solana/blob/7440ee434483977376c2b63a7a89884ff97c2575/core/src/banking_stage.rs#L1302
>     if there's an expensive batch of non-votes, the vote transactions will be blocked again
> 3. Sigverify/intermediate stage can own this logic https://github.com/solana-labs/solana/blob/7440ee434483977376c2b63a7a89884ff97c2575/core/src/banking_stage.rs#L1219-L1235
>     for identifying vote transactions, and directly forward to those vote threads
> 4. Vote threads can own any vote-specific ordering/forwarding logic

This sounds good idea, the vote servicing threads can also service gossip votes. Directing votes to dedicated threads instead of mixing with transactions, we are providing a good separation between cluster stability and its functionality. Perhaps creating an issue for this topic?

carllin:
> > Hmmm this seems like a good bandaid, but ultimately I think we want a separate vote queue and set of servicing voting threads. Some benefits seem like:
> > 
> > 1. Don't affect regular transaction processing if the votes conflict
> > 2. Won't block vote transactions behind expensive non-vote transactions. For instance here https://github.com/solana-labs/solana/blob/7440ee434483977376c2b63a7a89884ff97c2575/core/src/banking_stage.rs#L1302
> >    
> >    if there's an expensive batch of non-votes, the vote transactions will be blocked again
> > 3. Sigverify/intermediate stage can own this logic https://github.com/solana-labs/solana/blob/7440ee434483977376c2b63a7a89884ff97c2575/core/src/banking_stage.rs#L1219-L1235
> >    
> >    for identifying vote transactions, and directly forward to those vote threads
> > 4. Vote threads can own any vote-specific ordering/forwarding logic
> 
> This sounds good idea, the vote servicing threads can also service gossip votes. Directing votes to dedicated threads instead of mixing with transactions, we are providing a good separation between cluster stability and its functionality. Perhaps creating an issue for this topic?

> the vote servicing threads can also service gossip votes

Yup, gossip would just forward to these vote threads as well


> Perhaps creating an issue for this topic
Done! https://github.com/solana-labs/solana/issues/19971

core/src/banking_stage.rs
taozhu-chicago: unit tested

core/src/replay_stage.rs
```diff
@@ -1227,12 +1227,15 @@ impl ReplayStage {
...
 
+            let root_distance = poh_slot - root_slot;
+
             let tpu_bank = Self::new_bank_from_parent_with_notify(
                 &parent,
                 poh_slot,
                 root_slot,
                 my_pubkey,
                 subscriptions,
+                root_distance > 500,
```
carllin, line 1238: nit: move to a constant `MAX_UNTHROTTLED_ROOT_DISTANCE`

carllin: I was thinking about this a bit today, and was a little doubtful about how effective this would be since `root_slot` is local, so the `root_distance` is a short-lived, local concept. 

For instance, it's possible your root distance is < 500 because you're occasionally making roots, and then you switch to non-vote only mode, but then the rest of the cluster is not making roots.

I think this would be more effective if we had something like go into vote only mode if 

1. Our local root distance > 500, OR
2. > 1/3 of the vote accounts in the bank for `parent_slot` has a `root` where `root < poh_slot - 500`

This second one is a bit trickier to calculate, but I think can be done by computing the minimum root slot that has 1/3 of the stake here in `collect_vote_lockouts`. https://github.com/solana-labs/solana/blob/d0633c5547e4fd7a3099c45aedc6003397ff7db0/core/src/consensus.rs#L223. If we derive an ordered list of `(root, stake)` from smallest to largest, you can find the minimal root, `superminority_cluster_threshold_root` that `1/3` of the validators have. This can then be returned here: https://github.com/solana-labs/solana/blob/d0633c5547e4fd7a3099c45aedc6003397ff7db0/core/src/consensus.rs#L329-L335 and stored in the progress map, like the rest of the output here https://github.com/solana-labs/solana/blob/d0633c5547e4fd7a3099c45aedc6003397ff7db0/core/src/replay_stage.rs#L2184-L2200. 

We can then check that `superminority_cluster_threshold_root` for the parent slot here and see if we need to go into vote only mode

core/src/banking_stage.rs
```diff
@@ -1203,6 +1292,68 @@ impl BankingStage {
             .collect()
     }
 
+    fn generate_vote_packet_indexes(vers: &PinnedVec<Packet>) -> Vec<usize> {
+        vers.iter()
+            .enumerate()
+            .filter_map(|(index, ver)| {
+                if !ver.meta.discard {
+                    let tx: Transaction = limited_deserialize(&ver.data[0..ver.meta.size]).ok()?;
+                    if is_simple_vote_transaction(&tx) {
```
jstarry, line 1301: I think full deserialization is probably overkill here. I think at a minimum we want to know that the tx has:
1. only 1 instruction
2. the instruction calls the vote program

How about we do the above checks in sigverify and then tag packets like we do for `is_tracer_tx`?

jstarry: It's probably fine (and good) to prioritize non "Vote" vote program instructions through like "Init", "Authorize", etc as well

sakridge: +1 probably fine to prioritize other Vote instructions.
Painful to find the instructions and then the offsets manually. Basically wrote our own custom Transaction deserializer at that point. But it probably would be a lot faster especially with a large Transaction with a bunch of Vecs to allocate and copy.

jstarry: Yeah it's painful, zero copy deserialization would be really nice for this cc @t-nelson

mvines: I bet we can `bytemuck` here for a partial zero copy deserialization of the parts of the transaction we need to inspect

jstarry: I think we need the fields to be byte aligned for that

mvines: 🤔 ya, dang maybe not then

taozhu-chicago: yeah, manually count offset isn't fun, but doable. Doing this right now, hope to make bench-tps better.

core/src/replay_stage.rs
taozhu-chicago: @sakridge based on the offline chats, the `bank-only` change will be in its own PR for separate release, then I should remove its related change from this PR, should I?

taozhu-chicago: > @sakridge based on the offline chats, the `bank-only` change will be in its own PR for separate release, then I should remove its related change from this PR, should I?

merged latest v1.6, all clear

core/src/banking_stage.rs
taozhu-chicago: made suggested change, improved `bench-tps`, it is now back to v1.6 level. Please review the change

perf/src/sigverify.rs
```diff
@@ -141,6 +142,62 @@ fn verify_packet(packet: &mut Packet) {
...
+        .ok_or(PacketError::InvalidLen)?;
+
+    let instructions_len_offset = pubkey_len
+        .checked_mul(size_of::<Pubkey>())
+        .and_then(|v| v.checked_add(pubkey_start))
+        .and_then(|v| v.checked_add(size_of::<Hash>()))
+        .ok_or(PacketError::InvalidLen)?;
+
+    let (instruction_len, instruction_len_size) =
+        decode_shortu16_len(&packet.data[instructions_len_offset..])
```
jstarry, line 172: Need to ensure that `instructions_len_offset < packet.meta.size` before this access

perf/src/sigverify.rs
```diff
@@ -141,6 +142,62 @@ fn verify_packet(packet: &mut Packet) {
...
+    // skip if has more than 1 instruction
+    if instruction_len != 1 {
+        return Ok(());
+    }
+
+    let instruction_start = instructions_len_offset
+        .checked_add(instruction_len_size)
+        .ok_or(PacketError::InvalidLen)?;
+
+    let (instruction_id_index, _size) = decode_shortu16_len(&packet.data[instruction_start..])
```
jstarry, line 184: 1. Need to ensure that `instruction_start < packet.meta.size` before this access
2. Instead of `instruction_id_index`, can you name it `instruction_program_id_index`?

perf/src/sigverify.rs
```diff
@@ -141,6 +142,62 @@ fn verify_packet(packet: &mut Packet) {
...
+        .map_err(|_| PacketError::InvalidLen)?;
+
+    // accounts were already verified at do_get_packet_offsets,
+    // can just get the pubkey via instruction_id_index
+    let instruction_id_start = instruction_id_index
```
jstarry, line 189: Need to ensure that `instruction_id_index < pubkey_len`

perf/src/sigverify.rs
```diff
@@ -141,6 +142,62 @@ fn verify_packet(packet: &mut Packet) {
         pubkey_start = pubkey_end;
         sig_start = sig_end;
     }
+
+    check_for_simple_vote_transaction(packet, msg_start).ok();
```
jstarry, line 146: We should actually put this in `do_get_packet_offsets` because `verify_packet` isn't called when using GPU to do sigverify. Looks like we made the same mistake for the tracer pubkey check but we can fix that later

jstarry:
Also, I think that this PR should be split into one PR that identifies vote tx packets and one that does the prioritization. Reasoning is that your banking stage prioritization changes are probably just a temporary fix until we have a separate pipeline for vote tx's and that separate pipeline could benefit by having your sigverify code to filter out non votes.

perf/src/sigverify.rs
taozhu-chicago: done

perf/src/sigverify.rs
taozhu-chicago: Good call, changed.

perf/src/sigverify.rs
taozhu-chicago: done

perf/src/sigverify.rs
taozhu-chicago: Cool, didn't know that. created issue #20018 20018

taozhu-chicago:
> Also, I think that this PR should be split into one PR that identifies vote tx packets and one that does the prioritization. Reasoning is that your banking stage prioritization changes are probably just a temporary fix until we have a separate pipeline for vote tx's and that separate pipeline could benefit by having your sigverify code to filter out non votes.

Agree that changes to digverify and changes to prioritize vote are two different issues. I will split them. Thanks for pointing it out

taozhu-chicago:
> > Also, I think that this PR should be split into one PR that identifies vote tx packets and one that does the prioritization. Reasoning is that your banking stage prioritization changes are probably just a temporary fix until we have a separate pipeline for vote tx's and that separate pipeline could benefit by having your sigverify code to filter out non votes.
> 
> Agree that changes to digverify and changes to prioritize vote are two different issues. I will split them. Thanks for pointing it out

#20021

perf/src/sigverify.rs
jstarry: I meant we can move the tracer check later. We should move the vote check now

taozhu-chicago: then, any objection to move tracer check all together?

PR #19891, opened 2021-09-14
Add logging after taking a bank snapshot

brooksprumo:
#### Problem

From @CriesofCarrots 
> When I boot test clusters from intermediate snapshots, I need to know the expected bank hash to pass to each validator in order to sync/wait-for-supermajority on the snapshot slot. To this point, I've just been using a dummy hash on the first boot attempt, and letting the validator process the snapshot to get the right hash. But this takes a long time with the huge snapshots Jeff and I are using for our simulations.
Is there a better method?

#### Summary of Changes

As a simple first step, log the bank hash when AccountsBackgroundService takes a bank snapshot. Later, using the snapshot slot (or snapshot/accounts hash), you'll be able to go back through the log to get the bank hash for that snapshot.

### DIFF
--- a/runtime/src/accounts_background_service.rs
+++ b/runtime/src/accounts_background_service.rs
@@ -217,6 +217,12 @@ impl SnapshotRequestHandler {
                     }
                 }
                 snapshot_time.stop();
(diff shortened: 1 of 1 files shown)


CriesofCarrots (approved):
Thank you for tracking down this spot for me!

PR #19886, opened 2021-09-14
Use decimals for stake % while waiting for supermajority

mvines:
With the current distribution on mainnet, most nodes have their stake reported as "0%" and that's not really helpful.  Instead:
```
[2021-09-14T21:43:27.388776591Z INFO  solana_core::validator] 49.227% of active stake visible in gossip
[2021-09-14T21:43:27.388796459Z INFO  solana_core::validator] 0.388% of active stake has the wrong shred version in gossip
[2021-09-14T21:43:27.388799204Z INFO  solana_core::validator]     0.041% - 4FC7CT5ViFeoKbSXJVUa5AdhDtxgDcCK2rmgbdvX5ueQ
[2021-09-14T21:43:27.388803592Z INFO  solana_core::validator]     0.042% - ATiuGpmy9Yny8ebiWdGBjDLMMzEzN7Nab4qn5Rj3ysdf
[2021-09-14T21:43:27.388806828Z INFO  solana_core::validator]     0.009% - 3ngMxX3DpJ98HcZAAWLLAMDt8yo3eJkMr2XKaDojAkPf
[2021-09-14T21:43:27.388811016Z INFO  solana_core::validator]     0.041% - DMJ6YN8JVm9DeUuBTVycu7VQ1Jzob7b4RpjGz3jwFyME
[2021-09-14T21:43:27.388814593Z INFO  solana_core::validator]     0.042% - 48Dactjx3zuJujtBHZ5cd2w3maKFFGsKhn1sXzU59VGK
[2021-09-14T21:43:27.388817678Z INFO  solana_core::validator]     0.045% - 94HgFvsD8zm7UXJ5KJxN4zW5nsdhZWY2LFREjfaFEHwR
[2021-09-14T21:43:27.388821145Z INFO  solana_core::validator]     0.040% - BN7DyC1srWNesVNh8fLZsPYNxi362FwUUZ85iGf8e7AH
[2021-09-14T21:43:27.388824181Z INFO  solana_core::validator]     0.044% - J2jV3gQsvX2htBXHeNStAVvMJaPe3RgNotwfav9pyS6y
[2021-09-14T21:43:27.388828870Z INFO  solana_core::validator]     0.041% - 2uJL1mGvtMHznh1r9qykxdho9tw2oZjedwXdmmuWsRzq
[2021-09-14T21:43:27.388831945Z INFO  solana_core::validator]     0.043% - DL6URBwusvYUFwEYZK4VEaoaynWSduUA9e4N7WDuZawf
[...]

PR #19892, opened 2021-09-15
Use f64 for stake math in get_stake_percent_in_gossip

CriesofCarrots:
#### Problem
Mainnet-beta has so much active_stake that active_stake * 100 overflows a u64.

#### Summary of Changes
Use floating points to divide first
Also pulls in the display changes from https://github.com/solana-labs/solana/pull/19886

Closes #19886


mvines (approved):
lgtm

PR #19893, opened 2021-09-15
Log supermajority achieved status

michaelh-laine:
#### Problem
During a cluster restart the validator waits for the WAIT_FOR_SUPERMAJORITY_THRESHOLD_PERCENT to be achieved, periodically printing the current cluster stats to the log. Once the threshold is met the node starts with no log to indicate this change of state, making it difficult to ascertain the time the node started voting or that the state has even changed without digging further.

Additionally if the node starts with the --wait-for-supermajority flag but the cluster has already met the threshold there is no indication from the log what has happened.

#### Summary of Changes
Just added an info log in the conditional that succeeds once the threshold is met in wait_for_supermajority() in validator.rs

Fixes #


CriesofCarrots:
Thanks! Sorry we didn't get this into v1.6.25, but it'll be there going forward.

PR #19916, opened 2021-09-15
Allow only votes when root distance gets too high (v1.6)

sakridge:
#### Problem

Network can have a hard time creating roots with lots of transactions in the system.

#### Summary of Changes

Filter out non-vote transactions when the root distance gets too high.

Fixes #


aeyakovenko:
Stakes and system? Time limit?

Issue #19843, opened 2021-09-13
Shouldn't be possible to invoke ed25519 and secp256k1 instructions from a program


jstarry:
#### Problem
The `ed25519` and `secp256k1` program instructions are a no-op if not invoked from the tx level and so devs could erroneously believe that verification is happening.

#### Proposed Solution
Prevent cross program invocations of these two programs

tag @jackcmay

jackcmay:
This seems more like a documentation issue than anything else, we should make a clear delineation between programs and pre-compiles

PR #19918, opened 2021-09-15
Cleanup and standardize precompiles

jackcmay:
#### Problem

Precompiled programs take the same inputs and provide the same outputs but are handled separately.  They are also handled disjointly which makes working with pre-compiled programs as a whole messy

#### Summary of Changes

- Standardize the handling of pre-compiled programs
- Centralize the list of pre-compiled programs in prep for changes to not call pre-compiled program's `process_instruction` during regular instruction processing and to also allow blocking the calls to precompiled programs via cpi (#19843)
- Standardize the error type of pre-compiled programs in prep for returning better transaction error messages (#17960)

The last two points will be addressed in follow-up PRs.  This PR should contain no behavior changes and thus does not contain any featurization.

Fixes #


jstarry (approved):
Great cleanup!

PR #19922, opened 2021-09-15
Restore feature declaration

CriesofCarrots:
#### Problem
The demote_program_write_locks got insta-activation on mainnet-beta, but without the feature declaration, v1.6 is now incompatible with v1.7 and master.

#### Summary of Changes
Restore it. When this is released and 95% of stake has adopted the version, activate feature. Compatibility will be restored on next epoch boundary.


mvines:
> When this is released, activate feature.

To be clear, we need 95% on the release with the feature still even though it's a no-op. I think that's a given but being pedantic cause I don't want MB to die because of a double yolo :)

CriesofCarrots:
> To be clear, we need 95% on the release with the feature still even though it's a no-op

Yep, absolutely. I'll update the description to reiterate this.

jstarry:
Why is the feature declaration needed?

PR #19930, opened 2021-09-15
Dont call precompiled programs

### LINKED ISSUE #19843: Shouldn't be possible to invoke ed25519 and secp256k1 instructions from a program

jackcmay:
#### Problem

Should not be able to call a precompiled program directly

#### Summary of Changes

- Prevent calling precompiled program's via CPI
- Short-circuit runtime to returning `Ok` for all precompiled programs

Fixes #19843


runtime/src/bank.rs
```diff
@@ -5287,6 +5287,12 @@ impl Bank {
             .add_program(program_id, process_instruction_with_context);
     }
 
+    /// Replace a builtin instruction processor if it already exists
+    pub fn remove_builtin(&mut self, name: &str, program_id: Pubkey) {
+        debug!("Removing program {} under {:?}", name, program_id);
```
ryoqun, line 5292: i was thinking whether to remove or not `KeccakSecp256k11111111111111111111111111111` from AccountsDB here

- what about `Ed25519SigVerify111111111111111111111111111`? (upcoming precompiled program) Should it exist in AccountsDB or not?  (consistency with currently existing `KeccakSecp256k11111111111111111111111111111`)
- maybe, it's good idea for all program_ids in a tx to be `executable` for consistent parsing/displaying (i.e. explorer)/inspecting. So, all precompile programs should exist in AccountsDB?

so, my verdict is to leave it there? then, how about explicitly commenting why we're NOT removing from AccountsDB?

program-runtime/src/instruction_processor.rs
```diff
@@ -342,6 +342,16 @@ impl InstructionProcessor {
         }
     }
 
+        /// Add a static entrypoint to intercept instructions before the dynamic loader.
```
ryoqun, line 345: nits: `/// Remove ...`

runtime/src/bank.rs
jackcmay: Recent changes will fail any tx that contains a program who's account is not executable, so I think we need to add accounts as executable for all supported precompiles

program-runtime/src/instruction_processor.rs
jackcmay: Yeah, this pr is super wip, but want feedback on the gist before spending time honing

ryoqun: kk

runtime/src/bank.rs
ryoqun: so, commenting seems to the right path. :)

jackcmay:
Hey @ryoqun Can you take a detailed look at the capitalization and bank hash-related changes to make sure I didn't mess anything up?  Thanks!

runtime/src/bank.rs
```diff
@@ -2583,6 +2584,64 @@ impl Bank {
...
+                Maybe, inconsistent program activation is detected on snapshot restore?",
+            program_id
+        );
+
+        // Add a bogus executable account, which will be loaded and ignored.
+        let (lamports, rent_epoch) =
+            self.inherit_specially_retained_account_fields(&existing_genuine_program);
+        let account = AccountSharedData::from(Account {
+            lamports,
+            owner: solana_sdk::system_program::id(),
```
ryoqun, line 2635: how about being consistent with a precompile program existing already on mainnet-beta?:

```
a$ ./target/release/solana -um account KeccakSecp256k11111111111111111111111111111

Public Key: KeccakSecp256k11111111111111111111111111111
Balance: 0.000000001 SOL
Owner: NativeLoader1111111111111111111111111111111
Executable: true
Rent Epoch: 0
Length: 17 (0x11) bytes
0000:   73 65 63 70  32 35 36 6b  31 5f 70 72  6f 67 72 61   secp256k1_progra
0010:   6d                                       
```

I mean, owner should be changed to `native_loader::id()`?

runtime/src/bank.rs
```diff
@@ -2583,6 +2584,64 @@ impl Bank {
         debug!("Added native program {} under {:?}", name, program_id);
     }
 
+    /// Add a precompiled program
+    pub fn add_precompiled_program(&mut self, program_id: &Pubkey) {
```
ryoqun, line 2588: I think `add_native_program` isn't used for good reasons (maybe native program vs precompile program conceptual difference)...

how about adding a unit test for this new method (`add_precompiled_program`)?

runtime/src/bank.rs
ryoqun: could you comment about the exact reason we don't remove accounts from accountsdb?

programs/bpf_loader/src/syscalls.rs
```diff
@@ -2125,16 +2126,21 @@ fn check_account_infos(
...
+    invoke_context: &dyn InvokeContext,
 ) -> Result<(), EbpfError<BpfError>> {
+    #[allow(clippy::blocks_in_if_conditions)]
     if native_loader::check_id(program_id)
         || bpf_loader::check_id(program_id)
         || bpf_loader_deprecated::check_id(program_id)
         || (bpf_loader_upgradeable::check_id(program_id)
             && !(bpf_loader_upgradeable::is_upgrade_instruction(instruction_data)
                 || bpf_loader_upgradeable::is_set_authority_instruction(instruction_data)
-                || (close_upgradeable_program_accounts
```
ryoqun, line 2136: seems like this isn't bad merge conflict resolution. :)

program-runtime/src/instruction_processor.rs
```diff
@@ -335,6 +335,13 @@ impl InstructionProcessor {
         }
     }
 
+    /// Remove a program
+    pub fn remove_program(&mut self, program_id: Pubkey) {
+        if let Some(position) = self.programs.iter().position(|(key, _)| program_id == *key) {
+            self.programs.remove(position);
+        }
```
ryoqun, line 342: how about just using `self.programs.retain()`?

programs/bpf/rust/invoke/src/lib.rs
```diff
@@ -663,6 +667,12 @@ fn process_instruction(
             // reset executable account
             **(*accounts[ARGUMENT_INDEX].lamports).borrow_mut() += 1;
         }
+        TEST_CALL_PRECOMPILE => {
+            msg!("Test calling builtin from cpi");
```
ryoqun, line 671: typo: `builtin` => `precompile`?

ryoqun:
to be sure, this pr is compatible with mainnet-beta/testnet/devnet?

runtime/src/bank.rs
```diff
@@ -2583,6 +2584,64 @@ impl Bank {
...
+        let (lamports, rent_epoch) =
+            self.inherit_specially_retained_account_fields(&existing_genuine_program);
+        let account = AccountSharedData::from(Account {
+            lamports,
+            owner: solana_sdk::system_program::id(),
+            data: vec![],
+            executable: true,
+            rent_epoch,
+        });
+        self.store_account_and_update_capitalization(program_id, &account);
```
ryoqun, line 2640: looks okay :eyes:

runtime/Cargo.toml
```diff
@@ -59,6 +57,7 @@ name = "solana_runtime"
 
 [dev-dependencies]
 assert_matches = "1.5.0"
+ed25519-dalek = "=1.0.1"
```
ryoqun, line 60: version looks to be remained same as intended. :)

runtime/src/bank.rs
ryoqun: alternatively, I think enhancing add_native_program to be parameterized over precompile/native might be also desire to avoid code duplication...

runtime/src/bank.rs
jackcmay: Since the `KeccakSecp256k11111111111111111111111111111` account already exists on mainnet beta so  it will not be recreated, and therefore remain owned by the native loader.  Once the feature is activated it won't matter because precompiles only need an account that is executable

programs/bpf_loader/src/syscalls.rs
jackcmay: Nah, just moved from a passed in bool to calling `invoke_context.is_feature_active()` directly.

program-runtime/src/instruction_processor.rs
jackcmay: Could, the list is guaranteed to not contain duplicates and the current method only visits until the program is found, rather than visiting all members in the list.

runtime/Cargo.toml
jackcmay: Not sure what you mean, is there an action item here?

jackcmay:
> to be sure, this pr is compatible with mainnet-beta/testnet/devnet?

That's the question I was hoping you could help verify with a second set of eyes :-)

runtime/src/bank.rs
jackcmay: @ryoqun I think trying to combine those two functions would get too ugly and hard to read.  Instead I've pulled out some common code and cleaned them up.  Let me know what you think.  I also added more tests.

runtime/src/bank.rs
jackcmay: No reason in this case, I was originally thinking that the precompile would be added and then the builtin of the same name would be removed.  I've added back the remove of the builtin account.

program-runtime/src/instruction_processor.rs
ryoqun: i see.

runtime/Cargo.toml
ryoqun: there is no action needed. my above comment is strictly just a random comment. I tend to write something verbosely. in this case, I left a comment to indicate i checked there's no accidental version bump of underlying crate (which could potentially create cluster incompatibility).

runtime/src/bank.rs
ryoqun: got it

runtime/src/message_processor.rs
```diff
@@ -502,6 +504,14 @@ impl MessageProcessor {
             let mut time = Measure::start("execute_instruction");
             let pre_remaining_units = compute_meter.borrow().get_remaining();
 
+            let program_id = instruction.program_id(&message.account_keys);
+            if invoke_context.is_feature_active(&prevent_calling_precompiles_as_programs::id())
+                && is_precompile(program_id, |id| invoke_context.is_feature_active(id))
+            {
+                // Precompiled programs don't have an instruction processor
+                return Ok(());
```
ryoqun, line 512: hmm, maybe this should be `continue`? (at this moment, adding a failing unit test would be desirable...)

Is it ok to bail out early from `message`'s instruction processing? I mean, following instructions should be executed. and they could fail. in that case, we should revert all side-effects by preceding instruction so far?

programs/bpf/tests/programs.rs
```diff
@@ -812,7 +812,8 @@ fn test_program_bpf_invoke_sanity() {
     const TEST_WRITABLE_DEESCALATION_WRITABLE: u8 = 14;
     const TEST_NESTED_INVOKE_TOO_DEEP: u8 = 15;
     const TEST_EXECUTABLE_LAMPORTS: u8 = 16;
-    const TEST_RETURN_DATA_TOO_LARGE: u8 = 18;
+    const TEST_CALL_PRECOMPILE: u8 = 17;
+    const TEST_RETURN_DATA_TOO_LARGE: u8 = 19;
```
ryoqun, line 816: is there any reason, the number bumped from 17 => 19? why can't `TEST_RETURN_DATA_TOO_LARGE` be `18`?

programs/bpf/rust/invoke/src/lib.rs
```diff
@@ -32,7 +33,8 @@ const TEST_PRIVILEGE_DEESCALATION_ESCALATION_WRITABLE: u8 = 13;
 const TEST_WRITABLE_DEESCALATION_WRITABLE: u8 = 14;
 const TEST_NESTED_INVOKE_TOO_DEEP: u8 = 15;
 const TEST_EXECUTABLE_LAMPORTS: u8 = 16;
-const ADD_LAMPORTS: u8 = 17;
+const TEST_CALL_PRECOMPILE: u8 = 17;
+const ADD_LAMPORTS: u8 = 18;
```
ryoqun, line 37: here

programs/bpf/tests/programs.rs
ryoqun: oh, maybe because of this? https://github.com/solana-labs/solana/pull/19930/files#r717858614

programs/bpf/rust/invoke/src/lib.rs
```diff
@@ -45,6 +47,8 @@ const DERIVED_KEY2_INDEX: usize = 7;
 const DERIVED_KEY3_INDEX: usize = 8;
 const SYSTEM_PROGRAM_INDEX: usize = 9;
 const FROM_INDEX: usize = 10;
+const ED25519_PROGRAM_INDEX: usize = 11;
+// const INVOKE_PROGRAM_INDEX: usize = 12;
```
ryoqun, line 51: how about being explicit of the reason of commenting-out?:

`// const INVOKE_PROGRAM_INDEX: usize = 12; // not used currently` 

Otherwise, it's a bit hard to tell if this is accidental (like some debugging leftover) or intended addition..

runtime/src/message_processor.rs
jackcmay: Good catch, that was a merge error, that block used to be in a function that just processed an instruction.  I'l add a test to catch it too.

programs/bpf/tests/programs.rs
jackcmay: no reason, just not ordered, I'll order it to make it loo pretty :-)

programs/bpf/c/src/invoke/invoke.c
```diff
@@ -588,6 +591,16 @@ extern uint64_t entrypoint(const uint8_t *input) {
     *accounts[ARGUMENT_INDEX].lamports += 1;
     break;
   }
+  case TEST_CALL_PRECOMPILE: {
+    sol_log("Test calling builtin from cpi");
```
ryoqun, line 595: nit: `builtin` => `precompiled program` like the rust counterpart?

programs/bpf/tests/programs.rs
jackcmay: Hmm, @seanyoung , looks like maybe the data too large test case was never added to the test program so the test below passes due to an unrelated failure?
https://github.com/solana-labs/solana/commit/098585234d0051c080f2639088b924be340573d3

programs/bpf/rust/invoke/src/lib.rs
jackcmay: Same question for @seanyoung as above

ryoqun:
> > to be sure, this pr is compatible with mainnet-beta/testnet/devnet?
> 
> That's the question I was hoping you could help verify with a second set of eyes :-)

I stumbled on devnet check (testnet, mainnet-beta isn't done still): 

it seems master got incompatible?

```
arguments: --identity validator-keypair-devnet.json --known-validator dv1ZAGvdsz5hHLwWXsVnM94hWf1pjbKVau1QVkaMJ92 --known-validator dv2eQHeP4RFrJZ6UeiZWoc3XTtmtZCUKxxCApCDcRNV --known-validator dv4ACNkpYPcE3aKmYDqZm9G5EB3J4MRoeE7WNDRBVJB --known-validator dv3qDFk1DTF36Z62bNvrCXe9sKATA6xvVy6A798xxAS --only-known-rpc --ledger ledger-devnet --rpc-port 8899 --dynamic-port-range 8001-8011 --entrypoint entrypoint.devnet.solana.com:8001 --entrypoint entrypoint2.devnet.solana.com:8001 --entrypoint entrypoint3.devnet.solana.com:8001 --entrypoint entrypoint4.devnet.solana.com:8001 --entrypoint entrypoint5.devnet.solana.com:8001 --expected-genesis-hash EtWTRABZaYq6iMfeYKouRu166VU2xqa1wcaWoxPkrZBG --wal-recovery-mode skip_any_corrupted_record --limit-ledger-size

# both build failed to catchup; observable ever widening slot distance from:

$ ./solana-validator-no-cpi-precompile-before --version
solana-validator 1.8.0 (src:9d91a6f1; feat:173977236)
$ ./solana-validator-no-cpi-precompile --version
solana-validator 1.8.0 (src:09b9de1b; feat:3245883770)

$ solana catchup -ud --our-localhost=8899 --log --follow --commitment root

```

fyi, the same argument managed to catch up to the cluster with `solana-validator 1.7.12 (src:ca83167c; feat:2013646575)
`

runtime/src/bank.rs
```diff
@@ -2579,8 +2567,37 @@ impl Bank {
             self.inherit_specially_retained_account_fields(&existing_genuine_program),
         );
         self.store_account_and_update_capitalization(program_id, &account);
+    }
+
+    /// Add a precompiled program account
+    pub fn add_precompiled_account(&self, program_id: &Pubkey) {
```
ryoqun, line 2573: nice code refactoring. :)

runtime/src/bank.rs
```diff
@@ -11818,7 +11873,100 @@ pub(crate) mod tests {
             slot,
         );
 
-        bank.add_native_program("mock_program", &program_id, true);
+        bank.add_builtin_account("mock_program", &program_id, true);
+    }
+
+    #[test]
+    fn test_add_precompiled_account() {
```
ryoqun, line 11880: nice test coverage. :)

CriesofCarrots:
> it seems master got incompatible?

I believe @jeffwashington is looking into this as well.

jeffwashington:
> > it seems master got incompatible?
> 
> I believe @jeffwashington is looking into this as well.

I'm bisecting now. mnb and master with hack
failure results in not making roots at least.

jackcmay:
> > > it seems master got incompatible?
> > 
> > 
> > I believe @jeffwashington is looking into this as well.
> 
> I'm bisecting now. mnb and master with hack failure results in not making roots at least.

So probably unrelated to this pr, aka, not blocking the commit of this pr?

ryoqun:
@jackcmay 

good work. code looks ok after all comments are addressed. thanks for working nicely according to my review comments.

the remaining task before merging would be the rather very tedious live cluster manual testing (seems to be blocked by incompatible master..).

jackcmay:
> @jackcmay
> 
> good work. code looks ok after all comments are addressed. thanks for working nicely according to my review comments.
> 
> the remaining task before merging would be the rather very tedious live cluster manual testing (seems to be blocked by incompatible master..).

@jeffwashington Can you ping me when you get to the bottom of master's incompatibility?

ryoqun:
> > > > it seems master got incompatible?
> > > 
> > > 
> > > I believe @jeffwashington is looking into this as well.
> > 
> > 
> > I'm bisecting now. mnb and master with hack failure results in not making roots at least.
> 
> So probably unrelated to this pr, aka, not blocking the commit of this pr?

if this pr won't be backported soon, I think it's ok to merge this pr, forgoing the live-cluster compatibility stamp. as far as I reviewed, this should be okay... ;)

programs/bpf/tests/programs.rs
seanyoung: I don't think it was even called. Fixed in https://github.com/solana-labs/solana/pull/20340

sdk/src/feature_set.rs
```diff
@@ -223,6 +223,12 @@ pub mod do_support_realloc {
     solana_sdk::declare_id!("75m6ysz33AfLA5DDEzWM1obBrnPQRSsdVQ2nRmc8Vuu1");
 }
 
+// Note: when this feature is cleaned up, also remove the secp256k1 program from
+// the list of builtins and remove its files from /programs
```
ryoqun, line 227: cross ref: https://github.com/solana-labs/solana/pull/27100#issuecomment-1217972054 at #27100

PR #19941, opened 2021-09-16
rpc: performance fix for getProgramAccounts

ckamm:
#### Problem

The result accounts were pushed into a fresh vector one-by-one, which produced significant slowdowns for very large responses (visible when exceeding thousands of accounts).

#### Summary of Changes

Pre-allocate the right capacity - the result length is known.

### DIFF
--- a/rpc/src/rpc.rs
+++ b/rpc/src/rpc.rs
@@ -355,12 +355,10 @@ impl JsonRpcRequestProcessor {
         let encoding = config.encoding.unwrap_or(UiAccountEncoding::Base64);
         check_slice_and_encoding(&encoding, config.data_slice.is_some())?;
 
-        let mut accounts: Vec<Option<UiAccount>> = Vec::with_capacity(pubkeys.len());
-        for pubkey in pubkeys {
-            let response_account =
-                get_encoded_account(&bank, &pubkey, encoding, config.data_slice)?;
-            accounts.push(response_account)
-        }
+        let accounts = pubkeys
+            .into_iter()
+            .map(|pubkey| get_encoded_account(&bank, &pubkey, encoding, config.data_slice))
+            .collect::<Result<Vec<_>>>()?;
         Ok(new_response(&bank, accounts))
     }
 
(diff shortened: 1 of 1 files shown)


ckamm:
Optionally these could also be rewritten to use iterators, like
```
           keyed_accounts
                .into_iter()
                .map(|(pubkey, account)| {
                    Ok(RpcKeyedAccount {
                         pubkey: pubkey.to_string(),
                         account: encode_account(&account, &pubkey, encoding, data_slice_config)?,
                    })
                })
                .collect::<Result<Vec<_>>>()?
```

jstarry:
> Optionally these could also be rewritten to use iterators, like
> 
> ```
>            keyed_accounts
>                 .into_iter()
>                 .map(|(pubkey, account)| {
>                     Ok(RpcKeyedAccount {
>                          pubkey: pubkey.to_string(),
>                          account: encode_account(&account, &pubkey, encoding, data_slice_config)?,
>                     })
>                 })
>                 .collect::<Result<Vec<_>>>()?
> ```

I typically prefer this approach, do you mind updating?

ckamm:
@jstarry Not at all, that's what I did initially, then undid to keep the patch solely about the performance aspect. :)

jstarry:
lgtm once the checks are fixed, thanks for this!

CriesofCarrots (commented):
Looks like you just need to remove the `Ok(..?)` wrappers in two of your iterators, then gtg.
Thanks for this!

ckamm:
> Looks like you just need to remove the `Ok(..?)` wrappers in two of your iterators, then gtg.

Could you elaborate? I'm intentionally making the `map()` produce an iterator over `Result<_>`, so I can deal with the potential error inside the closure. Is there a better way to write it?

jstarry:
Screen Shot 2021-09-16 at 11 07 35 AM

brianlong:
When this is ready, please prepare a version of v1.6.25 with these changes for staging & user testing. Apps are eager to test this out on the large mainnet account set for Serum Program V3.

jstarry (approved):

CriesofCarrots (approved):
Thanks again!

PR #19944, opened 2021-09-16
Fix tpu service fanout

sakridge:
#### Problem

If `slots_per_epoch` is smaller than `MAX_FANOUT_SLOTS * 2`, then the tpu service can return an error.

#### Summary of Changes

Limit fanout to slots_per_epoch

Fixes #

### DIFF
--- a/client/src/tpu_client.rs
+++ b/client/src/tpu_client.rs
@@ -119,16 +119,21 @@ struct LeaderTpuCache {
     first_slot: Slot,
     leaders: Vec<Pubkey>,
     leader_tpu_map: HashMap<Pubkey, SocketAddr>,
+    slots_in_epoch: Slot,
+    last_epoch_info_slot: Slot,
 }
 
 impl LeaderTpuCache {
     fn new(rpc_client: &RpcClient, first_slot: Slot) -> Result<Self> {
-        let leaders = Self::fetch_slot_leaders(rpc_client, first_slot)?;
+        let slots_in_epoch = rpc_client.get_epoch_info()?.slots_in_epoch;
+        let leaders = Self::fetch_slot_leaders(rpc_client, first_slot, slots_in_epoch)?;
         let leader_tpu_map = Self::fetch_cluster_tpu_sockets(rpc_client)?;
         Ok(Self {
             first_slot,
             leaders,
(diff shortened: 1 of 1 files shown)


client/src/tpu_client.rs
```diff
@@ -187,7 +187,9 @@ impl LeaderTpuCache {
     }
 
     fn fetch_slot_leaders(rpc_client: &RpcClient, start_slot: Slot) -> Result<Vec<Pubkey>> {
-        Ok(rpc_client.get_slot_leaders(start_slot, 2 * MAX_FANOUT_SLOTS)?)
+        let epoch_info = rpc_client.get_epoch_info()?;
+        let fanout = (2 * MAX_FANOUT_SLOTS).min(epoch_info.slots_in_epoch * 2);
```
jstarry, line 191:
```suggestion
        let fanout = (2 * MAX_FANOUT_SLOTS).min(epoch_info.slots_in_epoch);
```
I don't think we want to multiple by 2 here right? You could be towards the end of an epoch and I think this fails if fetching slot leaders for current epoch + 2?

sakridge: yea. good point

jstarry (dismissed):
Thanks! Sorry I never tested on a new cluster

client/src/tpu_client.rs
```diff
@@ -187,7 +187,9 @@ impl LeaderTpuCache {
     }
 
     fn fetch_slot_leaders(rpc_client: &RpcClient, start_slot: Slot) -> Result<Vec<Pubkey>> {
-        Ok(rpc_client.get_slot_leaders(start_slot, 2 * MAX_FANOUT_SLOTS)?)
+        let epoch_info = rpc_client.get_epoch_info()?;
```
mvines, line 190: Let's cache `slots_in_epoch` please to reduce RPC server load, we don't need to be continually asking RPC for this value

sakridge: ok, now it should just call once per epoch.

mvines: Thanks, that's much better than what I was thinking of :)

mvines (approved):

PR #19955, opened 2021-09-16
AcctIdx: hold_range_in_memory

jeffwashington:
#### Problem
Accounts index needs help from higher level code to know ranges of items which will be accessed and need to be held in memory. This change is staging for when we don't hold everything in memory in the accts index.
#### Summary of Changes

Fixes #


brooksprumo (approved):
lgtm

PR #19958, opened 2021-09-16
stake: Add BorshDeserialize trait to structs

joncinque:
#### Problem

In its stake program copy-pasta, the stake pool program also has `BorshDeserialize` implemented for the stake program structs in order to cut down on compute cost when deserializing through stake accounts.  Since that implementation is not present in the sdk, stake pools currently can't move over to using it to resolve https://github.com/solana-labs/solana-program-library/issues/1865

#### Summary of Changes

Derive `BorshDeserialize` and `BorshSchema` where applicable for stake program structs.  The only important difference is `StakeState`, which can't use the default implementation because bincode serializes enums to 4 bytes, while borsh uses 1.

### DIFF
--- a/sdk/program/src/stake/state.rs
+++ b/sdk/program/src/stake/state.rs
@@ -11,6 +11,7 @@ use {
         },
         stake_history::StakeHistory,
     },
+    borsh::{maybestd::io, BorshDeserialize, BorshSchema},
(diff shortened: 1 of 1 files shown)


CriesofCarrots (approved):
lgtm!

PR #19966, opened 2021-09-17
AcctIdx: start calling flush on in mem cache buckets

jeffwashington:
#### Problem
Flushing will be the most complicated thing that happens with the disk based accounts index.
This is the beginnings of that behavior.
#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/in_mem_accounts_index.rs
+++ b/runtime/src/in_mem_accounts_index.rs
@@ -365,7 +365,7 @@ impl<T: IndexValue> InMemAccountsIndex<T> {
     }
 
     pub(crate) fn flush(&self) {
-        let flushing = self.flushing_active.swap(true, Ordering::Relaxed);
+        let flushing = self.flushing_active.swap(true, Ordering::Acquire);
         if flushing {
             // already flushing in another thread
             return;
@@ -373,11 +373,15 @@ impl<T: IndexValue> InMemAccountsIndex<T> {
 
         self.flush_internal();
 
-        self.flushing_active.store(false, Ordering::Relaxed);
+        self.flushing_active.store(false, Ordering::Release);
+    }
+
+    pub fn set_bin_dirty(&self) {
+        self.bin_dirty.store(true, Ordering::Release);
     }
 
     fn flush_internal(&self) {
-        let was_dirty = self.bin_dirty.swap(false, Ordering::Relaxed);
+        let was_dirty = self.bin_dirty.swap(false, Ordering::Acquire);
         if !was_dirty {
             // wasn't dirty, no need to flush
             return;
--- a/runtime/src/in_mem_accounts_index.rs
+++ b/runtime/src/in_mem_accounts_index.rs
@@ -7,7 +7,7 @@ use solana_measure::measure::Measure;
 use solana_sdk::{clock::Slot, pubkey::Pubkey};
 use std::collections::{hash_map::Entry, HashMap};
 use std::ops::{Bound, RangeBounds, RangeInclusive};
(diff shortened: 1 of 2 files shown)


runtime/src/in_mem_accounts_index.rs
```diff
@@ -264,6 +271,42 @@ impl<T: IndexValue> InMemAccountsIndex<T> {
         result
     }
 
+    pub(crate) fn flush(&self) {
+        let flushing = self.flushing_active.swap(true, Ordering::Relaxed);
```
brooksprumo, line 275: Since `flushing_active` is like a thread-safe flag, and this part raises the flag, I think this Ordering should be `Ordering::Acquire`. This ensures no loads/stores get reordered _before_ the load part of this swap. I.e. nothing happens before saying `start flushing`.

runtime/src/in_mem_accounts_index.rs
```diff
@@ -264,6 +271,42 @@ impl<T: IndexValue> InMemAccountsIndex<T> {
...
+        if flushing {
+            // already flushing in another thread
+            return;
+        }
+
+        self.flush_internal();
+
+        self.flushing_active.store(false, Ordering::Relaxed);
```
brooksprumo, line 283: Since `flushing_active` is like a thread-safe flag, and this part lowers the flag, I think this Ordering should be `Ordering::Release`. This ensures no loads/stores get reordered _after_ this store. I.e. this 'done flushing' only happens when the flushing is done, and nothing happens afterwards.

runtime/src/in_mem_accounts_index.rs
```diff
@@ -264,6 +271,42 @@ impl<T: IndexValue> InMemAccountsIndex<T> {
...
+    }
+
+    fn flush_internal(&self) {
+        let was_dirty = self.bin_dirty.swap(false, Ordering::Relaxed);
```
brooksprumo, line 287: Same comment here as with `flushing_active`. I think this should be `Ordering::Acquire`.

...But where does `bin_dirty` get set to true? That one should be `Ordering::Release`.

runtime/src/in_mem_accounts_index.rs
```diff
@@ -264,6 +271,42 @@ impl<T: IndexValue> InMemAccountsIndex<T> {
...
+                // If a parallel operation dirties the item again - even while this flush is occurring,
+                //  the last thing the writer will do, after updating contents, is set_dirty(true)
+                //  That prevents dropping an item from cache before disk is updated to latest in mem.
+                v.set_dirty(false);
+
+                // soon, this will update disk from the in-mem contents
+            }
+            */
+        }
+    }
```
brooksprumo, line 308: I guess this is OK to comment out for now?

brooksprumo (commented):
Comments about synchronization with Atomics, then wondering about set_dirty.

runtime/src/in_mem_accounts_index.rs
jeffwashington: yes, that code didn't/doesn't exist yet but will soon. Showing you the basics of what it does.

runtime/src/in_mem_accounts_index.rs
jeffwashington: done, and added set_bin_dirty, with Release.

runtime/src/in_mem_accounts_index.rs
jeffwashington: done

brooksprumo (approved):

PR #19967, opened 2021-09-17
AcctIdx: handle future config defaults

jeffwashington:
#### Problem
Config struct changes break all these places each time. This code needs to be more tolerant of changing options.
#### Summary of Changes

Fixes #

### DIFF
--- a/ledger-tool/src/main.rs
+++ b/ledger-tool/src/main.rs
@@ -1908,12 +1908,12 @@ fn main() {
             }
         }
         ("verify", Some(arg_matches)) => {
-            let accounts_index_config = value_t!(arg_matches, "accounts_index_bins", usize)
-                .ok()
-                .map(|bins| AccountsIndexConfig { bins: Some(bins) });
-
+            let mut accounts_index_config = AccountsIndexConfig::default();
+            if let Some(bins) = value_t!(matches, "accounts_index_bins", usize).ok() {
+                accounts_index_config.bins = Some(bins);
+            }
             let accounts_db_config = Some(AccountsDbConfig {
-                index: accounts_index_config,
+                index: Some(accounts_index_config),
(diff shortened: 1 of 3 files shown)


ledger-tool/src/main.rs
```diff
@@ -1908,12 +1908,12 @@ fn main() {
...
-                .ok()
-                .map(|bins| AccountsIndexConfig { bins: Some(bins) });
-
+            let mut accounts_index_config = AccountsIndexConfig::default();
+            if let Some(bins) = value_t!(matches, "accounts_index_bins", usize).ok() {
+                accounts_index_config.bins = Some(bins);
+            }
             let accounts_db_config = Some(AccountsDbConfig {
-                index: accounts_index_config,
+                index: Some(accounts_index_config),
```
brooksprumo, line 1916: Looks like now `AccountsDbConfig::index` is always `Some`. Is this going to change in future PRs? If not, would it worthwhile to remove the `Option<>` on the `index` field in `AccountsDbConfig`?

brooksprumo (commented):
Looks good, just the one question on desired usage going forward.

ledger-tool/src/main.rs
jeffwashington: the option to pass None overall is still a valid option. It is true that currently validator and ledger-tool (verify) will always pass Some(...), but tests or other code paths may still pass None.

jeffwashington: it is a rust hassle to figure out whether one of several optional args are present and if so make the overall struct a Some vs None. So, I chose to make it a Some(struct with Nones) where then optional args fill in the Nones with Some per arg. This code will likely get cleaned up and even consolidated at some point by someone because arg handling is messy.

PR #19970, opened 2021-09-17
Added web3 reference guide

cryptogosu:
#### Problem

Web3 was lacking a reference guide, and many ask for it in the support discord

#### Summary of Changes

Added web3 reference guide to the docs

### DIFF
--- a/docs/src/developing/clients/javascript-reference.md
+++ b/docs/src/developing/clients/javascript-reference.md
@@ -1,0 +1,729 @@
+---
+title: Web3 API Reference
+---
+
+## Web3 API Reference Guide
+
+## General
+
+### Connection
+
+[Source Documentation](https://solana-labs.github.io/solana-web3.js/classes/Connection.html)
+
+Connection is used to interact with the [Solana JSON RPC](https://docs.solana.com/developing/clients/jsonrpc-api). You can use Connection to confirm transactions, get account info, and more.
+
+You create a connection by defining the JSON RPC cluster endpoint and the desired commitment. Once this is complete, you can use this connection object to interact with any of the Solana JSON RPC API.
+
+#### Example Usage
+
+```javascript
+const web3 = require("@solana/web3.js");
+
+let connection = new web3.Connection(web3.clusterApiUrl('devnet'), 'confirmed');
+
+let slot = await connection.getSlot();
+console.log(slot);
+// 93186439
+
+let blockTime = await connection.getBlockTime(slot);
+console.log(blockTime);
+// 1630747045
+
+let block = await connection.getBlock(slot);
+console.log(block);
+
+/*
+{
+    blockHeight: null,
+    blockTime: 1630747045,
+    blockhash: 'AsFv1aV5DGip9YJHHqVjrGg6EKk55xuyxn2HeiN9xQyn',
+    parentSlot: 93186438,
+    previousBlockhash: '11111111111111111111111111111111',
+    rewards: [],
+    transactions: []
+}
+*/
+
+let slotLeader = await connection.getSlotLeader();
+console.log(slotLeader);
+//49AqLYbpJYc2DrzGUAH1fhWJy62yxBxpLEkfJwjKy2jr
+```
+
+The above example shows only a few of the methods on Connection. Please see the [source generated docs](https://solana-labs.github.io/solana-web3.js/classes/Connection.html) for the full list.
+
+### Transaction
+
+[SourceDocumentation](https://solana-labs.github.io/solana-web3.js/classes/Transaction.html)
+
+A transaction is used to interact with programs on the Solana blockchain. These transactions are constructed with TransactionInstructions, containing all the accounts possible to interact with, as well as any needed data or program addresses. Each TransactionInstruction consists of keys, data, and a programId. You can do multiple instructions in a single transaction, interacting with multiple programs at once.
+
+#### Example Usage
+
+```javascript
+const web3 = require('@solana/web3.js');
+const nacl = require('tweetnacl');
+
+// Airdrop SOL for paying transactions
+let payer = web3.Keypair.generate();
+let connection = new web3.Connection(web3.clusterApiUrl('devnet'), 'confirmed');
+
+let airdropSignature = await connection.requestAirdrop(
+    payer.publicKey,
+    web3.LAMPORTS_PER_SOL,
+);
+
+await connection.confirmTransaction(airdropSignature);
+
+let toAccount = web3.Keypair.generate();
+
+// Create Simple Transaction
+let transaction = new web3.Transaction();
+
+// Add an instruction to execute
+transaction.add(web3.SystemProgram.transfer({
+    fromPubkey: payer.publicKey,
+    toPubkey: toAccount.publicKey,
+    lamports: 1000,
+}));
+
+// Send and confirm transaction
+// Note: feePayer is by default the first signer, or payer, if the parameter is not set
+await web3.sendAndConfirmTransaction(connection, transaction, [payer])
+
+// Alternatively, manually construct the transaction
+let recentBlockhash = await connection.getRecentBlockhash();
+let manualTransaction = new web3.Transaction({
+    recentBlockhash: recentBlockhash.blockhash,
+    feePayer: payer.publicKey
+});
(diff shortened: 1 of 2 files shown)


docs/src/developing/clients/javascript-reference.md
```diff
@@ -0,0 +1,873 @@
...
+
+### Authorized
+
+[Source Documentation](https://solana-labs.github.io/solana-web3.js/classes/Authorized.html)
+
+Authorized is an object used when creating an authorized account for staking within Solana. You can designate a `staker` and `withdrawer` separately, allowing for a different account to withdraw other than the staker.
+
+You can find more usage of the `Authorized` object under [`StakeProgram`](javascript-api.md#StakeProgram)
+
+### BpfLoader
```
jstarry, line 15: Let's leave this out, I'd like to deprecate the bpf loader from web3 and encourage devs to use the cli instead

docs/src/developing/clients/javascript-reference.md
```diff
@@ -0,0 +1,873 @@
...
+//49AqLYbpJYc2DrzGUAH1fhWJy62yxBxpLEkfJwjKy2jr
+```
+
+The above example shows only a few of the methods on Connection. Please see the [source generated docs](https://solana-labs.github.io/solana-web3.js/classes/Connection.html) for the full list.
+
+### Enum
+
+[Source Documentation](https://solana-labs.github.io/solana-web3.js/classes/Enum.html)
+
+The Enum class is used to represent a Rust compatible Enum in javascript. The enum will just be a string representation if logged but can be properly encoded/decoded when used in conjunction with [Struct](javascript-api.md#Struct).
```
jstarry, line 114: It's specifically for borsh encoded data, can you mention that?

docs/src/developing/clients/javascript-reference.md
```diff
@@ -0,0 +1,873 @@
...
+
+Web3:
+```javascript
+import {Enum} from '@solana/web3.js';
+
+export class AccountType extends Enum {}
+```
+
+
+### EpochSchedule
```
jstarry, line 135: I don't think this one is too valuable

docs/src/developing/clients/javascript-reference.md
```diff
@@ -0,0 +1,873 @@
...
+//   166, 210, 195, 188,  60, 220, 210,  96, 136, 158,   6,
+//   205, 189, 165, 112,  32, 200, 116, 164, 234
+// ]
+```
+
+Using `generate` generates a random Keypair for use as an account on Solana. Using `fromSeed`, you can generate a Keypair using a deterministic constructor. `fromSecret` creates a Keypair from a secret Uint8array. You can see that the publicKey for the `generate` Keypair and `fromSecret` Keypair are the same because the secret from the `generate` Keypair is used in `fromSecret`.
+
+**Warning**: Do not use `fromSeed` unless you are creating a seed with high entropy. Do not share your seed. Treat the seed like you would a private key.
+
+### Loader
```
jstarry, line 230: Remove please

docs/src/developing/clients/javascript-reference.md
```diff
@@ -0,0 +1,873 @@
...
+```
+The above code creates a `createStakeAccountInstruction` to be used when creating an account with the `StakeProgram`. The Lockup is set to 0 for both the epoch and Unix timestamp, disabling lockup for the account.
+
+See [StakeProgram](javascript-api.md#StakeProgram) for more.
+
+### Message
+
+[Source Documentation](https://solana-labs.github.io/solana-web3.js/classes/Message.html)
+
+Message is used as another way to construct transactions. You can construct a message using the accounts, header, instructions, and recentBlockhash that are a part of a transaction.
```
jstarry, line 270: Can you also say that a transaction is the combination of a message and a list of signatures? It's not too clear that `Message` is basically the same thing as a transaction (minus the signatures)

docs/src/developing/clients/javascript-reference.md
```diff
@@ -0,0 +1,873 @@
...
+);
+
+await web3.sendAndConfirmTransaction(connection, transaction, [fromPublicKey])
+```
+
+### NonceAccount
+
+[Source Documentation](https://solana-labs.github.io/solana-web3.js/classes/NonceAccount.html)
+
+Normally a transaction is rejected if a transaction's `recentBlockhash` field is too old. To provide for certain custodial services, Nonce Accounts are used.
```
jstarry, line 336: Can you specify that "nonced" transactions (that is, transactions which start with an advance nonce ix) do not expire because the nonce account stores a recent blockhash on-chain which is then specified in the transaction's `recentBlockhash` field?

docs/src/developing/clients/javascript-reference.md
```diff
@@ -0,0 +1,873 @@
...
+let nonceAccountData = await connection.getNonce(
+    nonceAccount.publicKey,
+    'confirmed',
+);
+
+console.log(nonceAccountData);
+// NonceAccount {
+//   authorizedPubkey: PublicKey {
+//     _bn: <BN: 919981a5497e8f85c805547439ae59f607ea625b86b1138ea6e41a68ab8ee038>
+//   },
```
jstarry, line 393: Side note, it would be great if this logged in a prettier way. Do you mind writing an issue for that?

docs/src/developing/clients/javascript-reference.md
```diff
@@ -0,0 +1,873 @@
...
+});
+
+await web3.sendAndConfirmTransaction(connection, withdrawTransaction, [fromPublicKey, authorizedAccount]);
+```
+
+### Struct
+
+[SourceDocumentation](https://solana-labs.github.io/solana-web3.js/classes/Struct.html)
+
+The struct class is used to create Rust compatible structs in javascript.
```
jstarry, line 604: Again, can you please add a note that this is specific to borsh?

docs/src/developing/clients/javascript-reference.md
```diff
@@ -0,0 +1,873 @@
...
+
+### TransactionInstruction
+
+[SourceDocumentation](https://solana-labs.github.io/solana-web3.js/classes/TransactionInstruction.html)
+
+The TransactionInstruction object is used to create instructions to be used in transactions. They consist of keys, data, and a programId.
+
+See [Transaction](javascript-api.md#Transaction) for more use.
+
+### ValidatorInfo
```
jstarry, line 793: Don't think this is too important

jstarry (commented):
This is super helpful!

docs/src/developing/clients/javascript-reference.md
cryptogosu: Created #20022. I'll take a stab at it if it is still open tomorrow.

cryptogosu:
@jstarry Updated all documentation based on your suggestions.

docs/src/developing/clients/javascript-reference.md
```diff
@@ -0,0 +1,749 @@
...
+);
+
+await web3.sendAndConfirmTransaction(connection, transaction, [fromPublicKey])
+```
+
+### NonceAccount
+
+[Source Documentation](https://solana-labs.github.io/solana-web3.js/classes/NonceAccount.html)
+
+Normally a transaction is rejected if a transaction's `recentBlockhash` field is too old. To provide for certain custodial services, Nonce Accounts are used. Transactions created by a Nonce Account do not expire because they start with an `AdvanceNonceAccount` instruction, which updates the Nonce Account with an up-to-date `recentBlockhash`. This on-chain `recentBlockhash` is then used in the Nonce Account's transaction.
```
jstarry, line 246:
```suggestion
Normally a transaction is rejected if a transaction's `recentBlockhash` field is too old. To provide for certain custodial services, Nonce Accounts are used. Transactions which use a `recentBlockhash` captured on-chain by a Nonce Account do not expire as long at the Nonce Account is not advanced.
```

jstarry (commented):
I wish that the most important classes came first and the others were grouped by category. For example, starting off with `Authorized` makes it feel that it's a foundational important thing to learn, but it's really just a detail of how the stake program. How about splitting some of these classes into categories like "General", "Staking", etc

jstarry (approved):

jstarry:
Looks much better thanks!

PR #19990, opened 2021-09-17
client: Add retry logic on Pubsub 429 error during connect

joncinque:
#### Problem

The pubsub client can give a 429 (TOO_MANY_REQUESTS) when trying to connect, causing avoidable failures to users.

#### Summary of Changes

Respect the backoff from a 429, and retry the connection after waiting.  This mimics the send retry logic at https://github.com/solana-labs/solana/blob/74cc2e5ffac5c0c7bbb9126f27d25a1704366b41/client/src/http_sender.rs#L96-L117

Fixes #

### DIFF
--- a/client/src/pubsub_client.rs
+++ b/client/src/pubsub_client.rs
@@ -23,7 +23,8 @@ use {
             mpsc::{channel, Receiver},
             Arc, RwLock,
         },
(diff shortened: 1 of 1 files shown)


mvines (approved):
sweet!

Issue #19072, opened 2021-08-04
Web3.js: Subscribe errors are silent


tomjohn1028:
#### Problem
If a RPC request fails to subscribe an account, the failure is only logged. There is no onError callback or error propagation. Without the propagation users' and the Dapps cannot report these errors.

![Screen Shot 2021-07-02 at 12 18 22 PM](https://user-images.githubusercontent.com/32071703/124302898-c2053880-db2f-11eb-971d-ea72c0edb6f4.png)

[Current source](https://github.com/solana-labs/solana/blob/0b8d14b0fc19345411dabeface475d27a58779cb/web3.js/src/connection.ts#L3644)
#### Proposed Solution

Create an option _onSubscribeError_ argument that could be passed or attached to a connection instance.

jstarry:
@tomjohn1028 I also just realized that if one subscription fails, it will not be retried until either the whole connection gets interrupted or if the dapp makes a new subscription. I think that web3 should auto retry subscriptions (maybe configurable) and we could stick that into the heartbeat timer, thoughts?

jstarry:
Seems like the main value in a subscription error callback is when a new subscription is rate limited, is that right?

tomjohn1028:
> I think that web3 should auto retry subscriptions (maybe configurable) and we could stick that into the heartbeat timer, thoughts? 

Definitely think a configurable retry adds value

> Seems like the main value in a subscription error callback is when a new subscription is rate limited, is that right?

Rate limited or any other RPC issue that causes the client the fail to connect or disconnect. On devnet our UI was silently failing to work so a normal user would have had no idea what was happening if we couldn't report back.

jstarry:
Right so the websocket should automatically reconnect so I'm actually thinking that we could fix this in a few steps:
1. new config value for auto reconnecting rate limit issues with backoff
2. new apis for setting connection-wide open, error, and close callbacks to know when the connection is reconnecting

I'm undecided if subscription specific error callbacks would be useful or if it's sufficient to have just have a connection wide callback

steveluscher:
Hey @oJshua! Thanks for taking a crack at a PR. I see it's closed after lapsing. Is this free for someone else to work on now, or do you want to revive your old PR?

PR #19993, opened 2021-09-17
feat(web3.js): provide subscribe error notifications for websocket rpc

oJshua:
#### Problem
If a RPC request fails to subscribe an account, the failure is only logged. There is no onError callback or error propagation. Without the propagation users' and the Dapps cannot report these errors.

#### Summary of Changes
- add support for subscribing to websocket rpc subscribe errors

Fixes https://github.com/solana-labs/solana/issues/19072

### DIFF
--- a/web3.js/src/connection.ts
+++ b/web3.js/src/connection.ts
@@ -3905,7 +3905,6 @@ export class Connection {
   /**
    * Register a callback to be invoked when there is a subscribe error
    *
-   *
    * @param callback Function to invoke when a subscribe error occurs
    * @returns subscription id
    */
--- a/web3.js/src/connection.ts
+++ b/web3.js/src/connection.ts
@@ -3727,11 +3727,11 @@ export class Connection {
           sub.subscriptionId = null;
         }
 
-        Object.values(this._subscribeErrorCallbacks).forEach(callback =>
-          callback(rpcMethod, rpcArgs, error),
-        );
-
-        if (this._subscribeErrorCallbackCounter === 0) {
+        if (Object.values(this._subscribeErrorCallbacks).length) {
+          Object.values(this._subscribeErrorCallbacks).forEach(callback =>
+            callback(rpcMethod, rpcArgs, error),
+          );
+        } else {
           console.error(
             `${rpcMethod} error for argument`,
             rpcArgs,
--- a/web3.js/src/connection.ts
+++ b/web3.js/src/connection.ts
@@ -1721,6 +1721,15 @@ export type KeyedAccountInfo = {
 };
 
 /**
+ * Callback function for subscribe error notifications
+ */
+export type SubscribeErrorCallback = (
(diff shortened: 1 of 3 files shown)


web3.js/src/connection.ts
```diff
@@ -3707,12 +3721,23 @@ export class Connection {
...
-      } catch (err) {
+      } catch (error) {
         if (sub.subscriptionId === 'subscribing') {
           // eslint-disable-next-line require-atomic-updates
           sub.subscriptionId = null;
         }
-        console.error(`${rpcMethod} error for argument`, rpcArgs, err.message);
+
+        if (Object.values(this._subscribeErrorCallbacks).length) {
+          Object.values(this._subscribeErrorCallbacks).forEach(callback =>
```
jstarry, line 3731: nit: rather than creating the `Object.values` array twice, can you create it once and reuse it?

web3.js/src/connection.ts
```diff
@@ -1720,6 +1720,15 @@ export type KeyedAccountInfo = {
   accountInfo: AccountInfo<Buffer>;
 };
 
+/**
+ * Callback function for subscribe error notifications
+ */
+export type SubscribeErrorCallback = (
```
jstarry, line 1726: Some high level feedback:
1. It would be nice to have proper typing for the rpc args. I think an error info object could actually be the union of each subscription type plus its info fields rather than the built rpc args. Then developers can rely on the type system to pull info about the failed subscription out.
2. It'd be nice to return the subscription id here too for easier matching

web3.js/src/connection.ts
```diff
@@ -3877,6 +3902,31 @@ export class Connection {
     }
   }
 
+  /**
+   * Register a callback to be invoked when there is a subscribe error
+   *
+   * @param callback Function to invoke when a subscribe error occurs
+   * @returns subscription id
+   */
+  onSubscribeError(callback: SubscribeErrorCallback): number {
```
jstarry, line 3911: This feels useful but I believe the main failure recently is rate limited subscriptions so I think that an error callback on each subscription method would be more valuable. I started a conversation here about this: https://github.com/solana-labs/solana/issues/19072#issuecomment-923259968

oJshua: The connection-wide open and close events sound like a good plan. Would it make sense to have Connection extend EventEmitter or should we follow a similar pattern to our existing on.*, remove.* methods?

jstarry: I think the existing methods are better because they can be strongly typed

oJshua: Okay, cool. Returning a subscription id for these seemed a little strange since they aren't directly attached to a rpc subscription, but I suppose I could refer to them as callback ids.

jstarry: Sure, that name change sounds good

PR #20000, opened 2021-09-18
Ingest duplicate proofs from gossip into blockstore

### LINKED ISSUE #16508: Gossip proofs should be ingested into blockstore's, and used to notify ReplayStage of duplicates
#### Problem
Gossip proofs are propagated but not currently:

1) Ingested into blockstore's duplicate column family (`store_duplicate_if_not_existing()` only runs on received shreds from WindowService)
2) Not used to notify ReplayStage of duplicates

This is an issue if some nodes see duplicates and others don't.

AshwinSekar:
#### Problem
Duplicate proofs are not ingested from gossip, so there could be scenarios where a validator is not aware that there is duplicate contention ongoing.

#### Summary of Changes
Monitor gossip for valid gossip proofs, propagate into blockstore and notify replay stage. 

Fixes #16508

### DIFF
--- a/core/src/cluster_info_entries_listener.rs
+++ b/core/src/cluster_info_entries_listener.rs
@@ -1,0 +1,174 @@
+use crate::{result::Result, window_service::DuplicateSlotSender};
+use crossbeam_channel::{unbounded, Receiver as CrossbeamReceiver, Sender as CrossbeamSender};
+use log::*;
+use solana_gossip::{
+    cluster_info::{ClusterInfo, GOSSIP_SLEEP_MILLIS},
+    crds::Cursor,
+    crds_value::CrdsData,
+    duplicate_shred::{self, DuplicateShred},
+};
+use solana_ledger::blockstore::Blockstore;
+use solana_ledger::leader_schedule_utils::slot_leader_at;
+use solana_metrics::inc_new_counter_debug;
+use solana_poh::poh_recorder::PohRecorder;
+use solana_sdk::{clock::Slot, pubkey::Pubkey};
+use std::{
+    cmp::Ordering::{Equal, Greater},
+    collections::{hash_map::Entry, HashMap},
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        {Arc, Mutex},
+    },
+    thread::{self, sleep, Builder, JoinHandle},
+    time::Duration,
+};
+
+pub type DuplicateShredSender = CrossbeamSender<Pubkey>;
+pub type DuplicateShredReceiver = CrossbeamReceiver<Pubkey>;
+
+pub struct ClusterInfoEntriesListener {
+    thread_hdls: Vec<JoinHandle<()>>,
+}
+
+impl ClusterInfoEntriesListener {
+    pub fn new(
+        exit: &Arc<AtomicBool>,
+        cluster_info: Arc<ClusterInfo>,
+        poh_recorder: &Arc<Mutex<PohRecorder>>,
+        blockstore: Arc<Blockstore>,
+        duplicate_slot_sender: DuplicateSlotSender,
+    ) -> Self {
+        let (duplicate_shred_sender, duplicate_shred_receiver) = unbounded();
+        let exit_ = exit.clone();
+        let cluster_info_ = cluster_info.clone();
+
+        let listen_thread = Builder::new()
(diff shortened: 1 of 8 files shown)


core/src/tpu.rs
```diff
@@ -105,6 +109,13 @@ impl Tpu {
             bank_notification_sender,
             cluster_confirmed_slot_sender,
         );
+        let cluster_info_entries_listener = ClusterInfoEntriesListener::new(
+            exit,
+            cluster_info.clone(),
+            poh_recorder,
```
carllin, line 115: If we're just using this to get a bank, I think passing `bank_forks` is probably better than `poh_recorder`

core/src/cluster_info_entries_listener.rs
```diff
@@ -0,0 +1,174 @@
...
+                            }
+                            Entry::Occupied(mut entry) => {
+                                let (chunks, _) = entry.get_mut();
+                                chunks.push(chunk);
+                            }
+                        });
+
+                    chunks_per_slot.into_iter().for_each(|(slot, (chunks, num_chunks)): (Slot, (Vec<DuplicateShred>, usize))| {
+                        let leader = |slot : Slot| {
+                            poh_recorder.lock().unwrap().bank().as_ref().map(|bank| slot_leader_at(slot, bank)).flatten()
```
carllin, line 145: I think we can copy the bank out above the `for_each()` and save the locking

core/src/cluster_info_entries_listener.rs
```diff
@@ -0,0 +1,174 @@
...
+                                }
+                                Err(e) => {
+                                    warn!("Unable to ingest duplicate slot proof for {}: {}", slot, e)
+                                }
+                            },
+                            Greater => error!("Duplicate slot proof for {} is corrupt, expected {} chunks but got {} chunks", slot, num_chunks, chunks.len()),
+                            _ => (),
+                        }
+                    });
+                });
```
carllin, line 170: Would be good to extract this into a function and then write some unit tests

core/src/cluster_info_entries_listener.rs
```diff
@@ -0,0 +1,174 @@
...
+                                    if let Err(e) = duplicate_slot_sender.send(slot) {
+                                        error!("Unable to notify replay stage of duplicate slot {} from gossip proof: {}", slot, e)
+                                    }
+                                }
+                                Err(e) => {
+                                    warn!("Unable to ingest duplicate slot proof for {}: {}", slot, e)
+                                }
+                            },
+                            Greater => error!("Duplicate slot proof for {} is corrupt, expected {} chunks but got {} chunks", slot, num_chunks, chunks.len()),
+                            _ => (),
```
carllin, line 167: Maybe a comment here about what happens when we have insufficient chunks to reconstruct the shreds. Something along the lines of:

"We will retry when we get the missing chunks via another signal from `duplicate_shred_receiver`"

core/src/cluster_info_entries_listener.rs
```diff
@@ -0,0 +1,174 @@
...
+                        }
+                    }
+                });
+            }
+            sleep(Duration::from_millis(GOSSIP_SLEEP_MILLIS));
+        }
+        Ok(())
+    }
+
+    fn process_duplicate_shreds_loop(
```
carllin, line 103: maybe name `reconstruct_duplicate_shred_proofs_loop`

core/src/cluster_info_entries_listener.rs
```diff
@@ -0,0 +1,174 @@
...
+        duplicate_shred_sender: DuplicateShredSender,
+    ) -> Result<()> {
+        let mut cursor = Cursor::default();
+        while !exit.load(Ordering::Relaxed) {
+            let entries = cluster_info.get_entries(&mut cursor);
+            inc_new_counter_debug!("cluster_info_entries_listener-recv_count", entries.len());
+            if !entries.is_empty() {
+                entries.iter().for_each(|value| {
+                    if let CrdsData::DuplicateShred(_, _) = value.data {
+                        if let Err(e) = duplicate_shred_sender.send(value.pubkey()) {
```
carllin, line 92: any reason not to just directly reconstruct the shred proofs here? i.e. move the work from `process_duplicate_shreds_loop()` here

PR #20001, opened 2021-09-18
Prioritized transaction forwarding

jstarry:
#### Problem
Transaction forwarding doesn't prioritize critical consensus related messages

#### Summary of Changes
For each packet with the `is_simple_vote_tx` meta flag, extract the vote account address and use it to look up the stake weight. Then, forward those packets in order by stake weight followed by non vote transactions.

Fixes #


runtime/src/bank.rs
```diff
@@ -2985,6 +2985,34 @@ impl Bank {
...
     }
 
+    pub fn weight_packet(&self, packet: &Packet) -> u64 {
+        if !packet.meta.is_simple_vote_tx {
+            return 0;
+        }
+
+        fn vote_address_from_packet(packet: &Packet) -> Option<Pubkey> {
+            let tx: VersionedTransaction =
+                limited_deserialize(&packet.data[0..packet.meta.size]).ok()?;
```
taozhu-chicago, line 2995: Possible to get account pubkey by continue counting packet offsets instead of calling deserialization.

taozhu-chicago (approved):
lgtm

runtime/src/bank.rs
jstarry: Yeah agreed, I think this part could have a negative impact on perf. I'm not sure if our perf tests will see hit it though since they don't process many vote tx's. We'll probably have to bake on v1.7

jstarry:
@sakridge did you get a chance to take a look?

jstarry:
Since mainnet is going to v1.7 soon, let's hold off on the backport

PR #20004, opened 2021-09-18
AcctIdx: introduce startup to bucket map

jeffwashington:
#### Summary of Changes

Fixes #


brooksprumo (approved):
I don't immediately see why the startup boolean needs to be atomic, but I'm guessing its uses will make that clear later.

PR #20005, opened 2021-09-18
Fixup atomic ordering on BucketStorage Header

brooksprumo:
Lock-type operations should be atomic::acquire, and unlock-type
operations should be atomic::release, so setting that here.

Additionally, change `unlock()` so it cannot fail.

These changes should strictly increase correctness/safety. There are no functional changes.

### DIFF
--- a/bucket_map/src/bucket.rs
+++ b/bucket_map/src/bucket.rs
@@ -240,7 +240,7 @@ impl<T: Clone + Copy> Bucket<T> {
                 if best_bucket.uid(ix) == 0 {
                     let elem_loc = elem.data_loc(current_bucket);
                     if elem.num_slots > 0 {
-                        current_bucket.free(elem_loc, elem_uid).unwrap();
+                        current_bucket.free(elem_loc, elem_uid);
                     }
                     // elem: &mut IndexEntry = self.index.get_mut(elem_ix);
                     elem.storage_offset = ix;
@@ -266,10 +266,10 @@ impl<T: Clone + Copy> Bucket<T> {
                 let data_bucket = &self.data[elem.data_bucket_ix() as usize];
                 let loc = elem.data_loc(data_bucket);
                 //debug!(                    "DATA FREE {:?} {} {} {}",                    key, elem.data_location, data_bucket.capacity, elem_uid                );
-                data_bucket.free(loc, elem_uid).unwrap();
+                data_bucket.free(loc, elem_uid);
             }
             //debug!("INDEX FREE {:?} {}", key, elem_uid);
-            self.index.free(elem_ix, elem_uid).unwrap();
+            self.index.free(elem_ix, elem_uid);
         }
     }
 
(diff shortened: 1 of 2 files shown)


bucket_map/src/bucket.rs
```diff
@@ -266,10 +266,10 @@ impl<T: Clone + Copy> Bucket<T> {
                 let data_bucket = &self.data[elem.data_bucket_ix() as usize];
                 let loc = elem.data_loc(data_bucket);
                 //debug!(                    "DATA FREE {:?} {} {} {}",                    key, elem.data_location, data_bucket.capacity, elem_uid                );
-                data_bucket.free(loc, elem_uid).unwrap();
+                data_bucket.free(loc, elem_uid);
             }
             //debug!("INDEX FREE {:?} {}", key, elem_uid);
-            self.index.free(elem_ix, elem_uid).unwrap();
+            self.index.free(elem_ix, elem_uid);
```
brooksprumo, line 272: As you can see, all calls to `free()` immediately call `unwrap()`, so errors are never handled, we just panic.

bucket_map/src/bucket_storage.rs
```diff
@@ -41,13 +41,10 @@ impl Header {
...
-                .compare_exchange(0, uid, Ordering::Relaxed, Ordering::Relaxed)
+                .compare_exchange(0, uid, Ordering::Acquire, Ordering::Relaxed)
     }
-    fn unlock(&self, uid: u64) -> bool {
-        Ok(uid)
-            == self
-                .lock
-                .compare_exchange(uid, 0, Ordering::Relaxed, Ordering::Relaxed)
+    fn unlock(&self) -> u64 {
+        self.lock.swap(0, Ordering::Release)
```
brooksprumo, line 47: 1. Always unlock, unconditionally
2. Store using `Release`
3. Return the old value so we can still check for errors

bucket_map/src/bucket_storage.rs
```diff
@@ -164,16 +160,16 @@ impl BucketStorage {
...
-                self.used.fetch_sub(1, Ordering::Relaxed);
-                e = Ok(());
-            }
-        };
-        e
+            let previous_owner = hdr.as_ref().unwrap().unlock();
+            assert_ne!(
+                previous_owner, 0,
+                "free: unlocked a header that was already unlocked"
+            );
```
brooksprumo, line 170: Since errors from `free()` were always unwrapped, change that to an assert here. There's no recovery built in anyway if `free()` were to fail.

brooksprumo (commented):
Added some comments to the code that I thought may be helpful.

jeffwashington:
I think this looks right. I need to think about it some.

bucket_map/src/bucket_storage.rs
brooksprumo: As per our call, I've fixed this assert.

jeffwashington (commented):
lgtm

PR #20006, opened 2021-09-18
call set_startup and add metrics on generate_index

jeffwashington:
#### Problem
Startup of the accounts index requires special handling. During startup, we want to write inserted items to disk as quickly as possible. Otherwise, we could run out of memory with large account counts. Also, there are no (or fewer) thread limits during this time since account index flushing will not interfere with validator tx processing.

#### Summary of Changes
during generate_index, set startup to true and then false so the accounts index understands operating parameters. Flush and thread throttling will respond to this later.
Fixes #


brooksprumo (commented):
Lgtm.

I think for now set_startup can stay with its current Ordering. If it changes usage to gate other operations, it may need to change then. (I know that wasn't changing in this PR, but I'm just noticing it now.)

PR #20007, opened 2021-09-18
Separate tpu vote port

sakridge:
#### Problem

Votes can get effectively de-prioritized in the regular transaction queue.

#### Summary of Changes

Add a special vote-only TPU port.

Fixes #


mvines:
I'd add `tpu_vote` into here as well:
https://github.com/solana-labs/solana/blob/f579f73700ceead94fb8c2783ebaa16afd3e45e1/validator/src/main.rs#L690

sakridge:
> I'd add `tpu_vote` into here as well:
> https://github.com/solana-labs/solana/blob/f579f73700ceead94fb8c2783ebaa16afd3e45e1/validator/src/main.rs#L690

added, thanks

core/src/tpu.rs
```diff
@@ -87,6 +92,13 @@ impl Tpu {
             SigVerifyStage::new(packet_receiver, verified_sender, verifier)
         };
 
+        let (vote_verified_sender, vote_verified_receiver) = unbounded();
+
+        let vote_sigverify_stage = {
+            let verifier = TransactionSigVerifier::new_reject_non_vote();
+            SigVerifyStage::new(vote_packet_receiver, vote_verified_sender, verifier)
```
jstarry, line 99: Sigverify stage already uses 4 threads for verification, can we just half of those instead of adding 4 more?

perf/src/sigverify.rs
```diff
@@ -113,6 +122,66 @@ fn verify_packet(packet: &mut Packet) {
         return;
     }
 
+    if reject_non_vote {
+        if packet_offsets.sig_len != 1 {
```
jstarry, line 126: Do we know that all voters use the same account for paying fees as they do for their vote authority? It would be twice as expensive to use different accounts, but just wanted to be sure.

perf/src/sigverify.rs
```diff
@@ -113,6 +122,66 @@ fn verify_packet(packet: &mut Packet) {
...
+        }
+
+        let vote_index = vote_index.unwrap();
+        // First instruction should be the vote key
+        if packet.data[instruction0_start] as u32 != vote_index {
+            packet.meta.discard = true;
+            return;
+        }
+
+        if instructions_len != 1 {
```
jstarry, line 179: Should be faster to first check that there's only one instruction as you're doing here and _then_ look up that ix's program id index than it is to scan through each pubkey checking it against the vote program id.

perf/src/sigverify.rs
```diff
@@ -113,6 +122,66 @@ fn verify_packet(packet: &mut Packet) {
         return;
     }
 
+    if reject_non_vote {
```
jstarry, line 125: I think we want this check in `get_packet_offsets` to ensure that the gpu sigverify path hits this code too

core/src/banking_stage.rs
```diff
@@ -1033,7 +1041,7 @@ impl BankingStage {
             .filter_map(|tx_index| {
                 let p = &msgs.packets[*tx_index];
                 let tx: Transaction = limited_deserialize(&p.data[0..p.meta.size]).ok()?;
-                if votes_only && vote_transaction::parse_vote_transaction(&tx).is_none() {
+                if votes_only && !solana_runtime::bank::is_simple_vote_transaction(&tx) {
```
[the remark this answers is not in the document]
taozhu-chicago: commit `3243acc26d`

core/src/tpu.rs
taozhu-chicago: How constrained the resources are by adding 4 more threads. If necessary, perhaps can add a parameter to specify number of threads when creating sigverify stage.

taozhu-chicago:
local-cluster test `test_fork_choice_refresh_old_votes` is failing, it looks like the test creates fork that has root distance > 500, resulting votes-only bank, from that point on there is no new root. Not sure how to address this a this moment, looking into if possible to adjust test so the root distance wouldn't b too large. @sakridge wdyt?

taozhu-chicago:
> local-cluster test `test_fork_choice_refresh_old_votes` is failing, it looks like the test creates fork that has root distance > 500, resulting votes-only bank, from that point on there is no new root. Not sure how to address this a this moment, looking into if possible to adjust test so the root distance wouldn't b too large. @sakridge wdyt?

cc @AshwinSekar

taozhu-chicago:
perf test looks good: https://solanalabs.slack.com/archives/CP2L2S4KV/p1632411380026400

taozhu-chicago:
> local-cluster test `test_fork_choice_refresh_old_votes` is failing, it looks like the test creates fork that has root distance > 500, resulting votes-only bank, from that point on there is no new root. Not sure how to address this a this moment, looking into if possible to adjust test so the root distance wouldn't b too large. @sakridge wdyt?

After long debugging, turns out the PR#20082 was not merged into this branch, resulting vote TX with 2 signatures were dropped for vote-only-bank, hence no new roots. After merge, the test passed locally.

jstarry (commented):
Can you also expose the tpu votes port in the dockerfile?

core/src/tpu.rs
taozhu-chicago: Agree that cli args are better, perhaps adding two args, one for number of threads for transaction sigverify, one for vote sigverify, after all, these two don't need to have same throughputs. Maybe in a separate PR?

jstarry: Separate PR sounds good

sakridge: The actual sigverify operation on CPU is done with a Rayon threadgroup, so the number of cores used is basically the same.

sakridge: Multiple threads helps with the GPU path. Also, the throughput does need to be basically the same, because you can get DOS'ed on any port and any port can deliver roughly the network bandwidth of the system.

core/src/banking_stage.rs
```diff
@@ -283,13 +289,17 @@ impl BankingStage {
...
         let bank_thread_hdls: Vec<JoinHandle<()>> = (0..num_threads)
             .map(|i| {
-                let (verified_receiver, enable_forwarding) = if i < num_threads - 1 {
-                    (verified_receiver.clone(), true)
-                } else {
-                    // Disable forwarding of vote transactions, as votes are gossiped
-                    (verified_vote_receiver.clone(), false)
+                let (verified_receiver, enable_forwarding) = match i.cmp(&(num_threads - 2)) {
+                    std::cmp::Ordering::Less => (verified_receiver.clone(), true),
+                    std::cmp::Ordering::Equal => (vote_verified_receiver.clone(), false),
```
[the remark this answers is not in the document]
sakridge: Forwarding them is probably a good idea.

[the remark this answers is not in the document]
taozhu-chicago: if to forward, how about something like this? https://github.com/sakridge/solana/pull/4

sdk/src/feature_set.rs
```diff
@@ -170,6 +170,11 @@ pub mod demote_program_write_locks {
     solana_sdk::declare_id!("3E3jV7v9VcdJL8iYZUMax9DiDno8j7EWUVbhm9RtShj2");
 }
 
+pub mod send_to_tpu_vote_port {
+    // todo: update
+    solana_sdk::declare_id!("3E3jV7v9VcdJL8iYZUMax9DiDno8j7EWUVbhm9RtShj3");
```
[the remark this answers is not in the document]
taozhu-chicago: done :)

core/src/tpu.rs
taozhu-chicago: #20170

taozhu-chicago:
> Can you also expose the tpu votes port in the dockerfile?

added `8009/udp` to `sdk/docker-solana/Dockerfile`

core/src/banking_stage.rs
```diff
@@ -282,13 +289,17 @@ impl BankingStage {
...
         // Many banks that process transactions in parallel.
+        assert!(num_threads >= MIN_THREADS_VOTES + MIN_THREADS_BANKING);
         let bank_thread_hdls: Vec<JoinHandle<()>> = (0..num_threads)
             .map(|i| {
-                let (verified_receiver, enable_forwarding) = if i < num_threads - 1 {
-                    (verified_receiver.clone(), true)
-                } else {
-                    // Disable forwarding of vote transactions, as votes are gossiped
-                    (verified_vote_receiver.clone(), false)
+                let (verified_receiver, enable_forwarding) = match i.cmp(&(num_threads - 2)) {
```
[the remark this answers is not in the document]
jstarry: @taozhu-chicago ping on this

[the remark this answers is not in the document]
sakridge: Changed this: https://github.com/solana-labs/solana/pull/20007/commits/d3f16f3c872694f360dad3268c3eaf5d9521991e

jstarry (approved):

PR #20013, opened 2021-09-19
WaitableCondVar supports notify_one

jeffwashington:
#### Problem
When this helper struct was created, it did not need to support notify_one. There are callers coming which need that api. It is a natural pass-through implementation of this wrapper.
#### Summary of Changes

Fixes #


jeffwashington:
approves as part of another pr

brooksprumo (approved):

PR #20014, opened 2021-09-19
Add ability to update entire vote state

carllin:
#### Problem
Out of order votes can lead to dramatic difference in lockouts between local and on-chain tower.

For instance imagine this secnario:

Current on-chain state: 1, 5
New state: 1, 2 (lockout: 4), 3, 5, 7

Imagine the validator made two of these votes:
1) The first vote {1, 2, 3} didn't land in the old state, but didn't
land on chain
2) A second vote {1, 2, 5} was then submitted, which landed

2 is not popped off in the local tower because 3 doubled the lockout.
However, 3 did not land in the on-chain state, so the vote {1, 2, 6}
will immediately pop off 2.

#### Summary of Changes
Introduce ability to update entire vote state via a single transaction

Fixes #

### DIFF
--- a/programs/vote/src/vote_state/mod.rs
+++ b/programs/vote/src/vote_state/mod.rs
@@ -375,6 +375,139 @@ impl VoteState {
         Ok(())
     }
 
+    // TODO: `Ensure check_slots_are_valid()` runs on the slots in `new_state`
+    // before this
+    pub fn process_new_vote_state(
+        &mut self,
+        new_state: VecDeque<Lockout>,
+        new_root: Option<Slot>,
+        timestamp: Option<i64>,
+        epoch: Epoch,
+    ) -> Result<(), VoteError> {
+        assert!(!new_state.is_empty());
+
+        // check_slots_are_valid()` ensures we don't process any states
+        // that are older than the current state
+        if !self.votes.is_empty() {
+            assert!(new_state.back().unwrap().slot > self.votes.back().unwrap().slot);
+        }
+
+        match (new_root, self.root_slot) {
+            (Some(new_root), Some(current_root)) => {
+                if new_root < current_root {
+                    return Err(VoteError::RootRollBack);
+                }
+            }
+            (None, Some(_)) => {
+                return Err(VoteError::RootRollBack);
+            }
+            _ => (),
+        }
+
+        let mut previous_vote: Option<&Lockout> = None;
+
+        for vote in &new_state {
+            if vote.confirmation_count == 0 {
+                return Err(VoteError::ZeroConfirmations);
+            } else if let Some(new_root) = new_root {
+                if vote.slot <= new_root {
+                    return Err(VoteError::SlotSmallerThanRoot(vote.slot, new_root));
+                }
+            }
+
+            if let Some(previous_vote) = previous_vote {
+                if previous_vote.slot >= vote.slot {
+                    return Err(VoteError::SlotsNotOrdered);
+                } else if previous_vote.confirmation_count >= vote.confirmation_count {
+                    return Err(VoteError::LockoutsNotOrdered);
+                } else if vote.slot > previous_vote.last_locked_out_slot() {
+                    return Err(VoteError::NewVoteStateLockoutMismatch(
+                        previous_vote.slot,
+                        previous_vote.confirmation_count,
+                        vote.slot,
+                    ));
+                }
+            }
+            previous_vote = Some(vote);
+        }
+
+        // Find the first non-common ancestor between our current state and `new_state`
+        let first_unshared_ancestor_index = self
+            .votes
+            .iter()
+            .enumerate()
+            .zip(new_state.iter())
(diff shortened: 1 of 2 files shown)


programs/vote/src/vote_state/mod.rs
```diff
@@ -345,6 +375,166 @@ impl VoteState {
...
+    //
+    // 2) From 1), this means that for every vote `s` in the current state:
+    //    a) If there exists an `s'` in `new_state` where `s.slot == s'.slot`, then
+    //    we must gaurantee `s.confirmations <= s'.confirmations`
+    //
+    //    b) If there does not exist any such `s'` in `new_state`, then there exists
+    //    some `t` that is the smallest vote in `new_state` where `t.slot > s.slot`.
+    //    `t` must have expired/popped off s', so it must be guaranteed that
+    //    `s.last_locked_out_slot() < t`.
+
```
carllin, line 403: The invariants necessary to guarantee safety, should have a discussion to see if these are sufficient

programs/vote/src/vote_state/mod.rs
```diff
@@ -345,6 +375,166 @@ impl VoteState {
...
+    // is a vote state that could have been created by iteratively building a tower
+    // by processing one vote at a time. For instance, the tower:
+    //
+    // { slot 0, confirmations: 31 }
+    // { slot 1, confirmations: 30 }
+    //
+    // is a legal tower that could be submitted on top of a previously empty tower. However,
+    // there is no way to create this tower from the iterative process, because slot 1 would
+    // have to have at least one other slot on top of it, even if the first 30 votes were all
+    // popped off.
```
carllin, line 414: A note about "impossible towers" allowed by this check

programs/vote/src/vote_state/mod.rs
```diff
@@ -345,6 +375,166 @@ impl VoteState {
...
+    //`Ensure check_slots_are_valid()` runs on the slots in `new_state`
+    // before `process_new_vote_state()` is called
+
+    // This function should guarantee the following about `new_state`:
+    //
+    // 1) It's well ordered, i.e. the slots are sorted from smallest to largest,
+    // and the confirmations sorted from largest to smallest.
+    // 2) Confirmations `c` on any vote slot satisfy `0 < c <= MAX_LOCKOUT_HISTORY`
+    // 3) Lockouts are not expired by consecutive votes, i.e. for every consecutive
+    // `v_i`, `v_{i + 1}` satisfy `v_i.last_locked_out_slot() <= v_{i + 1}`.
```
jstarry, line 387:
```suggestion
    // `v_i`, `v_{i + 1}` satisfy `v_i.last_locked_out_slot() >= v_{i + 1}`.
```

programs/vote/src/vote_state/mod.rs
```diff
@@ -301,32 +301,59 @@ impl VoteState {
...
+        let mut i = 0;
+
+        // index into the slot_hashes, starting at the oldest known
+        // slot hash
+        let mut j = slot_hashes.len();
+
+        // Note:
+        //
+        // 1) `vote.slots` is sorted from oldest/smallest vote to newest/largest
+        // vote, due to the way votes are applied to the vote state (newest vots
```
AshwinSekar, line 315: vots -> votes

programs/vote/src/vote_state/mod.rs
AshwinSekar: what if we check that the last slot has confirmations == 1?

programs/vote/src/vote_state/mod.rs
```diff
@@ -345,6 +375,166 @@ impl VoteState {
...
+
+    // We also guarantee that compared to the current vote state, `new_state`
+    // introduces no rollback. This means:
+    //
+    // 1) The last slot in `new_state` is always greater than any slot in the
+    // current vote state.
+    //
+    // 2) From 1), this means that for every vote `s` in the current state:
+    //    a) If there exists an `s'` in `new_state` where `s.slot == s'.slot`, then
+    //    we must gaurantee `s.confirmations <= s'.confirmations`
```
AshwinSekar, line 397: gaurantee -> guarantee

programs/vote/src/vote_state/mod.rs
carllin: Yeah that would work for this case, but there are a few other types of impossible towers it doesn't account for I think. For instance, impossible towers based on creating more lockout than the number of slots that have elapsed on this particular fork:

```
1 (confirmations: 4) -> 2 -> 5
```

bji:
I propose a different implementation here:

https://github.com/solana-labs/solana/issues/27473

bji:
Just a note: the original assertion:

"Out of order votes can lead to dramatic difference in lockouts between local and on-chain tower."

is not illustrated by the example given.  In the example given, all votes are cast and land in-order, but there is a fork switch.

So it's not clear to me if this issue is about out-of-order votes or mismatched lockout states between Tower and cluster on fork switch.  I assumed it was the latter since that's what the example was about.

PR #20020, opened 2021-09-20
add bucketmap 'insert' to eliminate copies

jeffwashington:
#### Summary of Changes

Fixes #

### DIFF
--- a/bucket_map/src/bucket.rs
+++ b/bucket_map/src/bucket.rs
@@ -366,19 +366,10 @@ impl<T: Clone + Copy> Bucket<T> {
         //debug!(            "INDEX_IX: {:?} uid:{} loc: {} cap:{}",            key,            uid,            location,            index.num_cells()        );
     }
 
-    pub fn update<F>(&mut self, key: &Pubkey, updatefn: F)
-    where
-        F: Fn(Option<(&[T], RefCount)>) -> Option<(Vec<T>, RefCount)>,
-    {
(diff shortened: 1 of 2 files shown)


bucket_map/src/bucket.rs
```diff
@@ -366,19 +366,10 @@ impl<T: Clone + Copy> Bucket<T> {
...
-        F: Fn(Option<(&[T], RefCount)>) -> Option<(Vec<T>, RefCount)>,
-    {
-        let current = self.read_value(key);
-        let new = updatefn(current);
-        if new.is_none() {
-            self.delete_key(key);
-            return;
-        }
-        let (new, refct) = new.unwrap();
+    pub fn insert(&mut self, key: &Pubkey, value: (&Vec<T>, RefCount)) {
```
brooksprumo, line 369: Can `value` be `&[T]` here instead?

bucket_map/src/bucket_map.rs
```diff
@@ -166,6 +166,21 @@ impl<T: Clone + Copy + Debug> BucketMap<T> {
         }
     }
 
+    /// Update Pubkey `key`'s value with 'value'
+    pub fn insert(&self, key: &Pubkey, value: (&Vec<T>, RefCount)) {
```
brooksprumo, line 170: And same thing here: (have `value` be `&[T]`)

brooksprumo (approved):

PR #20021, opened 2021-09-20
sigverify to identify and mark simple vote transaction

taozhu-chicago:
#### Problem

Changes to sigverify to identify simple vote transaction should be standalone PR

#### Summary of Changes
Pull the relevant changes from #19884 

Fixes #

### DIFF
--- a/perf/Cargo.toml
+++ b/perf/Cargo.toml
@@ -24,6 +24,7 @@ solana-budget-program = { path = "../programs/budget", version = "=1.6.27" }
 solana-logger = { path = "../logger", version = "=1.6.27" }
 solana-measure = { path = "../measure", version = "=1.6.27" }
 solana-metrics = { path = "../metrics", version = "=1.6.27" }
+solana-vote-program = { path = "../programs/vote", version = "=1.6.27" }
 curve25519-dalek = { version = "2" }
 
 [lib]
--- a/perf/src/sigverify.rs
+++ b/perf/src/sigverify.rs
@@ -46,15 +46,23 @@ struct PacketOffsets {
     pub sig_start: u32,
     pub msg_start: u32,
     pub pubkey_start: u32,
+    pub pubkey_len: u32,
 }
 
 impl PacketOffsets {
-    pub fn new(sig_len: u32, sig_start: u32, msg_start: u32, pubkey_start: u32) -> Self {
+    pub fn new(
+        sig_len: u32,
+        sig_start: u32,
+        msg_start: u32,
+        pubkey_start: u32,
+        pubkey_len: u32,
+    ) -> Self {
         Self {
             sig_len,
             sig_start,
             msg_start,
             pubkey_start,
+            pubkey_len,
         }
     }
 }
@@ -68,6 +76,7 @@ pub enum PacketError {
     MismatchSignatureLen,
     PayerNotWritable,
     InvalidProgramIdIndex,
+    InvalidProgramLen,
 }
 
 impl std::convert::From<std::boxed::Box<bincode::ErrorKind>> for PacketError {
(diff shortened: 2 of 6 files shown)


perf/src/sigverify.rs
```diff
@@ -309,17 +238,89 @@ fn do_get_packet_offsets(
...
+        .checked_add(pubkey_len_size)
+        .ok_or(PacketError::InvalidLen)?;
+
+    let instructions_len_offset = pubkey_len
+        .checked_mul(size_of::<Pubkey>())
+        .and_then(|v| v.checked_add(pubkey_start))
+        .and_then(|v| v.checked_add(size_of::<Hash>()))
+        .ok_or(PacketError::InvalidLen)?;
+
+    // Packet should have 1 more byte for instructions.len
```
jstarry, line 276: shortu16 can read up to 3 bytes (look up `MAX_ENCODING_LENGTH`)

jstarry: hmm but could there only be 1 byte remaining if the length is 0. Maybe we skip the `decode_shortu16_len` call and read the next byte directly?

perf/src/sigverify.rs
```diff
@@ -309,17 +238,89 @@ fn do_get_packet_offsets(
...
+
+    // Packet should have at least 2 more byte for one instructions_program_id
+    let _ = instruction_start
+        .checked_add(2usize)
+        .filter(|v| *v <= packet.meta.size)
+        .ok_or(PacketError::InvalidLen)?;
+
+    let (instruction_program_id_index, _size) =
+        decode_shortu16_len(&packet.data[instruction_start..])
+            .map_err(|_| PacketError::InvalidLen)?;
```
jstarry, line 303: This is actually just a `u8`

perf/src/sigverify.rs
```diff
@@ -309,17 +238,89 @@ fn do_get_packet_offsets(
...
+
+    if instruction_program_id_index >= pubkey_len {
+        return Err(PacketError::InvalidPubkeyLen);
```
jstarry, line 306: Maybe add a new error for this, the pubkey len could be fine, maybe just the program id index that's corrupted

perf/src/sigverify.rs
```diff
@@ -309,17 +238,89 @@ fn do_get_packet_offsets(
...
+    }
+
+    let instruction_id_start = instruction_program_id_index
```
jstarry, line 309: nit: can you name these variables `instruction_program_id_start` and `instruction_program_id_end`

perf/src/sigverify.rs
taozhu-chicago: per our chat, will add a function does something like this: is_length_eq(&Packet, offset, value) -> bool

jstarry: Sorry, disregard this. If you look at `decode_shortu16_len`, it will error if there aren't enough bytes to read from, so this is safe.

taozhu-chicago: > Sorry, disregard this. If you look at `decode_shortu16_len`, it will error if there aren't enough bytes to read from, so this is safe.

lovely

perf/src/sigverify.rs
taozhu-chicago: done

perf/src/sigverify.rs
taozhu-chicago: added `InvalidProgramIdIndex`

perf/src/sigverify.rs
taozhu-chicago: done

perf/Cargo.toml
```diff
@@ -24,6 +24,7 @@ solana-budget-program = { path = "../programs/budget", version = "=1.6.27" }
 solana-logger = { path = "../logger", version = "=1.6.27" }
 solana-measure = { path = "../measure", version = "=1.6.27" }
 solana-metrics = { path = "../metrics", version = "=1.6.27" }
+solana-vote-program = { path = "../programs/vote", version = "=1.6.27" }
```
jstarry, line 27: don't need this dependency, the vote program id is in `solana_sdk::vote::program::id`

perf/src/sigverify.rs
```diff
@@ -237,17 +239,88 @@ fn do_get_packet_offsets(
...
+        .checked_add(MESSAGE_HEADER_LENGTH)
+        .ok_or(PacketError::InvalidLen)?;
+
+    let (pubkey_len, pubkey_len_size) =
+        decode_shortu16_len(&packet.data[message_account_keys_len_offset..])
+            .map_err(|_| PacketError::InvalidLen)?;
+
+    let pubkey_start = message_account_keys_len_offset
+        .checked_add(pubkey_len_size)
+        .ok_or(PacketError::InvalidLen)?;
```
jstarry, line 269: Can you mirror @sakridge's changes [here](https://github.com/solana-labs/solana/pull/20007/files#diff-c4686ade7f5e09568084fc8e5ff4df62961724ef52932e93a759e46d609b0d2fR49) which add a `pubkey_len` field to `PacketOffsets` and use that and the `pubkey_start` offset to replace this code.

perf/src/sigverify.rs
```diff
@@ -237,17 +239,88 @@ fn do_get_packet_offsets(
...
+        }
         offsets
     } else {
         // force sigverify to fail by returning zeros
         PacketOffsets::new(0, 0, 0, 0)
     }
 }
 
-pub fn generate_offsets(batches: &[Packets], recycler: &Recycler<TxOffset>) -> TxOffsets {
+fn check_for_simple_vote_transaction(
```
jstarry, line 255: This deserves full unit test coverage

perf/Cargo.toml
taozhu-chicago: It is better to ref to vote id with full path.

jstarry: Ah ok that's fine then

perf/src/sigverify.rs
taozhu-chicago: done

perf/src/sigverify.rs
taozhu-chicago: absolutely, done.

jstarry (approved):

perf/src/sigverify.rs
```diff
@@ -234,20 +245,88 @@ fn do_get_packet_offsets(
...
+    packet: &mut Packet,
+    packet_offsets: &PacketOffsets,
+    current_offset: usize,
+) -> Result<(), PacketError> {
+    if packet_offsets.sig_len != 1 {
```
jstarry, line 268: FYI (+ @sakridge) there are at least two validators on mainnet that use 2 sig votes:
- https://explorer.solana.com/address/p2pUJruVeKTuZ41zbGWcxienmzRFnUa9fjBHigsR3QC
- https://explorer.solana.com/address/DWvDTSh3qfn88UoQTEKRV2JnLt5jtJAVoiCo3ivtMwXP

I think it would be better to remove this restriction

taozhu-chicago: feels like checking number of sigs is a quick way to rule out vote, if it is not 100%, might as well remove this check. Unless `sig_len==1 || sig_len == 2` is all there is. 
BTW, how do I see the 2 sigs vote in via above links?

sakridge: Yea. `sig_len == 1 || sig_len == 2` should be fine, more than 2 doesn't make sense. Could be `sig_len <= 2` because I think we exclude `sig_len == 0` already in another place.

jstarry: > feels like checking number of sigs is a quick way to rule out vote, if it is not 100%, might as well remove this check. Unless `sig_len==1 || sig_len == 2` is all there is.
> BTW, how do I see the 2 sigs vote in via above links?

Any of those account's transactions have two account signers

taozhu-chicago: OK, putting in a new PR for this.

jstarry:
@taozhu-chicago I'm porting this change to master fyi

PR #20024, opened 2021-09-20
AcctIdx: age adds time component

jeffwashington:
#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/bucket_map_holder.rs
+++ b/runtime/src/bucket_map_holder.rs
@@ -3,17 +3,24 @@ use crate::bucket_map_holder_stats::BucketMapHolderStats;
 use crate::in_mem_accounts_index::SlotT;
 use crate::waitable_condvar::WaitableCondvar;
 use solana_bucket_map::bucket_map::{BucketMap, BucketMapConfig};
+use solana_sdk::clock::SLOT_MS;
+use solana_sdk::timing::AtomicInterval;
 use std::fmt::Debug;
 use std::sync::atomic::{AtomicBool, AtomicU8, AtomicUsize, Ordering};
 use std::sync::Mutex;
 pub type Age = u8;
 
+const AGE_MS: u64 = SLOT_MS; // match one age per slot time
+
+// will eventually hold the bucket map
 pub struct BucketMapHolder<T: IndexValue> {
     pub disk: Option<BucketMap<SlotT<T>>>,
 
     pub count_ages_flushed: AtomicUsize,
     pub age: AtomicU8,
     pub stats: BucketMapHolderStats,
+
+    age_timer: AtomicInterval,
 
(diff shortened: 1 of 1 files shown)


runtime/src/bucket_map_holder.rs
```diff
@@ -3,18 +3,24 @@ use crate::bucket_map_holder_stats::BucketMapHolderStats;
 use crate::in_mem_accounts_index::SlotT;
 use crate::waitable_condvar::WaitableCondvar;
 use solana_bucket_map::bucket_map::{BucketMap, BucketMapConfig};
+use solana_sdk::timing::AtomicInterval;
 use std::fmt::Debug;
 use std::sync::atomic::{AtomicBool, AtomicU8, AtomicUsize, Ordering};
 use std::sync::Mutex;
 pub type Age = u8;
 
+const AGE_MS: u64 = 400; // 400 ms per age increment, which matches slot time
```
brooksprumo, line 12: Is there already a slot time constant somewhere? Or, is the intent to have the Age time be tunable and different from the slot time?

runtime/src/bucket_map_holder.rs
```diff
@@ -53,6 +59,12 @@ impl<T: IndexValue> BucketMapHolder<T> {
     pub fn future_age_to_flush(&self) -> Age {
         self.current_age().wrapping_add(self.ages_to_stay_in_cache)
     }
+
+    fn has_age_interval_elapsed(&self) -> bool {
+        // note that when this returns true, state of age_timer is modified
+        self.age_timer.should_update(AGE_MS) // todo: make sure this interval has correct Ordering
```
brooksprumo, line 65: Should this todo be resolved before the PR gets merged?

jeffwashington: I was hoping you'd help me resolve this. I added tests and it appears to work. should_update uses Ordering::Relaxed.

runtime/src/bucket_map_holder.rs
jeffwashington: I wanted it different, but logically the same. Maybe it does make sense to use the clock constant. I'll investigate.

jeffwashington: I made it the same constant

runtime/src/bucket_map_holder.rs
jeffwashington: I got rid of the comment for now. Please do look at AtomicInterval and see if the orderings make sense.

brooksprumo: Gotcha, I'll take a look

brooksprumo: I think this looks good

brooksprumo (approved):

PR #20034, opened 2021-09-20
CPI without Account Refs

Lichtso:
#### Problem
#15410 introduced [this](https://github.com/solana-labs/solana/blob/7a785e067ac109bb065ea1170af4bd9f896adfc2/runtime/src/message_processor.rs#L178) `unsafe` expression.

#### Summary of Changes
* Removes search for accounts and `unsafe` lifetime transmute in `InvokeContext::push()`.
* Replaces account parameters of all CPI methods by indices, as they are easier to verify and compatible with ABIv2 (see #19191).

Fixes #

### DIFF
--- a/runtime/src/message_processor.rs
+++ b/runtime/src/message_processor.rs
@@ -212,7 +212,7 @@ impl<'a> InvokeContext for ThisInvokeContext<'a> {
         let (mut pre_sum, mut post_sum) = (0_u128, 0_u128);
         let mut work = |_unique_index: usize, index_in_instruction: usize| {
             if index_in_instruction < write_privileges.len()
-                && index_in_instruction < accounts.len()
+                && index_in_instruction < account_indices.len()
             {
                 let account_index = account_indices[index_in_instruction];
                 let (key, account) = &accounts[account_index];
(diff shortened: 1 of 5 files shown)


runtime/src/message_processor.rs
```diff
@@ -214,24 +194,63 @@ impl<'a> InvokeContext for ThisInvokeContext<'a> {
...
-        )
+        let accounts = &self.accounts;
+        let pre_accounts = &mut self.pre_accounts;
+        let timings = &mut self.timings;
+
+        // Verify the per-account instruction results
+        let (mut pre_sum, mut post_sum) = (0_u128, 0_u128);
+        let mut work = |_unique_index: usize, index_in_instruction: usize| {
+            if index_in_instruction < write_privileges.len()
+                && index_in_instruction < accounts.len()
```
jstarry, line 215:
```suggestion
                && index_in_instruction < account_indices.len()
```
I think this should be updated

jstarry (commented):
Nice cleanup, noticed something while reading your change

runtime/src/message_processor.rs
Lichtso: Good find!

jstarry (approved):
Looks great, it also looks like this change makes https://github.com/solana-labs/solana/issues/9711 trivial to implement

PR #20035, opened 2021-09-20
AcctIdx: delete from disk index

jeffwashington:
#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/in_mem_accounts_index.rs
+++ b/runtime/src/in_mem_accounts_index.rs
@@ -126,6 +126,14 @@ impl<T: IndexValue> InMemAccountsIndex<T> {
         if let Entry::Occupied(index_entry) = entry {
             if index_entry.get().slot_list.read().unwrap().is_empty() {
                 index_entry.remove();
+                // note there is a potential race here that has existed.
+                // if someone else holds the arc,
+                //  then they think the item is still in the index and can make modifications.
+                // We have to have a write lock to the map here, which means nobody else can get
+                //  the arc, but someone may already have retreived a clone of it.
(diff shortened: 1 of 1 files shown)


runtime/src/in_mem_accounts_index.rs
```diff
@@ -126,6 +126,14 @@ impl<T: IndexValue> InMemAccountsIndex<T> {
...
             if index_entry.get().slot_list.read().unwrap().is_empty() {
                 index_entry.remove();
+                // note there is a potential race here that has existed.
+                // if someone else holds the arc,
+                //  then they think the item is still in the index and can make modifications.
+                // We have to have a write lock to the map here, which means nobody else can get
+                //  the arc, but someone may already have retreived a clone of it.
+                if let Some(disk) = self.storage.disk.as_ref() {
+                    disk.delete_key(&pubkey)
+                }
```
brooksprumo, line 136: Is there still a race now? If yes, does deleting the key here add any risks?

jeffwashington: I believe the race existed previously and still exists now. I believe the current change has no impact on the correctness or incorrectness of the existing behavior. The result either way is that a client has a strong count to an arc where they can still add/unref or modify the slot list, but that modification will never get stored in the accounts index and will never be returned by a 'get', 'entry', etc. by anyone else.

jeffwashington: I figure an existing race that is now understood and documented is better than an undocumented or unknown race. ;-)

brooksprumo (approved):

PR #20037, opened 2021-09-20
Make bigtable delete-slots actually usable

ryoqun:
#### Problem

seems last-minute review adjustments made `solana-ledger-tool bigtable delete-slots` not workable at all (with/without `--force`)

#### Summary of Changes

context: https://github.com/solana-labs/solana/pull/19931

### DIFF
--- a/ledger-tool/src/bigtable.rs
+++ b/ledger-tool/src/bigtable.rs
@@ -439,7 +439,7 @@ pub fn bigtable_process_command(ledger_path: &Path, matches: &ArgMatches<'_>) {
         }
         ("delete-slots", Some(arg_matches)) => {
             let slots = values_t_or_exit!(arg_matches, "slots", Slot);
-            let dry_run = !value_t_or_exit!(arg_matches, "force", bool);
+            let dry_run = !arg_matches.is_present("force");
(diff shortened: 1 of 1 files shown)


ledger-tool/src/bigtable.rs
```diff
@@ -439,7 +439,7 @@ pub fn bigtable_process_command(ledger_path: &Path, matches: &ArgMatches<'_>) {
         }
         ("delete-slots", Some(arg_matches)) => {
             let slots = values_t_or_exit!(arg_matches, "slots", Slot);
-            let dry_run = !value_t_or_exit!(arg_matches, "force", bool);
```
ryoqun, line 442: no way to go past this `..._or_exit!()` given this:

https://github.com/solana-labs/solana/blob/d6d2840c9442420831a692391109b3c243b1d968/ledger-tool/src/bigtable.rs#L292

CriesofCarrots: Great catch

jstarry: Sorry about that lack of testing

jstarry (approved):

PR #20044, opened 2021-09-20
Fix memo handling and "blocks" key formatting

ryoqun:
#### Problem

While trying  to use `bigtable delete-slots`, there was some problems:

- Memo extraction was recently implemented. Old blocks can have been uploaded prior that. In that case, we should skip memo extraction to avoid unintended removal omission due to false mismatch.
- it seems that "blocks" bigtable key formatting code is simply out of sync (haven't fully fact-checked. just from code-reading). `slot.to_string()` would be decimal formatted; but it needs to be hex-formatted according to this: https://github.com/solana-labs/solana/blob/e12e8bc431c69bc9be7e32da4547cb1698e77e01/storage-bigtable/src/lib.rs#L632-L635

not tested at all. :) just barely compiles

#### Summary of Changes

context: #19931

### DIFF
--- a/ledger-tool/src/bigtable.rs
+++ b/ledger-tool/src/bigtable.rs
@@ -43,13 +43,13 @@ async fn upload(
     .await
 }
 
-async fn delete_slots(slots: Vec<Slot>, extract_memo: bool, dry_run: bool) -> Result<(), Box<dyn std::error::Error>> {
+async fn delete_slots(slots: Vec<Slot>, dry_run: bool) -> Result<(), Box<dyn std::error::Error>> {
     let read_only = dry_run;
     let bigtable = solana_storage_bigtable::LedgerStorage::new(read_only, None)
         .await
         .map_err(|err| format!("Failed to connect to storage: {:?}", err))?;
 
-    solana_ledger::bigtable_delete::delete_confirmed_blocks(bigtable, slots, extract_memo, dry_run).await
+    solana_ledger::bigtable_delete::delete_confirmed_blocks(bigtable, slots, dry_run).await
 }
 
 async fn first_available_block() -> Result<(), Box<dyn std::error::Error>> {
@@ -287,13 +287,6 @@ impl BigTableSubCommand for App<'_, '_> {
                                     .help("Slots to delete"),
                                 )
                             .arg(
-                                Arg::with_name("extract_memo")
-                                    .long("extract-memo")
-                                    .takes_value(false)
-                                    .help(
-                                        "Old transaction could have been persisted without memo, needed to adjust unmatched delete skip behavior"),
-                            )
-                            .arg(
                                 Arg::with_name("force")
                                     .long("force")
                                     .takes_value(false)
@@ -446,9 +439,8 @@ pub fn bigtable_process_command(ledger_path: &Path, matches: &ArgMatches<'_>) {
         }
         ("delete-slots", Some(arg_matches)) => {
             let slots = values_t_or_exit!(arg_matches, "slots", Slot);
-            let extract_memo = arg_matches.is_present("extract_memo");
             let dry_run = !arg_matches.is_present("force");
-            runtime.block_on(delete_slots(slots, extract_memo, dry_run))
+            runtime.block_on(delete_slots(slots, dry_run))
         }
         ("first-available-block", Some(_arg_matches)) => runtime.block_on(first_available_block()),
         ("block", Some(arg_matches)) => {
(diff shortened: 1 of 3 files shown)


storage-bigtable/src/lib.rs
```diff
@@ -695,10 +699,12 @@ impl LedgerStorage {
                     Some(Ok(fetched_tx_info)) if fetched_tx_info == &expected_tx_info => {
                         deletion_rows.push(signature);
                     }
-                    Some(Ok(_)) => {
+                    Some(Ok(fetched_tx_info)) => {
                         warn!(
-                            "skipped tx row {} because the bigtable entry did not match",
-                            signature
+                            "skipped tx row {} because the bigtable entry ({:?}) did not match to {:?}",
```
ryoqun, line 704: more helpful debugging.

storage-bigtable/src/lib.rs
```diff
@@ -629,7 +633,7 @@ impl LedgerStorage {
         // Store the block itself last, after all other metadata about the block has been
         // successfully stored.  This avoids partial uploaded blocks from becoming visible to
         // `get_confirmed_block()` and `get_confirmed_blocks()`
-        let blocks_cells = [(slot_to_key(slot), confirmed_block.into())];
+        let blocks_cells = [(slot_to_blocks_key(slot), confirmed_block.into())];
         bytes_written += self
             .connection
             .put_protobuf_cells_with_retry::<generated::ConfirmedBlock>("blocks", &blocks_cells)
```
ryoqun, line 639: here

storage-bigtable/src/lib.rs
```diff
@@ -731,7 +737,7 @@ impl LedgerStorage {
             }
 
             self.connection
-                .delete_rows_with_retry("blocks", &[slot.to_string()])
```
ryoqun, line 734: unless i'm reading code wrongly, this should be inconsistent with https://github.com/solana-labs/solana/pull/20044/files#r712516629?

storage-bigtable/src/lib.rs
```diff
@@ -652,7 +656,7 @@ impl LedgerStorage {
             let signature = transaction.signatures[0];
             let index = index as u32;
             let err = meta.as_ref().and_then(|meta| meta.status.clone().err());
-            let memo = extract_and_fmt_memos(&transaction.message);
+            let memo = if extract_memo { extract_and_fmt_memos(&transaction.message) } else { None };
```
ryoqun, line 659: unlike the `"blocks"` bug, I confirmed this is a bug.

ledger-tool/src/bigtable.rs
```diff
@@ -286,6 +286,13 @@ impl BigTableSubCommand for App<'_, '_> {
                                     .required(true)
                                     .help("Slots to delete"),
                                 )
+                            .arg(
+                                Arg::with_name("extract_memo")
+                                    .long("extract-memo")
```
ryoqun, line 291: maybe, saner default value could be "`true`". or  make this is required to force ops to think hard for this.

ryoqun:
@jstarry if you can polish this up, it would be great. :)

storage-bigtable/src/lib.rs
```diff
@@ -70,6 +70,10 @@ fn slot_to_key(slot: Slot) -> String {
     format!("{:016x}", slot)
 }
 
+fn slot_to_blocks_key(slot: Slot) -> String {
```
ryoqun, line 73: intention here is that DRY.

hopefully, `slot_to_tx_by_addr_key` and etc can be defined as well. (the `!slot` bit-or is quite easy to forget to do properly at call site...)

storage-bigtable/src/lib.rs
jstarry: Correct, thanks

jstarry:
I think it's sufficient to check tx rows by slot and inclusion index (but sure err too) and skip meta for the equality check since it was added recently. I added a commit that should make operation of this cli tool a little less complex and straightforward

storage-bigtable/src/lib.rs
ryoqun: done 429bbcb152b71f24f75c2dcb1534d75fb21cadc1

ryoqun:
@jstarry thanks for simplifying the memo bug fix! I like it. Could you sanity-check the pr? I think it's ready for merging and another cherry-pick onto the my custom build of `solana-ledger-tool`. :)

jstarry (approved):

PR #20047, opened 2021-09-20
AccountsDb plugin framework

lijunwangs:
#### Problem

It has been observed that a validator loaded with accounts queries via RPC services can fall behind the network. We have been working on replicating the whole AccountsDb to another node to serve the RPC calls -- #16986. That general solution of replicating everything can introduce additional unnecessary load on the validator as most use cases are only interested in subset of the programs and accounts. The alternative approach is to replicate the accounts update to external databases (Kafka, Postgres etc) for user defined portions of accounts.

#### Summary of Changes
Create a plugin mechanism in the accounts update path so that accounts data can be streamed out to external data stores (be it Kafka or Postgres). The plugin mechanism allows

Data stores of connection strings/credentials to be configured,
Accounts with patterns to be streamed
PostgreSQL implementation of the streaming for different destination stores to be plugged in.

The code comprises 4 major parts:

1. accountsdb-plugin-intf: defines the plugin interface which concrete plugin should implement.
2. accountsdb-plugin-manager: manages the load/unload of plugins and provide interfaces which the validator can notify of accounts update to plugins.
3. accountsdb-plugin-postgres: the concrete plugin implementation for PostgreSQL
4. The validator integrations: updated streamed right after snapshot restore and after account update from transaction processing or other real updates.

The plugin is optionally loaded on demand by new validator CLI argument -- there is no impact if the plugin is not loaded.

Fixes #


accountsdb-plugin-manager/src/accounts_update_notifier.rs
```diff
@@ -0,0 +1,215 @@
...
+
+    pub fn notify_slot_confirmed(&self, slot: Slot) {
+        let mut plugin_manager = self.plugin_manager.write().unwrap();
+        if plugin_manager.plugins.is_empty() {
+            return;
+        }
+
+        match self.bank_forks.read().unwrap().get(slot) {
+            None => {
+                info!("The slot is not found {:?}", slot);
```
CriesofCarrots, line 178: Should this be a little stronger than `info!`?

accountsdb-plugin-manager/src/accounts_update_notifier.rs
```diff
@@ -0,0 +1,215 @@
...
+    let data = stored_account_meta.data.to_vec();
+    Some(ReplicaAccountInfo {
+        account_meta,
+        hash: bs58::encode(stored_account_meta.hash.0).into_string(),
+        data,
+    })
+}
+
+#[allow(dead_code)]
+fn accountinfo_from_readable_account(
```
CriesofCarrots, line 103: What's to come for this?

accountsdb-plugin-manager/src/accountsdb_plugin_service.rs
```diff
@@ -0,0 +1,131 @@
...
+    ///    The `accounts_selector` section allows the user to controls accounts selections.
+    ///    accounts_selector = {
+    ///         accounts = \['pubkey-1', 'pubkey-2', ..., 'pubkey-n'\],
+    ///    }
+    ///    or:
+    ///    accounts_selector = {
+    ///         owners = \['pubkey-1', 'pubkey-2', ..., 'pubkey-m'\]
+    ///    }
+    ///    When accounts and owners are specified, the program only filters only using accounts
+    ///    as it is more selective. And owners field is ignored. When only owners is specified,
```
CriesofCarrots, line 48: I had some editing suggestions here, but first:
Is this true that the owners field is ignored? It looks to me that due to `AccountsSelector::is_account_selected()`, the results will be the union of accounts specified in `accounts` and those owned by `owners`.

accountsdb-plugin-postgres/src/accountsdb_plugin_postgres.rs
```diff
@@ -0,0 +1,146 @@
...
+        info!("Updating account {:?} at slot {:?}", account, slot);
+
+        match &mut self.client {
+            None => {
+                return Err(AccountsDbPluginError::DataStoreConnectionError {
+                    msg: "There is no connection to the PostgreSQL database.".to_string(),
+                });
+            }
+            Some(client) => {
+                let slot = slot as i64; // postgres only support i64
```
CriesofCarrots, line 88:
```suggestion
                let slot = slot as i64; // postgres only supports i64
```

accountsdb-plugin-intf/src/accountsdb_plugin_intf.rs
```diff
@@ -0,0 +1,71 @@
...
+}
+
+impl Eq for ReplicaAccountInfo {}
+
+#[derive(Clone, Default, PartialEq, Debug)]
+pub struct ReplicaAccountInfo {
+    pub account_meta: ReplicaAccountMeta,
+    pub hash: String,
+    pub data: Vec<u8>,
+}
```
CriesofCarrots, line 25: Any plans to dedupe with replica-lib?

accountsdb-plugin-manager/src/accountsdb_plugin_service.rs
```diff
@@ -0,0 +1,131 @@
...
+    ///    It is usually used to configure the connection information for the external data store.
+    ///    The accounts field support wildcard to select all accounts:
+    ///    accounts_selector = {
+    ///         accounts = \['*'\],
+    ///    }
+    pub fn new(
+        confirmed_bank_receiver: Receiver<Slot>,
+        bank_forks: Arc<RwLock<BankForks>>,
+        accountsdb_plugin_config_file: &Path,
+    ) -> Self {
```
CriesofCarrots, line 60: Based on the number of `unwrap`s below, it looks like this should probably return a Result and handle Errors gracefully.

CriesofCarrots (commented):
Here's my first pass. Mostly interface comments; the meat of this PR looks pretty good. Thanks for taking this one!

accountsdb-plugin-intf/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+authors = ["Solana Maintainers <maintainers@solana.foundation>"]
+edition = "2018"
+name = "solana-accountsdb-plugin-intf"
```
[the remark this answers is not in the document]
lijunwangs: Will do

accountsdb-plugin-intf/src/accountsdb_plugin_intf.rs
```diff
@@ -0,0 +1,71 @@
...
+
+#[derive(Error, Debug)]
+pub enum AccountsDbPluginError {
+    #[error("Error with opening config file.")]
```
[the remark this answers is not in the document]
lijunwangs: Will do

accountsdb-plugin-manager/src/accounts_update_notifier.rs
lijunwangs: This is fixed

accountsdb-plugin-manager/src/accountsdb_plugin_service.rs
lijunwangs: Corrected.

accountsdb-plugin-intf/src/accountsdb_plugin_intf.rs
lijunwangs: Yes. I am planning to use the 3rd installment of AccountsDb replication PR to address this

accountsdb-plugin-intf/src/accountsdb_plugin_intf.rs
```diff
@@ -0,0 +1,71 @@
+use std::any::Any;
+use std::io;
+/// The interface for AccountsDb plugins. A plugin must implement
+/// the AccountsDbPlugin trait to work with the Solana Validator.
+/// In addition the dynamic libraray must export a "C" function _create_plugin which
+/// creates the implementation of the plugin.
+use thiserror::Error;
```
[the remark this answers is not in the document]
lijunwangs: Done

aeyakovenko:
@lijunwangs had a request to be able to also pipe the transaction logs to a DB

runtime/src/accounts_db.rs
```diff
@@ -937,6 +937,26 @@ struct RemoveUnrootedSlotsSynchronization {
...
+
+    /// Notified when a slot is marked frozen.
+    fn notify_slot_processed(&self, slot: Slot, parent: Option<Slot>);
+
+    /// Notified when a slot is rooted.
+    fn notify_slot_rooted(&self, slot: Slot, parent: Option<Slot>);
+}
+
+pub type AccountsUpdateNotifier = Arc<RwLock<dyn AccountsUpdateNotifierIntf + Sync + Send>>;
+
```
carllin, line 959: I think this might make sense to move to a separate module

accountsdb-plugin-postgres/src/accountsdb_plugin_postgres.rs
```diff
@@ -0,0 +1,340 @@
...
+    /// Unload all plugins and loaded plugin libraries, making sure to fire
+    /// their `on_plugin_unload()` methods so they can do any necessary cleanup.
+    fn on_unload(&mut self) {
+        info!("Unloading plugin: {:?}", self.name());
+    }
+
+    fn update_account(&mut self, account: &ReplicaAccountInfo, slot: u64) -> Result<()> {
+        if let Some(accounts_selector) = &self.accounts_selector {
+            if !accounts_selector
+                .is_account_selected(&account.account_meta.pubkey, &account.account_meta.owner)
```
carllin, line 156: It would be nice if we could avoid the clones in `ReplicaAccountInfo` until we're sure the account is selected

lijunwangs: Yes -- I am working on that in the follow-on PR for performance. The ReplicaAccountInfo is still used -- I like the encapsulation -- would like to reduce exposing the internal data structures defined in the core validator code-- but the fields values (such as pubkey, owner, data) are using reference instead of copy so that we avoid the encoding.

runtime/src/accounts_db.rs
```diff
@@ -5921,6 +5950,16 @@ impl AccountsDb {
...
     pub fn store_cached(&self, slot: Slot, accounts: &[(&Pubkey, &AccountSharedData)]) {
         self.store(slot, accounts, self.caching_enabled);
+
+        if let Some(accounts_update_notifier) = &self.accounts_update_notifier {
+            let notifier = &accounts_update_notifier.read().unwrap();
+
+            for account in accounts {
+                let pubkey = account.0;
+                let account = account.1;
+                notifier.notify_account_update(slot, pubkey, account);
```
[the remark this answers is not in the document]
lijunwangs: Yes. I am looking into the parallel processing in the follow on PR.

runtime/src/accounts_db.rs
lijunwangs: okay

carllin:
I think it's a good idea to break up the 

1) Interface
2) Changes to AccountsDb

That way we can check in the interface code without having to incur any performance hits until the performance optimizations are in.

accountsdb-plugin-interface/src/accountsdb_plugin_interface.rs
```diff
@@ -0,0 +1,78 @@
...
+    pub executable: bool,
+    pub rent_epoch: u64,
+}
+
+impl Eq for ReplicaAccountInfo {}
+
+#[derive(Clone, Default, PartialEq, Debug)]
+pub struct ReplicaAccountInfo {
+    pub account_meta: ReplicaAccountMeta,
+    pub data: Vec<u8>,
```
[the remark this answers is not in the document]
mvines: I mean the answer is yes, we will be held to backwards compatibility where whether we want to or not

accountsdb-plugin-postgres/src/accountsdb_plugin_postgres.rs
```diff
@@ -0,0 +1,340 @@
...
+            {
+                return Ok(());
+            }
+        }
+        if self.accounts_selector.is_none() {
```
t-nelson, line 161:
```suggestion
        } else {
```

runtime/src/accounts_db.rs
```diff
@@ -937,6 +937,26 @@ struct RemoveUnrootedSlotsSynchronization {
 
 type AccountInfoAccountsIndex = AccountsIndex<AccountInfo>;
 
+pub trait AccountsUpdateNotifierIntf: std::fmt::Debug {
+    /// Notified when an account is updated at runtime, due to transaction activities
+    fn notify_account_update(&self, slot: Slot, pubkey: &Pubkey, account: &AccountSharedData);
+
+    /// Notified when the AccountsDb is initialized at start when restored
+    /// from a snapshot.
+    fn notify_account_data_at_start(&self, slot: Slot, account: &StoredAccountMeta);
```
t-nelson, line 946:
```suggestion
    fn notify_account_restore_from_snapshot(&self, slot: Slot, account: &StoredAccountMeta);
```

validator/src/main.rs
```diff
@@ -2564,6 +2579,21 @@ pub fn main() {
         None
     };
 
+    let accountsdb_plugin_config_file = if matches.is_present("enable_accountsdb_plugin") {
+        Some(PathBuf::from(
+            matches
+                .value_of("accountsdb_plugin_config")
+                .unwrap_or_else(|| {
+                    eprintln!(
+                        "The --accountsdb-plugin-config <plugin_config_file> is required when --enable-accountsdb-plugin is set.",
```
t-nelson, line 2588: `clap` can enforce this for us. https://docs.rs/clap/2.33.3/clap/struct.Arg.html#method.requires

t-nelson (commented):
This is looking really good!  Most of my comments are around generalizing the interface a bit so as not to assume it's going to necessarily be feeding a database, rather any application that might want to consume account update events

accountsdb-plugin-manager/src/accounts_update_notifier.rs
```diff
@@ -0,0 +1,128 @@
...
+        pubkey::Pubkey,
+    },
+    std::sync::{Arc, RwLock},
+};
+#[derive(Debug)]
+pub(crate) struct AccountsUpdateNotifierImpl {
+    plugin_manager: Arc<RwLock<AccountsDbPluginManager>>,
+}
+
+impl AccountsUpdateNotifierIntf for AccountsUpdateNotifierImpl {
```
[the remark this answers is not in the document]
lijunwangs: Will do

accountsdb-plugin-postgres/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+authors = ["Solana Maintainers <maintainers@solana.foundation>"]
+edition = "2018"
+name = "solana-accountsdb-plugin-postgres"
```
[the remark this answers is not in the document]
lijunwangs: Good idea!

validator/src/main.rs
lijunwangs: ok, will look into it

accountsdb-plugin-manager/src/accountsdb_plugin_service.rs
lijunwangs: Done

accountsdb-plugin-interface/src/accountsdb_plugin_interface.rs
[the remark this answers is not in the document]
lijunwangs: Thanks Trent, will look into that. This interface is binary contract using direct function calls. Such changes need to be communicated and solved via re-compiling the code, and fix potential conflicts. Out of this consideration, I declare the interfaces directly instead of using various existing account structures.

lijunwangs:
> I think it's a good idea to break up the
> 
> 1. Interface
> 2. Changes to AccountsDb
> 
> That way we can check in the interface code without having to incur any performance hits until the performance optimizations are in.

There is no performance penalty if the plugin is not loaded. That being said, I am planning to update this PR with passing the information via reference instead of clone to stave off penalty if an account is not selected when a validator is loaded -- this need to be done as it is part of the interface.

accountsdb-plugin-interface/src/accountsdb_plugin_interface.rs
```diff
@@ -0,0 +1,78 @@
+/// The interface for AccountsDb plugins. A plugin must implement
+/// the AccountsDbPlugin trait to work with the Solana Validator.
+/// In addition, the dynamic libraray must export a "C" function _create_plugin which
+/// creates the implementation of the plugin.
+use {std::any::Any, std::io, thiserror::Error};
+
+#[derive(Clone, PartialEq, Default, Debug)]
+pub struct ReplicaAccountMeta {
+    pub pubkey: String,
```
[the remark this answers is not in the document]
CriesofCarrots: Perhaps use `[u8; 32]` instead, if you're trying to avoid the dependency on sdk/`Pubkey`?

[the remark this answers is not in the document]
lijunwangs: Changed to use reference only

accountsdb-plugin-intf/Cargo.toml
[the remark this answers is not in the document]
lijunwangs: done

accountsdb-plugin-manager/src/accounts_update_notifier.rs
```diff
@@ -0,0 +1,215 @@
+use std::str::FromStr;
```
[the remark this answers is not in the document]
lijunwangs: done

accountsdb-plugin-intf/src/accountsdb_plugin_intf.rs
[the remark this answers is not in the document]
lijunwangs: done

runtime/src/accounts_db.rs
```diff
@@ -937,6 +937,26 @@ struct RemoveUnrootedSlotsSynchronization {
 
 type AccountInfoAccountsIndex = AccountsIndex<AccountInfo>;
 
+pub trait AccountsUpdateNotifierIntf: std::fmt::Debug {
```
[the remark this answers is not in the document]
lijunwangs: Done

runtime/src/accounts_db.rs
lijunwangs: Done

accountsdb-plugin-postgres/src/accountsdb_plugin_postgres.rs
lijunwangs: Passing by reference is done

accountsdb-plugin-interface/src/accountsdb_plugin_interface.rs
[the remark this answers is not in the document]
lijunwangs: Changed to use enum to pass versioned account info

accountsdb-plugin-interface/src/accountsdb_plugin_interface.rs
```diff
@@ -0,0 +1,78 @@
...
+
+#[derive(Error, Debug)]
+pub enum AccountsDbPluginError {
+    #[error("Error opening config file.")]
+    ConfigFileOpenError(#[from] io::Error),
+
+    #[error("Error reading config file.")]
+    ConfigFileReadError { msg: String },
+
+    #[error("Error connecting to the backend data store.")]
```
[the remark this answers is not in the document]
lijunwangs: Done

accountsdb-plugin-manager/src/accounts_update_notifier.rs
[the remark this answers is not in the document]
lijunwangs: Done

validator/src/main.rs
lijunwangs: Done

accountsdb-plugin-manager/src/accounts_update_notifier.rs
```diff
@@ -0,0 +1,131 @@
...
+    pub fn new(plugin_manager: Arc<RwLock<AccountsDbPluginManager>>) -> Self {
+        AccountsUpdateNotifierImpl { plugin_manager }
+    }
+
+    fn accountinfo_from_shared_account_data<'a>(
+        &self,
+        pubkey: &'a Pubkey,
+        account: &'a AccountSharedData,
+    ) -> Option<ReplicaAccountInfo<'a>> {
+        //let data = account.data().to_vec();
```
sakridge, line 60: remove

runtime/src/accounts_db.rs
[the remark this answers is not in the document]
sakridge: The plugin can handle that, no?

sakridge:
couple small nits, but I think it looks good!

accountsdb-plugin-manager/src/accountsdb_plugin_service.rs
```diff
@@ -0,0 +1,117 @@
...
+    ///    The `libpath` field should be pointed to the full path of the dynamic shared library
+    ///    (.so file) to be loaded. The shared library must implement the `AccountsDbPlugin`
+    ///    trait. And the shared library shall export a `C` function `_create_plugin` which
+    ///    shall create the implementation of `AccountsDbPlugin` and returns to the caller.
+    ///    The rest of the JSON fields' definition is up to to the concrete plugin implementation
+    ///    It is usually used to configure the connection information for the external data store.
+
+    pub fn new(
+        confirmed_bank_receiver: Receiver<BankNotification>,
+        accountsdb_plugin_config_file: &Path,
```
CriesofCarrots, line 57: Doesn't this need to handle multiple plugin configs?

accountsdb-plugin-manager/src/accountsdb_plugin_service.rs
```diff
@@ -0,0 +1,117 @@
...
+            accountsdb_plugin_config_file
+        );
+        let plugin_manager = AccountsDbPluginManager::new();
+        let plugin_manager = Arc::new(RwLock::new(plugin_manager));
+
+        let mut file = File::open(accountsdb_plugin_config_file).unwrap();
+        let mut contents = String::new();
+        file.read_to_string(&mut contents).unwrap();
+
+        let result: serde_json::Value = serde_json::from_str(&contents).unwrap();
```
CriesofCarrots, line 70: Let's pass these dependency errors up instead of just panicking.

accountsdb-plugin-manager/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+log = "0.4.11"
+serde = "1.0.130"
+serde_derive = "1.0.103"
+serde_json = "1.0.67"
+solana-logger = { path = "../logger", version = "=1.8.0" }
+solana-metrics = { path = "../metrics", version = "=1.8.0" }
+solana-rpc = { path = "../rpc", version = "=1.8.0" }
+solana-runtime = { path = "../runtime", version = "=1.8.0" }
+solana-sdk = { path = "../sdk", version = "=1.8.0" }
+solana-accountsdb-plugin-interface = { path = "../accountsdb-plugin-interface", version = "=1.8.0" }
```
CriesofCarrots, line 25: Nit: A-Z this

accountsdb-plugin-postgres/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+log = "0.4.14"
+postgres = { version = "0.19.1", features = ["with-chrono-0_4"] }
+
+serde = "1.0.130"
+serde_derive = "1.0.103"
+serde_json = "1.0.67"
+solana-accountsdb-plugin-interface = { path = "../accountsdb-plugin-interface", version = "=1.8.0" }
+solana-logger = { path = "../logger", version = "=1.8.0" }
+
+thiserror = "1.0.29"
```
CriesofCarrots, line 28:
```suggestion
postgres = { version = "0.19.1", features = ["with-chrono-0_4"] }
serde = "1.0.130"
serde_derive = "1.0.103"
serde_json = "1.0.67"
solana-accountsdb-plugin-interface = { path = "../accountsdb-plugin-interface", version = "=1.8.0" }
solana-logger = { path = "../logger", version = "=1.8.0" }
thiserror = "1.0.29"
```

accountsdb-plugin-manager/src/accountsdb_plugin_service.rs
```diff
@@ -0,0 +1,117 @@
...
+    #[error("Invalid plugin path")]
+    InvalidPluginPath,
+
+    #[error("Cannot load plugin shared library")]
+    PluginLoadError(String),
+}
+
+/// The service managing the AccountsDb plugin workflow.
+pub struct AccountsDbPluginService {
+    confirmed_slots_observer: SlotStatusObserver,
```
CriesofCarrots, line 36: This field name doesn't seem right, since processed slots may never be confirmed.
How about just slot_status_observer?

accountsdb-plugin-manager/src/slot_status_observer.rs
```diff
@@ -0,0 +1,80 @@
...
+    pub fn join(&mut self) -> thread::Result<()> {
+        self.exit_updated_slot_server.store(true, Ordering::Relaxed);
+        self.confirmed_bank_receiver_service
+            .take()
+            .map(JoinHandle::join)
+            .unwrap()
+    }
+
+    fn run_confirmed_bank_receiver(
+        confirmed_bank_receiver: Receiver<BankNotification>,
```
CriesofCarrots, line 46:
```suggestion
        bank_notification_receiver: Receiver<BankNotification>,
```

accountsdb-plugin-manager/src/slot_status_observer.rs
```diff
@@ -0,0 +1,80 @@
...
+            atomic::{AtomicBool, Ordering},
+            Arc,
+        },
+        thread::{self, Builder, JoinHandle},
+    },
+};
+
+#[derive(Debug)]
+pub(crate) struct SlotStatusObserver {
+    confirmed_bank_receiver_service: Option<JoinHandle<()>>,
```
CriesofCarrots, line 16:
```suggestion
    bank_notification_receiver_service: Option<JoinHandle<()>>,
```

accountsdb-plugin-postgres/src/accountsdb_plugin_postgres.rs
```diff
@@ -0,0 +1,348 @@
...
+                        return Ok(());
+                    }
+                } else {
+                    return Ok(());
+                }
+
+                debug!(
+                    "Updating account {:?} {:?} at slot {:?} using account selector {:?}",
+                    account.pubkey,
+                    account.owner,
```
CriesofCarrots, line 170: I think we need to stringify these fields somewhere before here, don't we?
We want to print pretty strings, plus your database schema is expecting varchar

core/src/validator.rs
```diff
@@ -545,13 +575,19 @@ impl Validator {
                 ));
             }
 
-            let (confirmed_bank_sender, confirmed_bank_receiver) = unbounded();
-
             let accountsdb_repl_service = config.accountsdb_repl_service_config.as_ref().map(|accountsdb_repl_service_config| {
+                let (confirmed_bank_sender, confirmed_bank_receiver) = unbounded();
+                confirmed_bank_senders.push(confirmed_bank_sender);
```
CriesofCarrots, line 580: I think the "confirmed_bank" prefix is no longer accurate, since the plugin service is using the frozen notification separately.
How about "bank_notification_sender/bank_notification_receiver", etc?

CriesofCarrots: Oh, I see you did make these kinds of nomenclature changes in optimistically_confirmed_bank_tracker.rs.
Let's do that everywhere!

CriesofCarrots (commented):
Should we include an example config file with the accountsdb-plugin-postgres crate?

accountsdb-plugin-postgres/src/accountsdb_plugin_postgres.rs
lijunwangs: The database is not using VARCHAR any longer for efficiency.

runtime/src/accounts_db.rs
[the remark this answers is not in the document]
lijunwangs: Yes. It is being handled in the plugin itself.

validator/src/main.rs
```diff
@@ -1867,6 +1867,22 @@ pub fn main() {
...
+                .requires("accountsdb_plugin_config")
+                .hidden(true)
+                .help("Enable AccountsDb plugin through which account data can streamed to external data store."),
+        )
+        .arg(
+            Arg::with_name("accountsdb_plugin_config")
+                .long("accountsdb-plugin-config")
+                .value_name("FILE")
+                .takes_value(true)
+                .hidden(true)
```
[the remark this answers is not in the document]
lijunwangs: Multiple plugin will be supported in a separate PR.

validator/src/main.rs
```diff
@@ -1867,6 +1867,22 @@ pub fn main() {
...
                 .help("Number of threads to use for servicing AccountsDb Replication requests"),
         )
+        .arg(
+            Arg::with_name("enable_accountsdb_plugin")
+                .long("enable-accountsdb-plugin")
+                .takes_value(false)
+                .requires("accountsdb_plugin_config")
+                .hidden(true)
+                .help("Enable AccountsDb plugin through which account data can streamed to external data store."),
+        )
```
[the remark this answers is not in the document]
lijunwangs: It has a little overlapping because of the "require". But I think this is a relatively significant feature for the validator and deserves a feature flag to control it explicitly to make the people more conscious about the risks/benefits of it.

accountsdb-plugin-manager/src/slot_status_observer.rs
lijunwangs: Done

accountsdb-plugin-postgres/src/accountsdb_plugin_postgres.rs
lijunwangs: I changed for the debug message.

core/src/validator.rs
lijunwangs: Done

accountsdb-plugin-manager/src/slot_status_observer.rs
lijunwangs: Done

accountsdb-plugin-manager/src/accountsdb_plugin_service.rs
lijunwangs: Done

accountsdb-plugin-manager/src/accounts_update_notifier.rs
lijunwangs: Done

accountsdb-plugin-postgres/src/accountsdb_plugin_postgres.rs
CriesofCarrots: I see VARCHAR in the schema sql...? https://github.com/solana-labs/solana/pull/20047/files#diff-27e23ad1496611c1e34cb0a0349fc572ab7dafcd9e9bbe4b8ae5d7c20b0c2e9c

validator/src/main.rs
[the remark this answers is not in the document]
CriesofCarrots: I think this complexity for the user is not worth it

accountsdb-plugin-postgres/src/accountsdb_plugin_postgres.rs
lijunwangs: Good catch -- I missed cherry-picking the schema change. Updated

validator/src/main.rs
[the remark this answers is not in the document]
lijunwangs: okay. removed. Thanks!

CriesofCarrots (approved):
lgtm to land and iterate on
Thanks for all the nit fixing :)

accountsdb-plugin-postgres/src/accountsdb_plugin_postgres.rs
```diff
@@ -0,0 +1,349 @@
...
+    /// Unload all plugins and loaded plugin libraries, making sure to fire
+    /// their `on_plugin_unload()` methods so they can do any necessary cleanup.
+    fn on_unload(&mut self) {
+        info!("Unloading plugin: {:?}", self.name());
+    }
+
+    fn update_account(&mut self, account: ReplicaAccountInfoVersions, slot: u64) -> Result<()> {
+        match account {
+            ReplicaAccountInfoVersions::V0_0_1(account) => {
+                if let Some(accounts_selector) = &self.accounts_selector {
```
fanatid, line 160: Why empty `accounts_selector` from config means select nothing? It looks strange.

PR #20048, opened 2021-09-21
Use a constant for unlocked header uid

brooksprumo:
Instead of just a literal `0` for UID when the header is unlocked, make a constant to give additional context.


jeffwashington (approved):
lgtm

PR #20053, opened 2021-09-21
feat(explorer): load and display transaction details when commitment is confirmed

oJshua:
#### Problem
Currently, transaction details are only displayed when the confirmation status is finalized. We have this information as soon as the confirmation status is confirmed, so it makes sense to display it as soon as possible. 

#### Summary of Changes
- slight refactor of details section
- load details with commitment `"confirmed"`

Fixes #


oJshua:
@jstarry did you have any comments on this one? Thanks!

jstarry:
Hey thanks for the ping, it fell off my radar with all the outage stuff. Changes look good!

oJshua:
@jstarry Not a problem! Thank you!

PR #20067, opened 2021-09-21
AcctIdx: get can load from disk

jeffwashington:
#### Problem
acct idx will soon put most items on disk. Get will need to be able to read from disk or in-mem.
#### Summary of Changes
allow get to read from disk index
Fixes #

### DIFF
--- a/runtime/src/in_mem_accounts_index.rs
+++ b/runtime/src/in_mem_accounts_index.rs
@@ -1,5 +1,6 @@
 use crate::accounts_index::{
-    AccountMapEntry, AccountMapEntryInner, IndexValue, SlotList, WriteAccountMapEntry,
+    AccountMapEntry, AccountMapEntryInner, AccountMapEntryMeta, IndexValue, RefCount, SlotList,
+    WriteAccountMapEntry,
 };
 use crate::bucket_map_holder::{Age, BucketMapHolder};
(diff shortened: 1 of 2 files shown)


runtime/src/in_mem_accounts_index.rs
```diff
@@ -115,7 +130,20 @@ impl<T: IndexValue> InMemAccountsIndex<T> {
...
+
+        // not in cache, look on disk
+        let new_entry = self.load_account_entry_from_disk(pubkey)?;
+        let mut map = self.map().write().unwrap();
+        let entry = map.entry(*pubkey);
+        let result = match entry {
+            Entry::Occupied(occupied) => Arc::clone(occupied.get()),
+            Entry::Vacant(vacant) => Arc::clone(vacant.insert(new_entry)),
+        };
+        Some(result)
```
brooksprumo, line 146: Minor, can the `Arc::clone()` get refactored out of each match arm and put in the return value instead? i.e. `Some(Arc::clone(result))`.

brooksprumo (commented):
Looks good overall

runtime/src/in_mem_accounts_index.rs
jeffwashington: No, that was my first attempt, but the return is a ref in both cases and the lifetimes didn't work.

brooksprumo (approved):

PR #20069, opened 2021-09-21
AcctIdx: remove_if_slot_list_empty uses disk backing

jeffwashington:
#### Problem
This function has to handle a new set of options where the item could exist on disk but not in the cache.
#### Summary of Changes

Fixes #


brooksprumo (approved):
lgtm

PR #20074, opened 2021-09-21
AcctIdx: flush inserts to disk

jeffwashington:
#### Problem
during bg flush, write dirty items to disk.
There is a placeholder here for removing items from in-mem after writing to disk. It has complexity, so I broke it into a separate pr.
#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/in_mem_accounts_index.rs
+++ b/runtime/src/in_mem_accounts_index.rs
@@ -419,6 +419,11 @@ impl<T: IndexValue> InMemAccountsIndex<T> {
         self.storage.wait_dirty_or_aged.notify_one();
     }
 
+    fn should_remove_from_mem(&self, current_age: Age, entry: &AccountMapEntry<T>) -> bool {
+        // this could be tunable dynamically based on memory pressure
+        current_age == entry.age()
+    }
+
     fn flush_internal(&self) {
         let was_dirty = self.bin_dirty.swap(false, Ordering::Acquire);
         let current_age = self.storage.current_age();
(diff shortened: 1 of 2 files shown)


runtime/src/in_mem_accounts_index.rs
```diff
@@ -419,6 +419,11 @@ impl<T: IndexValue> InMemAccountsIndex<T> {
         self.storage.wait_dirty_or_aged.notify_one();
     }
 
+    fn should_remove_from_mem(&self, current_age: Age, entry: &AccountMapEntry<T>) -> bool {
+        // this could be tunable dynamically based on memory pressure
+        current_age == entry.age()
+    }
+
     fn flush_internal(&self) {
         let was_dirty = self.bin_dirty.swap(false, Ordering::Acquire);
```
brooksprumo, line 428: Not related to this PR, but this swap can probably be Relaxed. No need to change it now, just wanted to put it out there for us to be aware of and think about.

runtime/src/in_mem_accounts_index.rs
```diff
@@ -428,21 +433,52 @@ impl<T: IndexValue> InMemAccountsIndex<T> {
...
+            flush_entries_updated_on_disk += 1;
+            disk.insert(&k, (&v.slot_list.read().unwrap(), v.ref_count()));
         }
+        Self::update_time_stat(&self.stats().flush_update_us, m);
+        Self::update_stat(
+            &self.stats().flush_entries_updated_on_disk,
+            flush_entries_updated_on_disk,
+        );
+
+        // loop that processes 'removes' will go here
```
brooksprumo, line 480: I assuming this will be in a different PR, yes?

brooksprumo (approved):
lgtm, assuming that "future code here" section will land in a subsequent PR.

runtime/src/in_mem_accounts_index.rs
jeffwashington: #20076

runtime/src/in_mem_accounts_index.rs
jeffwashington: can you tell me your thinking?

jeffwashington: also in description:
> There is a placeholder here for removing items from in-mem after writing to disk. It has complexity, so I broke it into a separate pr.

runtime/src/in_mem_accounts_index.rs
jeffwashington: also in description:
> There is a placeholder here for removing items from in-mem after writing to disk. It has complexity, so I broke it into a separate pr.

runtime/src/in_mem_accounts_index.rs
brooksprumo: Sorry, I think I'm wrong... I must've gotten mixed up with the `v.dirty()` and `v.set_dirty()` later down in the loop...

PR #20085, opened 2021-09-21
AcctIdx: move background() to bucket holder

jeffwashington:
#### Problem
this function makes more sense in the bucket holder. Most of the state is stored in bucket holder.
#### Summary of Changes

Fixes #


runtime/src/bucket_map_holder_stats.rs
```diff
@@ -59,7 +58,7 @@ impl BucketMapHolderStats {
         now.saturating_sub(last) // could saturate to 0. That is ok.
     }
 
-    fn ms_per_age<T: IndexValue>(&self, storage: &Arc<BucketMapHolder<T>>) -> u64 {
+    fn ms_per_age<T: IndexValue>(&self, storage: &BucketMapHolder<T>) -> u64 {
```
brooksprumo, line 61: 👏

brooksprumo (dismissed):

PR #20087, opened 2021-09-21
orderings on AcctIdx dirty

jeffwashington:
#### Problem
I think these were from a discussion we had.
#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/accounts_index.rs
+++ b/runtime/src/accounts_index.rs
@@ -175,11 +175,11 @@ impl<T: IndexValue> AccountMapEntryInner<T> {
     }
 
     pub fn dirty(&self) -> bool {
-        self.meta.dirty.load(Ordering::Relaxed)
+        self.meta.dirty.load(Ordering::Acquire)
     }
 
-    pub fn set_dirty(&self, value: bool) -> bool {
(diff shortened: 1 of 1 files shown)


runtime/src/accounts_index.rs
```diff
@@ -175,11 +175,11 @@ impl<T: IndexValue> AccountMapEntryInner<T> {
     }
 
     pub fn dirty(&self) -> bool {
-        self.meta.dirty.load(Ordering::Relaxed)
+        self.meta.dirty.load(Ordering::Acquire)
     }
 
     pub fn set_dirty(&self, value: bool) -> bool {
-        self.meta.dirty.swap(value, Ordering::Relaxed)
+        self.meta.dirty.swap(value, Ordering::Acquire)
```
brooksprumo, line 182: I think we want the 'store' part of this swap to be Release.

Also, it looks like we don't use the return value of `set_dirty()` anywhere. Assuming that's not going to change, I'd change the `swap()` to a `store()` too, and remove the return value.

```suggestion
        self.meta.dirty.store(value, Ordering::Release)
```

brooksprumo (approved):
lgtm!

PR #20096, opened 2021-09-21
Handle duplicate AppendVec IDs

### LINKED ISSUE #17088: Incremental Snapshots

### LINKED ISSUE #18641: [Incremental Snapshots] Handle duplicate AppendVec IDs
Is it guaranteed that between the full snapshot generated on one node, an the incremental snapshot generated/fetched from another node, that *none* the storage entries will have overlapping ids?

If this is not the case, should we re-map duplicate ids before reconstructing the storage map?

_Originally posted by @carllin in https://github.com/solana-labs/solana/pull/18504#discussion_r667286793_

brooksprumo:
When reconstructing the AccountsDb, if the storages came from full and
incremental snapshots generated on different nodes, it's possible that
the AppendVec IDs could overlap/have duplicates, which would cause the
reconstruction to fail.

This commit handles this issue by unconditionally remapping the
AppendVec ID for every AppendVec.

Fixes #18641

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -8669,11 +8669,6 @@ pub mod tests {
             accounts.write_version.load(Ordering::Relaxed)
         );
 
-        assert_eq!(
-            daccounts.next_id.load(Ordering::Relaxed),
-            accounts.next_id.load(Ordering::Relaxed)
-        );
-
         // Get the hash for the latest slot, which should be the only hash in the
         // bank_hashes map on the deserialized AccountsDb
         assert_eq!(daccounts.bank_hashes.read().unwrap().len(), 2);
--- a/runtime/src/serde_snapshot.rs
+++ b/runtime/src/serde_snapshot.rs
@@ -433,8 +433,8 @@ where
     }
 
     // Remap the deserialized AppendVec paths to point to correct local paths
-    let num_collisions = AtomicUsize::default();
-    let next_append_vec_id = AtomicUsize::default();
+    let num_collisions = AtomicUsize::new(0);
+    let next_append_vec_id = AtomicUsize::new(0);
     let mut measure_remap = Measure::start("remap");
     let mut storage = (0..snapshot_storages.len())
         .into_par_iter()
--- a/runtime/src/serde_snapshot.rs
+++ b/runtime/src/serde_snapshot.rs
@@ -472,7 +472,8 @@ where
                         break (new_append_vec_id, new_append_vec_path);
                     }
 
-                    // A file exists at the new path.  Try again.
+                    // If we made it this far, a file exists at the new path.  Record the collision
+                    // and try again.
                     num_collisions.fetch_add(1, Ordering::Relaxed);
                 };
                 // Only rename the file if the new ID is actually different from the original.
--- a/runtime/src/serde_snapshot.rs
+++ b/runtime/src/serde_snapshot.rs
@@ -376,13 +376,13 @@ fn reconstruct_single_storage<E>(
     slot: &Slot,
     append_vec_path: &Path,
     storage_entry: &E,
-    new_append_vec_id: Option<AppendVecId>,
+    remapped_append_vec_id: Option<AppendVecId>,
     new_slot_storage: &mut HashMap<AppendVecId, Arc<AccountStorageEntry>>,
 ) -> Result<(), Error>
 where
     E: SerializableStorage,
 {
-    let append_vec_id = new_append_vec_id.unwrap_or_else(|| storage_entry.id());
+    let append_vec_id = remapped_append_vec_id.unwrap_or_else(|| storage_entry.id());
     let (accounts, num_accounts) =
         AppendVec::new_from_file(append_vec_path, storage_entry.current_len())?;
     let u_storage_entry =
@@ -453,11 +453,11 @@ where
 
                 // Remap the AppendVec ID to handle any duplicate IDs that may previously existed
                 // due to full snapshots and incremental snapshots generated from different nodes
-                let (new_append_vec_id, new_append_vec_path) = loop {
-                    let new_append_vec_id = next_append_vec_id.fetch_add(1, Ordering::Relaxed);
-                    let new_file_name = AppendVec::file_name(*slot, new_append_vec_id);
-                    let new_append_vec_path =
-                        append_vec_path.parent().unwrap().join(&new_file_name);
+                let (remapped_append_vec_id, remapped_append_vec_path) = loop {
+                    let remapped_append_vec_id = next_append_vec_id.fetch_add(1, Ordering::Relaxed);
+                    let remapped_file_name = AppendVec::file_name(*slot, remapped_append_vec_id);
+                    let remapped_append_vec_path =
+                        append_vec_path.parent().unwrap().join(&remapped_file_name);
 
                     // Break out of the loop in the following situations:
                     // 1. The new ID is the same as the original ID.  This means we do not need to
@@ -466,10 +466,10 @@ where
                     //    rename the file to this new path.
                     //    **DEVELOPER NOTE:**  Keep this check last so that it can short-circuit if
                     //    possible.
-                    if storage_entry.id() == new_append_vec_id
-                        || std::fs::metadata(&new_append_vec_path).is_err()
+                    if storage_entry.id() == remapped_append_vec_id
+                        || std::fs::metadata(&remapped_append_vec_path).is_err()
                     {
-                        break (new_append_vec_id, new_append_vec_path);
+                        break (remapped_append_vec_id, remapped_append_vec_path);
                     }
 
                     // If we made it this far, a file exists at the new path.  Record the collision
@@ -477,15 +477,15 @@ where
                     num_collisions.fetch_add(1, Ordering::Relaxed);
                 };
                 // Only rename the file if the new ID is actually different from the original.
-                if storage_entry.id() != new_append_vec_id {
-                    std::fs::rename(append_vec_path, &new_append_vec_path)?;
+                if storage_entry.id() != remapped_append_vec_id {
+                    std::fs::rename(append_vec_path, &remapped_append_vec_path)?;
                 }
 
                 reconstruct_single_storage(
                     slot,
-                    &new_append_vec_path,
+                    &remapped_append_vec_path,
                     storage_entry,
-                    Some(new_append_vec_id),
+                    Some(remapped_append_vec_id),
                     &mut new_slot_storage,
                 )?;
             }
--- a/runtime/src/serde_snapshot.rs
+++ b/runtime/src/serde_snapshot.rs
@@ -520,7 +520,7 @@ where
         .fetch_add(snapshot_version, Ordering::Relaxed);
     accounts_db.generate_index(limit_load_slot_count_from_snapshot, verify_index);
 
-    datapoint_debug!(
+    datapoint_info!(
         "reconstruct_accountsdb_from_fields()",
         ("remap-time-us", measure_remap.as_us(), i64),
         (
--- a/runtime/src/serde_snapshot.rs
+++ b/runtime/src/serde_snapshot.rs
@@ -22,6 +22,7 @@ use {
     log::*,
     rayon::prelude::*,
     serde::{de::DeserializeOwned, Deserialize, Serialize},
+    solana_measure::measure::Measure,
     solana_program_runtime::InstructionProcessor,
     solana_sdk::{
         clock::{Epoch, Slot, UnixTimestamp},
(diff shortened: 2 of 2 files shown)


runtime/src/serde_snapshot.rs
```diff
@@ -442,51 +451,84 @@ where
...
+
+    datapoint_debug!(
+        "reconstruct_accountsdb_from_fields()",
+        ("remap-time-us", measure_remap.as_us(), i64),
+        (
+            "remap-collisions",
+            num_collisions.load(Ordering::Relaxed),
+            i64
+        ),
+    );
```
brooksprumo, line 531: I figured these datapoints would be helpful too. Knowing how long the remapping takes may later indicate if/when this needs to be sped up. Also, I think the number of collisions will be an interesting metric.

runtime/src/accounts_db.rs
```diff
@@ -8669,11 +8669,6 @@ pub mod tests {
             accounts.write_version.load(Ordering::Relaxed)
         );
 
-        assert_eq!(
-            daccounts.next_id.load(Ordering::Relaxed),
-            accounts.next_id.load(Ordering::Relaxed)
-        );
```
brooksprumo, line 8675: Lastly, this test was checking the `next_id` before and after serde to ensure the values were the same. But since that logic is now different, this invariant is no longer valid. Luckily `next_id` is not used anywhere (outside of getting the next AppendVec ID here in AccountsDb), so this invariant is not required.

brooksprumo (commented):
Here's a few spots of the code I've called out, hopefully for better/easier understanding and context.

runtime/src/serde_snapshot.rs
```diff
@@ -442,51 +451,84 @@ where
...
+                    // try again (and update the collision count), otherwise break out of the loop
+                    // with our new ID and path.
+                    let exists = std::fs::metadata(&new_append_vec_path).is_ok();
+                    if exists {
+                        num_collisions.fetch_add(1, Ordering::Relaxed);
+                    } else {
+                        break (new_append_vec_id, new_append_vec_path);
+                    }
+                };
+                std::fs::rename(append_vec_path, &new_append_vec_path)?;
```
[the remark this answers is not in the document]
brooksprumo: Tagging @carllin @sakridge @ryoqun if you have suggestions.

runtime/src/serde_snapshot.rs
carllin: `datapoint_debug!` won't show up by default in most reporting, should use at least `datapoint_info!`

brooksprumo: Oh, that's good to know. What/where is the "reporting" that you're referring to? I'd like to be able to see these datapoints on the grafana/metrics website (not the local console logs); that one needs `_info!` or higher?

runtime/src/serde_snapshot.rs
```diff
@@ -442,51 +451,84 @@ where
...
+                let (new_append_vec_id, new_append_vec_path) = loop {
+                    let new_append_vec_id = next_append_vec_id.fetch_add(1, Ordering::Relaxed);
+                    let new_file_name = AppendVec::file_name(*slot, new_append_vec_id);
+                    let new_append_vec_path =
+                        append_vec_path.parent().unwrap().join(&new_file_name);
+
+                    // Make sure there's not another AppendVec already at this path.  If there is,
+                    // try again (and update the collision count), otherwise break out of the loop
+                    // with our new ID and path.
+                    let exists = std::fs::metadata(&new_append_vec_path).is_ok();
```
carllin, line 465: Probably faster to just keep around a HashSet for all the appendvecs used in this slot, updated at the end of the loop here https://github.com/solana-labs/solana/blob/2e821371a588be0c9282b80b243d5a79138115df/runtime/src/serde_snapshot.rs#L480

carllin: Does this still allow the same `append_vec_id` across different slots?

runtime/src/serde_snapshot.rs
```diff
@@ -372,17 +376,19 @@ fn reconstruct_single_storage<E>(
     slot: &Slot,
     append_vec_path: &Path,
     storage_entry: &E,
+    new_append_vec_id: Option<AppendVecId>,
```
carllin, line 379: If `SerializableStorage` supported a `set_append_vec_id()` function, maybe we could just set the `new_append_vec_id` before this function is called and remove this additional argument?

runtime/src/serde_snapshot.rs
carllin: yup, info! or higher

runtime/src/serde_snapshot.rs
brooksprumo: This is awesome! I really wanted this to work, but unfortunately I don't think it will be worth it...

Since the `snapshot_storages` (and therefore `slot_storage`) is inside a parallel loop, if I want to mutate `SerializableStorage` that would require `E` (and therefore `SerializableStorage`) to be `Send` everywhere. I went down this rabbit hole for a while, adding `Send` to everything, until I got to the serde tests and the errors started to get more involved.

Check out the last line for what I needed to do, in this error here:
```
error[E0277]: `<C as serde_snapshot::TypeContext<'_>>::SerializableAccountStorageEntry` cannot be sent between threads safely
   --> runtime/src/serde_snapshot/tests.rs:76:5
    |
76  |     reconstruct_accountsdb_from_fields(
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `<C as serde_snapshot::TypeContext<'_>>::SerializableAccountStorageEntry` cannot be sent between threads safely
    |
    = help: the trait `Send` is not implemented for `<C as serde_snapshot::TypeContext<'_>>::SerializableAccountStorageEntry`
note: required by a bound in `serde_snapshot::reconstruct_accountsdb_from_fields`
   --> runtime/src/serde_snapshot.rs:407:50
    |
394 | fn reconstruct_accountsdb_from_fields<E>(
    |    ---------------------------------- required by a bound in this
...
407 |     E: SerializableStorage + std::marker::Sync + std::marker::Send,
    |                                                  ^^^^^^^^^^^^^^^^^ required by this bound in `serde_snapshot::reconstruct_accountsdb_from_fields`
help: consider further restricting the associated type
    |
68  |     R: Read, <C as serde_snapshot::TypeContext<'_>>::SerializableAccountStorageEntry: Send
    |            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

I could also solve this by:
1. Putting the mutation in an unsafe block, or
2. Cloning the `slot_storage` each iteration

but both those options seem bad. It feels like the current impl is the easiest. What do you think?

runtime/src/serde_snapshot.rs
brooksprumo: AppendVec IDs will always be unique since the `next_append_vec_id` is Atomic. So no, the same AppendVec ID will _not_ occur on different slots.

As far as the HashSet, I like that idea. It becomes complicated for a few reasons, mostly due to the parallelization. I think it may be easier to talk about that on a call to see if there's a way around it. Is that OK?

carllin: Makes sense! I misread the code, I thought we were only remapping ids for files that didn't exist. Oops!

carllin (approved):

PR #20099, opened 2021-09-22
windows: Make solana-test-validator work

joncinque:
#### Problem

`solana-test-validator` doesn't work on Windows.

#### Summary of Changes

Let's get it working! Here are the important changes:

* ledger lock needs to be done on a file instead of the directory -- let me know if you prefer to have this gated on `cfg[windows]`
* IPC service needs to use the Windows pipe naming scheme (I couldn't believe this one, who comes up with this stuff?)
* always disable the JIT
* file logging not possible yet because we can't redirect stderr, but this will change once env_logger fixes the pipe output target!

I tested this by deploying a program, doing some transfers, making sure the lock works, and running the `program-test` tests.  Let me know if there are any other bits worth testing.

Since logging to file still isn't possible, it's a huge spew of logs when it's running, but at least it works!

### DIFF
--- a/validator/src/bin/solana-test-validator.rs
+++ b/validator/src/bin/solana-test-validator.rs
@@ -24,8 +24,8 @@ use {
     },
     solana_streamer::socket::SocketAddrSpace,
     solana_validator::{
-        admin_rpc_service, dashboard::Dashboard, println_name_value, redirect_stderr_to_file,
-        test_validator::*,
+        admin_rpc_service, dashboard::Dashboard, ledger_lockfile, lock_ledger, println_name_value,
+        redirect_stderr_to_file, test_validator::*,
     },
     std::{
         collections::HashSet,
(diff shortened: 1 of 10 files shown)


validator/src/bin/solana-test-validator.rs
```diff
@@ -304,7 +304,14 @@ fn main() {
...
 
-    let mut ledger_fd_lock = fd_lock::RwLock::new(fs::File::open(&ledger_path).unwrap());
+    let lockfile = ledger_path.join("ledger.lock");
+    let mut ledger_fd_lock = fd_lock::RwLock::new(
+        fs::OpenOptions::new()
+            .write(true)
+            .create(true)
+            .open(&lockfile)
+            .unwrap(),
+    );
```
mvines, line 314: This duplication between `solana-test-validator` and `solana-validator` is now a little too big for my preference.  How about a shared function in validator/src/lib.rs for this

validator/src/bin/solana-test-validator.rs
```diff
@@ -396,6 +403,9 @@ fn main() {
         IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)),
         faucet_port,
     ));
+    #[cfg(target_family = "windows")]
+    let bpf_jit = false;
```
mvines, line 407: Can you add a comment in the code here that explains why the JIT is forced off for windows

mvines (approved):
heh, very cool!  

Want to ship the windows validator as well?  https://github.com/solana-labs/solana/blob/1b58166e9d1eca794e02839fa0230067cf4e6935/scripts/cargo-install-all.sh#L72

A 1.7 backport of this seems fine too

joncinque:
> Want to ship the windows validator as well? 

Definitely, will do!

PR #20113, opened 2021-09-22
AcctIdx: when stop_flush is disabled, prepare to age

jeffwashington:
#### Problem
stop_flush is set by a client call when a range is being held in memory. When the last stop_flush is cleared, a bucket that was previously prevented from flushing may now be ready to flush, so make sure a bg thread is awake to visit the bucket.
#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/in_mem_accounts_index.rs
+++ b/runtime/src/in_mem_accounts_index.rs
@@ -483,9 +483,10 @@ impl<T: IndexValue> InMemAccountsIndex<T> {
 
     fn start_stop_flush(&self, stop: bool) {
         if stop {
-            self.stop_flush.fetch_add(1, Ordering::Acquire);
-        } else {
-            self.stop_flush.fetch_sub(1, Ordering::Release);
+            self.stop_flush.fetch_add(1, Ordering::Release);
+        } else if 1 == self.stop_flush.fetch_sub(1, Ordering::Release) {
+            // stop_flush went to 0, so this bucket could now be ready to be aged
+            self.storage.wait_dirty_or_aged.notify_one();
         }
     }
 


runtime/src/in_mem_accounts_index.rs
```diff
@@ -481,8 +481,9 @@ impl<T: IndexValue> InMemAccountsIndex<T> {
     fn start_stop_flush(&self, stop: bool) {
         if stop {
             self.stop_flush.fetch_add(1, Ordering::Acquire);
```
brooksprumo, line 483: I know this isn't part of this PR, but I'm guessing this should probably be `Release` like the `else` case below. For one, we're not using the returned value from `fetch_add()` that we just read, which was a red flag. And if setting `stop_flush` to non-zero gates other logic, _writing_ this value is the important part for proper gating.

runtime/src/in_mem_accounts_index.rs
```diff
@@ -481,8 +481,9 @@ impl<T: IndexValue> InMemAccountsIndex<T> {
...
-        } else {
-            self.stop_flush.fetch_sub(1, Ordering::Release);
+        } else if 1 == self.stop_flush.fetch_sub(1, Ordering::Release) {
```
brooksprumo, line 484: This ordering is another spot that I'll need to keep in mind going forward. It's interesting because the case of `stop_flush == 0` is like an `unlock()`, which means Release is right. It looks like we're also using this as a "go" for the notify() call below, which usually means Acquire is right.

But it's also inside a conditional, so there's a control dependency... And the notify is a condition variable, so it's got its own synch stuff in it... Anyway, I think Release is good for now, just listing out my thoughts, but I don't have a conclusive "this is the definite right way". We can always bump this to AcqRel later.

brooksprumo (commented):
More ordering comments/questions 🙃

runtime/src/in_mem_accounts_index.rs
jeffwashington: done

brooksprumo (approved):

PR #20115, opened 2021-09-22
AcctIdx: Fix bad test

jeffwashington:
#### Problem
test construction was passing # bins for bin#. bins are 0..#bins. This results in indexing at [bins] which will fail.
#### Summary of Changes

Fixes #


brooksprumo (approved):
Good catch

PR #20121, opened 2021-09-22
AcctIdx: startup causes us to flush remove everything asap

jeffwashington:
#### Problem
during startup, we want to remove items from in-mem asap. The assumption is almost no pubkeys are in the index twice at startup.
#### Summary of Changes

Fixes #


brooksprumo (approved):
lgtm

PR #20122, opened 2021-09-22
AcctIdx: generation per thread starts at a random bucket

jeffwashington:
#### Problem
This helps a lot with startup time index generation by reducing lock contention as whole buckets are written to disk and resizing. Activity per bin is now more evenly distributed.
#### Summary of Changes

Fixes #


brooksprumo (dismissed):
Took me a good bit to figure out what was going on, but I think I get it and it looks good to me.

PR #20124, opened 2021-09-22
AcctIdx: metrics for loading from disk

jeffwashington:
#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/bucket_map_holder_stats.rs
+++ b/runtime/src/bucket_map_holder_stats.rs
@@ -14,6 +14,10 @@ pub struct BucketMapHolderStats {
     pub entries_from_mem: AtomicU64,
     pub entry_missing_us: AtomicU64,
     pub entries_missing: AtomicU64,
+    pub load_disk_found_count: AtomicU64,
+    pub load_disk_found_us: AtomicU64,
+    pub load_disk_missing_count: AtomicU64,
+    pub load_disk_missing_us: AtomicU64,
     pub updates_in_mem: AtomicU64,
     pub items: AtomicU64,
     pub keys: AtomicU64,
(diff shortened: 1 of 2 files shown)


runtime/src/bucket_map_holder_stats.rs
```diff
@@ -14,6 +14,10 @@ pub struct BucketMapHolderStats {
     pub entries_from_mem: AtomicU64,
     pub entry_missing_us: AtomicU64,
     pub entries_missing: AtomicU64,
+    pub load_disk: AtomicU64,
+    pub load_disk_us: AtomicU64,
+    pub load_disk_missing: AtomicU64,
+    pub load_disk_missing_us: AtomicU64,
```
brooksprumo, line 20: It look me a bit to understand the differences here. What about slightly different names?

- `load_disk_found_us` for the measurement when the item is found
- `load_disk_found_count` for the count of items found
- `load_disk_missing_us` for the measurement when the item is not found
- `load_disk_missing_count` for the count of items not found

runtime/src/in_mem_accounts_index.rs
```diff
@@ -106,10 +106,21 @@ impl<T: IndexValue> InMemAccountsIndex<T> {
...
+            let m = Measure::start("load_disk");
+            let entry_disk = disk.read_value(pubkey);
+            match &entry_disk {
+                Some(_) => {
+                    Self::update_time_stat(&self.stats().load_disk_us, m);
+                    Self::update_stat(&self.stats().load_disk, 1);
+                }
+                None => {
+                    Self::update_time_stat(&self.stats().load_disk_missing_us, m);
+                    Self::update_stat(&self.stats().load_disk_missing_us, 1);
```
brooksprumo, line 119: This one should be the count, not the timer, right?

brooksprumo (dismissed):

runtime/src/in_mem_accounts_index.rs
jeffwashington: yep, good catch.

runtime/src/bucket_map_holder_stats.rs
jeffwashington: done

PR #20128, opened 2021-09-23
Add Vote instruction to update on chain vote state

AshwinSekar:
#### Problem
To avoid out of order voting issues, we allow validators to update the on chain vote state directly through this new instruction. 

#### Summary of Changes
Plumbing for #20014 in the vote program

Fixes #

### DIFF
--- a/programs/vote/src/vote_state/mod.rs
+++ b/programs/vote/src/vote_state/mod.rs
@@ -375,6 +375,139 @@ impl VoteState {
         Ok(())
     }
 
+    // TODO: `Ensure check_slots_are_valid()` runs on the slots in `new_state`
+    // before this
+    pub fn process_new_vote_state(
+        &mut self,
+        new_state: VecDeque<Lockout>,
+        new_root: Option<Slot>,
+        timestamp: Option<i64>,
+        epoch: Epoch,
+    ) -> Result<(), VoteError> {
+        assert!(!new_state.is_empty());
+
+        // check_slots_are_valid()` ensures we don't process any states
+        // that are older than the current state
+        if !self.votes.is_empty() {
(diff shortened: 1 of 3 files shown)


programs/vote/src/vote_state/mod.rs
```diff
@@ -757,6 +997,33 @@ pub fn process_vote<S: std::hash::BuildHasher>(
...
+
+    if versioned.is_uninitialized() {
+        return Err(InstructionError::UninitializedAccount);
+    }
+
+    let mut vote_state = versioned.convert_to_current();
+    let authorized_voter = vote_state.get_and_update_authorized_voter(clock.epoch)?;
+    verify_authorized_signer(&authorized_voter, signers)?;
+
+    vote_state.check_slots_are_valid(&vote_state_update.get_legacy_vote(), slot_hashes)?;
```
carllin, line 1017: Nice, you read the todo :)

programs/vote/src/vote_state/mod.rs
```diff
@@ -96,6 +96,48 @@ impl Lockout {
...
+    pub fn last_voted_slot_hash(&self) -> Option<(Slot, Hash)> {
+        self.lockouts
+            .back()
+            .copied()
+            .map(|lockout| (lockout.slot, self.hash))
+    }
+
+    pub fn get_legacy_vote(&self) -> Vote {
+        Vote {
+            slots: self.lockouts.iter().map(|&lockout| lockout.slot).collect(),
```
carllin, line 134: If we implemented a common trait `VoteTransaction` (or a better name) for `VoteStateUpdate` and `Vote`, and passed that trait to `check_slots_are_valid()`, we could avoid this clone.

AshwinSekar: yeah I agree, seems wasteful right now, i'll do that

PR #20129, opened 2021-09-23
Add metric for duplicate retransmit

carllin:
#### Problem
Duplicate retransmits are not being logged in the retransmit metrics

#### Summary of Changes

Fixes #


behzadnouri:
@carllin why this was done on v1.7 as opposed to master and then backport to v1.7?

carllin:
@behzadnouri this issue didn't seem to be a problem on master because the `should_skip_retransmit()` covers this case in the `num_shreds_skipped` metric https://github.com/solana-labs/solana/blob/082d5dc5b267f8b36d6669bb5900a0f9e44a42e0/core/src/retransmit_stage.rs#L221-L224

PR #20140, opened 2021-09-23
AcctIdx: range takes a ref

jeffwashington:
#### Summary of Changes

Fixes #


brooksprumo (changes requested):
This looks odd to me. Why is a second reference needed?

brooksprumo (approved):
We talked offline, lgtm

PR #20143, opened 2021-09-23
Follow up to blockstore test cleanup (#19656)

steviez:
#### Summary of Changes
Use auto-deleting ledger so tests cleanup on disk after themselves - see #19656 for more details. The commit that added the test I'm updating came around same time as #19656, so they must have just missed each other.


brooksprumo (approved):
Thanks for the cleanup! lgtm

PR #20148, opened 2021-09-23
AcctIdx: condense upsert to always use read lock

jeffwashington:
#### Problem
Locking has changed on accts idx. Previously, upsert first attempted with the read lock, which would work if the pubkey was already in the in-mem hashmap. If it wasn't in the in-mem hashmap, upsert would get a write lock and call the upsert. Now that locking has changed, we don't need a write lock to call upsert. So, simplify the code path.
#### Summary of Changes

Fixes #


brooksprumo (approved):
Nice! lgtm

PR #20150, opened 2021-09-23
bucket_map: rename num_cells() to capacity()

brooksprumo:
I'm reading through bucket_map to make sure I understand it all. While doing that, I've been updating/renaming/documenting bits and pieces that has helped me.

**_No behavior changes_**

### DIFF
--- a/bucket_map/src/bucket_storage.rs
+++ b/bucket_map/src/bucket_storage.rs
@@ -123,7 +123,7 @@ impl BucketStorage {
     }
 
     pub fn uid(&self, ix: u64) -> Uid {
-        if ix >= self.num_cells() {
+        if ix >= self.capacity() {
             panic!("bad index size");
         }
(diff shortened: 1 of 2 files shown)


bucket_map/src/bucket_storage.rs
```diff
@@ -332,7 +332,9 @@ impl BucketStorage {
         self.stats.resizes.fetch_add(1, Ordering::Relaxed);
         self.stats.resize_us.fetch_add(m.as_us(), Ordering::Relaxed);
     }
-    pub fn num_cells(&self) -> u64 {
+
+    /// Return the number of cells currently allocated
+    pub fn capacity(&self) -> u64 {
```
brooksprumo, line 337: I think I was the one that changed this from capacity() to num_cells() initially, so it's only fair that I fix it because I was wrong 😅

jeffwashington (approved):
lgtm

PR #20168, opened 2021-09-24
AcctIdx: bg loop sleeps when all buckets flushed at current age

jeffwashington:
#### Problem
this allows the bg threads to remain sleep an optimal amount of time
#### Summary of Changes

Fixes #


brooksprumo (approved):
Nice. lgtm

PR #20171, opened 2021-09-24
Make rpc_bootstrap() aware of local incremental snapshots

### LINKED ISSUE #20169: Make `rpc_bootstrap()` aware of local incremental snapshots

brooksprumo:
#### Problem

When a validator boots up and runs `rpc_boostrap()`, it only looks at its local _full_ snapshots when checking if it needs to download a new snapshot or not.  This could result in unnecessarily downloading a snapshot.

#### Summary of Changes

Make `rpc_bootstrap()` aware of local incremental snapshots.

_NOTE:_ The diff is easier to read when ignoring whitespace.

Fixes #20169

### DIFF
--- a/validator/src/main.rs
+++ b/validator/src/main.rs
@@ -503,12 +503,7 @@ fn get_rpc_node(
         }
         blacklist_timeout = Instant::now();
 
-        let mut highest_snapshot_hash: Option<(Slot, Hash)> =
-            snapshot_utils::get_highest_full_snapshot_archive_info(snapshot_archives_dir).map(
-                |snapshot_archive_info| {
-                    (snapshot_archive_info.slot(), *snapshot_archive_info.hash())
-                },
-            );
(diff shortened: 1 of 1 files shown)


validator/src/main.rs
```diff
@@ -3072,3 +3070,29 @@ fn process_account_indexes(matches: &ArgMatches) -> AccountSecondaryIndexes {
         indexes: account_indexes,
     }
 }
+
+/// Get the Slot and Hash of the local snapshot with the highest slot.  Can be either a full
+/// snapshot or an incremental snapshot.
+fn get_highest_local_snapshot_hash(
```
brooksprumo, line 3076: Getting info about the highest local snapshot was used in two places, so I refactored 'em into this function. One needs the Slot and Hash, whereas the other only needs the Slot. I don't think the extra Hash will make a performance impact here.

PR #20176, opened 2021-09-24
AcctIdx: throttling bg flushing to spread out work over interval

jeffwashington:
#### Problem
Accounts index on disk works on a bg lru mechanism (aging) that has an age interval of 400ms (like slot time). There can be many threads required to allow the bg processes to keep up. However, we don't want a spike of active and busy threads every 400ms. We'd like the work to be spread evenly across threads and over the 400ms interval to keep system load more steady.
#### Summary of Changes
Put threads to sleep when bg processing is ahead of required progress.
Fixes #

### DIFF
--- a/runtime/src/bucket_map_holder.rs
+++ b/runtime/src/bucket_map_holder.rs
@@ -172,21 +172,65 @@ impl<T: IndexValue> BucketMapHolder<T> {
         AGE_MS
     }
 
+    /// return an amount of ms to sleep
+    fn throttling_wait_ms_internal(
+        &self,
+        interval_ms: u64,
+        elapsed_ms: u64,
+        bins_flushed: u64,
+    ) -> Option<u64> {
+        let target_percent = 90; // aim to finish in 90% of the allocated time
+        let remaining_ms = (interval_ms * target_percent / 100).saturating_sub(elapsed_ms);
+        let remaining_bins = (self.bins as u64).saturating_sub(bins_flushed);
+        if remaining_bins == 0 || remaining_ms == 0 || elapsed_ms == 0 || bins_flushed == 0 {
+            // any of these conditions result in 'do not wait due to progress'
+            return None;
+        }
+        let ms_per_s = 1_000;
+        let rate_bins_per_s = bins_flushed * ms_per_s / elapsed_ms;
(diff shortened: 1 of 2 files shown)


runtime/src/bucket_map_holder.rs
```diff
@@ -159,21 +159,65 @@ impl<T: IndexValue> BucketMapHolder<T> {
...
+        if remaining_bins == 0 || remaining_ms == 0 || elapsed_ms == 0 || bins_flushed == 0 {
+            // any of these conditions result in 'do not wait due to progress'
+            return None;
+        }
+        let ms_per_s = 1_000;
+        let rate_bins_per_s = bins_flushed * ms_per_s / elapsed_ms;
+        let expected_bins_processed_in_remaining_time = rate_bins_per_s * remaining_ms / ms_per_s;
+        if expected_bins_processed_in_remaining_time > remaining_bins {
+            // wait because we predict will finish prior to target
+            Some(1)
```
brooksprumo, line 181: Why pick a literal `1`here? I don't understand how this number was chosen.

brooksprumo (commented):
Looks good; had a question about the amount to wait.

runtime/src/bucket_map_holder.rs
jeffwashington: it is easy to think in terms of ms. 1ms is the smallest amount of time to wait. There are 400 ms divisions in 400ms. There are currently something like 10-16 threads and 8k buckets. So, avg 800 buckets per thread if we were evenly divided. That's 2 per ms per thread to hit our target of all buckets processed evenly during the 400ms. So, the rough idea was when we get ahead, wait this thread 1 ms, and then wake up, do another bucket, and check the system progress again. It seems like a simple strategy that won't get out of control if the calculations are wrong somehow.
So, this implementation is just wait 1 ms if we're ahead and don't wait if we're not ahead. This should be sufficient for now. We can make it more sophisticated, but a spikey, 16 pegged threads every 400ms is not good. This can transition to durations, different algorithms for variable waits, etc. if we need it to.

brooksprumo: Makes sense, thanks!

brooksprumo (dismissed):

jeffwashington:
Here's the load on lemond (128 cores) against mnb.
image

brooksprumo (approved):
lgtm

PR #20185, opened 2021-09-24
Rpc: filters performance improvement

### LINKED ISSUE #20177: Rpc: possible performance improvement for filters?
#### Problem

https://github.com/solana-labs/solana/blob/ca83167cfcc4f5cdb0837bf04d99041e7377c516/client/src/rpc_filter.rs#L49-L53

`MemcmpEncodedBytes` enum looks not healthy, why `Binary` variant has `String` as inner? And more, this `String` have bytes encoded in base58 :anguished: 

I'm not sure about the performance impact but decode base58 string each time for filter account on requests like `get_program_accounts` looks not good.

#### Proposed Solution

1. Rename `Binary` to `Base58` and add `Binary(Vec<u8>)`.
2. Left `Binary(String)` and add new variant with `Vec<u8>` (name?).

fanatid:
Closes #20177


fanatid:
CI should be fixed after https://github.com/solana-labs/solana/pull/20182

jstarry:
There's a temp fix on master already that you can rebase off of too

fanatid:
@ChrisBellew can I help somehow for review? Should I add tests or split PR for adding new enum variants and deprecation?

CriesofCarrots:
Thanks for all the work here!

PR #20186, opened 2021-09-24
AcctIdx: upsert uses get from mem first

jeffwashington:
#### Problem
getting write locks is expensive. Upsert can usually satisfy a request by just getting a read lock and finding items in memory.
#### Summary of Changes
upsert looks in memory first using only a read lock.
Fixes #


brooksprumo (approved):
lgtm

PR #20198, opened 2021-09-25
Fix blatant md formatting at the very top

tredondo:


jstarry (approved):
Thanks for fixing!

tredondo:
Happy to fix, but what's the CD cycle? I still see [week-old fixes not live](https://github.com/solana-labs/solana/commit/073c5359b0e39a4e27d085c4d222925f58c40c09#commitcomment-57045863).

jstarry:
Ah docs.solana.com is published from the v1.6 feature branch... I'll backport fixes now, hold tight

mvines:
> Ah docs.solana.com is published from the v1.6 feature branch... I'll backport fixes now, hold tight

It should be published from v1.6 actually, if v1.6 caused a publish then we may have a bug in CI

jstarry:
> > Ah docs.solana.com is published from the v1.6 feature branch... I'll backport fixes now, hold tight
> 
> It should be published from v1.6 actually, if v1.6 caused a publish then we may have a bug in CI

You said v1.6 twice 🤔

mvines:
😅 "It should be published from v1.7..."

PR #20200, opened 2021-09-25
Terminology: fix punctuation in CBC block

tredondo:
This terminology docs is not that friendly to noobs


jstarry:
Could you actually remove "CBC block" from the terminology list, please? We don't need it

tredondo:
Removed via https://github.com/solana-labs/solana/pull/20215

PR #20217, opened 2021-09-25
fix "unkown" Borsh typo

drbh:
#### Problem
Error typo

#### Summary of Changes
changes `BORSH_IO_ERROR` from `unkown` to `unknown` error


mvines:
Thanks for this PR!  Looks like the CI failure is due to this:
https://github.com/solana-labs/solana-program-library/blob/87b74cfa4c399fae428bce1f08e95d97fa625b8a/stake-pool/program/tests/vsa_remove.rs#L323
@joncinque - WDYT, should we just accept the speling mistake here or try to fix it with an update to that SPL test?

joncinque:
I can update the test to not check for the string.  Once https://github.com/solana-labs/solana-program-library/pull/2453 is merged, this can be retried no problem!

On the other hand, if some nodes are generating `"Unknown"` and others are generating `"Unkown"`, will that cause a blockhash mismatch?

jstarry:
> I can update the test to not check for the string. Once [solana-labs/solana-program-library#2453](https://github.com/solana-labs/solana-program-library/pull/2453) is merged, this can be retried no problem!
> 
> On the other hand, if some nodes are generating `"Unknown"` and others are generating `"Unkown"`, will that cause a blockhash mismatch?

I think you mean bank hash since blockhash is just the latest poh entry. The bank hash includes the blockhash, account delta hash, previous bank hash, etc. (check out `Bank::hash_internal_state`) but doesn't include tx metadata like errors. So updating the error description won't impact the bank hash and should be safe.

joncinque:
> I think you mean bank hash since blockhash is just the latest poh entry. The bank hash includes the blockhash, account delta hash, previous bank hash, etc. (check out Bank::hash_internal_state) but doesn't include tx metadata like errors. So updating the error description won't impact the bank hash and should be safe.

That's exactly what I meant, thanks for the clarification 😄

mvines:
Thanks @drbh

drbh:
Thank you @mvines! 🙏

PR #20222, opened 2021-09-26
Add get_snapshot_info() RPC function

### LINKED ISSUE #20173: [Incremental Snapshots] Add RPC fn to query a specific snapshot's info
At bootstrap, a node gets a list of snapshot Slot+Hashes, and then picks the highest Slot. If it needs to download the snapshot, we need to know if it's a full snapshot or an incremental snapshot. If it's an incremental snapshot, we need to download the full snapshot (if we don't have it already).

So we need a way to query the RPC node about the snapshot to figure out (1) its type, and (2) if it's an incremental snapshot then we also need the slot+hash of its corresponding full snapshot.

brooksprumo:
#### Problem

During bootstrap, the node gets a list of snapshot hashes to download a snapshot. The node assumes all these hashes correspond to full snapshots, which is not the case anymore. So (1) the node needs to query if the hash is a full or incremental snapshot in order to construct the download file path correctly, and (2), if the hash is an incremental snapshot, the node must then also download the full snapshot afterwards (or rather, it'll check if it has it locally already).

The changes to the download logic in bootstrap will happen in a subsequent PR (see issue #20225). This PR addresses the problem of not being able to query a snapshot hash for its type and associated information.

#### Summary of Changes

Add an RPC fn for querying a snapshot's information.

Fixes #20173

### DIFF
--- a/docs/src/developing/clients/jsonrpc-api.md
+++ b/docs/src/developing/clients/jsonrpc-api.md
@@ -53,6 +53,7 @@ gives a convenient interface for the RPC methods.
 - [getSlot](jsonrpc-api.md#getslot)
 - [getSlotLeader](jsonrpc-api.md#getslotleader)
 - [getSlotLeaders](jsonrpc-api.md#getslotleaders)
+- [getSnapshotInfo](jsonrpc-api.md#getsnapshotinfo)
 - [getStakeActivation](jsonrpc-api.md#getstakeactivation)
 - [getSupply](jsonrpc-api.md#getsupply)
 - [getTokenAccountBalance](jsonrpc-api.md#gettokenaccountbalance)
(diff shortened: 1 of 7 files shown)


docs/src/developing/clients/jsonrpc-api.md
```diff
@@ -2444,6 +2445,96 @@ The first leader returned is the leader for slot #100:
...
 ```
 
+### getSnapshotInfo
+
+**NEW: This method is only available in solana-core v1.8 or newer.**
+
+Returns the snapshot information for the given snapshot slot and hash.
+
+If the node has a snapshot with the given slot and hash, it will return the
+associated information about it.  For full snapshots, this is the slot and has.
```
mvines, line 2455:
```suggestion
associated information about it.  For full snapshots, this is the slot and hash.
```

brooksprumo: Hah, good eye!

docs/src/developing/clients/jsonrpc-api.md
```diff
@@ -2444,6 +2445,96 @@ The first leader returned is the leader for slot #100:
 }
 ```
 
+### getSnapshotInfo
```
mvines, line 2448: Will the caller know whether they are requesting information about a full or incremental snapshot?  If so I suspect that adding two RPC methods instead of one will be simpler from an implementation and interface perspective.

brooksprumo: Yeah, that's kinda the crux of this PR.

During bootstrap, the node (caller) gets a `(Slot, Hash)` for a snapshot to download, but it doesn't know if that corresponds to a full snapshot or an incremental snapshot. The caller needs to know the snapshot type in order to (1) call `download_snapshot_archive()` with the right params, and (2), if the type is an incremental snapshot, the caller may also need to download the associated full snapshot too.

So this RPC fn is returning info to answer both questions (1) and (2) for the caller. 

Maybe there's another way? This was just one way that I thought of doing it, but happy to change to something better!

mvines: > During bootstrap, the node (caller) gets a (Slot, Hash) for a snapshot to download, but it doesn't know if that corresponds to a full snapshot or an incremental snapshot

I see. It gets this `(Slot, Hash)` from gossip right?

mvines: `CrdsData:: SnapshotHashes` specifically?  The reason I ask is because I think we probably need a new CRDS data type to share incremental snapshot hashes over gossip for backwards compatibility, either by complimenting or deprecating `CrdsData:: SnapshotHashes`.  Once we do that the client will then know from a `(Slot, Hash)` whether the snapshot is full or incremental and won't need to ask RPC at all.

brooksprumo: > > During bootstrap, the node (caller) gets a (Slot, Hash) for a snapshot to download, but it doesn't know if that corresponds to a full snapshot or an incremental snapshot
> 
> I see. It gets this `(Slot, Hash)` from gossip right?

Yep!

brooksprumo:
Moving this back to a draft. I think a different implementation to solve the underlying issue will be required.

CriesofCarrots:
Feel free to re-ping me if this comes back around and needs eyes :)

brooksprumo:
Closing this PR. A new way to handle downloading incremental snapshots at bootstrap will be used and a different RPC function will be needed. Refer to this new PR, #20356.

PR #20223, opened 2021-09-26
Rpc: remove not required clone

fanatid:
#### Problem

For extract value from `serde_json::Value` `clone` call is not required, `take` method can be used.

#### Summary of Changes

`clone` replaced by `take`. In addition, I added `?` operator instead of `match`.

```rust
return Ok(json["result"].clone());
```
changed to
```rust
return Ok(json["result"].take());
```


mvines:
Nice cleanup, thanks

PR #20230, opened 2021-09-26
Update installer.md CLI output

tredondo:


jstarry:
I don't believe that those implemented proposals are meant to be updated, they are more for a historical record of protocol / tooling changes

PR #20234, opened 2021-09-26
Add stats to snapshot_bank()

brooksprumo:
I want to see snapshot metrics for (1) the number of snapshot storages, and (2) how long it takes to get those storages, for both incremental and full snapshots.

I've added `datapoint_info!` to `snapshot_bank()`.

### DIFF
--- a/runtime/src/snapshot_utils.rs
+++ b/runtime/src/snapshot_utils.rs
@@ -1594,6 +1594,7 @@ pub fn snapshot_bank(
     hash_for_testing: Option<Hash>,
     snapshot_type: Option<SnapshotType>,
 ) -> Result<()> {
+    let mut measure_snapshot_storages = Measure::start("snapshot-storages");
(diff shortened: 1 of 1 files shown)


runtime/src/snapshot_utils.rs
```diff
@@ -1594,6 +1594,7 @@ pub fn snapshot_bank(
     hash_for_testing: Option<Hash>,
     snapshot_type: Option<SnapshotType>,
 ) -> Result<()> {
+    let mut measure_snapshot_storages = Measure::start("snapshot-storages");
```
jeffwashington, line 1597: I think I'd rather see a separate function:
maybe
get_snapshot_storages(root_bank, snapshot_type);
This is lots of lines to ultimately call `root_bank.get_snapshot_storages`

brooksprumo: Great idea. Done!

jeffwashington (approved):
lgtm

PR #20253, opened 2021-09-27
AcctIdx: fix math error with stats calc

jeffwashington:
#### Problem
this is causing intermittent failures on stable ci (overflow) and was off by 1 anyway.
#### Summary of Changes

Fixes #


brooksprumo (approved):
Ah, interesting! lgtm

PR #20255, opened 2021-09-27
AcctIdx: handle race condition with flush and delete

jeffwashington:
#### Problem
I discovered a race condition with deleting items and the bg flushing. Ugh.
#### Summary of Changes
Hold the read lock while writing to disk to prevent deletes from occurring during writes to disk.
Fixes #


brooksprumo (approved):
Bummer about the race requiring the longer lock, but glad you found it!

PR #20257, opened 2021-09-27
AcctIdx: keep cached or multiple slot lists in cache

jeffwashington:
#### Problem
Going to disk to get an item in the acct idx can be expensive.
This may be tuned in the future based on memory budget.
#### Summary of Changes
Keep items that are likely to be used again soon in in-mem cache.
These include items with slot list.len != 1 and items with cached storages in their slot list.
Fixes #


brooksprumo (approved):
lgtm

PR #19877, opened 2021-09-14
Remove demote_program_write_locks feature 

CriesofCarrots:
#### Problem
Feature gating of this functionality won't be needed on v1.6

#### Summary of Changes
Remove it

PR #20263, opened 2021-09-27
v1.6: Restore ability for programs to upgrade themselves

CriesofCarrots:
#### Problem
In preventing inappropriate program write locks (https://github.com/solana-labs/solana/pull/19637 + https://github.com/solana-labs/solana/pull/19877 ), we assumed that programs would not need to upgrade themselves (ie. the program-id of a transaction is never writable). However, various governance programs depend on this ability, using their own governance mechanism for program upgrades.

#### Summary of Changes
Restore program-id write lock when the upgradeable loader id is present in the transaction; feature gate

### DIFF
--- a/programs/bpf/tests/programs.rs
+++ b/programs/bpf/tests/programs.rs
@@ -1801,8 +1801,6 @@ fn test_program_bpf_upgrade_and_invoke_in_same_tx() {
         message.clone(),
         bank.last_blockhash(),
     );
-    // program_id is automatically demoted to readonly, preventing the upgrade, which requires
-    // writeability
     let (result, _) = process_transaction_and_record_inner(&bank, tx);
     assert_eq!(
         result.unwrap_err(),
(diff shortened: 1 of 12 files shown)


sdk/program/src/message.rs
```diff
@@ -383,18 +383,22 @@ impl Message {
                 let key = self.account_keys[i];
                 sysvar::is_sysvar_id(&key) || BUILTIN_PROGRAMS_KEYS.contains(&key)
             }
-            && !self.is_key_called_as_program(i)
+            && (!self.is_key_called_as_program(i)
+                || (restore_write_lock_when_upgradeable && self.is_upgradeable_loader_present()))
```
CriesofCarrots, line 387: This is the actual logic change

programs/bpf/tests/programs.rs
```diff
@@ -2087,6 +2087,97 @@ fn test_program_bpf_upgrade_via_cpi() {
     );
 }
 
+#[cfg(feature = "bpf_rust")]
+#[test]
+fn test_program_bpf_upgrade_self_via_cpi() {
```
jackcmay, line 2092: Great!

jackcmay (approved):

PR #20264, opened 2021-09-27
AcctIdx: add age_interval_ms() to remove constant use

jeffwashington:
#### Problem
This may become dynamic later. Consolidate uses of the value.
#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/bucket_map_holder.rs
+++ b/runtime/src/bucket_map_holder.rs
@@ -12,7 +12,7 @@ use std::sync::{Arc, Mutex};
 use std::time::Duration;
 pub type Age = u8;
 
-pub const AGE_MS: u64 = SLOT_MS; // match one age per slot time
+const AGE_MS: u64 = SLOT_MS; // match one age per slot time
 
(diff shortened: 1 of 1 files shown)


runtime/src/bucket_map_holder.rs
```diff
@@ -153,6 +153,12 @@ impl<T: IndexValue> BucketMapHolder<T> {
         result
     }
 
+    /// prepare for this to be dynamic if necessary
+    /// For example, maybe startup has a shorter age interval.
+    fn age_interval_ms(&self) -> u64 {
+        AGE_MS
+    }
```
brooksprumo, line 160: I know this would be a pain to do now, but these time-like functions and constants could be `Duration` instead of a `u64`. And then only at the use in `AtomicInterval` convert back to a `u64` (or wherever that conversion is needed). Then the type system comes to our rescue and we hopefully don't have ms/us/ns mixups.

brooksprumo (dismissed):
lgtm. Added a comment for the future too.

PR #20265, opened 2021-09-27
Restore ability for programs to upgrade themselves

CriesofCarrots:
#### Problem
#20263 for master

In preventing inappropriate program write locks (#19637 + #19877 ), we assumed that programs would not need to upgrade themselves (ie. the program-id of a transaction is never writable). However, various governance programs depend on this ability, using their own governance mechanism for program upgrades.

#### Summary of Changes
Restore program-id write lock when the upgradeable loader id is present in the transaction; feature gate and require both `restore_write_lock_when_upgradeable` and  `demote_program_write_locks` features for new functionality.
v1.6 vs v1.7/master compatibility will be restored when https://github.com/solana-labs/solana/pull/20263 is released and these two features are activated on MNB

### DIFF
--- a/runtime/src/accounts.rs
+++ b/runtime/src/accounts.rs
@@ -241,7 +241,6 @@ impl Accounts {
             let rent_for_sysvars = feature_set.is_active(&feature_set::rent_for_sysvars::id());
             let demote_program_write_locks =
                 feature_set.is_active(&feature_set::demote_program_write_locks::id());
-            let is_upgradeable_loader_present = is_upgradeable_loader_present(message);
 
             for (i, key) in message.account_keys_iter().enumerate() {
                 let account = if !message.is_non_loader_key(i) {
@@ -280,7 +279,7 @@ impl Accounts {
                         if bpf_loader_upgradeable::check_id(account.owner()) {
                             if demote_program_write_locks
                                 && message.is_writable(i, demote_program_write_locks)
-                                && !is_upgradeable_loader_present
+                                && !message.is_upgradeable_loader_present()
                             {
                                 error_counters.invalid_writable_account += 1;
                                 return Err(TransactionError::InvalidWritableAccount);
@@ -1131,12 +1130,6 @@ pub fn prepare_if_nonce_account(
         }
     }
     false
-}
-
-fn is_upgradeable_loader_present(message: &SanitizedMessage) -> bool {
-    message
-        .account_keys_iter()
-        .any(|&key| key == bpf_loader_upgradeable::id())
 }
 
 pub fn create_test_accounts(
--- a/runtime/src/accounts.rs
+++ b/runtime/src/accounts.rs
@@ -199,9 +199,9 @@ impl Accounts {
     fn construct_instructions_account(
         message: &SanitizedMessage,
         is_owned_by_sysvar: bool,
-        demote_program_write_locks: bool,
+        demote_program_write_lock_features: bool,
     ) -> AccountSharedData {
-        let mut data = message.serialize_instructions(demote_program_write_locks);
+        let mut data = message.serialize_instructions(demote_program_write_lock_features);
         // add room for current instruction index.
         data.resize(data.len() + 2, 0);
         let owner = if is_owned_by_sysvar {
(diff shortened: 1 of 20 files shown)


sdk/program/src/message/legacy.rs
```diff
@@ -363,7 +363,8 @@ impl Message {
                 let key = self.account_keys[i];
                 sysvar::is_sysvar_id(&key) || BUILTIN_PROGRAMS_KEYS.contains(&key)
             }
-            && !(demote_program_write_locks && self.is_key_called_as_program(i))
+            && !(demote_program_write_lock_features
+                && (self.is_key_called_as_program(i) && !self.is_upgradeable_loader_present()))
```
CriesofCarrots, line 367: Logic change 1 of 2 here

sdk/program/src/message/mapped.rs
```diff
@@ -96,11 +97,12 @@ impl MappedMessage {
...
 
     /// Returns true if the account at the specified index was loaded as writable
-    pub fn is_writable(&self, key_index: usize, demote_program_write_locks: bool) -> bool {
+    pub fn is_writable(&self, key_index: usize, demote_program_write_lock_features: bool) -> bool {
         if self.is_writable_index(key_index) {
             if let Some(key) = self.get_account_key(key_index) {
                 return !(sysvar::is_sysvar_id(key) || BUILTIN_PROGRAMS_KEYS.contains(key)
-                    || (demote_program_write_locks && self.is_key_called_as_program(key_index)));
+                    || !(demote_program_write_lock_features
+                        && (self.is_key_called_as_program(key_index) && !self.is_upgradeable_loader_present())));
```
CriesofCarrots, line 105: Logic change 2 of 2 here

sdk/program/src/message/legacy.rs
CriesofCarrots: Sorry if these links suck. I did some rework to make the new cases easier to follow.

sdk/program/src/message/sanitized.rs
```diff
@@ -308,6 +308,14 @@ impl SanitizedMessage {
                 .saturating_add(num_secp256k1_signatures),
         )
     }
+
+    /// Inspect all message keys for the bpf upgradeable loader
+    pub fn is_upgradeable_loader_present(&self) -> bool {
+        match self {
+            Self::Legacy(message) => message.is_upgradeable_loader_present(),
```
jackcmay, line 315: Side note, but @jstarry this might be clearer if the messages implemented a trait containing `is_upgradeable_loader_present`?

jackcmay (dismissed):

sdk/program/src/message/sanitized.rs
jstarry: Sure, sounds good to me

runtime/src/accounts.rs
```diff
@@ -239,9 +239,9 @@ impl Accounts {
             let mut account_deps = Vec::with_capacity(message.account_keys_len());
             let mut rent_debits = RentDebits::default();
             let rent_for_sysvars = feature_set.is_active(&feature_set::rent_for_sysvars::id());
-            let demote_program_write_locks =
-                feature_set.is_active(&feature_set::demote_program_write_locks::id());
-            let is_upgradeable_loader_present = is_upgradeable_loader_present(message);
+            let demote_program_write_lock_features = feature_set
+                .is_active(&feature_set::demote_program_write_locks::id())
+                && feature_set.is_active(&feature_set::restore_write_lock_when_upgradeable::id());
```
jstarry, line 244: won't this cause forking if some nodes have updated to have this check and others don't?

CriesofCarrots: It could, if I screwed up and activated `demote_program_write_locks` before this lands and is adopted by testnet/devnet. However, my intent is to activate that feature together with `restore_write_lock_when_upgradeable`.
Please let me know if I've missed something, though.

jstarry: Ah got it, that makes sense. It feels to me that we don't need the additional feature switch then. Could we instead replace the old feature switch with this new one? Or just update the behavior of the old switch before enabling it? This approach is doable for sure though

CriesofCarrots: Oh yeah, duh, actually we could use only one feature. I did it this way to keep the features in line with what is in v1.6, but there's actually no reason to preserve 2 features there either. I'll clean up v1.6, and then update this to match. Thanks!

PR #20267, opened 2021-09-27
Fix typo

tredondo:


CriesofCarrots:
I think this one is correct as-is, actually: https://www.merriam-webster.com/dictionary/weight
See entry 2 (transitive verb), definitions 2 and 4

PR #20269, opened 2021-09-27
terminology.md: remove CBC block and unneeded filename

tredondo:
No need to repeat the filename for Markdown anchors. Removed "CBC block" upon a previous request.


jstarry (approved):
Thank you so much!

PR #20275, opened 2021-09-27
Add metric measuring number of successfully inserted push messages

carllin:
#### Problem
No easy way to tell ratio of new messages coming in from push vs pull to measure how effective push in the network currently is

#### Summary of Changes
Add a `process_push_success` metric that is a parallel to the `process_pull_response_success` metric

Fixes #


gossip/src/cluster_info.rs
```diff
@@ -2181,7 +2181,12 @@ impl ClusterInfo {
             messages
                 .into_iter()
                 .flat_map(|(from, crds_values)| {
-                    self.gossip.process_push_message(&from, crds_values, now)
+                    let (num_success, origins) =
+                        self.gossip.process_push_message(&from, crds_values, now);
+                    self.stats
+                        .process_pull_response_success
```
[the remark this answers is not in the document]
carllin: ugh, bad copy paste, thanks!

gossip/src/cluster_info_metrics.rs
```diff
@@ -109,6 +109,7 @@ pub(crate) struct GossipStats {
     pub(crate) new_push_requests2: Counter,
     pub(crate) new_push_requests: Counter,
     pub(crate) new_push_requests_num: Counter,
+    pub(crate) process_push_success: Counter,
```
behzadnouri, line 112: can we plz keep this list sorted?

gossip/src/crds_gossip.rs
```diff
@@ -49,12 +49,15 @@ impl CrdsGossip {
...
-        self.push
-            .process_push_message(&self.crds, from, values, now)
-            .into_iter()
-            .filter_map(Result::ok)
-            .collect()
+    ) -> (usize, HashSet<Pubkey>) {
+        let results = self
+            .push
+            .process_push_message(&self.crds, from, values, now);
+        let success_count = results.len();
```
behzadnouri, line 56: the inserted ones are the ones which are not error.
`results.len()` is the same as `values.len()`.

carllin: Ick, need to read more carefully. Fixed up, thanks!

gossip/src/cluster_info_metrics.rs
carllin: Updated

behzadnouri (approved):

PR #20276, opened 2021-09-27
cargo-build-bpf: Add Windows support

joncinque:
#### Problem

`cargo-build-bpf` doesn't support windows

#### Summary of Changes

Add support for it!

A few notes:
* `ld` doesn't ship by default with mingw, causing `linker not found` errors, so one of the changes is to force the usage of `ld.lld` from the sdk to link.  I have very little knowledge of different linker flavors, but this was the only one that worked.  If you think it should be another one, please let me know.
* it isn't possible to run bash scripts from the program, since they aren't natively executable by windows.  To get around that, this avoids calling `strip.sh` and just calls `llvm-objcopy` directly.
* for `dump.sh`, it's possible to run the script separately, and re-run `cargo build-bpf --dump` to do the postprocessing step.  it isn't elegant, but it actually works!

Fixes #

### DIFF
--- a/sdk/bpf/scripts/install.sh
+++ b/sdk/bpf/scripts/install.sh
@@ -3,11 +3,21 @@
 mkdir -p "$(dirname "$0")"/../dependencies
 cd "$(dirname "$0")"/../dependencies
 
-if [[ "$(uname)" = Darwin ]]; then
-  machine=osx
-else
-  machine=linux
-fi
+unameOut="$(uname -s)"
+case "${unameOut}" in
+  Linux*)
+    criterion_suffix=
+    machine=linux;;
+  Darwin*)
+    criterion_suffix=
+    machine=osx;;
+  MINGW*)
+    criterion_suffix=-mingw
+    machine=windows;;
+  *)
+    criterion_suffix=
+    machine=linux
+esac
 
 download() {
   declare url="$1/$2/$3"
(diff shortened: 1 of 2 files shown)


dmakarov (dismissed):

dmakarov:
This looks good to me, but I'm not quite sure why the usual CI checks were not executed for this PR. Let's wait until bpf-tools release assets are published ( https://github.com/solana-labs/bpf-tools/actions/runs/1281082328 ) before merging this.

joncinque:
Sounds good to me!  I've nudged a little update to the error message to restart the CI job

dmakarov (approved):

dmakarov:
Jon, I had to turn off the build on windows when I released bpf-tools v1.19.  The build failed artificats were copied in preparation for making a package tarball. The reason for the failure was the CI agent ran out of disk space.  I'm not sure how to request a bigger disk for a CI agent machine, and so for now CI doesn't build bpf-tools for Windows.

joncinque:
Ah that's too bad.  Thanks for noticing this.  It looks like we'll need to move the steps to a different directory for this to work as expected, if https://github.com/actions/virtual-environments/issues/1341 is actually possible, we can pass in a new `out` directory to checkout and build everything.  I'll take a longer look at this later.

dmakarov (dismissed):

gajojr:
So what is the solution for this?

TradeUpCards:
> So what is the solution for this?

I'd also like to know. Just getting into Solana coding and can't even do Hello World example because it says 

no such subcommand: `+bpf`

Been scouring the web, but seems most people getting it to work are on Linux and not Windows.

joncinque:
Be sure to install the solana command line tools and rust, then use a  bash-compatible shell like Git Bash.  If you need more support, please ask in the Solana Stack Exchange https://solana.stackexchange.com/

PR #20290, opened 2021-09-28
Simplify return_data accessors in InvokeContext

Lichtso:
#### Problem
Internally the handling of `return_data` in `InvokeContext` is more complex than it needs to be and it hinders integration in ABI v2.

#### Summary of Changes
* Automatically retrieves `program_id` from `invocation_stack`
* Removes the `Option::None` and instead use an empty data `Vec` for that
* Syscall interfaces remain the same

Fixes #


seanyoung (approved):
Looks good and works fine against Solang.

It does make the interface simpler which is nice.

PR #20297, opened 2021-09-28
ProgramTest now rejects transactions that are too large

mvines:
Make ProgramTest match RPC and reject transactions that are too large for a more realistic unit test environment

### DIFF
--- a/program-test/tests/toolarge.rs
+++ b/program-test/tests/toolarge.rs
@@ -1,0 +1,31 @@
+use {
+    solana_program_test::ProgramTest,
+    solana_sdk::{
+        instruction::Instruction, packet::PACKET_DATA_SIZE, pubkey::Pubkey, signature::Signer,
+        transaction::Transaction,
+    },
+};
+
+#[tokio::test]
+async fn test_transaction_too_large() {
+    let (mut banks_client, payer, recent_blockhash) = ProgramTest::default().start().await;
+
+    let mut transaction = Transaction::new_with_payer(
+        &[Instruction {
+            program_id: Pubkey::default(),
+            accounts: vec![],
+            data: [42; PACKET_DATA_SIZE * 2].to_vec(),
+        }],
+        Some(&payer.pubkey()),
+    );
+    transaction.sign(&[&payer], recent_blockhash);
+
+    assert_eq!(
+        "transport transaction error: Transaction too large",
+        &banks_client
+            .process_transaction(transaction)
+            .await
+            .unwrap_err()
+            .to_string(),
+    );
+}
--- a/sdk/src/transaction/mod.rs
+++ b/sdk/src/transaction/mod.rs
@@ -122,6 +122,10 @@ pub enum TransactionError {
     /// Transaction loads a writable account that cannot be written
     #[error("Transaction loads a writable account that cannot be written")]
     InvalidWritableAccount,
+
+    /// Transaction is too large
+    #[error("Transaction too large")]
+    TooLarge,
 }
 
 pub type Result<T> = result::Result<T, TransactionError>;
--- a/banks-server/src/banks_server.rs
+++ b/banks-server/src/banks_server.rs
@@ -1,7 +1,7 @@
 use solana_sdk::feature_set::FeatureSet;
 
 use {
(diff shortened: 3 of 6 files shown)


program-test/tests/toolarge.rs
```diff
@@ -0,0 +1,31 @@
...
+        &[Instruction {
+            program_id: Pubkey::default(),
+            accounts: vec![],
+            data: [42; PACKET_DATA_SIZE * 2].to_vec(),
+        }],
+        Some(&payer.pubkey()),
+    );
+    transaction.sign(&[&payer], recent_blockhash);
+
+    assert_eq!(
```
mvines, line 23: I'm not proud of this assert.  Going from a `TransportError` back into a typed `TransactionError` seems hard, and I didn't feel like reworking the signature of `process_transaction` just for this

sdk/src/transaction/mod.rs
```diff
@@ -122,6 +122,10 @@ pub enum TransactionError {
     /// Transaction loads a writable account that cannot be written
     #[error("Transaction loads a writable account that cannot be written")]
     InvalidWritableAccount,
+
+    /// Transaction is too large
+    #[error("Transaction too large")]
+    TooLarge,
```
CriesofCarrots, line 128: I was thinking there might be other places we'd want to use this for consistency -- like rpc simulateTransaction or RpcClient -- but I guess the return types are already too different, and there's no comparable `verify_transaction()`

program-test/tests/toolarge.rs
CriesofCarrots: Yeah, lack of PartialEq on TransportError is a pita; I've been running into that trying to update `tarpc` as well. I can't think of a better simple assert here.

CriesofCarrots (approved):

CriesofCarrots:
lgtm once CI is happy!

mvines:
Oh fun!  This PR caught some bugs in the SPL governance program or its tests, where it's creating transactions that are too large to send over the wire.

@SebastianBor - 👋🏼, can you please take a look at the failures here.  Is there an easy path to fixing them, or should we just ignore those tests in the SPL repo for now until their transaction sizes can be reduced

SebastianBor:
> Oh fun! This PR caught some bugs in the SPL governance program or its tests, where it's creating transactions that are too large to send over the wire.
> 
> @SebastianBor - 👋🏼, can you please take a look at the failures here. Is there an easy path to fixing them, or should we just ignore those tests in the SPL repo for now until their transaction sizes can be reduced

Looking at the tests which failed my guess is the problem is with deploying upgradable programs. The tests deploy a program  and then test it can be taken under governance by the governance program and upgraded through a proposal. [with_governed_program](https://github.com/solana-labs/solana-program-library/blob/b3684bbfe98e4fd42bf8ac7f82cc0c909c9877c0/governance/program/tests/program_test/mod.rs#L859)
I think It would be prohibitively to slow for tests to deploy the program using smaller transactions, so maybe we could explicitly opt out from that restriction for `ProgramTest` ? Or maybe there is a better way to deploy upgradable program to `ProgramTest`?

mvines:
> I think It would be prohibitively too slow for tests to deploy the program using smaller transactions

I don't see why that'd be the case?  But maybe we could instead use [ProgramTest::add_account](https://github.com/solana-labs/solana/blob/b507715d449acd5f04b1563957e4d52392abe700/program-test/src/lib.rs#L546) to populate those accounts beforehand?  

I'd like for the actual transaction simulation to be a close to the real thing as possible.

SebastianBor:
> > I think It would be prohibitively too slow for tests to deploy the program using smaller transactions
> 
> I don't see why that'd be the case? But maybe we could instead use [ProgramTest::add_account](https://github.com/solana-labs/solana/blob/b507715d449acd5f04b1563957e4d52392abe700/program-test/src/lib.rs#L546) to populate those accounts beforehand?
> 
> I'd like for the actual transaction simulation to be a close to the real thing as possible.

Can we use ProgramTest::add_account to add an upgradable program? I guess if we could just preload the program data buffer then it would be enough because this is what takes time (`bpf_loader_upgradeable::write`) when split into multiple transactions

PR #20302, opened 2021-09-28
AcctIdx: fix logic on should_remove_from_mem

jeffwashington:
#### Problem
logic was wrong resulting in us holding items in memory during startup
#### Summary of Changes

Fixes #


brooksprumo (approved):
lgtm, sorry for not catching this before

PR #20303, opened 2021-09-29
feat(docker): add `explorer` docker support

yasuoyuhao:
#### Problem
Add `explorer` Docker support

#### Summary of Changes

Add `explorer/Dockerfile`
Add `explorer/nginx.conf`

Fixes #


oJshua (changes requested):
Thank you for this contribution! I tested it out and it worked well. One comment. Could you also add newlines at the end of each file?

explorer/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+FROM node:16-alpine as react-build
+WORKDIR /app
+COPY ./package.json ./package-lock.json ./
+RUN npm install
+COPY . ./
+RUN npm run build
+
+# server environment
+FROM nginx:alpine
+COPY ./nginx.conf /etc/nginx/conf.d/configfile.template
```
oJshua, line 11: Would it be possible to do this without bringing in the nginx.conf template? For example, in this tutorial  https://medium.com/@shakyShane/lets-talk-about-docker-artifacts-27454560384f

jstarry:
Why do we need this?

explorer/Dockerfile
yasuoyuhao: It can be done, but this will lose the user's control over monitoring PORT. For the practice of Docker containerization, it may not be so flexible.

yasuoyuhao:
> 我們為什麼需要這個？

Maybe we can provide a more convenient option for the community to deploy on its own? Such as containerization?

yasuoyuhao:
Test Process

```bash
cd explorer
docker build -t explorer .
docker run -d -p 8080:80 explorer
```

and open broswer, enter `localhost:8080`

oJshua:
> Why do we need this?

We probably don't. I'm going to close this considering dockerization of React / create-react-app is well documented around the web. Thanks for your effort here @yasuoyuhao.

PR #20301, opened 2021-09-28
Revert "Refactor: Merge MessageProcessor into InvokeContext (#20165)"

jackcmay:
This reverts commit df6905c3a67e46c17d66a282d200741508c9e5f0.

#### Summary of Changes

Fixes #

PR #20308, opened 2021-09-29
Refactor: Merge message processor into invoke context

Lichtso:
#### Problem
#20165 (which was reverted in #20301) had the following problem:

The feature flag `tx_wide_compute_cap` was left in `ThisInvokeContext::new()`. So it was only applied per transaction, not applied to each instruction, as the constructor got hoisted out of the loop. It diverged on MB and not the CI tests, as the feature flag is activated by default in the tests, but not yet active on MB.

#### Summary of Changes
The application of the feature flag was moved from `ThisInvokeContext::new()` to `InvokeContext::set_instruction_index()`.
See: https://github.com/solana-labs/solana/commit/c579bd83a88ebc734d71b935bb5fcf9aa98490bf

Fixes #


jeffwashington:
I confirmed this makes new roots against mnb and rebased against master. @Lichtso asked me to merge this.

PR #20313, opened 2021-09-29
add unchecked invokes

### LINKED ISSUE #20311: Remove costly RefCell checks
The flat cost for CPI overhead is relatively expensive, sitting at about `9678` compute units when measuring a call from `solana_program::invoke_signed` to the entrypoint of the program  being invoked.

Approximately `5815` of these units are spent on this block of code here https://github.com/solana-labs/solana/blob/c02ef395ede79af57ff8eeacf7817ad2ab2d2e0f/sdk/program/src/program.rs#L23-L37, which may not be required since programs can be responsible for ensuring their code doesn't trigger a RefCell induced panic. 

Removing this block would be beneficial for programs that do a lot of CPI.

Tag @jstarry.

jackcmay:
#### Problem

`invoke` and `invoke_signed` check that the caller is not mishandling any `RefCell`s in `Accountinfo`.  Doing so can be quite expensive in terms of compute units.

#### Summary of Changes

Provide an unchecked version of those functions for folks who want to avoid those checks and use at their own risk.

Fixes #20311

### DIFF
--- a/sdk/program/src/program.rs
+++ b/sdk/program/src/program.rs
@@ -2,18 +2,36 @@ use crate::{
     account_info::AccountInfo, entrypoint::ProgramResult, instruction::Instruction, pubkey::Pubkey,
 };
 
-/// Invoke a cross-program instruction
+/// Invoke a cross-program instruction.
 ///
-/// Note that the program id of the instruction being issued must also be included in
-/// `account_infos`.
+/// Notes:
+/// - RefCell checking can be compute unit expensive, to avoid that expense use
+///   `invoke_unchecked` instead, but at your own risk.
+/// - The program id of the instruction being issued must also be included in
+///   `account_infos`.
 pub fn invoke(instruction: &Instruction, account_infos: &[AccountInfo]) -> ProgramResult {
     invoke_signed(instruction, account_infos, &[])
 }
 
+/// Invoke a cross-program instruction but don't enforce RefCell handling.
+///
+/// Notes:
+/// - The missing checks ensured that the invocation doesn't violate the borrow
+///   rules of the `AccountInfo` fields that are wrapped in `RefCell`s.  To
+///   include the checks call `invoke` instead.
+/// - The program id of the instruction being issued must also be included in
+///   `account_infos`.
+pub fn invoke_unchecked(instruction: &Instruction, account_infos: &[AccountInfo]) -> ProgramResult {
+    invoke_signed_unchecked(instruction, account_infos, &[])
+}
+
 /// Invoke a cross-program instruction with program signatures
 ///
-/// Note that the program id of the instruction being issued must also be included in
-/// `account_infos`.
+/// Notes:
+/// - RefCell checking can be compute unit expensive, to avoid that expense use
+///   `invoke_signed_unchecked` instead, but at your own risk.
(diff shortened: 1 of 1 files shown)


sdk/program/src/program.rs
```diff
@@ -35,6 +49,19 @@ pub fn invoke_signed(
         }
     }
 
+    invoke_signed_unchecked(instruction, account_infos, signers_seeds)
+}
+
+/// Invoke a cross-program instruction with program signatures but don't check
+/// Refcell handling.
```
armaniferrante, line 56:
```suggestion
/// RefCell handling.
```

sdk/program/src/program.rs
```diff
@@ -2,18 +2,32 @@ use crate::{
...
 ///
-/// Note that the program id of the instruction being issued must also be included in
-/// `account_infos`.
+/// Note that the program id of the instruction being issued must also be
+/// included in `account_infos`.
 pub fn invoke(instruction: &Instruction, account_infos: &[AccountInfo]) -> ProgramResult {
     invoke_signed(instruction, account_infos, &[])
 }
 
+/// Invoke a cross-program instruction but don't enforce Refcell handling.
```
armaniferrante, line 13:
```suggestion
/// Invoke a cross-program instruction but don't enforce RefCell handling.
```

sdk/program/src/program.rs
```diff
@@ -2,18 +2,32 @@ use crate::{
...
+///
+/// Notes:
+/// - The program id of the instruction being issued must also be included in
+///   `account_infos`.
+/// - Refcell checking can be compute unit expensive, to avoid that expense use
```
armaniferrante, line 18:
```suggestion
/// - RefCell checking can be compute unit expensive, to avoid that expense use
```

sdk/program/src/program.rs
```diff
@@ -2,18 +2,32 @@ use crate::{
...
+}
+
 /// Invoke a cross-program instruction with program signatures
 ///
-/// Note that the program id of the instruction being issued must also be included in
-/// `account_infos`.
+/// Notes:
+/// - The program id of the instruction being issued must also be included in
+///   `account_infos`.
+/// - Refcell checking can be compute unit expensive, to avoid that expense use
```
armaniferrante, line 29:
```suggestion
/// - RefCell checking can be compute unit expensive, to avoid that expense use
```

armaniferrante (approved):
Thank you!

sdk/program/src/program.rs
```diff
@@ -35,6 +49,19 @@ pub fn invoke_signed(
         }
     }
 
+    invoke_signed_unchecked(instruction, account_infos, signers_seeds)
+}
+
+/// Invoke a cross-program instruction with program signatures but don't check
+/// RefCell handling.
```
jstarry, line 56: Can you also add something like:
> These checks ensure that the invocation doesn't violate the borrow rules of the `AccountInfo` fields that are wrapped in `RefCell`'s

jstarry: Just so that people know why it's valuable to have those checks and they take on risk if they skip them

jstarry (approved):

Issue #19627, opened 2021-09-03
Build InfluxDB query to help set transaction costs


taozhu-chicago:
#### Problem
`cost_model` must calculate to the same `cost` for a given transaction from any validator machine, for stability of the cluster. Therefore the cost of each components in [proposed Fee structure](https://github.com/solana-labs/solana/pull/16984/files) must be assigned with a static (eg hardcoded) number, similar to [Compute Budget](https://github.com/solana-labs/solana/blob/master/sdk/src/compute_budget.rs#L82-L99). 

For consistency, transaction cost should be calculated closely following proposed fee schedule, therefore the need of a consistent method to assign/collaborate `compute units` to each cost elements, such as `signature`, `write lock` etc.

#### Method:

The raw data are metrics measuring each operation's CPU cost, mostly in `us`. When collecting metrics, we need to get cluster-aggregated data to ride of individual validator performance differences. It takes two steps to define `compute units` for each Transaction cost elements: 

1. The first step is get cluster-wide `compute unit` to `us` conversion rate. 
 
This rate essentially describes the relationship between hardcoded BPF instruction's operation budget (in `compute unit`) vs `us`, which should be the same relationship for signature verification, or built-in program. In another words, it can convert other transaction cost element from `us` to `compute unit`. 

PR #19624 added few more data points, if `tx_wide_compute_cap` feature is turned on, to add BPF instruction's `compute unit` and `us` into InfluxDB, 

Run query to get `conversion_rate`: 
```
select mean(rate) as conversion_rate from (SELECT "accumulated_units"::field / "execute_us"::field as rate FROM "tds"."autogen"."per_program_timings" WHERE (time > :dashboardTime: and time < :upperDashboardTime:) AND accumulated_units::field != 0 AND pubkey::field = 'all' Fill(0))
```
When I ran validator (PR #19624 with `tx_wide_compute_cap` on) on GCE VM for 15 mins, I got `conversion_rate = 15.03`:
![image](https://user-images.githubusercontent.com/82401714/132068764-1864a659-1ab7-4d0c-9b31-f1c698cfdac8.png)

2. Collect cluster wide execute time for each cost elements, using `conversion_rate` to convert into `computye unit`

- signature cost: 
```
[...]

PR #20314, opened 2021-09-29
Implement fee schedule in cost calculation

taozhu-chicago:
#### Problem
fee schedule should be implemented in cost calculation.

#### Summary of Changes
- update const cost values with data collected by #19627
- update cost calculation to closely proposed fee schedule #16984

Fixes #


jackcmay:
Great stuff!  We want to switch the fee calculation from just sigs to using these calculated tx costs.  

What do you think about passing a `(tx, cost)` pair into the bank batches?  The bank can then use a fee governor to calculate the final tx fee for each transaction?

We also need to make these fee calculations available for transaction simulation.  Maybe we give banks a ref to the cost tracker?

PR #20317, opened 2021-09-29
adds unprefixed_malloc_on_supported_platforms to jemalloc

behzadnouri:
#### Problem
Without this feature jemalloc is used only for Rust code but not for
bundled C/C++ libraries (like rocksdb).
https://github.com/solana-labs/solana/issues/14366#issuecomment-930404992

#### Summary of Changes
added the feature


mvines (approved):
I guess this means we can't blame rocksdb for heap fragmentation

PR #20322, opened 2021-09-29
explorer: Display confirmed block's child slot

jstarry:
#### Problem
Hard to navigate to the descendant block due to skipped slots

#### Summary of Changes

Fixes #


jstarry:
I love that github actions finish before the travis job even triggers sometimes haha

PR #20327, opened 2021-09-29
AcctIdx: optimize should_remove_from_mem

jeffwashington:
#### Problem
We only want to look at the slot list if we think we may throw the item out.
#### Summary of Changes

Fixes #


brooksprumo (approved):
lgtm

PR #20328, opened 2021-09-29
AcctIdx: bucket perf improvements

jeffwashington:
#### Problem
Add ix to insert. This avoids a potentially unnecessary index calculation for each insert.
Add try_insert, which returns an error if the map needs to grow.
Add grow, which separately grows the map according to the error. This allows callers to release their own external locks during the time the map is growing. Perhaps later, we could use just a read lock on the bucket map while growing is taking place. This would allow read access/lookups in the map while it is growing. Growing can be very expensive.
#### Summary of Changes

Fixes #

### DIFF
--- a/bucket_map/src/bucket.rs
+++ b/bucket_map/src/bucket.rs
@@ -198,6 +198,12 @@ impl<T: Clone + Copy> Bucket<T> {
         data: &[T],
         ref_count: u64,
     ) -> Result<(), BucketMapError> {
+        let best_fit_bucket = IndexEntry::data_bucket_from_num_slots(data.len() as u64);
+        if self.data.get(best_fit_bucket as usize).is_none() {
+            // fail early if the data bucket we need doesn't exist - we don't want the index entry partially allocated
+            //error!("resizing because missing bucket");
+            return Err(BucketMapError::DataNoSpace((best_fit_bucket, 0)));
+        }
         let index_entry = self.find_entry_mut(key);
         let (elem, elem_ix) = match index_entry {
             None => {
@@ -213,11 +219,6 @@ impl<T: Clone + Copy> Bucket<T> {
             }
         };
         let elem_uid = self.index.uid(elem_ix);
-        let best_fit_bucket = IndexEntry::data_bucket_from_num_slots(data.len() as u64);
-        if self.data.get(best_fit_bucket as usize).is_none() {
-            //error!("resizing because missing bucket");
-            return Err(BucketMapError::DataNoSpace((best_fit_bucket, 0)));
-        }
         let bucket_ix = elem.data_bucket_ix();
         let current_bucket = &self.data[bucket_ix as usize];
         if best_fit_bucket == bucket_ix && elem.num_slots > 0 {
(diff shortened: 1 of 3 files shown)


brooksprumo (changes requested):
I'm not sure I understand the need for adding the param to `insert()` and `try_insert()`, but not to `update()`. The bucket_map API now feels inconsistent both with itself, and with HashMap (not saying it needs to be, but an observation).

Maybe there's a new fn in bucket_map to get the `Bucket` from the bucket ix, and then the caller has a `Bucket` object to call `insert` on. Would that work?

(If not, maybe new functions like `insert_in_bucket()` and `try_insert_in_bucket()` could work?)

bucket_map/src/bucket.rs
```diff
@@ -366,22 +367,30 @@ impl<T: Clone + Copy> Bucket<T> {
         //debug!(            "INDEX_IX: {:?} uid:{} loc: {} cap:{}",            key,            uid,            location,            index.capacity()        );
     }
 
+    /// if result is a grow error, then grow the appropriate piece and return true
+    pub fn grow_from_error(&mut self, result: Result<(), BucketMapError>) -> bool {
```
brooksprumo, line 371: I get the intent here, but it feels like this needs some massaging. Because even though the name is "grow from error", it doesn't _always_ grow, and the parameter _isn't_ an error, it's a result (which may not be an error).

Maybe this fn can be 
```fn grow(&mut self, err: BucketMapError) { ... }```
so that it'll always grow. Then in `insert()`, it could be (inside the `loop`):

```
match self.try_write(key, new, refct) {
    Ok(()) => return,
    Err(err) => self.grow(err),
}
```

File: bucket_map/src/bucket.rs [hunk 2]
```diff
@@ -198,6 +198,12 @@ impl<T: Clone + Copy> Bucket<T> {
         data: &[T],
         ref_count: u64,
     ) -> Result<(), BucketMapError> {
+        let best_fit_bucket = IndexEntry::data_bucket_from_num_slots(data.len() as u64);
+        if self.data.get(best_fit_bucket as usize).is_none() {
+            // fail early if the data bucket we need doesn't exist - we don't want the index entry partially allocated
+            //error!("resizing because missing bucket");
+            return Err(BucketMapError::DataNoSpace((best_fit_bucket, 0)));
+        }
```
brooksprumo, line 206: Nice!

brooksprumo (commented):
I must not have added these to my previous review...

bucket_map/src/bucket_map.rs
```diff
@@ -176,8 +180,24 @@ impl<T: Clone + Copy + Debug> BucketMap<T> {
...
+        value: (&[T], RefCount),
+    ) -> Result<(), BucketMapError> {
+        let mut bucket = self.get_bucket(ix);
+        bucket.as_mut().unwrap().try_write(key, value.0, value.1)
+    }
+
+    /// if result is a grow error, then grow the appropriate piece
+    pub fn grow_from_error(&self, ix: usize, result: Result<(), BucketMapError>) {
+        let mut bucket = self.get_bucket(ix);
+        bucket.as_mut().unwrap().grow_from_error(result);
```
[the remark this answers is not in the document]
jeffwashington: it will be used. It is part of the api.

bucket_map/src/bucket_map.rs
```diff
@@ -176,8 +180,24 @@ impl<T: Clone + Copy + Debug> BucketMap<T> {
                 Arc::clone(&self.stats),
             ));
         }
-        let bucket = bucket.as_mut().unwrap();
-        bucket.insert(key, value)
+        bucket
+    }
```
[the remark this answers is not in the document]
brooksprumo: Talked offline. Resolved.

bucket_map/src/bucket.rs
jeffwashington: > I'm not sure I understand the need for adding the param to `insert()` and `try_insert()`, but not to `update()`. The bucket_map API now feels inconsistent both with itself, and with HashMap (not saying it needs to be, but an observation).
> 
> Maybe there's a new fn in bucket_map to get the `Bucket` from the bucket ix, and then the caller has a `Bucket` object to call `insert` on. Would that work?
> 
> (If not, maybe new functions like `insert_in_bucket()` and `try_insert_in_bucket()` could work?)

yes, some inconsistencies. update is not used by any current callers anymore. Insert won't be used either soon (replaced by try_insert). Ultimately, we can add a self-ref struct which allows us a better apis behind RwLock

jeffwashington:
> I'm not sure I understand the need for adding the param to `insert()` and `try_insert()`, but not to `update()`. The bucket_map API now feels inconsistent both with itself, and with HashMap (not saying it needs to be, but an observation).
> 
> Maybe there's a new fn in bucket_map to get the `Bucket` from the bucket ix, and then the caller has a `Bucket` object to call `insert` on. Would that work?
> 
> (If not, maybe new functions like `insert_in_bucket()` and `try_insert_in_bucket()` could work?)

yes, comments later. Brooks and I discussed offline.

bucket_map/src/bucket.rs
jeffwashington: renamed grow

brooksprumo (approved):
lgtm

PR #20330, opened 2021-09-29
Use batch_send in forward_buffered_packets

jbiseda:
#### Problem
forward_buffered_packets in banking_stage sends packets individually inline on a banking thread.

#### Summary of Changes
reduce syscalls to send packets by using sendmmsg

Fixes #


sakridge (approved):
lgtm

PR #20333, opened 2021-09-29
AcctIdx: when disk index insert needs to grow, drop locks, grow, then restart flush

jeffwashington:
#### Summary of Changes

Fixes #


brooksprumo (approved):
nice, lgtm

PR #20335, opened 2021-09-29
explorer: Add epoch details page

jstarry:
#### Problem
Hard to discover when an epoch started or ended

#### Summary of Changes
- Added epoch details page which fetches the confirmed blocks at the start and end for easy navigation
- Linkified all places where we display an epoch

Screen Shot 2021-09-29 at 7 46 23 PM

Fixes #


jstarry:
@t-nelson any feedback on the info displayed?

oJshua (approved):

PR #20339, opened 2021-09-30
v1.6: Optimize reward collection at epoch boundaries

jstarry:
#### Summary of Changes

Fixes #


runtime/src/bank.rs
```diff
@@ -1907,85 +1962,108 @@ impl Bank {
...
-                        vote_pubkey, err
-                    );
-                    continue;
-                }
-            };
-            let commission = Some(vote_state.commission);
-
-            for (stake_pubkey, stake_account) in stake_group.iter_mut() {
-                // curry closure to add the contextual stake_pubkey
-                let mut reward_calc_tracer = reward_calc_tracer.as_mut().map(|outer| {
```
jstarry, line 1931: @ryoqun is it ok to take this out for v1.7 or should we keep it in?

aeyakovenko:
any reason to put this into 1.6 or 1.7?  seems like we can run it on 1.8 and give it some soak time.  I am a bit terrified at stake rewards calculations getting a bug.

jstarry:
> any reason to put this into 1.6 or 1.7? seems like we can run it on 1.8 and give it some soak time. I am a bit terrified at stake rewards calculations getting a bug.

It should go into v1.6 due to the forking and very expensive replay that happens on epoch boundaries. This issue is basically freezing replay stage processing for 12s per fork that crosses the epoch boundary which brings us close to not making progress at all.

aeyakovenko:
@jstarry got it.  looks fairly low risk, not touching the inflation calculator from what i can tell.  @ryoqun can you verify?

jstarry:
> @jstarry got it. looks fairly low risk, not touching the inflation calculator from what i can tell. @ryoqun can you verify?

Before merging definitely want @ryoqun's eyes and we will double check the bank hash before/after this change on a mainnet snapshot (which we pulled for benchmarking cc @t-nelson)

runtime/src/bank.rs
```diff
@@ -1907,85 +1962,108 @@ impl Bank {
...
-            }
-        }
-        self.rewards.write().unwrap().append(&mut rewards);
+                    ))
+                },
+            )
+            .collect();
+
+        self.rewards.write().unwrap().append(&mut vote_rewards);
+        self.rewards.write().unwrap().append(&mut stake_rewards);
```
sakridge, line 2066: just take the lock once?

upload-perf/src/upload-perf.rs
```diff
@@ -74,7 +74,6 @@ fn main() {
                         ("commit", git_commit_hash.trim().to_string(), String)
                     );
                     */
-
```
ryoqun, line 77: yeah, i also know this rustfmt differences between master,v1.6 rust tool versions... xD

ryoqun:
> > any reason to put this into 1.6 or 1.7? seems like we can run it on 1.8 and give it some soak time. I am a bit terrified at stake rewards calculations getting a bug.
> 
> It should go into v1.6 due to the forking and very expensive replay that happens on epoch boundaries. This issue is basically freezing replay stage processing for 12s per fork that crosses the epoch boundary which brings us close to not making progress at all.

I don't think the slow inflation rewards calc. problem needs to be fixed as soon as possible. in my opinion, stalls at epoch boundaries are bad indeed. but this doesn't kill the cluster. that's because every other forks at the stall are likewise required to process the heavy inflation calc. so, forkiness isn't severe unlike the bot-induced ones. so, there should be no death spiral loop trigger condition as well. also, past epoch boundaries were similar and the cluster didn't dead indeed.

That initial impression from the symptom was the reason i was rather late to come here... Anyway, I'm doing very in-depth review now.

But, we'd like to bake this with 2 or 3 epochs, if possible. Even single line plain bug or something could be counted as one of costliest one in the human history....

runtime/src/bank.rs
```diff
@@ -1815,57 +1825,90 @@ impl Bank {
     ///   ( Vec<(staker info)> (voter account) ) keyed by voter pubkey
     ///
     /// Filters out invalid pairs
-    fn stake_delegation_accounts(
```
ryoqun, line 1818: (yeah, i didn't like this getter-like name; albeit this heavy-lifting... xD)

runtime/src/bank.rs
ryoqun: yeah, reward_calc_tracer isn't critical at all. it's just my toy..

runtime/src/bank.rs
```diff
@@ -1907,85 +1958,114 @@ impl Bank {
...
+                let vote_state = Arc::new(vote_state);
+                delegations
+                    .into_iter()
+                    .map(move |delegation| (vote_pubkey, Arc::clone(&vote_state), delegation))
+            },
+        );
+
+        let mut stake_rewards = thread_pool.install(|| {
+            stake_delegation_iterator
+                .par_bridge()
```
ryoqun, line 1984: something new to me. :)

runtime/src/bank.rs
```diff
@@ -1907,85 +1958,114 @@ impl Bank {
...
-                            },
-                        ));
+                    delegations,
+                },
+            )| {
+                vote_account_rewards.insert(vote_pubkey, (vote_account, vote_state.commission, 0));
+                let vote_state = Arc::new(vote_state);
+                delegations
+                    .into_iter()
+                    .map(move |delegation| (vote_pubkey, Arc::clone(&vote_state), delegation))
```
ryoqun, line 1978: nit: can't just be `Arc::new(vote_state)` here without `Arc::clone()`?

runtime/src/bank.rs
```diff
@@ -1907,85 +1958,114 @@ impl Bank {
...
-                            lamports,
-                            post_balance,
-                            commission,
+                            lamports: vote_rewards as i64,
+                            post_balance: vote_account.lamports,
+                            commission: Some(commission),
                         },
-                    ));
-                }
-                self.store_account(&vote_pubkey, &vote_account);
```
ryoqun, line 1985: here

runtime/src/bank.rs
```diff
@@ -1907,85 +1958,114 @@ impl Bank {
...
+                    },
+                )
+                .collect()
+        });
+
+        let mut vote_rewards = vote_account_rewards
+            .into_iter()
+            .filter_map(
+                |(vote_pubkey, (mut vote_account, commission, vote_rewards))| {
+                    if vote_rewards == 0 {
```
ryoqun, line 2048: hmm, sadly, `voters_reward > 0` can't be true substitute for the old `vote_account_changed` flag.

this new condition would create differing bank hash (due to the fact certain kind of vote accounts are stored or no... so, this is not inflation rewards discrepancy per se), if i read the code correctly..

this is because previous code always `store_account()`s vote accounts even if (total) `vote_rewards` is 0: https://github.com/solana-labs/solana/pull/20339/files#r719652698

the new code doesn't `store_account()` vote accounts if (total) `vote_rewards` is 0.

jstarry: I checked that vote_account isn't modified besides lamport changes so I don't think we need the store account here. But it doesn't seem like it would hurt, I'll add it back

runtime/src/bank.rs
jstarry: I'm not sure what you mean, we can only capture `vote_state` once and once it's `Arc` wrapped we have to clone a reference for each flat map iteration

runtime/src/bank.rs
sakridge: The hashing will look at all accounts, so even if it's the same value, the hash will be different if the account update is present in the new slot or not.

ledger-tool/src/main.rs
```diff
@@ -2362,15 +2363,17 @@ fn main() {
                             skipped_reasons: String,
                         }
                         use solana_stake_program::stake_state::InflationPointCalculationEvent;
-                        let mut stake_calcuration_details: HashMap<Pubkey, CalculationDetail> =
-                            HashMap::new();
-                        let mut last_point_value = None;
+                        // TODO: Double-check this operation is thread-safe with Ryo
```
ryoqun, line 2366: this code looks thread-safe to me. :)

ledger-tool/src/main.rs
```diff
@@ -2505,17 +2509,21 @@ fn main() {
                             },
                         );
 
-                        let mut unchanged_accounts = stake_calcuration_details
-                            .keys()
-                            .collect::<HashSet<_>>()
+                        // Ensure DashMap lock is dropped
```
ryoqun, line 2512: nits: `// Ensure DashMap lock is dropped by copying the actual inner pubkeys out of it`

ledger-tool/src/main.rs
```diff
@@ -2505,17 +2509,21 @@ fn main() {
...
+                        let unchanged_accounts = stake_calculation_details
+                            .iter()
+                            .map(|v| *v.key())
+                            .collect::<HashSet<Pubkey>>();
+
+                        let mut unchanged_accounts = unchanged_accounts
```
ryoqun, line 2518: nit: `// Ensure DashMap lock is dropped by copying the actual inner pubkeys and account data out of it`

runtime/src/bank.rs
ryoqun: oh, never mind..

ryoqun:
1st round of review is done: so far only https://github.com/solana-labs/solana/pull/20339#discussion_r719653951 needs to be addressed. I need some sleep to fully LGTM. also, this pr is actively developed as of now. :)

i wonder how this is faster compared to the current implementation?

runtime/src/bank.rs
jstarry: Learned from @carllin!

Issue #19885, opened 2021-09-14
Figure out how a new node discovers and then downloads full+incremental snapshots


brooksprumo:

lcgogo:
@brooksprumo 

according to https://docs.google.com/document/d/1l4Prip29I-QxpSqSzxMwYFKLP78fqDLH_eNAjZzvVps/edit

after delete ledge folder and restart sol, met followed error

[2021-09-16T10:32:09.435394057Z INFO  solana_validator] Searching for an RPC service with shred version 8573 (Retrying: No snapshots available)...

PR #20356, opened 2021-09-30
Add RPC get_highest_incremental_snapshot_info()

### LINKED ISSUE #20357: Get highest incremental snapshot info over RPC

brooksprumo:
#### Problem

During bootstrap, a node gets snapshot hashes from CRDS. These snapshot hashes are for full snapshots only. After downloading a full snapshot (or if the node already has one), it needs to download an incremental snapshot. However, at the moment there's no way to discover information about incremental snapshots on other nodes.

#### Summary of Changes

Add an RPC fn for getting the highest incremental snapshot information.

For more information on the algorithm to be used during bootstrap—and how this new RPC fn will be used—see #19885 

Fixes #20357

### DIFF
--- a/client/src/mock_sender.rs
+++ b/client/src/mock_sender.rs
@@ -2,18 +2,8 @@
 
 use {
     crate::{
-        client_error::Result,
-        rpc_config::RpcBlockProductionConfig,
-        rpc_request::RpcRequest,
-        rpc_response::{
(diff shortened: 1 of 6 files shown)


brooksprumo:
Closing this PR. As per our discussion in Discord, I'll be changing the impl to use gossip instead of rpc for bootstrap.

PR #20359, opened 2021-09-30
AcctIdx: introduce BucketApi for access to a specific bucket

jeffwashington:
#### Summary of Changes

Fixes #

### DIFF
--- a/bucket_map/src/bucket_api.rs
+++ b/bucket_map/src/bucket_api.rs
@@ -1,0 +1,134 @@
+use crate::bucket::Bucket;
+use crate::bucket_item::BucketItem;
+use crate::bucket_map::BucketMapError;
+use crate::bucket_stats::BucketMapStats;
+use crate::{MaxSearch, RefCount};
+use solana_sdk::pubkey::Pubkey;
+use std::ops::RangeBounds;
+use std::path::PathBuf;
+
+use std::sync::Arc;
+use std::sync::{RwLock, RwLockWriteGuard};
+
+type LockedBucket<T> = Arc<RwLock<Option<Bucket<T>>>>;
+
+pub struct BucketApi<T: Clone + Copy> {
+    drives: Arc<Vec<PathBuf>>,
+    max_search: MaxSearch,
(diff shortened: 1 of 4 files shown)


jeffwashington:
@brooksprumo thoughts?

brooksprumo:
> @brooksprumo thoughts?

Overall, I definitely like it.

I don't know Rust well enough yet to be able to visualize and internalize all the different `Arc` and `Mutex`/`RwLock` combos happening here. That and the `'static` lifetime/bounds.

How fast do you want this to go in? If fast, then maybe a call is best to go over the code. Otherwise I can try reading over this multiple times and look up the Rust bits I'm not sure about. Happy to do what's most effective for you.

jeffwashington:
> > @brooksprumo thoughts?
> 
> Overall, I definitely like it.
> 
> I don't know Rust well enough yet to be able to visualize and internalize all the different `Arc` and `Mutex`/`RwLock` combos happening here. That and the `'static` lifetime/bounds.
> 
> How fast do you want this to go in? If fast, then maybe a call is best to go over the code. Otherwise I can try reading over this multiple times and look up the Rust bits I'm not sure about. Happy to do what's most effective for you.

``static` wasn't necessary any more. At one point I was using `self-referencing`.

bucket_map/src/bucket_api.rs
```diff
@@ -0,0 +1,134 @@
...
+use crate::bucket_stats::BucketMapStats;
+use crate::{MaxSearch, RefCount};
+use solana_sdk::pubkey::Pubkey;
+use std::ops::RangeBounds;
+use std::path::PathBuf;
+
+use std::sync::Arc;
+use std::sync::{RwLock, RwLockWriteGuard};
+
+type LockedBucket<T> = Arc<RwLock<Option<Bucket<T>>>>;
```
brooksprumo, line 13: As a TODO, we talked offline about removing the `Arc` from `LockedBucket` here. (For the next PR.)

brooksprumo (approved):
lgtm

PR #20363, opened 2021-10-01
v1.6: Optimize epoch boundary

jstarry:
#### Problem
- `Bank::stake_delegation_accounts` loads vote accounts once for each delegated stake which is very inefficient
- Deserialized vote and stake states are not reused
- Stake activation and reward calculation on epoch boundaries is done serially

#### Summary of Changes
- Don't reload vote accounts more than once in `Bank::stake_delegation_accounts`
- Cache deserialized vote and stake state during rewards collection
- Add thread pool to load accounts, calculate rewards, and write changes in parallel

Fixes #


jstarry:
Closing for now to get https://github.com/solana-labs/solana/pull/20432 into master and v1.7 first

Issue #18863, opened 2021-07-23
`solana-validator exit` feature request: force snapshot, delinquency ignore


nyetwurk:
#### Problem

* When doing solana-validator exit, sometimes we have to wait a long time for a conjunction of snapshot request and leader slot gap.
* There is also no way to simply tell the validator to take snapshot "right now", e.g. for orderly shutdown, or if we want a snapshot to serve to someone else "right now"
* There is no way to force an `exit` when the delinquency is too high
* Non-voting validators also should not care if there is delinquency

#### Proposed Solutions

* `exit` should trigger a snapshot as soon as there is a gap in the leader schedule (or shortly before one)
*  There should be a way to manually trigger a snapshot (regardless of exit)
*  There should be way to ignore delinquency
*  Non-voting validators should always ignore delinquency

ryoqun:
cc: @mvines

nyetwurk:
Also, because there is no way to trigger a manual snapshot, `exit` can get stuck in a loop waiting for a leader gap, then waiting to take a snapshot, then waiting for *another* leader gap because the snapshot took too long to trigger.

bji:
The first issue is largely addressed by incremental snapshots, which complete quickly and often enough that completing a snapshot shortly within a target window is the common case instead of the uncommon case.

The second issue I do not have experience with because I have never set min-idle-time "too low", not that I am aware of anyway.

The third issue is still an issue and would be a great feature enhancement.

The fourth issue I consider a non-issue, validators who are savvy enough to make correct decisions about whether or not to ignore cluster delinquency, can probably find fairly easy workarounds, and those validators not savvy enough to make correct decisions, are disallowed from making bad decisions.

The fifth issue I am not familiar with as I don't run an RPC server.

TL;DR: The largest problem, which is the difficulty of getting snapshots to complete in a timely enough manner within target windows, appears to largely be mitigated by incremental snapshots.

michaelh-laine:
> The fourth issue I consider a non-issue, validators who are savvy enough to make correct decisions about whether or not to ignore cluster delinquency, can probably find fairly easy workarounds, and those validators not savvy enough to make correct decisions, are disallowed from making bad decisions.


The 4th issue was fixed in https://github.com/solana-labs/solana/pull/20367 over 6 months ago

PR #20367, opened 2021-10-01
Ignore delinquent stake on exit

michaelh-laine:
#### Problem
https://github.com/solana-labs/solana/issues/18863 mentions some issues with the exit command, this PR addresses one of these issues only: ability to ignore delinquency when exiting

#### Summary of Changes
Add an --ignore delinquent-stake (-i) flag to the solana-validator exit and wait-for-restart-window sub commands which if passed will cause the wait_for_restart_window function to ignore delinquent stake when in a leader schedule hole.

### DIFF
--- a/validator/src/main.rs
+++ b/validator/src/main.rs
@@ -2351,6 +2351,8 @@ pub fn main() {
         ("wait-for-restart-window", Some(subcommand_matches)) => {
             let min_idle_time = value_t_or_exit!(subcommand_matches, "min_idle_time", usize);
             let identity = pubkey_of(subcommand_matches, "identity");
+            let ignore_delinquency = subcommand_matches.is_present("ignore_delinquency");
+            
             wait_for_restart_window(&ledger_path, identity, min_idle_time, ignore_delinquency).unwrap_or_else(|err| {
                 println!("{}", err);
                 exit(1);
--- a/validator/src/main.rs
+++ b/validator/src/main.rs
@@ -2300,9 +2300,7 @@ pub fn main() {
                 .unwrap_or(DEFAULT_MAX_DELINQUENT_STAKE_PERCENT);
 
             if max_delinquent_stake > 100 {
-                println!(
-                    "Maximum permitted delinquency can't exceed 100"
-                );
+                println!("Maximum permitted delinquency can't exceed 100");
                 exit(1);
             }
 
@@ -2373,9 +2371,7 @@ pub fn main() {
                 .unwrap_or(DEFAULT_MAX_DELINQUENT_STAKE_PERCENT);
 
             if max_delinquent_stake > 100 {
-                println!(
-                    "Maximum permitted delinquency can't exceed 100"
-                );
+                println!("Maximum permitted delinquency can't exceed 100");
                 exit(1);
             }
 
--- a/validator/src/main.rs
+++ b/validator/src/main.rs
@@ -154,7 +154,6 @@ fn wait_for_restart_window(
         "Maximum permitted delinquency: {}%",
         max_delinquency_percentage
     );
-    
 
     let mut current_epoch = None;
     let mut leader_schedule = VecDeque::new();
@@ -2295,8 +2294,8 @@ pub fn main() {
             let min_idle_time = value_t_or_exit!(subcommand_matches, "min_idle_time", usize);
             let force = subcommand_matches.is_present("force");
             let monitor = subcommand_matches.is_present("monitor");
-            let max_delinquent_stake = value_t_or_exit!(subcommand_matches, "max_delinquent_stake", u8);
-
+            let max_delinquent_stake =
+                value_t_or_exit!(subcommand_matches, "max_delinquent_stake", u8);
 
             if !force {
                 wait_for_restart_window(&ledger_path, None, min_idle_time, max_delinquent_stake)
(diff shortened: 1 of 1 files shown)


michaelh-laine:
@mvines can you have a look at this, it adds a flag to the solana-validator exit and wait-for-restart-window sub commands, the check " ./scripts/cargo-for-all-lock-files.sh stable check --locked --tests --bins --examples" works locally but seems to fail on the CI so not sure where to go from here, any tips?

validator/src/main.rs
```diff
@@ -2119,6 +2125,13 @@ pub fn main() {
                     .default_value("10")
                     .help("Minimum time that the validator should not be leader before restarting")
             )
+            .arg(
```
mvines, line 2128: How about adding a `--min-delinquency` argument that defaults to 5% instead?  Great idea overall though!

michaelh-laine: should be max though right? Even the current min_delinquency_percentage should probably be max, because this is the maximum delinquency to accept?

validator/src/main.rs
```diff
@@ -2126,11 +2131,12 @@ pub fn main() {
...
-                    .short("i")
-                    .long("ignore-delinquent-stake")
-                    .takes_value(false)
-                    .help("Ignore delinquent stake threshold when exiting")
+                Arg::with_name("max_delinquent_stake")
+                    .long("max-delinquent-stake")
+                    .takes_value(true)
+                    .validator(is_parsable::<u8>)
+                    .value_name("PERCENT")
+                    .help("The maximum delinquent stake % permitted for an exit [default: 5]")
```
mvines, line 2139:
```suggestion
                    .validator(solana_clap_utils:: input_validators::is_valid_percentage)
                    .default_value("5")
                    .value_name("PERCENT")
                    .help("The maximum delinquent stake % permitted for an exit")
```

validator/src/main.rs
```diff
@@ -105,6 +105,8 @@ const INCLUDE_KEY: &str = "account-index-include-key";
 const DEFAULT_MIN_SNAPSHOT_DOWNLOAD_SPEED: u64 = 10485760;
 // The maximum times of snapshot download abort and retry
 const MAX_SNAPSHOT_DOWNLOAD_ABORT: u32 = 5;
+// The maximum delinquent stake to permit for a safe validator exit/restart
+const DEFAULT_MAX_DELINQUENT_STAKE_PERCENT: u8 = 5;
```
mvines, line 109:
```suggestion
```

validator/src/main.rs
```diff
@@ -150,8 +151,12 @@ fn wait_for_restart_window(
...
         ),
     );
-    if ignore_delinquent_stake {
-        println!("{}", style("Ignoring delinquent stake").yellow());
+
+    if max_delinquency_percentage != DEFAULT_MAX_DELINQUENT_STAKE_PERCENT {
+        println!(
+            "Maximum permitted delinquency: {}%",
+            max_delinquency_percentage
+        );
```
mvines, line 159:
```suggestion
    println!(
        "Maximum permitted delinquency: {}%",
        max_delinquency_percentage
    );
```

michaelh-laine:
@mvines check it out, 

- changed to --max-delinquent-stake (no short flag as m is taken by monitor), 

- I couldn't find a way to validate the input range within the argument definition so I've had to use a clunky conditional (twice) further down to prevent a value >100 (though in reality maybe it doesn't even matter whether we validate for this, it doesn't change anything ultimately), 

- I also added DEFAULT_MAX_DELINQUENT_STAKE_PERCENT (as a u8 so it's not the previous decimal value of 0.05, to make user input easier, converting it at the time of comparison seemed easier (and in line with how the supermajority f64 change was done with the mainnet crash a few weeks ago IIRC)).

validator/src/main.rs
```diff
@@ -2289,10 +2296,18 @@ pub fn main() {
...
-            let ignore_delinquent_stake = subcommand_matches.is_present("ignore_delinquent_stake");
+            let max_delinquent_stake = value_t!(subcommand_matches, "max_delinquent_stake", u8)
+                .unwrap_or(DEFAULT_MAX_DELINQUENT_STAKE_PERCENT);
+
+            if max_delinquent_stake > 100 {
+                println!(
+                    "Maximum permitted delinquency can't exceed 100"
+                );
+                exit(1);
+            }
```
mvines:
```suggestion
            let max_delinquent_stake = value_t_or_exit!!(subcommand_matches, "max_delinquent_stake", u8);
```

mvines (commented):
lgtm, just made some comments about how we can do the clap.rs arg parsing a little cleaner

validator/src/main.rs
michaelh-laine: I didn't set the default and used value_t!() as this was used elsewhere (e.g. --tpu-coalesce-ms), I guess no need for the constant if it's not used anywhere else, let me adjust

validator/src/main.rs
```diff
@@ -2119,6 +2126,15 @@ pub fn main() {
...
                     .help("Minimum time that the validator should not be leader before restarting")
             )
+            .arg(
+                Arg::with_name("max_delinquent_stake")
+                    .long("max-delinquent-stake")
+                    .takes_value(true)
+                    .validator(solana_clap_utils:: input_validators::is_valid_percentage)
+                    .default_value("5")
+                    .value_name("PERCENT")
+                    .help("The maximum delinquent stake % permitted for an exit [default: 5]")
```
mvines, line 2136: You don't need the " [default: 5]" here, clap automatically adds it when the `default_value()` arg is present

validator/src/main.rs
```diff
@@ -2119,6 +2126,15 @@ pub fn main() {
                     .default_value("10")
                     .help("Minimum time that the validator should not be leader before restarting")
             )
+            .arg(
+                Arg::with_name("max_delinquent_stake")
+                    .long("max-delinquent-stake")
+                    .takes_value(true)
+                    .validator(solana_clap_utils:: input_validators::is_valid_percentage)
```
mvines, line 2133: Sorry, just throw `is_valid_percentage` in the global `use` at the top of the file and then this and the other `.validator()` reduce down to:
```suggestion
                    .validator(is_valid_percentage)
```

validator/src/main.rs
michaelh-laine: yeah missed that, adjusted now. Should I worry about the CI failures? looks like an RPC test timeout

mvines (approved):
Thanks for doing this! I'll usher this into master and the v1.7 branch for the next v1.7 release

PR #20372, opened 2021-10-01
Move bootstrap code into its own module

brooksprumo:
This PR moves the bootstrap logic into its own module.

The bootstrap logic inside validator/src/main.rs is quite large, which makes going through the non-bootstrap code a bit harder. I'll also be modifying the bootstrap logic to support incremental snapshots, so having easier-to-read bootstrap code will be helpful.

NOTE: Zero behavior has changed. In fact, I've tried to make zero text changes at all. There some `pub` attributes added to a fn and a struct+fields, but that's it.

### DIFF
--- a/validator/src/bootstrap.rs
+++ b/validator/src/bootstrap.rs
@@ -1,0 +1,715 @@
+use {
+    log::*,
+    rand::{seq::SliceRandom, thread_rng, Rng},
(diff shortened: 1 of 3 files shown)


mvines (approved):
nice cleanup, thanks

PR #20374, opened 2021-10-01
Add CrdsData::IncrementalSnapshotHashes

brooksprumo:
#### Summary of Changes

Fixes #


gossip/src/crds_value.rs
```diff
@@ -204,6 +206,39 @@ impl SnapshotHash {
         }
     }
 }
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, AbiExample)]
+pub struct IncrementalSnapshotHash {
+    pub base_slot: Slot,
+    pub snapshot_hash: SnapshotHash,
```
mvines, line 213: nit: I'd rather we inline `SnapshotHash` here for clarity.

brooksprumo: Sure thing, will do!

brooksprumo: Done!

gossip/src/crds_value.rs
```diff
@@ -204,6 +206,44 @@ impl SnapshotHash {
...
+    pub wallclock: u64,
+}
+
+impl Sanitize for IncrementalSnapshotHash {
+    fn sanitize(&self) -> Result<(), SanitizeError> {
+        sanitize_wallclock(self.wallclock)?;
+        if self.base_slot >= MAX_SLOT {
+            return Err(SanitizeError::ValueOutOfBounds);
+        }
+        for (slot, _) in &self.hashes {
```
mvines, line 224: Restricting `self.hashes.len()` to a sane value here seems nice (and I see we don't do this for `SnapshotHash` either!)

mvines (commented):
lgtm, but let's get Behzad's review too

gossip/src/crds_value.rs
brooksprumo: What would be a good value. I assume I should just use `MAX_SNAPSHOT_HASHES`, right?

gossip/src/crds_value.rs
```diff
@@ -169,6 +171,9 @@ pub struct SnapshotHash {
 impl Sanitize for SnapshotHash {
     fn sanitize(&self) -> Result<(), SanitizeError> {
         sanitize_wallclock(self.wallclock)?;
+        if self.hashes.len() > MAX_SNAPSHOT_HASHES {
+            return Err(SanitizeError::ValueOutOfBounds);
+        }
```
behzadnouri, line 176: `MAX_SNAPSHOT_HASHES` is _supposed to be_ the largest number of hashes that you can include in SnapshotHash and send it through push messages or pull responses while the whole thing fits into a packet:
https://github.com/solana-labs/solana/blob/d6fdc4b63/gossip/src/cluster_info.rs#L3170-L3194

But if you have already received a packet and successfully deserialized it, then there is no point in checking the number of hashes in the sanitize here.
Regardless, this change is not related to this commit, so please remove it.

gossip/src/crds_value.rs
```diff
@@ -204,6 +209,47 @@ impl SnapshotHash {
         }
     }
 }
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, AbiExample)]
+pub struct IncrementalSnapshotHash {
+    pub from: Pubkey,
+    pub base_slot: Slot,
```
behzadnouri, line 216: imo, `_slot` in `base_slot` is redundant. The type already says it is `: Slot,`.
same way that we do not say:
```rust
pub from_pubkey: Pubkey,
```

gossip/src/crds_value.rs
```diff
@@ -91,6 +91,7 @@ pub enum CrdsData {
     Version(Version),
     NodeInstance(NodeInstance),
     DuplicateShred(DuplicateShredIndex, DuplicateShred),
+    IncrementalSnapshotHashes(IncrementalSnapshotHash),
```
behzadnouri, line 94: Why one is `IncrementalSnapshotHashes`, the other is `IncrementalSnapshotHash`?

If you are following `SnapshotHashes`, I don't think that is a good example to follow.

gossip/src/crds_value.rs
```diff
@@ -204,6 +209,47 @@ impl SnapshotHash {
...
+
+impl Sanitize for IncrementalSnapshotHash {
+    fn sanitize(&self) -> Result<(), SanitizeError> {
+        sanitize_wallclock(self.wallclock)?;
+        if self.base_slot >= MAX_SLOT {
+            return Err(SanitizeError::ValueOutOfBounds);
+        }
+        if self.hashes.len() > MAX_SNAPSHOT_HASHES {
+            return Err(SanitizeError::ValueOutOfBounds);
+        }
```
behzadnouri, line 229: If you want to add such a check here, you need to find the right max value for this new type;
and some tests that the max value is correct. similar to:
https://github.com/solana-labs/solana/blob/d6fdc4b63/gossip/src/cluster_info.rs#L3170-L3194

gossip/src/crds_value.rs
```diff
@@ -204,6 +209,47 @@ impl SnapshotHash {
...
+    }
+}
+
+impl IncrementalSnapshotHash {
+    pub fn new(from: Pubkey, base_slot: Slot, hashes: Vec<(Slot, Hash)>) -> Self {
+        Self {
+            from,
+            base_slot,
+            hashes,
+            wallclock: timestamp(),
```
behzadnouri, line 248: I would rather take `wallclock` as an argument rather than pulling it here. e.g. it would be more suitable for testing.

gossip/src/crds_value.rs
```diff
@@ -204,6 +209,47 @@ impl SnapshotHash {
         }
     }
 }
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, AbiExample)]
+pub struct IncrementalSnapshotHash {
+    pub from: Pubkey,
+    pub base_slot: Slot,
+    pub hashes: Vec<(Slot, Hash)>,
+    pub wallclock: u64,
```
behzadnouri, line 218: please remove `pub` from fields unless totally needed, or at leas use `pub(crate)`.

gossip/src/crds_value.rs
```diff
@@ -204,6 +209,47 @@ impl SnapshotHash {
         }
     }
 }
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, AbiExample)]
+pub struct IncrementalSnapshotHash {
```
behzadnouri, line 214: I am also wondering what exactly are we saving here by not including the base slot hash here, ie.
```rust
base: (Slot, Hash),
```

gossip/src/crds_value.rs
brooksprumo: Done!

gossip/src/crds_value.rs
brooksprumo: Done!

gossip/src/crds_value.rs
brooksprumo: Resolved here: https://github.com/solana-labs/solana/pull/20374#discussion_r721397268

gossip/src/crds_value.rs
brooksprumo: Done!

gossip/src/crds_value.rs
brooksprumo: Done! I've removed the `new()` fn, since every field is passed directly to the struct instance.

gossip/src/crds_value.rs
brooksprumo: Done. I've renamed the struct to `IncrementalSnapshotHashes`.

gossip/src/crds_value.rs
brooksprumo: Yeah, this was something I went back and forth on a few times.

When creating the actual incremental snapshot archive (which is where the snapshot hashes are pushed to CRDS), the incremental snapshot package does not know the full snapshot's hash, just the slot. So, if we want to have the base Hash here too, then either (1) the base (full) snapshot hash needs to be looked up* when pushing to CRDS, or (2) the incremental snapshot package needs to have the full snapshot hash added to it (which would need changes in AccountsBackgroundService to keep that info and put it into the incremental snapshot package).

Both of those options are possible, but the full snapshot hash isn't needed here, since we also have the `SnapshotHashes` for full snapshots.

*To handle the look up, this is somewhat tricky. We could look up the full snapshot archives on disk and then find the one with the matching slot and get its hash. I had an impl a while back that did this for a different reason, but design feedback was to avoid going to disk to do the lookup, for correctness and performance.

Since bootstrap doesn't need `IncrementalSnapshotHashes` to have the base hash, and the Incremental Snapshot in concept doesn't need the base hash, I didn't add it. If later there's a new unified snapshot hashes type for both full and incremental snapshots, then I think we'd get the base slot and hash for free.

I understand that putting the base hash here could be used during bootstrap, but I don't know of a reason that the base hash _must_ be here. Is there a use-case that you have in mind that would need the base hash here?

behzadnouri: I think the benefit would be that it will make the struct self-contained; and does not introduce cross-dependency between crds values.
It shouldn't be too much work for whatever the code which receives base-slot to also do the book-keeping for the snapshot hash for that slot. no?

brooksprumo: Done! I've updated the struct to also include the hash for the base as well.

mvines: We're good with only supporting one full snapshot that incrementals can be based on?  I can't think of anything immediately problematic with this.  Just pointing out that we're about to design away that possibility.

brooksprumo: Yeah, creating the incremental snapshot requires specifying what (single) slot it is based on. So the impl would have to change quite a lot to support something else.

brooksprumo: Oh, I think I misinterpreted the question. Yeah I see what you mean. I originally had _every_ IncrementalSnapshotHash also store its full snapshot (i.e. base) slot. We could do that, but it just means we consume more bandwidth.

brooksprumo: That, or it could be a HashMap of base slot to everything. I.e.

```struct IncrementalSnapshotHahes {
    from: Pubkey,
    hashes: HashMap<Slot, (Hash, (Slot, Hash)),
    wallclock: u64,
}
```

behzadnouri (approved):

PR #20379, opened 2021-10-01
Don't gossip more shapshot hashes than what we retain

brooksprumo:
#### Problem

SnapshotPackagerService is responsible for pushing snapshot hashes to the cluster. It pushes at most `MAX_SNAPSHOT_HASHES`. But, the node may be configured to only retain `X` snapshots, where `X < MAX_SNAPSHOT_HASHES`. In that case, we should only push out a max of `X` snapshot hashes.

#### Summary of Changes

The maximum number of snapshot hashes to push should be `min(MAX_SNAPSHOT_HASHES, snapshots_to_retain)`.

### DIFF
--- a/core/src/snapshot_packager_service.rs
+++ b/core/src/snapshot_packager_service.rs
@@ -29,6 +29,10 @@ impl SnapshotPackagerService {
     ) -> Self {
         let exit = exit.clone();
         let cluster_info = cluster_info.clone();
+        let max_snapshot_hashes = std::cmp::min(
+            MAX_SNAPSHOT_HASHES,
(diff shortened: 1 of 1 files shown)


brooksprumo:
@mvines @behzadnouri @sakridge Here's a PR to fix the number of snapshot hashes pushed to the cluster by SnapshotPackagerService. We talked about it in Discord [here](https://discord.com/channels/428295358100013066/478692221441409024/893258774264053831). Hopefully the problem description provides enough information. Let me know if there are any questions/issues. Thanks!

mvines (approved):

PR #20392, opened 2021-10-03
Add struct and convenience methods to track stake activation status

jstarry:
#### Problem
Tuples are not very dev-friendly

#### Summary of Changes
- Add `StakeActivationStatus` (type alias to StakeHistoryEntry) to represent `(effective, activating, deactivating)` tuple
- Implement `Add` operation for `StakeHistoryEntry` for easy summation
- Add some convenience constructors to `StakeHistoryEntry`

Fixes #

### DIFF
--- a/programs/stake/src/stake_state.rs
+++ b/programs/stake/src/stake_state.rs
@@ -1757,19 +1757,19 @@ mod tests {
         // assert that this stake follows step function if there's no history
         assert_eq!(
             stake.stake_activating_and_deactivating(stake.activation_epoch, Some(&stake_history),),
-            StakeActivationStatus::activating(0, stake.stake),
+            StakeActivationStatus::with_effective_and_activating(0, stake.stake),
         );
         for epoch in stake.activation_epoch + 1..stake.deactivation_epoch {
             assert_eq!(
                 stake.stake_activating_and_deactivating(epoch, Some(&stake_history)),
-                StakeActivationStatus::fully_active(stake.stake),
+                StakeActivationStatus::with_effective(stake.stake),
             );
         }
         // assert that this stake is full deactivating
         assert_eq!(
             stake
                 .stake_activating_and_deactivating(stake.deactivation_epoch, Some(&stake_history),),
-            StakeActivationStatus::deactivating(stake.stake),
+            StakeActivationStatus::with_deactivating(stake.stake),
         );
         // assert that this stake is fully deactivated if there's no history
         assert_eq!(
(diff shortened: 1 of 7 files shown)


sdk/program/src/stake/state.rs
```diff
@@ -342,28 +344,26 @@ impl Delegation {
...
         let (effective_stake, activating_stake) = self.stake_and_activating(target_epoch, history);
 
         // then de-activate some portion if necessary
         if target_epoch < self.deactivation_epoch {
             // not deactivated
-            (effective_stake, activating_stake, 0)
+            StakeActivationStatus::activating(effective_stake, activating_stake)
         } else if target_epoch == self.deactivation_epoch {
             // can only deactivate what's activated
-            (effective_stake, 0, effective_stake.min(delegated_stake))
```
joncinque, line 366: Getting rid of the `min` is a big win.  I double-checked the logic to be sure, and it seems like there's no need for it.

programs/stake/src/stake_state.rs
```diff
@@ -1838,7 +1824,8 @@ mod tests {
                 stake.deactivation_epoch + 1,
                 Some(&stake_history),
             ),
-            (stake.stake, 0, stake.stake) // says "I'm still waiting for deactivation"
+            // says "I'm still waiting for deactivation"
```
joncinque, line 1827: nano-nit: with your clarifications, no need for this comment!

sdk/program/src/stake_history.rs
```diff
@@ -15,6 +15,42 @@ pub struct StakeHistoryEntry {
...
 
+impl StakeHistoryEntry {
+    pub fn fully_active(effective: u64) -> Self {
+        Self {
+            effective,
+            ..Self::default()
+        }
+    }
+
+    pub fn activating(effective: u64, activating: u64) -> Self {
```
joncinque, line 26: nit: up to you, but I wonder if it would be clearer to call this `effective_and_activating`, or if just calling it `activating` clearly implies that there may be some fully active.

joncinque (approved):
Just some nits, great job!

sdk/program/src/stake/state.rs
jstarry: Thanks for double checking!

solana-grimes:

sdk/program/src/stake_history.rs
jstarry: yeah I wanted to shrug this off but I think your intuition about this not being very clear is correct. I changed the method names to `with_*` and so I have a `with_effective_and_activating` method which is pretty clear now.

PR #20397, opened 2021-10-04
AcctIdx: bucket map grows with read lock

jeffwashington:
#### Summary of Changes
write lock insert/update can require a grow.
Grow can now take place with a read lock. The grow uses a mutex to mark the bucket as grown.
The next write lock can check to see if there is a grown bucket available and switch to the grown bucket before returning the write lock
Fixes #


brooksprumo (approved):
Looks good. If you end up making other commits on the PR, we talked about adding a `pow2` on the capacity. But no worries if that doesn't happen here.

PR #20401, opened 2021-10-04
Explain what `solana-stake-accounts new` does

tredondo:
#### Problem
https://docs.solana.com/cli/manage-stake-accounts is unclear. The authors didn't put themselves in the shoes of a noob.

#### Summary of Changes
Try to add explanation. Might wrong. Please fix.


CriesofCarrots:
I don't think this is quite right. The new stake account is at an address derived from BASE_KEYPAIR and an index (`0` if cli arg not provided). @jstarry

jstarry:
> I don't think this is quite right. The new stake account is at an address derived from BASE_KEYPAIR and an index (`0` if cli arg not provided). @jstarry 

Thanks, I'll follow up with a correction

Issue #20370, opened 2021-10-01
GetConfirmedBlock sometimes returns -32004 on blocks that should exist


leoluk:
Wormhole first requests the current slot height with commitment level finalized: 

> 2021-10-01T12:17:53.508Z	INFO	wormhole.root.solwatch-finalized	solana/client.go:176	fetched current Solana height	{"commitment": "finalized", "slot": 99245280, "lastSlot": 99245278, "pendingSlots": 2, "took": 0.00160544}

It then calculates a range of slots to request:

> 2021-10-01T12:17:53.508Z	INFO	wormhole.root.solwatch-finalized	solana/client.go:186	fetching slots in range	{"from": 99245279, "to": 99245280, "took": 0.00169402, "commitment": "finalized"}

At this point, we assume that all blocks in this range are either finalized, or were skipped.

However, in some cases the first request fails:

> 2021-10-01T12:17:53.509Z	INFO	wormhole.root.solwatch-finalized	solana/client.go:252	empty slot	{"slot": 99245280, "code": -32004, "commitment": "finalized"}

... but a retry after 1m works:

> 2021-10-01T12:18:54.135Z	INFO	wormhole.root.solwatch-finalized	solana/client.go:281	fetched block	{"slot": 99245280, "num_tx": 2670, "took": 0.625404706, "commitment": "finalized"}
2021-10-01T12:18:54.135Z	ERROR	wormhole.root.solwatch-finalized	solana/client.go:365	SOLANA BUG: skipped or unavailable block retrieved on second attempt	{"slot": 99245280, "commitment": "finalized"}

This has been causing Wormhole consensus issues, since we assumed that if the getBlocks call returns -30004 or -30007, then the block was skipped. This is true most of the time, but not always.

CriesofCarrots:
@leoluk , can you provide an update here? Is this still occurring? Are you now seeing error -32014 instead?

leoluk:
Still happening as of 1.8.5 with finalized:

```
2021-11-23T19:45:14.228Z	INFO	wormhole.root.solwatch-finalized	solana/client.go:294	fetched block	{"slot": 108366349, "num_tx": 1117, "took": 0.625026025, "commitment": "finalized"}
2021-11-23T19:45:13.603Z	DEBUG	wormhole.root.solwatch-finalized	solana/client.go:237	requesting block	{"slot": 108366349, "commitment": "finalized", "empty_retry": 1}
[...]

PR #20407, opened 2021-10-04
Different error if block is not available

sakridge:
#### Problem

If the block is queried before transaction status service can complete the write, then it could return an error which indicates the block is not available at all.

#### Summary of Changes

Return an error which just indicates the transaction status has not been written yet.

Fixes #


rpc/src/rpc.rs
```diff
@@ -1001,30 +1001,34 @@ impl JsonRpcRequestProcessor {
             } else if commitment.is_confirmed() {
                 // Check if block is confirmed
                 let confirmed_bank = self.bank(Some(CommitmentConfig::confirmed()));
-                if confirmed_bank.status_cache_ancestors().contains(&slot)
-                    && slot
+                if confirmed_bank.status_cache_ancestors().contains(&slot) {
+                    if slot
                         <= self
                             .max_complete_transaction_status_slot
                             .load(Ordering::SeqCst)
```
jstarry, line 1008: I don't really get this check. Isn't it possible that we froze a bank on a higher fork and the requested slot is still not frozen/available?

sakridge: This transaction status writing happens asynchronously by a worker thread, so a block can be frozen or rooted and the status is still not written. In that case, we cannot construct the return data for the RPC call.

sakridge: A more involved and better fix could be to go look up the status from the in-memory state about the block if it's not written to blockstore yet.

jstarry: Got it, well this change already seems like a good improvement, but seems like we need the more involved fix (later) to fully close out the linked issue

CriesofCarrots:
Agreed, this change looks good to me for starters.
I'm wondering if this issue is related: https://github.com/solana-labs/solana/issues/20370 If the transaction-status queue is getting backed up enough to affect finalized blocks, we may want similar logic to return this error in the finalized path.

rpc/src/rpc.rs
sakridge: Yea. I'll leave it open, just link for now.

sakridge:
> Edit: Oh, I see you've linked that issue already. Not sure this fix will resolve that issue with `finalized` blocks. The rpc module assumes that if a block is old enough to be finalized, all the various metadata caching will be complete.

Are there other checks that you know of? Or case where the block wouldn't be ready by the time the slot is rooted?

macotouenaca (commented):
rpc/src/rpc.rs

PR #20410, opened 2021-10-04
Add clean metrics

jeffwashington:
#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/accounts_db.rs
+++ b/runtime/src/accounts_db.rs
@@ -1182,6 +1182,8 @@ struct LatestAccountsIndexRootsStats {
     roots_range: AtomicU64,
     rooted_cleaned_count: AtomicUsize,
     unrooted_cleaned_count: AtomicUsize,
+    clean_unref_from_storage_us: AtomicU64,
+    clean_dead_slot_us: AtomicU64,
 }
 
 impl LatestAccountsIndexRootsStats {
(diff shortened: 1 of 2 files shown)


runtime/src/accounts_db.rs
```diff
@@ -5827,6 +5854,8 @@ impl AccountsDb {
                 *slot
             })
             .collect();
+        measure.stop();
+        accounts_index_root_stats.clean_dead_slot_us += measure.as_us();
```
carllin, line 5858: Unrelated to this PR, I noticed an existing bug I think while reviewing this 😃  

This code path `clean_dead_slots_from_accounts_index()` is potentially executed  by both `clean_accounts()` from clean, but also by `purge_slots()` from Bank::drop(). However, both paths will add to the `self.clean_accounts_stats` below.

The way we get around this for the other functions that *eventually* call down to this path is by passing in the correct purge stats to use, either `self.external_purge_slots_stats` in `purge_slots()`, or `self.clean_accounts_stats.purge_stats` in clean_accounts(). For example, https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_db.rs#L2115 and https://github.com/solana-labs/solana/blob/master/runtime/src/accounts_db.rs#L3980

The bug here is that these `purge_stats` are not passed as an option to `remove_dead_slots_metadata()` or any of its subsequent function calls like this call to  `clean_dead_slots_from_accounts_index`  here.

carllin (approved):

runtime/src/accounts_db.rs
jeffwashington: will look into this

PR #20419, opened 2021-10-04
Accountsdb stream plugin improvement

lijunwangs:
#### Summary of Changes

Support using connection pooling and use multiple threads to do Postgres db operations. The performance is improved from 1500 RPS to 40,000 RPS measured during validator start.

Support multiple plugins at the same time.

Fixes #


accountsdb-plugin-postgres/src/accountsdb_plugin_postgres.rs
```diff
@@ -1,27 +1,30 @@
...
-    std::{fs::File, io::Read, sync::Mutex},
+    std::{fs::File, io::Read},
     thiserror::Error,
 };
 
-struct PostgresSqlClientWrapper {
-    client: Client,
-    update_account_stmt: Statement,
+enum PostgresClientEnum {
+    Simple(SimplePostgresClient),
```
carllin, line 21: Can this `SimplePostgresClient` this just be replaced with a `ParallelPostgresClient` of only 1 worker thread?

accountsdb-plugin-postgres/src/postgres_client.rs
```diff
@@ -0,0 +1,447 @@
...
+    }
+
+    fn update_account<T: ReadableAccountInfo>(
+        &mut self,
+        account: &T,
+        slot: u64,
+    ) -> Result<(), AccountsDbPluginError> {
+        if let Err(err) = self
+            .sender
+            .send(DbWorkItem::UpdateAccount(UpdateAccountRequest {
```
carllin, line 398: Might be good to add a comment here that because sender is a bounded channel, this will block if there is a large queue of pending work.

This means we should ensure in future changes that any calls to `notify_account_update() -> notify_plugins_of_account_update() -> update_account()` from AccountsDb doesn't block `AccountsDb::store_cached()`.

accountsdb-plugin-postgres/src/accountsdb_plugin_postgres.rs
lijunwangs: There is a subtle but important difference. Using the SimplePostgresClient -- the account info input is not cloned because it does not need to be sent over to different threads. It would be more performant that the single threaded ParallelPostgresClient which still requires cloning.

accountsdb-plugin-postgres/src/postgres_client.rs
lijunwangs: okay

accountsdb-plugin-postgres/src/postgres_client.rs
```diff
@@ -0,0 +1,447 @@
...
+
+        Ok(Self {
+            workers,
+            exit_worker,
+            sender,
+        })
+    }
+}
+
+impl PostgresClient for ParallelPostgresClient {
```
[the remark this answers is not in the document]
lijunwangs: Will look into how to insulate the Postgres DB part which is not unit testable.

